PR #2979, opened 2018-10-26
Removed hadoop modules derived from driverjar.gradle which have effectively no added value

pkozelka:
The flag `notYarn` is no longer used, as @michalkurka confirmed.

That was at the start of this PR. But as I continued to eliminate stuff that was no longer necessary after being unlocked by this, I ended up dropping the `driverjar.gradle` and half of the hadoop modules.


Pscheidl:
I approved the "cleanup", not tested it.

PR #2980, opened 2018-10-26
Adds name parameter to Python and R init methods

mdymczyk:
Adds the option to specify `cloud_name` in the `h2o.init()` methods in Python and R. Main motivation is because we are using the `-name` parameter in MLI and would like to move from bootstrapping via `java -jar h2o.jar` to `h2o.init()` calls.

Proposed logic:

* If there is an instance under given address:port with the same cloud name - connect as normal
* If there is an instance with a different cloud name - fail in case of R, in case of Python fail if port passed explicitly and as_port=True passed, otherwise try to bootstrap on consecutive ports
* If no instance then try to start an H2O-3 instance using name as cloud name.

Added also `as_port` to the `kwargs` so the `port` argument acts like the actual `-port` parameter instead of `-base_port`. 

Comments?

I also would like to add the `log_dir` parameter if this one looks good in the next PR.

### DIFF
--- a/h2o-py/h2o/backend/connection.py
+++ b/h2o-py/h2o/backend/connection.py
@@ -70,7 +70,7 @@ class H2OConnectionConf(object):
     """List of allowed property names exposed by this class"""
     allowed_properties = ["ip", "port", "https", "context_path", "verify_ssl_certificates",
                           "proxy", "auth", "cookies", "verbose"]
-    
+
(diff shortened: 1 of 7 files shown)


mdymczyk:
@michalkurka @Pscheidl now that H2O World is done could I get a quick review of this PR so I can merge if there are no objections?

PR #2985, opened 2018-10-29
PUBDEV-5705: added skipped column to csv parsers

wendycwong:
This PR completes part of the work for JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-5705?filter=myopenissues.

I have done the following:
1. added skipped_column support for csv parser;
2. added client interface in R and Python;
3. added R/Python tests.

The most heavy duty work is in CsvParser.java.

=======================================================================
Tips for code review:

To enable skipping columns during parse, introduced two parameters to ParseSetup.java

_skipped_columns (int[] with inputs from user listing column indices to skip)
_parse_columns_indices (int[] representing columns indices to be included in the final frame). Size of this array is number of columns in file minus number of skipped columns.
Added one parameter to parser.java
boolean[] _keepColumns with size = number of columns in file. For skipped columns indices, its value is false.

During guessSetup or when users did not specify skipped_columns, _parse_columns_indices will contain all column indices and _keepColumns will contain all True. Remember that the various parsers associated with the different parsers will be called once to attempt to parse the file with that parser type. Hence, whatever change I made MUST NOT affect this process. By setting up the _parse_columns_indices and _keepColumns the way I did will guarantee this process will run smoothly.

I have not changed anything associate with reading the input file because all the columns must be read even though not all of them will be included in the final parse file.

The major change is in method makedout in ParseDataset.java. Inside method FVecParseWriter of FVecParseWriter.java, the following parameters will have dimension equals to the parse_column_indices: _ctypes, _categoricals, _vec, _nvs.

There are special cases to setting them when they are called from gzip files.

The stage is now set for parseChunk method. Basically, we use two indices to refer to the columns. One is to index through all the columns in the input file, the other one refers only to the columns that are to be written to dout.

After parsing and during reconciliation of categorical levels, the same thing apply. We only look at column indices that are inside parse_column_indices even though the Categorical variable contains all the column indices. Only the columns that are being parsed will have non-empty hashmap representing the categorical value and integer levels.

Hope this helps, Wendy

### DIFF
--- a/h2o-core/src/main/java/water/parser/CsvParser.java
+++ b/h2o-core/src/main/java/water/parser/CsvParser.java
@@ -11,7 +11,6 @@ import java.io.ByteArrayOutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 
-import static water.parser.DefaultParserProviders.ARFF_INFO;
 import static water.parser.DefaultParserProviders.CSV_INFO;
 
 class CsvParser extends Parser {
@@ -59,7 +58,8 @@ class CsvParser extends Parser {
     boolean decimal = false;
     int fractionDigits = 0;
     int tokenStart = 0; // used for numeric token to backtrace if not successful
-    int colIdx = 0;
+    int columnCounter = 0;  // index into parsed columns only, exclude skipped columns.
+    int colIdx = 0; // count each actual column in the dataset including the skipped columns
     byte c = bits[offset];
     // skip comments for the first chunk (or if not a chunk)
     byte[] nonDataLineMarkers = nonDataLineMarkers();
@@ -80,10 +80,18 @@ class CsvParser extends Parser {
     dout.newLine();
 
     final boolean forceable = dout instanceof FVecParseWriter && ((FVecParseWriter)dout)._ctypes != null && _setup._column_types != null;
+    int colIndexNum = _keepColumns.length-1;
+
+    if (_setup._parse_columns_indices==null) {  // _parse_columns_indices not properly set
+      _setup.setParseColumnIndices(_setup.getNumberColumns(), _setup.getSkippedColumns());
+    }
+    int parseIndexNum = _setup._parse_columns_indices.length-1;
 MAIN_LOOP:
     while (true) {
-      final boolean forcedCategorical = forceable && colIdx < _setup._column_types.length && _setup._column_types[colIdx] == Vec.T_CAT;
-      final boolean forcedString = forceable && colIdx < _setup._column_types.length && _setup._column_types[colIdx] == Vec.T_STR;
+      final boolean forcedCategorical = forceable && colIdx < _setup._column_types.length &&
+              _setup._column_types[_setup._parse_columns_indices[columnCounter]] == Vec.T_CAT;
+      final boolean forcedString = forceable && colIdx  < _setup._column_types.length &&
+              _setup._column_types[_setup._parse_columns_indices[columnCounter]] == Vec.T_STR;
 
       switch (state) {
         // ---------------------------------------------------------------------
@@ -149,21 +157,24 @@ MAIN_LOOP:
             str.addBuff(bits);
           }
           if( !isNa &&
-              _setup.isNA(colIdx, str.toBufferedString())) {
+              _setup.isNA(columnCounter, str.toBufferedString())) {
             isNa = true;
           }
-          if (!isNa) {
-            dout.addStrCol(colIdx, str.toBufferedString());
+
+          if (!isNa && (colIdx <= colIndexNum) && _keepColumns[colIdx]) {
+            dout.addStrCol(columnCounter, str.toBufferedString());
             if (!isAllASCII)
-              dout.setIsAllASCII(colIdx, isAllASCII);
+              dout.setIsAllASCII(columnCounter, isAllASCII);
           } else {
-            dout.addInvalidCol(colIdx);
+            if ((colIdx <= colIndexNum) && _keepColumns[colIdx])
+              dout.addInvalidCol(columnCounter);
             isNa = false;
           }
           str.set(null, 0, 0);
           quotes = 0;
           isAllASCII = true;
-          ++colIdx;
+          if ((colIdx <= colIndexNum) &&  _keepColumns[colIdx++] && (columnCounter < parseIndexNum)) // only increment if not at the end
+            columnCounter++;
           state = SEPARATOR_OR_EOL;
           // fallthrough to SEPARATOR_OR_EOL
         // ---------------------------------------------------------------------
@@ -183,11 +194,13 @@ MAIN_LOOP:
           } else if (quoteCount > 2) {
             String err = "Unmatched quote char " + ((char) quotes);
             dout.invalidLine(new ParseWriter.ParseErr(err, cidx, dout.lineNum(), offset + din.getGlobalByteOffset()));
-            colIdx = 0;
+            columnCounter=0;
+            colIdx=0;
             quotes = 0;
           } else if (colIdx != 0) {
             dout.newLine();
-            colIdx = 0;
+            columnCounter = 0;
+            colIdx=0;
           }
           state = (c == CHAR_CR) ? EXPECT_COND_LF : POSSIBLE_EMPTY_LINE;
           if( !firstChunk )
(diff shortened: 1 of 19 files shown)


h2o-core/src/main/java/water/parser/Parser.java
```diff
@@ -66,10 +66,23 @@
...
-  protected Parser( ParseSetup setup, Key<Job> jobKey ) { _setup = setup;  CHAR_SEPARATOR = setup._separator; _jobKey = jobKey;}
+  protected Parser( ParseSetup setup, Key<Job> jobKey ) {
+    _setup = setup;  CHAR_SEPARATOR = setup._separator; _jobKey = jobKey;
+    if (_setup!=null && _setup._number_columns > 0) {
+      _keepColumns = new boolean[_setup._number_columns];
+      for (int colIdx = 0; colIdx < _setup._number_columns; colIdx++)
+        _keepColumns[colIdx] = true;
+      if (_setup._skipped_columns!=null) {
+        for (int colIdx : _setup._skipped_columns)
+          if (colIdx < _setup._number_columns)
```
Pscheidl: Silent failure may lead to successful parsing, but skipping invalid column IDs (those out of range) without telling the user might lead to unwanted results.

I'd throw an exception here. Our strategy everywhere should be to immediately notify the user if there is something wrong with the arguments provided.

h2o-core/src/main/java/water/parser/CsvParser.java
```diff
@@ -59,7 +58,8 @@
     boolean decimal = false;
     int fractionDigits = 0;
     int tokenStart = 0; // used for numeric token to backtrace if not successful
-    int colIdx = 0;
+    int columnCounter = 0;  // index into parsed columns only, exclude skipped columns.
```
Pscheidl: Please let's name this `parsedColumnCounter`. The difference between `colIdx` and `columnCounter` is not really obvious. Or just `parsedCols` or even `parsedCntr` is ok.

h2o-core/src/main/java/water/parser/CsvParser.java
```diff
@@ -149,21 +157,24 @@
...
           }
           if( !isNa &&
-              _setup.isNA(colIdx, str.toBufferedString())) {
+              _setup.isNA(columnCounter, str.toBufferedString())) {
             isNa = true;
           }
-          if (!isNa) {
-            dout.addStrCol(colIdx, str.toBufferedString());
+
+          if (!isNa && (colIdx <= colIndexNum) && _keepColumns[colIdx]) {
```
Pscheidl: I am not sure if the condition `(colIdx <= colIndexNum)` isn't in vain ? colIdx will always be at most the number of columns minus one in the original CSV file, as `int colIndexNum = _keepColumns.length-1;`.

This condition occurs on multiple lines.

h2o-core/src/main/java/water/parser/CsvParser.java
```diff
@@ -149,21 +157,24 @@
...
-            dout.addInvalidCol(colIdx);
+            if ((colIdx <= colIndexNum) && _keepColumns[colIdx])
+              dout.addInvalidCol(columnCounter);
             isNa = false;
           }
           str.set(null, 0, 0);
           quotes = 0;
           isAllASCII = true;
-          ++colIdx;
+          if ((colIdx <= colIndexNum) &&  _keepColumns[colIdx++] && (columnCounter < parseIndexNum)) // only increment if not at the end
```
Pscheidl: Incresing `colIdx` at the end of column `c` instead of increasing it at the beginning of column `c + 1` seems like a good change.

h2o-core/src/main/java/water/parser/FVecParseWriter.java
```diff
@@ -27,15 +26,60 @@
...
+          if (ctypes != null)       // happens with SVMlight
+            _ctypes[index] = ctypes[parse_columns_indices[index]];
+          if (categoricals != null)   // happens with SVMlight
+            _categoricals[index] = categoricals[parse_columns_indices[index]]; // categoricals calculated for all columns
+          _vecs[index] = ctypesShrunk?avs[index]:avs[parse_columns_indices[index]];
+        }
+      }
+      _parse_columns_indices = parse_columns_indices;
+      int num_parse_columns = parse_columns_indices.length;
+      _nvs = new NewChunk[num_parse_columns];
```
Pscheidl: May use `parse_columns_indices.length` directly. Not a big deal.

h2o-core/src/main/java/water/parser/FVecParseWriter.java
```diff
@@ -27,15 +26,60 @@
...
+      for (int i = 0; i < num_parse_columns; ++i) {
```
Pscheidl: May use parse_columns_indices.length directly. Not a big deal.

h2o-core/src/main/java/water/parser/Parser.java
wendycwong: > Good idea.  Done.

h2o-core/src/main/java/water/parser/CsvParser.java
wendycwong: Pavel, sooooooooo picky!  Done.

h2o-core/src/main/java/water/parser/CsvParser.java
wendycwong: Pavel:

I agreed with you at first.  However, after looking at this and checking out the runs in detail, this condition is essential.  Basically, during the run, the colIdx will be increased beyond colIndexNum if I do not put the condition in there.  It is intentionally designed by Tomas.

I used to have colIdx < colIndexNum but this creates the error of doubling the number of rows of certain files.  This test will fail:

Java 10 JUnit / water.parser.ParserTest.testParserRespectsSpecifiedColNum.

After checking the failure, I found the problem is that I need to let colIdx increment beyond its legal limit: colIdx == colIndexNum.  The reason for colIdx to increase beyond the actual number of columns is due to the finite state machine that Tomas has designed.  An extra row will be added if the condition colIdx < colIndexNum is satisfied.  However, this extra row will not be added if we have colIdx >= colIndexNum.

h2o-core/src/main/java/water/parser/CsvParser.java
wendycwong: Agree.

h2o-core/src/main/java/water/parser/FVecParseWriter.java
wendycwong: Was told from another code review that this provides small speedup?

h2o-core/src/main/java/water/parser/FVecParseWriter.java
wendycwong: Was told from another code review that this provides small speedup?

h2o-py/h2o/h2o.py
```diff
@@ -404,6 +409,8 @@ def import_file(path=None, destination_frame=None, parse=True, header=0, sep=Non
     assert_is_type(col_names, [str], None)
     assert_is_type(col_types, [coltype], {str: coltype}, None)
     assert_is_type(na_strings, [natype], {str: natype}, None)
+    assert (skipped_columns==None) or isinstance(skipped_columns, list), \
```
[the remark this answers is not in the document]
wendycwong: Thanks, you are right!  Changed.

h2o-py/h2o/h2o.py
```diff
@@ -572,13 +580,36 @@ def parse_setup(raw_frames, destination_frame=None, header=0, separator=None, co
     # TODO: clean up all this
```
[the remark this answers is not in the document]
wendycwong: Done. I left the other one in.

h2o-core/src/main/java/water/parser/FVecParseWriter.java
Pscheidl: A very tiny, small one. No need to allocate some memory on stack and then destroy it. A negligible one in this case. Doesn't matter really.

h2o-core/src/main/java/water/parser/CsvParser.java
Pscheidl: I see. Did not realize there is an extra row at the end. You're right Wendy.

wendycwong: Pavel:  As an experiment, you can change the condition from colIdx <= colIndexNum to colIdx < colIndexNum and see the doubling of the number of rows in that test.

h2o-core/src/main/java/water/parser/FVecParseWriter.java
wendycwong: Got it.

PR #2986, opened 2018-10-29
Remove pandas as a dependency for frame.py.

wendycwong:


h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -412,6 +412,9 @@ def _metrics_class(model_json):
 
     def convert_H2OXGBoostParams_2_XGBoostParams(self):
         '''
+        This method and convert_H2OFrame_2_DMatrix will require to import the following toolboxes:
+        xgboost, pandas, numpy and scipy.sparse.
+
```
angela0xdata: In order to use convert_H2OXGBoostParams_2_XGBoostParams and convert_H2OFrame_2_DMatrix, you must import the following toolboxes:

h2o-py/h2o/frame.py
```diff
@@ -2568,6 +2567,9 @@ def isax(self, num_words, max_cardinality, optimize_card=False, **kwargs):
 
     def convert_H2OFrame_2_DMatrix(self, predictors, yresp, h2oXGBoostModel):
         '''
+        This method will require to import the following toolboxes:
+        xgboost, pandas, numpy and scipy.sparse.
+
```
angela0xdata: This method requires that you import the following toolboxes:

h2o-py/h2o/estimators/estimator_base.py
wendycwong: Thanks.  Done

h2o-py/h2o/frame.py
wendycwong: Done

michalkurka:
@wendycwong looks good to me!

Can you please also make fix on `rel-xia`?

wendycwong:
michalk:  Will add it to rel-xia

PR #2998, opened 2018-11-01
Log dir and log level Python/R parameters

mdymczyk:
Adds the option to pass `log_dir` and `log_level` to both Python and R APIs during `h2o.init`. If connecting to an existing cluster/node they are skipped. No special validations performed assuming H2O-3's behaviour should take care of it all.

### DIFF
--- a/h2o-py/tests/testdir_apis/H2O_Module/pyunit_h2oinit.py
+++ b/h2o-py/tests/testdir_apis/H2O_Module/pyunit_h2oinit.py
@@ -4,7 +4,10 @@ sys.path.insert(1,"../../../")
 from tests import pyunit_utils
 import h2o
 from h2o.utils.typechecks import assert_is_type
-from h2o.exceptions import H2OConnectionError, H2OServerError
+from h2o.exceptions import H2OConnectionError, H2OServerError, H2OValueError
+import tempfile
+import shutil
+import os
 
 def h2oinit():
     """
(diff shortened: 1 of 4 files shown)


mdymczyk:
@michalkurka @Pscheidl similarly to #2980 adding a few other parameters I could use for mli-2.

h2o-py/tests/testdir_apis/H2O_Module/pyunit_h2oinit.py
```diff
@@ -75,10 +78,56 @@ def h2oinitname():
...
         assert_is_type(e, H2OConnectionError)
+        h2o.cluster().shutdown()
+
+
+def h2oinit_default_log_dir():
+    tmpdir = tempfile.mkdtemp()
+    try:
+        h2o.init(strict_version_check=False, name="default_log", ice_root=tmpdir)
+    except H2OConnectionError as e:  # some errors are okay like version mismatch
+        print("error message type is {0} and the error message is {1}\n".format(e.__class__.__name__, e.args[0]))
```
Pscheidl: I'd make these error messages more informative. Mention error during startup with custom log dir. My only reason is these are customer facing. The more informative our errors are, the more are users able to solve themselves.

Pscheidl:
@mdymczyk Is this just the client part ? As I can see no link to H2O backend. The code itself looks good and I have no problem approving it. The check for directory existence should be of course done on backend.

mdymczyk:
@Pscheidl yes only Client (Python and R) changes to expose the startup parameters (similarly to the `name` PR I just merged in today).

mdymczyk:
@Pscheidl just to make sure - this is ok to be merged?

PR #3006, opened 2018-11-02
Eliminate generating another extra jar which is not used for each h2o-hadoop

jakubhava:
We use just the shadow JAR. The `jar` provided by `java` plugin is generated and never used. Less artifacts and should also speed the build a bit

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ buildscript {
     //noinspection GroovyAssignabilityCheck
     dependencies {
         classpath 'org.ow2.asm:asm:5.1'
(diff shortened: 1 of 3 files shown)


jakubhava:
Feel free to merge if you're ok with this change after your review. I might be on holidays already

jakubhava:
I also upgraded shadow jar to 4.0.2 as that speeds up generating shadowJar of Hadoop from to 34 seconds from 100 seconds @michal-raska. Checked the files and they are identical @michal-raska

pkozelka:
Cool!

michal-raska:
Awesome!

PR #3010, opened 2018-11-02
PUBDEV-5705: add skip column support to ARFF parser.

wendycwong:
This PR is related to JIRA: https://0xdata.atlassian.net/browse/PUBDEV-5705?filter=-1

It completes the following:
1. Added tests to verify that columns can be skipped with ARFF files in R and Python.
2. Added skip column capability to parse_test_file in testUtils.java.  This is mainly for testing only.

In addition, more tests are added to:
1. Verify that skipped columns are supported correctly when importing folders in Java/R/Python.
2. Test pyunit_PUBDEV_5705_drop_columns_high_enums_large.py contains columns with cardinality > 10000000.  When the column with high cardinality is skipped, the file was parsed successfully.

### DIFF
--- a/h2o-core/src/test/java/water/parser/ParserTestARFF.java
+++ b/h2o-core/src/test/java/water/parser/ParserTestARFF.java
@@ -4,20 +4,17 @@ import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Ignore;
 import org.junit.Test;
-import water.DKV;
-import water.H2O;
-import water.Key;
-import water.TestUtil;
+import water.*;
 import water.fvec.Frame;
 import water.fvec.NFSFileVec;
 import water.fvec.Vec;
-
-import static water.parser.ParserTest.makeByteVec;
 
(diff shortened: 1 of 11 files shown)


h2o-core/src/test/java/water/parser/ParserTestARFF.java
```diff
@@ -4,21 +4,18 @@
 import org.junit.BeforeClass;
 import org.junit.Ignore;
 import org.junit.Test;
-import water.DKV;
-import water.H2O;
-import water.Key;
-import water.TestUtil;
+import water.*;
```
Pscheidl: Watch out for those bulk imports, I believe we had a policy about that with @michalkurka . I try not to use them. But this is not an issue for performance really. Mostly this is an issue of being unable to see new imports in new PRs.

wendycwong: Tried to fix it but IntelliJ just replaced it with import water.*;

PR #3013, opened 2018-11-02
Get more useful results from testMultiNode on failure

pkozelka:
When running `testMultiNode` task, the case when it fails gives no useful information in the log, because the testrunner keeps running the suite no matter what happens.

This fix changes the behavior quite significantly, to match the way how tests are typically executed. That is:
- if test passes, next one can execute
- if test fails on assertion (or other "recoverable" error), failure is reported and next test can execute
- if any other (= non-recoverable) exception happens, the testsuite gets killed; which, in this specific test runner, means forced JVM exit (I found no other "gentle" way to break execution). 

That causes the stacktrace and the causing exception to be exposed in the console.

It is critical that tests provide richer information than just PASSED/FAILED. Please consider approving this or proposing better way to achieve something like this.

### DIFF
--- a/h2o-core/src/test/java/water/junit/FatalCrashListener.java
+++ b/h2o-core/src/test/java/water/junit/FatalCrashListener.java
@@ -1,0 +1,19 @@
+package water.junit;
+
+import org.junit.runner.notification.Failure;
+import org.junit.runner.notification.RunListener;
+
+class FatalCrashListener extends RunListener {
+
+  @Override
+  public void testFailure(Failure failure) throws Exception {
+    final Throwable exception = failure.getException();
+
+    // ignore recoverable exceptions
+    if (exception instanceof AssertionError) return;
+
+    // any other causes JVM exit
+    exception.printStackTrace();
+    System.exit(-1);
+  }
+}
--- a/h2o-core/src/test/java/water/junit/H2OTestRunner.java
+++ b/h2o-core/src/test/java/water/junit/H2OTestRunner.java
@@ -40,6 +40,10 @@ public class H2OTestRunner {
     jcore.addListener(new TextListener(jsystem));
     // Add XML generator listener
     jcore.addListener(new XMLTestReporter());
+
+    // Add listener that kills JVM when there is an unhandled exception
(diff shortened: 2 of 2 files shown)


michalkurka:
Thanks for the effort @pkozelka! I am not clear on 2 topics, could you please explain?

> if test fails on assertion (or other "recoverable" error), failure is reported and next test can execute

I thought this is already done by https://github.com/h2oai/h2o-3/blob/master/h2o-core/src/test/java/water/junit/XMLTestReporter.java#L233 - maybe it can be extended there if there is not enough information?

> It is critical that tests provide richer information than just PASSED/FAILED. Please consider approving this or proposing better way to achieve something like this.

If the XML report is generated all the information is available on Jenkins, eg.: http://mr-0xc1:8080/view/H2O-3/job/h2o-3-pipeline/job/master/1109/testReport/junit/hex.deeplearning/DeepLearningProstateTest/Java_10_JUnit___run/ or here http://mr-0xc1:8080/blue/organizations/jenkins/h2o-3-pipeline/detail/master/1109/tests

pkozelka:
@michalkurka you are right, the tests are already providing a lot more information than is available in the console. I was probably somewhat misled with my specific problem where no exception was in console neither in testresults. Therefore the reasoning provided above is probably not very good.

The proposed behavior can be useful by trying to cover following:
- if the tests setup phase fails, it makes no sense to continue at all
- in such case, it's useful to expose the problem in console
- problems in setup can occasionally be completely invisible in tests output

Currently, it will often stop the testsuite prematurely due to considering almost all exceptions fatal.
Now I think that this behavior should be used only for a subset of `Throwable`, for instance, `java.lang.Error` - which includes OOM, and may be perhaps thrown from some H2O initialization steps.

jakubhava:
also @michal-raska is working on improving multi node tests, just FYI to avoid collisions

michalkurka:
@pkozelka for OOM handling we can also use the JVM option `-XX:OnOutOfMemoryError="kill -9 %p"`

pkozelka:
looks like steps in wrong direction, closing

PR #3015, opened 2018-11-05
PUBDEV-5705: added skipped column support to gzip/gz parsers.

wendycwong:
This PR covers the work for JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-5705?filter=myopenissues.

Work done:
1. Implemented skipped column support for gzip/gz parsers.
2. Discussed with Michalk that SVMlight and Avro will not support skipped columns.
3. Added Junit/R/Python unit tests to ensure correct implementation.
4. Added R/Python unit tests to make sure that users can set column names and types with skipped columns.

### DIFF
--- a/h2o-parsers/h2o-avro-parser/src/test/java/water/parser/ParseTestAvro.java
+++ b/h2o-parsers/h2o-avro-parser/src/test/java/water/parser/ParseTestAvro.java
@@ -2,7 +2,6 @@ package water.parser;
 
 
 import com.google.common.io.Files;
-
 import org.apache.avro.Schema;
 import org.apache.avro.SchemaBuilder;
 import org.apache.avro.file.DataFileWriter;
@@ -12,17 +11,16 @@ import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.io.DatumWriter;
 import org.junit.BeforeClass;
 import org.junit.Test;
+import water.TestUtil;
+import water.fvec.Frame;
+import water.fvec.Vec;
+import water.util.StringUtils;
 
 import java.io.File;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 
 import static org.junit.Assert.*;
-
-import water.TestUtil;
-import water.fvec.Frame;
-import water.fvec.Vec;
-import water.util.StringUtils;
 
 /**
  * Test suite for Avro parser.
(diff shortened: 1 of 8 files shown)


h2o-parsers/h2o-avro-parser/src/test/java/water/parser/ParseTestAvro.java
```diff
@@ -34,6 +32,16 @@
   @BeforeClass
   static public void setup() { TestUtil.stall_till_cloudsize(5); }
 
+  @Test
+  public void testSkippedColumns() {
+    try { // specify skipped columns, not allowed!
+      Frame f1 = parse_test_file("smalldata/parser/avro/sequence100k.avro", new int[]{0,1});
+      assert 1==2:"Parser should have thrown an exception but did not!";
```
Pscheidl: This is obviously a leftover.

Pscheidl: Please use JUNIT's fail() method.

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_gz_large.py
```diff
@@ -0,0 +1,90 @@
+from __future__ import print_function
+import sys
+
+sys.path.insert(1, "../../")
+import h2o
+from tests import pyunit_utils
+import random
+
+def import_zip_skipped_columns():
```
Pscheidl: Why does the filename of this test has "large" postfix ?

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_gzip_large.py
```diff
@@ -0,0 +1,63 @@
+from __future__ import print_function
+import sys
+
+sys.path.insert(1, "../../")
+import h2o
+from tests import pyunit_utils
+import random
+
+def import_gzip_skipped_columns():
```
Pscheidl: Why the "large" postfix in the filename ?

Pscheidl:
No functionality added ? Just tests ? Also, I don't understand the "large" postfix in nearly all the test filenames.

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_gz_large.py
wendycwong: large is added to tests that takes longer to complete.  They will be put into different directories and give more time to finish

h2o-parsers/h2o-avro-parser/src/test/java/water/parser/ParseTestAvro.java
wendycwong: Done.

wendycwong:
Pavel:

This PR is more like a panic attack and I need to prove that everything works.  I added all tests that I thought needed to be there here.

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_gzip_large.py
wendycwong: It depends on the developer's body mass.  In my case, I am fat and therefore my tests are also large.

PR #3023, opened 2018-11-08
PUBDEV-5352 - Cannot use checkpoint on new data in DRF. Added enhance…

Pscheidl:
…ment to Python checkpoint setter.

https://0xdata.atlassian.net/browse/PUBDEV-5352

Problem is column skipping. The method `protected void ignoreBadColumns(int npredictors, boolean expensive){ ... }`

ignores column names present in checkpointed model and just deletes/skips those anyway.

### DIFF
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -1152,8 +1152,9 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
     Vec vvecs[] = new Vec[names.length];
     int good = 0;               // Any matching column names, at all?
     int convNaN = 0;  // count of columns that were replaced with NA
-    for( int i=0; i<names.length; i++ ) {
-      Vec vec = test.vec(names[i]); // Search in the given validation set
+    final Frame.FrameVecRegistry frameVecRegistry = test.frameVecRegistry();
+    for (int i = 0; i < names.length; i++) {
+      Vec vec = frameVecRegistry.findByColName(names[i]); // Search in the given validation set
       boolean isResponse = response != null && names[i].equals(response);
       boolean isWeights = weights != null && names[i].equals(weights);
       boolean isOffset = offset != null && names[i].equals(offset);
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -28,10 +28,7 @@ import water.util.*;
 import java.io.*;
 import java.lang.reflect.Field;
 import java.net.URI;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Comparator;
-import java.util.Random;
+import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -1252,6 +1249,12 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
                                            final String[][] domains) {
 
     boolean haveCategoricalPredictors = false;
+    final Map<String, Integer> namesIndicesMap = new HashMap<>(names.length);
+
+    for (int i = 0; i < names.length; i++) {
+      namesIndicesMap.put(names[i], i);
+    }
+
     for (int i = 0; i < frame.numCols(); ++i) {
       if (frame.names()[i].equals(responseName)) continue;
       if (frame.names()[i].equals(wieghtsName)) continue;
@@ -1263,7 +1266,7 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
         break;
       }
       // or a equally named column of the training set is categorical, but the test column isn't (e.g., numeric column provided to be converted to a factor)
-      int whichCol = ArrayUtils.find(names, frame.name(i)); // TODO: speedup this O(n) search
+      final int whichCol = namesIndicesMap.get(frame.name(i));
       if (whichCol >= 0 && domains[whichCol] != null) {
         haveCategoricalPredictors = true;
         break;
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -1096,9 +1096,12 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
    * @param interactionBldr Column names to create pairwise interactions with
    * @param catEncoded Whether the categorical columns of the test frame were already transformed via categorical_encoding
    */
-  public static String[] adaptTestForTrain(Frame test, String[] origNames, String[][] origDomains, String[] names, String[][] domains,
-                                           Parameters parms, boolean expensive, boolean computeMetrics, InteractionBuilder interactionBldr, ToEigenVec tev,
-                                           IcedHashMap<Key, String> toDelete, boolean catEncoded) throws IllegalArgumentException {
+  public static String[] adaptTestForTrain(final Frame test, final String[] origNames, final String[][] origDomains,
+                                           String[] names, String[][] domains, final Parameters parms,
+                                           final boolean expensive, final boolean computeMetrics,
+                                           final InteractionBuilder interactionBldr, final ToEigenVec tev,
+                                           final IcedHashMap<Key, String> toDelete, final boolean catEncoded)
+          throws IllegalArgumentException {
     String[] msg = new String[0];
     if (test == null) return msg;
     if (catEncoded && origNames==null) return msg;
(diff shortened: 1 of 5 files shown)


h2o-py/h2o/estimators/random_forest.py
```diff
@@ -627,7 +627,10 @@ def checkpoint(self):
     @checkpoint.setter
     def checkpoint(self, checkpoint):
         assert_is_type(checkpoint, None, str, H2OEstimator)
-        self._parms["checkpoint"] = checkpoint
+        if(issubclass(checkpoint, H2OEstimator)):
```
Pscheidl: This is enhancement. Makes it possible to insert just the model to be checkpointed and not to explicitely extract its `model_id`.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -906,17 +906,43 @@ protected boolean ignoreUuidColumns() {
...
-      }.doIt(_train,"Dropping bad and constant columns: ",expensive);
+      }.doIt(_train, "Dropping bad and constant columns: ", expensive);
+
+
+    } else if(_parms._checkpoint != null){ // If there is a checkpoint, keep exactly the same columns
+      final Model checkpointedModel = _parms._checkpoint.get();
+      assert checkpointedModel != null;
+
+      // Avoid O(n) lookup
+      final Set<String> checkpointNames = new HashSet<>(checkpointedModel._output._names.length);
```
Pscheidl: For Lookup, a method `find` of `Frame` class could be used. However, the complexity of that method is *O(n)* now. To reduce the comp. complexity of Frame's `find`, a HashSet must be introduced inside Frame. It would hold redudant values to `_names` string array. For many columns, it also means "slightly" bigger memory overhead we have no explicit reason to introduce right now.

This is the reason I did it this way in this place. Quickly instantiate a HashSet with columns listed in the checkpointed model, use it to search few lines later and then drop it. Minimal change right now.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -906,17 +906,43 @@ protected boolean ignoreUuidColumns() {
...
+      for (String col : checkpointedModel._output._names){
+        checkpointNames.add(col);
+      }
+
+      // Number of columns skipped equals to the difference in sizes of those columns
+      // Missing columns check is done later
```
Pscheidl: The check is done in `init` method of `SharedTree` class.

h2o-core/src/main/java/hex/ModelBuilder.java
wendycwong: Yeap, we are always trading off memory/computation in all kinds of algos.  Feels like the fight of good vs evil going on again.

h2o-py/tests/testdir_jira/pyunit_pubdev_5352.py
```diff
@@ -0,0 +1,21 @@
...
+    mnist_original = h2o.import_file(pyunit_utils.locate("smalldata/flow_examples/mnist/test.csv.gz"));
+    mnist_original['C785'] = mnist_original['C785'].asfactor()
+    predictors = mnist_original.columns[0:-1]
+    target = 'C785'
+    train, valid, new_data = mnist_original.split_frame(ratios=[.5, .2], seed=1234)
+    drf = H2ORandomForestEstimator(model_id='drf', ntrees=3, seed=1234)
+    drf.train(x=predictors, y=target, training_frame=train, validation_frame=valid)
+    drf_checkpoint = H2ORandomForestEstimator(model_id='drf_checkpoint',checkpoint=drf, ntrees=4, seed=1234)
+    drf_checkpoint.train(x=predictors, y=target, training_frame=new_data, validation_frame=valid)
+
```
wendycwong: Is there a way to break up a DRF training into two parts using checkpoint and then compare with the results when no checkpoint are used but trained to the bitter end?  If the checkpoint is done correctly, the two models should give us the same answer?

Pscheidl: Great addition. I'll add such test.

michalkurka:
@Pscheidl please don't merge yet - I would like to read through it one more time

Pscheidl:
> @Pscheidl please don't merge yet - I would like to read through it one more time

Sure @michalkurka . Waiting for your green light.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -906,17 +906,43 @@ protected boolean ignoreUuidColumns() {
...
+        protected boolean filter(Vec v) {
           boolean isBad = v.isBad();
           boolean skipConst = ignoreConstColumns() && v.isConst();
           boolean skipString = ignoreStringColumns() && v.isString();
           boolean skipUuid = ignoreUuidColumns() && v.isUUID();
           boolean skip = isBad || skipConst || skipString || skipUuid;
           return skip;
         }
-      }.doIt(_train,"Dropping bad and constant columns: ",expensive);
+      }.doIt(_train, "Dropping bad and constant columns: ", expensive);
```
michalkurka: I think we should drop bad columns and columns the algo doesn't support (strings).

If the user changes the type of the column (eg. from Number to UUID) then this will fail in the model training. I think this needs a little different approach - but you are on the right path.

Pscheidl: This is old code you are probably referring to. 

The code below only deletes features not present in the checkpointed model. That's it. I can see the limits of this approach. The goals are:

- Implement approach checking column types of shared columns being equal 
- Use list of columns in the checkpointed model as a blacklist - if the same column is in the original dataset, it can not be blacklisted when resuming from that checkpoint (the root cause for this error we're trying to solve here).
- Do not delete new columns, as it might make sense for some algos to leverage them  ? For example DRF (trees may be added with new splits on features unseen in previously created trees ?).

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1094,14 +1094,26 @@ public void init(boolean expensive) {
...
-    ignoreBadColumns(separateFeatureVecs(), expensive);
-    ignoreInvalidColumns(separateFeatureVecs(), expensive);
+    
+    if(_parms._checkpoint != null){
+      if(DKV.get(_parms._checkpoint) == null){
+          error("_checkpoint", "Checkpoint has to point to existing model!");
+      }
+      // Do not ignore bad columns, as only portion of the training data might be supplied (e.g. continue from checkpoint)
+      final Model checkpointedModel = _parms._checkpoint.get();
+      final String[] warnings = checkpointedModel.adaptTestForTrain(_train, expensive, false);
```
Pscheidl: 1. I believe using the whole `adaptTestForTrain` method as-is is useful as-is.
2. The method has O(n) search for `n` columns. Tried to avoid this in the previous commit. We could do better really.

h2o-core/src/main/java/hex/ModelBuilder.java
Pscheidl: This is all a matter of how we define our operations. Let `f(x)` be a model-creating function. Also, let `A` be a training set. Let's consider a model `M = f(A)`.

Now, split the training set `A` into two subsets `B,C`, so that `B + C = A`. Creating a checkpointed model by `Mc = f(B)` and then finishing the model from where we left off by `Mf = f(C, Mc)`, the resulting model `Mf` should be equal to the original `M`. (`M = Mf`).

If these models should be equal, than we may not drop bad columns in all cases, as a column from a subset of the original data might look like a bad column, bad when the whole original dataset is fed, the column no longer looks "bad", e.g. being full of 0s. This is exactly the case with this PR.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1095,13 +1095,25 @@ public void init(boolean expensive) {
...
+      }
+    } else {
+      // Drop all non-numeric columns (e.g., String and UUID).  No current algo
+      // can use them, and otherwise all algos will then be forced to remove
+      // them.  Text algos (grep, word2vec) take raw text columns - which are
+      // numeric (arrays of bytes).
+      ignoreBadColumns(separateFeatureVecs(), expensive);
+    }
     checkResponseVariable();
+    ignoreInvalidColumns(separateFeatureVecs(), expensive);
```
michalkurka: is this necessary even in the case `adaptTestForTrain` was called? it should take care of the BAD columns as well, right?

Pscheidl: Both checks are duplicated and should be moved into the "else" block.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1223,41 +1224,59 @@ else if (isWeights && computeMetrics) {
...
-        // or a equally named column of the training set is categorical, but the test column isn't (e.g., numeric column provided to be converted to a factor)
-        int whichCol = ArrayUtils.find(names, test.name(i));
-        if (whichCol >= 0 && domains[whichCol] != null) {
-          haveCategoricalPredictors = true;
-          break;
-        }
+      final boolean hasCategoricalPredictors = hasCategoricalPredictors(test, response, weights, offset, fold, names, domains);
+
+      // check if we first need to expand categoricals before calling this method again
+      if (hasCategoricalPredictors) {
```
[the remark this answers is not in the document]
Pscheidl: It can only get to this point if expansive is true. The check before was duplicated, thus I removed it.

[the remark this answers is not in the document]
michalkurka: I see! Sorry @Pscheidl! my mistake

PR #3024, opened 2018-11-08
PUBDEV-6036: Fix `offset_column` parameter in the Python client while initializing an estimator

maurever:
The issue was focused on `offset_column` parameter, but it is the same problem for `weights_column` and `fold_column` parameters. Now there is a check if the parameters have been already set during initialization of estimator. If yes, it checks the new value is the same, if the values are different, it sets the last passed value to `params` dictionary and throws `UserWarning`.

There was also a bug in the documentation about the `offset_column` parameter (the wrong name of the column was assigned to the parameter), so I also fix that.

https://0xdata.atlassian.net/browse/PUBDEV-6036

### DIFF
--- a/h2o-docs/src/product/data-science/algo-params/offset_column.rst
+++ b/h2o-docs/src/product/data-science/algo-params/offset_column.rst
@@ -95,7 +95,7 @@ Example
 
 	# try using the `offset_column` parameter:
 	# initialize the estimator then train the model
-	boston_gbm = H2OGradientBoostingEstimator(offset_column = "offset_column", seed = 1234)
+	boston_gbm = H2OGradientBoostingEstimator(offset_column = "offset", seed = 1234)
 	boston_gbm.train(x=predictors, y=response, training_frame=train, validation_frame=valid)
 
 	# print the mse for validation set
--- a/h2o-py/h2o/estimators/estimator_base.py
+++ b/h2o-py/h2o/estimators/estimator_base.py
@@ -182,10 +182,9 @@ class H2OEstimator(ModelBase):
                             raise H2OValueError("Column %s not in the training frame" % xi)
                         xset.add(xi)
             x = list(xset)
-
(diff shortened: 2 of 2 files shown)


michalkurka:
Interesting - some tests started to fail after this change:

```
stacktrace:
water.exceptions.H2OModelBuilderIllegalArgumentException: Illegal argument(s) for XGBoost model: XGBoost_model_python_1541709076839_1. Details: ERRR on field: _offset_column: Offset is not supported for XGBoost.
```

I don't think that is your fault. Probably the tests need to be fixed. Can you please look into it?

Also: please consider adding tests of the new functionality.

maurever:
That is interesting it didn't cause failure before, because the XBOOST implementation in H2O does not support `offset_column`  parameter, the test is testing it and does not expect that it will fail... In the test, the `train` method is used and the parameter `offset_column` in the `train` method should work the same as before my commit... 

I can remove the testing `offset_column` parameter from [pyunit_insurance_tweedie_small.py](https://github.com/h2oai/h2o-3/blob/master/h2o-py/tests/testdir_algos/xgboost/pyunit_insurance_tweedie_small.py) but I am not sure if it is the right solution...

And of course, I will add tests...

h2o-py/tests/testdir_algos/xgboost/pyunit_insurance_tweedie_small.py
```diff
@@ -11,13 +11,15 @@ def xgboost_insurance_tweedie_small():
     test_frame = h2o.import_file(pyunit_utils.locate("smalldata/testng/insurance_validation1.csv"))
     x = ['District', 'Group', 'Age', 'Holders']
     y = 'Claims'
-    offset_column = 'offset'
+    # maurever: offset column parameter is not supported in XGBOOST
```
michalkurka: I would just delete the offset parameters.

PR #3027, opened 2018-11-08
PUBDEV-6046: Failed to get metric: auc from ModelMetrics type BinomialGLM

michalkurka:
This PR makes sure the AUC2 object is created even in edge cases when the data have 0 total weight. This prevents possible NPEs.

### DIFF
--- a/h2o-core/src/main/java/hex/ModelMetrics.java
+++ b/h2o-core/src/main/java/hex/ModelMetrics.java
@@ -154,16 +154,15 @@ public class ModelMetrics extends Keyed<ModelMetrics> {
     if (null == method)
       throw new H2OIllegalArgumentException("Failed to find ModelMetrics for criterion: " + criterion);
 
-    double c;
     try {
-      c = (double) method.invoke(obj);
+      return (double) method.invoke(obj);
     } catch (Exception e) {
+      Log.err(e);
       throw new H2OIllegalArgumentException(
               "Failed to get metric: " + criterion + " from ModelMetrics object: " + mm,
-              "Failed to get metric: " + criterion + " from ModelMetrics object: " + mm + ", criterion: " + method, e
+              "Failed to get metric: " + criterion + " from ModelMetrics object: " + mm + ", criterion: " + method + ", exception: " + e.getMessage()
       );
     }
-    return c;
   }
 
 
--- a/h2o-core/src/main/java/hex/ModelMetrics.java
+++ b/h2o-core/src/main/java/hex/ModelMetrics.java
@@ -127,11 +127,16 @@ public class ModelMetrics extends Keyed<ModelMetrics> {
   }
 
   static public double getMetricFromModelMetric(ModelMetrics mm, String criterion) {
-    if (null == criterion || criterion.equals("")) throw new H2OIllegalArgumentException("Need a valid criterion, but got '" + criterion + "'.");
+    if (null == criterion || criterion.equals(""))
+      throw new H2OIllegalArgumentException("Need a valid criterion, but got '" + criterion + "'.");
+
     Method method = null;
+    Object obj = null;
+    criterion = criterion.toLowerCase();
     ConfusionMatrix cm = mm.cm();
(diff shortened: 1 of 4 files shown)


h2o-core/src/main/java/hex/ModelMetricsBinomial.java
```diff
@@ -210,12 +210,14 @@ private ModelMetrics makeModelMetrics(Model m, Frame f, GainsLift gl) {
       double mse = Double.NaN;
       double logloss = Double.NaN;
       double sigma = Double.NaN;
-      AUC2 auc = null;
+      final AUC2 auc;
```
pkozelka: cool, I like immutable vars wherever possible ;)

h2o-core/src/main/java/hex/ModelMetrics.java
```diff
@@ -150,16 +156,12 @@ static public double getMetricFromModelMetric(ModelMetrics mm, String criterion)
...
-    } catch(Exception fallthru) {
-      try {
-        c = (double)method.invoke(cm);
-      } catch (Exception e) {
-        throw new H2OIllegalArgumentException(
-                "Failed to get metric: " + criterion + " from ModelMetrics object: " + mm,
-                "Failed to get metric: " + criterion + " from ModelMetrics object: " + mm + ", criterion: " + method + ", exception: " + e
-        );
-      }
+      c = (double) method.invoke(obj);
```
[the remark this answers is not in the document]
michalkurka: fixed. thank you

h2o-core/src/main/java/hex/ModelMetrics.java
```diff
@@ -127,19 +127,25 @@ static public double getMetricFromModel(Key<Model> key, String criterion) {
   }
 
   static public double getMetricFromModelMetric(ModelMetrics mm, String criterion) {
-    if (null == criterion || criterion.equals("")) throw new H2OIllegalArgumentException("Need a valid criterion, but got '" + criterion + "'.");
+    if (null == criterion || criterion.equals(""))
+      throw new H2OIllegalArgumentException("Need a valid criterion, but got '" + criterion + "'.");
+
```
pkozelka: this hunk seems to be reformat-only; wouldn't you consider adding curly braces?
Maybe my personal problem, but I find the python-like formatting slightly confusing in Java code.

h2o-core/src/main/java/hex/ModelMetrics.java
```diff
@@ -127,19 +127,25 @@ static public double getMetricFromModel(Key<Model> key, String criterion) {
...
+    Object obj = null;
+    criterion = criterion.toLowerCase();
     ConfusionMatrix cm = mm.cm();
     try {
-      method = mm.getClass().getMethod(criterion.toLowerCase());
+      method = mm.getClass().getMethod(criterion);
+      obj = mm;
     }
     catch (Exception e) {
       // fall through
```
pkozelka: I am aware that this is outside your PR, just noting: here the exception is _ignored_, perhaps it's my poor English but _fall through_ gives me completely different intuition for handling than what happens here. 
Besides, such frequent exception swallowing, even in obvious cases like this, is a little bit disturbing. I guess that in most these cases, at least `IllegalStateException` could have been thrown.

h2o-core/src/main/java/hex/ModelMetrics.java
michalkurka: added

h2o-core/src/main/java/hex/ModelMetrics.java
michalkurka: the execution of this method through to the next step ;)

an exception is eventually thrown if neither of the attempts succeeds

mjorgen1:
I am getting this error "Failed to find ModelMetrics for criterion: auc" but for when I run H2OAutoML, after I set sort_metric='auc'. Any solutions for this? @pkozelka kozelka @wendycwong

PR #3028, opened 2018-11-08
[PUBDEV-5770] implement auto-saving of models for AutoML and GridSearch

honzasterba:


michalkurka:
Perhaps we can make available to every model? If it is a common ModelBuilder parameter - the users will be able to specify it for grid search as well.

h2o-algos/src/main/java/hex/schemas/PCAV3.java
```diff
@@ -10,23 +10,24 @@
...
+        "pca_method",
+        "pca_impl",
+        "k",
+        "max_iterations",
+        "use_all_factor_levels",
+        "compute_metrics",
+        "impute_missing",
+        "seed",
+        "max_runtime_secs",
+        "export_checkpoints_dir"
```
michalkurka: Unnecessary reformating makes git-blame useless. I try to keep the original formatting unless it is horrible ;)

h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
```diff
@@ -425,62 +425,27 @@ Model getLeader() {
...
         Key frame_key = m._output._cross_validation_metrics.frame()._key;
         long frame_checksum = m._output._cross_validation_metrics.frame().checksum();
-        other_metrics[i++] = ModelMetrics.getMetricFromModelMetric(leaderboard_set_metrics.get(ModelMetrics.buildKey(model_key, model_checksum, frame_key, frame_checksum)), other_metric);
+        other_metrics[i++] = ModelMetrics.getMetricFromModelMetric(leaderboard_set_metrics.get(ModelMetrics.buildKey(model_key, model_checksum, frame_key, frame_checksum)), metric);
       }
     }
     return other_metrics;
   }
 
-  private static double[] getSortMetrics(String sort_metric, IcedHashMap<Key<ModelMetrics>, ModelMetrics> leaderboard_set_metrics, Frame leaderboardFrame, Model[] models) {
```
michalkurka: good clean-up, thanks

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -234,6 +236,17 @@ private void setFinalState() {
     }
   }
 
+  private void saveModelCheckpointIfConfigured() {
+    Model model = _result.get();
+    if (model != null && !StringUtils.isNullOrEmpty(model._parms._export_checkpoints_dir)) {
+      try {
+        model.exportBinaryModel(model._parms._export_checkpoints_dir + "/" + model._key.toString(), true);
+      } catch (IOException e) {
+        throw new H2OIllegalArgumentException("export_checkpoints_dir", "saveModelIfConfigured", e);
```
michalkurka: Hmm, I think this might be a costly mistake if the user inputs the directory incorrectly. This exception will stop CV models and prevent the main model from being built.

I think we should pre-create the target subdirectory for the model before it is actually built. This will make sure we have a write-access to the directory and that the directory exists.

I would also consider a special handling of the CV models - produce just a warning if we cannot write the CV model but fail if we cannot write the main model. But I am not 100% sure if this helps anything.

michalkurka:
@honzasterba looks good - one thing I would like to add is a better validation of the target directory - we should ideally make sure we have a write-access to the directory before we attempt to build the model.

michalkurka:
I also didn't notice a test for this feature. Would be good to have a small test in both py & r.

Perhaps one more idea: can we have a bit more structured output? Eg. put main model in a directory and CV models in sub-directories? If we have some kind of structure then we can add functions to automatically load main model & the cv models (= load all models from a directory).

h2o-algos/src/main/java/hex/schemas/PCAV3.java
honzasterba: thats very subjective, I try to keep code in better shape when I touch it

h2o-core/src/main/java/hex/ModelBuilder.java
honzasterba: good point. I think we should keep any advanced structuring of the checpoints dir to another task
as per not failing on CV models its would be the same case for main model, so we would only delay the exception as such I think its better to fail sonner rather than later

honzasterba:
this PR is getting big enough, I would focus on just saving the checkpoints for now and figuroe out structuing in another PR

h2o-algos/src/main/java/hex/schemas/PCAV3.java
michalkurka: yep, it is subjective - no question about that

but it is objectively harder to figure out what happened with the code if there is a lot of unrelated refactorings ;)

I don't want to discourage you from improving - but I do suggest to make 2 PRs - one with refactorings and the second one with functional changes (not necessary here in the case of modified indentation in this list but certainly recommended for the automl changes). It is easy to hide an issue in a large PR (65 modified files).

h2o-core/src/main/java/hex/ModelBuilder.java
michalkurka: sure, we can do the nested directories in another PR - definitely not necessary in this one

regarding the CV models - I think I agree - let's make it simpler (at least for now). My point of view is that the model checkpointing will most likely be used for long-running tasks (eg. overnight) - you don't want to find out that you ended up without a trained model just because you enabled checkpointing (that was supposed to prevent you from losing your work in the first place ;))

h2o-py/tests/testdir_algos/automl/pyunit_automl_checkpoints.py
```diff
@@ -0,0 +1,43 @@
...
+    aml = H2OAutoML(project_name="py_aml0", stopping_rounds=3, stopping_tolerance=0.001, stopping_metric="AUC",
+                    max_models=2, seed=1234, export_checkpoints_dir=checkpoints_dir)
+    aml.train(y="CAPSULE", training_frame=train)
+
+    get_aml = get_automl(aml.project_name)
+    num_files = len([f for f in os.listdir(checkpoints_dir) if "_cv" not in f])  # do not count CV models
+    shutil.rmtree(checkpoints_dir)
+
+    assert aml.project_name == get_aml["project_name"]
+    assert get_aml["leaderboard"].nrows == num_files, "Not all generated autoML models were saved."
```
michalkurka: would be good to check that at least one model was built

h2o-r/tests/testdir_algos/glm/runit_GLM_grid_checkpoints.R
```diff
@@ -0,0 +1,29 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../../scripts/h2o-r-test-setup.R")
+
+check.glm.grid.checkpoints <- function(conn) {
+  cars <- h2o.uploadFile(locate("smalldata/junit/cars_20mpg.csv"))
+  seed <- sample(1:1000000, 1)
```
michalkurka: deterministic tests are better, I would use a fixed seed

michalkurka: no need to subset the training frame at all for this test

h2o-r/tests/testdir_algos/glm/runit_GLM_grid_checkpoints.R
```diff
@@ -0,0 +1,29 @@
...
+  Log.info(paste0("runif seed: ",seed))
+  r <- h2o.runif(cars,seed=seed)
+  train <- cars[r > 0.2,]
+
+  hyper_params = list()
+  hyper_params$alpha <- lapply(round(runif(sample(2:3,1)),6), function (x) x)
```
michalkurka: same as above, no need to be random

h2o-py/tests/testdir_algos/automl/pyunit_automl_checkpoints.py
honzasterba: good idea

h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
```diff
@@ -37,7 +37,8 @@
       "metalearner_fold_column",
       "keep_levelone_frame",
       "metalearner_params",
-      "seed"
+      "seed",
+      "export_checkpoints_dir"
```
seb-h2o: does it make sense to expose this for SE? I'd rather encourage users to pass it as a metalearner param

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -793,6 +795,7 @@ private void setCommonModelBuilderParams(Model.Parameters params) {
 
     params._keep_cross_validation_models = buildSpec.build_control.keep_cross_validation_models;
     params._keep_cross_validation_fold_assignment = buildSpec.build_control.nfolds != 0 && buildSpec.build_control.keep_cross_validation_fold_assignment;
+    params._export_checkpoints_dir = buildSpec.build_control.export_checkpoints_dir;
```
seb-h2o: for Stacked Ensembles, this will probably be ignored. If we want this to apply to SEs as well, the property needs to be propagated to the metalearner in the stack() method

PR #3029, opened 2018-11-09
PUBDEV-5261: Allow custom_metric to be used for early stopping

michalkurka:
@angela0xdata the issue with PR is that custom metric is only allowed for DRF/GBM & GLM but I cannot have different early stopping metrics for each algo. That means stopping_criteria = custom is available everywhere even though it is not actually supported.

I will need to add some checks to the source code.

I am adding 2 new stopping criteria:
- custom - for custom metric functions where "less is better", it is expected that the lower bound is 0
- custom_increasing  - for custom metric functions where "more is better"

### DIFF
--- a/h2o-core/src/main/java/hex/ScoreKeeper.java
+++ b/h2o-core/src/main/java/hex/ScoreKeeper.java
@@ -8,9 +8,6 @@ import water.util.Log;
 import water.util.MathUtils;
 
 import java.util.Arrays;
-import java.util.Comparator;
-
-import static hex.ScoreKeeper.StoppingMetric.deviance;
 
 /**
  * Low-weight keeper of scores
@@ -111,12 +108,16 @@ public class ScoreKeeper extends Iced {
       ((ScoreKeeperAware) m).fillTo(this);
     }
     if (m._custom_metric != null )
-    _custom_metric =  m._custom_metric.value;
+      _custom_metric =  m._custom_metric.value;
   }
 
-  public enum StoppingMetric { AUTO, deviance, logloss, MSE, RMSE,MAE,RMSLE, AUC, lift_top_group, misclassification, mean_per_class_error, custom} //, r2}
-  public static boolean moreIsBetter(StoppingMetric criterion) {
-    return (criterion == StoppingMetric.AUC || criterion == StoppingMetric.lift_top_group); // || criterion == StoppingMetric.r2);
+  public enum StoppingMetric { AUTO, deviance, logloss, MSE, RMSE,MAE,RMSLE, AUC, lift_top_group, misclassification, mean_per_class_error, custom, custom_increasing} //, r2}
+  static boolean moreIsBetter(StoppingMetric criterion) {
+    return (criterion == StoppingMetric.AUC || criterion == StoppingMetric.lift_top_group || criterion == StoppingMetric.custom_increasing); // || criterion == StoppingMetric.r2);
+  }
+  // only for those metrics where "less is better"
+  private static boolean hasLowerBound(StoppingMetric criterion) {
+    return criterion != StoppingMetric.deviance;
   }
 
   /** Based on the given array of ScoreKeeper and stopping criteria should we stop early? */
@@ -126,10 +127,11 @@ public class ScoreKeeper extends Iced {
     if (len < 2*k) return false; //need at least k for SMA and another k to tell whether the model got better or not
 
     if (criterion==StoppingMetric.AUTO) {
-      criterion = classification ? StoppingMetric.logloss : deviance;
+      criterion = classification ? StoppingMetric.logloss : StoppingMetric.deviance;
     }
 
     boolean moreIsBetter = moreIsBetter(criterion);
+    boolean hasLowerBound = hasLowerBound(criterion);
     double movingAvg[] = new double[k+1]; //need one moving average value for the last k+1 scoring events
     double lastBeforeK = moreIsBetter ? -Double.MAX_VALUE : Double.MAX_VALUE;
     double bestInLastK = moreIsBetter ? -Double.MAX_VALUE : Double.MAX_VALUE;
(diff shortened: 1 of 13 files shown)


angela0xdata:
What do "custom" and "custom_metric" mean? What happens when each of these is selected? (This info will help me with PUBDEV-6048.)

michalkurka:
@angela0xdata added details in the PR description

angela0xdata:
okay, this is starting to make sense. And I just realized that I need to add "custom_metric_func" to my docs. I'll include that as part of PUBDEV-6048.

h2o-core/src/main/java/water/api/schemas3/ModelParametersSchemaV3.java
```diff
@@ -162,7 +162,7 @@
    * Metric to use for convergence checking, only for _stopping_rounds > 0
    */
 //  @API(help = "Metric to use for early stopping (AUTO: logloss for classification, deviance for regression)", values = {"AUTO", "deviance", "logloss", "MSE", "RMSE","MAE","RMSLE", "AUC", "lift_top_group", "misclassification", "mean_per_class_error", "r2"}, level = API.Level.secondary, direction=API.Direction.INOUT, gridable = true)
-  @API(help = "Metric to use for early stopping (AUTO: logloss for classification, deviance for regression)", values = {"AUTO", "deviance", "logloss", "MSE", "RMSE","MAE","RMSLE", "AUC", "lift_top_group", "misclassification", "mean_per_class_error"}, level = API.Level.secondary, direction=API.Direction.INOUT, gridable = true)
+  @API(help = "Metric to use for early stopping (AUTO: logloss for classification, deviance for regression)", values = {"AUTO", "deviance", "logloss", "MSE", "RMSE","MAE","RMSLE", "AUC", "lift_top_group", "misclassification", "mean_per_class_error", "custom", "custom_increasing"}, level = API.Level.secondary, direction=API.Direction.INOUT, gridable = true)
```
seb-h2o: @michalkurka, FYI I created an EnumValuesProvider to avoid having to sync those enums.
see [usage example](https://github.com/h2oai/h2o-3/blob/29cb7b4ecbcf5b0714ea72732f7b048c29541e58/h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java#L126)

h2o-core/src/main/java/hex/ScoreKeeper.java
```diff
@@ -253,83 +259,6 @@ public static boolean stopEarly(ScoreKeeper[] sk, int k, boolean classification,
             && MathUtils.compare(_lift, o._lift, 1e-6, 1e-6);
   }
 
-  public static Comparator<ScoreKeeper> comparator(StoppingMetric criterion) {
```
seb-h2o: always feel good to delete long useless code

h2o-py/h2o/estimators/deeplearning.py
```diff
@@ -1003,13 +1003,14 @@ def stopping_metric(self):
         Metric to use for early stopping (AUTO: logloss for classification, deviance for regression)
 
         One of: ``"auto"``, ``"deviance"``, ``"logloss"``, ``"mse"``, ``"rmse"``, ``"mae"``, ``"rmsle"``, ``"auc"``,
-        ``"lift_top_group"``, ``"misclassification"``, ``"mean_per_class_error"``  (default: ``"auto"``).
+        ``"lift_top_group"``, ``"misclassification"``, ``"mean_per_class_error"``, ``"custom"``, ``"custom_increasing"``
```
seb-h2o: still don't understand why we keep those generated filed in repo, confuses me every time

h2o-core/src/main/java/water/api/schemas3/ModelParametersSchemaV3.java
michalkurka: That looks nice, it is useful in those case when you want to expose all the values in the Enum (=> none of them are internal). Thank you

h2o-py/h2o/estimators/deeplearning.py
michalkurka: Yeah, it has pros and cons.

@seb-h2o The main reason why we do that is the ability to clearly see the effect of a change in the generator/parameter definition in the pull request. This way it is impossible to miss an accidental change, both you and the reviewer can go over it. It also makes sure the generated code doesn't change randomly (eg. unpredictable order of methods) and you can see how the file was changed in history and what changes were made.

Another benefit is that it makes the files visible to the outside world, you can search in it and you can provide links to the source code. That is very useful sometimes. You can checkout any branch and you will see the source code without running the build (you don't even need the right environment to run the build - which can change a lot over time - I am still able to look at the source code of h2o2 even though my build is failing ;)).

seb-h2o: ok, you convinced me: pros look useful indeed

h2o-core/src/main/java/water/api/schemas3/ModelParametersSchemaV3.java
michalkurka: @seb-h2o I tried, it doesn't work for ModelBuilders (values annotation is expected). Needs a modification. I will submit a PR. But do we really need the Provider classes? Wouldn't it be enough to just have a static method that converts enum to list of string values? like this:

    public static <E extends Enum<E>> String[] getValuesOf(Class<E> clazz) {
      E[] enumValues = clazz.getEnumConstants();
      String[] values = new String[enumValues.length];
      for (int i = 0; i < enumValues.length; i++) {
        values[i] = enumValues[i].name();
      }
      return values;
    }

it could be defined on SchemaV3 class

seb-h2o: you can't use those kind of methods directly in Java annotations (or any kind of dynamic value), but classes are allowed, hence the ValuesProvider class to circumvent this.

not sure why it didn't work for ModelBuilder, processing logic in SchemaMetadata is rather simple:
```java
this.values = annotation.valuesProvider() == ValuesProvider.NULL ? annotation.values() : getValues(annotation.valuesProvider());
...
// If the field is an enum then the values annotation field had better be set. . .
if (is_enum && (null == this.values || 0 == this.values.length)) {
   throw H2O.fail("Didn't find values annotation for enum field: " + this.name);
}
 ```

PR #3032, opened 2018-11-13
Pubdev-3451: add quiet argument to init

fvermeij:
Hello h2o people,

Jira here: https://0xdata.atlassian.net/browse/PUBDEV-3451

This is my first contribution to h2o, so if I did anything wrong, just point it out. I hope I followed all the guidelines.

This is a new feature, that adds an option to the 'h2o.init' function, to surpress output. As mentioned in the Jira, this has advantages for R, because the outputs interfere with knitr documents, making it hard to use h2o in latex. Furthermore, the output can interfere with certain unit test or build pipelines. 
The progress bars can be disabled with "h2o.no_progress()". It would be nice if the startup connection messages could be disabled too.

The option is implemented slightly differently for R and Python. In Python, 'verbose' is set to False, meaning no connection messages are printed at all, including when shutting down, etcetera. In R, there is no such verbose mode, so only the messages during init are hidden, to match the Jira ticket.

### DIFF
--- a/h2o-py/h2o/h2o.py
+++ b/h2o-py/h2o/h2o.py
@@ -174,7 +174,7 @@ def init(url=None, ip=None, port=None, name=None, https=None, insecure=None, use
     :param strict_version_check: If True, an error will be raised if the client and server versions don't match.
     :param ignore_config: Indicates whether a processing of a .h2oconfig file should be conducted or not. Default value is False.
     :param extra_classpath: List of paths to libraries that should be included on the Java classpath when starting H2O from Python.
-    :param quiet: If True, the connection will not show messages when starting up. Default value is False.
+    :param quiet: If True, the connection will not show connection status messages. Default value is False.
     :param kwargs: (all other deprecated attributes)
     :param jvm_custom_args Customer, user-defined argument's for the JVM H2O is instantiated in. Ignored if there is an instance of H2O already running and the client connects to it.
     """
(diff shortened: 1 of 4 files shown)


michalkurka:
@fvermeij thank you for your contribution. It is very appreciated.

At a first glance, things look good!

@ledell was the requester of this change. Erin, can you please review?

michalkurka:
Mr. Jenkins, please test!

ledell:
Thanks, I will review this soon!  Very nice idea, I think people have complained about this before.

swt2c:
@fvermeij any chance you could resolve the merge conflicts?  I would love to see this change implemented.  When running h2o from cron, I don't want to see any output unless there is a problem.

fvermeij:
> 
> 
> @fvermeij any chance you could resolve the merge conflicts? I would love to see this change implemented. When running h2o from cron, I don't want to see any output unless there is a problem.

Yes I fixed them :) I still think this would be a useful extra option for h2o, even though this PR has been inactive for a while.

bscan:
Hey, I'd love to see this implemented too. I commented this on the Jira ticket as well, but an important use case is Kaggle Notebooks. They are becoming popular ways to show others how to use an algorithm or software package. However, when using H2O for simple examples, much of the output consists of cluster diagnostics irrelevant to the algorithm itself and often poorly formatted. This gives the impression to novices that H2O is "heavy" or complicated when in reality it's user friendly due to things like automatic categorical feature encoding, lambda_search, and AutoML.

valenad1:
Done in #7079 with this #15653

PR #3034, opened 2018-11-13
Vaclav pubdev 5897

vaclavbelak:
This PR upgrades the examples in R package to match the contemporary style guidelines. I modified only the source files, so the PDF would need to be generated (happy to do so if you want me to).

https://0xdata.atlassian.net/browse/PUBDEV-5897?filter=-1


ledell:
@vaclavbelak Thanks, I will review this very soon.

vaclavbelak:
The issues should be fixed now. Can you please have a look again? Thanks!

angela0xdata:
From a doc perspective, the changes look good. No need to rebuild the PDF (or html). Those will be built beginning with the nightly build after this is merged.

PR #3035, opened 2018-11-13
Pavel random forest manual changes revert

Pscheidl:


Pscheidl:
@seb Well, it misbehaves. You can definitely supply just models everywhere else. This way, we put the burden of extracting model_id to H2O users. 

I'll do the change in a separate PR while reviewing R API.

PR #3040, opened 2018-11-13
PUBDEV-5766: Add monotonicity constraints to H2O XGBoost

michalkurka:


h2o-py/tests/testdir_algos/xgboost/pyunit_xgboost_monotone.py
```diff
@@ -0,0 +1,36 @@
...
+    h2oModelS = H2OXGBoostEstimator(**h2oParamsS)
+    h2oModelS.train(x=myX, y=y, training_frame=trainFile)
+
+    native_params = h2oModelS._model_json["output"]["native_parameters"].as_data_frame()
+    print(native_params)
+
+    constraints = (native_params[native_params['name'] == "monotone_constraints"])['value'].values[0]
+
+    assert constraints == u'(-1,0,1,0,0,0,1,0,0,0)'
+
```
wendycwong: Michal:  Do you want to check the output here with native model when you enable monotone_constraint?

michalkurka: that is a good idea - I can certainly do that

PR #3042, opened 2018-11-14
PUBDEV-6039 add mojo version check

maurever:
Add mojo version check to throw an error after the version of h2o does not support it. 
- add a new abstract method `mojoVersion()` to `ModelMojoReader` similar to `ModelMojoWriter`
- implement the `mojoVersion()` method in all readers
- add public parameter `_mojo_version` into `MojoModel` class to check compatibility
- edit `SharedTreeMojoModel` to use `_mojo_version` from `MojoModel`, no defining protected one
- implement `check_mojo_version()` method in `ModelMojoReader` and use it in `readAll()` method after `_mojo_version` parameter is assigned.


wendycwong:
Veronica:

This is an excellent PR.  Can you add tests for it?

Wendy

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
```diff
@@ -49,6 +49,13 @@ public static MojoModel readFrom(MojoReaderBackend reader) throws IOException {
 
   protected abstract M makeModel(String[] columns, String[][] domains, String responseColumn);
 
+  /**
+   * Version of the mojo file produced. Follows the <code>major.minor</code>
```
michalkurka: Is this version of the mojo file?

wendycwong: I think this version is associated with genmodel.jar and represented the h2o-3 version that actually generated the genmodel.jar.

michalkurka: yeah, but the documentation of this method is not correct in my opinion - it is not a version of "a mojo file produced" because this class doesn't produce any mojo file ;) also it is not the version of the mojo file that is being parsed by the reader; that is why I think we need to revise the method documentation

michalkurka: perhaps even rename, this comparison `_model._mojo_version > Double.parseDouble(mojoVersion())` is hard to understand

would be easier if it was maxSupportedMojoVersion() instead or something like that - then there is no question about what the statement is doing. wdyt?

maurever: Oh, that is my mistake. I only copied the doc from the same method from `MojoWriter,` and I didn't check the text. I will change it.

And I also agree with rename the check method. Thanks for catching it.

maurever:
> Veronica:
> 
> This is an excellent PR. Can you add tests for it? You can increase your mojo model version to be higher than we support and make sure it fails or something like that.
> 
> Wendy

Thank you, Wendy. I have not added test yet, but I will do it.

maurever:
I fixed docstring, the name of the method and added a test. The build shows some unsuccessful checks, but I think my changes don't cause them.

h2o-genmodel/src/main/java/hex/genmodel/MojoModel.java
```diff
@@ -18,6 +18,7 @@
   public double _defaultThreshold;
   public double[] _priorClassDistrib;
   public double[] _modelClassDistrib;
+  public double _mojo_version;
```
wendycwong: nice moving this one here instead of in every model!

PR #3043, opened 2018-11-14
PUBDEV-4737 grid search returns wrong model type

maurever:
Edit python client to return correct model type in a grid search.

### DIFF
--- a/h2o-py/h2o/grid/grid_search.py
+++ b/h2o-py/h2o/grid/grid_search.py
@@ -220,6 +220,9 @@ class H2OGridSearch(backwards_compatible()):
         is_unsupervised = is_auto_encoder or algo == "pca" or algo == "svd" or algo == "kmeans" or algo == "glrm"
         if is_auto_encoder and y is not None: raise ValueError("y should not be specified for autoencoder.")
         if not is_unsupervised and y is None: raise ValueError("Missing response")
+        if not is_unsupervised:
+            y = y if y in training_frame.names else training_frame.names[y]
+            self.model._estimator_type = "classifier" if training_frame.types[y] == "enum" else "regressor"
         self._model_build(x, y, training_frame, validation_frame, algo_params)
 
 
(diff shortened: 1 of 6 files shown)


h2o-py/h2o/grid/grid_search.py
```diff
@@ -220,6 +220,9 @@ def build_model(self, algo_params):
         is_unsupervised = is_auto_encoder or algo == "pca" or algo == "svd" or algo == "kmeans" or algo == "glrm"
```
michalkurka: This statement is missing 'isolation forest', this code is hard to maintain - similar a code is found in the train() inside EstimatorBase: https://github.com/h2oai/h2o-3/blob/master/h2o-py/h2o/estimators/estimator_base.py#L134

Can you please find out if the code can be refactored so that there is just a single place where we detect if the algo is supervised? Same holds true for estimator type (that should be in one place too).

h2o-py/tests/testdir_algos/deeplearning/pyunit_iris_grid_deeplearning.py
```diff
@@ -37,6 +37,10 @@ def iris_dl_grid():
     assert combo in total_grid_space
     total_grid_space.remove(combo)
 
+  print("Check correct type value....")
+  model_type = gs[0].type
+  true_model_type = "classifier"
```
michalkurka: perfect!

michalkurka:
Looks good, thank you. I think it can be improved because similar code can be found in 2 places. If you want to please go ahead and merge this PR - but I would like to see if there is a way to avoid duplicating the code.

h2o-py/h2o/grid/grid_search.py
maurever: I found this part of the code, and that is why this bug occurred - the _estimator_type is set after requesting the `h2o.api()` only in the `Estimator.train()` method in Python but this method is not called in grid search...

I think a better way is returning _estimator_type in JSON response - I looked into the response, but there was no information about the type... It could also improve controlling which type of algorithm which model uses (so it should not happen some type is missing like isolation forest). I realized that it would be a much more significant change, so I did it similar to `train` method, and I wanted to discuss it after the build is done and ok... 

Another question is how is this issue implemented in R; there could also be some code duplicates... I will look at it and I will try to find a better solution.

michalkurka: I am not sure why the REST API is not returning the information but some algos (deeplearning) can be both supervised and unsupervised (autoencoder). Perhaps that is why. Would be nice to have a better solution than this kind of explicit list.

PR #3045, opened 2018-11-15
JDBC test improvement - keep derby tempfiles inside build

pkozelka:
Little cleanup to reduce source tree polution caused by JDBC test execution:

- `derby.log` is now created inside `build` directory
- derby databases are created in a temporary folder, also under `build` directory
- each DB testsuite uses its own database name, to reduce influence between the tests

### DIFF
--- a/h2o-core/src/test/java/water/jdbc/SQLManagerIntegTest.java
+++ b/h2o-core/src/test/java/water/jdbc/SQLManagerIntegTest.java
@@ -12,13 +12,16 @@ import water.fvec.Frame;
 import water.fvec.TestFrameBuilder;
 import water.fvec.Vec;
 
+import java.io.File;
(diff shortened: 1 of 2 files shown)


pkozelka:
what I found in existing `build/tmp` dirs looks more like compilation related, so I thought it's better to leave this at top, so that it is easier to notice. We can easily change it anytime later, for instance if we introduce any kind of conventional directory structure for test runtime files. However, if you insist, I can do that anyway, I have no strong opinions here - except that it should be somewhere under `build`.

michalkurka:
oh, okay - if the purpose is not to archive the file (on jenkins) - I would suggest to put it in the junit temporary directory instead - that is cleared after each test execution

that sounds like a better option to me, the test already does that for the database itself

PR #3051, opened 2018-11-16
PUBDEV-6054 - Fixed unclosed leaked socket in RabitWorker

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6054

Connections are not being closed, as pointer to tracker thread is being deleted prematurely. The resources were garbage collected, but sometimes not quickly enough. In case of building many small models at once, the resources were never freed in time, resulting in SocketChannels holding file descriptors allocated on Linux.


Pscheidl:
An attempt to make the test quicker using Ulimit.

```
The latest docker supports setting ulimits through the command line and the API. For instance, docker run takes --ulimit <type>=<soft>:<hard> and there can be as many of these as you like. So, for your nofile, an example would be --ulimit nofile=262144:262144
```

Pscheidl:
Moved the stress test to the smoke stage (thanks @michal-raska ) to see the output fast: http://mr-0xc1:8080/job/h2o-3-pipeline/job/PR-3051/9/

Now moved to nigthly. The test itself is not that slow, finishes in +- 5 minutes. Yet I'd leave it in nightly.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitTrackerH2O.java
```diff
@@ -71,9 +71,8 @@ public boolean start(long timeout) {
             throw new IllegalStateException("Rabit tracker already started.");
         }
         RabitTrackerH2OThread trackerThread = new RabitTrackerH2OThread(this);
-        trackerThread.setDaemon(true);
```
Pscheidl: There is no reason for this thread to be a daemon thread. It should always be killed in time. And if JVM with H2O is being stopped, the thread should be stopped too.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitTrackerH2O.java
```diff
@@ -71,9 +71,8 @@ public boolean start(long timeout) {
...
-        trackerThread.start();
         this.trackerThread = trackerThread;
+        trackerThread.start();
```
Pscheidl: First assign it, then start it. If start() throws an unechecked exception, the pointer is still there in `this.trackerThread` and can be dealth with later.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitTrackerH2O.java
```diff
@@ -20,7 +20,7 @@
 
     private Map<String, String> envs = new HashMap<>();
 
-    private RabitTrackerH2OThread trackerThread;
+    private volatile RabitTrackerH2OThread trackerThread;
```
Pscheidl: Red from both the thread and this class. Better guarantee single place of existence.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitTrackerH2O.java
```diff
@@ -71,26 +71,28 @@ public boolean start(long timeout) {
...
         this.trackerThread = trackerThread;
+        trackerThread.start();
         return true;
     }
 
     @Override
     public void stop() {
-        if(null != this.trackerThread) {
-            this.trackerThread.interrupt();
+        assert this.trackerThread != null;
```
Pscheidl: Null check removed. This is never assigned null before this method's invocation now.

In fact, it should never ever be null at this point. That's why there is this new assert.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitTrackerH2O.java
```diff
@@ -71,26 +71,28 @@ public boolean start(long timeout) {
...
-                this.sock.close();
-            } catch (IOException e) {
-                throw new RuntimeException("Failed to close Rabit tracker socket.", e);
-            } finally {
-                this.port = 9091;
-            }
+
+        try {
+            this.sock.close();
+            this.port = 9091;
```
Pscheidl: Only reset the port to default when socket closing is successful. 

When unsuccessful, log an error and continue, as there may be enough resources to finish the task. The error will be in the logs still.

scripts/jenkins/groovy/defineTestStages.groovy
```diff
@@ -311,6 +311,10 @@ def call(final pipelineContext) {
     [
       stageName: 'R3.3 CMD Check as CRAN', target: 'test-r-cmd-check-as-cran', rVersion: '3.3.3',
       timeoutValue: 10, hasJUnit: false, component: pipelineContext.getBuildConfig().COMPONENT_R
+    ],
+    [ // These run with reduced number of file descriptors for early detection of FD leaks
+      stageName: 'XGBoost Stress tests', target: 'test-pyunit-xgboost-stress', pythonVersion: '3.5', timeoutValue: 20,
+      component: pipelineContext.getBuildConfig().COMPONENT_PY, customDockerArgs: [ '--ulimit nofile=100:100' ]
```
Pscheidl: Notice the `'--ulimit nofile=100:100'`. Makes this test fail fast if there is something wrong with resources not being freed. If these tests are ran without the fixes, they fail in less than 100 models built (as some of the descriptors are already taken by the system).

h2o-py/tests/testdir_algos/xgboost/stress/pyunit_xgboost_airlines_stress.py
```diff
@@ -0,0 +1,18 @@
+import h2o
```
michalkurka: should we mark this test as large?

michalkurka: sorry, I didn't notice there was already a special category for it

scripts/jenkins/Makefile.jenkins
```diff
@@ -29,7 +29,7 @@ test-package-py:
 	zip -q -r test-package-py.zip h2o-py/tests/ h2o-py/demos/ \
 	h2o-docs/src/booklets/v2_2015/source h2o-py/build/dist/*.whl \
 	h2o-genmodel/build/libs/h2o-genmodel.jar h2o-assemblies/genmodel/build/libs/genmodel.jar scripts/run.py \
-	$(JENKINS_MAKEFILE_PATH) tests/pyunitSmokeTestList tests/pyunitAutoMLList tests/pyunitSingleNodeList h2o-py/scripts/h2o-py-test-setup.py
```
michalkurka: @michal-raska can you please check this?

scripts/jenkins/Makefile.jenkins
```diff
@@ -59,6 +59,10 @@ test-pyunit-single-node:
 	cd h2o-py/tests/ && rm -f -r testdir_hdfs
 	export NO_GCE_CHECK=True && cd h2o-py/tests/ && ../../scripts/run.py --wipeall --geterrs --testsize s --numclouds 1 --numnodes 1 --jvm.xmx 3g --testlist ../../tests/pyunitSingleNodeList
 
+test-pyunit-xgboost-stress:
```
michalkurka: great!

michalkurka:
great work! thank you

approved from my side - please let @michal-raska review the build changes

Pscheidl:
@pkozelka also reviewed, gave thumbs up :)

PR #3052, opened 2018-11-16
PUBDEV-6065: Leaderboard throws ArrayIndexOutOfBoundsException if sort_metric is specified but no model was built

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6065


seb-h2o:
@ledell : FYI, discovered this bug when working on benchmarks during the week

PR #3054, opened 2018-11-16
PUBDEV-6062: fix multinomial coefficient names

wendycwong:
This PR completes the work demanded in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6062?filter=-1.

Work done:
1. In Java backend, force coefficient name to contain the actual class name instead of 0, 1, 2,...
2. Added R/Python unit test to verify correct coefficient names are returned.

### DIFF
--- a/h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
+++ b/h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
@@ -31,6 +31,8 @@ public class GLMModelV3 extends ModelSchemaV3<GLMModel, GLMModelV3, GLMModel.GLM
     @API(help="Lambda best + 1 standard error. Only applicable with lambda search and cross-validation")
     double lambda_1se;
 
+    int _responseColumn; // indicates the response column index, do not use, internal only
+
     private GLMModelOutputV3 fillMultinomial(GLMOutput impl) {
       if(impl.get_global_beta_multinomial() == null)
         return this; // no coefificients yet
@@ -42,9 +44,10 @@ public class GLMModelV3 extends ModelSchemaV3<GLMModel, GLMModelV3, GLMModel.GLM
       if(impl.isStandardized()){
         int n = impl.nclasses();
         String [] cols = new String[n*2];
+        String[] classNames = impl._domains[_responseColumn];
         for(int i = 0; i < n; ++i) {
-          cols[i] = "Coefs_class_" + i;
-          cols[n+i] = "Std_Coefs_class_" + i;
+          cols[i] = "Coefs_class_" + classNames[i];
+          cols[n+i] = "Std_Coefs_class_" + classNames[i];
         }
         String [] colTypes = new String[cols.length];
         Arrays.fill(colTypes, "double");
--- a/h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
+++ b/h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
@@ -22,6 +22,9 @@ public class GLMModelV3 extends ModelSchemaV3<GLMModel, GLMModelV3, GLMModel.GLM
     @API(help="Table of Coefficients")
     TwoDimTableV3 coefficients_table;
 
+    @API(help="Table of Coefficients with coefficients denoted with class names for GLM multinonimals only.")
+    TwoDimTableV3 coefficients_table_multinomials_with_class_names;  // same as coefficients_table but with real class names.
+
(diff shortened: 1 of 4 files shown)


ledell:
Here's an R example of current behavior:

```
> library(h2o)
> h2o.init()
> fit <- h2o.glm(y = 5, training_frame = as.h2o(iris), family = "multinomial")
> fit@model$coefficients_table
Coefficients: glm multinomial coefficients
         names coefs_class_0 coefs_class_1 coefs_class_2 std_coefs_class_0 std_coefs_class_1 std_coefs_class_2
1    Intercept      6.496784     -0.408705    -18.679065         -3.969984         -0.016699         -8.596793
2 Sepal.Length     -1.522256      1.109945      0.000000         -1.260529          0.919108          0.000000
3  Sepal.Width      3.757638     -0.684351     -3.575675          1.637828         -0.298285         -1.558516
4 Petal.Length     -2.059447     -0.596759      3.326552         -3.635537         -1.053457          5.872356
5  Petal.Width     -4.436355     -1.466536      7.098211         -3.381557         -1.117849          5.410524
```

ledell:
This could definitely break people's model post-processing code if we change it, so we need to consider this.

wendycwong:
@ledell   I agree with you.  Added a new twodim table: coefficients_table_multinomial_with_class_names

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -39,13 +42,26 @@ private GLMModelOutputV3 fillMultinomial(GLMOutput impl) {
       String [] ns = ArrayUtils.append(new String[]{"Intercept"},Arrays.copyOf(names,names.length-1));
 
       coefficients_table = new TwoDimTableV3();
+      if (impl._responseColumn >= 0) // _responseColumn -1 for regression and binomial.
```
michalkurka: can you use _nClass > 2 instead? do we really need _responseColumn>

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -816,6 +816,7 @@ public void setZValues(double [] zValues, double dispersion, boolean dispersionE
     double [][] _vcov;
     private double _dispersion;
     private boolean _dispersionEstimated;
+    public int _responseColumn=-1; // represent response column index, only >= 0 when nclass > 2
```
michalkurka: I think this is a little misleading, it is not natural to expect that _responseColumn would be populated only for multiclass problems.

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
wendycwong: Yes, changed it.

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: Removed .

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -113,9 +143,26 @@ public int compare(Integer o1, Integer o2) {
...
       }
+
       return this;
     }
+
+    public void revertCoeffNames(String[] colNames, int nclass, TwoDimTableV3 coeffs_table) {
+      String newName = coeffs_table.name+" with class names";
+      coeffs_table.name = newName;
+      boolean bothCoeffStd = colNames.length==(2*nclass);
+      for (int tableIndex = 1; tableIndex <= nclass;  tableIndex++) {
```
Pscheidl: Why do we iterate from 1 to nclass and then do `tableIndex - 1 at each line of occurance ? Wouldn't it be better to just iterate from 0 ?

Pscheidl:
@wendycwong Approved but please see my one comment. I found a small issue (possibly).

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
wendycwong: Pavel:  The reason for the tableIndex starting from 1 is because for the coeffs_table.columns array, the first name is always "name".  The only useful assignment starts from index 1 to nclass.  So, if I set tableIndex to start from 1, I will need to do coeffs_table.columns[tableIndex+1].  Either way, I have to manipulate the tableIndex to either +1 or -1.  Since we are not looking for speed here, I just pick to do the tableIndex to start from 1.  Good observation!

PR #3056, opened 2018-11-16
Cleanup - move temporary directory 'lib' under the build directory

pkozelka:
Temporary directories should always be under a `build` dir.

This PR does this for toplevel `lib` which is now `build/lib` instead.

### DIFF
--- a/h2o-algos/testSSL.sh
+++ b/h2o-algos/testSSL.sh
@@ -108,7 +108,7 @@ pwd
 tshark -i ${INTERFACE} -T fields -e data -w ${OUTDIR}/h2o-nonSSL.pcap 1> /dev/null 2>&1 & PID_4=$!
 
 java -Dai.h2o.name=$CLUSTER_NAME -ea \
-    -cp "build/libs/h2o-algos-test.jar${SEP}build/libs/h2o-algos.jar${SEP}../h2o-core/build/libs/h2o-core.jar${SEP}../h2o-core/build/libs/h2o-core-test.jar${SEP}../h2o-genmodel/build/libs/h2o-genmodel.jar${SEP}../lib/*" \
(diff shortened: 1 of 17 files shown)


michalkurka:
I wonder if the `lib` directory is really needed. It is still used by `testSSL.ssh` (this PR should fail in that test if it is executed somewhere still ;)).

Otherwise other testMulti(/Single)Node.sh have been migrated and take classpath from elsewhere. @michal-raska 

We could probably get rid off it.

h2o-algos/testSSL.sh
```diff
@@ -108,7 +108,7 @@ pwd
 tshark -i ${INTERFACE} -T fields -e data -w ${OUTDIR}/h2o-nonSSL.pcap 1> /dev/null 2>&1 & PID_4=$!
 
 java -Dai.h2o.name=$CLUSTER_NAME -ea \
-    -cp "build/libs/h2o-algos-test.jar${SEP}build/libs/h2o-algos.jar${SEP}../h2o-core/build/libs/h2o-core.jar${SEP}../h2o-core/build/libs/h2o-core-test.jar${SEP}../h2o-genmodel/build/libs/h2o-genmodel.jar${SEP}../lib/*" \
+    -cp "build/libs/h2o-algos-test.jar${SEP}build/libs/h2o-algos.jar${SEP}../h2o-core/build/libs/h2o-core.jar${SEP}../h2o-core/build/libs/h2o-core-test.jar${SEP}../h2o-genmodel/build/libs/h2o-genmodel.jar${SEP}../build/lib/*" \
```
michalkurka: Can we use project classpath just like we do it in multiNode/singleNode tests? Then we wouldn't need to have a build directory at all.

michalkurka:
Seems like we can get rid of the `lib` directory completely by fixing `testSSL.sh`

PR #3058, opened 2018-11-20
PUBDEV-5236 add the missing transform method

maurever:
In the `H2OPrincipalComponentAnalysisEstimator` (`pca.py`) the method `transform` is missing and the `@property transform` was called instead. It caused `None` was returned. The H2OPCA works well because it doesn't have `@property transform` (the parameter `transform` is set in the `init `method and cannot be changed). 

JIRA: https://0xdata.atlassian.net/browse/PUBDEV-5236

**The first idea**
- generate transform method into `pca.py`
- problem: overloading doesn't work even though the tests passed. When the user sets the transform property, a pipeline uses this property method when is running. 

**The second idea**
- rename the  `@property transform` to `@property transformation`
- problem: it is not user-friendly, it breaks all estimators usage, the doc should be changed, etc...

**The third idea**
- use in pipeline different object (adapter or wrapper) which implement `fit` and `transform` method properly
- problem: user have to use this object instead of estimator object
- solution: use in `sklearn.Pipeline` the old implementation of `H2OPCA` which works well, but can't be customized after initialization -> implement method `init_for_pipeline` in 'H2OPrincipalComponentAnalysisEstimator' which return `H2OPCA` initialized from 'H2OPrincipalComponentAnalysisEstimator' properties. 

The final solution is not so pretty, but enable to users to use a transform estimator in `sklearn.Pipeline` with a little change. 

**This is not the problem only with PCA but also with GLRM, SVD and Aggregator.**

### DIFF
--- a/h2o-bindings/bin/gen_python.py
+++ b/h2o-bindings/bin/gen_python.py
@@ -559,9 +559,12 @@ def class_extra_for(algo):
 
             :returns: H2OPCA object
             \"\"\"
-            
+            import inspect
             from h2o.transforms.decomposition import H2OPCA
-            return H2OPCA(**self._parms)
+            # check which parameters can be passed to H2OPCA init
+            var_names = list(dict(inspect.getmembers(H2OPCA.__init__.__code__))['co_varnames'])
+            parameters = {k: v for k, v in self._parms.items() if k in var_names}
+            return H2OPCA(**parameters)
         """
 
 
--- a/h2o-bindings/bin/gen_python.py
+++ b/h2o-bindings/bin/gen_python.py
@@ -552,18 +552,24 @@ def class_extra_for(algo):
         """
     elif algo == "pca":
         return """
-        def fit_transform(self, X, y=None, **params):
+        def init_for_pipeline(self):
             \"\"\"
-            Fit and transform the given H2OFrame with the fitted PCA model.
-            This method has to be implemented to be used with sklearn.pipeline module.
+            Returns H2OPCA object which implements fit and transform method to be used in sklearn.Pipeline properly.
 
-            :param H2OFrame X: May contain NAs and/or categorical data.
-            :param H2OFrame y: Ignored for PCA. Should be None.
-            :param params: Ignored.
-
-            :returns: The input H2OFrame transformed by the Principal Components.
+            :returns: H2OPCA object
             \"\"\"
-            return self.fit(X, y, **params).predict(X)
+            from h2o.transforms.decomposition import H2OPCA
+            return H2OPCA(model_id=self.model_id,
+                            k=self.k,
+                            max_iterations=self.max_iterations,
+                            seed=self.seed,
+                            transform="None" if self.transform is None else self.transform,
+                            use_all_factor_levels=False if self.use_all_factor_levels is None else self.use_all_factor_levels,
+                            pca_method="GramSVD" if self.pca_method is None else self.pca_method,
+                            pca_impl="mtj_evd_symmmatrix" if self.pca_impl is None else self.pca_impl,
+                            ignore_const_cols=True if self.ignore_const_cols is None else self.ignore_const_cols,
+                            impute_missing=False if self.impute_missing is None else self.impute_missing,
+                            compute_metrics=True if self.compute_metrics is None else self.compute_metrics)
         """
 
 
--- a/h2o-bindings/bin/gen_python.py
+++ b/h2o-bindings/bin/gen_python.py
@@ -552,9 +552,9 @@ def class_extra_for(algo):
         """
     elif algo == "pca":
         return """
-        def transform(self, X, y=None, **params):
+        def fit_transform(self, X, y=None, **params):
             \"\"\"
-            Transform the given H2OFrame with the fitted PCA model.
+            Fit and transform the given H2OFrame with the fitted PCA model.
             This method has to be implemented to be used with sklearn.pipeline module.
 
             :param H2OFrame X: May contain NAs and/or categorical data.
@@ -563,7 +563,7 @@ def class_extra_for(algo):
 
             :returns: The input H2OFrame transformed by the Principal Components.
             \"\"\"
-            return self.predict(X)
+            return self.fit(X, y, **params).predict(X)
         """
 
 
--- a/h2o-bindings/bin/gen_python.py
+++ b/h2o-bindings/bin/gen_python.py
@@ -555,21 +555,13 @@ def class_extra_for(algo):
         def init_for_pipeline(self):
             \"\"\"
             Returns H2OPCA object which implements fit and transform method to be used in sklearn.Pipeline properly.
+            All parameters defined in self.__params, should be input parameters in H2OPCA.__init__ method.
 
             :returns: H2OPCA object
             \"\"\"
+            
             from h2o.transforms.decomposition import H2OPCA
-            return H2OPCA(model_id=self.model_id,
-                            k=self.k,
-                            max_iterations=self.max_iterations,
-                            seed=self.seed,
-                            transform="None" if self.transform is None else self.transform,
-                            use_all_factor_levels=False if self.use_all_factor_levels is None else self.use_all_factor_levels,
-                            pca_method="GramSVD" if self.pca_method is None else self.pca_method,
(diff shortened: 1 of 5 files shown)


h2o-py/h2o/estimators/pca.py
```diff
@@ -268,4 +268,16 @@ def max_runtime_secs(self, max_runtime_secs):
         assert_is_type(max_runtime_secs, None, numeric)
         self._parms["max_runtime_secs"] = max_runtime_secs
 
+    def transform(self, X, y=None, **params):
```
maurever: It is not saved to have the property named the same as this method. This method is used in `sklearn.pipeline` and cannot be refactored. How important is to have this property in H2OPrincipalComponentAnalysisEstimator? https://github.com/h2oai/h2o-3/blob/682fdfcc7375e2dfab15b45c512491ecd21985d7/h2o-py/h2o/estimators/pca.py#L117

michalkurka: Not sure if I understand. What do you suggest to do instead?

h2o-py/h2o/estimators/pca.py
```diff
@@ -269,3 +269,15 @@ def max_runtime_secs(self, max_runtime_secs):
         self._parms["max_runtime_secs"] = max_runtime_secs
 
 
+
+    def transform(self, X, y=None, **params):
+        """
+        Transform the given H2OFrame with the fitted PCA model.
```
michalkurka: Maybe we can state the purpose of the method? that it is used in sklearn pipelines?

h2o-py/h2o/estimators/pca.py
maurever: I am not a Python guru, but I think there should not be two methods with the same name and absolutely different functionality. There are property `transform(self)` which return value from `parms` dictionary; and now also new method `transform(self, X, y=None, **params)` which calculates something. Before this change, the problem was `sklearn.Pipeline` at one moment call `transform(X)` function on estimator and the property `transform(self)` method returned None (or string with a different setting) instead of a frame... So my question is, how important is to have `transform(self)` property called `transform`? For example, the better name for this property is `transformation` or something like this. I am not sure, how big impact this refactoring will have on usage PCA estimator.

michalkurka: I see. This is unfortunate. I think we should deprecate the transform parameter and introduce a new name instead. We will have to keep the old one for a little bit for backwards compatibility. Can you please make a jira issue?

Otherwise if everything is working - I would tolerate the conflict for a while. But we do need to address it eventually.

h2o-py/h2o/estimators/pca.py
maurever: Yes, that is a good point. I changed it.

h2o-py/h2o/estimators/pca.py
maurever: It seems to be working, but it doesn't. If the user set the `transform` property after creating estimator and before using in a pipeline, it will cause the same error.

```python
pca = H2OPCA(k=2)
pca.transform = None
pipe = Pipeline([("standardize", H2OScaler()),
                     ("pca", pca),
                     ("rf", H2ORandomForestEstimator(seed=42,ntrees=50))])
pipe.fit(iris[:4],iris[4])`
```

maurever:
New problems occurred, it is not ready to merge, and this issue is still in progress.

maurever:
New usage of `H2OPrincipalComponentAnalysisEstimator` in a `sklearn.Pipeline:`

```python
  from h2o.transforms.preprocessing import H2OScaler
  from h2o.estimators.pca import H2OPrincipalComponentAnalysisEstimator as H2OPCA
  from h2o.estimators.random_forest import H2ORandomForestEstimator
  from sklearn.pipeline import Pipeline

  iris = h2o.import_file(path=pyunit_utils.locate("smalldata/iris/iris_wheader.csv"))
 
  pipe = Pipeline([("standardize", H2OScaler()),
  ("pca", H2OPCA(k=2).init_for_pipeline()),
  ("rf", H2ORandomForestEstimator(seed=42,ntrees=50))])
  pipe.fit(iris[:4], iris[4])
```
Can you check if it is a acceptable solution, @wendycwong, @michalkurka, please? Thank you.

If yes, I will change GLRM, SVD, and Aggregator Estimators too, and will add a new pyunit test for this changes (e.g., there is no test of these algos in the sklearn.Pipeline yet).

h2o-bindings/bin/gen_python.py
```diff
@@ -550,6 +550,28 @@ def _determine_vec_size(self):
...
+    elif algo == "pca":
+        return """
+        def init_for_pipeline(self):
+            \"\"\"
+            Returns H2OPCA object which implements fit and transform method to be used in sklearn.Pipeline properly.
+
+            :returns: H2OPCA object
+            \"\"\"
+            from h2o.transforms.decomposition import H2OPCA
+            return H2OPCA(model_id=self.model_id,
```
michalkurka: Are all the possible parameters propagated to H2OPCA?

We should make sure none of the parameters get lost.

h2o-py/tests/testdir_pipeline/pyunit_scale_pca_rf.py
```diff
@@ -21,10 +18,50 @@ def scale_pca_rf_pipe():
                    ("pca", H2OPCA(k=2)),
                    ("rf", H2ORandomForestEstimator(seed=42,ntrees=50))])
   pipe.fit(iris[:4],iris[4])
+  print(pipe)
+
+
+def scale_pca_rf_pipe_new_import():
+
+  from h2o.transforms.preprocessing import H2OScaler
+  from h2o.estimators.pca import H2OPrincipalComponentAnalysisEstimator as H2OPCA
```
michalkurka: I think the alias "H2OPCA" makes the test a little bit harder to read (because there already is an H2OPCA class)

michalkurka:
@maurever I like this approach, please see inline comments

h2o-py/tests/testdir_pipeline/pyunit_scale_pca_rf.py
maurever: Ok, I will change it.

h2o-bindings/bin/gen_python.py
maurever: That's the question. Because if we want to keep both algos (`H2OPCA` and `H2OPrincipalComponentAnalysisEstimator`), a developer would have always checked if the parameters fit in both of implementation. But I understand and try to find a better solution. Thank you.

maurever:
So, I edit `init_for_pipeline` method to create `H2OPCA` dynamically with parameters which  `H2OPCA.__init__ `accepts. 

I found out the parameters for `H2OPCA` and `H2OPrincipalComponentAnalysisEstimator` are different. First I unified it, but then I decided not to do this change even though the tests passed. That is the question: Should be the parameters the same or not? If yes, the new JIRA should be created for this bug. 

What do you think about it, @michalkurka? 

If this change is acceptable, I will finally implement also this in GLRM, SVD, and Aggregator Estimators.

maurever:
The problem for the PCA algorithm is solved so that this PR can be merged.

But after a quick exploration of the other algorithms (GLRM, SVD, and Aggregator Estimators) I found out, there are another problems. 

The main problem still is, that these estimators don't implement `transform` method properly and can't be used in `sklearn.pipeline` properly. 

But for example, only initialization `H2OSingularValueDecompositionEstimator` throws an error - no test exists for its implementation let alone for its usage in a sklern.pipeline. But the solution of transform problem will be the same as for PCA.

Tests for `H2OAggregatorEstimator` exist, but there is no pipeline test case. When H2OAggregatorEstimator is passed to pipeline, it failed when the `H2OEstimator.fit` method is called (the `self` is `DataFrame` instead of `H2OEstimator`). 

For `H2OAggregatorEstimator` and `H2OGeneralizedLowRankEstimator` should be used a different approach to solve transform problem because there is no equivalent like H2OPCA or H2OSVD.

Also, documentation about usage H2OEstimators in the sklearn.pipeline is missing.

I will create new JIRAs to inspect these problems and solve them. 

Thanks @michalkurka to help me solve this PR.

PR #3060, opened 2018-11-20
[WIP-PUBDEV-6068] Use shared library for pipelines.

michal-raska:


michalkurka:
@michal-raska is this still relevant?

michal-raska:
It'll be good to use the shared library, some of the features from H2O-3 pipeline has been ported (and improved) there. But this PR is kind of stale, probably will be easier to just make a new one.

PR #3062, opened 2018-11-20
PUBDEV-5956: Fixing keys leakage in KeepCrossValidationFoldAssignment test

deil87:
New version of PR https://github.com/h2oai/h2o-3/pull/2904 but rel-xia based.

### DIFF
--- a/h2o-algos/src/test/java/hex/CrossValidFoldAssignmentsTest.java
+++ b/h2o-algos/src/test/java/hex/CrossValidFoldAssignmentsTest.java
@@ -6,10 +6,13 @@ import hex.tree.gbm.GBMModel;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import water.DKV;
+import water.MRTask;
 import water.TestUtil;
+import water.fvec.Chunk;
 import water.fvec.Frame;
 import water.fvec.Vec;
 import water.rapids.ast.prims.advmath.AstKFold;
+import water.util.ArrayUtils;
 
 import static junit.framework.TestCase.assertNotNull;
 import static org.junit.Assert.assertEquals;
@@ -98,4 +101,62 @@ public class CrossValidFoldAssignmentsTest extends TestUtil {
       if (cvFoldAssignmentFrame != null) cvFoldAssignmentFrame.delete();
     }
   }
+
+  // Checks that implicitly generated fold column is preserved after model is built
+  @Test public void checkImplicitFoldAssignmentsAreKeptWithoutMakeCopy() {
+    final int nfolds = 3;
+    Frame tfr = null;
+    Frame cvFoldAssignmentFrame = null;
+    GBMModel gbm = null;
+    try {
+      tfr = parse_test_file("smalldata/iris/iris_wheader.csv");
+
+      GBMModel.GBMParameters parms = new GBMModel.GBMParameters();
+      parms._train = tfr._key;
+      parms._response_column = "class";
+      parms._ntrees = 1;
+      parms._max_depth = 1;
+      parms._nfolds = nfolds;
+      parms._distribution = DistributionFamily.multinomial;
+      parms._keep_cross_validation_predictions=false;
+      parms._keep_cross_validation_fold_assignment=true;
+      GBM job = new GBM(parms);
+      gbm = job.trainModel().get();
+
+      assertNotNull(gbm._output._cross_validation_fold_assignment_frame_id);
+      cvFoldAssignmentFrame = DKV.getGet(gbm._output._cross_validation_fold_assignment_frame_id);
+      assertNotNull(cvFoldAssignmentFrame);
+      assertEquals(tfr.numRows(), cvFoldAssignmentFrame.numRows());
+      assertEquals(tfr.numRows(), ArrayUtils.sum(new CheckFoldTask(nfolds).doAll(cvFoldAssignmentFrame)._foldCnt));
+    } finally {
+      if (tfr != null) tfr.remove();
+      if (gbm != null) {
+        gbm.delete();
+        gbm.deleteCrossValidationModels();
+      }
+      if (cvFoldAssignmentFrame != null) cvFoldAssignmentFrame.delete();
+    }
+  }
+
+  private static class CheckFoldTask extends MRTask<CheckFoldTask> {
+    private long[] _foldCnt;
+    private CheckFoldTask(int nfolds) {
+      _foldCnt = new long[nfolds];
+    }
+    @Override
+    public void map(Chunk c) {
+      for (int i = 0; i < c._len; i++) {
+        double val = c.atd(i);
+        if ((int) val != val || val < 0 || val > _foldCnt.length) {
+          throw new IllegalStateException("Unexpected value: " + val);
+        }
+        _foldCnt[(int) val]++;
+      }
+    }
+    @Override
+    public void reduce(CheckFoldTask mrt) {
+      _foldCnt = ArrayUtils.add(_foldCnt, mrt._foldCnt);
+    }
+  }
+
 }
--- a/h2o-algos/src/test/java/hex/CrossValidFoldAssignmentsTest.java
+++ b/h2o-algos/src/test/java/hex/CrossValidFoldAssignmentsTest.java
@@ -1,0 +1,101 @@
+package hex;
+
+import hex.genmodel.utils.DistributionFamily;
+import hex.tree.gbm.GBM;
+import hex.tree.gbm.GBMModel;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import water.DKV;
+import water.TestUtil;
+import water.fvec.Frame;
+import water.fvec.Vec;
+import water.rapids.ast.prims.advmath.AstKFold;
+
+import static junit.framework.TestCase.assertNotNull;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class CrossValidFoldAssignmentsTest extends TestUtil {
+  @BeforeClass() public static void setup() { stall_till_cloudsize(1); }
+
+  // Check that we should not make a copy of fold assignments Vec in ModelBuilder.
+  // Fold column as a part of train dataset is not being tracked by Scope.enter() and is still available
+  // after the model is trained.
+  @Test public void checkFoldAssignmentsAreKeptWithoutMakeCopy() {
+    final int nfolds = 3;
+    Frame tfr = null;
+    Frame cvFoldAssignmentFrame = null;
+    Frame foldId = null;
+    GBMModel gbm = null;
+    try {
+      tfr = parse_test_file("smalldata/iris/iris_wheader.csv");
+      foldId = new Frame(new String[]{"foldId"}, new Vec[]{AstKFold.kfoldColumn(tfr.vec("class").makeZero(), nfolds, 543216789)});
+      tfr.add(foldId);
+      DKV.put(tfr);
+
+      GBMModel.GBMParameters parms = new GBMModel.GBMParameters();
+      parms._train = tfr._key;
+      parms._response_column = "class";
+      parms._ntrees = 1;
+      parms._max_depth = 1;
+      parms._fold_column = "foldId";
+      parms._distribution = DistributionFamily.multinomial;
+      parms._keep_cross_validation_predictions=false;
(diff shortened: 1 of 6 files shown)


h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -465,7 +465,7 @@ public Vec cv_AssignFold(int N) {
       }.doAll(2*N,Vec.T_NUM,folds_and_weights).outputFrame().vecs();
 
     if (_parms._keep_cross_validation_fold_assignment)
-      DKV.put(new Frame(Key.<Frame>make("cv_fold_assignment_" + _result.toString()), new String[]{"fold_assignment"}, new Vec[]{foldAssignment.makeCopy()}));
```
michalkurka: Why do we don't need to make a copy? Can you please explain this change?

seb-h2o: my 2 cents here: foldAssignment is not manipulated after this line, so making a copy looks over-defensive coding.
Although, especially in relation with the following line, it could be make more explicit:
```
if (_parms._keep_cross_validation_fold_assignment)
      DKV.put(new Frame(Key.<Frame>make("cv_fold_assignment_" + _result.toString()), new String[]{"fold_assignment"}, new Vec[]{foldAssignment}));
else if( _parms._fold_column == null) foldAssignment.remove();
```
even wondering if last line should not be simply
`else foldAssignment.remove();`

deil87: I would put it this way, I don't understand why .makeCopy() was there before. This code was added in scope of PUBDEV-2774 task. And there is no test that checks this makeCopy addition. We also don't have test in Vec for makeCopy method. I have following considerations:
1) Do we really need to copy? if there is a case when we remove (train) data and still need fold assignments? as I understand that fold assignment frame contains only one vector (with number of fold as values) and it is of no use without data.
2) I'm not sure that makeCopy works as expected.... i tried to add `Vec leaked = foldAssignment.makeCopy();` line and the test did not fail with leaking keys error.
3) I can't check that all the tests are fine without makeCopy() ... CI tests keep hang for different reasons. But If the tests are fine then I kind of can change the code however I want to :slightly_smiling_face:
Summary: making copy is expensive and we should try to avoid doing it whenever we can. Using fold assignments without data doesn't make much sense ( at least based on my current understanding).

michalkurka: @seb-h2o @deil87 good points, I also don't see an obvious reason. 

One think to keep in mind is Scope, the Vec might be tracked by Scope and can be nuked after the Scope exits. One possibility is that the author wanted to prevent this - that would be my guess.

I don't junits will fail on this because they usually don't do anything after the model is built. But we might see something in runits/pyunits.

I will look into it as well.

deil87: @michalkurka yes, you are right. foldassignment is being tracked by Scope from computeCrossValidation method. If author had an intention like you mentioned above then they should have used Scope.untrack on the copy.

michalkurka: @deil87 yes, please add a test - that is the best way to figure this out

deil87: @michalkurka done, please review the tests

michalkurka: thanks, will review!

h2o-automl/src/test/java/ai/h2o/automl/AutoMLTest.java
```diff
@@ -138,27 +140,61 @@
...
       aml = AutoML.makeAutoML(Key.<AutoML>make(), new Date(), autoMLBuildSpec);
       AutoML.startAutoML(aml);
       aml.get();
 
-      assertTrue(aml.leader() !=null && aml.leader()._parms._keep_cross_validation_fold_assignment);
-      assertTrue(aml.leader() !=null && aml.leader()._output._cross_validation_fold_assignment_frame_id != null);
+      leader = aml.leader();
+
+      assertTrue(leader !=null && leader._parms._keep_cross_validation_fold_assignment);
+      assertNotNull(leader._output._cross_validation_fold_assignment_frame_id);
```
michalkurka: This tests checks the frame key is defined but doesn't make sure it actually has any data.

h2o-algos/src/test/java/hex/CrossValidFoldAssignmentsTest.java
```diff
@@ -0,0 +1,101 @@
...
+      parms._ntrees = 1;
+      parms._max_depth = 1;
+      parms._fold_column = "foldId";
+      parms._distribution = DistributionFamily.multinomial;
+      parms._keep_cross_validation_predictions=false;
+      parms._keep_cross_validation_fold_assignment=true;
+      GBM job = new GBM(parms);
+      gbm = job.trainModel().get();
+
+      //Let's check that if we remove training frame we will also remove Vec for fold assignments.
```
michalkurka: Why do we want to remove the fold assignments when training frame is removed? I would not expect such behavior (not I think it is really possible in H2O).

h2o-algos/src/test/java/hex/CrossValidFoldAssignmentsTest.java
```diff
@@ -0,0 +1,101 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class CrossValidFoldAssignmentsTest extends TestUtil {
+  @BeforeClass() public static void setup() { stall_till_cloudsize(1); }
+
+  // Check that we should not make a copy of fold assignments Vec in ModelBuilder.
+  // Fold column as a part of train dataset is not being tracked by Scope.enter() and is still available
+  // after the model is trained.
+  @Test public void checkFoldAssignmentsAreKeptWithoutMakeCopy() {
```
michalkurka: Thumns up! Thank you

michalkurka:
@deil87 thank you

one formal comment: PUBDEV-5956 is not marked for a fix release, should this go to master instead? Usually, we put such changes only to master unless there is a good reason.

h2o-algos/src/test/java/hex/CrossValidFoldAssignmentsTest.java
```diff
@@ -0,0 +1,101 @@
...
+      parms._max_depth = 1;
+      parms._fold_column = "foldId";
+      parms._distribution = DistributionFamily.multinomial;
+      parms._keep_cross_validation_predictions=false;
+      parms._keep_cross_validation_fold_assignment=true;
+      GBM job = new GBM(parms);
+      gbm = job.trainModel().get();
+
+      assertNotNull(gbm._output._cross_validation_fold_assignment_frame_id);
+      cvFoldAssignmentFrame = DKV.getGet(gbm._output._cross_validation_fold_assignment_frame_id);
```
michalkurka: A small comment here: you should ideally have a second copy of the "foldId" Vec, this test would pass even if the modelbuilder completely overwrote the fold column.

h2o-automl/src/test/java/ai/h2o/automl/AutoMLTest.java
deil87: @michalkurka this is how we have been checking it so far. If id is available then we assume that our property `keep_cross_validation_fold_assignment` is working fine. See for example **pyunit_automl_args.py** test **test_keep_cross_validation_fold_assignment_enabled_with_nfolds_neq_0()**. 
Are you suggesting to change this approach of checking?

@seb I think that we should rename **test_keep_cross_validation_fold_assignment_enabled_with_nfolds_eq_0** -> **test_keep_cross_validation_fold_assignment_DISABLED_with_nfolds_eq_0**

h2o-algos/src/test/java/hex/CrossValidFoldAssignmentsTest.java
deil87: @michalkurka there are so many things that we can check. Name of the test says `checkFoldAssignmentsAreKeptWithoutMakeCopy` so I think that I've checked this case. This case you are talking about should have been tested in the scope of adding fold_assignment functionality.... check that models are using correct folds. Of course it is easy to extend responsibility of my test....but where is the limit then? I prefer to keep the responsibility/scope as narrow as possible.

h2o-automl/src/test/java/ai/h2o/automl/AutoMLTest.java
michalkurka: I would suggest adding at least

   assertNotNull(DKV.getGet(leader._output._cross_validation_fold_assignment_frame_id))

or something like that. The existence of the key doesn't imply the existence of the data.

michalkurka:
@deil87 thanks for adding the tests. I added one more test to CrossValidFoldAssignmentsTest - this one checks the case when the folds are assigned by the algo itself (not given explicitly by the user).

michalkurka:
> @michalkurka I thought that fixes should go to rel-xia. Also we were discussing that I will prepare PR that is fixing only leakage without changing any semantics in ModelBuilder and create JIRA & another PR where I'm proposing changes that will probably go into 3.24 release ( hence master branch) since it is going to change semantic of the remove_impl method. So I would say that we need to set fix release for this jira.

Release branch should only have fixes for issues that are marked for a fix release. PUBDEV-5956 doesn't have a fix release assigned. My proposal is to set it to 3.24.0.1 - next major release, unless you think otherwise.

h2o-algos/src/test/java/hex/CrossValidFoldAssignmentsTest.java
michalkurka: My point was that using `isBitIdentical` could be misleading here because you are running it on the very same frame/vec. I will leave this one up to you. Not a big deal.

deil87:
> > @michalkurka I thought that fixes should go to rel-xia. Also we were discussing that I will prepare PR that is fixing only leakage without changing any semantics in ModelBuilder and create JIRA & another PR where I'm proposing changes that will probably go into 3.24 release ( hence master branch) since it is going to change semantic of the remove_impl method. So I would say that we need to set fix release for this jira.
> 
> Release branch should only have fixes for issues that are marked for a fix release. PUBDEV-5956 doesn't have a fix release assigned. My proposal is to set it to 3.24.0.1 - next major release, unless you think otherwise.

@michalkurka I don't think otherwise. I'm just wondering how it is working. Why we can't just put fix release version now. It was me who created this jira... maybe I should have added it in the beginning. Or probably this jira was forgotten somehow and we did not mark it to go into some fix release( how to check that we have reviewed it actually and did not put any fix version to it intentionally?).  Because for me it is a fix for something that was released in some major release. So if you insist that it should not go into fix release then I need more criteria to be able to differentiate between at least two types of fixes.

michalkurka:
@deil87 usually we set the fix release in the jira before we start working on it (there is always an exception, of course!). The process of setting the fix release is a group decision, we discuss internally and we reach conclusion.

Most of the fixes are for issues that are affecting either paying customers or community users. This is what usually goes into fix release. Everything else is discussed individually. The goal is to deliver stable fix releases and make upgrades safe for our users.

h2o-algos/src/test/java/hex/CrossValidFoldAssignmentsTest.java
deil87: @michalkurka Thank you for paying attention to this. I knew this. I even have the test where I'm removing train frame and fold assignment vec is being removed as well. I'm just checking here that we are getting actually the SAME vec as initial vec and assuming we don't mess with vec along the way.

deil87:
> @deil87 usually we set the fix release in the jira before we start working on it (there is always an exception, of course!). The process of setting the fix release is a group decision, we discuss internally and we reach conclusion.
> 
> Most of the fixes are for issues that are affecting either paying customers or community users. This is what usually goes into fix release. Everything else is discussed individually. The goal is to deliver stable fix releases and make upgrades safe for our users.

1) There is no way easy way to know if the jira was considered/discussed by the group. Probably if it is not pushed by customers or users than ideally it should be forgotten intentionally.
2) There is no clear relation between `The goal is to deliver stable fix releases` and not including jiras that were not pushed by customer/community.  With this PR i think we can say that we are improving stability. But I understand that any changes are dangerous and I see why we are probably trying to minimize the scope of the fix releases. 
Thank you for explanation, it is more clear for me right now.

PR #3065, opened 2018-11-20
Fix two intermittent tests

wendycwong:
The following tests have been failing:

pyunit_PUBDEV_4090_grid_model_seeds.py
pyunit_pubdev_4628_pubdev_4950_deeplearning_mojo_pojo_large.py.

Also, I fix the h2o-r/tests/testdir_parser/runit_PUBDEV_5705_drop_columns_parser_gz.R which set skipp_columns to all columns and hence encounter the error. I make sure the skipped_columns will never contain all the column indices.

Incorporate Seb's suggestion to fix them.

### DIFF
--- a/h2o-py/tests/pyunit_utils/utilsPY.py
+++ b/h2o-py/tests/pyunit_utils/utilsPY.py
@@ -3173,7 +3173,7 @@ def model_run_time_sorted_by_time(model_list):
     return model_runtime_sec_list
 
 
-def model_seed_sorted_by_time(model_list):
+def model_seed_sorted(model_list):
     """
     This function is written to find the seed used by each model in the order of when the model was built.  The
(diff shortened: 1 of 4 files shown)


seb-h2o:
does it also fix pyunit_pubdev_4628_pubdev_4950_deeplearning_mojo_pojo_large.py?

wendycwong:
Seb:  I just finished fixing the DL mojo/pojo test to make it test all the different DL modes.  I shorted the training dataset to make it faster.

Also, I fix the  h2o-r/tests/testdir_parser/runit_PUBDEV_5705_drop_columns_parser_gz.R which set skipp_columns to all columns and hence encounter the error.  I make sure the skipped_columns will never contain all the column indices.

PR #3069, opened 2018-11-21
[HEXDEV-724] add headers for cross-origin AJAX requests

honzasterba:
- needed for Flow UI development (UI server on different port than back-end)
- also removed some dead code

### DIFF
--- a/h2o-core/src/main/java/water/server/ServletUtils.java
+++ b/h2o-core/src/main/java/water/server/ServletUtils.java
@@ -166,29 +166,6 @@ public class ServletUtils {
     }
   }
 
-  public static boolean isXhrRequest(final HttpServletRequest request) {
-    final String requestedWithHeader = request.getHeader("X-Requested-With");
-    return "XMLHttpRequest".equals(requestedWithHeader);
-  }
-
-  public static void setCommonResponseHttpHeaders(HttpServletResponse response, final boolean xhrRequest) {
-    if (xhrRequest) {
-      response.setHeader("Cache-Control", "no-cache");
-    }
-    response.setHeader("X-h2o-build-project-version", H2O.ABV.projectVersion());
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/server/ServletUtils.java
```diff
@@ -166,29 +166,6 @@ public static String getDecodedUri(HttpServletRequest request) {
     }
   }
 
-  public static boolean isXhrRequest(final HttpServletRequest request) {
```
michalkurka: Thanks for noticing, there are copies of these methods in JettyHTTPD.

michalkurka:
> needed for Flow UI development (UI server on a different port than back-end)

Is there another way to do the same? We should prevent the JS libraries we bundle from making requests elsewhere than H2O backend.

Enabling CORS like this opens the way for the libraries to send data outside of the users' network. I think we should prevent that.

In development, you can make H2O serve the web resources from any location (see JarHash#getResource2). Would that help?

honzasterba:
@michalkurka regarding your security concerns, the bad thing about CORS headers is that it applies only to browser based requests. So yes, in theory you are right, should someone open a webpage it can start calling localhost:554321 and possibly leak sensitive data. But so can any other computer on the network (not via browser) since we by default listen on *:54321

I have thought about hiding this behaviour behind a property that you need to explicitly enable but decided not to because our API can be called from anywhere anyway and adding this ability to browsers seemed like not such a big risk.

I understand your concern and will disable this by default.

honzasterba:
@michalkurka  this should be good to merge

michalkurka:
@honzasterba couple of small comments

- to me, it looks like it enables CORS, not disables - perhaps you can change the name of the property
- can you add a comment that explains the purpose of the property?
- maybe instead of true/false we can make it more general and let it control the value of the header `Access-Control-Allow-Origin` - so you could enable it eg. for localhost and a specific port - this would make it more useful than just for development - but by no means necessary - true/false is clear and easy to use

PR #3070, opened 2018-11-21
PUBDEV-6081: Reducing the Invocation JMH level setup/teardown to only the training…

DiegoEliasCosta:
… model (PUBDEV-6081)

This patch is related to [PUBDEV-6081](https://0xdata.atlassian.net/browse/PUBDEV-6081)

In summary, we should avoid using Invocation level JMH setup/teardown on fast benchmarks. As explained in the JIRA ticket, it often offsets the measurements of the benchmarks quite a bit.
This patch reduces the invocation level to only the model training, as it is the only part required to be run on every invocation.

I ketp the original exception handling as it is to avoid causing any undesired behavior, but it can be further simplified.


michalkurka:
Thanks for contributing! I will look into that

michalkurka:
Great changes, thank you. We will take a lesson from that and follow this pattern from now on.

Thank you for contributing!

PR #3071, opened 2018-11-21
PUBDEV 6082 proposal to remove cv related objects with main model

deil87:
https://0xdata.atlassian.net/browse/PUBDEV-6082


deil87:
Still need to add tests that cover these changes

PR #3080, opened 2018-11-23
PUBDEV-6076 keep timestamp format after group_by

maurever:
**Problem:** 
The type of group_by column is changed to `int` even though there is `time` type before group by operation. The problem is the new calculated group by vectors are saved in the DKV with default type `int`. The original type is decoded from values in the vector to write it to a schema. But `timestamp` type is `int` by default and can't be easily decoded back to a `timestamp`. 

**Solution:** 
Don't use the default `int` type when the new vectors are created. Group by columns are not changed during calculation so that its type can be preserved.

### DIFF
--- a/h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
@@ -465,7 +465,9 @@ public class AstGroup extends AstPrimitive {
     }
     Vec v = Vec.makeZero(ngrps); // dummy layout vec
     // Convert the output arrays into a Frame, also doing the post-pass work
-    return mrfill.doAll(types, new Frame(v)).outputFrame(names, domains);
+    Frame f =  mrfill.doAll(types, new Frame(v)).outputFrame(names, domains);
+    v.remove();
+    return f;
   }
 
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -453,18 +453,19 @@ public static Frame buildOutput(int[] gbCols, int noutCols, Frame fr, String[] f
...
-    for (int i = 0; i < fcnames.length; i++)
+    for (int i = 0; i < fcnames.length; i++) {
       names[i + gbCols.length] = fcnames[i];
+      types[i + gbCols.length] = Vec.T_NUM;
+    }
     Vec v = Vec.makeZero(ngrps); // dummy layout vec
-
     // Convert the output arrays into a Frame, also doing the post-pass work
-    Frame f = mrfill.doAll(nCols, Vec.T_NUM, new Frame(v)).outputFrame(names, domains);
-    v.remove();
```
michalkurka: Do we really not need to remove the zero Vec? Are no keys leaking?

maurever: You are right! I overlooked it and were suprised what is leaked. Thanks.

PR #3082, opened 2018-11-23
PUBDEV-5790 - H2O imports headers into first row of the dataframe

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-5790

Introduced new ParseSetup field `first_row_len`.

However, this `first_row_len` parameters mostly doesn't need to originate from the user. In case of CSV parser, where this is mostly needed (non-binary), we already know length of headers in advance from the guessSetup phase. These heuristics of CSVParser's guessSetup mostly do work and are the most reliable source. 

R and Python API had to be adjusted to pass the `first_line_len` variable from ParseSetup/GuessSetup phase to Parse phase.

Right now, the APIs only pass the variable between those two phases. This setting is not exposed to users yet. Whether to expose it as an expert-level workaround for those very edge cases we don't know about it is to be discussed. I wouldn't do it just now.

The MultiFileParseTask is able to leverage this `first_line_len` property to determine if reading first chunk is enough to parse the headers from a file or if more bytes need to be fetched.


h2o-core/src/main/java/water/parser/ParseDataset.java
```diff
@@ -774,7 +774,16 @@ private FVecParseWriter makeDout(ParseSetup localSetup, int chunkOff, int nchunk
       byte[] zips = vec.getFirstBytes();
       ZipUtil.Compression cpr = ZipUtil.guessCompressionMethod(zips);
```
Pscheidl: Do we really need to fetch these first bytes just to determine compression ?

h2o-core/src/main/java/water/parser/ParseDataset.java
```diff
@@ -774,7 +774,16 @@ private FVecParseWriter makeDout(ParseSetup localSetup, int chunkOff, int nchunk
...
       if (localSetup._check_header == ParseSetup.HAS_HEADER) { //check for header on local file
-        byte[] bits = decryptionTool.decryptFirstBytes(ZipUtil.unzipBytes(zips, cpr, localSetup._chunk_size));
+        byte[] bits = zips;
+        if (localSetup._firstLineLen > localSetup._chunk_size) {
+          // If length of first line is known (from guess phase) and chunk size is smaller,
+          // there is a possibility of not being able to load enough data to check the whole header.
+          // Therefore, more bytes are loaded here.
+          bits = extractHeaderBytes(localSetup, decryptionTool, vec);
+        } else {
+          bits = decryptionTool.decryptFirstBytes(ZipUtil.unzipBytes(bits, cpr, localSetup._chunk_size));
```
Pscheidl: The compression does not seem to be required to be known at this point. We could avoid fetching first bytes from the vec twice, but this is a low-cost operation, thus kept in place right now.

h2o-core/src/main/java/water/parser/ParseDataset.java
seb-h2o: I think so: magic number is common way to identify file type

seb-h2o: although usually only first FEW bytes are needed :)

seb-h2o:
@Pscheidl , you write that `These heuristics of CSVParser's guessSetup mostly do work and are the most reliable source.`
Although, according to your tests it currently doesn't work for very long CSV headers, right?
Do we really need a new user param for this or shouldn't we improve the header detection for CSV in this case.
I just had the same issue with ARFF and was able to fix it by looping through chunks until we reach the end of the header. I believe it should be even easier with CSV because even if the header is spread over multiple chunks, it ends with the first EOL.
To be clear, my question is: what is the reason why CSV parser fails detecting large headers today?

h2o-core/src/main/java/water/fvec/FileVec.java
```diff
@@ -39,6 +39,12 @@ public void setNChunks(int n){
     _nChunks = n;
     setChunkSize((int)length()/n);
   }
+
+  @Override
+  public int chunkLen(int cidx) {
```
michalkurka: Please explain, this function is supposed to return the length of the chunk. In case of FileVec that is the byte-size of the chunk. The size of the chunk is predetermined before the parse. This override doesn't seem to do the right unless I am missing something.

michalkurka:
@Pscheidl can this be generalized? The issue here seems to be that H2O determines the size of the chunk but this size doesn't necessarily be good for the first chunk. We could make the first chunk bigger in certain cases because we need to compensate for too long headers.

So instead of your first_line_size you could introduce a first_chunk_size (we already do have chunk_size). This would also be backward compatible (you could simply use chunk_size if fist_chunk_size is not provided).

Another point of view: If the headers are so big that the default chunk size is too small, does it mean that the dataset is too small? Should we just boost the chunk size instead? If we have too many tiny chunks the performance will actually be worse than if we have fewer amount reasonably sized chunks.

h2o-core/src/main/java/water/parser/ParseDataset.java
Pscheidl: If you see the IO methods I used in my new pieces of code in this class, maybe then you'll ask the same question or see an obvious error in my thinking. :)

seb-h2o: if you mean that it could now be moved just before L784 instead, then apparently yes.
but it's a slightly different question from the one you asked, sorry for not reading your mind ;)

Pscheidl: The answer is no we don't need to fetch these first bytes to determine first compression, as those bytes are fetched anyway in the coder later. But the IO of the old part would need to work like my new part then. This is what I meant.

h2o-core/src/main/java/water/fvec/FileVec.java
Pscheidl: Damn, this is a leftover of my experiments. Sorry !

PR #3085, opened 2018-11-26
Shortened runit_PUBDEV_5705_drop_columns_parser_arff_large.R test.

wendycwong:
Inside runit_PUBDEV_5705_drop_columns_parser_arff_large.R test, there are two tests.  For both tests, I will only either test h2o.uploadFile or h2o.importFile but not both.

### DIFF
--- a/h2o-r/tests/testdir_parser/runit_PUBDEV_5705_drop_columns_parser_arff_large.R
+++ b/h2o-r/tests/testdir_parser/runit_PUBDEV_5705_drop_columns_parser_arff_large.R
@@ -60,14 +60,6 @@ test.parseSkippedColumnsARFF<- function() {
   skipall <- onePermute
   skip99Per <- onePermute[1:floor(h2o.ncol(f1) * 0.99)]
 
-  # test skipall for h2o.importFile
-  e <-
-    tryCatch(
-      assertCorrectSkipColumns(fileName, fullFrameR, skipall, TRUE, h2o.getTypes(f1)),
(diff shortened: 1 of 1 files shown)


h2o-r/tests/testdir_parser/runit_PUBDEV_5705_drop_columns_parser_arff_large.R
```diff
@@ -60,14 +60,6 @@ test.parseSkippedColumnsARFF<- function() {
...
 
-  # test skipall for h2o.importFile
-  e <-
-    tryCatch(
-      assertCorrectSkipColumns(fileName, fullFrameR, skipall, TRUE, h2o.getTypes(f1)),
-      error = function(x)
-        x
-    )
-  print(e)
   # test skipall for h2o.uploadFile
```
seb-h2o: this looked fine to me, could be kept or not, was mainly wondering why the first 16 lines of this file all look the same.
and why this function test.parseSkippedColumnsARFF looks copy/pasted multiple times later in this file

wendycwong: I planned to do the following:

1. skipped all columns (this should generate an error)
2. skipped 99% of the columns.

Then, I want to test both h2o.uploadFile and h2o.importFile.

Based on this calculation, I will need to have 4 parseSkippedColumnsARFF since the actual tests are performed in this function.

PR #3086, opened 2018-11-26
fix seeds for data generation and model building.

wendycwong:
Fix the test pyunit_pubdev_5058_glrm_mojo_pubdev_5470_load_model_large.py by
1. fixed the seed in dataframe generation;
2. fixed the seed in model generation.


wendycwong:
This is fixed in another PR.  I am going to abandon it here.

PR #3087, opened 2018-11-27
PUBDEV-3847 add new test defined in issue

maurever:
I added a new test to check if a `gbm` model still fails with `java.lang.AssertionError: Coldata xxx out of range`.

### DIFF
--- a/h2o-r/tests/testdir_jira/runit_pubdev_3847.R
+++ b/h2o-r/tests/testdir_jira/runit_pubdev_3847.R
@@ -1,0 +1,28 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../scripts/h2o-r-test-setup.R")
+######################################################################################
+# Turn the PUBDEV-3847 issue into the test and check if it fails
+######################################################################################
+pubdev.3847.test <-
+  function() {
+      file <- locate("smalldata/gbm_test/pubdev-3847.zip")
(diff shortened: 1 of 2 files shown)


h2o-r/tests/testdir_jira/runit_pubdev_3847.R
```diff
@@ -0,0 +1,30 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../scripts/h2o-r-test-setup.R")
+######################################################################################
+# Turn the PUBDEV-3847 issue into the test and check if it fails
+######################################################################################
+pubdev.3847.test <-
+  function() {
+      file <- locate("smalldata/jira/pubdev_3847.csv")
+      data <- h2o.importFile(file, destination_frame = "pubdev3847.data", parse = FALSE)
```
michalkurka: This is causing the test failure. Please remove `parse = FALSE` - it should resolve it.

maurever:
Should the test be NOPASS or not? @michalkurka

michalkurka:
@maurever I would be optimistic and assume the issue doesn't exist anymore (and also close the related JIRA). If it happens again we can re-open the jira.

michalkurka:
@maurever the test is not using a seed and is thus not reproducible

One approach you can take is to generate a random seed and log it for future reference. We need to be able to reproduce the test if it fails. To do that we need to know the seed and the environment where it was running (number of CPU cores, number of nodes). Please make sure this information is available in the test output. cc: @michal-raska

PR #2938, opened 2018-10-18
PUBDEV-5820: Jetty refactor

pkozelka:
NOTE that due to postponing support of CDH6 / Jetty9, this effort continues in smaller steps that will be easier to review and debug.
The first step of this divided approach is PR #2976 - please consider reviewing that one instead.

-----

Brings support for CDH 6 and Jetty 9. Support for Jetty 8 will remain selectively possible.

Following steps are included here:

- [x] add module for CDH6.0 build (derived from branch `cdh6`, PR #2830)
- [x] upgrade Jetty to version 9 (re-introduced former attempt from @jakubhava + improvements)
- [ ] minimize code requiring direct dependency on Jetty
- [x] separate Jetty stuff into its own module, to enable support of both versions
- [x] tune default versions per environment
- [ ] fix problem with configuring header size etc in jetty 9 (the way available in jetty 8 is no longer usable)
- [ ] make sure that `h2odriver.jar` is executable and stuffed with all runtime deps, and matching hadoop deps

PR #3092, opened 2018-11-28
PUBDEV-5820 - Added Jetty 9 implementation of IFacade

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-5820


Pscheidl:
Two little todos, otherwise done.

michalkurka:
Pavel, this is a good start. Thank you. Can we do something about the code duplication?

Pscheidl:
@michalkurka I am not sure a common codebase between jetty 8 and 9 is really what we want. The portion of common code might be rather reduced in future by the needs to modify Jetty 8 & 9 code differently. Some  common logic could be extract for sure.

Pscheidl:
@pkozelka Good point with the dependencies. Yes, but that is solveable, but rather dirty and I'd also prefer not to do that. Our goal should be to get rid of this misuse of servlet standard and restore a proper component-oriented frontend. This will introduce much less work with maintenance and is also upgradeable.

The common code might not become common in future, as there will definitely be changes in Jetty 9. There are API-breaking changes in minor releases in Jetty, could be observed in 8.x already.

h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9ProxyServerAdapter.java
```diff
@@ -0,0 +1,86 @@
...
+      throw e;
+    } catch (Exception e) {
+      throw new IOException(e);
+    }
+  }
+
+  private void registerHandlers(HandlerWrapper handlerWrapper, ServletContextHandler context, Credentials credentials, String proxyTo) {
+    // setup authenticating proxy servlet (each request is forwarded with BASIC AUTH)
+    final ServletHolder proxyServlet = new ServletHolder(TransparentProxyServlet.class);
+    proxyServlet.setInitParameter("ProxyTo", proxyTo);
```
pkozelka: I believe that `ProxyTo` works only on Jetty 8, but Jetty 9 requires `proxyTo` parameter - that's why I originally did it https://github.com/h2oai/h2o-3/pull/2938/commits/a4d26b5cae69b7ba9e6b15873ea38a65521dcee2

h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
```diff
@@ -73,6 +72,7 @@ Server createJettyServer(String ip, int port) {
         final ArrayList<ConnectionFactory> connectionFactories = new ArrayList<>();
         ConnectionFactory connectionFactory = new HttpConnectionFactory();
         final HttpConfiguration httpConfiguration = new HttpConfiguration();
+        httpConfiguration.setSendServerVersion(true);
```
pkozelka: I think this is good idea - and we can add the same to jetty8 version in this PR, WDYT?

Pscheidl: We already have this in Jetty 8. This is only done in Jetty 9 in a completely different way, like this ^^^.

h2o-assemblies/main/build.gradle
```diff
@@ -16,7 +16,7 @@ configurations {
 // Dependencies
 dependencies {
     compile project(":h2o-app")
-    runtime project(":h2o-jetty-8")
```
michalkurka: @Pscheidl I would keep the default as Jetty 8 in this PR and only enable it in CDH 6.

Pscheidl:
@michalkurka We will keep the test. As the only module with Jetty 9 functionality will be CDH 6.0, this test may be preserved now.

PR #3093, opened 2018-11-28
Fixed DL mojo test by surrounding DL autoencoder mode with try/except.

wendycwong:
Incorporate Michal K code review comments to check error message is related to unstable model only.  For other error messages, an exception will be raised.


wendycwong:
This is the same as PR 3089 which I am not going to merge into master and was approved by Michalk and seb

PR #3095, opened 2018-11-28
PUBDEV-5876: finished speeding up binomial COD

wendycwong:
This PR completes the work for JIRA: https://0xdata.atlassian.net/browse/PUBDEV-5876?filter=-1.

Worked done:

speedup GLM COD for binomial;
added Python unit tests to verify correct results are obtained after my change.
speedup GLM COD for multinomial;
added R unit tests to verify correct results.

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/ComputationState.java
+++ b/h2o-algos/src/main/java/hex/glm/ComputationState.java
@@ -557,8 +557,6 @@ public final class ComputationState {
     public double yy;
     public final double likelihood;
 
-
-
     public GramXY(Gram gram, double[] xy, double [] grads, double[] beta, int[] activeCols, int [] newActiveCols, double yy, double likelihood) {
       this.gram = gram;
       this.xy = xy;
@@ -568,6 +566,21 @@ public final class ComputationState {
       this.newCols = newActiveCols;
       this.yy = yy;
       this.likelihood = likelihood;
+    }
+
+    public final double [] getCODGradients(double l2pen){
+      int lastGradInd = xy.length-1;
+      if(grads == null){
+        grads = new double[xy.length];
+        for(int i = 0; i < lastGradInd; ++i)
+          grads[i] = xy[i] - l2pen * beta[i];
+        grads[lastGradInd] = xy[lastGradInd]; // intercept does not have l1/l2 penality
+      }
+      if(newCols != null) {
+        for (int i : newCols)
+          grads[i] = xy[i] - (i==lastGradInd?0:l2pen * beta[i]);
+      }
+      return grads;
     }
 
     public final double [] getCODGradients(){
@@ -630,7 +643,10 @@ public final class ComputationState {
   protected GramXY computeNewGram(DataInfo activeData, double [] beta, GLMParameters.Solver s){
     double obj_reg = _parms._obj_reg;
     if(_glmw == null) _glmw = new GLMModel.GLMWeightsFun(_parms);
-    GLMTask.GLMIterationTask gt = new GLMTask.GLMIterationTask(_job._key, activeData, _glmw, beta,_activeClass).doAll(activeData._adaptedFrame);
+    boolean useCODBM = s.equals(GLMParameters.Solver.COORDINATE_DESCENT) &&
+            (_glmw._family.equals(Family.binomial) || _glmw._family.equals(Family.multinomial));
+    GLMTask.GLMIterationTask gt = new GLMTask.GLMIterationTask(_job._key, activeData, _glmw, beta,_activeClass,
+            useCODBM).doAll(activeData._adaptedFrame);
     gt._gram.mul(obj_reg);
     ArrayUtils.mult(gt._xy,obj_reg);
     int [] activeCols = activeData.activeCols();
@@ -685,7 +701,10 @@ public final class ComputationState {
         }
       }
       if(!weighted || matches) {
-        GLMTask.GLMIncrementalGramTask gt = new GLMTask.GLMIncrementalGramTask(newColsIds, activeData, _glmw, beta).doAll(activeData._adaptedFrame); // dense
+        boolean useCODBM = s.equals(GLMParameters.Solver.COORDINATE_DESCENT) &&
+                (_glmw._family.equals(Family.binomial) || _glmw._family.equals(Family.multinomial));
+        GLMTask.GLMIncrementalGramTask gt = new GLMTask.GLMIncrementalGramTask(newColsIds, activeData, _glmw, beta,
+                useCODBM, _activeClass).doAll(activeData._adaptedFrame); // dense
         for (double[] d : gt._gram)
           ArrayUtils.mult(d, obj_reg);
         ArrayUtils.mult(gt._xy, obj_reg);
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -642,79 +642,50 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       return xy;
     }
 
-    private void fitCOD_multinomial(Solver s) {
-      double[] beta = _state.betaMultinomial();
+    private void fitIRLSM_multinomial(Solver s){
+      assert _dinfo._responses == 3:"IRLSM for multinomial needs extra information encoded in additional reponses, expected 3 response vecs, got " + _dinfo._responses;
+      double [] beta = _state.betaMultinomial();
       LineSearchSolver ls;
-
       do {
         beta = beta.clone();
-           for (int c = 0; c < _nclass; ++c) {
-            _state.setActiveClass(c);
-            boolean onlyIcpt = _state.activeDataMultinomial(c).fullN() == 0;
-
+        for (int c = 0; c < _nclass; ++c) {
+          boolean onlyIcpt = _state.activeDataMultinomial(c).fullN() == 0;
+          _state.setActiveClass(c);
+          if (s.equals(Solver.COORDINATE_DESCENT)) {
             if (_state.l1pen()==0) {
               if (_state.ginfoNull())
                 _state.updateState(beta, _state.gslvr().getGradient(beta));
               ls = new MoreThuente(_state.gslvrMultinomial(c), _state.betaMultinomial(c, beta), _state.ginfoMultinomial(c));
             } else
               ls = new SimpleBacktrackingLS(_state.gslvrMultinomial(c), _state.betaMultinomial(c, beta), _state.l1pen());
-/*            LineSearchSolver ls = (_state.l1pen() == 0)
-                    ? new MoreThuente(_state.gslvrMultinomial(c), _state.betaMultinomial(c, beta), _state.ginfoMultinomial(c))
-                    : new SimpleBacktrackingLS(_state.gslvrMultinomial(c), _state.betaMultinomial(c, beta), _state.l1pen());*/
-
-            new GLMMultinomialUpdate(_state.activeDataMultinomial(), _job._key, beta, c).doAll(_state.activeDataMultinomial()._adaptedFrame);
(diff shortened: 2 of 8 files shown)


h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -570,6 +568,21 @@ public GramXY(Gram gram, double[] xy, double [] grads, double[] beta, int[] acti
...
+        for(int i = 0; i < lastGradInd; ++i)
+          grads[i] = xy[i] - l2pen * beta[i];
+        grads[lastGradInd] = xy[lastGradInd]; // intercept does not have l1/l2 penality
+      }
+      if(newCols != null) {
+        for (int i : newCols)
+          grads[i] = xy[i] - (i==lastGradInd?0:l2pen * beta[i]);
+      }
+      return grads;
+    }
```
wendycwong: This is different from getCODGradients(). Here, I am trying to calculate the gradient as the partial derivatives of the negative log likelihood + l2pen and I need this to update coefficients for COD. This is equaition 4.24 of ESLII-print4 without the l2/l1 penalty

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -642,82 +642,53 @@ private void doCleanup() {
...
+          } else {
+            ls = (_state.l1pen() == 0)
                     ? new MoreThuente(_state.gslvrMultinomial(c), _state.betaMultinomial(c, beta), _state.ginfoMultinomial(c))
-                    : new SimpleBacktrackingLS(_state.gslvrMultinomial(c), _state.betaMultinomial(c, beta), _state.l1pen());*/
-
-            new GLMMultinomialUpdate(_state.activeDataMultinomial(), _job._key, beta, c).doAll(_state.activeDataMultinomial()._adaptedFrame);
-            ComputationState.GramXY gram = _state.computeGram(_state.betaMultinomial(c, beta), s);
-            double[] betaCnd = COD_solve(gram, _state._alpha, _state.lambda());
+                    : new SimpleBacktrackingLS(_state.gslvrMultinomial(c), _state.betaMultinomial(c, beta), _state.l1pen());
+          }
```
wendycwong: I am combining back the IRLSM and COORDINATE_DESCENT into one part again.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1466,6 +1437,20 @@ else if(_parms._lambda_search) { // lambda search prefers coordinate descent
...
     int numStart = activeData.numStart();
+      double temp;
+      if (_parms._family.equals(Family.binomial) || _parms._family.equals(Family.multinomial)) {
+        for (int i = 0; i < P; ++i) {
+          temp = xx[i][i] + l2pen;
+          diagInv[i] = 1.0 / temp;
+          grads[i] += temp * beta[i];
+        }
+        grads[P] += (xx[P][P]) * beta[P];
+      } else { // no change for regression
```
wendycwong: I am converting the gradient calculated in equation 4.24 of ESLII-print4 into last line of equation 4.26. Now, the gradient is transpose(X)Wz. This is what Tomas used in his calculation in later parts.  By this transformation, I don't need to do any extra work.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -720,14 +722,26 @@ public final double likelihood(double yr, double ym) {
...
+
+    public GLMWeights computeWeightsCOD(double y, double eta, double off, double w, GLMWeights x) {
+      double etaOff = eta + off;
+      x.mu = linkInv(etaOff); // prob estimated response = 1
+      double temp = x.mu*(1-x.mu);
+      x.respMinusPr = w*(y-x.mu);
+      x.prOneMinpr = w*(temp < 1e-6?1e-6:temp);
+      x.w = x.prOneMinpr;
+      x.z = x.respMinusPr;
+      likelihoodAndDeviance(y,x,w);
```
wendycwong: I am adding this function instead of using the old one because it is simplier and more numerically stable. For example, x.w is equivalent to x.mu*(1-x.mu). However, it is calculated as w/(var * d * d) which can be simplified to x.mu * (1-x.mu) for Binomials only.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -794,46 +765,45 @@ private void fitLSM(Solver s){
...
           firstIter = false;
           long t3 = System.currentTimeMillis();
-          if(_state._lsNeeded) {
-            if(ls == null)
+          if (_state._lsNeeded) {
+            if (ls == null)
               ls = (_state.l1pen() == 0 && !_state.activeBC().hasBounds())
-                 ? new MoreThuente(_state.gslvr(),_state.beta(), _state.ginfo())
-                 : new SimpleBacktrackingLS(_state.gslvr(),_state.beta().clone(), _state.l1pen(), _state.ginfo());
+                      ? new MoreThuente(_state.gslvr(), _state.beta(), _state.ginfo())
```
Pscheidl: It is very hard to read the actual changes because of the formatting.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -721,13 +721,23 @@ public final double likelihood(double yr, double ym) {
...
       likelihoodAndDeviance(y,x,w);
       return x;
     }
+
+    public GLMWeights computeWeightsCOD(double y, double eta, double off, double w, GLMWeights x) {
+      double etaOff = eta + off;
+      x.mu = linkInv(etaOff); // prob estimated response = 1
+      double temp = x.mu*(1-x.mu);
+      x.z = w*(y-x.mu);
+      x.w = w*(temp < 1e-6?1e-6:temp);
```
Pscheidl: This part could go wrong (concentrating only at numerics). Let's use Math.max, which handles all the edge cases including NaNs correctly.

Pscheidl:
While going through the PR, I found many issues in the code that are NOT RELATED to this PR. Yet I'd like to mention them for us to consider improving it.

- Numerics - unnecessary propagation of error
- Dead code - unused methods and other pieces of code. The 1st derivative of inverse linking functions method is unused for example.
- public methods that could be private (does not cause us that much trouble)

Lots of commented code. E.g. why do we have this comment in `linkDeriv` function:

```plain
//          double res = _tweedie_link_power == 0
//            ?Math.max(2e-16,Math.exp(x))
//            // (1/lambda) * eta^(1/lambda - 1)
//            :(1.0/_tweedie_link_power) * Math.pow(link(x), 1.0/_tweedie_link_power - 1.0);
```

At first side, I have no idea to tell if this equation is just something unused for tweedie or something to be used at any time near future.

I don't understand why `1e-6` has been chosen as a point for "numerical instability", to quote the author precisely. We should use Math.max() at the places in the code where this is checked.


Also, I *believe I found a bug*. Again, not related to this PR. There are many checks like this one in `GLMModel`:

```java
          return (divsor<1e-6)?1e6:(1.0/divsor);
```

What bothers me is not the Math.max() unused mostly, but the value returned when the condition `<1e-6` is evaluated as true. I believe the purpose of this is to avoid operation with numbers with less decimal places than this "somehow" chosen `point of numerical instability`. Numbers lower than `- 1/1000000` suddenly become `1 000 000` (one million!), because the corrected value is `1e6`. I believe this error had beed done once and than copy-pasted.

Am I correct ?

Also, in GLMOutput class, there are unused methods as well, e.g. `lambda_selected`.

If I am right, maybe we should do a JIRA for that and clean up a little bit. Just a tiny little bit of absolutely necessary cleaning.

h2o-algos/src/main/java/hex/glm/GLMModel.java
Pscheidl: Also, is 1e6 really the desired value returned ?

h2o-algos/src/main/java/hex/glm/GLM.java
Pscheidl: Tiny observation: you can calculate `xx[i][i] + l2pen;` once before the `ifelse` block starts. :)

h2o-algos/src/main/java/hex/glm/GLMTask.java
```diff
@@ -1533,36 +1544,60 @@ protected void processRow(Row r) { // called for every row in the chunk
...
+        } else if (_beta != null) {
+          _glmf.computeWeights(y, r.innerProduct(_beta) + _sparseOffset, r.offset, r.weight, _w);
+          w = _w.w;
+          wz = w * _w.z;
+          _likelihood += _w.l;
+        } else {
+          w = r.weight;
+          wz = w * (y - r.offset);
+        }
+        wsum += w;
```
Pscheidl: `wsum += w + r.weight` ?

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: x.mu is a probability from 0 to 1 and x.mu*(1-x.mu) can be small.  Tomas put it in there to make sure the algorithm is numerically table.  I changed it per your suggestion and made more changes to the code to remove all such similar comparisons in GLMModel.java

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: I changed it back to what it was before.

wendycwong:
Pavel:

I noticed that there are a lot of commented out codes as well.  However, since I do not totally understand GLM, I was reluctantly to delete them at this point.  A clean up will be good.  

Regarding the 1e-6, it is basically saying that if Prob(yi=1) is less than 1e-6, I don't really care anymore and I am just going to cap all probability to be 1e-6.  Since 1e-6 was chosen and changing this may change the accuracy of the algo going forward, I feel that we are kind of stuck at that value.

Pscheidl:
@wendycwong Please look at the value. If the value is less than 1e-6, it would make sense to set it to `1e-6` to me, not `1e6`. Notice the missing `-` sign.

In other notation, if value is less than `0.0000006`, thank we should set it to `0.0000006`. The numerical goal there is not to let the value there to be closer to zero than `0.0000006`, as then the error might be huge. I know there are holes in standards following IEEE 754, so I believe Tomas had a good reason to do that. 

However, whan I believe is done is to set the value to the closest possible one, which is the `0.0000006` in this case, because that equals `1e-6`. What we do on many lines in GLM is to limit it to `1e6`, which equals to one million (1 000 000). **You know that**, as you probably intentionally used `1e-6` in your code, which seems correct to me. But there are lines which set the value to `1e6`, which I believe is a mistake. Huge mistake possibly.

wendycwong:
I have preformed a hadoop profiling and here is the result:


airline run time |   |   |   |  
-- | -- | -- | -- | --
before_wendy | rel_xia | Improvement | glm_COD | Improvement to rel-xia
129.392 | 110.886 | 14.30227526 | 113.121 | -2.015583572
127.563 | 110.767 | 13.16682737 | 111.147 | -0.343062464
131.382 | 112.841 | 14.11228327 | 111.227 | 1.430331174
126.739 | 110.079 | 13.14512502 | 111.511 | -1.300883911
127.944 | 109.488 | 14.42506096 | 114.404 | -4.489989771
124.729 | 110.944 | 11.05196065 | 109.523 | 1.280826363
127.788 | 108.15 | 15.36764015 | 174.509 | -61.35829866
128.471 | 109.616 | 14.67646395 | 110.712 | -0.999854036
127.768 | 114.609 | 10.29913593 | 115.433 | -0.718966224
128.961 | 108.85 | 15.59463714 | 110.325 | -1.355075792
132.385 |   |   | 112.925 |  
128.4656364 | 110.623 | 13.88903435 | 117.7124545 | -6.408662345

wendycwong:
The performance is just not there for this PR.  I am going to close it without merging it.  So sad!

PR #3097, opened 2018-11-29
PUBDEV-5865: Add monotonicity constraints to H2O GBM

michalkurka:


.gitignore
```diff
@@ -107,7 +107,6 @@ native
 
 # Ignore Jupyter Notebook project files
 *.ipynb_checkpoints/
-*.ipynb
```
Pscheidl: :+1:

h2o-algos/src/main/java/hex/schemas/GBMV3.java
```diff
@@ -84,6 +86,9 @@
     @API(help="Column sample rate (from 0.0 to 1.0)", level = API.Level.critical, gridable = true)
     public double col_sample_rate;
 
+    @API(help = "A mapping representing monotonic constraints. Use +1 to enforce an increasing constraint and -1 to specify a decreasing constraint.", level = API.Level.secondary)
+    public KeyValueV3[] monotone_constraints;
```
Pscheidl: I wonder how checkpointing will work.

Is it possible to continue "improving" a model with monotonic constraints ? Or even better, remove them selectively by setting them to zero ?

michalkurka: Good question. I think people can keep training a model with constraints. It would even be possible to remove the constraints (or define them in the opposite direction - not much sense though). The new trees will respect the new parameters.

h2o-algos/src/main/java/hex/tree/TreeUtils.java
```diff
@@ -0,0 +1,32 @@
...
+import hex.ModelBuilder;
+import water.fvec.Frame;
+import water.fvec.Vec;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class TreeUtils {
+
+  public static void checkMonotoneConstraints(ModelBuilder<?, ?, ?> mb, Frame train, KeyValue[] constraints) {
```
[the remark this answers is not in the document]
michalkurka: The code will take any number, only the sign matters. We can restrict it more but IMHO not necessary.

PR #3100, opened 2018-11-29
PUBDEV-6097 - Jetty 8 updated to version 8.1.21.v20160908

Pscheidl:
Updated Jetty 8 to latest fix version.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -38,9 +38,6 @@ localNexusLocation=http://nexus:8081/nexus/repository
 # Version of Apache Parquet dependency (should be kept in sync with the version used in current Spark releases)
 defaultParquetVersion=1.8.1
 
-# Jetty Version
-jettyVersion=9.2.24.v20180105
-
 # Default Hadoop client version
 defaultHadoopClientVersion=2.8.4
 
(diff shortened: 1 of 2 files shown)


michalkurka:
@jakubhava that is a good suggestion

Do we need to spread the jetty version across multiple modules? Ideally h2o-jetty-8 module should be the only that defines the version. The overall goal is to be able to switch between h2o-jetty-8 and h2o-jetty-9 at assembly level. Having jetty versions in different modules would make it difficult (if not impossible).

Pscheidl:
Removed the common Jetty version. Good idea @jakubhava . Thanks !

Pscheidl:
@jakubhava we're solving the dependency definition and also the shape of the dependency graph as a whole in https://github.com/h2oai/h2o-3/pull/3101 .

Are you ok with that ? This PR is only about updating the verison.

jakubhava:
@Pscheidl thanks for the note :) I'm all good with this PR

pkozelka:
@Pscheidl @michalkurka is there anything we are waiting for, with this PR? I suggest to proceed with merge.

jakubhava:
@Pscheidl should we merge?

PR #1344, opened 2017-07-07
PUBDEV-4661: Reconstruction error for AutoEncoder in EasyWrapper

michalkurka:
this is just a simple quick fix, per-feature error will be added later

PR #3102, opened 2018-11-30
PUBDEV-6030 get reconstruction error for auto encoders from easy wrapper

maurever:
Edit `EasyPredictModelWrapper` to be able to calculate reconstruction error for autoencoder. Edit `DeepLearningAutoEncoderTest` to test if the error from model and mojo model is the same. 

Cherry pick [commit ](https://github.com/h2oai/h2o-3/pull/3102/commits/915ce9399ecee905ec764eb7b76150e48406a422) of @michalkurka, but find a slightly different solution.

JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6030

### DIFF
--- a/h2o-algos/src/test/java/hex/deeplearning/DeepLearningAutoEncoderCategoricalTest.java
+++ b/h2o-algos/src/test/java/hex/deeplearning/DeepLearningAutoEncoderCategoricalTest.java
@@ -14,6 +14,7 @@ import water.TestUtil;
 import water.fvec.Frame;
 import water.fvec.NFSFileVec;
 import water.fvec.Vec;
+import water.parser.BufferedString;
 import water.parser.ParseDataset;
 import water.util.Log;
 
@@ -84,33 +85,6 @@ public class DeepLearningAutoEncoderCategoricalTest extends TestUtil {
 
     Assert.assertEquals(l2vec.mean(), mymodel.mse(), 1e-8*mymodel.mse());
 
-    try {
-      DeeplearningMojoModel mojoModel = (DeeplearningMojoModel) mymodel.toMojo();
-      EasyPredictModelWrapper model = new EasyPredictModelWrapper(mojoModel);
-      AutoEncoderModelPrediction tmpPrediction;
-      RowData tmpRow;
-      double calcNormMse = 0;
-      for (int r = 0; r < train.numRows(); r++) {
-        tmpRow = new RowData();
-        for (int c = 0; c < train.numCols(); c++) {
-          tmpRow.put(train.names()[c], train.vec(c).at(r));
-        }
-        tmpPrediction = model.predictAutoEncoder(tmpRow);
-        calcNormMse += tmpPrediction.mse;
-      }
-      double mojoMeanError = calcNormMse/train.numRows();
-      sb.append("Mojo mean reconstruction error (train): ").append(mojoMeanError).append("\n");
-      sb.append("Mean reconstruction error should be the same from model compare to mojo model " +
-              "reconstruction error: ");
-      sb.append(mymodel.mse()).append(" == ").append(mojoMeanError).append("\n");
-      Assert.assertEquals(mymodel.mse(), mojoMeanError, 1e-7);
-
-    } catch (IOException error) {
-      sb.append("IOError: ").append(error.toString()).append("\n");
-    } catch (PredictException error){
-      sb.append("PredictException: ").append(error.toString()).append("\n");
-    }
-
     // Create reconstruction
     Log.info("Creating full reconstruction.");
     final Frame recon_train = mymodel.score(train);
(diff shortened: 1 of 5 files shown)


h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -383,6 +387,13 @@ private RowData reconstructedToRowData(double[] reconstructed) {
     return result;
   }
 
+  private double calculateReconstructionError(double [] original, double [] reconstructed, double [] normmul){
+    double l2 = 0;
+    for (int i = 0; i < original.length; i++)
+      l2 += Math.pow((reconstructed[i] - original[i]) * normmul[i], 2);
+    return  l2 / original.length;
```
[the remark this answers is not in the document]
maurever: The requirement was to add a method to calculated reconstruction error from MOJO AutoEncoder. I have found that every input data are normalized at the beginning of algorithm in the AutoEncoder model, see [DeepLearningModel.java line 873](https://github.com/h2oai/h2o-3/blob/29133ce8539c350c501a4b5cc594f25214bd2e2d/h2o-algos/src/main/java/hex/deeplearning/DeepLearningModel.java#L872) and [Neurons.java lines 529-533](https://github.com/h2oai/h2o-3/blob/79965ac8bf830063cb1aab4a42031f008f2d7292/h2o-algos/src/main/java/hex/deeplearning/Neurons.java#L529)).

I tested the calculation of MSE (which @michalkurka added in this [PR](https://github.com/h2oai/h2o-3/pull/1344), with train data, but the normalization is missing (if you compare to MSE in `model._output.errors.scored_train._mse`), and results were not the same. After adding the normalization to the calculation, the result is correct.

If you look into [DeepLearningAutoEncoderTest.java lines 100-108](https://github.com/h2oai/h2o-3/blob/79965ac8bf830063cb1aab4a42031f008f2d7292/h2o-algos/src/test/java/hex/deeplearning/DeepLearningAutoEncoderTest.java#L100), there is also a manual calculation of MSE with normalization.

h2o-algos/src/test/java/hex/deeplearning/DeepLearningAutoEncoderTest.java
```diff
@@ -151,12 +157,37 @@ public void run() {
...
+              tmpRow = new RowData();
+              for (int c = 0; c < train.numCols(); c++) {
+                tmpRow.put(train.names()[c], train.vec(c).at(r));
+              }
+              tmpPrediction = model.predictAutoEncoder(tmpRow);
+              calcNormMse += tmpPrediction.mse;
+            }
+            double mojoMeanError = calcNormMse/train.numRows();
+            sb.append("Mojo mean reconstruction error (train): ").append(mojoMeanError).append("\n");
+            Assert.assertTrue("Mean reconstruction error should be the same from model compate to mojo model " +
```
michalkurka: Maybe `assertEquals` with a tolerance would be better?

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -383,6 +387,18 @@ private RowData reconstructedToRowData(double[] reconstructed) {
     return result;
   }
 
+  private double calculateReconstructionError(double [] original, double [] reconstructed, double [] normmul, int numNumeric){
```
michalkurka: Maybe we can move this function to DeeplearningMojoModel? Then you would not need to access `mojoModel._normmul` and `mojoModel._nums`. This would also expose the functionality to more users.

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -383,6 +387,18 @@ private RowData reconstructedToRowData(double[] reconstructed) {
...
+    double l2 = 0;
+    int numIndex = 0;
+    double norm = 0;
+    for (int i = 0; i < numNumeric; i++) {
+      numIndex = i + (original.length - numNumeric);
+      norm = (normmul != null) ? normmul[i] : 1;
+      l2 += Math.pow((reconstructed[numIndex] - original[numIndex]) * norm, 2);
+    }
+    return  l2 / numNumeric;
```
michalkurka: What if there are no numeric features (numNumeric == 0)?

h2o-genmodel/src/main/java/hex/genmodel/easy/prediction/AutoEncoderModelPrediction.java
```diff
@@ -24,4 +24,6 @@
    * Example: input RowData([sex: "Male", ..]) will produce output RowData([sex: [Male: 0.9, Female: 0.1], ..]
    */
   public RowData reconstructedRowData;
-}
\ No newline at end of file
+
+  public double mse;
```
michalkurka: Missing documentation ;)

michalkurka:
Looks good @maurever, minor improvements requested.

h2o-algos/src/test/java/hex/deeplearning/DeepLearningAutoEncoderTest.java
maurever: Yes, definitely.

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
maurever: Yes, it is a better solution and I like it.

I was just inspired by code in the wrapper, where similar issues are solved by this way... I will look at it...

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
maurever: OMG, of course...

h2o-algos/src/test/java/hex/deeplearning/DeepLearningAutoEncoderCategoricalTest.java
```diff
@@ -84,16 +85,43 @@ public void run() {
...
     try {
       DeeplearningMojoModel mojoModel = (DeeplearningMojoModel) mymodel.toMojo();
       EasyPredictModelWrapper model = new EasyPredictModelWrapper(mojoModel);
       AutoEncoderModelPrediction tmpPrediction;
       RowData tmpRow;
+      BufferedString bStr;
       double calcNormMse = 0;
       for (int r = 0; r < train.numRows(); r++) {
         tmpRow = new RowData();
+        bStr = new BufferedString();
```
michalkurka: It doesn't matter in test but BufferedString is meant to be reused, it shouldn't be created inside of a loop.

h2o-algos/src/test/java/hex/deeplearning/DeepLearningAutoEncoderCategoricalTest.java
```diff
@@ -84,16 +85,43 @@ public void run() {
 
     Assert.assertEquals(l2vec.mean(), mymodel.mse(), 1e-8*mymodel.mse());
 
+    // Create reconstruction
+    Log.info("Creating full reconstruction.");
+    final Frame recon_train = mymodel.score(train);
+    Assert.assertTrue(mymodel.testJavaScoring(train,recon_train,1e-5));
+
+    Frame df1 = mymodel.scoreDeepFeatures(train, 0);
+    Assert.assertTrue(df1.numCols() == 10);
```
michalkurka: the frame will leak if the assertion fails

also I wouldn't use assertTrue for checking equality of numbers, assertEquals is better before it is more verbose

maurever: Yes, that is true.

h2o-algos/src/test/java/hex/deeplearning/DeepLearningAutoEncoderCategoricalTest.java
maurever: Ok.

PR #3103, opened 2018-12-03
PUBDEV-5829 fix model schema V3

maurever:
During fill from implementation `fillFromImpl(...)` method in the ModelSchemaV3, timestamp assignment is missing. 

JIRA: https://0xdata.atlassian.net/browse/PUBDEV-5829


Pscheidl:
JIRA: https://0xdata.atlassian.net/browse/PUBDEV-5829
This issue "blocks" https://0xdata.atlassian.net/browse/PUBDEV-6091 - we need the timestamp exposed as well.

PR #3105, opened 2018-12-04
[Experimental] Allows to use MOJO2 Pipelines to transform H2O Frames

michalkurka:


jakubhava:
Nice!

gradle.properties
```diff
@@ -11,6 +11,9 @@ doAnimalSniffer=false
 # The flag to include ORC support inside default h2o.jar.
 doIncludeOrc=false
 
+# The flag to include MOJO Pipeline support inside default h2o.jar.
+doIncludeMojoPipeline=false
```
jakubhava: Do we even want such a option? It's a proprietary file. 

This might be really helpful for internal testing, but we need to be sure it does not leak anywhere.

h2o-extensions/mojo-pipeline/src/main/java/hex/mojopipeline/MojoPipelineExtension.java
```diff
@@ -0,0 +1,46 @@
...
+    _enabled = hasMojoRuntime();
+    _initialized = true;
+    if (! _enabled) {
+      Log.debug("MOJO Runtime not found");
+    }
+  }
+
+  private boolean hasMojoRuntime() {
+    // relying on implementation - need to improve MOJO2 API
+    return ServiceLoader.load(MojoPipelineFactory.class).iterator().hasNext();
```
jakubhava: Maybe safer would be check we have exactly one implementation? To avoid class collisions

gradle.properties
michalkurka: The dependency is only on mojo-runtime-api (just like it is included in SW I believe). @jakubhava do I get it right?

h2o-extensions/mojo-pipeline/src/main/java/hex/mojopipeline/MojoPipelineExtension.java
michalkurka: I think this is okay, it will fail with a better message when people actually try to use the MOJO. This is just a quick check if we have the implementation at all.

gradle.properties
jakubhava: I see :) The API is fine to use anywhere. I thought this means the implementation. All good then, thx!

michalkurka:
@jakubhava I modified the build script (and code) in order to never directly reference the implementation jar. The implementation jar is extracted from the mojo zip file itself.

Can you please review the code? The buildscript will probably need some changes as well + I will need to work with @michal-raska to mount the right directories in the docker.

PR #3057, opened 2018-11-16
PUBDEV-5959: Client Mode Improvement [WIP!!!]

michalkurka:

PR #3108, opened 2018-12-04
 [PUBDEV-5959] PySparking client is hanging after re-connecting to the H2O external backend

jakubhava:
This PR is for rel-xia to follow our current PR cycle, from https://github.com/h2oai/h2o-3/pull/3057.

Original comment:

## The Issue

The original issue was that when a client disconnected and connected to the cluster before the cluster discovered the original client is gone, it blocked the connection of the new client because of previously opened connections

This was caused by using the old client entry in the Client hash map as the cluster was still trying to  contact the old client on the old sockets. We need to make sure the new sockets are created.

##  The Fix
In order to fix this, we needed to add a new entry the the `AutoBuffer` head. This entry contains the information whether the node is client and the unique id. This entry is of size 2 bytes to minimize to communication overhead. The 1. bit is info whether we are client or not and the last 15 bits are the unique id.

This is needed because we need to be able to discover whether the node is client and if it is, different one, before we call `INTERN.put` of this newly created node.

## Implications

In the future, this additional info can help us separate client from the H2O core as we can now tell early whether the node is client or not

Also as part of this PR, the Client hash map was removed as it was duplication information already available in INTERN. 

The method `reportClient` was also removed. The information about the connected client is now printed as part of the `intern` method as that is always the first place where the new node is created. So the code is not scattered around.

### DIFF
--- a/h2o-core/src/main/java/water/AutoBuffer.java
+++ b/h2o-core/src/main/java/water/AutoBuffer.java
@@ -183,7 +183,10 @@ public final class AutoBuffer {
     _read = true;
     _firstPage = true;
     _chan = null;
-    _h2o = H2ONode.intern(pack.getAddress(), getPort());
+    boolean isClient = AutoBuffer.decodeIsClient(getUniqueId());
+    char uniqueId = AutoBuffer.decodeUniqueId(getUniqueId());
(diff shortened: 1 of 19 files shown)


jakubhava:
@h2o-ops please test!

jakubhava:
So merging, fingers crossed. Thanks a ton @michalkurka for the help, learned some new tricks from your code :)

PR #3110, opened 2018-12-04
PUBDEV-5959: PySparking client is hanging after re-connecting to the H2O external backend

michalkurka:


Tagar:
Thank you so much @jakubhava and @michalkurka for getting this done.
Looking forward to try out 3.22.0.3 release (when matching Sparkling Water release will be out).

PR #3112, opened 2018-12-05
PUBDEV-6091 - expose timestamp in Python & R API

Pscheidl:
JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6091


Pscheidl:
Note to self: To be discussed with @grigory93 first !

h2o-r/tests/testdir_jira/runit_pubdev_6019.R
```diff
@@ -0,0 +1,13 @@
...
+
+
+test.timestamp <- function() {
+  prostate.data <- h2o.importFile(path = locate('smalldata/logreg/prostate_train.csv'))
+  gbm.model <- h2o.gbm(x=c("AGE","RACE","DPROS"),y="CAPSULE",training_frame=prostate.data ,model_id="gbm_prostate_model", ntrees = 1, max_depth = 1, seed = 1)
+  expect_false(is.null(gbm.model@model$timestamp))
+
+}
+
+doTest("H2OTree visitor", test.timestamp)
```
ledell:
```suggestion
doTest("Test model timestamp", test.timestamp)
```

ledell:
@Pscheidl So.... I think the model already has model start time information stored in `start_time`.  In R it does, at least:

```
> gbm.model@model$timestamp
$timestamp
[1] 1.544048e+12
> gbm.model@model$start_time
[1] 1.544048e+12
> gbm.model@model$end_time
[1] 1.544048e+12
> gbm.model@model$run_time
[1] 146
```

Pscheidl:
Sorry, this PR is a mess I admit. I was doing it in a rush in Amsterdam and that's not good.

1. I deleted R API changes.
2. The exposure in python is still needed, so I made it reflect R API at least.
3. Grigory wants it in the summary table. Is everyone ok with it ? @ledell @michalkurka

michalkurka:
@grigory93 can you please explain why is a timestamp helpful in the grid summary? I don't fully understand the use case of multiple runs.

If there are worries about making it too cluttered we can add a parameter "verbose" to the grid summary method or something along those lines.

General guidance is to optimize for the typical use case. Lets see what Gregory has to say about the multiple runs but if it is an edge case then I would agree with @ledell .

Pscheidl:
@michalk see what gregory said ^^^^

Copy&paste:
```plain
When getting models added to the grid object on multiple runs and analyzing its summary it's not clear when the models were ran (created), that's why timestamp would have been helpful to me (and I expect to others who are re-using grid objects).
```

ledell:
I also don't see a huge value in returning timestamp by default - I'd prefer a tidy grid summary object with just the most essential info: model id and param values.

ledell:
@Pscheidl If Python models don't have the `start_time` `end_time` and `run_time` objects, then it would be good to add them (I haven't checked if they are already there).

ledell:
@Pscheidl Ok, sounds good -- let's add these to Python (only) then.

h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -257,6 +257,9 @@ def _resolve_model(self, model_id, model_json):
         m._metrics_class = metrics_class
         m._parms = self._parms
         m._estimator_type = self._estimator_type
+        m._start_time = model_json.get('output', {}).get('start_time', None)
```
michalkurka: just to make are these always present?

Pscheidl: If not, the outcome is None. But yes, these are added to all models.

Pscheidl: Model has methods startClock and stopClock. As `Model` is parent to all other mdoels, these are always in place.

PR #3113, opened 2018-12-05
[PUBDEV-6110] Prepare build for new Gradle, and mainly unblock SW-H2O builds

jakubhava:


build.gradle
```diff
@@ -29,6 +28,7 @@ buildscript {
 plugins {
     id "java"
     id "com.github.hierynomus.license" version "0.14.0"
+    id "net.ltgt.apt" version "0.19"
```
jakubhava: need to upgrade the plugin to resolve the bug which occurs when building SW as `./gradlew build -x check --include-build /path/to/h2o`

build.gradle
```diff
@@ -292,11 +292,6 @@ subprojects {
     }
 }
 
-task wrapper(type: Wrapper) {
```
jakubhava: Custom wrapper task is not required and actually prevents the build

jakubhava:
thanks!

PR #3114, opened 2018-12-05
Rel-xia-PUBDEV-5990-arrange-OOM

wendycwong:
This PR resolves JIRA: https://0xdata.atlassian.net/browse/PUBDEV-5990?filter=-1

Work done:
1. Instead of saving all the rows for a MSB in a RPC call, I have broken down the rows into batches.  Only one batch need to be saved at a time.  This reduces the memory requirement.  However, this can potentially increase the number of batches that need to be fetched from neighboring nodes and increase run time.  The batchsize decision is important in this case.  I will perform experiment to capture this.
2. Added R and Python unit tests to ensure that sorting is done correctly and no OOM message is received.  The datasets are generated from Lauren's parameters.

### DIFF
--- a/h2o-core/src/main/java/water/rapids/RadixOrder.java
+++ b/h2o-core/src/main/java/water/rapids/RadixOrder.java
@@ -48,7 +48,7 @@ class RadixOrder extends H2O.H2OCountedCompleter<RadixOrder> {
     // it when aligning two keys in Merge()
     int keySize = ArrayUtils.sum(_bytesUsed);
     // 256MB is the DKV limit.  / 2 because we fit o and x together in one OXBatch.
-    int batchSize = 256*1024*1024 / Math.max(keySize, 8) / 2 ;
+    int batchSize = 1048576 ; // larger, requires more memory with less remote row fetch and vice versa for smaller
     // The Math.max ensures that batches of o and x are aligned, even for wide
     // keys.  To save % and / in deep iteration; e.g. in insert().
     System.out.println("Time to use rollup stats to determine biggestBit: " + ((t1=System.nanoTime()) - t0) / 1e9); t0=t1;
--- a/h2o-core/src/main/java/water/MemoryManager.java
+++ b/h2o-core/src/main/java/water/MemoryManager.java
@@ -1,14 +1,15 @@
 package water;
 
+import jsr166y.ForkJoinPool;
+import jsr166y.ForkJoinPool.ManagedBlocker;
+import water.util.Log;
+import water.util.PrettyPrint;
+
+import javax.management.Notification;
+import javax.management.NotificationEmitter;
 import java.lang.management.*;
 import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicLong;
-import javax.management.Notification;
-import javax.management.NotificationEmitter;
-import jsr166y.ForkJoinPool.ManagedBlocker;
-import jsr166y.ForkJoinPool;
-import water.util.Log;
-import water.util.PrettyPrint;
 
 /**
  * Manages memory assigned to key/value pairs. All byte arrays used in
--- a/h2o-core/src/main/java/water/rapids/BinaryMerge.java
+++ b/h2o-core/src/main/java/water/rapids/BinaryMerge.java
@@ -26,11 +26,12 @@ class BinaryMerge extends DTask<BinaryMerge> {
   private transient long _retLen[/*n2GB*/][];   // How many rows does it match to?
 
   final FFSB _leftSB, _riteSB;
+  final boolean _onlyLeftFrame;  // denote if only left frame is available which implies sorting.
   private transient KeyOrder _leftKO, _riteKO;
 
   private final int _numJoinCols;
   private transient long _leftFrom;
-  private transient int _retBatchSize;
+  private transient int _retBatchSize; // no need to match batchsize of RadixOrder.
 
   private final boolean _allLeft, _allRight;
   private boolean[] _stringCols;
@@ -65,7 +66,7 @@ class BinaryMerge extends DTask<BinaryMerge> {
       _base = base;
       // Create fast lookups to go from chunk index to node index of that chunk
       Vec vec = _vec = frame.anyVec();
-      _chunkNode = vec==null ? null : new int[vec.nChunks()];
+      _chunkNode = vec==null ? null : MemoryManager.malloc4(vec.nChunks());
       if( vec == null ) return; // Zero-columns for Sort
       for( int i=0; i<_chunkNode.length; i++ )
         _chunkNode[i] = vec.chunkKey(i).home_node().index();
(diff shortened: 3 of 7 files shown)


h2o-r/tests/testdir_munging/runit_pubdev_5990_arrange_OOM.R
```diff
@@ -0,0 +1,17 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../scripts/h2o-r-test-setup.R")
+
+test_arrange_OOM<- function(){
```
Pscheidl: I think we can drop either the Python test or the R test @wendycwong . It does the same thing and only prolongs the time it takes to finish the test suite.

h2o-core/src/main/java/water/rapids/RadixOrder.java
```diff
@@ -48,7 +48,8 @@ public void compute2() {
     // it when aligning two keys in Merge()
     int keySize = ArrayUtils.sum(_bytesUsed);
     // 256MB is the DKV limit.  / 2 because we fit o and x together in one OXBatch.
-    int batchSize = 256*1024*1024 / Math.max(keySize, 8) / 2 ;
+  //  int batchSize = 256*1024*1024 / Math.max(keySize, 8) / 2 ; // original value
+    int batchSize = 10000 ; // larger, requires more memory with less remote row fetch and vice versa for smaller
```
Pscheidl: This is the actual fix ? I don't see any memory consumption changes in the Radix sort.

Pscheidl: - Isn't this change going to actually slow down the algorithms on machines with plenty of memory ?
- Shouldn't we calculate this dynamically based on available memory (whole cluster has to be taken into account) ?

h2o-r/tests/testdir_munging/runit_pubdev_5990_arrange_OOM.R
wendycwong: Removed the R test.  Thanks Pavel

h2o-core/src/main/java/water/rapids/RadixOrder.java
wendycwong: Pavel:

The batchsize of RadixOrder.java will not affect memory consumption in this case.  We are still going to look at the complete sort_cols before determining the orders.  However, you are looking at very few columns and hence, reading the whole frame here is fine.  

The actual change occurs in BinaryMerge.java where we know the correct row order and where the rows are in terms of chunk index and node index.  In this case, I am changing the code so that it will only fill up one batch at a time.  In this case, you are reading in a batch of the frame that contains ALL the columns.  Here, the batchsize matters.  I am setting the batchsize to be just one number.

Small batchsize will generate many batches and hence, you are collecting batches more often from all the nodes/chunks.  This can slow down the sort.  However, in normal data frame, you do not expect to see duplicate values inside a row and hence the total number of batches are mostly one.  It is when you hit the large duplicated rows that you will run into the slow down.  However, if you set the batchsize to be too big, your operation may not complete as the machine will run out of memory for the thread.  My change does not affect the merge operation.

I did a small experiment off the Hadoop and found this using Lauren dataset:
BatchSize, with 10GB Hadoop |   | Original with 30GB Hadoop
-- | -- | --
pow(2,20)=1048576 | pow(2, 14) = 16384 |  
65.13598084 | 77.55316401 | 80.4890089
68.44324303 | 74.94142485 | 84.53140998
65.77826309 | 74.94142485 | 69.37319779
71.68452096 | 77.12972713 | 64.26113796
66.98057103 | 80.23118186 | 64.36953998
68.62468004 | 75.68329501 | 68.41652203
70.24415207 | 76.96755791 | 67.93818879
  | 79.98730803 | 71.41601896
  |   |  
68.12734444 | 77.17938545 | 71.34937805

Note that the slow down is not too sensitive to batchsize as long as it is not too small.  To make things easy, I will choose between 16384 or above.  I will do another set of profiling with merge to just make sure nothing was affected.  Is this clear to you?

wendycwong: Pavel:

I thought about using dynamic batchsize but choose not to do it.  The reason is because a RPC is called for merging the portion of the frame with the same MSB.  Since I do not know how many MSBs there are in the frame, I do not know the number of RPCs that will be invoked.  Hence, with different machine cores, memory sizes, this just gets too complicated.  Instead I performed a small measurement experiment and choose the batch size from there.  Once the batch size is determined, I did another set of experiments to measure the performance of sort and merge before and after my change and here are the results:


Merge run time with 50GB |   |  
-- | -- | --
rel-xia | arrange |  
729.1819911 | 463.903742 |  
719.8644481 | 458.936582 |  
702.2259722 | 461.642022 |  
642.1265712 | 460.872243 |  
649.1899869 | 462.709643 |  
631.3483431 | 459.360536 |  
640.1016922 | 458.373506 |  
636.0594559 | 458.688039 |  
657.282526 | 462.229692 |  
766.295264 | 457.157273 |  
677.3676251 | 461.453146 |  
677.3676251 | 460.48422 | Average(s)
sort with 30G |   |  
rel-xia | arrange |  
18.12790608 | 14.62477 |  
18.40748 | 14.68489 |  
21.90696502 | 14.25093 |  
19.80440092 | 14.93063 |  
19.18315411 | 14.84824 |  
18.22094703 | 15.56067 |  
17.98486495 | 16.17139 |  
21.51360393 | 16.32791 |  
20.96344399 | 19.08546 |  
19.32334495 | 13.82125 |  
19.5436111 | 15.4306141 | Average(s)

wendycwong: Since the two results are comparable in orders of magnitude, I consider my implementation did not slow down the sort/merge by too much.  The hadoop cluster changes all the time and I consider the results comparable if they have the same order of magnitude.  Does this help alleviate your concern?  Let me know.  Thanks, Wendy

Pscheidl: @wendycwong Of course, the slow down I was talking about is within the same order of magnitude and thus we can potentially ignore it. This is good enough. Definitely an improvement over what we had previously.

Having the value dynamically set would be benificial - there will be no slowdown (if not even performance increase) for machines with large memory. I understand your reasons. It would be too complicated to do right now and there is no real benefit.

wendycwong:
Okay, after a few experimentation, the final batchsize is going to be 1048576.  Here are the speed profiles of this batchsize in merging and sorting against master:


Merge run time with 50GB |   |  
-- | -- | --
rel-xia | arrange-131072 |  
729.1819911 | 458.873219 |  
719.8644481 | 465.4779956 |  
702.2259722 | 468.4030662 |  
642.1265712 | 462.910486 |  
649.1899869 | 469.7552221 |  
631.3483431 | 468.4336309 |  
640.1016922 | 461.788543 |  
636.0594559 | 468.0818651 |  
657.282526 | 470.2982311 |  
766.295264 | 464.8148432 |  
677.3676251 | 470.4183922 |  
677.3676251 | 466.295954 | Average(s)
sort with 30G |   |  
rel-xia | arrange-131072 |  
18.12790608 | 14.748353 |  
18.40748 | 15.35378981 |  
21.90696502 | 17.97243905 |  
19.80440092 | 14.50938606 |  
19.18315411 | 14.25310612 |  
18.22094703 | 15.14144087 |  
17.98486495 | 14.57082486 |  
21.51360393 | 16.26566386 |  
20.96344399 | 18.47945499 |  
19.32334495 | 14.97219896 |  
  | 14.65059495 |  
19.5436111 | 15.53793205 | Average(s)

PR #3101, opened 2018-11-30
PUBDEV-5894 / PUBDEV-5820 - CDH 6.0

Pscheidl:
This branch contains commits from https://github.com/h2oai/h2o-3/pull/3092, as this PR is not merged yet and we wanted to test it.

PR #3115, opened 2018-12-06
PUBDEV-5820: Core dependency on Jetty must be just testRuntimeOnly

pkozelka:
... otherwise two jettys get into runtime, right after introducing Jetty 9.


pkozelka:
All java tests fail on missing library:
```
==> sandbox/multi/out.5.3 <==
Exception in thread "main" java.lang.NoClassDefFoundError: javax/servlet/ServletOutputStream
	at water.init.NetworkInit.initializeNetworkSockets(NetworkInit.java:76)
	at water.H2O.startLocalNode(H2O.java:1544)
	at water.H2O.main(H2O.java:2001)
Caused by: java.lang.ClassNotFoundException: javax.servlet.ServletOutputStream
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:582)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:190)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:499)
	... 3 more
```

@michal-raska we need to add jetty jars to `test-package-java` (it should take all transitive deps of `h2o-jetty-8`).
Can you please do it within this PR?

Pscheidl:
As Jetty 9 has been merged into master, I rebased this PR onto latest master as well.

michal-raska:
> All java tests fail on missing library:
> 
> ```
> ==> sandbox/multi/out.5.3 <==
> Exception in thread "main" java.lang.NoClassDefFoundError: javax/servlet/ServletOutputStream
> 	at water.init.NetworkInit.initializeNetworkSockets(NetworkInit.java:76)
> 	at water.H2O.startLocalNode(H2O.java:1544)
> 	at water.H2O.main(H2O.java:2001)
> Caused by: java.lang.ClassNotFoundException: javax.servlet.ServletOutputStream
> 	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:582)
> 	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:190)
> 	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:499)
> 	... 3 more
> ```
> @michal-raska we need to add jetty jars to `test-package-java` (it should take all transitive deps of `h2o-jetty-8`).
> Can you please do it within this PR?

I don't think it's enough to add them to test-package-java. We need gradle to put them on classpath. That's why I added the `testCompile project(":h2o-jetty-${defaultJettyVersion}")` dependencies in #3101.

jakubhava:
@michal-raska the issue is that adding testCompile of specific jetty does not really solve the problem. True, it solves the problem for production artifacts but the tests of h2o-core will still `hardcode` specific version.

We need only fat jars ( standalone and hadoop jars) to pick specific version, h2o-core needs to be version independent also in a way that we don't specify it in build.gradle.

jakubhava:
Also I think that we decided to remove `${defaultJettyVersion}"` from gradle.properties? To avoid the confusion between versions

michal-raska:
@jakubhava after discussion with @pkozelka we decided to move the implementation to #3101. Also here is short changelog:

- remove `defaultJettyVersion` and use `defaultWebserverModule` (the motivation is, once you have to change the Jetty, you can do it in one place)
- the `testCompile` dependency on Jetty 8 will become `testRuntimeOnly`

The decision has been made because of the following:
We need jetty on classpath; that can be solved either by adding the `testRuntimeOnly` dependency or by adding it to the classpath manually (in `gradle/multiNodeTesting.gradle`). We agreed that adding `testRuntimeOnly` dependency is ok for now, and will be modified later.

Is there any other possibility how to add the jetty to classpath? Maybe I'm overlooking something.

jakubhava:
I still believe this will cause issues. h2o-core should not require any specific jetty in it's dependencies specifications.

For Sparkling Water, this is fine as it's just related to  h2o tests now. If this is just temporary solution, then the intention and the evil side effects need to be clearly mentioned somewhere and fixed asap so we don' create another technical debt which will be forgotten

PR #3118, opened 2018-12-06
[HEXDEV-717] enable SPNEGO authentication

honzasterba:
back end part done, working on Python API support

### DIFF
--- a/h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
+++ b/h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
@@ -8,9 +8,18 @@ import org.eclipse.jetty.security.DefaultIdentityService;
 import org.eclipse.jetty.security.HashLoginService;
 import org.eclipse.jetty.security.IdentityService;
 import org.eclipse.jetty.security.LoginService;
+import org.eclipse.jetty.security.SpnegoLoginService;
 import org.eclipse.jetty.security.authentication.BasicAuthenticator;
 import org.eclipse.jetty.security.authentication.FormAuthenticator;
-import org.eclipse.jetty.server.*;
+import org.eclipse.jetty.security.authentication.SpnegoAuthenticator;
+import org.eclipse.jetty.server.ConnectionFactory;
+import org.eclipse.jetty.server.Connector;
+import org.eclipse.jetty.server.Handler;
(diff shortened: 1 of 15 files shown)


h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
```diff
@@ -1,15 +1,10 @@
...
 
 import org.eclipse.jetty.jaas.JAASLoginService;
-import org.eclipse.jetty.security.Authenticator;
-import org.eclipse.jetty.security.ConstraintMapping;
-import org.eclipse.jetty.security.ConstraintSecurityHandler;
-import org.eclipse.jetty.security.DefaultIdentityService;
-import org.eclipse.jetty.security.HashLoginService;
-import org.eclipse.jetty.security.IdentityService;
-import org.eclipse.jetty.security.LoginService;
+import org.eclipse.jetty.security.*;
```
pkozelka: pls keep imports explicit

honzasterba:
now how do hell do we test this? :)

h2o-py/requirements.txt
```diff
@@ -2,6 +2,8 @@ pip
 colorama>=0.3.8
 future
 requests
+winkerberos >= 0.5.0; sys.platform == 'win32'
```
michalkurka: Just please keep in mind It shouldn't be a hard requirement, only an optional one.

h2o-py/h2o/auth.py
```diff
@@ -0,0 +1,60 @@
+# -*- encoding: utf-8 -*-
+"""
+Data class for SPNEGO authentication.
+
+:copyright: (c) 2016 H2O.ai
+:license:   Apache License Version 2.0 (see LICENSE for details)
+"""
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+try:
```
michalkurka: kerberos should be optional, we need to make sure the code doesn't fail if kerberos is not installed

and produce an user-understandable error message

michalkurka:
As perfect as anything kerberos related can be ;)

h2o-py/h2o/auth.py
honzasterba: added a nicer message, this fails only when user tries to explicitly import from h2.auth, does not influence any other imports

honzasterba:
@michalkurka should be ready to merge, I would like to do the tests in another PR

PR #3120, opened 2018-12-07
Fix editor config to make diffs meaningful

michalkurka:


pkozelka:
Maybe it's better to retarget this to `rel-xia`

michalkurka:
@pkozelka I see, makes sense - will do

PR #3122, opened 2018-12-08
Rel-Xia-HEXDEV-715 spencer PR

wendycwong:
Review is being done on spencer_2309-2310.  This PR is here to make sure it runs with rel-xia branch.

Add final to variable declaration.
HEXDEV-715: Fixing new chunk.
HEXDEV0715: Fixed the error for testNumberFormat and C1 and C2 chunk tests are passing.
HEXDEV-715: Fixed sorting bug.
HEXDEV-715: Fixed bug in NewChunk.java when using double is not lossy but really should be using Long when the integer fits into Long format.
HEXDEV-715: Fixed NewChunk.java to only update long min/max when appropriate.
HEXDEV-715: This is a working version before any optimization.  Always come back here.
HEXDEV-715: Fixed speed tests.
HEXDEV-715: Replace BigInteger with long.
HEXDEV-715: Minor optimization and incorporate Michalk code review.
HEXDEV-715: Incorporate Michalk code reviews.
            Added parsing of Long.MAX_VALUE and Long.MIN_VALUE to make sure it works.

### DIFF
--- a/h2o-core/src/main/java/water/fvec/NewChunk.java
+++ b/h2o-core/src/main/java/water/fvec/NewChunk.java
@@ -9,7 +9,11 @@ import water.util.PrettyPrint;
 import water.util.StringUtils;
 import water.util.UnsafeUtils;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.HashMap;
+import java.util.UUID;
+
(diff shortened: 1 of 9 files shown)


wendycwong:
May not want this one into fix release.

PR #3126, opened 2018-12-10
[PUBDEV-6115] Fix output of rankTsv in Leaderboard

jakubhava:
Fix output of rankTsv in Leaderboard. Previously, it tried to print Arrays of metrics directly (`defaultMetricForModel(m)` returns array), however Java arrays don't have readable` toString` method. Just `hashcode` of that array was printed out

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
@@ -524,7 +524,7 @@ public class Leaderboard extends Keyed<Leaderboard> {
     return new String[] {"unknown"};
   }
 
-  public String rankTsv() {
+  String rankTsv() {
     String lineSeparator = "\n";
 
     StringBuilder sb = new StringBuilder();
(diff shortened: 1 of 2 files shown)


jakubhava:
It was used by one user from withing Sparkling Water, but its not used inside Sparkling Water internals. The users used it on their own

seb-h2o:
interesting... ideally we should have extension points for those kind of customizations: on our side it's weird to have public methods like this one that are not exposed through rest api and not used internally...

jakubhava:
I will leave this decision up on you :) Right now I just need to fix this tiny thing so it gives reasonable output to the person who is using that

seb-h2o:
sure! didn't suggest you to do anything, was just thinking out loud :)

jakubhava:
yup, but back to that argument, I agree. We need to think a lot before we make anything public, even package-private. Once people start using that there's no way back. I should probably keep this method package-private as it was before as well

seb-h2o:
yes please, keep it package-private: if user was able to access it with this visibility until now, let's not tempt anyone else.

jakubhava:
@seb-h2o added test :)

seb-h2o:
thx @jakubhava , feel free to merge (I guess build stage is failing on legacy test? didn't check personally)

ledell:
@jakubhava Can you add to the PR description what this change does?  It's not clear from looking at the code.

jakubhava:
@ledell thx done, it's fairly tiny & straightforward fix

jakubhava:
@h2o-ops please test!

PR #3127, opened 2018-12-10
[HEXDEV-717] add kerberos login to hadoop tests

honzasterba:


michal-raska:
@honzasterba please let me know before merging, we'll need to make sure we have latest build of hadoop images.

PR #3128, opened 2018-12-10
[PUBDEV-6116] Relocate guava inside h2o-persist-gcs

jakubhava:
Shadow the libraries and depend on those. Ultimately enabling GCS in Sparkling Water

### DIFF
--- a/h2o-persist-google-cloud-storage-shadow/build.gradle
+++ b/h2o-persist-google-cloud-storage-shadow/build.gradle
@@ -1,0 +1,26 @@
+description = "Google Cloud Storage library in one single shadow jar, relocated under ai.h2o"
+
+import com.github.jengelman.gradle.plugins.shadow.tasks.ConfigureShadowRelocation
+apply plugin: 'com.github.johnrengelman.shadow'
+
+dependencies {
+    compile 'com.google.cloud:google-cloud-storage:1.19.0'
+}
+
+shadowJar {
+    mergeServiceFiles()
+    classifier = ''
+}
+
+
+task relocateShadowJar(type: ConfigureShadowRelocation) {
+    target = tasks.shadowJar
+    prefix = "ai.h2o"
+
+}
+
+tasks.shadowJar.dependsOn tasks.relocateShadowJar
+
+
+jar.enabled = false
+build.dependsOn shadowJar
--- a/build.gradle
+++ b/build.gradle
@@ -78,6 +78,7 @@ ext {
       project(':h2o-ext-krbstandalone'),
       project(':h2o-ext-authsupport'),
       project(':h2o-ext-jython-cfunc'),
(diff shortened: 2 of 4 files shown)


h2o-persist-google-cloud-storage-shadow/build.gradle
```diff
@@ -0,0 +1,26 @@
+description = "Google Cloud Storage library in one single shadow jar, relocated under ai.h2o"
```
michalkurka: This will shadow all possible dependencies. Are you sure that is not an issue for any of them?

jakubhava: Yup, we were discussing this solution with @michal-raska and it seemed liked easiest and the most manageable solution. I would wait for tests results to tell us more and check if persist-gcs tests passed

michalkurka: it also relocates `jodatime` which used to be relocated in h2o (it still might be. If it is there would be a clash of dependencies in the final assembly - not a good thing. Generally speaking, doing this can be dangerous because you do lose control over the dependencies you are using and if you are relocating some of them in your own project you can run into weird issues.

jakubhava: Good point. In that case we need to find a way how to relocate just guava. The issue is that a lot of google libraries are using that guava version and we need to to force them to use ours

michalkurka:
@Pscheidl do we have a test for GCS using Python/R? We do need a test that is using the `h2o.jar`

h2o-persist-google-cloud-storage-shadow/build.gradle
jakubhava: I will change the script to just relocate guava, not everything.

I didn't do this directly as I didn't find a way how to relocate transitive dependencies. But maybe shadowJar has support for that

michalkurka:
Note: This should also be tested using Hadoop tests.

h2o-persist-google-cloud-storage-shadow/build.gradle
michalkurka: I will think about an alternative solution.

Usually, the relocation should happen at the point where the library causes an issue. This is a bit too early - it might be well justified by our use case but I don't think this is the common pattern.

jakubhava: I will try to relocate as part of Sparkling Water

Pscheidl:
@michalkurka No we don't. I'll add them.

jakubhava:
Closing as this is being done in Sparkling Water. We should however thing whether to relocate guava in the future or not as I'm still not sure it's safe to have 2 guava ( ours & hadoop one) on the classpath.

PR #3129, opened 2018-12-10
Eliminate Guava from Core

jakubhava:
With this change, h2o-core does not depend on `guava`


jakubhava:
It's possible I missed something, let's wait for tests to finish

build.gradle
```diff
@@ -136,15 +135,6 @@ ext {
       project(':h2o-algos')
     ]
     
-    // Apt required
```
michalkurka: Why are you removing apt?

jakubhava:
It brings another guava dependency. Not related to runtime & compile as it's just related to annotations. But the major point is that we are not using auto service and therefore, this annotation-processing is dead code

h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/ParquetInt96TimestampConverter.java
```diff
@@ -50,8 +48,8 @@ public static long getTimestampMillis(Binary timestampBinary) {
     byte[] bytes = timestampBinary.getBytes();
 
     // little endian encoding - bytes are red in inverted order
-    long timeOfDayNanos = Longs.fromBytes(bytes[7], bytes[6], bytes[5], bytes[4], bytes[3], bytes[2], bytes[1], bytes[0]);
```
michalkurka: We can also use UnsafeUtils.get8/get4 to do the same.

michalkurka:
@jakubhava if it is not affecting the dependencies, can we remove it in a separate PR? 

I think removing apt is not a bad idea but I would do it independently.

michalkurka:
Great job @jakubhava, definitely a good effort.

h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/ParquetInt96TimestampConverter.java
jakubhava: These functions are the ones used in Guava for exact same functionality

jakubhava:
NEW PR for just removing APT https://github.com/h2oai/h2o-3/pull/3130

michalkurka:
@jakubhava PR looks good. If any of the code is copied from another project - we need to make sure all the license information is preserved.

I suggest separating the copied functions to their own files. To make the distinction clear.

jakubhava:
@michalkurka  good point about licensing and the separation. I also found that a few tests are failing but that's because of guava is missing and I forgot to replace all occurrences of guava code. Will fix tomorrow

jakubhava:
Guava is removed from `compile` & `runtime` libraries. Guava is used in tests, which I think is fine as it does not influences the production code

jakubhava:
@h2o-ops please test!

jakubhava:
I'm happy with the change, waiting for tests

jakubhava:
@michalkurka I thought that just referencing the source file is enough. I can add the full license of the code we are using, just to be safe. Thank you!

jakubhava:
@michalkurka added the licenses, please have a look of this looks good for you :)

jakubhava:
As the outcome, the final `h2o.jar` is 5mb smaller

PR #3130, opened 2018-12-11
remove unused auto-service (also bringing guava dependency

jakubhava:
(cherry picked from commit 4a0904a25abd1d5575c2c9459a79ea379f94eecc)


jakubhava:
New PR for removing APT.

It is not used at this stage, so basically just dead code.

PR #3136, opened 2018-12-13
PUBDEV-6129 Added support for using indices when specifying columns in parameters of Target Encoding in Python's API

deil87:
Purpose of PR is to unify R and Python's target encoding API's
1) added support for indices for specifying columns in parameters of TE ( adjust descriptions)
2) @angela0xdata found mismatch in naming https://0xdata.atlassian.net/browse/PUBDEV-6126  . Changed blending_avg -> blended_avg to match with the first version of R's TE.
3) correct example in description for target encoding

### DIFF
--- a/h2o-py/h2o/targetencoder.py
+++ b/h2o-py/h2o/targetencoder.py
@@ -28,7 +28,7 @@ class TargetEncoder(object):
 
     Sample usage:
 
-    >>> targetEncoder = TargetEncoder(x=te_columns, y=responseColumnName, blending_avg=True, inflection_point=3, smoothing=1)
+    >>> targetEncoder = TargetEncoder(x=te_columns, y=responseColumnName, blended_avg=True, inflection_point=3, smoothing=1)
     >>> targetEncoder.fit(trainFrame) 
     >>> encodedTrain = targetEncoder.transform(frame=trainFrame, holdout_type="kfold", seed=1234, is_train_or_valid=True)
     >>> encodedValid = targetEncoder.transform(frame=validFrame, holdout_type="none", noise=0.0, is_train_or_valid=True)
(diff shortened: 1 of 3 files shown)


h2o-py/h2o/targetencoder.py
```diff
@@ -28,25 +28,26 @@ class TargetEncoder(object):
...
     # Construction
     #-------------------------------------------------------------------------------------------------------------------
 
-    def __init__(self, x=None, y=None, fold_column='', blending_avg=True, inflection_point=3, smoothing=1):
+    def __init__(self, x=None, y=None, fold_column='', blended_avg=True, inflection_point=3, smoothing=1):
         """
         Creates instance of the TargetEncoder class and setting parameters that will be used in both `train` and `transform` methods.
 
-        :param List[str] x: List of categorical column names that we want apply target encoding to
+        :param List[str] or List[int] x: List of categorical column names or indicess that we want apply target encoding to
```
angela0xdata: change "indicess" to "indices"

angela0xdata:
Thanks @deil87 :)

h2o-py/h2o/targetencoder.py
deil87: @angela0xdata done

deil87:
@michalkurka Could you please take a look at this PR as well?

michalkurka:
@deil87 what do you think about keeping the old parameter `blending_avg` around to maintain backward compatibility?

We can produce a warning if someone uses the old parameter.

michalkurka:
Yes, I think it makes sense to keep both. One marked as deprecated.

I don't think we have guidelines about alpha-stage features. The only guideline is to be mindful to our users and make migrating as easy as possible.

michalkurka:
@deil87 are you planning to make further changes in this PR? Eg. keep the old parameter around as well as the new one?

deil87:
Going to close this branch and open a new one that is rel-xia based

PR #3137, opened 2018-12-13
[PUBDEV-6137] Fix GCS support on Hadoop

jakubhava:


jakubhava:
Before merging, will add hadoop test to verify this actually works

jakubhava:
Added test, waiting for Hadoop tests to run

jakubhava:
@Pscheidl @michalkurka Hadoop tests are passing as well

jakubhava:
Applied the change on rel-xia so this fix can affect also next Sparkling Water release

PR #3138, opened 2018-12-13
Removed is_train_or_valid parameter from Target encoding

deil87:


michalkurka:
@angela0xdata this changes the TE API, technically a breaking change but we didn't promise the API to be stable (was released as alpha quality).

Can we document the change somewhere? WDYT?

angela0xdata:
Hmm... I don't recall stating that it was alpha quality. But this change will be included in the release notes when I note the change for PUBDEV-6130.

PR #3143, opened 2018-12-14
HEXDEV_725: fixed pojo binary comments - show binary representation of bitset completely

pkozelka:
Simple fix.
There was a completely useless check for boundaries, and it contained a bug.
Fixed by just removing it.


pkozelka:
After approval, I would like to add one more commit with some IDEA warnings resolved (and python-like code replaced with more java-like style) - check here if you care: daaf8d614e23cf4a132fb71a2604d468daad0327 (i did not want to push it here now, to make sure it does not disturb review)

michalkurka:
https://github.com/h2oai/h2o-3/commit/daaf8d614e23cf4a132fb71a2604d468daad0327 looks good, feel free to merge

PR #3150, opened 2018-12-19
PUBDEV-6123: fixed SQLManager import queries for MS SQLServer / Azure SQL

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6123

fixed import queries, now using following syntax, both fully supported from SQL Server 2012:

for column names retrieval (supported from SQLS 2008):
```sql
select top(1) <columns> from <table>;
```

for data chunks retrieval (supported from SQLS 2012):
```sql
select <columns> from <table> order by row_number() over (order by (select 0)) offset <start> rows fetch next <N> rows only;
```

see ticket comments for testing protocol

### DIFF
--- a/h2o-core/src/main/java/water/jdbc/SQLManager.java
+++ b/h2o-core/src/main/java/water/jdbc/SQLManager.java
@@ -280,7 +280,9 @@ public class SQLManager {
     switch(databaseType) {
       case SQL_SERVER_DB_TYPE:  // requires ORDER BY clause with OFFSET/FETCH NEXT clauses, syntax supported since SQLServer 2012
         sqlText += " ORDER BY ROW_NUMBER() OVER (ORDER BY (SELECT 0))";
-        //fall through oracle syntax for offset logic
+        sqlText += " OFFSET " + start + " ROWS FETCH NEXT " + length + " ROWS ONLY";
+        break;
+        
       case ORACLE_DB_TYPE:
         sqlText += " OFFSET " + start + " ROWS FETCH NEXT " + length + " ROWS ONLY";
         break;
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/jdbc/SQLManager.java
```diff
@@ -276,8 +278,10 @@ static String buildSelectChunkSql(String databaseType, String table, long start,
 
     String sqlText = "SELECT " + columns + " FROM " + table;
     switch(databaseType) {
+      case SQL_SERVER_DB_TYPE:  // requires ORDER BY clause with OFFSET/FETCH NEXT clauses, syntax supported since SQLServer 2012
+        sqlText += " ORDER BY ROW_NUMBER() OVER (ORDER BY (SELECT 0))";
+        //fall through oracle syntax for offset logic
```
michalkurka: this is will work, do you think there is a chance someone can broke SQL Server by modifying Oracle statement? Maybe it would be better to duplicate instead. WDYT?

seb-h2o: yeah, was hesitating... probably safer to duplicate, it's just one line after all

grigory93: agree, better to duplicate and not mix sql flavors

h2o-core/src/main/java/water/jdbc/SQLManager.java
```diff
@@ -239,8 +239,10 @@ public void compute2() {
   static String buildSelectSingleRowSql(String databaseType, String table, String columns) {
 
     switch(databaseType) {
+      case SQL_SERVER_DB_TYPE: //syntax supported since SQLServer 2008
+        return "SELECT TOP(1) " + columns + " FROM " + table;
+
```
grigory93: should specify in the docs that we support SQL Server 2008 or later

PR #3151, opened 2018-12-19
PUBDEV-6122 - Add a parameter to pdp to save the plot

Pscheidl:
### About
[PUBDEV-6122](https://0xdata.atlassian.net/browse/PUBDEV-6122)
[SUPPORT PORTAL Ticket #93297](https://support.h2o.ai/helpdesk/tickets/93297)
### Python
- I rather chose to modify a current test (almost zero slowdown of test pipeline) that constructs PDPs anyway.
- Existing files are overwritten
- PNG output format (unified with R).

Example call:
```python
pdpOrig = gbm_model.partial_plot(data=data,cols=['AGE', 'RACE', 'DCAPS'],server=True, plot=True, save_to_file="/home/pavel/pdps.png")
```

### R 
- Saves each plot to a separate file. Automatically adds posfix `${FEATURE_NAME}.png`, where `${FEATURE_NAME}` is obviously name of the feature the plot corresponds to. For `n` plots being generated in R, there are `n` files created.
- Plots can still be seen in R, `dev.copy()` is used.
- Existing files are overwritten
- PNG format (unified with Python)
- Test from scratch, uses temp files

Example call:

```R
    plot <- h2o.partialPlot(object = model, data = data, save_to = "/home/pavel/pdp")
```

If the user accidentally uses any of the following prefixes: [png,jpg,pdf], then those are automatically removed.

```R
    plot <- h2o.partialPlot(object = model, data = data, save_to = "/home/pavel/pdp.pdf)
```

Both calls would end in three files created on filesystem:

```plain
/home/pavel/pdp_AGE.png
/home/pavel/pdp_RACE.png
/home/pavel/pdp_DCAPS.png
```

Passing an empty string to `save_to` argument saves just the feature name with `.png` suffix in current workspace, e.g. `/home/pavel/RStudio/_AGE.png`

### DIFF
--- a/h2o-r/h2o-package/R/models.R
+++ b/h2o-r/h2o-package/R/models.R
@@ -3081,7 +3081,7 @@ h2o.cross_validation_predictions <- function(object) {
 #' Inside each list, the first element is the column name followed by values defined by the user.
 #' @param save_to Fully qualified prefix of the image files the resulting plots should be saved to, e.g. '/home/user/pdp'.
 #'  Plots for each feature are saved separately in PNG format, each file receives a suffix equal to the corresponding feature name, e.g. `/home/user/pdp_AGE.png`.
-#'  If the files already exists, they will be overridden.
+#'  If the files already exists, they will be overridden. Files are only saves if plot argument is set to True.
 #' @return Plot and list of calculated mean response tables for each feature requested.
 #' @examples
 #' \donttest{
--- a/h2o-r/h2o-package/R/models.R
+++ b/h2o-r/h2o-package/R/models.R
@@ -3081,7 +3081,7 @@ h2o.cross_validation_predictions <- function(object) {
 #' Inside each list, the first element is the column name followed by values defined by the user.
 #' @param save_to Fully qualified prefix of the image files the resulting plots should be saved to, e.g. '/home/user/pdp'.
 #'  Plots for each feature are saved separately in PNG format, each file receives a suffix equal to the corresponding feature name, e.g. `/home/user/pdp_AGE.png`.
-#'  If the files already exists, they will be overridden.
+#'  If the files already exists, they will be overridden. Files are only saves if plot = TRUE (default).
 #' @return Plot and list of calculated mean response tables for each feature requested.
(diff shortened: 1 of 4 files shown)


h2o-r/h2o-package/R/models.R
```diff
@@ -3096,7 +3101,8 @@ h2o.cross_validation_predictions <- function(object) {
 #' }
 #' @export
 
-h2o.partialPlot <- function(object, data, cols, destination_key, nbins=20, plot = TRUE, plot_stddev = TRUE, weight_column=-1, include_na=FALSE, user_splits=NULL) {
+h2o.partialPlot <- function(object, data, cols, destination_key, nbins=20, plot = TRUE, plot_stddev = TRUE,
```
michalkurka: it seems like the new parameter is called save_to in R and save_to_file in Python, is that intended?

h2o-r/h2o-package/R/models.R
```diff
@@ -3209,6 +3215,14 @@ h2o.partialPlot <- function(object, data, cols, destination_key, nbins=20, plot
       } else {
         plot(pp[,1:2], type = "l", main = attr(x,"description") )
       }
+      if(!is.null(save_to)){
```
michalkurka: Maybe the documentation should say the file will only get written if plot==TRUE

h2o-r/h2o-package/R/models.R
Pscheidl: Yes, this is intentional, as there is rarely one file in R.

h2o-r/h2o-package/R/models.R
Pscheidl: Fixed. Thank you.

h2o-r/h2o-package/R/models.R
michalkurka: Makes sense, thank you @Pscheidl

ledell: Do we have any existing naming conventions in R that we should consider?  (regarding parameters that save to disk...)

PR #3152, opened 2018-12-19
HEXDEV-726: Improve performance of Grid Search in client mode (for SW)

michalkurka:


h2o-core/src/main/java/water/H2O.java
```diff
@@ -1217,16 +1215,42 @@ private PrioritizedForkJoinPool(int p, int cap) {
...
+
+  // package-private for unit tests
+  static <T extends RemoteRunnable> T runOnH2ONode(H2ONode node, T runnable) {
+    if (node == H2O.SELF) {
+      // run directly
+      runnable.run();
+      return runnable;
+    } else {
+      RunnableWrapperTask<T> task = new RunnableWrapperTask<>(runnable);
+      H2ONode leader = H2O.CLOUD.leader();
```
jakubhava: Small note: Should we use the node from the argument?

michalkurka: Good catch, yes, that was the intention. Thanks for catching that.

PR #3158, opened 2018-12-20
Added a note about GBM reproducibility in multinode setting

michalkurka:


h2o-docs/src/product/data-science/gbm-faq/reproducibility.rst
```diff
@@ -7,4 +7,8 @@ Reproducibility
 
 - **How deterministic is GBM?**
 
- As long as you set the seed, GBM is deterministic up to floating point rounding errors (out-of-order atomic addition of multiple threads during histogram building). This means that if you set a seed, your results will be reproducible (even if, for example, you change the number of nodes in your cluster, change the way you ingest data, or change the number of files your data lives in, among many other examples).
```
michalkurka: Not sure why this appears as modified. I didn't change anything AFAIK

h2o-docs/src/product/data-science/gbm-faq/reproducibility.rst
```diff
@@ -7,4 +7,8 @@ Reproducibility
...
+ As long as you set the seed, GBM is deterministic up to floating point rounding errors (out-of-order atomic addition of multiple threads during histogram building). This means that if you set a seed, your results will be reproducible (even if, for example, you change the number of nodes in your cluster, change the way you ingest data, or change the number of files your data lives in, among many other examples).
+
+- **Can reproducibility be guaranteed in multinode clusters?**
+
+Yes, as long as the cluster configuration is the same and the model training is triggered from the leader node of the cluster the resulting model will be the same given the other requirements are satisfied (seed, no early stopping). For two clusters to be considered the same they need to have the same number of nodes and the nodes need to have the same number of CPU cores available (or the same restriction on number of threads).
```
meganjkurka: I think we should add here how to identify the leader node IP and after what version was this IP returned automatically.  I can add this.

maurever:
Related to this JIRA https://0xdata.atlassian.net/browse/PUBDEV-6024. 
Or what else is expected to improve in reproducibility section in the JIRA ^^^?

michalkurka:
@maurever a lot is expected ;)

This only talks about reproducibility using a single version of H2O, no between versions. Quoting from JIRA:

> For example this section should include why you might see difference between version, and what you can do to ensure that you are actually running the same model between versions (note: we do not guarantee that the model between versions will be the same, since we are always trying to improve the performance of your models when we release new versions of h2o)

I expect you will also elaborate on the role of scoring intervals and how reproducibility can be accomplished with early stopping enabled.

Simple examples should illustrate rounding errors and how order of mathematical operations can affect the result. Eg. calculating (x + y) + z and x + (y + z) can give different results because computers work with a limited precision. What role does that play in distributed computation based on map-reduce and so on.

PR #3160, opened 2018-12-20
PUBDEV-6111 Fixed euclidean distance formula in AstDistance rapids expression

Pscheidl:
## About
[PUBDEV-6111](https://0xdata.atlassian.net/browse/PUBDEV-6111)

Simple fix of missing `Math.sqrt()` in euclidean distance. Originally reported by community.

### DIFF
--- a/h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstDistance.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstDistance.java
@@ -129,7 +129,7 @@ public class AstDistance extends AstBuiltin<AstDistance> {
             }
           } else if (l2) {
             for (int c = 0; c < p; ++c) { //cols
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstDistance.java
```diff
@@ -129,7 +129,7 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
             }
           } else if (l2) {
             for (int c = 0; c < p; ++c) { //cols
-              distRQ += Math.pow(cs[c].atd(r) - Qs[c].at(q), 2);
+              distRQ += Math.sqrt(Math.pow(cs[c].atd(r) - Qs[c].at(q), 2));
```
michalkurka: Looks good but please add a test

michalkurka: this is not correct

PR #3165, opened 2018-12-21
PUBDEV-6153: Add support for Cloudera Hadoop 6.1

michal-raska:
There are also some minor changes:

- hive start script is moved to common space (it's the same for CDH and HDP)
- h2odrivers are removed after the stage finishes


h2o-hadoop-3/h2o-cdh6.1-assembly/build.gradle
```diff
@@ -0,0 +1,16 @@
+ext {
+    hadoopVersion = 'cdh6.0.1'
```
michalkurka: We need CDH 6.1, this is 6.0.1

PR #3166, opened 2018-12-21
PUBDEV-6129 Added support for using indices when specifying columns in parameters of Target Encoding in Python's API

deil87:
1) added support for using indices for column-related parameters
2) renamed parameter for blending ( deprecated previous one)


deil87:
@michalkurka it looks like there is no proper overloading in Python especially based on parameter's name and not on its type.  I decided to overcome make it with kwargs. I could have made it by keeping both parameters in the same method ( and use False if one of them was set to false otherwise use default value - >True) but not sure if it looks fine when hint for signature of the method pops up.

deil87:
Moved to https://github.com/h2oai/h2o-3/pull/3189

PR #3172, opened 2018-12-21
PUBDEV-4305 allow long vector to pass

maurever:
Fix R communication to be able to work with a large dataset in R without throwing "Long vectors not supported yet." error.

JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-4305


maurever:
I tried to used `multiple = TRUE` parameter in `rawToChar` method (https://stat.ethz.ch/R-manual/R-devel/library/base/html/rawConversion.html), but after using `paste0` method (to join the result chars together), a similar problem has occurred. I haven't found a better solution yet...

maurever:
I implemented a new approach:
1. using `.h2o.doSafeGET(urlSuffix = urlSuffix, binary=TRUE)` get the data in binary format 
2. get the size of data and then decide if the data are small or too big to use the `rawToChar` method
3. if the data are too big, save them into a temporary file using the `writeBin` method to write the binary data in a chunked form

h2o-r/h2o-package/R/frame.R
```diff
@@ -3395,49 +3431,62 @@ as.h2o.Matrix <- function(x, destination_frame="", ...) {
...
-    ttt2 <- substr(ttt, 1L, n-chars_to_trim)
-    ttt <- ttt2
-  }
-
+    
   # Get column types from H2O to set the dataframe types correctly
   colClasses <- attr(x, "types")
   colClasses <- gsub("numeric", NA, colClasses) # let R guess the appropriate numeric type
-  colClasses <- gsub("int", NA, colClasses) # let R guess the appropriate numeric type
+  colClasses <- gsub("int", NA, colClasses) # let R guess the appropriate numeric typepodobne jako mame
```
michalkurka: please fix the comment

h2o-r/h2o-package/R/frame.R
```diff
@@ -3395,49 +3431,62 @@ as.h2o.Matrix <- function(x, destination_frame="", ...) {
...
+                      '&hex_string=', as.numeric(useHexString))
+  
+  verbose <- getOption("h2o.verbose", FALSE)
+    
+  if (verbose) pt <- proc.time()[[3]]
+  
+  # Get data in binary format for case the data are too big to load in character format  
+  payload <- .h2o.doSafeGET(urlSuffix = urlSuffix, binary=TRUE)
+  
+  if(length(payload) < .Machine$integer.max)  {
```
michalkurka: Good idea! I think it would be good to parametrize this in order to be able to use this code path even for smaller datasets (for testing purposes).

    getOption("h2o.as.data.frame.max.in-memory.payload.size", .Machine$integer.max) 

something like this?

h2o-r/h2o-package/R/frame.R
maurever: Ups.. :)

h2o-r/h2o-package/R/frame.R
```diff
@@ -3471,7 +3527,16 @@ as.data.frame.H2OFrame <- function(x, ...) {
...
 #' }
 #' @export
-as.matrix.H2OFrame <- function(x, ...) as.matrix(as.data.frame.H2OFrame(x, ...))
+as.matrix.H2OFrame <- function(x, ...) {
+  .fetch.data(x,1L)   
+  nCol <- attr(x, "ncol")
+  nRow <- attr(x, "nrow")
+
+  if(nCol * nRow > .Machine$integer.max){
+    stop("It is not possible convert H2OFrame to data.frame/data.table. The H2OFrame is bigger than vector size limit for R.")
```
michalkurka: Isn't this comment misleading? The user is attempting to convert to a `matrix`.

I would also suggest what to do instead -> convert to data.frame which is not limited in the same way.

maurever: Yes, that is absolutely true.

PR #3183, opened 2018-12-27
PUBDEV-5742 Integration of Target Encoding into AutoML

deil87:
This PR implements integration of TE into AutoML's pipeline. Target encoding aspect is being controlled by multiple strategies ( default values of which user can change):

**1. Application strategy** 
Decides which columns to encode in the original frame.

- AllCategoricalTEApplicationStrategy

- ThresholdTEApplicationStrategy

**2. TE parameters selection strategy (hyperparameters search)** 
Decides which set of parameters for target encoding will be used for encoding. This search should be done with respect to the baseline performance without TargetEncoding, as we don't want to apply TE that decreases performance. 

- Current version is using Random Grid Search(RGS) with early stopping.  ( SMBO version with DRF or GP surrogate models is in a separate branch https://github.com/h2oai/h2o-3/pull/3394 )

- After filtering out (with application strategy) categorical columns that we don't want to encode  we incorporate rest of the columns into our grid of hyperparameters so that we can try different combinations of the columns.

- Safety(unfortunately not in terms of time) of applying TE is enforced by having `empty` combination of columns to encode in the grid. But this does not guarantee that RGS will evaluate this scenario so we might want to eagerly evaluate this cornerstone case. We might take this baseline performance into consideration as another criteria for early stopping.

This PR is subject for decomposition. We can separate some functionality to make this PR easier to review. 
Sub PRs:
- Broadcast Join For TE https://github.com/h2oai/h2o-3/pull/3233
- Threshold/All TE application strategy ( select columns that are subject for encoding) https://github.com/h2oai/h2o-3/pull/3975
- Small refactoring in AutoML class :white_check_mark:
- Support for  ModelBuilder's cloning :white_check_mark:
- Random Grid Search for TE hyperparameters
- AstGroup consistency issue ( fixed and closed ) :white_check_mark:
And this PR will be dedicated to integration only.


michalkurka:
Great PR! Thank you @deil87 

What will be the major thing to figure out is how to productionalize such models. The users will need to be able to apply the encoding to new observations. We will definitely need a MOJO for TE.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -225,40 +226,95 @@ public AutoML(Key<AutoML> key, Date startTime, AutoMLBuildSpec buildSpec) {
...
+      String responseColumnName = trainingFrame.name(trainingFrame.find(getResponseColumn()));
+      String foldColumnName = trainingFrame.name(trainingFrame.find(getFoldColumn()));
+
+      Map<String, Frame> encodingMap = tec.prepareEncodingMap(trainingFrame, responseColumnName, foldColumnName, imputeNAsWithNewCategory);
+      switch (holdoutType) {
+        case TargetEncoder.DataLeakageHandlingStrategy.KFold:
+          this.trainingFrame = tec.applyTargetEncoding(trainingFrame, responseColumnName, encodingMap, holdoutType, foldColumnName, withBlendedAvg, imputeNAsWithNewCategory, seed);
+          this.validationFrame = tec.applyTargetEncoding(getValidationFrame(), responseColumnName, encodingMap, TargetEncoder.DataLeakageHandlingStrategy.None, foldColumnName, withBlendedAvg, imputeNAsWithNewCategory, seed);
+          this.leaderboardFrame = tec.applyTargetEncoding(getLeaderboardFrame(), responseColumnName, encodingMap, TargetEncoder.DataLeakageHandlingStrategy.None, foldColumnName, withBlendedAvg, imputeNAsWithNewCategory, seed);
+          break;
```
seb-h2o: I don't know how other strategies work, but do you think it would make sense to have this done in 2 steps?
The reason is that I see a lot of duplicate params between from `prepareEncodingMap` to all those `applyTargetEncoding`, for example:
```
TETask task = tec.setup(trainingFrame, responseColumnName, foldColumnName, imputeNAsWithNewCategory, withBlendedAvg, seed);
this.trainingFrame = task.apply(trainingFrame, strategy);
this.validationFrame = task.apply(validationFrame, strategy2);
...
```
maybe the fact that non-training frames always have None strategy could be implicit, dunno...
wdyt?

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -225,40 +226,95 @@ public AutoML(Key<AutoML> key, Date startTime, AutoMLBuildSpec buildSpec) {
...
+    
+    boolean hasCategoricalVecs = false;
+    for(Vec vec : trainingFrame.vecs()) {
+      if(vec.isCategorical()) {
+        hasCategoricalVecs = true;
+        break;
+      }
+    }
+    
+    if(hasCategoricalVecs) {
```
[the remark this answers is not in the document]
deil87: Good point! In most cases I'm trying not to introduce dependencies between components( especially on the early stages of development). And later if it makes sense I'm optimising. Strategies in general could be different and some of them could be expensive to calculate and we don't wan't to execute it when there are some cheap-calculated evidences that we should not run heavy code. But absence of the columns at all is an edge case that will be the same for all the strategies and you are right. Later I will probably add this simple check to the common ancestor of Strategies. Thank you!

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -225,40 +226,95 @@ public AutoML(Key<AutoML> key, Date startTime, AutoMLBuildSpec buildSpec) {
...
+
+      //TODO Either perform random grid search over parameters or introduce evolutionary selection algo
+      BlendingParams blendingParams = new BlendingParams(5, 1);
+      String[] columnsToEncode = strategy.getColumnsToEncode();
+      boolean withBlendedAvg = true;
+      boolean imputeNAsWithNewCategory = true;
+      int seed = 1234; // TODO make it a parameter for users to set
```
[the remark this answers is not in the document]
deil87: Thanks! I was looking for something that is dedicated to a single purpose. Or at least to rely on the seed that is closely related to my scope. Will take into account your recommendation!

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
deil87: I had an attempt to move some common parameters to the TE class. Use can see with an example of BlendingParams class that it is not being passed twice. But the problem is that we want this two-methods type of API for Target Encoding and we don't have a stateful backend/sessions. When we call from the clients the first method and instantiate TargetEncoding class we can keep it and connect to the second/following calls from the client. For AutoML case it could be possible (and would be a right thing to do) bc there is no communication with the client during TE phase.

deil87: Regarding None and none-training frames. There is no special mechanism for now to separate training and non-training frames. I had a parameter/flag that is close to this but we decided to remove it. Ideally I would have the following API:  TE.fitAndTranform() - for training and transforming of training frame and TE.transform() - for other frames.  But again the first method would have to return frame and not the encoding map..... but we need it to pass with the second method's call. Of course we could return both encoding map and encoded training frame but this would complicate things with serialisation before sending to the client and also it is quite uncommon for users to expect tuples as a return value.

ledell:
@deil87 Can you create a description of what this PR implements?  Also please provide some performance results that show that this could improve the performance of AutoML (on at least a few datasets).  We also need a strategy for avoiding hurting performance if TE does not help.

deil87:
@ledell Sure I will add benchmarks. I always write unit tests, integration tests and benchmarks whenever it is possible. I'm just not there yet in terms of benchmarking.

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_automl_gbm_target_encoding.py
```diff
@@ -86,12 +86,9 @@ def titanic_with_te_loostrategy(frame = None, seeds = None):
     ds = split_data(frame, current_seed)
     targetColumnName = "survived"
 
-    foldColumnName = "kfold_column"
```
deil87: This should probably go into a separate PR that will be fixing  wrong test

deil87:
@ledell Are we planning to merge this PR for upcoming release? I can try to shape it but the most important obstacle is that I guess there is not enough time for others to review it. Good thing is that most of the changes are in automl package and @seb-h2o will be able to review them. Most of the changes to the core from this PR are actually represented in as separate PRs( broadcast join, astgroup inconsistency). Those PRs are added as blockers to this PR.
Also wanted to discuss R/Py support. For now I have multiple parameters in Java to configure TE integration. We can enable TE by default and don't provide R/Py api for turning it on/off... otherwise exposing 'enable/disable' parameter (or even all parameters) is going to take some extra time.

ledell:
@deil87 No, I think we probably want to wait for 3.26 because it seems like a lot of work to finish in 2 days.

deil87:
Merged with master branch to resolve conflicts. Fixed some tests that were failing due to merge. There are still couple of tests that  are leaking/failing. Going to fix them later when this PR is decomposed into smaller pieces.

PR #3185, opened 2018-12-27
Custom early stopping: validate that custom metric was actually defined

michalkurka:


michalkurka:
@angela0xdata just FIY, this doesn't have a documentation impact

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1341,6 +1341,11 @@ else if (_parms._weights_column != null && _weights != null && !_weights.isBinar
         }
       }
     }
+    if (_parms._stopping_metric == ScoreKeeper.StoppingMetric.custom || _parms._stopping_metric == ScoreKeeper.StoppingMetric.custom_increasing) {
```
Pscheidl: A single if would do as well ? 

```java
    if ((_parms._stopping_metric == ScoreKeeper.StoppingMetric.custom || _parms._stopping_metric == ScoreKeeper.StoppingMetric.custom_increasing)
            && _parms._custom_metric_func == null) {
        error("_stopping_metric", "Custom metric function needs to be defined in order to use it for early stopping.");
    }
```

But don't worry about it. :)

michalkurka: yep, seems to me a bit clearer with the nested if

PR #3189, opened 2019-01-02
PUBDEV-6129 Added support for using indices when specifying column-related parameters of target encoding(Python API)

deil87:
1) added support for using indices for column-related parameters
2) renamed parameter for blending ( deprecated previous one)

### DIFF
--- a/h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_automl_gbm_target_encoding.py
+++ b/h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_automl_gbm_target_encoding.py
@@ -53,7 +53,7 @@ def titanic_with_te_kfoldstrategy(frame = None, seeds = None):
 
       teColumns = ["home.dest", "cabin", "embarked"]
       targetEncoder = TargetEncoder(x= teColumns, y= targetColumnName,
-                                    fold_column= foldColumnName, blending_avg= True, inflection_point = 3, smoothing = 1)
+                                    fold_column= foldColumnName, blended_avg= True, inflection_point = 3, smoothing = 1)
       targetEncoder.fit(frame=ds['train'])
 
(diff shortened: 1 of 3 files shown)


h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_automl_gbm_target_encoding.py
```diff
@@ -53,7 +53,7 @@ def titanic_with_te_kfoldstrategy(frame = None, seeds = None):
 
       teColumns = ["home.dest", "cabin", "embarked"]
       targetEncoder = TargetEncoder(x= teColumns, y= targetColumnName,
-                                    fold_column= foldColumnName, blending_avg= True, inflection_point = 3, smoothing = 1)
+                                    fold_column= foldColumnName, blended_avg= True, inflection_point = 3, smoothing = 1)
```
ledell: Is this to match the R API?  Should this be in a separate PR?

ledell:
@deil87 Just double checking -- you are using 0-index for Python, correct?  Also, is this already supported in R (if not, do you plan to add support?).

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_automl_gbm_target_encoding.py
deil87: @ledell the goal was to match the R API. Ideally it should be in a separate PR, yes. Is it ok to  leave it here?

deil87:
@ledell Yes it is supported in Py and R, and I'm using 0-based indexing in both cases.

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_automl_gbm_target_encoding.py
ledell: Yeah we can leave it here.  Just wanted to check.  Thanks.

PR #3194, opened 2019-01-03
[PUBDEV-6166] In External Backend, we need to explicitly pass timestamp from the Spark Driver node

jakubhava:


jakubhava:
@h2o-ops please test!

jakubhava:
Also manually tested on Sparkling Water side with required changes https://0xdata.atlassian.net/browse/SW-1107 and worked.

michalkurka:
Thank you @jakubhava.

Tagar:
Thank you @jakubhava for staying on top of this problem. 👍 


ps. Would be nice to have connectivity to backend cluster from SW covered by automated tests, if possible. It's a second upgrade attempt that we had to roll back ..

jakubhava:
@Tagar Good suggestion, we have a tests, but not exactly for this specific use-case. I will add it to my pipeline, however, Sparkling Water is also open source project, feel free to provide us with a PR :)

jakubhava:
Just for reference: Both internal and external backend are fully tested. What is tested in one mode is tested in second one and vice versa. The reason why the critical bug got it was because 1 pipeline stage was ignoring the external cluster option - 2 pysparkling yarn integration tests. This was introduced when I was porting code to run in docker machines instead on real machines. Fixing now, however just wanted to make sure we have the code fully covered.

This should not happen in the future again.

Tagar:
Thanks a lot @jakubhava !

PR #3199, opened 2019-01-03
PUBDEV-4680: support for blending in StackedEnsemble

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-4680

- changed StackedEnsemble logic to support 2 drivers:
  - `StackedEnsembleCVDriver`: for default SE requiring base models being trained using CV.
  - `StackedEnsembleBlendingDriver`: for SE not requiring preconditions on base models (can be trained with or without CV) outside being trained on a compatible training frame.
- added a `blending_frame` parameter to `StackedEnsembleModel`: if provided, this triggers blending. Note that `training_frame` is still required to compute `training_metrics` for the final SE model, and is also used to verify that the base models were trained with a compatible frame.
- added logic to cache base models predictions to avoid having to recompute them when creating multiple SEs from same base models: esp. useful for AutoML, but not only. By default, this is disabled and not exposed to client: need to figure out if we want to expose this.
- removed duplicate logic from `Metalearner` as I needed to add logic common to all metalearner algos.
- added `stacking_strategy` field to model output to identify models generated using cross_validation stacking from those using blending


seb-h2o:
@michalkurka : created https://0xdata.atlassian.net/browse/PUBDEV-6176 for the refactoring part.
I was (and still am) planning to create the migration document for those changes, but really don't want to put those changes aside (I invested some work to understand the scoring subtleties there and I'm sure the refactored version is actually better).
I'll put the following changes in a new PR (maybe though on top of the current one, will see...):
- move of `StackedEnsembleModel` class.
- changes in its `predictScoreImpl` methods.
- methods visibility changes allowed by this move.
- move of `StackedEnsemble.MetalearnerAlgorithm` to `Metalearner` for better encapsulation of metalearner logic.

angela0xdata:
Will this be also added to the Flow UI? Or does that require a separate PR and Jira ticket?

ledell:
@angela0xdata Yep, good call -- this should require a separate JIRA and PR for the Flow addition.  I created the JIRA [here](https://0xdata.atlassian.net/browse/PUBDEV-6180), currently assigned to @seb-h2o but could also be re-assigned to @michal-raska.

h2o-r/h2o-package/R/stackedensemble.R
```diff
@@ -47,6 +49,7 @@ h2o.stackedEnsemble <- function(x, y, training_frame,
                                 metalearner_fold_assignment = c("AUTO", "Random", "Modulo", "Stratified"),
                                 metalearner_fold_column = NULL,
                                 keep_levelone_frame = FALSE,
+                                blending_frame = NULL,
                                 seed = -1,
```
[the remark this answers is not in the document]
seb-h2o: makes sense, will do

h2o-algos/src/main/java/hex/StackedEnsembleModel.java
```diff
@@ -59,13 +61,18 @@ public StackedEnsembleModel(Key selfKey, StackedEnsembleParameters parms, Stacke
 
     // base_models is a list of base model keys to ensemble (must have been cross-validated)
     public Key<Model> _base_models[] = new Key[0];
-    // Should we keep the level-one frame of cv preds + repsonse col?
+    // Should we keep the level-one frame of cv preds + response col?
     public boolean _keep_levelone_frame = false;
+    // internal flag if we want to avoid having the base model predictions rescored multiple times, esp. for blending.
+    public boolean _keep_base_model_predictions = false;
```
seb-h2o: For now, I chose not to expose this field through REST API, so there's no possibility for user to enable predictions caching.
Main intent in mind is to use this for SEs in AutoML as we compute multiple SEs reusing same models.

h2o-algos/src/main/java/hex/StackedEnsembleModel.java
```diff
@@ -343,43 +347,50 @@ public void checkAndInheritModelProperties() {
...
-          throw new H2OIllegalArgumentException("Base models are inconsistent: they use different response columns.  Found: " + responseColumn + " and: " + aModel._parms._response_column + ".");
+          throw new H2OIllegalArgumentException("Base models are inconsistent: they use different response columns." +
+              " Found: "+responseColumn+"(StackedEnsemble) and "+aModel._parms._response_column+" (model "+k+").");
+        
+//        if (blending_mode && _parms._blending.equals(aModel._parms._train)) {
+//          throw new H2OIllegalArgumentException("Base model `"+k+"` was trained with the StackedEnsemble blending frame.");
+//        }
 
         // TODO: we currently require xval; loosen this iff we add a separate holdout dataset for the ensemble
-
```
seb-h2o: this block has only changes in indentation due to inclusion in `if (cv_required_on_base_model)` condition

h2o-algos/src/main/java/hex/ensemble/Metalearner.java
```diff
@@ -25,366 +22,222 @@
 import water.fvec.Frame;
 import water.util.Log;
 
-class Metalearner {
```
seb-h2o: DRYing this class:
- providing factory method to retrieve metalearner instance.
- each algo may custom logic in its own impl class: GLM and AUTO are currently the main ones requiring custom logic.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -82,7 +84,7 @@ public static void addModelPredictionsToLevelOneFrame(Model aModel, Frame aModel
     }
   }
 
-  private class StackedEnsembleDriver extends Driver {
```
seb-h2o: abstracted this driver to provide 2 implementations:
- one for cross-validation stacking strategy.
- one for blending.
StackedEnsembleModel itself is agnostic.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -143,108 +145,83 @@ private Frame prepareLevelOneFrame(String levelOneKey, Model[] baseModels, Frame
...
-          String[] multClassNames = predColNames.toArray(new String[0]);
-          baseModelPredictions.add(aPred.subframe(multClassNames));
-        }
-      }
-
-      Frame levelOne = prepareLevelOneFrame(levelOneKey, baseModels.toArray(new Model[0]), baseModelPredictions.toArray(new Frame[0]), actuals);
-
-      // remove baseModelPredictions frames and all the non-preds vecs from the DKV
-      for (Frame aPred : baseModelPredictions)
-        Frame.deleteTempFrameAndItsNonSharedVecs(aPred, levelOne);
```
seb-h2o: please note that those predictions have now a slightly longer lifetime: they're cleaned up by default only once the metalearner model is computed.

h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -104,7 +104,14 @@ def train(self, x=None, y=None, training_frame=None, offset_column=None, fold_co
...
         """
+        self._train(x=x, y=y, training_frame=training_frame, offset_column=offset_column, fold_column=fold_column,
+                    weights_column=weights_column, validation_frame=validation_frame, max_runtime_secs=max_runtime_secs, 
+                    ignored_columns=ignored_columns, model_id=model_id, verbose=verbose)
 
+
+    def _train(self, x=None, y=None, training_frame=None, offset_column=None, fold_column=None,
+              weights_column=None, validation_frame=None, max_runtime_secs=None, ignored_columns=None,
+              model_id=None, verbose=False, extend_parms_fn=None):
         assert_is_type(training_frame, None, H2OFrame)
```
seb-h2o: needed to "hide" train implementation to add a hook for subclasses (StackedEnsembleEstimator)

h2o-bindings/bin/gen_python.py
```diff
@@ -513,6 +513,22 @@ def levelone_frame_id(self):
...
             print("No levelone_frame_id for this model")         
+            
+        def stacking_strategy(self):
+            model = self._model_json["output"]
+            if "stacking_strategy" in model and model["stacking_strategy"] is not None:
+                return model["stacking_strategy"]
+            print("No stacking strategy for this model")  
+        
+        # Override train method to support blending 
+        def train(self, x=None, y=None, training_frame=None, blending_frame=None, **kwargs):
```
seb-h2o: please tell if any concern for usage of **kwargs here, as I didn't want to duplicate all params again to forward them to super impl.

h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -104,7 +104,14 @@ def train(self, x=None, y=None, training_frame=None, offset_column=None, fold_co
...
         :param bool verbose: Print scoring history to stdout. Defaults to False.
         """
+        self._train(x=x, y=y, training_frame=training_frame, offset_column=offset_column, fold_column=fold_column,
+                    weights_column=weights_column, validation_frame=validation_frame, max_runtime_secs=max_runtime_secs, 
+                    ignored_columns=ignored_columns, model_id=model_id, verbose=verbose)
 
+
+    def _train(self, x=None, y=None, training_frame=None, offset_column=None, fold_column=None,
+              weights_column=None, validation_frame=None, max_runtime_secs=None, ignored_columns=None,
+              model_id=None, verbose=False, extend_parms_fn=None):
```
seb-h2o: needed to "hide" train implementation to add a hook for subclasses (StackedEnsembleEstimator)

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -492,11 +519,33 @@ def standalone_test(test):
...
     h2o.log_and_echo("------------------------------------------------------------")
     h2o.log_and_echo("")
-    h2o.log_and_echo("STARTING TEST")
+    h2o.log_and_echo("STARTING TEST "+test.__name__)
     h2o.log_and_echo("")
     h2o.log_and_echo("------------------------------------------------------------")
     test()
 
+def run_tests(tests, run_in_isolation=True):
+    #flatten in case of nested tests/test suites
```
seb-h2o: simplistic "test suite" runner: a bit frustrated that when creating Py test files with multiple tests, those don't run is isolation by default.

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -492,11 +519,33 @@ def standalone_test(test):
 
     h2o.log_and_echo("------------------------------------------------------------")
     h2o.log_and_echo("")
-    h2o.log_and_echo("STARTING TEST")
+    h2o.log_and_echo("STARTING TEST "+test.__name__)
     h2o.log_and_echo("")
     h2o.log_and_echo("------------------------------------------------------------")
     test()
 
+def run_tests(tests, run_in_isolation=True):
```
seb-h2o: simplistic "test suite" runner: a bit frustrated that when creating Py test files with multiple tests, those don't run is isolation by default.

h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_binary.py
```diff
@@ -0,0 +1,115 @@
...
+    t = StackedEnsembleTest() if not blending else StackedEnsembleBlendingTest()
+
+    def test_saved_binary_model_produces_same_predictions_as_original():
+        ds = t.prepare_data()
+        base_models = t.train_base_models(ds)
+        se_model = t.train_stacked_ensemble(ds, base_models)
+        
+        #Predict in ensemble in Py client
+        preds_py = se_model.predict(ds.test)
+
```
seb-h2o: `pyunit_NOPASS_stackedensemble_binary.py` was failing due to usage of save model from old H2O-3 version.
Fixed it by saving binary model in tmp dir first.

h2o-r/tests/runitUtils/utilsR.R
```diff
@@ -376,6 +376,36 @@ function(testDesc, test) {
...
+          # do.call(test_name, list(), envir=envir)
+        })
+        if(length(warnings) > 0)
+            warning(paste("\n", warnings, "\n"))
+        if(length(errors) > 0)
+            stop(paste("Failing tests:\n", errors, "\n"), call. = FALSE)
+    }
+    doTest(suiteDesc, suiteTest)
+}
+
```
seb-h2o: same as for Python, added this small runner to run R tests functions in isolation and see clearly which test fails if any.

seb-h2o:
@ledell , @michalkurka : when you have time, please have a look, thx

h2o-algos/src/main/java/hex/StackedEnsembleModel.java
michalkurka: Is this meant for grid search over SE hyper-parameters?

h2o-algos/src/main/java/hex/StackedEnsembleModel.java
```diff
@@ -119,7 +114,14 @@ public void initMetalearnerParams(MetalearnerAlgorithm algo) {
...
     public Model _metalearner;
-    public Frame _levelone_frame_id;
+    public Frame _levelone_frame_id; //set only if StackedEnsembleParameters#_keep_levelone_frame=true
+    public StackingStrategy _stacking_strategy;
+    
+    //Set of base model predictions that have been cached in DKV to avoid scoring the same model multiple times,
+    // it is then the responsibility of the client code to delete those frames from DKV.
+    //This especially useful when building SE models incrementally (e.g. in AutoML).
+    //The Set is instantiated and filled only if StackedEnsembleParameters#_keep_base_model_predictions=true.
+    public NonBlockingHashSet<String> _base_model_predictions_keys;
```
michalkurka: Is there a good reason to store Keys as Strings? Why a HashSet? It would be good to be able to link the prediction frame back to the model - using an array and preserving the same order of models would enable it.

h2o-algos/src/main/java/hex/StackedEnsembleModel.java
seb-h2o: It's not what I had in mind: just that AutoML currently builds 2 SEs, one from all models, and the other one from bests of kind, therefore reusing same models in each SE, besides when running the AutoML instance multiple times, new SEs need to be built, including previous models again. This flag could prevent having to recompute predictions in those cases.

But if we would want to do grid search over SE hyper-params, this flag would provide similar benefits.

h2o-algos/src/main/java/hex/StackedEnsembleModel.java
seb-h2o: @michalkurka
Regarding the Hashset, I just wanted to ensure there's no duplicate: predictions are not only associated to a model, but to a model+frame combination; in CV-stacking mode, this frame can only be the validation frame, but in blending mode, it could be the blending frame or the validation frame. 
For now, the purpose of those keys is only to track them for deletion later. The associated predictions are not exposed through REST API (validation frame predictions also were not until now): if we want to expose them, then we can think about a way, but I don't think that simply sorting by model order would be convenient enough anyway. 
From backend, it is currently as simple to do a reverse lookup: given list of models, a prediction key and a frame, build model prediction keys for the model+frame until it matches the given key... not extremely efficient, but as there's no use-case currently anyway...

Regarding the String key: In a early implementation, I started with a Key but I had some issue with serialization then, so I switched to String... will try again, but it may be related with the use of a NonBlockingHashSet... this Iced mechanism is actually seriously limiting our usage of data structures.

h2o-algos/src/main/java/hex/StackedEnsembleModel.java
michalkurka: Does AutoML delete the cached predictions? I couldn't find it in the code. I only see it deleted when the SE model is deleted, the base predictions can be quite big I think to keep around without the user knowing. It is dense numerical data that do not compress well.

michalkurka:
Looks good to me, good refactoring also - my only suggestion is to think about the lifecycle of the cached predictions. IMHO it could be made more straightforward how the predictions are re-used and when they get deleted.

seb-h2o:
@michalkurka 
AutoML integration is a separate ticket: for now the predictions are always deleted by default immediately once the model has been built.
For AutoML, there are 2 use-cases:
- single AutoML run: I'll probably systematically activate the predictions cache for the first SE (ideally should be the Best of each model SE, which may require switching order of execution), then run the second SE benefiting from the cache, and finally clearing the cache at the end of the run.
- multiple runs: at some point we will need to modify the current API if we want to optimize AutoML (and avoid some issues we're having currently) when user is re-running an AutoML instance to obtain more models. In this case, we would be able to explicitly activate the  predictions cache. Outside this use-case, I'm not sure I want to expose this caching behaviour to user.

ledell:
@seb-h2o Looks good, and the performance makes sense to me.

I ran an example here, will post the results below:
```
# 5-fold CV ensemble
[1] "Best Base-learner Test AUC:  0.76979821502548"
[1] "Ensemble Test AUC:  0.773548270328318"

# Using a blending frame 75/25
[1] "Best Base-learner Test AUC:  0.762783080146884"
[1] "Ensemble Test AUC:  0.769013357143719"
```

Code:
```
library(h2o)
h2o.init()

y <- "response"
x <- setdiff(names(train), y)

test[,y] <- as.factor(test[,y])

# Number of CV folds (to generate level-one data for stacking)
nfolds <- 5

# Train & Cross-validate a GBM
my_gbm <- h2o.gbm(x = x,
                  y = y,
                  training_frame = train,
                  distribution = "bernoulli",
                  ntrees = 10,
                  max_depth = 3,
                  min_rows = 2,
                  learn_rate = 0.2,
                  nfolds = nfolds,
                  fold_assignment = "Modulo",
                  keep_cross_validation_predictions = TRUE,
                  seed = 1)

# Train & Cross-validate a RF
my_rf <- h2o.randomForest(x = x,
                          y = y,
                          training_frame = train,
                          ntrees = 50,
                          nfolds = nfolds,
                          fold_assignment = "Modulo",
                          keep_cross_validation_predictions = TRUE,
                          seed = 1)

# Train a stacked ensemble using the GBM and RF above
ensemble <- h2o.stackedEnsemble(x = x,
                                y = y,
                                training_frame = train,
                                model_id = "my_ensemble_binomial",
                                base_models = list(my_gbm, my_rf))

# Eval ensemble performance on a test set
perf <- h2o.performance(ensemble, newdata = test)

# Compare to base learner performance on the test set
perf_gbm_test <- h2o.performance(my_gbm, newdata = test)
perf_rf_test <- h2o.performance(my_rf, newdata = test)
baselearner_best_auc_test <- max(h2o.auc(perf_gbm_test), h2o.auc(perf_rf_test))
ensemble_auc_test <- h2o.auc(perf)
print(sprintf("Best Base-learner Test AUC:  %s", baselearner_best_auc_test))
print(sprintf("Ensemble Test AUC:  %s", ensemble_auc_test))


# Now create a blending frame and compare results

ss <- h2o.splitFrame(train, seed = 1)
train <- ss[[1]]
blend <- ss[[2]]

# Train & Cross-validate a GBM
my_gbm <- h2o.gbm(x = x,
                  y = y,
                  training_frame = train,
                  distribution = "bernoulli",
                  ntrees = 10,
                  max_depth = 3,
                  min_rows = 2,
                  learn_rate = 0.2,
                  nfolds = nfolds,
                  fold_assignment = "Modulo",
                  keep_cross_validation_predictions = TRUE,
                  seed = 1)

# Train & Cross-validate a RF
my_rf <- h2o.randomForest(x = x,
                          y = y,
                          training_frame = train,
                          ntrees = 50,
                          nfolds = nfolds,
                          fold_assignment = "Modulo",
                          keep_cross_validation_predictions = TRUE,
                          seed = 1)

# Train a stacked ensemble using the GBM and RF above
ensemble <- h2o.stackedEnsemble(x = x,
                                y = y,
                                training_frame = train,
                                blending_frame = blend,
                                base_models = list(my_gbm, my_rf))

# Eval ensemble performance on a test set
perf <- h2o.performance(ensemble, newdata = test)

# Compare to base learner performance on the test set
perf_gbm_test <- h2o.performance(my_gbm, newdata = test)
perf_rf_test <- h2o.performance(my_rf, newdata = test)
baselearner_best_auc_test <- max(h2o.auc(perf_gbm_test), h2o.auc(perf_rf_test))
ensemble_auc_test <- h2o.auc(perf)
print(sprintf("Best Base-learner Test AUC:  %s", baselearner_best_auc_test))
print(sprintf("Ensemble Test AUC:  %s", ensemble_auc_test))
```

ledell:
@seb-h2o The results for the blended ensemble are exactly the same whether the base models use 5-fold CV or not.  If you comment out the `nfolds` arg, and re-run the code in the second half, you will see that it doesn't change the ensemble performance.  The final model (trained on the whole training set) which is used in prediction is the same regardless of whether CV was performed or not, therefore the performance of the base models is identical.  

I was trying to replicate a real-world scenario -- the user only has so much training data, and so they could choose to make the most use of it by doing CV regular stacked ensemble, or they can train models more quickly by doing a blended ensemble (which will always result in a degradation of model performance compared to a regular stacked ensemble).

Just for curiosity, here are the results if we pretend we have unlimited training data and we compare CV stacked ensemble to the blended ensemble, where the base models have access to the exact same amount of data:

```
# 5-fold CV ensemble
[1] "Best Base-learner Test AUC:  0.76979821502548"
[1] "Ensemble Test AUC:  0.773548270328318"

# Blended ensemble: Using a blending frame 75/25
[1] "Best Base-learner Test AUC:  0.762783080146884"
[1] "Ensemble Test AUC:  0.769013357143719"

# 5-fold CV ensemble: Using training frame of 75%
[1] "Best Base-learner Test AUC:  0.762783080146884"
[1] "Ensemble Test AUC:  0.769149301575427"
```

h2o-algos/src/main/java/hex/StackedEnsembleModel.java
seb-h2o: @michalkurka : now that client mode tests are failing, I know why it used to be String instead of key:
```
Caused by: java.io.InvalidClassException: water.Key; no valid constructor
	at java.io.ObjectStreamClass$ExceptionInfo.newInvalidClassException(ObjectStreamClass.java:169)
	at java.io.ObjectStreamClass.checkDeserialize(ObjectStreamClass.java:874)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2043)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1573)
	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:431)
	at water.nbhm.NonBlockingHashMap.readObject(NonBlockingHashMap.java:1421)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at java.io.ObjectStreamClass.invokeReadObject(ObjectStreamClass.java:1170)
	at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2178)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2069)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1573)
	at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2287)
	at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2211)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2069)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1573)
	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:431)
	at water.AutoBuffer.javaSerializeReadPojo(AutoBuffer.java:1514)
```

I believe this is because the NonBlockingHashset is only Serializable... so, going to replace this with an array tmrw...

seb-h2o:
@ledell oh right! I forgot that main models were actually trained against all training set and that cv models were used for validation only when we don't plan to build a stacking SE on top of them... then comparison definitely makes sense, sorry for that...
note that on some unit tests, blending is sometimes better than CV-stacking indeed, so that I had to make assertions that were just logging warnings in this case as I didn't want the test give the impression that we have always CV > blending.

ledell:
@seb-h2o 
> note that on some unit tests, blending is sometimes better than CV-stacking indeed,
That's interesting, but I suppose that can happen on very small datasets.

ledell:
@seb-h2o This looks good, is it ready to merge?

PR #3201, opened 2019-01-04
PUBDEV-6170: Add a technote explaining what to do when Jetty is missing

michalkurka:


michalkurka:
Please also see https://0xdata.atlassian.net/browse/TN-13

PR #3205, opened 2019-01-07
PUDEV-6162 - S3A bucket with credentials

Pscheidl:
[JIRA PUBDEV-6162](https://0xdata.atlassian.net/browse/PUBDEV-6162)

Make sure to visit [Support portal ticket 93348](https://support.h2o.ai/helpdesk/tickets/93348) for more details.

### DIFF
--- a/h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
+++ b/h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
@@ -3,14 +3,8 @@ package water.persist;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.*;
 
-import java.io.EOFException;
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.net.SocketTimeoutException;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.io.*;
+import java.net.*;
 import java.util.ArrayList;
 import java.util.concurrent.Callable;
 import java.util.regex.Matcher;
@@ -36,6 +30,8 @@ public final class PersistHdfs extends Persist {
   public static final Configuration CONF;
   /** Root path of HDFS */
   private final Path _iceRoot;
+  
+  private static final String DEFAULT_CHARSET = "utf-8";
 
   /**
    * Filter out hidden files/directories (dot files, eg.: .crc).
@@ -286,24 +282,39 @@ public final class PersistHdfs extends Persist {
   }
 
   public static void addFolder(Path p, ArrayList<String> keys,ArrayList<String> failed) throws IOException {
-    FileSystem fs = FileSystem.get(p.toUri(), PersistHdfs.CONF);
+    FileSystem fs = null;
+    final AmazonCredentials amazonCredentials = extractAmazonCredentials(p.toString());
+    try {
+      fs = FileSystem.get(new URI(p.toString()), PersistHdfs.CONF);
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException("Given S3 URI is malformed. Please check special characters and encoding.", e);
+    }
     if(!fs.exists(p)){
       failed.add("Path does not exist: '" + p.toString() + "'");
       return;
     }
-    addFolder(fs, p, keys, failed);
+    addFolder(fs, p, keys, failed, amazonCredentials);
   }
 
-  private static void addFolder(FileSystem fs, Path p, ArrayList<String> keys, ArrayList<String> failed) {
+  private static void addFolder(FileSystem fs, Path p, ArrayList<String> keys, ArrayList<String> failed, final AmazonCredentials amazonCredentials) {
     if (fs == null) return;
     Futures futures = new Futures();
     try {
       for( FileStatus file : fs.listStatus(p, HIDDEN_FILE_FILTER) ) {
         Path pfs = file.getPath();
-        if(file.isDirectory()) {
-          addFolder(fs, pfs, keys, failed);
+        if(file.isDirectory() && !isDuplicatedRootFolderPresent(file, p)) {
+          addFolder(fs, pfs, keys, failed, amazonCredentials);
         } else if (file.getLen() > 0){
-          Key k = HDFSFileVec.make(pfs.toString(), file.getLen(), futures);
+          Key k;
+          if(amazonCredentials != null) {
+            URI uri = pfs.toUri();
+            //Decode the concatenated credentials in case those contain encoded chars. URI constructor is going to encode
+            // them once again
+            uri = new URI(uri.getScheme(), URLDecoder.decode(amazonCredentials.concatenated, DEFAULT_CHARSET), uri.getHost(), uri.getPort(), uri.getPath(),uri.getQuery(), uri.getFragment());
+            k = HDFSFileVec.make(uri.toString(), file.getLen(), futures);
+          } else {
+            k = HDFSFileVec.make(pfs.toString(), file.getLen(), futures);
+          }
           keys.add(k.toString());
           Log.debug("PersistHdfs: DKV.put(" + k + ")");
         }
@@ -315,34 +326,34 @@ public final class PersistHdfs extends Persist {
       futures.blockForPending();
     }
   }
+  
+  private static boolean isDuplicatedRootFolderPresent(final FileStatus fileStatus, final Path referencePath){
+    return fileStatus.getPath().getName().equals(referencePath.getName());
+  }
 
   @Override
   public Key uriToKey(URI uri) throws IOException {
     assert "hdfs".equals(uri.getScheme()) || "s3".equals(uri.getScheme())
             || "s3n".equals(uri.getScheme()) || "s3a".equals(uri.getScheme()) : "Expected hdfs, s3 s3n, or s3a scheme, but uri is " + uri;
 
+      final String uriWithCredentials = encodeCredentialsIfPresent(uri.toString());
+      if(uriWithCredentials != null) {
+        uri = URI.create(uriWithCredentials);
+      }
+
     FileSystem fs = FileSystem.get(uri, PersistHdfs.CONF);
     FileStatus[] fstatus = fs.listStatus(new Path(uri));
     assert fstatus.length == 1 : "Expected uri to single file, but uri is " + uri;
-
-    return HDFSFileVec.make(fstatus[0].getPath().toString(), fstatus[0].getLen());
+    // The FileSystem API may return the path with decoded credentials - thus it is required to re-encode.
+    return HDFSFileVec.make(encodeCredentialsIfPresent(fstatus[0].getPath().toString()), fstatus[0].getLen());
   }
-
-  public static FileSystem getFS(String path) throws IOException {
-    try {
-      return getFS(new URI(path));
-    } catch (URISyntaxException e) {
-      throw new RuntimeException(e);
-    }
-  }
-  public static FileSystem getFS(URI uri) throws IOException {
-    return FileSystem.get(uri, PersistHdfs.CONF);
-  }
-
+  
+  
   // Is there a bucket name without a trailing "/" ?
+  private static final Pattern BUCKET_ONLY_PATTERN = Pattern.compile("s3n://.@{0,1}[^/]*");
   private boolean isBareS3NBucketWithoutTrailingSlash(String s) {
     String s2 = s.toLowerCase();
-    Matcher m = Pattern.compile("s3n://[^/]*").matcher(s2);
+    Matcher m = BUCKET_ONLY_PATTERN.matcher(s2);
     return m.matches();
   }
 //  // We don't handle HDFS style S3 storage, just native storage.  But all users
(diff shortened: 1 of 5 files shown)


Pscheidl:
Let's test if credentials present in the environment variables kick in. Those are already used for the very same `test.0xdata.com` bucket by the environment and should be visible.

Test unable to be ran locally right now. Could be optionally moved to nightly.

Pscheidl:
Things to check:
Accessibility of public buckets with out credential provider chain.

Pscheidl:
I've tried Flow manually as well.
snimek z 2019-01-08 15-09-31

Pscheidl:
## Summary of changes done

1. Updated AWS SDK to latest fix release.
1. Using new method `doesBucketExistv2` which correctly fails under certain conditions (see method documentation).
1. Lock mechanism in `getClient` method fixed. Before checking if there is no instance, one has to be sure there is noone other holding the lock, trying to create it. Before, the check was done before the attempt to hold the lock. Once the lock is obtained and the check passes, it is freed immediately.
1. Replaced old deprecated Client API with new builder-style API. Configured it to mimic the old behavior. Old configuration methods preserved with the business logic inside, only refactored a little bit.
1. **The actual fix** - support for credentials (access key ID & secret access key pair) provided in the URL added. For example `s3://AKIIJPD44LO5HMR7NZEQ:7k3eWIQkoFwShOPEOGYP4ThmNKBqKM9zO2QuZK84@pavel-h2o-test/iris.csv` (not real credentials in the URL). The support is added for GET operations only (s3a style).  In order to support credentials in URL, I've modified `decodePath` method to extract these credentials from the URL obtained from the user (most likely by calling importPath). This method no longer returns array of strings, but new object `S3Path` which has all the parts of `s3://` url properly named. Replaced all the places where the plain string array was used with this object. Much more readable, as now the array is not of size 2 (bucket name, file/item name), but the array's size may climb up to 4 members (bucket name, file/item name, secret key ID, secret access key). Method `getClient` then uses these credentials to construct the client. If those credentials are provided, a new `AwsCredentialsProvider` with these in-url credentials is inserted into the chain, otherwise not the list of providers remains the same. Because specifying users credentials in URL might be a mean for the users to load from several different buckets (with different credentials) - the method `getClient` checks if the credentials change (or are not present anymore) and constructs new client on demand. To compare the credentials used, those are not kept in memory by H2O directly - a hash (SHA-256) is used instead.
1. Added a test with credentials in URL. The test has a JUNIT assumption at the beginning - if credentials are not provided by the environment, the test is marked as "assumption failed".

h2o-persist-s3/src/test/java/water/persist/PersistS3Test.java
```diff
@@ -34,17 +38,67 @@
   }
   @Test
   public void testS3Import() throws Exception {
+    Key k = null, k2 = null;
```
Pscheidl: This test remained the same, I only took the opportunity to do explicit cleanup.

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -64,13 +117,23 @@ public static AmazonS3 getClient() {
...
+    private static List<AWSCredentialsProvider> getProviders(final String accessKeyId, final String accessSecretKey) {
+      final List<AWSCredentialsProvider> providers = new ArrayList<>();
+      if (accessKeyId != null && accessSecretKey != null) {
+        providers.add(new AWSStaticCredentialsProvider(new BasicAWSCredentials(accessKeyId, accessSecretKey)));
+      }
+      providers.add(new H2OArgCredentialsProvider());
+      providers.add(InstanceProfileCredentialsProvider.getInstance());
+      providers.add(new EnvironmentVariableCredentialsProvider());
+      providers.add(new SystemPropertiesCredentialsProvider());
+      providers.add(new SystemPropertiesCredentialsProvider());
```
michal-raska: it's the same as line above, copy error?

Pscheidl: Oh yes, a copy error ! Thank you ! This needs to be resolved !

Pscheidl:
https://issues.apache.org/jira/browse/HADOOP-14114

h2o-persist-s3/build.gradle
```diff
@@ -1,16 +1,9 @@
...
-//dependencyManagement {
-//  imports {
-//    mavenBom 'com.amazonaws:aws-java-sdk-bom:1.10.47'
-//  }
-//}
 
 dependencies {
   compile project(":h2o-core")
-  compile 'com.amazonaws:aws-java-sdk-s3:1.10.77'
+  compile 'com.amazonaws:aws-java-sdk-s3:1.11.478'
```
jakubhava: I would be careful about updating the versions. Hadoop & AWS work together just on specific combinations.

jakubhava: however if the tests pass, then I guess we are good to go

michalkurka: This is a good point, the versions are selected to be compatible with the versions of the Hadoop libraries we bundle in H2O (currently 2.8.x).

It would be good to make sure they are in sync. Even if the tests pass.

Pscheidl:
The difference between Hadoop client version 2.7.4 used in H2O version 3.18.0.10 is in PersistHDFS class, method `calcTypeheadMatches`. This specific piece of code:

```java
      FileSystem fs = FileSystem.get(p.toUri(), conf);
      for( FileStatus file : fs.listStatus(expand) ) {
        Path fp = file.getPath();
        if( fp.toString().startsWith(p.toString()) ) {
          array.add(fp.toString());
        }
        if( array.size() == limit) break;
      }
```

specifically `fp.toString()` behaves differently. This is the file path of the retrieved file. In the newer version of the client library we use in newer versions of H2O (reporter used `3.22.0.2`), this filepath no longer contains user credentials, but just the filename. Therefore, this comparison fails.

This bug must be fixed, however I am not 100% sure this is the end of this issue. The reporter also stated `parseSetup` phase was not working for him after using `core-site.xml` configuration. Should be investigated.

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -1,5 +1,6 @@
 package water.persist;
 
+import com.sun.jndi.toolkit.url.Uri;
```
michalkurka: Probably unused? I would hope so ;)

Pscheidl: Of course, unused. Thanks.

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -377,26 +378,37 @@ private boolean isBareS3NBucketWithoutTrailingSlash(String s) {
...
+      final URI uri = p.toUri();
+      final String comparedPath = p.toString();
       if( !filter.endsWith("/") ) expand = p.getParent();
-      FileSystem fs = FileSystem.get(p.toUri(), conf);
+      FileSystem fs = FileSystem.get(uri, conf);
       for( FileStatus file : fs.listStatus(expand) ) {
-        Path fp = file.getPath();
-        if( fp.toString().startsWith(p.toString()) ) {
-          array.add(fp.toString());
+        final URI s3File = file.getPath().toUri();
```
Pscheidl: This was the main problem. The returned file.getPath() only contained the protocol, bucket name and file name, e.g. `s3:/bucketname/iris.csv`. The credentials got lost. So those are re-inserted below. 

The re-insertion serves two purpose:
1. Correct URI is shown to the User (e.g. in Flow)
1. Later, when those are imported, the full URL is still required for the fetching process.

michalkurka:
@Pscheidl I have a bunch of questions:

- modified function calcTypeaheadMatches is only used from Flow and it seems not tested by this PR? Is that correct?
- does the credentials work from R/Python? it would be good to have a test proving it
- what about a recursive import of objects from S3? will it work or will the credentials also get stripped? addFolder is similar to calcTypeaheadMatches - which makes me thing if it needs fix too

Pscheidl:
@michalkurka 

1. There is a test `public void testS3CalcTypeaheadMatches() throws Exception {`
1. Will add Python & R tests
1. Yes, this one needs fixing too. But importing from whole bucket via `CalcTypeaheadMatches` works.

Pscheidl:
We also need to encode the credentials before sending them.

Pscheidl:
Original @michalkurka 's questions ^^^ and final answers to them
- [ ] _modified function calcTypeaheadMatches is only used from Flow and it seems not tested by this PR? Is that correct?_ No, it is tested and was tested almost from the very beginning. There were other useful tests added however.
- [ ] - _does the credentials work from R/Python? it would be good to have a test proving it_ - Python and R both call the same functions, so a Python `importFiles` test is enough. What we need to test is both a single file retrieval and a whole folder fetching inside one bucket, because the addFolder method might get called recursively in that case. These tests have been added as well (Python).
- [ ] - _what about a recursive import of objects from S3? will it work or will the credentials also get stripped? addFolder is similar to calcTypeaheadMatches - which makes me thing if it needs fix too_ - Answer in previous point. I wrote a Python test to prove it works recursively for two files in folder inside one bucket. These files are iris dataets named `iris.csv` and `iris2.csv`. First, the test tries to import one of the files directly. Then import the whole folder. In the case of whole folder import operation, the resulting frame must have twice as much rows as the first one (fetching two identical files, only names are different). This also tests the recursive calls do work.

Pscheidl:
Also checked the secret credentials do not leak into our key names.

Pscheidl:
Besides `Python` "integration" test, I added JUNIT test for `importFiles` functionality. There are two tests in fact - one imports a single file from a bucket's folder, one imports the whole folder (2 files). This JUNIT test is more fine-grained. Directly checks there are only two keys added (and later cleaned up) and there are no failures in the process. Also makes sure the `importFiles` method is called in a constant manner.

Pscheidl:
There is one issue in tests that seems to be resolved by their client, but it looks like it is not resolved everywhere: `java.lang.NullPointerException: null uri host. This can be caused by unencoded / in the password string`

https://issues.apache.org/jira/browse/HADOOP-3733

We're of course doing the URL encoding ourselves.

Pscheidl:
TODO: Fix `isBareS3NBucketWithoutTrailingSlash` - the bucket name detection is just horribly wrong + the slash at the end should also be appended for folders, not just for whole buckets.

Pscheidl:
TODO: Fix isBareS3NBucketWithoutTrailingSlash - the bucket name detection is far from perfect + the slash at the end should also be appended for folders, not just for whole buckets.

Pscheidl:
JUnit test failures unrelated ! Detected one last issue with URI encoding in addFolder. Detection of credentials should be done via regexp, not with path.toUri() method.

Pscheidl:
Secret keys stripping marked as `Won't fix`, because: "too few people complain about it". 
https://issues.apache.org/jira/browse/HADOOP-14439

Pscheidl:
Test failures unrelated again:
```
Java 8 JUnit / hex.deeplearning.DeepLearningGradientCheck.gradientCheck
```

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -315,34 +326,34 @@ private static void addFolder(FileSystem fs, Path p, ArrayList<String> keys, Arr
...
-    }
-  }
-  public static FileSystem getFS(URI uri) throws IOException {
-    return FileSystem.get(uri, PersistHdfs.CONF);
-  }
-
+  
+  
   // Is there a bucket name without a trailing "/" ?
+  private static final Pattern BUCKET_ONLY_PATTERN = Pattern.compile("s3n://.@{0,1}[^/]*");
```
Pscheidl: Historically limited to s3n. These APIs hide so many traps I'd not be surprised if s3a and s3n behaved differently and this method was really meant to check just s3n.

Pscheidl:
Please review.

Potential problem: These credentials might get into our logs :) I redacted them in `PersistHDFS` code, but these may still leak. And not only in our code, Hadoop client is probably also able to log them. User's responsibility ? 

Please note Hadoop people talk about removing this option in 3.3 (one of the links ^^^). There are still warnings printed about how unsafe this is. Main reason of the issue reporter was to import  files from multiple buckets with single configuration. However, Hadoop 2.8.x should support multiple bucket configurations. I made a JIRA to introduce tests for that : https://0xdata.atlassian.net/browse/PUBDEV-6193

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -286,24 +282,39 @@ private static void ignoreAndWait(final Exception e, boolean printException) {
   }
 
   public static void addFolder(Path p, ArrayList<String> keys,ArrayList<String> failed) throws IOException {
-    FileSystem fs = FileSystem.get(p.toUri(), PersistHdfs.CONF);
+    FileSystem fs = null;
+    final AmazonCredentials amazonCredentials = extractAmazonCredentials(p.toString());
+    try {
+      fs = FileSystem.get(new URI(p.toString()), PersistHdfs.CONF);
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException("Given S3 URI is malformed. Please check special characters and encoding.", e);
```
michalkurka: This class deals with all Hadoop filesystems, not just S3. Please also include the path/(=user input) in the message.

PR #3206, opened 2019-01-07
ITA-362: move test lookups into Makefile

pkozelka:
The goal is to maintain the same Makefile behavior in Jenkins and local execution.

This PR solves the problem for all affected targets - many R and python unit tests.
In addition to this, another problem is fixed: some exclusions were implemented by simply deleting certain tests from the filesystem, which is not such a problem on jenkins agents, but would be very dangerous when executed on developer's machine. Instead, cummulative exclussion is used now, see new helper targets `init-tests`, `lookup-*-tests` and usage of `.tests-*.txt` files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -141,3 +141,5 @@ h2o-docs/src/booklets/source/PythonBooklet.tui
 *.gz(busy)
 
 h2o-web/package-lock.json
+
+h2o-*/tests/.lookup.txt
--- a/scripts/jenkins/Makefile.jenkins
+++ b/scripts/jenkins/Makefile.jenkins
@@ -25,21 +25,36 @@ warmup-caches:
 	./gradlew cpLibs $$ADDITIONAL_GRADLE_OPTS
 
 test-package-py:
-	cd h2o-py/tests/ && rm -f -r testdir_hdfs
 	zip -q -r test-package-py.zip h2o-py/tests/ h2o-py/demos/ \
 	h2o-docs/src/booklets/v2_2015/source h2o-py/build/dist/*.whl \
 	h2o-genmodel/build/libs/h2o-genmodel.jar h2o-assemblies/genmodel/build/libs/genmodel.jar scripts/run.py \
 	$(JENKINS_MAKEFILE_PATH) tests/pyunit*List h2o-py/scripts/h2o-py-test-setup.py
 
 test-package-r:
-	cd h2o-r/tests/ && rm -f -r testdir_hdfs
 	zip -q -r test-package-r.zip h2o-r h2o-docs/src/booklets/v2_2015/source \
 		h2o-assemblies/genmodel/build/libs/genmodel.jar scripts/run.py \
 		$(JENKINS_MAKEFILE_PATH) tests/runitSmokeTestList tests/runitAutoMLList \
 		scripts/validate_r_cmd_check_output.R scripts/validate_r_cmd_check_output.py h2o-3-DESCRIPTION
 
+init-lookup:
+	rm -f h2o-{r,py}/tests/.lookup.txt
+
+lookup-automl-tests: init-lookup
+	mkdir -p h2o-{r,py}/tests
+	find h2o-r/tests -name '*automl*.R' -exec basename {} \; >>h2o-r/tests/.lookup.txt
+	find h2o-py/tests -name '*automl*.py' -exec basename {} \; >>h2o-py/tests/.lookup.txt
+
+lookup-hdfs-tests: init-lookup
+	mkdir -p h2o-{r,py}/tests/testdir_hdfs
+	find h2o-r/tests/testdir_hdfs -name '*.R' -exec basename {} \; >>h2o-r/tests/.lookup.txt
+	find h2o-py/tests/testdir_hdfs -name '*.py' -exec basename {} \; >>h2o-py/tests/.lookup.txt
+
+lookup-demos-tests: init-lookup
+	mkdir -p h2o-{r,py}/tests/testdir_demos
+	find h2o-r/tests/testdir_demos -name '*.R' -exec basename {} \; >>h2o-r/tests/.lookup.txt
+	find h2o-py/tests/testdir_demos -name '*.py' -exec basename {} \; >>h2o-py/tests/.lookup.txt
+
 test-py-smoke:
-	cd h2o-py/tests/ && rm -f -r testdir_hdfs
 	export NO_GCE_CHECK=True && cd h2o-py/tests/ && ../../scripts/run.py --wipeall --geterrs --testlist ../../tests/pyunitSmokeTestList --numclouds 6 --jvm.xmx 3g
 
 test-py-init:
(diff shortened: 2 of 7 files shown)


pkozelka:
The problem was reported for make target `test-r-small-client-mode`, but applies to many other R and python test targets; I did the fix on all of them

scripts/jenkins/Makefile.jenkins
```diff
@@ -38,8 +38,24 @@ test-package-r:
 		$(JENKINS_MAKEFILE_PATH) tests/runitSmokeTestList tests/runitAutoMLList \
 		scripts/validate_r_cmd_check_output.R scripts/validate_r_cmd_check_output.py h2o-3-DESCRIPTION
 
+init-tests:
+	>.tests-r.txt
+	>.tests-py.txt
+
+lookup-automl-tests:
+	find h2o-r/tests -name '*automl*.R' -printf '%f\n' >>.tests-r.txt
```
michalkurka: -printf doesn't work on mac, the makefile should be posix-compliant

scripts/jenkins/Makefile.jenkins
```diff
@@ -38,8 +38,24 @@ test-package-r:
 		$(JENKINS_MAKEFILE_PATH) tests/runitSmokeTestList tests/runitAutoMLList \
 		scripts/validate_r_cmd_check_output.R scripts/validate_r_cmd_check_output.py h2o-3-DESCRIPTION
 
+init-tests:
+	>.tests-r.txt
```
michalkurka: why do we need to append to the tests? would it be better to just write the test file once? without appending?

pkozelka: this makes combining multiple sets of exclusions (or potentially inclusions) possible - for example, see targets `test-r-small` or `test-r-small-client-mode` which have multiple lookup dependencies

michalkurka: HDFS testes are not run as a part of test-r-small or test-r-small-client-mode.

pkozelka: sure, that's why they are excluded - the lookup is universal (for both include and exclude) and then the list is used with parameter, either `--testlist` or `--excludelist`

pkozelka: I know it's not ideal, but should be simpler than before... I will add some brief explanatory comments

pkozelka:
@michalkurka is it better to rebase this to release branch?

.gitignore
```diff
@@ -141,3 +141,6 @@ h2o-docs/src/booklets/source/PythonBooklet.tui
 *.gz(busy)
 
 h2o-web/package-lock.json
+
+.tests-r.txt
```
michalkurka: If we put these file in the results directory they will be automatically ignored. Storing them in the results directory would have one more benefit - they would be archived by the jenkins job and provide another info when debugging.

scripts/jenkins/groovy/buildH2O3.groovy
```diff
@@ -26,15 +26,13 @@ def call(final pipelineContext) {
                             makefilePath = pipelineContext.getBuildConfig().MAKEFILE_PATH
                             activatePythonEnv = true
                         }
-                        findAutoMLTests(pipelineContext, pipelineContext.getBuildConfig().COMPONENT_PY)
```
michalkurka: Great!

michalkurka:
> @michalkurka is it better to rebase this to release branch?

It is usually better to not to maintain 2 versions of the jenkins files. I think it is safe to include in rel-xu as well. But please do check carefully that the executed tests are the same as before (regardless of what is the target branch).

.gitignore
pkozelka: good point, I will change it

pkozelka:
ok I rebased it against `rel-xu`.
I am not merging this yet, because I want to check the sets of executed tests really carefully.
The best way I can think of is, gather all `.../results/summary.txt` files from jenkins tests - but right now, the results are disappearing which is probably related to [ITA-363](https://0xdata.atlassian.net/browse/ITA-363) - so I will come back here after fixing that one.

.gitignore
pkozelka: ... well actually, the `results` directory is being wiped at the beginning of `run.py` which makes it unsuitable for storing configuration.
Instead, I changed the location to be `h2o-*/tests/.lookup.txt` and added that into set of artifacts archives by jenkins. So it will appear on UI, just from slightly different location.

michalkurka: right!

pkozelka:
With a little scripting, I extracted numbers of configured, scheduled, and actually performed tests per testsuite. I did this for `rel-xia` build 21 and `PR-3206` build 21, in situation where that PR was rebased on `rel-xia` branch.
I made the script to produce simple TSV tables (scheduled cnt, testsuite name) for easy textual comparison, and this is the result:

```
$ diff -y -W70 rel-xu-21-HC.tsv PR-3206-21-HC.tsv 
1	flow-headless			1	flow-headless
1	flow-headless-medium	  <
1	flow-headless-small	  <
1	flow-headless-smoke		1	flow-headless-smoke
1	info-check			1	info-check
2	java-10-junit		  <
2	java-11-junit		  <
1	py2.7-booklets			1	py2.7-booklets
23	py2.7-demos			23	py2.7-demos
48	py2.7-smoke			48	py2.7-smoke
20	py3.5-single-node		20	py3.5-single-node
616	py3.5-small			616	py3.5-small
14	py3.5-small-automl		14	py3.5-small-automl
79	py3.6-medium-large		79	py3.6-medium-large
5	py3.6-test-demos		5	py3.6-test-demos
1	r3.4-booklets			1	r3.4-booklets
1	r3.4-datatable		  <
11	r3.4-demos-medium-larg		11	r3.4-demos-medium-larg
28	r3.4-demos-small		28	r3.4-demos-small
9	r3.4-demos-small	  <
69	r3.4-medium-large		69	r3.4-medium-large
660	r3.4-small			660	r3.4-small
10	r3.4-small-automl		10	r3.4-small-automl
660	r3.4-small-client-mode		660	r3.4-small-client-mode
10	r3.4-small-client-mode		10	r3.4-small-client-mode
31	r3.4-smoke			31	r3.4-smoke
```

This comparison shows that:
- some testsuites present on release build are not executed on this PR build at all, which is OK
- others indicate exactly the same number of scheduled tests which is OK
My conclusion therefore is that this PR does not change which tests are selected for running by the changed lookup mechanism. This comparison is a bit weak, because it works with number of tests per testsuite, not with the explicit tests, but it is very unlikely that such mistake would not impact number of tests.

The reason why I could not use comparison of individual tests is, that it is very rare that all tests in all testsuites are even executed; that is, execution of a testsuite is often aborted and the remaining tests are not listed in the `summary.txt` file, although they were scheduled (I think changing this behavior would be beneficial - can be listed just as "skipped").

Further analysis of that same data has shown that there is a high duplicity in executed tests. While it might be useful to execute certain tests in various environments, I am not sure if it is always the case - for instance, test `runit_deeplearning_iris_basic.R` executed 4 times in `rel-xu` build 21:

```
$ grep runit_deeplearning_iris_basic.R rel-xu-21.tsv
info-check	runit_deeplearning_iris_basic.R
r3.4-small-client-mode	runit_deeplearning_iris_basic.R
r3.4-small	runit_deeplearning_iris_basic.R
r3.4-smoke	runit_deeplearning_iris_basic.R
```

IMHO, duplications like this are undesired - WDYT @michalkurka @michal-raska ?

With some more effort, this tooling can be adjusted to perform such analysis conveniently, or even automatically. The datasets are good enough to provide far more information - like time spent in duplications etc.

michalkurka:
@pkozelka thanks for the thorough investigation

Regarding the duplicities, are there any tests that take a significant amount of time?

info-check and smoke-tests are special, they are reusing portion of the tests to do some basic check sanity check. They should generally run very fast (seconds). The duplicities would be an issue if they were taking a long time. In that case they should be removed from smoke/info-check tests.

There will be always duplicities in small and client-small tests. That is intended.

pkozelka:
@michalkurka following list shows most significant duplications - selected are those taking more than 120 seconds in total (number before name is seconds):

**headless-test.js**:
pass	948	flow-headless
pass	586	flow-headless-medium
pass	109	flow-headless-smoke
**pyunit_H2OXGBoost_native_XGBoost_enum_multinomial_compare_sparse_singleNode.py**:
pass	46	py3.5-single-node
pass	89	py3.5-small
**pyunit_random_attack_medium.py**:
pass	37	py3.6-medium-large
pass	22	py3.6-medium-large
pass	99	py3.6-medium-large
**runit_as.h2o_sparse.R**:
pass	213	r3.4-small
pass	281	r3.4-small-client-mode
**runit_automl_args.R**:
pass	250	r3.4-small-automl
pass	272	r3.4-small-client-mode-automl
**runit_automl_leaderboard.R**:
pass	273	r3.4-small-automl
pass	182	r3.4-small-client-mode-automl
**runit_glrm_nnmf.R**:
pass	86	r3.4-small
pass	74	r3.4-small-client-mode
**runit_glrm_orthonnmf.R**:
pass	138	r3.4-small
pass	156	r3.4-small-client-mode
**runit_merge_overlap.R**:
pass	164	r3.4-small
pass	635	r3.4-small-client-mode
**runit_partialPlot.R**:
pass	44	r3.4-small
pass	153	r3.4-small-client-mode
**runit_pubdev_1674_scale_invariance_all_algo.R**:
pass	88	r3.4-small
pass	130	r3.4-small-client-mode
**runit_pubdev_3847.R**:
pass	410	r3.4-small
pass	443	r3.4-small-client-mode
**runit_pubdev_4870_sort_bug_pubdev4404_desc.R**:
pass	90	r3.4-small
pass	126	r3.4-small-client-mode
**runit_PUBDEV_5008_5386_ordinal_glm_nidhi_R.R**:
pass	55	r3.4-small
pass	130	r3.4-small-client-mode
**runit_pubdev_5074_glm-no-weights.R**:
pass	56	r3.4-small
pass	68	r3.4-small-client-mode
**runit_PUBDEV_5705_drop_columns_parser_gz.R**:
pass	297	r3.4-small
pass	317	r3.4-small-client-mode
**runit_p-value_synthetic.R**:
pass	92	r3.4-small
pass	116	r3.4-small-client-mode
**runit_repeating_merge.R**:
pass	364	r3.4-datatable
pass	295	r3.4-small
pass	332	r3.4-small-client-mode

michalkurka:
Thanks @pkozelka, in that case, I don't see a significant room for improvement because these are almost exclusively small and client-mode-small tests which are ok to be duplicated (same goes for automl-small and client-mode-automl-small).

PR #3207, opened 2019-01-07
PUBDEV-6176: moved StackedEnsembleModel to hex.ensemble package 

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6176

This implies changes in how SE scoring is done in predictScoreImpl, especially deactivated computeMetrics for base models at it appeared useless (we only need predictions).

Also fixes https://0xdata.atlassian.net/browse/PUBDEV-6266

Adding migration document for those changes:
This impacts only Java API:
- `hex.StackedEnsembleModel` -> `hex.ensemble.StackedEnsembleModel`
- `hex.StackedEnsembleModel.MetalearnerAlgorithm` -> `hex.ensemble.Metalearner.Algorithm`
- removed public access to following internal methods:
  - `StackedEnsembleModel.checkAndInheritModelProperties`.
  - `StackedEnsembleModel.doScoreOrCopyMetrics`.
  - `StackedEnsembleModel.doScoreMetricsOneFrame`.
  - `StackedEnsemble.addModelPredictionsToLevelOneFrame`.


h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -135,44 +119,22 @@ public void initMetalearnerParams(MetalearnerAlgorithm algo) {
    */
   @Override
   protected Frame predictScoreImpl(Frame fr, Frame adaptFrm, String destination_key, Job j, boolean computeMetrics, CFuncRef customMetricFunc) {
```
seb-h2o: @michalkurka : please review the rewrite of this `predictScoreImpl` method for Stacked Ensemble, that's also what fixes the issue PUBDEV-6266 (cf. test below).

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
```diff
@@ -417,6 +415,73 @@ public void testBaseModelPredictionsCaching() {
       }
     }
 
+    
+    @Test
+    public void test_SE_with_GLM_can_do_predictions_on_frames_with_unseen_categorical_values() {
```
seb-h2o: tested that this test fails on current master.

h2o-docs/src/product/migrating.rst
```diff
@@ -1,4 +1,30 @@
 .. _migration:
+.. include:: <isonum.txt>
+
+Migration Guide
+###############
+
+Migrating between H2O-3 versions
```
seb-h2o: @angela0xdata : I introduced this new section, and put it on top of the file as I believe that there are way more people today likely to migrate between h2o-3 versions than from h20-2 to 3.
Of course, any suggestion welcome.

angela0xdata: @seb-h2o I don't understand the include statement. What is <isonum.txt>?

h2o-docs/src/product/migrating.rst
```diff
@@ -1,4 +1,30 @@
...
+--------------------------
+
+Java API
+~~~~~~~~
+
+The following classes were moved and/or renamed:
+
+- ``hex.StackedEnsembleModel`` |rarr| ``hex.ensemble.StackedEnsembleModel``.
+- ``hex.StackedEnsembleModel.MetalearnerAlgorithm`` |rarr| ``hex.ensemble.Metalearner.Algorithm``.
+
```
angela0xdata: What is |rarr| in the above two bullets? These references are causing an error when building the docs.

angela0xdata:
@seb-h2o I made some updates to the heading levels. Also, please see comments regarding references and the include statement in the doc.

h2o-docs/src/product/migrating.rst
seb-h2o: @angela0xdata those are non-ascii characters (right-arrow in this case), hence the `isonum` import above.
I built the docs locally before and it worked fine, that's why I included it... maybe it depends on the environment though. If it doesn't work everywhere, I'll remove it.

seb-h2o: for reference: http://docutils.sourceforge.net/docs/ref/rst/definitions.html

seb-h2o:
@angela0xdata I removed the import for special characters and using a table formatting instead: feel free to try building the docs again.
Also thanks for the other changes! btw, don't know if you did the header changes mainly for consistency, but FYI rST format doesn't formally require to change all headers in this case, as it's rebuilding the hierarchy when parsing the file (matching similar underlines): it's a cool feature indeed as it allows shifting entire sub-trees just by adding/removing one single header and without having to modify the existing... pretty neat.

angela0xdata:
@seb-h2o Thanks for the info. Everything does build locally, but the build produced additional errors (which aren't harmful; I just don't like to see them if I can avoid them). But I agree that rst can be pretty nifty.

seb-h2o:
@michalkurka would appreciate if you could find time to review this today: tests are fixed, and this ticket is blocking me to complete https://github.com/h2oai/h2o-3/pull/3273 (leaking keys in UT after scoring, fixed by this scoring rewrite).

h2o-docs/src/product/migrating.rst
```diff
@@ -1,9 +1,39 @@
...
+Java API
+''''''''
+
+The following classes were moved and/or renamed:
+
+=================================================   ======================================
+  Until 3.22                                          From 3.24
+=================================================   ======================================
+``hex.StackedEnsembleModel``                        ``hex.ensemble.StackedEnsembleModel``
```
michalkurka: @jakubhava how does this affect Sparkling Water?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
michalkurka: Looks good to me so far - I will check one more time

h2o-core/src/main/java/water/util/ArrayUtils.java
```diff
@@ -1437,7 +1437,7 @@ public int compare(Integer x, Integer y) {
     if(s == null)return ary;
     int cnt = 0;
     int idx = find(ary,s);
-    while(idx > 0) {
+    while(idx >= 0) {
```
michalkurka: wow!

h2o-docs/src/product/migrating.rst
```diff
@@ -1,9 +1,39 @@
...
+``hex.StackedEnsembleModel.MetalearnerAlgorithm``   ``hex.ensemble.Metalearner.Algorithm``
+``ai.h2o.automl.AutoML.algo``                       ``ai.h2o.automl.Algo``
+=================================================   ======================================
+
+Some internal methods of ``StackedEnsemble`` and ``StackedEnsembleModel`` are no longer public, but this should not impact anyone.
+
+Migrating to H2O-3
```
michalkurka: @angela0xdata can we remove the h2o-2 to h2o-3 section, it is really not relevant anymore

h2o-docs/src/product/migrating.rst
jakubhava: nope, we're good there 👍 

Thanks for the heads up though!

michalkurka: How about the other 2? I remember last time we renamed `algos` it caused SW tests to fail.

jakubhava: the answer applies to other ones as well, the `algo->Algo` change is already in a few last SW releases, green. @seb-h2o did that in a smart backwards-compatible way.

angela0xdata:
@michalkurka I'm fine removing that legacy migrating section, but I would also defer to @tomkraljevic.

angela0xdata:
@seb-h2o @michalkurka @jakubhava, Tom has said agreed that we can remove that "Migrating to H2O 3" section.

h2o-docs/src/product/migrating.rst
michalkurka: perfect, thanks!

seb-h2o:
@angela0xdata merging now, will remove the migration section on a separate tiny PR

PR #3208, opened 2019-01-07
PUBDEV-6171: split the pyunit_pubdev_3500_max_k_large.py test into 3 …

wendycwong:
This PR completes the work in https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6171?filter=myopenissues

This JIRA specifies that the highest number of k is selected if it is set to -1.  This implies, we will set k=column number and perform PCA.  Hence, there is no need to use the huge dataset which only slows down the calculation.  I used a much smaller dataset and this speeds up the test.  I did not split the test in this case either.


Pscheidl:
Failed test seems to be unrelated: `352  FAIL  40010   19s pyunit_h2oinit.py`

wendycwong:
test run time off my personal mac for pyunit_pubdev_3500_max_k.py:

36s, 34s, 26s, 27s, 34s, 27s with mean 30.67second.

PR #3210, opened 2019-01-08
PUBDEV-6099 edit early stopping doc to be more clear

maurever:
Edit documentation about early stopping calculation to be clear how it is implemented in H2O.

JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6099

Related code: https://github.com/h2oai/h2o-3/blob/677ebe4336eaa5b9c5bfb0a05668b215d3f171c5/h2o-core/src/main/java/hex/ScoreKeeper.java#L123-L227


michalkurka:
@maurever please move this to rel-xu, the jira is marked for 3.22.1.x

PR #3213, opened 2019-01-09
PUBDEV-6172: Add capability to compare a smaller number of columns. (…

wendycwong:
Move fixed tests from master to rel-xia


wendycwong:
Need to push to rel-xu and not rel-xia

PR #3218, opened 2019-01-11
PUBDEV-5809: H2O-core changes to ensure per-model runtime limit

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-5809

To be able to enforce a `max_runtime` per model in AutoML, the following changes were required to get the feature working as expected:
1. improved/fixed GLM not respecting time limits.
2. ensure that time allocation for CV models is done at `ModelBuilder` level (instead of Schema, or Grids).
3. basic heuristic to allocate time to CVs even if they run in parallel.
4. standardized and encapsulate time/timeout measurements in new `Countdown` class.

see also https://github.com/h2oai/h2o-3/pull/3339 for AutoML feature requiring those changes.

### DIFF
--- a/h2o-algos/src/main/java/hex/deepwater/DeepWater.java
+++ b/h2o-algos/src/main/java/hex/deepwater/DeepWater.java
@@ -69,7 +69,7 @@ public class DeepWater extends ModelBuilder<DeepWaterModel,DeepWaterParameters,D
 
   @Override public boolean isSupervised() { return !_parms._autoencoder; }
 
-  @Override protected int nModelsInParallel() { return 1; }
+  @Override protected int nModelsInParallel(int folds) { return 1; }
 
   @Override protected DeepWaterDriver trainModelImpl() { return new DeepWaterDriver(); }
 
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -111,6 +111,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
    */
   @Override
   public void cv_computeAndSetOptimalParameters(ModelBuilder[] cvModelBuilders) {
+    if(_parms._max_runtime_secs != 0) _parms._max_runtime_secs = 0;
     if(_parms._lambda_search) {
       _xval_test_deviances = new double[_parms._lambda.length];
       _xval_test_sd = new double [_parms._lambda.length];
@@ -515,7 +516,6 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       GLMGradientInfo ginfo = new GLMGradientSolver(_job,_parms, _dinfo, 0, _state.activeBC()).getGradient(beta);
       _lmax = lmax(ginfo._gradient);
       _state.setLambdaMax(_lmax);
-      _model = new GLMModel(_result, _parms, GLM.this, _state._ymu, _dinfo._adaptedFrame.lastVec().sigma(), _lmax, _nobs);
       if (_parms._lambda_min_ratio == -1) {
         _parms._lambda_min_ratio = (_nobs >> 4) > _dinfo.fullN() ? 1e-4 : 1e-2;
         if(_parms._alpha[0] == 0)
@@ -553,10 +553,17 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
         _parms._gradient_epsilon = _parms._lambda[0] == 0 ? 1e-6 : 1e-4;
         if(_parms._lambda_search) _parms._gradient_epsilon *= 1e-2;
       }
-      // clone2 so that I don't change instance which is in the DKV directly
-      // (clone2 also shallow clones _output)
-      _model.clone2().delete_and_lock(_job._key);
+      buildModel();
     }
+  }
+
+  // FIXME: contrary to other models, GLM output duration includes computation of CV models:
+  //  ideally the model should be instantiated in the #computeImpl() method instead of init
+  private void buildModel() {
+    _model = new GLMModel(_result, _parms, this, _state._ymu, _dinfo._adaptedFrame.lastVec().sigma(), _lmax, _nobs);
+    // clone2 so that I don't change instance which is in the DKV directly
+    // (clone2 also shallow clones _output)
+    _model.clone2().delete_and_lock(_job._key);
   }
 
   protected static final long WORK_TOTAL = 1000000;
@@ -1211,6 +1218,8 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
         init(true);
       if (error_count() > 0)
         throw H2OModelBuilderIllegalArgumentException.makeFromBuilder(GLM.this);
+//      buildModel();
+      _model._output._start_time = System.currentTimeMillis(); //quickfix to align output duration with other models
       if(_parms._lambda_search) {
         if (_parms._family == Family.ordinal)
           _nullDevTrain = new GLMResDevTaskOrdinal(_job._key,_state._dinfo,getNullBeta(), _nclass).doAll(_state._dinfo._adaptedFrame).avgDev();
@@ -1260,7 +1269,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       if(!_parms._lambda_search)
         updateProgress(false);
       // lambda search loop
-      for (int i = 0; i < _parms._lambda.length; ++i) {  // lambda search
+      for (int i = 0; i < _parms._lambda.length && !stop_requested(); ++i) {  // lambda search
         if(_parms._max_iterations != -1 && _state._iter >= _parms._max_iterations)
           break;
         Submodel sm = computeSubmodel(i,_parms._lambda[i]);
@@ -1291,6 +1300,13 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
           scoreAndUpdateModel(); // update partial results
         }
         _job.update(_workPerIteration,"iter=" + _state._iter + " lmb=" + lambdaFormatter.format(_state.lambda()) + "deviance trn/tst= " + devFormatter.format(trainDev) + "/" + devFormatter.format(testDev) + " P=" + ArrayUtils.countNonzeros(_state.beta()));
+      }
+      if (stop_requested()) {
+        if (timeout()) {
+          Log.info("Stopping GLM training because of timeout");
+        } else {
+          throw new Job.JobCancelledException();
+        }
       }
       if(_state._iter >= _parms._max_iterations)
         _job.warn("Reached maximum number of iterations " + _parms._max_iterations + "!");
@@ -1331,7 +1347,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
         _state.updateState(beta, gginfo);
         if (!_parms._lambda_search)
           updateProgress(false);
-        return !timeout() && !_job.stop_requested() && _state._iter < _parms._max_iterations;
+        return !stop_requested() && _state._iter < _parms._max_iterations;
       } else {
         GLMGradientInfo gginfo = (GLMGradientInfo) ginfo;
         if(gginfo._gradient == null)
@@ -1342,7 +1358,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
           updateProgress(true);
         boolean converged = _state.converged();
         if (converged) Log.info(LogMsg(_state.convergenceMsg));
-        return !timeout() && !_job.stop_requested() && !converged && _state._iter < _parms._max_iterations;
+        return !stop_requested() && !converged && _state._iter < _parms._max_iterations;
       }
     }
 
@@ -1353,7 +1369,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
         updateProgress(true);
       boolean converged = _state.converged();
       if(converged) Log.info(LogMsg(_state.convergenceMsg));
-      return !_job.stop_requested() && !converged && _state._iter < _parms._max_iterations ;
+      return !stop_requested() && !converged && _state._iter < _parms._max_iterations ;
     }
 
     private transient long _scoringInterval = SCORING_INTERVAL_MSEC;
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -1269,9 +1269,9 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       if(!_parms._lambda_search)
         updateProgress(false);
       // lambda search loop
-      for (int i = 0; i < _parms._lambda.length && !stop_requested(); ++i) {  // lambda search
-        if(_parms._max_iterations != -1 && _state._iter >= _parms._max_iterations)
-          break;
+      for (int i = 0; i < _parms._lambda.length; ++i) {  // lambda search
+        if (_job.stop_requested() || (timeout() && _model._output._submodels.length > 0)) break;  //need at least one submodel on timeout to avoid issues.
+        if(_parms._max_iterations != -1 && _state._iter >= _parms._max_iterations) break;
         Submodel sm = computeSubmodel(i,_parms._lambda[i]);
         double trainDev = sm.devianceTrain;
         double testDev = sm.devianceTest;
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -167,7 +167,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
   public static void startAutoML(AutoML aml) {
     // Currently AutoML can only run one job at a time
     if (aml.job == null || !aml.job.isRunning()) {
-      H2OJob j = new H2OJob(aml, aml._key, aml.timeRemainingMs());
+      H2OJob j = new H2OJob(aml, aml._key, aml.runCountdown.remainingTime());
       aml.job = j._job;
       j.start(aml.workAllocations.remainingWork());
       DKV.put(aml);
@@ -420,10 +420,6 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     if (buildSpec.build_control.stopping_criteria.max_models() == 0)
       return Integer.MAX_VALUE;
     return buildSpec.build_control.stopping_criteria.max_models() - modelCount.get();
-  }
-
-  private boolean timingOut() {
-    return timeRemainingMs() <= 0;
   }
 
   @Override
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -582,16 +582,17 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     jobs.add(subJob);
 
     long lastWorkedSoFar = 0;
-    int gridLastCount = 0;
+    Set<Key<Model>> lastGridModels = new HashSet<>();
+    long lastGridModelCreation = subJob.start_time();
 
     while (subJob.isRunning()) {
       if (null != parentJob) {
         if (parentJob.stop_requested()) {
-          userFeedback.info(Stage.ModelTraining, "AutoML job cancelled; skipping " + name);
+          userFeedback.info(stage, "AutoML job cancelled; skipping " + name);
           subJob.stop();
         }
         if (!ignoreTimeout && timingOut()) {
-          userFeedback.info(Stage.ModelTraining, "AutoML: out of time; skipping " + name);
+          userFeedback.info(stage, "AutoML: out of time; skipping " + name);
           subJob.stop();
         }
       }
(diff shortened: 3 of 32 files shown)


h2o-core/src/main/java/water/util/Countdown.java
```diff
@@ -0,0 +1,60 @@
+package water.util;
+
+import java.io.Serializable;
+
+/**
+ * Simple countdown to encapsulate timeouts and durations.
+ * time_limit_millis <= 0 is interpreted as infinite countdown (no timeout)
+ */
+public class Countdown implements Serializable {
```
michalkurka: should be Iced instead of Serializable (or nothing ;))

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -220,6 +218,7 @@ public AutoML(Key<AutoML> key, Date startTime, AutoMLBuildSpec buildSpec) {
     super(key);
 
     this.startTime = startTime;
+    this.runCountdown = new Countdown(Math.round(1000 * buildSpec.build_control.stopping_criteria.max_runtime_secs()));
```
michalkurka: perhaps it would be better to keep the countdown in seconds instead of ms?

h2o-core/src/main/java/water/util/Countdown.java
seb-h2o: well, I definitely forgot to provide empty constructor so Serializable wouldn't work... but is it really necessary to implement Iced for such simple classes (2 long fields)? I mean technically, the Weaver delegates to Java serialization for Serializable objects...
Just asking to personal info ;) switching to Iced anyway, thx for noticing.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: I prefer to keep it in ms, there's information loss when converting ms to s, so I don't want this to be done implicitely: for example 50ms could be converted either to 0s or 1s, and if 0s, then the semantic changes (here it's interpreted as no timeout like almost everywhere in our code).

The best I can do is provide a static factory method maybe: `Countdown.fromSeconds()`

ledell:
@seb-h2o I'm wondering if the name `max_model_runtime_secs` could be misleading because it is also the combination of the names of our two existing per-run limit variables,`max_models` and `max_runtimes_secs`.  Maybe we could do something like `max_runtime_secs_per_model`?  It's longer but might be more explicit... ?  What do you think?

seb-h2o:
@ledell I have to agree with you indeed... when writing tests, I found myself intuitively searching (and expecting from autocompletion) for `max_runtime_single_model`, `max_runtime_individual_model` and so on... 
so yes, I'll rename it: `max_runtime_secs_per_model` sounds good.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1261,8 +1270,8 @@ public void computeImpl() {
         updateProgress(false);
       // lambda search loop
       for (int i = 0; i < _parms._lambda.length; ++i) {  // lambda search
-        if(_parms._max_iterations != -1 && _state._iter >= _parms._max_iterations)
-          break;
+        if (_job.stop_requested() || (timeout() && _model._output._submodels.length > 0)) break;  //need at least one submodel on timeout to avoid issues.
+        if(_parms._max_iterations != -1 && _state._iter >= _parms._max_iterations) break;
```
seb-h2o: @wendycwong , can you please verify that those interruptions can't do any harm?
Ideally, I'd like for GLM to stop gracefully in case of timeout, possibly with a not-fully trained model.
In case of stop user request, we raise an exception, so I'm less concerned...

h2o-core/src/main/java/water/api/schemas3/ModelParametersSchemaV3.java
```diff
@@ -209,7 +209,6 @@ public P fillImpl(P impl) {
 
     impl._train = (this.training_frame == null) ? null : Key.<Frame>make(this.training_frame.name);
     impl._valid = (this.validation_frame == null) ? null : Key.<Frame>make(this.validation_frame.name);
-    impl._max_runtime_secs = nfolds > 0 ? max_runtime_secs / (nfolds+1) : max_runtime_secs;
```
seb-h2o: @michalkurka : for your concern, this is one of the main changes related with model building, (same in GridSearch): this time-repartition logic is now fully done in ModelBuilder.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -56,6 +56,10 @@
   public XGBoost(boolean startup_once) { super(new XGBoostModel.XGBoostParameters(),startup_once); }
   public boolean isSupervised(){return true;}
 
+  @Override protected int nModelsInParallel(int folds) {
+    return nModelsInParallel(folds, 2);
```
seb-h2o: @michalkurka : at that time, the default parallelization for XGBoost was 2, same as GBM. Should I changed it to 1? For info, seems to work fine with 2.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -389,10 +394,22 @@ public void run() {
...
-    return 1; //safe fallback
+  protected int nModelsInParallel(int folds) {
+    return nModelsInParallel(folds, 1);
+  }
+
+  protected int nModelsInParallel(int folds, int defaultParallelization) {
+    if (!_parms._parallelize_cross_validation) return 1; //user demands serial building (or we need to honor the time constraints for all CV models equally)
+    if (_train.byteSize() < 1e6) return folds; //for small data, parallelize over CV models
+    // TODO: apply better heuristic, estimating parallelization based on H2O.getCloudSize() and H2O.ARGS.nthreads
+    return defaultParallelization;
```
seb-h2o: maybe we should ensure that the parallelization is always <= ncores.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
ledell: @seb-h2o Does this mean that only 2 models are trained in parallel even if `nfolds = 5`?  I think we should really review the defaults for CV parallelization based on the size of the data vs size of the H2O cluster, but that's a conversation for another day!

seb-h2o: that's correct, and afair only GBM (and in this PR, XGB) have parallelization activated for large data. It's also always activated for most algos when data are very small, like < 1MB.
I also already added a comment in ModelBuilder that goes in the same direction: we could really improve this heuristic based on cluster size, nthreads, data size... maybe we should just create a ticket for this.

seb-h2o: @ledell created https://0xdata.atlassian.net/browse/PUBDEV-6312 for future reference

ledell: @seb-h2o Thanks.  We have a simplified ticket here (based on data size only), I'll link them in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-5783

seb-h2o:
@wendycwong , @michalkurka : anything to add to this PR?

seb-h2o:
@wendycwong , @michalkurka : pretty please... any comment?

h2o-algos/src/main/java/hex/deepwater/DeepWater.java
```diff
@@ -69,7 +69,7 @@ static boolean haveBackend(DeepWaterParameters.Backend b) {
 
   @Override public boolean isSupervised() { return !_parms._autoencoder; }
 
-  @Override protected int nModelsInParallel() { return 1; }
+  @Override protected int nModelsInParallel(int folds) { return 1; }
```
michalkurka: Do we need this new argument? Can it be retrieved from model parameters instead?

seb-h2o: it's the result of `ModelBuilder#nFoldWork()` that is computed during xval,  so as it's a protected method only, it's safer to pass this result as a parameter.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: I feel this should work.  As long as your fix pass Jenkin GLM tests, I feel confident.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -111,6 +111,7 @@ public void computeCrossValidation() {
    */
   @Override
   public void cv_computeAndSetOptimalParameters(ModelBuilder[] cvModelBuilders) {
+    if(_parms._max_runtime_secs != 0) _parms._max_runtime_secs = 0;
     if(_parms._lambda_search) {
```
wendycwong: I thought setting _parms._max_runtime_secs to 0 will disable maximum run time restriction?

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -111,6 +111,7 @@ public void computeCrossValidation() {
    */
   @Override
   public void cv_computeAndSetOptimalParameters(ModelBuilder[] cvModelBuilders) {
+    if(_parms._max_runtime_secs != 0) _parms._max_runtime_secs = 0;
```
wendycwong: I thought setting max_runtime_secs to 0 will disable run time restriction.  Why are you setting it here?

michalkurka:
@seb-h2o please consider splitting into 2 PRs: automl changes and code changes separate

h2o-algos/src/main/java/hex/glm/GLM.java
seb-h2o: @wendycwong , this if for consistency with other algos which do the same timeout cancellation in this method: this cancellation applies only for the main model, which means that currently (and for some reason not fully clear to me) when we compute a model with CV=5 for example, with a time limit of (to make it simple) 12min, we assign 2 min (effective) for each CV model, and 2 min for the main model, but those last 2 min are kind of virtual as we will effectively use as much time as necessary...

Please see other algos for comparison: if in GLM's case, there is a better reason to keep the time limit for the main model, then I'll be happy to revert this change and add a comment explaining why this is not necessary here.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: This part looks fine.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: seb:  setting max_runtime_secs=0 will disable the time restriction.  I am confused to why you are setting it to 0 here.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -517,9 +520,9 @@ public MP nextModelParameters(Model previousModel) {
               double timeleft = this.time_remaining_secs();
               if (timeleft > 0)  {
                 if (params._max_runtime_secs > 0) {
-                  params._max_runtime_secs = (long) floor(min(params._max_runtime_secs, timeleft));
+                  params._max_runtime_secs = min(params._max_runtime_secs, timeleft);
```
seb-h2o: removed rounding that can lead to have small timeout converted to 0 (no timeout): tried to keep double all the way to avoid those annoying rounding issue

h2o-core/src/main/java/water/api/Schema.java
```diff
@@ -341,6 +341,21 @@ public S fillFromParms(Properties parms, boolean checkRequiredFields) {
...
+   * We use PojoUtils.fillFromJson() rather than just using "schema = Gson.fromJson(post_body)"
+   * so that we have defaults: we only overwrite fields that the client has specified.
+   *
+   * @param body the post body (can't be null), converted to JSON by default
+   * @return the filled schema
+   */
+  public S fillFromBody(String body) {
+    PojoUtils.fillFromJson(this, body);
+    return (S) this;
+  }
```
seb-h2o: rationale for this new method:
- provide a simple way to customize parsing of body params: necessary in AutoML.
- method name symmetric with `fillFromParams`

h2o-core/src/main/java/water/api/Handler.java
```diff
@@ -40,20 +40,17 @@ public Schema handle(int version, Route route, Properties parms, String post_bod
...
-    // with the exception of control-type query parameters.
-    //
-    // We use PojoUtils.fillFromJson() rather than just using "schema = Gson.fromJson(post_body)"
-    // so that we have defaults: we only overwrite fields that the client has specified.
-    if (is_post_of_json) {
-      PojoUtils.fillFromJson(schema, post_body);
+    //Fill from JSON body, if there is one.  NOTE: there should *either* be a JSON body *or* parms,
+    //with the exception of control-type query parameters.
+    if (has_body) {
+      schema = schema.fillFromBody(post_body);
```
seb-h2o: see `Schema` for rationale behind those changes

h2o-core/src/main/java/water/api/EnumValuesProvider.java
```diff
@@ -1,26 +1,34 @@
...
 
-  private String[] getValuesOf(Class<E> clazz) {
+  private String[] getValuesOf(Class<E> clazz, List<E> excluded) {
     E[] values = clazz.getEnumConstants();
     List<String> names = new ArrayList<>(values.length);
     for (E val : values) {
-      names.add(val.name());
+      if (!excluded.contains(val)) {
+        names.add(val.name());
+      }
```
seb-h2o: used in AutoML

h2o-core/src/main/java/hex/schemas/HyperSpaceSearchCriteriaV99.java
```diff
@@ -3,6 +3,8 @@
 import hex.ScoreKeeper;
 import hex.grid.HyperSpaceSearchCriteria;
 import water.api.API;
+import water.api.EnumValuesProvider;
+import water.api.schemas3.ModelParamsValuesProviders.StoppingMetricValuesProvider;
```
seb-h2o: using the values provider in this class as I was creating some of them for `AutoMLStoppingCriteriaV99`

h2o-algos/src/main/java/hex/glm/GLM.java
seb-h2o: @michalkurka do you know why we disable timeout for main model when training with CV for most algos? 
Is it really necessary? also, did we forget to disable it on GLM or was it missing on purpose?

h2o-algos/src/main/java/hex/deepwater/DeepWater.java
michalkurka: Can we still keep nModelsInParallel in the no-parameter version as well and mark it as deprecated? There are internal projects that depend on it -> people out there might be using it as well.

The method could produce a warning and the behavior doesn't have to be perfect. WDYT?

h2o-algos/src/main/java/hex/glm/GLM.java
michalkurka: @seb-h2o Are you referring to GBM/DRF and Deep Learning which are doing this?

    if( _parms._stopping_rounds == 0 && _parms._max_runtime_secs == 0) return; // No exciting changes to stopping conditions

`cv_computeAndSetOptimalParameters` in those models takes into consideration the CV models and builds the main model bases on the parameters of the CV models - it calculates the average number of trees or epochs in deep learning.

The main model should not be restricted by time because you want to build a good model, not eg. an empty model while your CV models are good.

GLM is currently not using `max_runtime_secs` AFAIK

h2o-algos/src/main/java/hex/deepwater/DeepWater.java
seb-h2o: @michalkurka sure, I need to hook it into the one with param though as it's the last one that is used by `ModelBuilder`
 
change proposal in ModelBuilder:
```
@Deprecated protected int nModelsInParallel() { return 0; }

protected int nModelsInParallel(int folds) { 
   if (nModelsInParallel() > 0) return nModelsInParallel();
   // current logic goes there
}

Comment (seb-h2o) [reply h2o-algos/src/main/java/hex/deepwater/DeepWater.java]: when you say internal projects are using it: do you mean they're overriding it? or are they consuming/calling it?
my proposal works in the former case but not the later.

File: h2o-algos/src/main/java/hex/glm/GLM.java (hunk 9 above)
Comment (michalkurka) [reply h2o-algos/src/main/java/hex/glm/GLM.java]: sorry, GLM is using max_runtime_secs

File: h2o-algos/src/main/java/hex/deepwater/DeepWater.java (hunk 7 above)
Comment (seb-h2o) [reply h2o-algos/src/main/java/hex/deepwater/DeepWater.java]: in case or consuming the default impl would rather be sth like:
```
@Deprecated protected int nModelsInParallel() {
  return nModelsInParallel(_parms._nfolds)
}
```

Comment (michalkurka) [reply h2o-algos/src/main/java/hex/deepwater/DeepWater.java]: I think your proposal makes sense

File: h2o-core/src/main/java/hex/Model.java [hunk 15]
```diff
@@ -1530,7 +1535,7 @@ protected Frame predictScoreImpl(Frame fr, Frame adaptFrm, String destination_ke
                                    j,
                                    customMetricFunc).doAll(names.length, Vec.T_NUM, adaptFrm);
 
-    if (computeMetrics)
+    if (computeMetrics && bs._mb != null) //metric builder can be null if training was interrupted/cancelled
```
michalkurka: hmm, what other issues can there be? the model will be invalid basically - can we make sure the model will have metrics if there was a timeout? Not sure where else can it crash

michalkurka:
I think this is overall a great change. My only concern is that it looks like we are able to train models that don't have metrics and it seems like this might violate some invariants we have in the code.

h2o-algos/src/main/java/hex/deepwater/DeepWater.java
seb-h2o: @michalkurka which one? :)
the first one right? I expect this being mainly overridden, not called internally as it seems to have been introduced for xval only.

h2o-algos/src/main/java/hex/glm/GLM.java
seb-h2o: @michalkurka Yes, just wanted to have the confirmation that we should therefore also disable `max_runtime_secs` for CV main model in GLM as well

h2o-core/src/main/java/hex/Model.java
seb-h2o: yes, ensuring `bs._mb != null` looks simple, it's just about moving one line in `BigScore.map` below. Will give it a try.

h2o-algos/src/main/java/hex/glm/GLM.java
michalkurka: yes, disabling the limit here makes sense to me

h2o-algos/src/main/java/hex/deepwater/DeepWater.java
michalkurka: I don't see the use case for calling directly, but overriding the method in a custom implementation makes sense for me (we do it as well in the code).

h2o-core/src/main/java/hex/Model.java
seb-h2o: @michalkurka I doubled checked this code: there should be no issue on timeout:
- there's no timeout on model scoring, it's handled only at model driver level.
- in case of training timeout, we throw an exception, we don't call `job.stop()`.
- only AutoML may call `job.stop()` in case of global timeout. I can ensure that `BigScore.map` always returns a `MetricBuilder` in this case, but I'm more comfortable with the null-check above: if there's no metric, the model won't appear in our leaderboard anyway.
wdyt?

h2o-algos/src/main/java/hex/deepwater/DeepWater.java
seb-h2o: @michalkurka just added it to `ModelBuilder`: see https://github.com/h2oai/h2o-3/pull/3218/files#diff-ab9d2bd762456be481b3fa49912d6b64R391

h2o-core/src/main/java/hex/Model.java
michalkurka: @seb-h2o Okay, I think that is reassuring.

What about models that override `predictScoreImpl`? Do they need similar handling as well?

seb-h2o: no, not necessary: the main concern is about result of `makeBigScoreTask`

seb-h2o: @michalkurka imo, it's currently almost impossible to ensure a completely graceful training cancellation: I don't think we have it perfect currently. This NPE was just an exception I've seen several times in the logs when AutoML times out.
But for core concerns, it's relevant only for job cancellation, so the less exceptions, the nicer it is, but I'm sure there are still a few hiding here and there.

michalkurka: yeah, that is what I meant - `protected BigScore makeBigScoreTask` - makeBigScoreTask is protected and the models can call it (most likely in their overridden predictScoreImpl or other scoring functions).

XGBoost used to call it but it doesn't anymore, I checked. Just wondering about the other places.

seb-h2o:
@michalkurka what's the status with this PR? can it be merged?
regarding `makeBigScoreTask`, I can create an additional ticket to deep delve into it: the potential NPE has been there for years given the current `BigScore.map` logic, so this is a bit out of scope.
The only models I've seen susceptible of encountering a similar NPE on cancellation are DeepLearning and DeepWater (if null result from call to `scoreMetrics`.

michalkurka:
@seb-h2o if your PR doesn't change the current behavior then my concern expressed in https://github.com/h2oai/h2o-3/pull/3218/files#r264899479 doesn't have to be addressed in this PR and we are okay to merge.

PR #3221, opened 2019-01-12
PUBDEV-6177 - Python API no longer fails on empty/unfilled categorical levels.

Pscheidl:
[PUBDEV-6177](https://0xdata.atlassian.net/browse/PUBDEV-6177)

The root cause will be solved in: https://0xdata.atlassian.net/browse/PUBDEV-6140


Pscheidl:
Introduced a new test just for IRF with the specified dataset. Made sure Python does not fail on this error, as R does.

However, the root cause seems to be IRF not filling these values correctly or the conversion mechanism not working properly for IRF (not seen this / was unable to reproduce with DRF/GBM/XGBoost).

PR #3223, opened 2019-01-12
PUBDEV-6188 - Amazon S3 fetching with credentials in URL

Pscheidl:
[PUBDEV-6188](https://0xdata.atlassian.net/browse/PUBDEV-6188)

This is the **S3** PR, not S3A. This PR touches **PersistS3**, not PersistHDFS.

### DIFF
--- a/h2o-persist-s3/src/main/java/water/persist/PersistS3.java
+++ b/h2o-persist-s3/src/main/java/water/persist/PersistS3.java
@@ -6,11 +6,14 @@ import com.amazonaws.ClientConfiguration;
 import com.amazonaws.Protocol;
 import com.amazonaws.auth.*;
 import com.amazonaws.auth.profile.ProfileCredentialsProvider;
+import com.amazonaws.internal.StaticCredentialsProvider;
 import com.amazonaws.regions.RegionUtils;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3Client;
 import com.amazonaws.services.s3.S3ClientOptions;
 import com.amazonaws.services.s3.model.*;
+import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.commons.lang.builder.ToStringBuilder;
 import water.*;
 import water.fvec.FileVec;
 import water.fvec.S3FileVec;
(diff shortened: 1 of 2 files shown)


Pscheidl:
TODO: Client caching strategy. As credentials are supplied via URI, a new client is built with static credential provider. This means the bucket cache must also contain caches. We should also make sure those are destroyed over time. Could be a significant resource consumer for long-running instances with data loaded from lots of different buckets (client re-init time after time is not really bad, main point of the cache is not to call S3 while user is selecting the file e.g. in Flow).

Pscheidl:
Sidenote: Unlike Hadoop's wrapper, Amazon's client is able to handle special characters just fine, no need to do encoding/decoding on our own.

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -25,35 +28,74 @@
...
+    if (accessKeyId != null && accessSecretKey != null) {
+      StringBuilder keyDigestBuilder = new StringBuilder(accessKeyId);
+      keyDigestBuilder.append(accessSecretKey);
+      digest = _messageDigest.digest(keyDigestBuilder.toString().getBytes());
+    } else {
+      digest = null;
+    }
+
+    synchronized (_lock) {
+      if (_s3 == null || digest != _credentialsDigest) {
```
michalkurka: `_credentialsDigest ` is never updated and stays `null` in this PR.

h2o-persist-s3/src/test/java/water/persist/PersistS3Test.java
```diff
@@ -9,17 +9,21 @@
...
  * Created by tomas on 6/27/16.
  */
 public class PersistS3Test extends TestUtil {
-
+  private static final String AWS_ACCESS_KEY_PROPERTY_NAME = "AWS_ACCESS_KEY_ID";
+  private static final String AWS_SECRET_KEY_PROPERTY_NAME = "AWS_SECRET_ACCESS_KEY";
 
   @BeforeClass
-  public static void setup() { stall_till_cloudsize(5); }
+  public static void setup() { stall_till_cloudsize(1); }
```
michalkurka: I would keep 5 there

h2o-persist-s3/src/test/java/water/persist/PersistS3Test.java
```diff
@@ -9,17 +9,21 @@
...
 import static junit.framework.TestCase.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeTrue;
 
 /**
  * Created by tomas on 6/27/16.
  */
 public class PersistS3Test extends TestUtil {
-
+  private static final String AWS_ACCESS_KEY_PROPERTY_NAME = "AWS_ACCESS_KEY_ID";
```
michalkurka: The names of these properties coincide with the names AWS libraries are using to lookup the credentials.

That is why test `testS3UriToKeyUriCredentials` doesn't actually prove that credentials encoded in URL will work (the S3 libraries might take it from the system properties). Aka if I intentionally remove the credentials the test will still pass.

h2o-persist-s3/src/test/java/water/persist/PersistS3Test.java
```diff
@@ -57,4 +61,89 @@ public void testS3Import() throws Exception {
       Scope.exit();
     }
   }
+
+  @Test
+  public void testS3UriToKeyUriCredentials() throws Exception {
+    // This test is only runnable in environment with Amazon credentials properly set {AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY}
+    final String accessKey = System.getenv(AWS_ACCESS_KEY_PROPERTY_NAME);
```
michalkurka: Not a good idea imho, reasoning please see above ^^^

michalkurka: Ideally, the test should also show that the import won't work without the credentials in the URL. When a test depends on an external system (or generally has an external assumption) it is good to validate the assumptions before running the test.

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
Pscheidl: I know. Latest patch not applied. WIP.

PR #3233, opened 2019-01-17
PUBDEV-6213: Added broadcast join for Target Encoding

deil87:


h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,137 @@
...
+   * @param leftCatColumnsIdxs indices of the categorical columns from `leftFrame` we want to use to calculate encodings
+   * @param leftFoldColumnIdx index of the fold column from `leftFrame` or `-1` if we don't use folds
+   * @param broadcastedFrame supposedly small frame that we will broadcast to all nodes and use it as lookup table for joining
+   * @param rightCatColumnsIdxs indices of the categorical columns from `broadcastedFrame` we want to use to calculate encodings 
+   * @param rightFoldColumnIdx index of the fold column from `broadcastedFrame` or `-1` if we don't use folds
+   * @return
+   */
+  static Frame join(Frame leftFrame, int[] leftCatColumnsIdxs, int leftFoldColumnIdx, Frame broadcastedFrame, int[] rightCatColumnsIdxs, int rightFoldColumnIdx) {
+    int numeratorIdx = broadcastedFrame.find("numerator");
+    int denominatorIdx = broadcastedFrame.find("denominator");
```
seb-h2o: not a big fan of magic strings in public classes, especially when they rely on some logic done in another public class (here TargetEncoder)

h2o-automl/src/test/java/ai/h2o/automl/targetencoding/BroadcastJoinTest.java
```diff
@@ -0,0 +1,115 @@
...
+      
+      Frame joined = BroadcastJoinForTargetEncoder.join(fr, new int[]{0}, 1, rightFr, new int[]{0}, 1);
+
+      Scope.enter();
+      assertStringVecEquals(cvec("a", "c", "b"), joined.vec("ColA"));
+      assertEquals(22, joined.vec("numerator").at(0), 1e-5);
+      assertEquals(42, joined.vec("numerator").at(1), 1e-5);
+      assertEquals(44, joined.vec("denominator").at(0), 1e-5);
+      assertEquals(84, joined.vec("denominator").at(1), 1e-5);
+      Scope.exit();
```
seb-h2o: aren't those assertions true also without fold columns?
I mean 3rd row should be the one with different behaviour, like null values, no?

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override
+    public int hashCode() {
+      return (_levelValue + _foldValue).hashCode();
+    }
+  }
+
+  static class EncodingData {
+    private long numerator;
+    private long denominator;
```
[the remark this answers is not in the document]
deil87: Right!

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,137 @@
...
+   * 
+   * @param leftFrame frame that we want to keep order of
+   * @param leftCatColumnsIdxs indices of the categorical columns from `leftFrame` we want to use to calculate encodings
+   * @param leftFoldColumnIdx index of the fold column from `leftFrame` or `-1` if we don't use folds
+   * @param broadcastedFrame supposedly small frame that we will broadcast to all nodes and use it as lookup table for joining
+   * @param rightCatColumnsIdxs indices of the categorical columns from `broadcastedFrame` we want to use to calculate encodings 
+   * @param rightFoldColumnIdx index of the fold column from `broadcastedFrame` or `-1` if we don't use folds
+   * @return
+   */
+  static Frame join(Frame leftFrame, int[] leftCatColumnsIdxs, int leftFoldColumnIdx, Frame broadcastedFrame, int[] rightCatColumnsIdxs, int rightFoldColumnIdx) {
```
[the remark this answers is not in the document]
deil87: I was thinking about overloading even with cases when we don't have fold column. Probably I will add one. And yes, I will try to support it in the near future.

michalkurka:
@deil87 looks good to me!

There is one thing I would consider changing - using Strings in CompositeLookupKey. Target Encoding is applied to categorical values, it would make sense to create the Map based on current  indexes of the categorical values in the column you want to encode and keep ints instead of Strings in the CompositeLookupKey. The issue I foresee here is that the Map could get big (and length of the String values would matter). Using indexes yields data structures whose size is more predictable and is a function of the number of rows only.

h2o-automl/src/test/java/ai/h2o/automl/targetencoding/BroadcastJoinTest.java
deil87: Yes you are right, but I was checking that I'm able to get what I expected to get(I was more concerned on the right values in each particular row). To make it more clear I will add extra checks, thank you!

deil87: it is just difficult to create vec of ints with null in it....  inconvenient to check this kind of things

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: Me neither, but in this case I would have to pass them as parameter to a main class( or worse to method) that would complicate signatures/API. But yes, these TargetEncoding and BroadcastJoiner classes are tightly coupled....but I made it explicit by adding `ForTargetEncoding` to the naming of the class.

deil87:
@seb Agree, I like generalisation too but I think since I don't have two use cases(which would make it a requirement)  for now  I'm free not to do this. This it actually called `premature generalization` and considered as a bad thing to do.

michalkurka:
@deil87 I don't think it is a must-have for the first version if it adds too much complexity. Better to have it simpler and improve later.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,137 @@
...
+    CompositeLookupKey(String levelValue, long fold) {
+      this._levelValue = levelValue;
+      this._foldValue = fold;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (obj != null && obj instanceof CompositeLookupKey) {
+        CompositeLookupKey s = (CompositeLookupKey) obj;
+        return _levelValue.equals(s._levelValue) && _foldValue.equals(s._foldValue);
```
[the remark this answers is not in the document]
deil87: I don't know for sure but my reasoning told me that if we have categorical column at hand then it shouldn't contain nulls in it. Time will answer to this question or maybe not :)

deil87:
@michalkurka agree. I will add comment to the code about possible option to improve things

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
seb-h2o: or just defined as constants to show where it comes from?

deil87: @seb-h2o yep, good idea

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,139 @@
...
+import water.fvec.Frame;
+import water.util.IcedHashMap;
+
+class BroadcastJoinForTargetEncoder {
+  
+  static class CompositeLookupKey {
+    // Note: Consider using indexes of levels instead of strings to save space. It will probably require 
+    // to use CategoricalWrappedVec.computeMap() since indexes for levels could differ in both frames for the same level values.
+    private String _levelValue;
+    private Long _foldValue;
```
michalkurka: Does this need to be a Long? The constructor takes a `long`. Shouldn't' int be enough?

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,139 @@
...
+      if (obj != null && obj instanceof CompositeLookupKey) {
+        CompositeLookupKey s = (CompositeLookupKey) obj;
+        return _levelValue.equals(s._levelValue) && _foldValue.equals(s._foldValue);
+      }
+      return false;
+    }
+
+    @Override
+    public int hashCode() {
+      return (_levelValue + _foldValue).hashCode();
```
michalkurka: Creating a new object (String) to calculate a hashCode is very inefficient. I strongly suggest to change it. A simple auto-generated hashCode should be good enough for your purposes and provide good performance.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,139 @@
...
+      Chunk numeratorChunk = cs[_numeratorIdx];
+      Chunk denominatorChunk = cs[_denominatorIdx];
+      for (int i = 0; i < categoricalChunk.len(); i++) {
+        long levelValue = categoricalChunk.at8(i);
+        String factor = categoricalChunk.vec().factor(levelValue);
+        long foldValue = _foldColumnIdx != -1 ? cs[_foldColumnIdx].at8(i) : -1;
+        _encodingDataMap.put(new CompositeLookupKey(factor, foldValue), new EncodingData(numeratorChunk.at8(i), denominatorChunk.at8(i)));
+      }
+    }
+  }
```
michalkurka: this needs a reducer, will not work in multinode environment, will need a fix

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: @michalkurka `Long` instead of `long` I wanted to use to be able to pass `null` for the case when we don't have fold column, but then forgot about it :). I used long since in general fold column could be defined by user and those values could be unexpectedly big ones( they might not be continuous sequence of numbers that starts from zero). Felt safer with long. But I agree, int will be more than enough in most cases. Thanks!

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
michalkurka: @deil87 I didn't mean the default java hash code, that won't work. I was refering to an auto-generated hash code that IntelliJ will give you (or any other IDE). Eg.:

    @Override
    public int hashCode() {
      int result = _levelValue.hashCode();
      result = 31 * result + _foldValue.hashCode();
      return result;
    }

deil87:
@michalkurka I've taken multiNodeTest related changes from master and pushed with `BroadcastJoinDistributedTest` test to see how it will work.

michalkurka:
@deil87 sure, you can now rebase on master - I merged the changes for the multinode testing

deil87:
@michalkurka sorry for confusion. That is actually how I already have taken these changes. Looking into logs now.

michalkurka:
do one more rebase - there is an issue with Java 7 right now - we disabled the step for now

deil87:
@michalkurka ok, I see that there is some SSL issues.

deil87:
@michalkurka For some reasons I can't see stages graph in Blue Ocean UI.

deil87:
@michalkurka Probably just waits for available executor.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,168 @@
...
+          assert foldValueFromVec <= Integer.MAX_VALUE && foldValueFromVec >= 0 : FOLD_VALUE_IS_OUT_OF_RANGE_MSG;
+          foldValue = (int) foldValueFromVec;
+        }
+        _encodingDataMapPerNode.put(new CompositeLookupKey(factor, foldValue), new EncodingData(numeratorChunk.at8(i), denominatorChunk.at8(i)));
+      }
+    }
+
+    @Override
+    public void reduce(FrameWithEncodingDataToHashMap mrt) {
+      _encodingDataMapFinal.putAll(mrt.getEncodingDataMap());
```
michalkurka: this will work but it is better to do 

_encodingDataMapPerNode.putAll(mrt.getEncodingDataMap())

because that minimizes necessary data transfer between nodes - _encodingDataMapPerNode is serialized as well as the MRTask result so data is partially duplicated in _encodingDataMapPerNode and _encodingDataMapFinal

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,168 @@
...
+        String factor = categoricalChunk.vec().factor(levelValue);
+        
+        int foldValue = -1;
+        if(_foldColumnIdx != -1) {
+          long foldValueFromVec = cs[_foldColumnIdx].at8(i);
+          assert foldValueFromVec <= Integer.MAX_VALUE && foldValueFromVec >= 0 : FOLD_VALUE_IS_OUT_OF_RANGE_MSG;
+          foldValue = (int) foldValueFromVec;
+        }
+        
+        CompositeLookupKey lookupKey = new CompositeLookupKey(factor, foldValue);
```
michalkurka: We usually avoid allocating new objects on a per-row level if it is not necessary. This could be done by mutating CompositeLookupKey. Not necessary just a remark ;)

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: @michalkurka Could you please help me to understand it better.... Reduce phase will be performed on a single node once all map tasks are finished and sent back to a reduce node. It means that map nodes are sending only empty serialised hashmaps of `_encodingDataMapFinal` to the reducer. After reduce phase is completed we will end up with all data in `_encodingDataMapFinal` hashmap and some data in '_encodingDataMapPerNode" - and it is a partial duplicate of the data. But I don't understand where would we be sending unnesessary data between nodes during MapReduce. 
Did you mean those empty serialised maps are going to be send around or something else? Or maybe you meant that once reduce phase is finished and final map is filled it will be redistributed across all nodes? 
Actually I would like to  know if we wiat for all maps to finish before starting doing something (shuffling, reducing) ?

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: @michalkurka  echoes of my Scala background -> generating tons of objects is not a problem as long as everything is immutable. But you are right, I will rewrite it into a mutable version.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: @michalkurka I also found a bug in the test and probably in the code. Reduce method was not being called at all. Trying to understand why. I can't wrap my head around the comment of the MRTasl.reduce method. Is says `Reduce is optional if, e.g., the result is some output vector`.

michalkurka: MRTask are always serialized as a whole, there is no difference between input and output - if you have 2 structures with duplicate data both structures will be serialized and send back - which is not necessary

in the hypothetical scenario of 2 nodes: the second node will do all the local reduces then serialized the MRTask send it to the first node and that will call the final reduce - because you have 2 structures populated in your task it will in worst case double the amount of data that need to be transferred

reduce is called always when there are actually data to reduce on, if the map is not applied reduce will not be called

deil87: @michalkurka I think that in 1 node scenario reduce method is not supposed to be called and that is another reason not to use separate hashMap for final aggregation -  it would not be filled up in single node scenario bc reduce would not be called. With my distributed test... I think that setting size of cloud to 2 does not mean that task will be distributed, I need to force it to do it somehow. Maybe with layout or size of the frame.

michalkurka: reduce will be called even in a single node scenario but you need at least 2 chunks - if there is nothing to reduce on (just a single map() invocation) it won't be called

you are right in your assessment that this is also a good reason not to have the second "final" structure

deil87: I meant that it can have multiple reduces on 1 node but it is not splitting into multiple tasks in my test scenario.

deil87: @michalkurka got it, thanks! I pushed commit with fixes. All single node tests are working but my `BroadcastJoinDistributedTest` is failing - I don't understand why. Maybe It is because of the improper layout ( but I tried different ones). I set it in a way that sum of them is equal to the number of rows. I can see in debug mode that reduce is being called twice as expected but in one call map is empty.

deil87: @michalkurka have you had a chance to look into  ^^?  I was not able to see if it is passing or not on Jenkins as Junit tests support for AutoML that you have added is still not working(last time I checked January 30th). I was pushing Michal Raska month ago but this stuff was out of priority .... but I think we need to get back to this.

deil87:
@michalkurka what you can suggest me to do with FrameTest.testNonEmptyChunks  test that I had  mentioned before? https://github.com/h2oai/h2o-3/pull/3233#issuecomment-458887736

deil87:
@seb-h2o Even though you have already approved this PR I want to ask you to help me get answers to some open questions

deil87:
@michal-raska as we still don't have properly set up JUnit tests for automl module I will have to merge this PR as is as I've got an approval from @michalkurka . But I would really like to have those tests' opinions as well.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,184 @@
...
+import water.fvec.Frame;
+import water.util.DistributedException;
+import water.util.IcedHashMap;
+
+import java.util.Objects;
+
+class BroadcastJoinForTargetEncoder {
+  
+  static class CompositeLookupKey extends Iced {
+    // Note: Consider using indexes of levels instead of strings to save space. It will probably require
```
michalkurka: I agree with this comment. It would be more H2O-like approach to represent the values by their indices in the domain.

You also don't need this structure, you can have a Map per-fold instead of one big Map. When categoricals are represented as indices, this map can be implemented using a plain java array.

If the Vec is cached on the node, the domain doesn't have to be sent over the wire. Having a structure that doesn't need to represent the domain thus saves possibly a lot of memory for large domains.

Same applies to `-1` fold - if you have this structure per fold then the memory is used more efficiently and much less data is sent over the wire.

I would suggest making this change.

deil87: @michalkurka I have changed representation and store everything in arrays now. Locally single/multinode tests are fine.... waiting for Jenkins. Could you please take a look at the implementation? The only thing that bothers me is a reduce phase when we have to merge arrays. It is O(m*2n), where `m` is number of categorical levels and `n` is number of folds. In most cases number of folds is not that big so it is probably fine.

deil87: @michalkurka I've tested one stress scenario and time performance is almost the same. I'm going to run scenario that was causing OOM and see if approach with arrays improved situation.

deil87:
@michalkurka What about these changes for upcoming release?

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,185 @@
...
+      Chunk denominatorChunk = cs[_denominatorIdx];
+      for (int i = 0; i < categoricalChunk.len(); i++) {
+        int levelValue = (int) categoricalChunk.at8(i);
+        int[] arrForNumerators = null;
+        int[] arrForDenominators = null;
+
+        if (_foldColumnIdx != -1) {
+          long foldValueFromVec = cs[_foldColumnIdx].at8(i);
+          if (!(foldValueFromVec <= Integer.MAX_VALUE && foldValueFromVec > 0)) {
+            throw new DistributedException(new AssertionError(FOLD_VALUE_IS_OUT_OF_RANGE_MSG + " but was " + foldValueFromVec));
```
michalkurka: use `assert` instead

also: this doesn't have to be checked for every single row - just check the min-max before you actually submit the task

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,185 @@
...
+      Chunk categoricalChunk = cs[_categoricalColumnIdx];
+      int numOfVecs = cs.length;
+      Chunk num = cs[numOfVecs - 2];
+      Chunk den = cs[numOfVecs - 1];
+      for (int i = 0; i < num.len(); i++) {
+        int levelValue = (int) categoricalChunk.at8(i);
+        int mappedLevelValue = -1;
+        try {
+          mappedLevelValue = _levelMappings[0][levelValue];
+        } catch (Exception ex) {
```
michalkurka: catching ArrayIndexOutOfBoundsException can certainly be avoided

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,185 @@
...
+          Log.info("Categorical level with index `" + levelValue + "` was not present in training data. ");
```
michalkurka: logging events on a row-level should be avoided, it will have an adverse effect on performance and might pollute logs and render them unusable

h2o-automl/src/test/java/ai/h2o/automl/targetencoding/TargetEncodingDistributedTest.java
```diff
@@ -0,0 +1,116 @@
...
+public class TargetEncodingDistributedTest extends TestUtil {
+
+  @BeforeClass
+  public static void setup() {
+    stall_till_cloudsize(1);
+  }
+
+  private Frame fr = null;
+
+  /*@Test
```
deil87: Going to consider to move tests from automl module to ext_te module as a next step

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,189 @@
...
+    public void map(Chunk[] cs) {
+      Chunk categoricalChunk = cs[_categoricalColumnIdx];
+      Chunk numeratorChunk = cs[_numeratorIdx];
+      Chunk denominatorChunk = cs[_denominatorIdx];
+      for (int i = 0; i < categoricalChunk.len(); i++) {
+        int levelValue = (int) categoricalChunk.at8(i);
+        int[] arrForNumerators = null;
+        int[] arrForDenominators = null;
+
+        synchronized (_encodingDataPerNode) { // Sync on whole array as there is no way to sync on two objects ( two rows of 2D array)
```
deil87: Going to consider to rewrite this. If we were storing numerators and denominators in one row we would be able to sync on single row of the 2D array. But for that we will have to "share" row and maximum number of categorical levels will be around (2^31-1) / 2 instead of (2^31-1). Which is probably quite fine.

deil87:
@michalkurka I have improved my previous solution by storing numerator and denomination in the same array(row). This way we can `synchronize` only on a single row without blocking whole 2D array . Tests went fine. In last commit I've just removed old empty test suites that were merged into another test suites.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,174 @@
...
+      Chunk numeratorChunk = cs[_numeratorIdx];
+      Chunk denominatorChunk = cs[_denominatorIdx];
+      for (int i = 0; i < categoricalChunk.len(); i++) {
+        int levelValue = (int) categoricalChunk.at8(i);
+
+          // We are allowed to do casting to `int` as we have validation before submitting this MRTask
+          int foldValue = _foldColumnIdx != -1 ? (int) cs[_foldColumnIdx].at8(i) : 0;
+          int[] arrForNumeratorsAndDenominators = _encodingDataPerNode[foldValue];
+
+        synchronized (arrForNumeratorsAndDenominators) { // Sync on whole array as there is no way to sync on two objects ( two rows of 2D array)
```
seb-h2o: @deil87 trying to figure out why this would be necessary: modifying the same value looks a bit alien to the map-reduce approach. Here you don't care if 2 parallel maps are modifying the `_encodingDataPerNode` 2d array concurrently as soon as they're not modifying the same cell.
But I don't see how it should be possible here as you never read the previous content of the cell. I'm missing something?

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,174 @@
...
+      }
+    }
+
+    @Override
+    public void reduce(FrameWithEncodingDataToArray mrt) {
+      int[][] leftArr = getEncodingDataArray();
+      int[][] rightArr = mrt.getEncodingDataArray();
+      for(int rowIdx = 0 ; rowIdx < leftArr.length; rowIdx++) {
+        for(int colIdx = 0 ; colIdx < leftArr[rowIdx].length; colIdx++) {
+          synchronized (leftArr[rowIdx]) {
```
seb-h2o: same here, reduce is single threaded and no other thread should access `_encodingDataPerNode` at that point: all maps are completed.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,174 @@
...
+    int _categoricalColumnIdx, _foldColumnIdx, _cardinalityOfCatCol;
+    int[][] _encodingDataArray;
+    int[][] _levelMappings;
+
+    BroadcastJoiner(int[] categoricalColumnsIdxs, int foldColumnIdx, int[][] encodingDataMap, int[][] levelMappings, int cardinalityOfCatCol) {
+      assert categoricalColumnsIdxs.length == 1 : "Only single column target encoding(i.e. one categorical column is used to produce its encodings) is supported for now";
+
+      _categoricalColumnIdx = categoricalColumnsIdxs[0];
+      _foldColumnIdx = foldColumnIdx;
+      _encodingDataArray = encodingDataMap;
```
seb-h2o: name change confused me when reading code below

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: @seb-h2o first thing... I forgot to remove comment and it is outdated. 

> modifying the same ref looks a bit alien to the map-reduce approach

In a multinode scenario refs will be different. 

> But I don't see how it should be possible here as you never read the previous content of the cell. I'm missing something?

Here we are reading row/array
```
int[] arrForNumeratorsAndDenominators = _encodingDataPerNode[foldValue];
```
Here we are modifying
```
          arrForNumeratorsAndDenominators[levelValue] = (int) numeratorChunk.at8(i);
          arrForNumeratorsAndDenominators[_cardinalityOfCatCol + levelValue] = (int) denominatorChunk.at8(i);
```
So I think there is an opportunity for race condition for the maps that are being run concurrently on the same mode.  
Another question is whether we operate on the same data( cells) as you have said. And it looks like we are operating with just the cells of the array for a given  fold and we should be fine even without sync. But I have test that would fail without this sync block(if you run this test multiple times) - `TargetEncoderBuilderTest.transform_KFold_scenario`. 

I think that the reason for race condition is that reduce phase might be intertwined with map phase and as we read and write in reduce method we need sync there as well. Do we wait for all maps to finish and only then run reduce phase? From my understanding( debugging) we are not waiting.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: @seb-h2o Are you sure about single thread? I think it depends on how many maps we have and we can reduce in parallel as well. At least it is my current understanding.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,174 @@
...
+class BroadcastJoinForTargetEncoder {
+
+  static class FrameWithEncodingDataToArray extends MRTask<FrameWithEncodingDataToArray> {
+    
+    /**
+     * Numerators and denominators are being stored in the same row-array of the 2D array.
+     * 
+     *        _encodingDataPerNode[k][0 .. numOfCatLevels -1] will be storing numerators
+     *        _encodingDataPerNode[k][numOfCatLevels .. (2*numOfCatLevels -1)] will be storing denominators,
+     *
```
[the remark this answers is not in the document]
deil87: @seb-h2o I think it is....no..no...no :) I'm not going to use this word again :)  Values could be located in many different ways I think. Don't think it makes sense to change current approach.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: @seb-h2o Yes, it is confusing. Will fix.

h2o-extensions/target-encoder/src/test/java/ai/h2o/targetencoding/TEMojoIntegrationTest.java
```diff
@@ -71,6 +71,7 @@ public void withoutBlending() throws PredictException, IOException {
       targetEncoderBuilder.trainModel().get();
 
       targetEncoderModel = targetEncoderBuilder.getTargetEncoderModel();
+      Scope.track_generic(targetEncoderModel);
```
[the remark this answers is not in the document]
deil87: @seb-h2o yeah. this one is useful!

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
seb-h2o: @deil87 ref was a bad wording sorry, I meant instance, and anyway `synchronized` is no use in multinode.
Yes, I noticed the global read/write, but what for me would have justified the `synchronized` is sth like:
```
arrForNumeratorsAndDenominators[levelValue] += (int) numeratorChunk.at8(i);
```

> Do we wait for all maps to finish and only then run reduce phase? From my understanding( debugging) we are not waiting.

You're right, F/J applies reduce for every fork, no wait.
The sync is necessary here not because of the `map` (we both agree that only one thread can change a given cell), but because reduce is modifying this 2d array in-place.
If your `FrameWithEncodingDataToArray` has 2 fields, one for map, and one for the result of `reduce`, you wouldn't need the `sync`: I'd be curious about the performance impact

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
seb-h2o: I was wrong, answered in previous comment.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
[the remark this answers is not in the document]
seb-h2o: np, was just wondering if you considered it. As I said, the advantage of even/odd approach is that you don't need to carry around the split point to use the array.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: @seb-h2o 
> ref was a bad wording sorry, I meant instance, and anyway

I think that both refs and instances will be different. We fill this array(or better these arrays) on different nodes independently.

> and anyway synchronized is no use in multinode.

I know that `synchronized` is not working across nodes because it can't work through serialisation/deserialisation processes. But here `synchronized` is needed(it is not actually needed if we use another approach - see below) when we operate within one node and, as I mentioned before, our `map` and `reduce` methods are getting intertwined. So reduce in multinode is  being done both locally and between instances from different nodes 

> If your FrameWithEncodingDataToArray had 2 encodingData fields, one for map, and one for the result of reduce, you wouldn't need the sync: I'd be curious about the performance impact

We would need to have twice as much memory and for cases when we have really high cardinalities and also many fold values this will be quite memory expensive. 

Gotcha! I think I come up with a better solution. Instead of all synchronisations or extra arrays 
we can do following in reduce method:
```
if (leftArr != rightArr) {
            int valueFromLeftArr = leftArr[rowIdx][colIdx];
            int valueFromRIghtArr = rightArr[rowIdx][colIdx];
            leftArr[rowIdx][colIdx] = Math.max(valueFromLeftArr, valueFromRIghtArr);
}
```
as when we are reducing lets call it within node we don't need to merge arrays at all, and we will not have no concurrency issues as we don't do anything

seb-h2o: @deil87 I think we have a misunderstanding with the ref/instance thing: of course I know they should be different, but apparently they weren't always different, otherwise you wouldn't have needed the `synchronized` lock.
Now can you please explain your fix?
If you add a condition `if (leftArr != rightArr)`, it's because you suggest that they could be equal... which is exactly what I was pointing at.
M/R should try to avoid that imo.

Regarding the memory usage, yes, my suggestion would temporarily double the memory usage for the 2D array, but once the reduction is done, you can reassign it to the result array.
There's always a trade-off between memory and perf, in this case, I would happily favour perf against memory. You have a 2D int array, |cat| * 2 * |folds| * 4 bytes < 1 MB for 10 folds, 10000 categories (before, you were creating 3*100000 locks for the same map/reduce operation).

Actually, I looked into details at the `MRTask` impl, and I didn't expect it to (shallow) clone the original instance for both left and right map operations, I guess that's why you can have `leftArr == rightArr` in the reduce operation (I find this to be very against M/R model, which is supposed to avoid those kind of conflicts).
And then, your latest change makes sense somehow: as the result is shared by lhs and rhs within node, then the reduction is actually done during mapping... not very M/Rish but efficient.

Btw, if you know in advance that `if (leftArr != rightArr)`, you can as well move the condition before the loop, and completely skip the reduce within node, it would be more clear.
Also please add comment explaining why: this really works because of the implementation details of `MRTask`. Usually, when reducing to `Frame`, this is handled by the `MRTask.map(Chunk, NewChunk)` methods. I believe that reducing to arrays is unusual in h2o-3, that's why those races arose.

seb-h2o: Also, as you said that categorical cardinality can be high (for folds, I don't think users have enough time in their life to use TE with 100 xval), maybe you should use `MemoryManager.malloc` then when creating your arrays.

deil87: @seb-h2o 
> I think we have a misunderstanding with the ref/instance thing

Yes, there was a misunderstanding. I'm sorry. Now I see what you were referring to.

> M/R should try to avoid that imo

Agree, maps should be pure and don't share same instance. In given use case we don't have conflicts during map phase as we work with different cells and there might not be any collisions. But due to an ability to reduce before all maps are done we can have race conditions. Current design of MRTask has its benefits of saving memory during computations....but we pay for it with higher carefulness during coding. It is always the case when you share something :)

> There's always a trade-off between memory and perf, in this case, I would happily favour perf against memory

I would not favour any of them. As it is called "trade-off" it by definition always depends on the situation. It is not number of folds that might be high, most likely it is number of categorical levels. I had mentioned two terms because even having relatively small number of folds it affects total value significantly due to a multiplication operation between them. Toy example: 0.5Gb for cat. levels and having 10 folds will make it 5Gb.

> Actually, I looked into details at the MRTask impl, and I didn't expect it to (shallow) clone the original instance for both left and right map operations, I guess that's why you can have `leftArr == rightArr` in the reduce operation (I find this to be very against M/R model, which is supposed to avoid those kind of conflicts).

Yes, I also had a look into implementation details of our MR framework. With shallow copy we can't do check `leftArr != rightArr` before  `reduce` is being called  as MRTask instances are always different but internal fields referencing the same instances(in case of MRTasks from the same node) and we don't know particular fields of MRTask of course => we can only check inside reduce method. 

>  then the reduction is actually done during mapping... not very M/Rish but efficient.

 Yes, feels like this. 

> Btw, if you know in advance that if (leftArr != rightArr), you can as well move the condition before the loop, and completely skip the reduce within node, it would be more clear.

Yes, good catch! 

> I believe that reducing to arrays is unusual in h2o-3, that's why those races arose.

I think that it is not about arrays. It is about mutable vs immutable approach. If you do it in immutable way with new NewChunk `MRTask.map(Chunk, NewChunk)` then no race conditions. If you do `MRTask.map(Chunk[])` then RCs are very welcome.
I have seen reducing to arrays in AstGroup functionality. Actually if you take a look at the comment for `GBTaskAggsPerNode.reduce` method you will see quite similar story. 

> MemoryManager.malloc then when creating your arrays

I was considering to use `MemoryManager` but there were no method for 2D int array. But once you are also suggesting this I've decided to add one. Btw, there is a code duplication between `MemoryManager.malloc8d` and `CoxPHUtils.malloc2DArray`. Also `CoxPHUtils.malloc3DArray` maybe could be moved to `MemoryManager` so we can get rid of `CoxPHUtils`.

Thank you for the discussion! I think It was a good opportunity for both of us to learn more about our MRTask framework.

deil87: @seb-h2o added comment for check in reduce method. Let me know if it is not clear. Also created separate PR for malloc4 https://github.com/h2oai/h2o-3/pull/3994/files

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
```diff
@@ -0,0 +1,178 @@
...
+      }
+    }
+
+    @Override
+    public void reduce(FrameWithEncodingDataToArray mrt) {
+      int[][] leftArr = getEncodingDataArray();
+      int[][] rightArr = mrt.getEncodingDataArray();
+      
+      // Note: we need to add this check due to the following circumstances: 
+      //       1) there is a chance that reduce phase will start before map phase is finished.
```
seb-h2o: Thanks, let's go into more details for future unknown reader:
For a given MR task T: we have T -> (Tl, Tr), and let's say Tl -> (Tll, Tlr).
Tll, Tlr `map` is always completed before Tl `reduce`, and T `reduce` needs to wait for completion of both Tl and Tr.
Tll and Tlr manipulate the same array, but not the same cells -> no race condition.
The race arises because Tl `reduce` (of both Tll and Tlr) can occur in parallel with Tr `map` (and `reduce`).
Once both Tl and Tr are completed, T `reduce` itself is safe.

Don't know how to summarize this, but it's far from being intuitive, so I think it's worth writing it explicitly.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
seb-h2o: @deil87 definitely useful conversation, thanks.

Maybe for future improvement, I'm still wondering if there's a way to avoiding passing the full 2D array to the forked tasks. It probably requires support at `MRTask` level in a similar way to what is done with `MRTask.map(Chunk, NewChunk)` though...

> I think that it is not about arrays. It is about mutable vs immutable approach

Yeah, totally agree, I just meant that if reducing to arrays was more common, someone would have implemented at `MRTask` level to avoid those races, like mentioned above.

michalkurka: @deil87 @seb-h2o 

Synchronization is IMHO not needed here, I have a simple "proof" :) data of `arrForNumeratorsAndDenominators` are never read in this map() they are only being written. Even if it was was somehow important in what order will the write-only operations be executed, synchronization doesn't address that because the order of execution of the independent map tasks is arbitrary.

A single h2o node can have a shared instance of this array, 2 different nodes will just need to do a reduce, the reduce step doesn't need synchronization either. Reduce step should only be executed if `this != mrt` (that is when we are working with a result of remote node).

seb-h2o: @michalkurka FYI the code you're reviewing is not up to date, the `sync` was removed after discussion.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: @seb-h2o 
> Don't know how to summarize this, but it's far from being intuitive, so I think it's worth writing it explicitly.

Agree. I just wonder whether these details should be here... it is actually quite general information and might be useful even outside of the BroadcastJoin scope. wdyt?

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
michalkurka: @seb-h2o okay, thank you for fixing that

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/BroadcastJoinForTargetEncoder.java
deil87: @seb-h2o added more details to the comment to make it more explicit

seb-h2o: fantastic, thanks

deil87:
Tests are fine. Going to merge it tomorrow if there is no objections

PR #3238, opened 2019-01-21
PUBDEV-6188 - Dynamic S3 credentials

Pscheidl:
## Description

https://0xdata.atlassian.net/browse/PUBDEV-6188

Once user credentials are set for Amazon S3 (in environment variables, configuration file etc.), there is no way to change them (currently). This PR makes it possible to change them while H2O is still running.

## How it is done

- New client is created each time credentials are changed
- All caches are invalidated once the new client is created
- Better locking mechanism - introduced timeout. The credentials change require the S3 client variable and caches not to be used. However, there might be a long-running task (e.g. cache update of a large bucket). Right now, there is a timeout of 60 seconds, otherwise it would block for long. This value is just a proposal.

### R

```R
    h2o.setS3Credentials(accessKeyId, accesSecretKey)
    [1] "Credentials successfully set."
```

### Python
```python
from h2o.persist import set_s3_credentials
set_s3_credentials(access_key_id, secret_access_key)
> Credentials successfully set.
```


h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -6,6 +6,7 @@
 import com.amazonaws.Protocol;
 import com.amazonaws.auth.*;
 import com.amazonaws.auth.profile.ProfileCredentialsProvider;
+import com.amazonaws.internal.StaticCredentialsProvider;
```
michalkurka: internal? is this safe to use? seems deprecated in more recent version of AWS libs

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -35,13 +39,41 @@
...
   private static final String KEY_PREFIX = "s3://";
   private static final int KEY_PREFIX_LEN = KEY_PREFIX.length();
+  private static final int LOCK_TIMEOUT_SEC = 60;
 
-  private static final Object _lock = new Object();
+  private static final Lock _lock = new ReentrantLock();
+  
   private static volatile AmazonS3 _s3;
 
+  public static void changeClientCredentials(final String accesKeyId, final String accessSecretKey) {
```
michalkurka: hmm, can we just throw out an existing instance of PersistS3 and start with a new one? keeping this consistent seems error-prone

just thinking out loud, don't take it too seriously

michalkurka: another thought: all the locking seems to be done for Cache - if we separate all the caching to a new object, we can just swap the object to a new instance and avoid any locking

I believe there is a simpler approach that doesn't require locking

ledell:
@Pscheidl I see that we have two functions in R that use camelCase with `h2o.setSomething()` (we typically used camel case in the beginning of the project in R and snake case in Python always) but in the past few years we have just tried to name the functions in an identical way unless there is a strong objection/reason not to.  I am curious if anyone else has an opinion on this, but I would lean toward naming the function `h2o.set_s3_credentials()`.

Pscheidl:
@ledell Let's not change the standards and do it the way you say. Converted to snake case in the latest commit.

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
Pscheidl: You're right. This one was super easy to replace with our own implementation. So I did it :) 

Just a sidenote: The whole API as we use it is deprecated in newer versions.

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
Pscheidl: There is no need to have locking mechanism at all in this method. This was my mistake. Once the cache is updated, it has a local copy of the pointer to the S3 client unless the operation is finished. Any subsequent  calls to this method will NOT trigger the update. Which means if the client is swapped in the middle of an update, it will still finished properly.

Having the caches in separate  class doesn't help anything. Not without proper cache deprecation mechanism, which in fact requires checks on current amazon S3 client or event-based deprecation. Which again requires some degree of locking.

This method creates short-lived caches. Those are linked to a bucket and a bucket is linked to credentials. Changing credentials mostly means accessing a different bucket (typical use case). This renders the old cache obsolete, as it is meant to be short-lived and only prevent DoS attack on Amazon while the user is typing. This means it is perfectly OK to delete the caches once new client is created.

**So I completely removed the lock.**

There is one corner case when the user hits `Import file` button in Flow and changes S3 credentials between the file search phase and hitting the `Import file` button. This can be hardly prevent and the error is very clear:

snimek z 2019-01-22 11-12-39

Pscheidl: We definitely can not, that is why there is some degree of locking introduced. The timeout is there for a simple reason: The call must fail if the new client is not created withing a reasonable period of time, so Flow or other clients don't get stuck on a single call.

Pscheidl:
Please note the original cache updating is not thread safe anyway. The whole class does not solve all the problems anyway.

For example `HashMap<String, KeyCache> _keyCaches` is potentially concurrently modified in `calcTypeaheadMatches`, as this method might be called very quickly and `_keyCaches.put(parts[0], new KeyCache(parts[0]));` operation might be called from several threads at once. This also means it might run the `update` operation twice under some very rare circumstances. This is all original code.

I propose not to focus on this right now. This class needs a major rewrite with proper caching and multiple client handling. I'd do it once we migrate to the new AWS client, as the API we use right now is long deprecated.

WDYT @michalkurka ?

Pscheidl:
I did the credentials change process via MRTask. The values are red dynamically by our custom credentials provider placed into the AWS credential provider chain.

For now, the values reside in memory on each node. I completely avoided DKV for this task.

However, I do think current solution has flaws. Some origin from flaws/tradeoffs made in current implementation.

Made a little helper painting for us to continue discussion personally.
[RB2019-01-25 12.13.05.pdf](https://github.com/h2oai/h2o-3/files/2795798/RB2019-01-25.12.13.05.pdf)

h2o-persist-s3/src/main/java/water/persist/PersistS3Handler.java
```diff
@@ -0,0 +1,33 @@
...
+        
+        new S3CredentialsMRTask(s3Credentials.secret_key_id, s3Credentials.secret_access_key).doAllNodes();
+        return s3Credentials;
+    }
+
+    /**
+     * Checks for basic mistakes users might make when providing S3 credentials.
+     * @param s3Credentials S3 credentials provided by the user
+     */
+    private void validateS3Credentials(final PersistS3CredentialsV3 s3Credentials){
```
Pscheidl: Do not log credentials !

Pscheidl:
`setS3Credentials_fail` test is expect to fail, as the credentials might get picked up from the environment.

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -65,11 +69,47 @@ public static AmazonS3 getClient() {
...
+
+  /**
+   * Holds basic credentials (Secret key ID + Secret access key) pair.
+   */
+  private static final class H2ODynamicCredentialsProvider implements AWSCredentialsProvider {
+
+
+    @Override
+    public AWSCredentials getCredentials() {
+      BufferedString keyId = DKV.getGet(S3_SECRET_KEY_ID_DKV_KEY);
```
michalkurka: Sorry for nitpicking but it would be better to store this in a single object. To guarantee consistency.

Pscheidl:
API style remains the same. We already got it approved by @ledell in the previous PR which I cancelled. :)

Pscheidl:
There are test failures.

### PersistS3HandlerTest

Noticed some weird behavior with cloud status. Looks like it complains the cloud size is under 5, but the test only requires cloud size of 1 (PeristS3Test does proper testing of all the functionality on cloud of size 5)

```
01-29 01:01:24.803 192.168.254.5:44008 24220 FJ-126-15 WARN: Killing /192.168.254.5:44000 because the cloud is no longer accepting new H2O nodes.
01-29 01:01:24.824 192.168.254.5:44008 24220 FJ-126-15 WARN: Killing /192.168.254.5:44004 because the cloud is no longer accepting new H2O nodes.
01-29 01:01:24.836 192.168.254.5:44008 24220 FJ-126-15 WARN: Killing /192.168.254.5:44006 because the cloud is no longer accepting new H2O nodes.
```

### PersistS3Test

There is a stacktrace:

```
Stacktrace
com.amazonaws.AmazonClientException: No Amazon S3 credentials set directly.
at water.persist.PersistS3Test.testS3Import()
``` 

This is an exception commonly thrown by amazon credential providers. Looks like it gets logged and Jenkins thinks it failed . This should be fixed.

Re-running the tests in the meantime.

PR #3239, opened 2019-01-21
Enable multinode testing on OS X

michalkurka:


michalkurka:
@seb-h2o please check the fix in WorkAllocations (tests were failing because of use of LinkedList inside of an Iced class)

PR #3240, opened 2019-01-21
Adds a new stage for AutoML Junit tests

michalkurka:
Fixes multinode issues


michalkurka:
@seb-h2o please check the fix in `WorkAllocations` (tests were failing because of use of LinkedList inside of an Iced class)

deil87:
@michalkurka great, thank you!

PR #3241, opened 2019-01-22
PUBDEV-6216 - implement hive delegation token generation for HIVE on …

honzasterba:
…hadoop with kerberos

### DIFF
--- a/h2o-hadoop-2/h2o-mapreduce-2-generic/build.gradle
+++ b/h2o-hadoop-2/h2o-mapreduce-2-generic/build.gradle
@@ -1,6 +1,7 @@
 ext {
   hadoopVersion = 'cdh5.2'
   hadoopMavenArtifactVersion = '2.5.0-cdh5.2.0'
+  hiveMavenArtifactVersion = '0.13.1-cdh5.2.0'
 }
 
 apply plugin: 'java'
@@ -15,7 +16,8 @@ compileJava {
 }
 
 dependencies {
-  compile('org.apache.hadoop:hadoop-client:' + hadoopMavenArtifactVersion)
+  compileOnly "org.apache.hive:hive-jdbc:$hiveMavenArtifactVersion"
+  compile "org.apache.hadoop:hadoop-client:$hadoopMavenArtifactVersion"
   compile(project(':h2o-app')) {
     exclude module: "${defaultWebserverModule}"
   }
--- a/h2o-hadoop-2/h2o-mapreduce-2-generic/src/main/java/water/hadoop/h2odriver.java
+++ b/h2o-hadoop-2/h2o-mapreduce-2-generic/src/main/java/water/hadoop/h2odriver.java
@@ -150,6 +150,8 @@ public class h2odriver extends Configured implements Tool {
   static String keytabPath = null;
   static boolean reportHostname = false;
   static boolean driverDebug = false;
+  static String hiveHost = null;
+  static String hivePrincipal = null;    
 
   String proxyUrl = null;
   // Runtime state that might be touched by different threads.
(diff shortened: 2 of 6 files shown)


h2o-hadoop-3/h2o-mapreduce-3-generic/build.gradle
```diff
@@ -1,6 +1,7 @@
 ext {
   hadoopVersion = 'cdh6.0.1'
   hadoopMavenArtifactVersion = '3.0.0-cdh6.0.1'
+  hiveMavenArtifactVersion = '2.1.1-cdh6.0.1'
```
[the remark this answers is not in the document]
honzasterba: this is compileOnly it does not get bundled with the actual driver and is needed only to have HiveConf and HiveConnection classes at compile time - so even though there is different version of the class at runtime, it should be fine since the only method used is HiveConnection#getDelegationToken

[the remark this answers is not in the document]
michal-raska: Oh, didn't see that it's compileOnly. Thanks.

h2o-hadoop-2/h2o-mapreduce-2-generic/src/main/java/water/hadoop/h2odriver.java
```diff
@@ -1821,6 +1830,8 @@ private JobWrapper submitJob(Configuration conf) throws Exception {
     j.setOutputKeyClass(Text.class);
     j.setOutputValueClass(Text.class);
 
+    new HiveTokenGenerator(driverDebug).addHiveDelegationToken(j, hiveHost, hivePrincipal);
```
michalkurka: I would probably leave it on - at least for now. Otherwise, it doesn't generate any output.

h2o-hadoop-2/h2o-mapreduce-2-generic/build.gradle
```diff
@@ -1,6 +1,7 @@
 ext {
   hadoopVersion = 'cdh5.2'
```
michalkurka: I have no idea why we have CDH 5.2 here. Seems incorrect to me.

h2o-hadoop-2/h2o-mapreduce-2-generic/build.gradle
```diff
@@ -1,6 +1,7 @@
...
   hadoopMavenArtifactVersion = '2.5.0-cdh5.2.0'
+  hiveMavenArtifactVersion = '0.13.1-cdh5.2.0'
```
michalkurka: I guess this is inspired by above ^^^ hadoop version. I think the issue is that `hadoopMavenArtifactVersion` is overridden in each assembly but `hiveMavenArtifactVersion` is defined just here. We should override the definition on the assembly level for each supported distribution.

h2o-hadoop-2/h2o-mapreduce-2-generic/src/main/java/water/hadoop/h2odriver.java
```diff
@@ -630,6 +632,7 @@ static void usage() {
                     "          [-jobname <name of job in jobtracker (defaults to: 'H2O_nnnnn')>]\n" +
                     "              (Note nnnnn is chosen randomly to produce a unique name)\n" +
                     "          [-principal <kerberos principal> -keytab <keytab path> [-run_as_user <impersonated hadoop username>] | -run_as_user <hadoop username>]\n" +
+                    "          [-hiveHost <hostname:port> -hivePrincipal <hive server principal>]\n" +
```
michalkurka: Not sure if we have to put it here. Not all options are documented. If we do include it in the help we should mark it as experimental (it could change).

h2o-hadoop-2/h2o-mapreduce-2-generic/build.gradle
honzasterba: I dont think one project can override ext in another so these stay exactly the same and this project is always compiled against these libs

h2o-hadoop-2/h2o-mapreduce-2-generic/build.gradle
honzasterba: agree, but thats for another PR

h2o-hadoop-2/h2o-mapreduce-2-generic/src/main/java/water/hadoop/h2odriver.java
honzasterba: it can be enabled by -driver_debug so we dont need to have it always on

h2o-hadoop-2/h2o-mapreduce-2-generic/src/main/java/water/hadoop/h2odriver.java
honzasterba: removed

h2o-hadoop-2/h2o-mapreduce-2-generic/build.gradle
michalkurka: It used to work that way (until last December)... if it doesn't anymore I think that might be an issue. Let me check

PR #3251, opened 2019-01-24
[ITA-369] Gradle 5.1.1.

michal-raska:
Test Gradle 5.1.1 and let's see if there are any issues.

Bump images version to 9.


michalkurka:
Interesting... gradle 5.1.1 ingores my setting in `~/.gradle/gradle.properties`:

    org.gradle.java.home=/Library/Java/JavaVirtualMachines/jdk1.8.0_101.jdk/Contents/Home

and fails because I have Java 7 as the system Java

michal-raska:
@michalkurka is it ok to merge to `rel-xu`, or should I merge this to `master` only?

PR #3252, opened 2019-01-24
PUBDEV-6186 - handle NoClassDefFoundError when intializing JDBC connection

honzasterba:
- this improves user exprience when using thin hive jdbc driver jar


michalkurka:
@honzasterba should this go also to rel-xu?

PR #3254, opened 2019-01-25
Make SQL ingest including setup fully asynchronous

michalkurka:
+ improve progress reporting - Job will have info about what
  is happening on the backend

### DIFF
--- a/h2o-core/src/main/java/water/jdbc/SQLManager.java
+++ b/h2o-core/src/main/java/water/jdbc/SQLManager.java
@@ -37,11 +37,11 @@ public class SQLManager {
    * @param username (Input)
    * @param password (Input)
    * @param columns (Input)
-   * @param sqlFetchMode (Input)
+   * @param fetch_mode (Input)
    */
   public static Job<Frame> importSqlTable(final String connection_url, final String table, final String select_query,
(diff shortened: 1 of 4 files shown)


honzasterba:
do we want this in rel xu? if so rebasing and mergibg will be problematic

h2o-core/src/main/java/water/jdbc/SQLManager.java
```diff
@@ -38,207 +37,244 @@
...
 
-  private static Connection getConnectionSafe(String url, String username, String password) throws SQLException {
-    try {
-      return DriverManager.getConnection(url, username, password);
-    } catch (NoClassDefFoundError e) {
-      throw new RuntimeException("Failed to get database connection, probably due to using thin jdbc driver jar.", e);
+      DKV.put(fr);
+      ParseDataset.logParseResults(fr);
+      if (importTable.equals(SQLManager.TEMP_TABLE_NAME))
+        dropTempTable(_connection_url, _username, _password);
```
seb-h2o: I thought this drop was done in a finally clause at some point: as it is not a true temporary table, it's important to have this always executed otherwise it prevents the user from running imports later, even if s/he fixed the driver issue for example.

PR #3258, opened 2019-01-28
Demo tests to show inconsistency between TestFrameBuilder and TestUtil.parse_test_file

deil87:


michalkurka:
By adding a Vec with a different layout to your frame you cause the original frame to be rebalanced.

Your synthetic frame has a single chunk layout and so does the constant Vec you create, nothing is rebalanced and nothing leaks.

In case of your parsed frame the layout is different, and you are forcing the parsed frame to rebalance. The original chunks then leak.

Rebalancing a frame is an expensive operation and should be avoided. I suggest to fix your code so that it doesn't trigger a rebalance.

deil87:
@michalkurka could you please take a look at my last comment? I wanted to close the PR but it is still interesting to try to get answers. @seb-h2o did you run into this rebalancing stuff? wdyt?

seb-h2o:
@deil87 agree that we should make it clear (javadoc and/or higher-level methods semantic) when this triggers or may trigger rebalancing: it's not obvious currently from method signature and devs can't look deep into the code each time they need to use an existing method.

PR #3259, opened 2019-01-28
[PROPOSAL] Demo tests to highlight unexpected behaviour of the ShuffleSplitFrame.shuffleSplitFrame

deil87:
https://0xdata.atlassian.net/browse/PUBDEV-6241

### DIFF
--- a/h2o-algos/src/test/java/hex/splitframe/ShuffleSplitFrameTest.java
+++ b/h2o-algos/src/test/java/hex/splitframe/ShuffleSplitFrameTest.java
@@ -6,14 +6,12 @@ import org.junit.Test;
 import water.Key;
 import water.MRTask;
 import water.TestUtil;
-import water.fvec.Chunk;
-import water.fvec.Frame;
-import water.fvec.NewChunk;
-import water.fvec.Vec;
+import water.fvec.*;
 import water.parser.BufferedString;
 
 import java.util.Arrays;
 
+import static org.junit.Assert.assertEquals;
 import static water.fvec.FrameTestUtil.createFrame;
 import static water.fvec.FrameTestUtil.collectS;
 import static water.util.ArrayUtils.flat;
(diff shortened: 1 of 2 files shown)


michalkurka:
By adding a Vec with a different layout to your frame you cause the original frame to be rebalanced.

Your synthetic frame has a single chunk layout and so does the constant Vec you create, nothing is rebalanced and nothing leaks.

In case of your parsed frame the layout is different, and you are forcing the parsed frame to rebalance. The original chunks then leak.

Rebalancing a frame is an expensive operation and should be avoided. I suggest to fix your code so that it doesn't trigger a rebalance.

h2o-algos/src/test/java/hex/splitframe/ShuffleSplitFrameTest.java
```diff
@@ -115,5 +113,82 @@ static void testScenario(Frame f, String[] expValues, MRTask chunkAssertions) {
...
+              .withDataForCol(0, ar("a", "b"))
+              .withDataForCol(1, ar("yellow", "blue"))
+              .withDataForCol(2, ar("2", "6"))
+              .build();
+
+      double[] splitRatio = {0.9, 0.1};
+      Key[] keysForSplits = {Key.make("key_" + leaderboardFrame._key), Key.make("key2_" + leaderboardFrame._key)};
+      Frame[] splits = ShuffleSplitFrame.shuffleSplitFrame(leaderboardFrame, keysForSplits, splitRatio, 1234);
+      System.out.println("Number of row in original frame: " + leaderboardFrame.numRows());
+      assertEquals(1, splits[0].numRows());
```
michalkurka: What do you propose the result should be?

deil87: Lets say for the edge case of 2 rows and ratios [0.9, 0.1]  I would expect two splits with one row in each because it is I believe a higher priority to fulfil implicit user's request of getting 2 non-empty splits . But the main point is that we should not end up with empty split when let's say I'm splitting 6-rows dataset with [0.8, 0.2] ratios. In current implementation it is quite often the case.

PR #3262, opened 2019-01-29
updating gbm reproducibility document

meganjkurka:


angela0xdata:
I just pushed a couple of changes. Here's what that FAQ looks like when built. Let me know if you think this is okay.

screen shot 2019-01-29 at 1 37 24 pm

meganjkurka:
Thank you @angela0xdata that looks great to me! Would it be possible to update the docs website with this change once it's approved or would we need to wait until a new release?

angela0xdata:
@meganjkurka I'm not able to build h2o-3 docs. They're tied to a release. But after the merge, it will be available in the docs that are built for each nightly build release. 
Can you also cherry-pick into rel-xu?

maurever:
I like the changes; it is more clear for me now.

PR #3263, opened 2019-01-30
[BUILD] Add AutoML files to test-package-java.zip

michal-raska:
Adds the required files for AutoML tests to `test-package-java.zip`.


deil87:
@michal-raska Thank you!

michal-raska:
@deil87 does this output looks ok please? I'm not sure how long the tests are expected to run:

```
> Task :h2o-automl:testMultiNode
Running h2o-automl junit tests...
h2o-automl junit tests PASSED
:h2o-automl:testMultiNode took 27.244 secs

> Task :h2o-automl:testSingleNode
Running h2o-automl junit tests...
h2o-automl junit tests PASSED
:h2o-automl:testSingleNode took 3.265 secs
```

deil87:
@michal-raska in the snippet that I've sent you before
```
> Task :h2o-automl:testMultiNode
./testMultiNode.sh: line 101: cd: src/test/java: No such file or directory
Running h2o-automl junit tests...
h2o-automl junit tests PASSED
:h2o-automl:testMultiNode took 10.702 secs

> Task :h2o-automl:testSingleNode
./testSingleNode.sh: line 94: cd: src/test/java: No such file or directory
Running h2o-automl junit tests...
h2o-automl junit tests PASSED
```
it was only 10 seconds for multiNode and who knows how much for a single node. At least some positive changes could be seen :) But to be honest it should take longer I think. Can we increase verbosity of the logging?

deil87:
@michal-raska can we check ``cat $OUTDIR/tests.txt` this location? 
`OUTDIR=sandbox/multi`

deil87:
@michal-raska I can't see files here http://mr-0xc1:8080/blue/organizations/jenkins/h2o-3-pipeline/detail/PR-3233/10/artifacts

deil87:
@michal-raska I ssh as you suggested and I can see that `tests.txt` has all the tests ( even more then expected). Some of them definitely should have failed... going to look into other artifacts.

deil87:
@michal-raska this PR is about adding everything to test package...so yes. But I need to figure out why for example `ai.h2o.automl.targetencoding.TargetEncodingTitanicBenchmark` got into the list of tests. Because the is no suffix `Test` or `Suite` or whatever we are using for filtering tests out of the other files(if we do) in it.

deil87:
@michal-raska and probably tests are not being run, at least `tests.txt` does not represent the tests that are actually being run.

michal-raska:
@deil87 ok, so let's keep this open, I'll check that as well.

deil87:
@michal-raska from files `out.0` and `out.1` I can say that only one test class was executed -> `ai.h2o.automl.targetencoding.TargetEncodingExceptionsHandlingTest` with 6 subtests in it

deil87:
@michal-raska let's proceed with this one?

PR #3270, opened 2019-02-01
adding seed information to reproducibility doc

meganjkurka:
Spoke with @michalkurka and the seed does not need to be explicitly set at time of training for the model to reproducible.  The seed will be randomly assigned and saved in the binary model.

I've updated the document to reflect this information.


h2o-docs/src/product/data-science/gbm-faq/reproducibility.rst
```diff
@@ -64,3 +62,33 @@ Reproducibility
    
    - Refer to :ref:`save_and_load_model` for more information.
    - The binary model will contain the H2O version and the parameters used to train the model.
+
+ 4. Train the GBM model using the same:
+	
+   - training data (information about the training data saved in Step 1)
+   - cluster configuration (information about the cluster configuration saved in Step 2)
+   - same parameters (information about the parameters saved in Step 3)
```
angela0xdata: I think you can remove "same" from this bullet:
> same parameters (information about the parameters saved in Step 3)

meganjkurka: thanks for the catch! Just fixed it

maurever:
After the discussion with @meganjkurka, it is waiting for https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6262.

meganjkurka:
@maurever I added a section on reproducing the model if CPU's was different.  I was not able to figure out how to do this in Python so I only have an R example.  The function `h2o.parse_raw` in Python does not have the `chunk_size` argument.

angela0xdata:
@meganjkurka @maurever I made some updates to the formatting. This looks good when built locally.

maurever:
I finally solved and merged the problem with getting seed in R (https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6262). So I edit the part of the documentation here to reflect that. In R there the new method `h2o.get_seed(model)` to get seed correctly now. A user can also use `model@parameters$seed,` however, this could be unsafe in the future, so use the `h2o.get_seed(model)` is the proper way.

I also edited the part about using chunk size in Python. It is working now. 

I think it is ready to merge before release. :)

PR #3272, opened 2019-02-01
PUBDEV-6183: Doc fix for Tweedie deviance equation

angela0xdata:
- Updated the equation for deviance in the Tweedie subsection in the user guide and GLM booklet
- p=3 is Inverse-Gaussian instead of Gaussian in the UG
- Fixed character issue in GLM booklet


laurendiperna:
looks good! thanks for making these updates!

PR #3273, opened 2019-02-01
PUBDEV-6254: AutoML basic support for blending

seb-h2o:
https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6254

Exposing blending frame in AutoML..: expected behaviour when blending frame is provided, is that we don't enforce some CV-related params necessary for SE otherwise, and SE are still computed if nfolds=0 as we don't use cv predictions to train the SEs in this case.

When CV is disabled (common use-case when using blending for time series), we now force the creation of a validation frame to avoid overfitting of the base models.
Also made small changes to leaderboard so that it can do scoring based on validation metrics then, and not only by scoring the leaderboard frame, making this latter one potentially optional also internally.

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
@@ -295,6 +295,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
         Log.warn("Failed to find base model; skipping: " + k);
         continue;
       }
+      Log.debug("Checking properties for model " + k);
       if (!aModel.isSupervised()) {
         throw new H2OIllegalArgumentException("Base model is not supervised: " + aModel._key.toString());
       }
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
@@ -135,6 +135,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
       
       StackedEnsemble.addModelPredictionsToLevelOneFrame(base, basePreds, levelOneFrame);
       DKV.remove(basePreds._key); //Cleanup
+      Frame.deleteTempFrameAndItsNonSharedVecs(basePreds, levelOneFrame);
     }
 
     // Add response column to level one frame
@@ -162,6 +163,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
       ModelMetrics mmStackedEnsemble = lastComputedMetric.deepCloneWithDifferentModelAndFrame(this, fr);
       this.addModelMetrics(mmStackedEnsemble);
     }
+    Frame.deleteTempFrameAndItsNonSharedVecs(levelOneFrame, adaptFrm);
     return predictFr;
   }
 
@@ -182,7 +184,8 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
 
   ModelMetrics doScoreMetricsOneFrame(Frame frame, Job job) {
       Frame pred = this.predictScoreImpl(frame, new Frame(frame), null, job, true, CFuncRef.from(_parms._custom_metric_func));
-      pred.remove();
+//      Frame pred = this.score(frame, null, job, true,  CFuncRef.from(_parms._custom_metric_func));
+      pred.delete();
       return ModelMetrics.getFromDKV(this, frame);
   }
 
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -1131,20 +1131,17 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     pollAndUpdateProgress(Stage.ModelTraining, "DeepLearning hyperparameter search 3", work, this.job(), dlJob);
   }
 
-  Job<StackedEnsembleModel> stack(String modelName, Key<Model>[]... modelKeyArrays) {
+  Job<StackedEnsembleModel> stack(String modelName, Key<Model>[] modelKeyArrays, boolean use_cache) {
     WorkAllocations.Work work = workAllocations.getAllocation(Algo.StackedEnsemble, JobType.ModelBuild);
     if (work == null) return null;
 
-    List<Key<Model>> allModelKeys = new ArrayList<>();
-    for (Key<Model>[] modelKeyArray : modelKeyArrays)
-      allModelKeys.addAll(Arrays.asList(modelKeyArray));
     // Set up Stacked Ensemble
     StackedEnsembleParameters stackedEnsembleParameters = new StackedEnsembleParameters();
-    stackedEnsembleParameters._base_models = allModelKeys.toArray(new Key[0]);
+    stackedEnsembleParameters._base_models = modelKeyArrays;
     stackedEnsembleParameters._valid = (getValidationFrame() == null ? null : getValidationFrame()._key);
     stackedEnsembleParameters._blending = (getBlendingFrame() == null ? null : getBlendingFrame()._key);
     stackedEnsembleParameters._keep_levelone_frame = true; //TODO Why is this true? Can be optionally turned off
-    stackedEnsembleParameters._keep_base_model_predictions = true; //avoids recomputing some base predictions for each SE
+    stackedEnsembleParameters._keep_base_model_predictions = use_cache; //avoids recomputing some base predictions for each SE
     // Add cross-validation args
     if (buildSpec.input_spec.fold_column != null) {
       stackedEnsembleParameters._metalearner_fold_column = buildSpec.input_spec.fold_column;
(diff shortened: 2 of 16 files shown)


h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -1302,8 +1298,11 @@ public void learn() {
       for (int i = 0; i < bestModelsOfEachType.size(); i++)
         bestModelKeys[i] = bestModelsOfEachType.get(i)._key;
 
-      Job<StackedEnsembleModel> bestEnsembleJob = stack("StackedEnsemble_BestOfFamily", bestModelKeys);
+      Job<StackedEnsembleModel> bestEnsembleJob = stack("StackedEnsemble_BestOfFamily", bestModelKeys, true);
       pollAndUpdateProgress(Stage.ModelTraining, "StackedEnsemble build using top model from each algorithm type", seWork, this.job(), bestEnsembleJob, true);
+      
+      Job<StackedEnsembleModel> ensembleJob = stack("StackedEnsemble_AllModels", notEnsembles, false);
+      pollAndUpdateProgress(Stage.ModelTraining, "StackedEnsemble build using all AutoML models", seWork, this.job(), ensembleJob, true);
```
seb-h2o: @ledell is it a concern if I swap training order between BestOf and AllModels?
I do this to avoid having to recompute some of the predictions twice as `BestOfModels ⊂ AllModels`.
As SE are not time-bound, I don't see any inconvenient in doing this (just saying as AllModels is usually better than bestof so we definitely don't want to skip it).

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -135,6 +135,7 @@ protected Frame predictScoreImpl(Frame fr, Frame adaptFrm, String destination_ke
       
       StackedEnsemble.addModelPredictionsToLevelOneFrame(base, basePreds, levelOneFrame);
       DKV.remove(basePreds._key); //Cleanup
+      Frame.deleteTempFrameAndItsNonSharedVecs(basePreds, levelOneFrame);
```
seb-h2o: this omission is not due to refactoring nor blending, it was just not tested correctly: discovered that we were leaking keys when scoring SEs only with the AutoML test...
same below

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
ledell: No, i don't have any issue with it and it sounds like a nice little optimization.

seb-h2o:
no one to have a look/approve? thx @ledell , @navdeep-G , @michalkurka

ledell:
@seb-h2o Something fishy going on here because there are not enough models showing up when I run it with `blending_frame` and `nfolds = 0`.  Let's see if we can track down what's happening here.

```
library(h2o)

h2o.init()

y <- "response"
x <- setdiff(names(train), y)

test[,y] <- as.factor(test[,y])

# Create a blending frame and leaderboard frame
ss <- h2o.splitFrame(test, ratios = 0.5, seed = 1)
bl_frame <- ss[[1]]
lb_frame <- ss[[2]]

# Run AutoML for 5 minutes
aml <- h2o.automl(x = x, y = y, project_name = "aml_cv",
                  training_frame = train,
                  leaderboard_frame = lb_frame,
                  max_runtime = 5*60,
                  seed = 1)

aml@leaderboard
# model_id       auc   logloss mean_per_class_error      rmse       mse
# 1    StackedEnsemble_AllModels_AutoML_20190221_131918 0.8006533 0.5419577            0.2875455 0.4267136 0.1820845
# 2 StackedEnsemble_BestOfFamily_AutoML_20190221_131918 0.8005042 0.5414494            0.2916961 0.4266881 0.1820627
# 3                    XGBoost_3_AutoML_20190221_131918 0.7952302 0.5484351            0.2879172 0.4303396 0.1851922
# 4                    XGBoost_1_AutoML_20190221_131918 0.7947747 0.5505018            0.2884810 0.4309972 0.1857586
# 5                    XGBoost_2_AutoML_20190221_131918 0.7946405 0.5492224            0.2960475 0.4304096 0.1852524
# 6                        GBM_5_AutoML_20190221_131918 0.7942125 0.5498891            0.3060453 0.4306719 0.1854783
# 
# [32 rows x 6 columns]


# Run AutoML for 5 minutes
aml3 <- h2o.automl(x = x, y = y, project_name = "aml_blend_nfolds0",
                   training_frame = train,
                   blending_frame = bl_frame,
                   leaderboard_frame = lb_frame,
                   nfolds = 0,
                   max_runtime = 5*60,
                   seed = 1)

aml3@leaderboard
# model_id       auc   logloss mean_per_class_error      rmse       mse
# 1    StackedEnsemble_AllModels_AutoML_20190221_134302 0.7996554 0.5440628            0.2837578 0.4276189 0.1828579
# 2 StackedEnsemble_BestOfFamily_AutoML_20190221_134302 0.7996554 0.5440628            0.2837578 0.4276189 0.1828579
# 3                    XGBoost_1_AutoML_20190221_134302 0.7857660 0.5918340            0.3037245 0.4423410 0.1956656
# 4                        XRT_1_AutoML_20190221_134302 0.7786058 0.5745400            0.3219016 0.4417962 0.1951839
# 5                        DRF_1_AutoML_20190221_134302 0.7780982 0.5746432            0.3246467 0.4420167 0.1953787
# 6           GLM_grid_1_AutoML_20190221_134302_model_1 0.6975841 0.6275564            0.3762506 0.4679069 0.2189369
# 
# [6 rows x 6 columns] 
```

seb-h2o:
@ledell the validation issue is fixed now (was not specific to blending actually, it's just that we don't usually disable CV):
for reference, explanation of your issue above
```
XGBoost_1 was overfitting with nfolds=0, monopolizing all the time left.... so we need to split training frame in this case and ensure all base models are trained with a validation frame.
```
I fixed it by extracting a validation frame from training frame when CV disabled if former was missing.
Please note that leaderboard frame doesn't look necessary at all anymore, and we should think about removing it (future PR please :) as this is not really related to blending either).

Also took the responsibility to reorganize the code a bit in AutoML: this is a large class and I always waste a lot of time navigating there, so now it looks like so:
------
static stuff
------
constructor with base public functions: run, stop, high level logic
-------
data preparation
-------
generic sub jobs logic: create, configure, run, poll
-------
default models definitions
-------
cleanup + misc utility methods
-------
at least now , there is some logic in the code layout and I'm much more comfortable navigating in it, I hope you'll agree.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -279,73 +330,197 @@ WorkAllocations planWork() {
...
+        splitRatios = new double[]{ 0.9, 0.1, 0 };
+        userFeedback.info(Stage.DataImport,
+            "Since cross-validation is disabled, and no validation frame was provided, " +
+                "automatically split the training data into training and validation frames in the ratio 90/10");
+      } else if (null == this.leaderboardFrame) {
+        splitRatios = new double[]{ 0.9, 0, 0.1 };
+        userFeedback.info(Stage.DataImport,
+            "Since cross-validation is disabled, and no leaderboard frame was provided, " +
+                "automatically split the training data into training and leaderboard frames in the ratio 90/10");
+      }
```
seb-h2o: @ledell please double check ratios

h2o-py/h2o/automl/autoh2o.py
```diff
@@ -275,6 +275,9 @@ def train(self, x = None, y = None, training_frame = None, fold_column = None,
         :param leaderboard_frame: H2OFrame with test data for scoring the leaderboard.  This is optional and
             if this is set to None (the default), then cross-validation metrics will be used to generate the leaderboard 
             rankings instead.
+        :param blending_frame: H2OFrame used to train the Stacked Ensembles in the absence of cross-validated base models.
```
```suggestion
        :param blending_frame: H2OFrame used to train the the metalearning algorithm in Stacked Ensembles (can use with nfolds = 0).
```

h2o-r/h2o-package/R/automl.R
```diff
@@ -16,6 +16,8 @@
 #'        validation_frame will be ignored.
 #' @param leaderboard_frame Leaderboard frame (H2OFrame or ID); Optional.  If provided, the Leaderboard will be scored using 
 #'        this data frame intead of using cross-validation metrics, which is the default.
+#' @param blending_frame Blending frame (H2OFrame or ID) used to train the Stacked Ensembles in the absence of cross-validated base models; Optional.
```
```suggestion
#' @param blending_frame Blending frame (H2OFrame or ID) used to train the the metalearning algorithm in Stacked Ensembles (instead of relying on cross-validated predicted values); Optional.
```

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
```diff
@@ -75,6 +75,9 @@ public AutoMLBuildControlV99() {
     @API(help = "ID of the validation data frame (used for early stopping in grid searches and for early stopping of the AutoML process itself).", direction=API.Direction.INPUT)
     public KeyV3.FrameKeyV3 validation_frame;
 
+    @API(help = "ID of the blending frame used to train the Stacked Ensembles in the absence of cross-validated base models. When provided, it is also recommended to disable cross validation by setting `nfolds=0` and to provide a leaderboard frame for scoring purposes.", direction = API.Direction.INPUT)
```
ledell:
```suggestion
    @API(help = "ID of the H2OFrame used to train the the metalearning algorithm in Stacked Ensembles (instead of relying on cross-validated predicted values). When provided, it is also recommended to disable cross validation by setting `nfolds=0` and to provide a leaderboard frame for scoring purposes.", direction = API.Direction.INPUT)
```

ledell:
@seb-h2o I think the one thing that is still missing here is the updates to automl.rst User Guide file.  The section which talks about automatic splitting should be updated.  Do you want to do that or me?

seb-h2o:
@ledell you mean the "Notes" section? Agree, we should probably add that we're also extracting a validation frame when CV is disabled. Feel free to edit :)
Also maybe the property descriptions themselves need to be updated: I can read this for `leaderboard_frame` for example: 
> then a leaderboard frame will be generated automatically from the validation frame (if provided) 

"from validation frame"? maybe it was the case in the past, but not anymore... should have I done it this way?

a last concern: in `xu` the missing validation_frame is still a possibility. Should I apply a fix there as well?

ledell:
@seb-h2o Yep, in the "Notes" section.  I will also update to acknowledge the validation frame update.  Regarding the the leaderboard frame definition w/r/t no CV -- this is how we used to do it (at least that's what's documented).  My thought behind breaking off the leaderboard frame from validation instead of training, is that the validation data is less "important" than the training data.  What do you think?

ledell:
@seb-h2o Follow-up user guide PR: https://github.com/h2oai/h2o-3/pull/3319

PR #3277, opened 2019-02-05
[HEXDEV-720] Hive MetaStore connector

honzasterba:
- use hive MetaStoreClient to get table metadata and provide correct column names and types


h2o-core/src/main/java/water/api/ImportHiveTableHandler.java
```diff
@@ -0,0 +1,25 @@
...
+    String DEFAULT_DATABASE = "default";
+    
+    String ALL_COLUMNS = "*";
+
+    ParseV3 loadHiveTable(String database, String tableName, String columns) throws Exception;
+
+  }
+  
+  public ParseV3 importHiveTable(int version, ImportHiveTableV99 request) throws Exception {
+    HiveTableImporter importer = (HiveTableImporter) Class.forName(IMPORTER_CLASS).newInstance();
```
honzasterba: please tell me there is a better way (I copied this pattern from PersistManager.java)

michalkurka: better way would be SPI, or better making hive an extension, see ExtensionManager

honzasterba:
added support for partition filter
 current issue with partitioned tables is that partition-key columns are not present in the data so they are not imported

h2o-assemblies/main/build.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
     compile project(":h2o-persist-s3")
     compile project(":h2o-persist-http")
     compile project(":h2o-persist-hdfs")
+    compile project(":h2o-hive")
```
michalkurka: Do we want the support for hive in standalone as well? What does it do to dependencies? Are we going to be compatible with any version of Hive?

I would rather start just with Hadoop support.

h2o-core/src/main/java/water/fvec/Vec.java
```diff
@@ -606,6 +606,7 @@ public static Vec makeVec(long [] vals, String [] domain, Key<Vec> vecKey){
    *  and initialized to the given constant value.  */
   public Vec makeCon(final double d) { return makeCon(d, group(), _rowLayout, T_NUM); }
   public Vec makeCon(final double d, byte type) { return makeCon(d, group(), _rowLayout, type); }
+  public Vec makeCon(final long l, byte type) { return makeCon(l, null, group(), _rowLayout, type); }
```
michalkurka: what is the difference?

h2o-assemblies/main/build.gradle
honzasterba: good idea, will move it

h2o-core/src/main/java/water/fvec/Vec.java
honzasterba: there was no way how to create a String vec with the same layout as the current vec

h2o-core/src/main/java/water/parser/ParseSetup.java
```diff
@@ -40,6 +41,9 @@
   String[][] _na_strings;       // Strings for NA in a given column
   String[][] _data;           // First few rows of parsed/tokenized data
   int[] _parse_columns_indices; // store column indices to be parsed into the final file
+  
+  String[] _synthetic_column_names;
```
michalkurka: please explain in a comment the purpose of these 2 fileds

ParseSetup is Iced and `_synthetic_column_values` don't seem to be iced-serializable

h2o-core/src/main/java/water/parser/SyntheticColumnGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+import water.Job;
+import water.Key;
+import water.MRTask;
+import water.fvec.Chunk;
+import water.fvec.Frame;
+import water.fvec.Vec;
+
+public class SyntheticColumnGenerator extends MRTask<SyntheticColumnGenerator> {
+  
+  public static Frame addSyntheticColumns(Job<Frame> job, Vec[] parsedVecs, ParseSetup setup, Key[] fkeys, int[] fileChunkOffsets) {
```
michalkurka: I would like to have this decoupled from ParseDataset 

this call is confusing to the reader:

      - fr = new Frame(job._result, setup._column_names, AppendableVec.closeAll(avs));
      + fr = SyntheticColumnGenerator.addSyntheticColumns(job, AppendableVec.closeAll(avs), setup, fkeys, mfpt._fileChunkOffsets);

because it in the common case only creates the frame without any synthetics columns

you can have an interface, called eg. ParseFinalizer that would have have 2 implementations based on `setup._synthetic_column_names` with a factory-method somewhere

h2o-core/src/main/java/water/parser/SyntheticColumnGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+    _fileChunkOffsets = fileChunkOffsets;
+    _fkeys = fkeys;
+  }
+
+  @Override
+  public void map(Chunk[] cs) {
+    int synColCnt = _setup._synthetic_column_names.length;
+    for (int colIdx = 0; colIdx < synColCnt; colIdx++) {
+      String fkey = findFileForChunk(cs[0].cidx());
+      String colValue = _setup._synthetic_column_values.get(fkey)[colIdx];
```
michalkurka: was this tested multi-node? `_synthetic_column_values` don't seem Freezable to me

h2o-hadoop-common/tests/python/pyunit_gbm_on_hive.py
```diff
@@ -30,24 +30,19 @@ def gbm_on_hive():
 
     # read from S3
     airlines_dataset_original = h2o.import_file(path="https://s3.amazonaws.com/h2o-public-test-data/smalldata/airlines/AirlinesTest.csv.zip")
-    # read from Hive Distributed
```
michalkurka: why is this deleted? I think it should stay

honzasterba: moved to a separate test

h2o-core/src/main/java/water/parser/ParseSetup.java
honzasterba: converted to array

h2o-core/src/main/java/water/fvec/Vec.java
michalkurka: I don't think this works in the general case... if your goal is to create a constant NA columns with String please add a method without the `long` parameter, eg. `makeConStr(byte type)`

this creates a Vec made of `C0DChunks`

please also add a test that shows that rollups can be calculated over such column

honzasterba: I think this will work. What you are suggesting is pretty much what 

public Vec makeCon(final double d, byte type) 
https://github.com/h2oai/h2o-3/pull/3277/files/3957805df8a43235c18228836d37a0b6111d6410#diff-8087560966ac7843cf17ceb4410c0be2L607

does, but inside is a check
if( (long)d==d ) return makeCon((long)d, null, group, rowLayout, type)

so it ends up creating C0LChunk chunks, so we should be fine.

I have changed the API to hide the 0L inside the call. Added a test as well.

PR #3279, opened 2019-02-06
[BUILD] Fix git diff check.

michal-raska:


michal-raska:
Looks like the issue is caused by something in reference repos:

```
+--Git Diff Check--+

if [[ $(git diff) ]]; then \

	echo "Non-zero git diff after running gradlew build:"; \

	git diff; \

	exit 1; \

fi

error: object directory /home/jenkins/repos/h2o-3/.git/objects does not exist; check .git/objects/info/alternates.

fatal: unable to read 01f6778e5aaead14a51e403eee6f9f6b4e63d23a
```

michal-raska:
Docker mount of the reference folder was missing, therefore the failure in previous comment.

PR #3281, opened 2019-02-06
PUBDEV-6138: Add scikit-learn Integration topic to UG - WIP

ABartzGit:
Note that the import statement for RandomizedSearchCV has changed. Also, I'm getting errors when I run fit in both examples.


angela0xdata:
@laurendiperna, can you run the examples and see if you get errors? 
I also get the same error when I run the PCA test: https://github.com/h2oai/h2o-3/blob/master/h2o-py/tests/testdir_scikit_grid/pyunit_scal_pca_rf_grid.py

PR #3283, opened 2019-02-07
PUBDEV-6267 fix autoencoder mse error calculation

maurever:
Fix Autoencoder `calculateReconstructionErrorPerRowData` error and set the default value of the result MSE to -1.

JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6267


maurever:
@michalkurka I added the test, but it is just duplication of the [runit_PUBDEV_4628_deeplearning_mojo_large.R ](https://github.com/h2oai/h2o-3/blob/rel-xu/h2o-r/tests/testdir_javapredict/runit_PUBDEV_4628_deeplearning_mojo_large.R) test in case` params$standardize <- FALSE`. The PUBDEV_4628 test just randomly selects the parameters, that is why sometimes failed and sometimes not. Is better to create the new test or edit the old one?

PR #3290, opened 2019-02-12
HEXDEV-732: GLRM Memory Allocation Reduction

wendycwong:
This PR aims to solve this issue: https://0xdata.atlassian.net/projects/HEXDEV/issues/HEXDEV-732?filter=allopenissues .

There are a lot of array allocation at various places in the algo, I plan to get rid of them.

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/algos/glrm/GlrmMojoModel.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/algos/glrm/GlrmMojoModel.java
@@ -3,6 +3,7 @@ package hex.genmodel.algos.glrm;
 import hex.ModelCategory;
 import hex.genmodel.MojoModel;
 
+import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.Random;
 
@@ -32,6 +33,13 @@ public class GlrmMojoModel extends MojoModel {
   public boolean _reverse_transform;
   public double _accuracyEps = 1e-10; // reconstruction accuracy A=X*Y
   public int _iterNumber = 100; // maximum number of iterations to perform X update.
+  public double[] _grad;  // used in gradientL 
+  public double[] _tgrad; // used in calling glrmloss
+  public double[] _txy;
+  public double[] _tx;
+  public double[] _tu;
+  public double[] _ta;
+  public double[] _tbest;
 
   // We don't really care about regularization of Y since it is changed during scoring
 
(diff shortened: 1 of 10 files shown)


h2o-genmodel/src/main/java/hex/genmodel/algos/glrm/GlrmMojoModel.java
```diff
@@ -151,7 +174,9 @@ public GlrmMojoModel(String[] columns, String[][] domains, String responseColumn
    * @return
    */
   public double applyBestAlpha(double[] u, double[] x, double[] grad, double[] a, double oldObj, Random random) {
-    double[] bestX = new double[x.length];
+    //double[] bestX = new double[x.length];
```
Pscheidl: Remove this ?

h2o-genmodel/src/main/java/hex/genmodel/algos/glrm/GlrmMojoModel.java
```diff
@@ -151,7 +174,9 @@ public GlrmMojoModel(String[] columns, String[][] domains, String responseColumn
...
+    double[] bestX = _tbest;
```
Pscheidl: The original code creates new array which `bestX` variable points to. This piece of code together with `Arrays.fill(bestX, 0.0);` will make `bestX` point to the shared `_tbest` array and fill it with zeros. Is this intentional ?

Pscheidl: Not crucial though ...

h2o-genmodel/src/main/java/hex/genmodel/algos/glrm/GlrmMojoModel.java
```diff
@@ -278,24 +305,28 @@ public static int rank(boolean transposed, double[][] archetypes_raw) {
    */
   private double[] gradientL(double[] x, double[] a) {
     // Prepate output row
-    double[] grad = new double[_ncolX];
-
+    //double[] grad = new double[_ncolX];
```
Pscheidl: remove the old line ?

wendycwong:
Discussed with @pscheidl.  Getting rid of the  memory allocation may not gain us anything.  Let's do a test in Hadoop to find out if this thing helps with anything at all.  Please post timing information here before completing this PR.

wendycwong:
I have completed a timing test on mr-0xd2, 4 cluster with 40G each.  Pavel is right, I do not have a lot of gain in speed.  However, this fix is more to enable users to run without running out of memory.  Previous implementation performs a lot of new double[] operation.  I generate 3 enum columns with various cat levels and then perform GLRM on the synthetic frame.  Refer to the JIRA for a detail timing result.  Here, I am just adding the highlights:


cat level | glrm runtime gain (%)
3000      | 1.848860781
2000      | 2.685828768
1000       | 3.34021701

h2o-genmodel/src/main/java/hex/genmodel/algos/glrm/GlrmMojoModel.java
wendycwong: Done

h2o-genmodel/src/main/java/hex/genmodel/algos/glrm/GlrmMojoModel.java
wendycwong: Done

wendycwong:
Small improvement in memory.  Not worth the change for the time being.

PR #3291, opened 2019-02-12
[PUBDEV-6274] fix hive auth keytab distribution

honzasterba:
- distributed cache is not good to use since it requires the file to be on hdfs
- use our internal mechanism for distributing security files to mappers

### DIFF
--- a/h2o-mapreduce-generic/src/main/java/water/hadoop/DelegationTokenRefresher.java
+++ b/h2o-mapreduce-generic/src/main/java/water/hadoop/DelegationTokenRefresher.java
@@ -11,6 +11,7 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
+import java.io.File;
 import java.io.IOException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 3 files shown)


h2o-mapreduce-generic/src/main/java/water/hadoop/h2omapper.java
```diff
@@ -126,6 +127,19 @@ private static void repairNullArgsAndWarnIfNecessary(String[] args) {
     }
   }
   
+  static boolean makeSureIceRootExists(String iceRoot) {
+    File f = new File(iceRoot);
+    if (!f.exists()) {
+      boolean success = f.mkdirs();
+      if (!success) {
+        Log.POST(103, "mkdirs(" + f.toString() + ") failed");
```
michalkurka: These Log messages should probably stay in the original location because now the method is called from different context

h2o-mapreduce-generic/src/main/java/water/hadoop/DelegationTokenRefresher.java
```diff
@@ -20,37 +21,47 @@
 
 public class DelegationTokenRefresher implements Runnable {
   
-  public static final String KEYTAB_FILE = "hive_keytab";
-  private static final String KEYTAB_PATH = "./" + KEYTAB_FILE;
-
-  public static void setup(Configuration conf) {
+  public static void setup(Configuration conf, String iceRoot) throws IOException {
```
michalkurka: do we need to worry about creating the iceRoot? The only place where this is used is in the h2omapper which will make the directory for us - I would rather simplify and remove the call of `makeSureIceRootExists` + rename iceRoot to something else, eg. keytabDirectory

h2o-mapreduce-generic/src/main/java/water/hadoop/h2omapper.java
honzasterba: moved

h2o-mapreduce-generic/src/main/java/water/hadoop/DelegationTokenRefresher.java
honzasterba: ice_root is created only when any H2O_MAPPER_CONF_* vars are set, so yes, we need to make sure it exists

h2o-mapreduce-generic/src/main/java/water/hadoop/h2omapper.java
honzasterba: actually I have decided to keep them here because we want to have it behave exactly the same, and these logs will only appear once anyways (the first time this method is called.

PR #3293, opened 2019-02-13
XGBoost: Java backend can calculate approximate contributions

michalkurka:
Exposes approximate contributions to Java API audience.

These are not Tree SHAP values, these are values that are returned by XGBoost when `approx_contribs = TRUE`

### DIFF
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -91,6 +91,10 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
     Frame scoreStagedPredictions(Frame frame, Key<Frame> destination_key);
   }
 
+  public interface Contributions {
+    Frame scoreContributions(Frame frame, Key<Frame> destination_key);
+  }
+
   public interface ExemplarMembers {
     Frame scoreExemplarMembers(Key<Frame> destination_key, int exemplarIdx);
   }
--- a/h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostScoreTask.java
+++ b/h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostScoreTask.java
@@ -14,6 +14,8 @@ import static water.fvec.Vec.T_NUM;
 
 public class XGBoostScoreTask extends MRTask<XGBoostScoreTask> {
 
+    enum OutputType { PREDICT, PREDICT_CONTRIB_APPROX}
+    
     private final XGBoostModelInfo _sharedmodel;
     private final XGBoostOutput _output;
     private final XGBoostModel.XGBoostParameters _parms;
@@ -142,7 +144,8 @@ public class XGBoostScoreTask extends MRTask<XGBoostScoreTask> {
 
     private static ScoreResult scoreChunkExt(final XGBoostModelInfo sharedmodel, final XGBoostModel.XGBoostParameters parms,
                                              final BoosterParms boosterParms, final XGBoostOutput output,
-                                             final Frame fr, final Chunk[] cs) {
+                                             final Frame fr, final Chunk[] cs,
+                                             final OutputType outputType) {
         DMatrix data = null;
         Booster booster = null;
         try {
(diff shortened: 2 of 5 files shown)


h2o-core/src/main/java/hex/Model.java
```diff
@@ -91,6 +91,10 @@ public boolean filter(KeySnapshot.KeyInfo k) {
     Frame scoreStagedPredictions(Frame frame, Key<Frame> destination_key);
   }
 
+  public interface Contributions {
+    Frame scoreContributions(Frame frame, Key<Frame> destination_key);
```
Pscheidl: I like the idea to have a custom key for the contribution matrix.

Pscheidl: In future, we have the possibility to add `Frame scoreContributions(Frame frame);` method, which will create the key.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -91,6 +91,10 @@ public boolean filter(KeySnapshot.KeyInfo k) {
     Frame scoreStagedPredictions(Frame frame, Key<Frame> destination_key);
   }
 
+  public interface Contributions {
```
Pscheidl: Definitely a common interface for them all.

h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostScoreTask.java
```diff
@@ -14,6 +14,8 @@
 
 public class XGBoostScoreTask extends MRTask<XGBoostScoreTask> {
 
+    public enum OutputType { PREDICT, PREDICT_CONTRIB }
```
Pscheidl: Tiny detail: protected is enough.

h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostSetupTask.java
```diff
@@ -38,6 +40,15 @@ protected void execute() {
     final DMatrix matrix;
     try {
       matrix = makeLocalMatrix();
+      if (_parms._save_matrix_directory != null) {
+        File directory = new File(_parms._save_matrix_directory);
+        if (directory.mkdirs()) {
```
Pscheidl: This might return false if the directory already exists, but it will also return false if the directory could not be created. Something like (quickly written):

```java
        if(directory.exists()){
          // continue
        }
        else if (directory.mkdirs()) {
          Log.debug("Created directory for matrix export: " + directory.getAbsolutePath());
        } else{
          throw new IllegalStateException("Unable to create directory for matrix export: " + directory.getAbsolutePath());
        }
```

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostTest.java
```diff
@@ -1526,6 +1528,86 @@ public void testMakeDataInfo() {
     }
   }
 
+  @Test
+  public void testScoreContributions() throws IOException, XGBoostError {
```
Pscheidl: I'm missing end-to-end test that proves the new parameter works. E.g. in Python.

michalkurka: it is not exposed, internal at this point

h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostScoreTask.java
```diff
@@ -170,8 +173,17 @@ private static ScoreResult scoreChunkExt(final XGBoostModelInfo sharedmodel, fin
 
             // Predict
             ScoreResult result = new ScoreResult();
-            result._preds = booster.predict(data);
-            result._labels = data.getLabel();
+            switch (outputType) {
+                case PREDICT:
```
Pscheidl: Like the split, leads to maximum code reusal.

h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostSetupTask.java
michalkurka: hmm, I think it is okay :)

h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostScoreTask.java
michalkurka: Fixed

h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostSetupTask.java
michalkurka: Fixed

michalkurka:
@Pscheidl issues fixed

This is not meant to be exposed just yet. So it will not have python/R test

michalkurka:
@Pscheidl I will have to revert your suggestion of checking return value of `mkdirs` - this is a race condition on multi-node setups that are running on the same physical machine.

PR #3295, opened 2019-02-14
PUBDEV-6277: Make constant response check optional for all models

navdeep-G:
* Currently if a training frame has a constant response column, then an error is thrown based on the below snippet: 

```
        if (_response.isConst())
          error("_response","Response cannot be constant.");
```

* A good solution for this is to make the above check optional for instances where this behavior is not desired. 
* **NOTE: This is only exposed in Java**

### DIFF
--- a/h2o-core/src/main/java/hex/ModelBuilder.java
+++ b/h2o-core/src/main/java/hex/ModelBuilder.java
@@ -36,7 +36,7 @@ abstract public class ModelBuilder<M extends Model<M,P,O>, P extends Model.Param
    *  we can return built Model. */
   public final M get() { assert _job._result == _result; return _job.get(); }
   public final boolean isStopped() { return _job.isStopped(); }
-
+  
   // Key of the model being built; note that this is DIFFERENT from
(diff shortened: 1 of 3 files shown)


h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -822,6 +822,7 @@ public void clearInitState() {
   protected transient String[] _origNames; // only set if ModelBuilder.encodeFrameCategoricals() changes the training frame
   protected transient String[][] _origDomains; // only set if ModelBuilder.encodeFrameCategoricals() changes the training frame
 
+  public boolean _checkConstResponse(){return _parms._check_constant_response;}
```
michalkurka: please get rid off this method, just use the field directly just like here

    if (expensive && error_count() == 0 && _parms._auto_rebalance) {

michalkurka:
please include a test - eg. in DRF

navdeep-G:
@michalkurka I have added a test. Thx!

PR #3302, opened 2019-02-15
PUBDEV-6294: Expose check_constant_response for DRF and GBM

navdeep-G:
* Expose `check_constant_response` for DRF and GBM
* This variable is set to `true` by default, which is behavior today, i.e., thrown an exception if response column is a constant value
* Added simple unit tests in R/Python client which basically ensures GBM and DRF respect this argument if set to `false`

### DIFF
--- a/h2o-algos/src/main/java/hex/schemas/DRFV3.java
+++ b/h2o-algos/src/main/java/hex/schemas/DRFV3.java
@@ -68,5 +68,8 @@ public class DRFV3 extends SharedTreeV3<DRF, DRFV3, DRFV3.DRFParametersV3> {
 
         @API(help = "Row sample rate per tree (from 0.0 to 1.0)", gridable = true)
         public double sample_rate;
+
+        @API(help="Check for constant response", level = API.Level.expert, direction = API.Direction.INOUT)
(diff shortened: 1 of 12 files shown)


angela0xdata:
@navdeep-G Just an FYI that this needs to be changed in the docs and in Flow. I opened a Jira ticket for the docs (https://0xdata.atlassian.net/browse/PUBDEV-6287), and I can work on that when this is merged. Should the change in Flow be included in this PR or in a separate PR/Jira?

navdeep-G:
@angela0xdata I think Flow should be another ticket. However, I thought these changes will be picked up Flow automatically? Right @michalkurka @michal-raska ?

michal-raska:
@navdeep-G yes, it's there automatically in this case (because it's `boolean` and Flow knows how to handle this type). But for some more complex types (for example `KeyValue[]`) it might have to be added manually.

angela0xdata:
@michal-raska @navdeep-G ahh, I didn't know that about Booleans in Flow. Thanks for the info!

navdeep-G:
@michalkurka I think this PR is good to go. Agree?

PR #3303, opened 2019-02-18
updating sum documentation

meganjkurka:


meganjkurka:
@michalkurka how do I check that the documentation is okay? 

I tried `./gradlew build -x test` but the R documentation fails even without my change.

This is the output I see on master:
```
> Task :h2o-r:buildPackageDocumentation FAILED

FAILURE: Build failed with an exception.
```

angela0xdata:
@meganjkurka I'm not sure why you're hitting that error. In the h2o-r folder, I just ran `../gradlew build` and it worked fine. 
But I normally check the nightly build the day after merging and check the docs, just in case.

PR #3305, opened 2019-02-19
PUBDEV-6290: Fixing AutoML leader slot in R for single model

ledell:
- Fixed a bug on the AutoML R client described in [PUBDEV-6290](https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6290)
- Removed the single NA row for empty leaderboards in R
- Added warning for when there's an empty leaderboard
- The `h2o.getAutoML()` function was also not working for empty leaderboards in R (now fixed)
- Fixed unit test for empty leaderboard (was checking for 1 row, now checks for 0 rows)
- There are a lines changed because of deleted spaces (must have been introduced by my editor...)

### DIFF
--- a/h2o-r/h2o-package/R/automl.R
+++ b/h2o-r/h2o-package/R/automl.R
@@ -246,49 +246,13 @@ h2o.automl <- function(x, y, training_frame,
       params <- list(input_spec = input_spec, build_control = build_control, build_models = build_models)
   }
 
-  # POST call to AutoMLBuilder
+  # POST call to AutoMLBuilder (executes the AutoML job)
   res <- .h2o.__remoteSend(h2oRestApiVersion = 99, method = "POST", page = "AutoMLBuilder", autoML = TRUE, .params = params)
   .h2o.__waitOnJob(res$job$key$name)
 
-  # GET AutoML job and leaderboard for project
-  automl_job <- .h2o.__remoteSend(h2oRestApiVersion = 99, method = "GET", page = paste0("AutoML/", res$job$dest$name))
-  #project <- automl_job$project  # This is not functional right now, we can get project_name from user input instead
-  leaderboard <- as.data.frame(automl_job["leaderboard_table"]$leaderboard_table)
-  row.names(leaderboard) <- seq(nrow(leaderboard))
-
-
-  # Intentionally mask the progress bar here since showing multiple progress bars is confusing to users.
-  # If any failure happens, revert back to user's original setting for progress and display the error message.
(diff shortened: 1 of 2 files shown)


ledell:
Can test it out using this code here (doesn't really need to be a unit test though):

```
library(h2o)
h2o.init()

train <- h2o.importFile("http://s3.amazonaws.com/erin-data/higgs/higgs_train_10k.csv")
y <- "response"
x <- setdiff(names(train), y)

aml <- h2o.automl(x = x, y = y,
                  training_frame = train,
                  max_runtime_secs = 3,
                  seed = 1)
```
Now we have a zero row leaderboard (just like Python), and `h2o.getAutoML()` gives an error message now:
```
> aml@leaderboard
[1] model_id             auc                  logloss              mean_per_class_error rmse                
[6] mse                 
<0 rows> (or 0-length row.names)

[0 row x 6 columns] 
> aa <- h2o.getAutoML(aml@project_name)
Warning message:
In h2o.getAutoML(aml@project_name) :
  The leaderboard contains zero models: try running AutoML for longer (the default is 1 hour).
```

h2o-r/h2o-package/R/automl.R
```diff
@@ -266,15 +267,20 @@ h2o.automl <- function(x, y, training_frame,
...
+    leader <- h2o.getModel(automl_job$leaderboard$models[[1]]$name)
   } else {
-      # create a phony leader
-      Class <- paste0("H2OBinomialModel")
-      leader <- .newH2OModel(Class = Class,
-                             model_id = "dummy")
+    # create a phony leader
+    Class <- paste0("H2OBinomialModel")
+    leader <- .newH2OModel(Class = Class,
+                           model_id = "dummy")
```
seb-h2o: what prevents you from reusing `h2o.getAutoML` directly from L254? 
Would be better than having to duplicate code:
```
aml <- h2o.getAutoML(res$job$dest$name)
aml$project_name <- build_control$project_name
return(aml)
```

ledell:
@seb-h2o You're right, I should just wrap the code to retrieve and process the leaderboard and leader in function and call it twice.

h2o-r/h2o-package/R/automl.R
ledell: @seb-h2o Yeah you're right, I can just do that.  Thanks!

ledell: @seb-h2o I changed the `h2o.automl()` function to make use of `h2o.getAutoML()`. Does everything else look good?

PR #3311, opened 2019-02-21
Improves accuracy of GBM models trained with monotone constraints

michalkurka:


h2o-algos/src/main/java/hex/tree/gbm/GBM.java
```diff
@@ -965,18 +972,22 @@ public GammaPass(FrameMap frameMap, DTree[] trees, int[] leafs, Distribution dis
     }
 
     double gamma(int tree, int nid) {
+      return calcGamma(tree, nid, _num[tree][nid]);
+    }
+
+    double calcGamma(int tree, int nid, double num) {
```
maurever: Don't you like overloading? :) If not, change the name and add a description, please. It is not clear why two calculations of gamma are here.

PR #3312, opened 2019-02-21
[HEXDEV-720] update docker images for easier hive import testing

honzasterba:


docker/hadoop/cdh/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # pull base image
-FROM ubuntu:16.04
+FROM harbor.h2o.ai/opsh2oai/h2o-3/dev-r-3.4.1-jdk-8:10
```
michal-raska: Using this base image, might be a problem. Sparkling Water images are based on these images, and some of them need Java 7. What is the reason to use the R image as base please?

docker/hadoop/hdp/scripts/sbin/get_hive_home.sh
```diff
@@ -0,0 +1 @@
+export HIVE_HOME=/usr/hdp/$(ls -1 /usr/hdp | grep -v current)/${HIVE_VERSION}
```
michal-raska: Can we set the HIVE_HOME via env in this case as well, so we don't need this file?

docker/hadoop/cdh/scripts/startup/51_hive_start
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -ex
+
+source /usr/sbin/get_hive_home.sh
```
michal-raska: I think the `source` here is not necessary, the file is empty

docker/hadoop/common/sbin/hive_version_check.sh
```diff
@@ -0,0 +1,18 @@
...
+    HIVE=hive
+else
+    HIVE="$HIVE_HOME/bin/hive"
+fi
+HIVE_VERSION_MAJOR=$($HIVE --version | grep -E '^Hive' | sed -E 's/Hive ([0-9])\..*/\1/')
+if [ "$HIVE_VERSION_MAJOR" = "2" ]
+then
+    exit 0
+else
+    # no hive or hive 1.x or hive 0.x
```
michal-raska: Could we print a message here please?

docker/hadoop/common/startup/50_hive_init
```diff
@@ -0,0 +1,17 @@
...
+
+set -e -x
+
+sudo -E -u hdfs hadoop fs -mkdir -p /user/hive/warehouse
+sudo -E -u hdfs hadoop fs -chmod a+w /user/hive/warehouse
+sudo -E -u hdfs hadoop fs -chmod a+w /tmp
+
+# init mysql database for metastore
+
+rm -rf /var/lib/mysql
```
michal-raska: I think this should be moved to a separate file like `..._mysql_init`.

h2o-dist/buildinfo.json
```diff
@@ -20,91 +20,109 @@
     "hadoop_distributions" : [
         {
             "distribution" : "cdh5.4",
+            "hiveVersion"  : "hive",
```
michal-raska: I think this should be elsewhere, the information is relevant only to our images. Probably the best place would be https://github.com/h2oai/h2o-3/blob/fef47a9386b6971190e2967a6d3f8cdb5dd86b2d/scripts/jenkins/groovy/pipelineUtils.groovy#L84.

docker/hadoop/cdh/Dockerfile
honzasterba: well we need R, first I tried copy&pasting the code to install R but 
- image build failed because I reached CRAN API limits 
- building the images took forever, this speeds it up a lot
- I dont understand the note about java7, we are using a specific java8+R341 image
- less code/reuse = win

docker/hadoop/cdh/scripts/startup/51_hive_start
honzasterba: true, will remove

docker/hadoop/common/sbin/hive_version_check.sh
honzasterba: done

docker/hadoop/hdp/scripts/sbin/get_hive_home.sh
honzasterba: I wanted to but docker ENV only supports static values so we have to find the true value at runtime

docker/hadoop/common/startup/50_hive_init
honzasterba: will do

h2o-dist/buildinfo.json
honzasterba: removed

michal-raska:
Thanks @honzasterba!

PR #3316, opened 2019-02-25
PUBDEV-6308: AUC on leader model returns value calculated not on leaderboard_frame

deil87:


h2o-automl/src/test/java/ai/h2o/automl/AutoMLModelMetricsTest.java
```diff
@@ -0,0 +1,76 @@
...
+    autoMLBuildSpec.build_control.stopping_criteria.set_max_models(1);
+    autoMLBuildSpec.build_control.stopping_criteria.set_seed(7890);
+    autoMLBuildSpec.build_control.keep_cross_validation_models = false;
+    autoMLBuildSpec.build_control.keep_cross_validation_predictions = false;
+
+    AutoML aml = AutoML.startAutoML(autoMLBuildSpec);
+    aml.get();
+    Leaderboard leaderboard = aml.leaderboard();
+
+    // It is expected that auc() called on leader will return value estimated on `leaderboard_frame` but it is not the case.
```
michalkurka: I don't think expectation is reasonable, the contract of the getLeader method is to return the leader model.

I does seem natural to expect the leader metrics but this should not be handled on the model level.

seb-h2o: @michalkurka we had this discussion with @deil87 : of course as soon as Leader is just a Model instance, there's no reason to change the metrics, but to avoid the confusion you mention, we could also return a different type of object, probably proxying Model, with:
- `getLeader().auc()`  would return leaderboard metric.
- `getLeader().model.auc()` would return training/validation metric.
the problem is backwards compatibility of course, maybe we should make our users vote for this kind of change...

PR #3318, opened 2019-02-25
PUBDEV-6196: expose include_algos parameter to AutoML Py+R clients

seb-h2o:
https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6196

this is a client-only feature

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/Algo.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/Algo.java
@@ -3,7 +3,7 @@ package ai.h2o.automl;
 // if we need to make the Algo list dynamic, we should just turn this enum into a class...
 // implementation of AutoML.algo can be safely removed once we get rid of this interface: current purpose
 // is to keep backward compatibility with {@link AutoML.algo}
-public enum Algo implements AutoML.algo {
+public enum Algo {
   GLM,
   DRF,
   GBM,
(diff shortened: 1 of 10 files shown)


angela0xdata:
@seb-h2o can you add R and Python tests for this?

seb-h2o:
@angela0xdata what kind of tests are you thinking about? I already added Py+R UT, cf. changes in `runit_automl_args.R` and `pyunit_automl_args.py`

angela0xdata:
@seb-h2o oh, sorry. I didn't see those in here. (How the heck did I miss those =/)

ledell:
@seb-h2o Will this throw a client-side error if the user sets both args?  Thx.

seb-h2o:
@ledell yes, it will, there are even tests checking this behaviour ;)

ledell:
@seb-h2o Can you make the changes (add the new param) to automl.rst?

seb-h2o:
@ledell, I just added support for `include_algos` on backend, so it's now available through Java API or REST API but... still not in Flow, as automl doesn't have any API discovery in Flow, contrary to other algos.
However, I'm not even sure we should have it in Flow at all: 
1. the `exclude_algos` field is implemented as checkboxes, so adding another field with the same options will just make the form more cumbersome: we don't offer than many algorithms, so it's easy to click a couple more times to get include logic.
I think an `include_algos` could make sense if it were implemented like tags in Jira for example (cf. https://selectize.github.io/selectize.js/#content) but with basic checkboxes, I don't see any benefit.
2. if we had both, the recommended UI behaviour would be to hide one as soon as we start making a selection in the other to avoid exceptions: not usually difficult, if it were not for Flow...

ledell:
@seb-h2o I totally agree with you -- I think we should only have the `exclude_algos` option in Flow.  Having both is confusing.

PR #3322, opened 2019-02-27
PUBDEV-6319: Test that shows that AstGroup can return unexpected values

deil87:
During Integration of target encoding( TE) into AutoML I have noticed some unexpected inconsistency in results for the same TE parameters. So I tracked it down to the method that is doing grouping for calculation of the frequencies. I checked out revision that was before my refactoring in AstGroup class to see if it was something that I had introduced when I was implementing TE. The bug was there before. 
Behaviour is quite interesting... only from time to time AstGroup returns only one line with different aggregated values( maybe it is specific for a titanic dataset...but still it seems that it is always one line with inconsistency).

### DIFF
--- a/h2o-core/src/test/java/water/rapids/ast/prims/mungers/AstMergeTest.java
+++ b/h2o-core/src/test/java/water/rapids/ast/prims/mungers/AstMergeTest.java
@@ -20,30 +20,6 @@ public class AstMergeTest extends TestUtil {
     stall_till_cloudsize(1);
   }
 
-
-  @Test(timeout = 10000) // This merge is not going to finish in reasonable time. Check if it is n*log(n)
-  public void AutoMergeAllLeftStressTest() {
-
-    fr = new TestFrameBuilder()
-            .withName("leftFrame")
-            .withColNames("ColA", "ColB")
-            .withVecTypes(Vec.T_NUM, Vec.T_STR)
-            .withRandomIntDataForCol(0, 1000000, 0, 100)
-            .withRandomBinaryDataForCol(1, 1000000)
-            .build();
-
-    Frame frRight = new TestFrameBuilder()
-            .withName("rightFrame")
-            .withColNames("ColA_R", "ColB_R")
-            .withVecTypes(Vec.T_NUM, Vec.T_STR)
-            .withRandomIntDataForCol(0, 1000000, 0, 100)
-            .withRandomBinaryDataForCol(1, 1000000)
-            .build();
-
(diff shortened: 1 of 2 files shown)


michalkurka:
thanks, looks like there indeed is a race condition. @deil87 please make a jira

h2o-core/src/test/java/water/rapids/ast/prims/mungers/AstMergeTest.java
```diff
@@ -20,30 +20,6 @@ static public void setup() {
...
-            .withColNames("ColA_R", "ColB_R")
-            .withVecTypes(Vec.T_NUM, Vec.T_STR)
-            .withRandomIntDataForCol(0, 1000000, 0, 100)
-            .withRandomBinaryDataForCol(1, 1000000)
-            .build();
-
-    String tree = "(merge leftFrame rightFrame TRUE FALSE [0.0] [0.0] 'auto' )";
-    Rapids.exec(tree);
-  }
-
```
wendycwong: Is this the failing test?  I remembered you added a JIRA to fix this right?

deil87: @wendycwong this branch was branched from commit before my small refactoring in AstGroup to show that this behaviour is not related to my changes. Refactored version is failing in the same fashion. So you can adapt the test for the new version of using grouping api and work with it.

wendycwong: Looks like you removed two tests and added one.  Is there a reason to remove the two tests?  Also, for the test that you add, if it is failing, you can add @Ignore in front of it.  I want to make sure the Jenkin tests pass for your PR but at the same time, I want to keep your test for me to work with.  Thanks, Wendy

deil87: @wendycwong I removed other tests because they did not compile(and Ignore annotation would not help here) because I branched from some intermediate commit in my TE development. So don't pay attention to those tests. My goal was to show that it was failing before my changes in AstGroup. Now we need to fix it in the current master.

wendycwong: Got it and thanks.  I was actually thinking about pushing this PR to master and then check it out to make the change.  Now, I will just port your test over and make sure it runs.  Thanks.

wendycwong:
Andrey:

Can you close this PR?  I already copied your test over to my branch.


Wendy

PR #3323, opened 2019-02-28
PUBDEV-5801: Added column names and types to model output

wendycwong:
This PR completes the work described in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-5801?filter=-1 .

I have added the Java backend to copy the column types to model output.
I have added a pyunit test to check for all H2O algos that the column names and column types are correct.

For stackedensemble, I added the check to one existing pyunit test: pyunit_stackedensemble_regression.py

For xgboost, I added the check to one existing pyunit test: pyunit_insurance_poisson_small.py

### DIFF
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -620,9 +620,22 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
     /** Columns used in the model and are used to match up with scoring data
      *  columns.  The last name is the response column name (if any). */
     public String _names[];
+    public String _column_types[];
 
+    /**
+     * @deprecated as of March 6, 2019, replaced by (@link #setNames(String[] names, String[] columnTypes))
+     * 
+     */
+    @Deprecated
     public void setNames(String[] names) {
       _names = names;
+      _column_types = new String[names.length];
+      Arrays.fill(_column_types, "NA");
+    }
+
+    public void setNames(String[] names, String[] columntypes) {
+      _names = names;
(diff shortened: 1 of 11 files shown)


h2o-core/src/main/java/hex/Model.java
```diff
@@ -663,6 +680,7 @@ protected Output(ModelBuilder b, Frame train) {
         throw new IllegalArgumentException(b.validationErrors());
       // Capture the data "shape" the model is valid on
       setNames(train != null ? train.names() : new String[0]);
+      setColumnTypes(train);
```
michalkurka: this can cause NullPointerException - train can be null

wendycwong: I added the null pointer check in the method setColumnTypes.  Sounds okay?

wendycwong: I changed it to setColumnTypes(train!=null?train.typesStr():new String[0]).  This should fix it.

h2o-core/src/main/java/water/api/schemas3/ModelOutputSchemaV3.java
```diff
@@ -19,6 +19,9 @@
   @API(help="Column names", direction=API.Direction.OUTPUT)
   public String[] names;
 
+  @API(help="Column types", direction=API.Direction.OUTPUT)
+  public String[] column_types;
```
Pscheidl: The output of this should be a finite set, basically defined by `Vec.T_*` constants. Which means it should be documented somewhere which values our users might expect.

PR #3326, opened 2019-02-28
PUBDEV-6300 Flows throws error on an empty path while performing 'Import Files' operation

Pscheidl:
JIRA: 
https://0xdata.atlassian.net/browse/PUBDEV-6300

Depends on changes in Flow:
https://github.com/h2oai/h2o-flow/pull/130

### DIFF
--- a/h2o-core/src/main/java/water/persist/PersistManager.java
+++ b/h2o-core/src/main/java/water/persist/PersistManager.java
@@ -15,6 +15,7 @@ import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.regex.Pattern;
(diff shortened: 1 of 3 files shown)


h2o-core/src/main/java/water/persist/PersistManager.java
```diff
@@ -247,6 +247,11 @@ private static boolean httpUrlExists(String URLName){
    * @return List of matches
    */
   public List<String> calcTypeaheadMatches(String filter, int limit) {
+    filter = filter.trim();
```
Pscheidl: Let's "calculate" with trimmed value anyway.

h2o-core/src/main/java/water/persist/PersistManager.java
```diff
@@ -247,6 +247,11 @@ private static boolean httpUrlExists(String URLName){
...
+    if (filter.isEmpty()) {
+      return new ArrayList<>();
```
Pscheidl: Default behavior is to return nothing if the path is non-existing. As whitespaces or empty path means non-existing, return nothing as well.

Pscheidl:
Waiting for https://github.com/h2oai/h2o-flow/pull/130 to be accepted by more people, then releasing new version of Flow. Then `WIP` label will be removed.

h2o-core/src/main/java/water/persist/PersistManager.java
honzasterba: Collections.emptyList() is better

Pscheidl:
@honzasterba Resolved.

PR #3328, opened 2019-02-28
PUBDEV-6292: remove NA row from GLM varimp

wendycwong:
This work fixes the problem described in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6292?filter=-1

Tasked completed:
1. remove the intercept term from the standardized coefficient magnitude calculation from the backend.  The intercept term magnitude is assigned NA anyway.
2. Redirected model.R or model_base.py to look for standardized_coefficient_magnitudes when calculating variable importances.  GLM Binomial and GLM Multinomial have different coefficient fields.
3. Added Runit and pyunit tests to make sure varimp for GLM look the same as for other algos and varimp_plot works.

### DIFF
--- a/h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
+++ b/h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
@@ -103,8 +103,9 @@ public class GLMModelV3 extends ModelSchemaV3<GLMModel, GLMModelV3, GLMModel.GLM
               return 0;
             }
           });
-          String [] names2 = new String[names.length];
-          for(int i = 0; i < names2.length-1; ++i)
+          int len = names.length-1;
+          String [] names2 = new String[len]; // this one decides the length of standardized table length
+          for(int i = 0; i < len; ++i)
             names2[i] = names[indices[i]];
           tdt = new TwoDimTable("Standardized Coefficient Magnitudes", "standardized coefficient magnitudes", names2, new String[]{"Coefficients", "Sign"}, new String[]{"double", "string"}, new String[]{"%5f", "%s"}, "names");
           for (int i = 0; i < magnitudes.length - 1; ++i) {
@@ -235,8 +236,9 @@ public class GLMModelV3 extends ModelSchemaV3<GLMModel, GLMModelV3, GLMModel.GLM
             return 0;
           }
         });
-        String[] names2 = new String[names.length];
-        for (int i = 0; i < names2.length - 1; ++i)
+        int len = names.length-1;
+        String[] names2 = new String[len];
+        for (int i = 0; i < len; ++i)
           names2[i] = names[indices[i]];
         tdt = new TwoDimTable("Standardized Coefficient Magnitudes", "standardized coefficient magnitudes", names2, new String[]{"Coefficients", "Sign"}, new String[]{"double", "string"}, new String[]{"%5f", "%s"}, "names");
         for (int i = 0; i < beta.length - 1; ++i) {
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -414,9 +414,24 @@ class ModelBase(backwards_compatible()):
         :returns: A list or Pandas DataFrame.
         """
         model = self._model_json["output"]
-        if "variable_importances" in list(model.keys()) and model["variable_importances"]:
-            vals = model["variable_importances"].cell_values
-            header = model["variable_importances"].col_header
+        if self.algo=='glm' or "variable_importances" in list(model.keys()) and model["variable_importances"]:
+            if self.algo=='glm':
+                tempvals = model["standardized_coefficient_magnitudes"].cell_values
+                maxVal = 0
+                sum=0
+                for item in tempvals:
+                    sum=sum+item[1]
+                    if item[1]>maxVal:
+                        maxVal = item[1]
+                vals = []
+                for item in tempvals:
+                    tempT = (item[0], item[1], item[1]/maxVal, item[1]/sum)
+                    vals.append(tempT)
+                header = ["variable", "relative_importance", "scaled_importance", "percentage"]
+            else:
+                vals = model["variable_importances"].cell_values
+                header = model["variable_importances"].col_header
+                
             if use_pandas and can_use_pandas():
                 import pandas
                 return pandas.DataFrame(vals, columns=header)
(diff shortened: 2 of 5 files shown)


h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -103,8 +103,9 @@ public int compare(Integer o1, Integer o2) {
               return 0;
             }
           });
-          String [] names2 = new String[names.length];
-          for(int i = 0; i < names2.length-1; ++i)
+          int len = names.length-1;
+          String [] names2 = new String[len]; // this one decides the length of standardized table length
+          for(int i = 0; i < len; ++i)
```
wendycwong: Note that the for loop goes from 0 to names2.length-2.  Hence, the intercept term is not assigned anything and therefore set to NA.  I just make sure we generate names2 to be the exact length without the intercept term.  Hence, nothing is assigned to NA.

h2o-py/h2o/model/model_base.py
```diff
@@ -414,9 +414,24 @@ def varimp(self, use_pandas=False):
...
+                        maxVal = item[1]
+                vals = []
+                for item in tempvals:
+                    tempT = (item[0], item[1], item[1]/maxVal, item[1]/sum)
+                    vals.append(tempT)
+                header = ["variable", "relative_importance", "scaled_importance", "percentage"]
+            else:
+                vals = model["variable_importances"].cell_values
+                header = model["variable_importances"].col_header
+
```
wendycwong: Here I just grabbed the standardized_coefficient_magnitudes and convert them to the format used for variable importance of other algos.  Now, model.varimp() will provide the same output for all algos including GLM.

h2o-py/h2o/model/model_base.py
```diff
@@ -529,10 +544,16 @@ def coef_norm(self):
...
+        if self._model_json["output"]["model_category"]=="Multinomial":
+            tbl = self._model_json["output"]["standardized_coefficient_magnitudes"]
+            if tbl is None:
+                return None
+            return {name: coef for name, coef in zip(tbl["names"], tbl["coefficients"])}
+        else:
+            tbl = self._model_json["output"]["coefficients_table"]
+            if tbl is None:
+                return None
+            return {name: coef for name, coef in zip(tbl["names"], tbl["standardized_coefficients"])}
```
wendycwong: I cannot combine the two because multinomial coef_norm are always +ve while for other families, they can be either positive or negative

h2o-py/h2o/model/model_base.py
```diff
@@ -1139,6 +1153,9 @@ def varimp_plot(self, num_of_features=None, server=False):
         elif self._model_json["algo"] == "deeplearning":
             plt.title("Variable Importance: H2O Deep Learning", fontsize=20)
             if not server: plt.show()
+        elif self._model_json["algo"] == "glm":
+            plt.title("Variable Importance: H2O GLM", fontsize=20)
+            if not server: plt.show()            
         else:
             raise H2OValueError("A variable importances plot is not implemented for this type of model")
```
wendycwong: Need to add this so that GLM will be recognized to be an algo with variable importance.

h2o-r/h2o-package/R/models.R
```diff
@@ -1542,7 +1542,18 @@ h2o.varimp <- function(object) {
...
+      maxCoeff <- max(tvi$coefficients)
+      sumCoeff <- sum(tvi$coefficients)
+      scaledCoeff <- tvi$coefficients/maxCoeff
+      percentageC <- tvi$coefficients/sumCoeff
+      variable <- tvi$names
+      relative_importance <- tvi$coefficients
+      scaled_importance <- scaledCoeff
+      percentage <- percentageC
+      vi <- data.frame(variable, relative_importance, scaled_importance, percentage)
+      }  # no true variable importances, maybe glm coeffs? (return standardized table...)
```
wendycwong: All these calculations are performed to make sure the varimp for glm contains the exact same field as the other algos.  Namely, they need to have relative_importance, scaled_importance and percentage.

h2o-r/h2o-package/R/models.R
```diff
@@ -2730,9 +2736,17 @@ h2o.varimp_plot <- function(model, num_of_features = NULL){
...
   if(model@algorithm[1] != "glm") stop("Warning: model must be a GLM")
-
+  maxcoeff = 1
+  if (model@model$model_summary["family"]=="multinomial") {
+    coeff_table <- model@model$standardized_coefficient_magnitudes
+    sorted_table <- coeff_table[order(abs(coeff_table$coefficients)),]
+    norm_coef <- sorted_table$coefficients
+    sort_norm <- norm_coef
+    maxcoeff = max(norm_coef)
+  } else {
```
wendycwong: ensured that the standard coefficient plot of multinomial is grabbing the correct field. It is standardized-coefficient_magnitude and not the coefficients_table

h2o-r/tests/testdir_algos/glm/runit_pubdev_6292_varimp_check.R
```diff
@@ -0,0 +1,34 @@
...
+  myX = c("ID","AGE","RACE","CAPSULE","DCAPS","PSA","VOL","DPROS")
+  Log.info("Checking varimp for GLM Gaussian")
+  buildModelVarimpCheck("gaussian", h2o.data, x_indices=myX, y_index=myY)
+}
+
+buildModelVarimpCheck <- function(family, training_frame, x_indices, y_index) {
+  model <- h2o.glm(y=y_index, x=x_indices, training_frame=training_frame, family=family)
+  varimp <- h2o.varimp(model)
+  expect_true(sum(is.na(varimp$relative_importance))==0, "NA still appears in varimp")
+  h2o.varimp_plot(model)
```
wendycwong: test to make sure no NAs are found in the varimp returned.

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
michalkurka: @wendycwong what happens when the model is trained without an intercept (parameter `_intercept = false`) - is this still going to work or will it cause one coefficient to be skipped?

h2o-py/h2o/model/model_base.py
michalkurka: Perfect, thank you!

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
michalkurka: Just wanted to make sure the intercept is included in the total lenght even when it was disabled in model training. Perhaps adding a test would be a good idea?

michalkurka:
@wendycwong looks good but can you please add a test that makes sure this works correctly when GLM is trained *without* intercept?

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
wendycwong: Michal:  If you check the original code, the implementation will always assume that there is an intercept term.  That is why the for loop to assign values to name2 goes from 0 to names2.length-2.   I did not change anything in my implementation except to assign name to of length names2.length-1.  So, if you do train without the intercept term, both the original implementation and my implementation here will exclude one predictor.

michalkurka:
Discussed in person, thank you

PR #3329, opened 2019-02-28
[PUBDEV-6305] fix context_path related bugs

honzasterba:
- h2o.init now correctly respects context_path when starting new h2o instance
- h2o.downloadAllLogs now respects context_path
- h2o.downloadCSV now respects context_path


Pscheidl:
I only double-checked our `TestUtils` are unaffected, as I had strong feeling it could be used there. It's not. Captain obvious out.

PR #3331, opened 2019-03-01
PUBDEV-6283 - Stratified split fails when ran on the same column twice

Pscheidl:
## Jira
https://0xdata.atlassian.net/browse/PUBDEV-6283

## Reproducibility
Original script to reproduce the issue from the JIRA

```python
import h2o
h2o.init()
df_h2o = h2o.import_file("https://s3.amazonaws.com/h2o-public-test-data/smalldata/covtype/covtype.20k.data")
stratified = df_h2o[54].stratified_split()
train = df_h2o[stratified=="train"]
test  = df_h2o[stratified=="test"]
df_h2o['C3'] = df_h2o['C3'].asfactor()
col = 'C3'

split=df_h2o[col].stratified_split(test_frac=0.3)
train=df_h2o[split=='train']
test=df_h2o[split=='test']
print(len(train))
print(train[col].nacnt())
print(train[col].nlevels())
train['C3'].stratified_split(test_frac=0.1)
```

### DIFF
--- a/h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstStratifiedSplit.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstStratifiedSplit.java
@@ -57,7 +57,7 @@ public class AstStratifiedSplit extends AstPrimitive {
     // Collect input vector domain
     final long[] classes = new VecUtils.CollectIntegerDomain().doAll(stratifyingColumn).domain();
     // Number of output classes
-    final int numClasses = stratifyingColumn.isNumeric() ? classes.length : stratifyingColumn.domain().length;
+    final int numClasses = classes.length;
     
     // Make a new column based on input column - this needs to follow layout of input vector!
     // Save vector into DKV
--- a/h2o-core/src/test/java/water/util/VecUtilsTest.java
+++ b/h2o-core/src/test/java/water/util/VecUtilsTest.java
@@ -7,6 +7,8 @@ import org.junit.Test;
 import water.DKV;
 import water.TestUtil;
 import water.fvec.Frame;
+import water.fvec.TestFrameBuilder;
+import water.fvec.Vec;
 
 /**
  * Test VecUtils interface.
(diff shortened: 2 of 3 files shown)


h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstStratifiedSplit.java
```diff
@@ -57,7 +57,7 @@ public static Vec split(Vec stratifyingColumn, double splittingFraction, long ra
     // Collect input vector domain
     final long[] classes = new VecUtils.CollectIntegerDomain().doAll(stratifyingColumn).domain();
     // Number of output classes
-    final int numClasses = stratifyingColumn.isNumeric() ? classes.length : stratifyingColumn.domain().length;
+    final int numClasses = classes.length;
```
Pscheidl: This is the actual fix. Domain of the stratifyingColumn (the Vec instance) remains the same, but there are less rows afterwards, means "high" probability of reducing the domain cardinality. This is the case here. CollectIntegerDomain always provides the up-to-date cardinality no matter what.

h2o-core/src/test/java/water/util/VecUtilsTest.java
```diff
@@ -32,4 +34,25 @@ public void testStringVec2Categorical() {
       f.delete();
     }
   }
+
+  @Test
+  public void collectIntegerDomain() {
```
Pscheidl: I wrote this test to test collectIntegerDomain works. Why not keep it ? :question:

h2o-py/tests/testdir_jira/pyunit_pubdev_6283.py
```diff
@@ -0,0 +1,26 @@
+import h2o
+from tests import pyunit_utils
+
+
+def pubdev_6283():
```
Pscheidl: Polished version of the provided script with some additional asserts.

h2o-py/tests/testdir_jira/pyunit_pubdev_6283.py
```diff
@@ -0,0 +1,26 @@
...
+    test  = data[stratified=="test"]
+    assert train.nrows + test.nrows == data.nrows
+    
+    split=data['C3'].stratified_split(test_frac=0.3)
+    train=data[split=='train']
+    test=data[split=='test']
+    assert train.nrows + test.nrows == data.nrows
+    
+    split = train['C3'].stratified_split(test_frac=0.1)
+    split.show()
```
Pscheidl: Trigger the lazy init.

Pscheidl:
Test failures unrelated. Restarting ( because why not ?)

PR #3147, opened 2018-12-15
Improve Leaked Keys handling in JUNIT tests

Pscheidl:

PR #3332, opened 2019-03-01
PUBDEV-6542 - Distributed JUnit key leak test with RapidsTest as the first testsuite using it

Pscheidl:
Defines a new JUNIT runner named `H2ORunner`. Aims to replace the old logic in TestUtil, but is heavily inspired by it. Main flaws of the previous solutions were:

- **Leaked keys detected at the end of each test suite.** Keys leaked in the middle of a test suite would be detected at the end of the whole suite, not pointing at the right test which actually caused the leak.
- **Tests with shared resources not supported.** The support can be done even for current tests using TestUtils by introducing a new type of parent, as I did in my old & experimental PR, which is now closed: https://github.com/h2oai/h2o-3/pull/3147  (I named it `TestUtilsSharedResources`).
- Stall_till_cloud_size presence not enforced.
- Output not really pretty. Most of the time, it is unremoved `Frame` which causes the leak. The rest of the keys are related to the frame (vecs, chunks, rollup stats ...).

This new algo works in a distributed manner. Each node holds a set of keys introduced at the time the test suit is initialized (this basically means `@BeforeClass` phase). Each node works with Local Test Runtime in that phase. After each test, each node checks if there are any new keys left uncleaned. Please note, even the original algorithm in TestUtil ignores some keys, as those are deleted asynchronously and waiting for those to be cleaned means making them synchronous or ignoring them. Typically rollup stats. For similarities between this algo and the old one, please see `TestUtil` - `  public static void checkLeakedKeys()`. My logic should reflect it.

### Enforced `@CloudSize`

snimek z 2019-03-01 18-04-26

### `@CloudSize` values sanity check

snimek z 2019-03-01 18-06-43

### Key leak detection

snimek z 2019-03-01 18-07-55

### Highlights the exact test that caused the leak
snimek z 2019-03-01 18-08-37

* Output of leaked keys is prettier - I groups leaked keys by frame. I'll print Frame key, then the vec & chunk keys, all with proper indentation. This way, it is easy to recognize what exactly leaked and which keys are only transitive leak.


h2o-core/src/test/java/water/runner/CheckKeysTask.java
```diff
@@ -0,0 +1,45 @@
+package water.runner;
+
+import water.*;
+import water.util.ArrayUtils;
+
+import java.util.*;
+
+public class CheckKeysTask extends MRTask<CheckKeysTask> {
+
+    Key[] leakedKeys;
```
honzasterba: Is there anyl reason you use an array instead of a List (or Set which is used below)? all this code would be much simpler with a List and I dont see any benefit of an array

Pscheidl: Yes, we do not have means for moving Set around the cluster implemented (serialization/deserialization) in our core. I used Sets for the local computation.

Pscheidl:
One major flaw caught: The `runner.*` classes are test-only, so the H2O cloud started does not contain them and thus is unable to invoke the MR tasks.

h2o-core/src/test/java/water/runner/CleanNewKeysTask.java
```diff
@@ -0,0 +1,26 @@
...
+        for (Key actualKey : actualKeys){
+            final Value keyValue = Value.STORE_get(actualKey);
+            if(initKeys.contains(actualKey) || isIgnorableKeyLeak(actualKey, keyValue)) continue;
+            H2O.STORE.remove(actualKey);
+            actualKey.remove();
+        }
+        
+    }
+
+    private static boolean isIgnorableKeyLeak(final Key key, final Value keyValue) {
```
wendycwong: This function here and the one in CheckKeysTask looks exactly the same.  Wonder if you can combine them.

Pscheidl: Yes, this is doable.

Pscheidl: Done.

Pscheidl:
Rebased onto `rel-yau`

h2o-core/src/test/java/water/runner/LocalTestRuntime.java
```diff
@@ -0,0 +1,16 @@
...
+import water.Key;
+
+import java.util.HashSet;
+import java.util.Set;
+
+@Ignore
+class LocalTestRuntime {
+    protected static Set<Key> initKeys;
+
+    static{
```
honzasterba: unnecessary static block

h2o-core/src/test/java/water/runner/LocalTestRuntime.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.junit.Ignore;
+import water.Key;
+
+import java.util.HashSet;
+import java.util.Set;
+
+@Ignore
+class LocalTestRuntime {
+    protected static Set<Key> initKeys;
```
honzasterba: I think package private is enough here, we dont need protected access

h2o-core/src/test/java/water/runner/LocalTestRuntime.java
Pscheidl: Adjusted to your code style preferences.

h2o-core/src/test/java/water/runner/LocalTestRuntime.java
Pscheidl: Thanks. We don't.

Pscheidl:
So, I got green tests and everything should work. Let's see if I can repeat that again.

Pscheidl:
Rebased on latest master. The related `RapidsTest` finished successfully.

Please have a look if you'd give it a green light or not @honzasterba @michalkurka .

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -122,4 +122,27 @@ protected void setupLocal() {
 
 
   }
+
+  /**
+   * Retains given keys locally, not cluster-wide.
+   *
+   * @param retainedKeys Keys to retain
+   */
+  public static void retainLocal(final Set<Key> retainedKeys) {
```
michalkurka: Is this safe to be included in core? I am the current usage is harmless however, there might be hidden issues:

- Models can make assumption about the state of DKV - they can eg. assume that a specific is Frame is present in memory - then Model#remove would lead to NPE if the Frame was deleted. This is certainly a bad design but possible.

- this code might suffer from race conditions, eg. line 143 - if eg. a CV model is deleted as a part of another model (that is homed on another node), the model deletion can happen between lines 142-143 and an exception in the remove method.

michalkurka:
LGTM in general - I would just be careful about exposing certain functions to h2o-core outside of the scope of tests.

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: Ha! There might be a new way to do this. I just realized that. Next monday, I'll do some experiments. Thank you.

Pscheidl: Turns out we don't really need to do this step. Why ? Let's just clean everything after the whole test class is executed - I can't imagine any links among separate test classes.

h2o-core/src/test/java/water/runner/H2ORunner.java
```diff
@@ -0,0 +1,165 @@
...
+import water.util.Log;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+
+@Ignore
+public class H2ORunner extends BlockJUnit4ClassRunner {
+    private TestClass testClass;
```
honzasterba: both could be final

h2o-core/src/test/java/water/runner/H2ORunner.java
```diff
@@ -0,0 +1,165 @@
...
+        for (Key key : leakedKeysSet) {
+            Log.err(String.format("Key '%s'", key.toString()));
+        }
+    }
+
+
+
+    private int fetchCloudSize() {
+        final CloudSize annotation = testClass.getAnnotation(CloudSize.class);
+        if (annotation == null) throw new IllegalStateException("@CloudSize annotation is missing for test class: " + testClass.getName());
```
honzasterba: why so restrictive, I think this should do nothing if the annotation is missing

h2o-core/src/test/java/water/runner/KeysMRTask.java
```diff
@@ -0,0 +1,23 @@
+package water.runner;
+
+import org.junit.Ignore;
+import water.Job;
+import water.Key;
+import water.MRTask;
+import water.Value;
+
+@Ignore
+public abstract class KeysMRTask<T extends KeysMRTask<T>> extends MRTask<T> {
```
honzasterba: is this class really necessary? seems like it has only one child

h2o-core/src/test/java/water/runner/H2ORunner.java
Pscheidl: There was a discussion about it. The original design was `@Cloudsize(1)` behavior if there is no annotation. There were discussions and the main argument for this requirement was "lack of hidden magic".

h2o-core/src/test/java/water/runner/KeysMRTask.java
Pscheidl: Thanks. Removed.

h2o-core/src/test/java/water/runner/H2ORunner.java
Pscheidl: Thanks. Declared as final.

PR #3334, opened 2019-03-04
PUBDEV-6262 edit R api to be able get/pass long seed as string

maurever:
Edit R API to be able to pass long seed autogenerated from h2o server.

It is only R related. 
- algos are able to get seed as a string
- if the user wants to use a bigger number than the integer size, he/she has to pass seed as a string
- if the user wants to use autogenerated seed from h2o server, model@parameters$seed also returns this number as a string
- if user pass possibly longer number than the integer size, it produces the warning

JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6262


h2o-r/h2o-package/R/communication.R
```diff
@@ -583,17 +582,56 @@ print.H2OTable <- function(x, header=TRUE, ...) {
...
+    parameters = jsonParsed[[modelId]][[1]]$parameters
+    for (i in 1:length(parameters)){
+        if(parameters[[i]]$name=="seed"){
+            seedIndex = i
+            break
+        }
+    }
+    seedParsed = parameters[[seedIndex]]$actual_value
+    # test if seed is corrupted by R, if yes convert it to string 
+    if (is.numeric(seedParsed) && toString(seedParsed) != seedString){
```
michalkurka: The drawback of this approach is that people will sometimes get a number back and sometimes a string. That can be surprising and perhaps also cause issues.

Maybe we should keep seed as String always. WDYT?

The issue is also not limited to seed, any integer number can have this issue.

maurever: So input can be long and string too, and output always string? Only in R?  I think it is also confusing, however still better than now.

maurever:
I reset back to the first solution to pass parameter seed as the numeric/string on input and return seed as the numeric/string on output. There is a parameter in jsonlite::fromJSON to parse long integer as a string by default. It solves the problems with long in R JSON in general, however, it put some inconsistency into the output.

h2o-r/h2o-package/R/models.R
```diff
@@ -259,13 +262,23 @@ h2o.getFutureModel <- function(object,verbose=FALSE) {
...
         type <-  "character"
-    if (!inherits(paramValue, type)) {
-      e <- paste0("\"", name , "\" must be of type ", type, ", but got ", class(paramValue), ".\n")
-    } else if ((length(paramDef$values) > 1L) && !(paramValue %in% paramDef$values)) {
-      e <- paste0("\"", name,"\" must be in")
-      for (fact in paramDef$values)
-        e <- paste0(e, " \"", fact, "\",")
-      e <- paste(e, "but got", paramValue)
+    if (name == "seed") {
+	  if(is.character(paramValue) && suppressWarnings(is.na(bit64::as.integer64((paramValue)))))
```
michalkurka: package bit64 is not part of the default R installation, it should certainly be optional for H2O to work

h2o-r/tests/testdir_jira/runit_pubdev_6262_long_seed_to_string.R
```diff
@@ -0,0 +1,43 @@
...
+    logloss2 <- gbm_model2@model$scoring_history$training_logloss
+
+    # Compare results with same string seed setting autogenerated from h2o server
+    expect_equal(auto_seed, gbm_model2@parameters$seed)
+    expect_equal(rmse1, rmse2)
+    expect_equal(logloss1, logloss2)
+
+    # Expect warning when using long number in R
+    params$seed <- -8664354335142703762
+    expect_warning(do.call(h2o.gbm, params))
```
michalkurka: please check the warning - to make sure you are getting the right one

h2o-r/tests/testdir_jira/runit_pubdev_6262_long_seed_to_string.R
```diff
@@ -0,0 +1,43 @@
...
+    logloss1 <- gbm_model1@model$scoring_history$training_logloss
+
+    # Train model with string seed
+    params$seed <- auto_seed
+    gbm_model2 <- do.call(h2o.gbm, params)
+    rmse2 <- gbm_model2@model$scoring_history$training_rmse
+    logloss2 <- gbm_model2@model$scoring_history$training_logloss
+
+    # Compare results with same string seed setting autogenerated from h2o server
+    expect_equal(auto_seed, gbm_model2@parameters$seed)
```
michalkurka: please add a convenience function, eg. h2o.get_seed(model) to retrieve the seed - the function's documentation can mention that the seed can mention the different ways a seed can be represented

h2o-r/h2o-package/R/communication.R
```diff
@@ -583,20 +582,17 @@ print.H2OTable <- function(x, header=TRUE, ...) {
...
     rawREST <- .h2o.doSafeREST(h2oRestApiVersion = h2oRestApiVersion, urlSuffix = page, parms = .params, method = method, timeout = timeout)
   }else if(autoML == TRUE){
     rawREST <- .h2o.doSafeREST(h2oRestApiVersion = h2oRestApiVersion, urlSuffix = page, parms = .params, method = method,autoML=autoML)
   }else{
     rawREST <- .h2o.doSafeREST(h2oRestApiVersion = h2oRestApiVersion, urlSuffix = page, parms = .params, method = method)
   }
-
   if( raw ) rawREST
-  else      .h2o.fromJSON(jsonlite::fromJSON(rawREST,simplifyDataFrame=FALSE))
+  else      .h2o.fromJSON(jsonlite::fromJSON(rawREST,simplifyDataFrame=FALSE, bigint_as_char=TRUE))
```
michalkurka: this changes all the number outputs, something to definitely mention in release notes and possible document the impacts

cc: @angela0xdata

h2o-bindings/bin/gen_R.py
```diff
@@ -65,6 +65,8 @@ def gen_module(schema, algo, module):
         if param["name"] == "seed":
             if algo != "stackedensemble":
                 yield "#' @param seed Seed for random numbers (affects certain parts of the algo that are stochastic and those might or might not be enabled by default)"
+                yield "#'        The type is long, however, R cannot handle long number correctly, so the type can also be a string."
```
michalkurka: I would leave the function documentation as is. The matter not very clear to a regular user and also won't affect the regular user. Instead of having the topic explained in each algo, I would prefer to put it in a small section in the documetation.

h2o-r/tests/testdir_jira/runit_pubdev_6262_long_seed_to_string.R
maurever: I added it, however, it does the same functionality as model@parameters$seed right now. Is that what you want?

h2o-r/h2o-package/R/communication.R
maurever: To be clear: It changes only outputs, which are long in JAVA to string in R.

h2o-r/h2o-package/R/communication.R
```diff
@@ -583,20 +582,24 @@ print.H2OTable <- function(x, header=TRUE, ...) {
...
     rawREST <- .h2o.doSafeREST(h2oRestApiVersion = h2oRestApiVersion, urlSuffix = page, parms = .params, method = method,autoML=autoML)
   }else{
     rawREST <- .h2o.doSafeREST(h2oRestApiVersion = h2oRestApiVersion, urlSuffix = page, parms = .params, method = method)
   }
-
   if( raw ) rawREST
-  else      .h2o.fromJSON(jsonlite::fromJSON(rawREST,simplifyDataFrame=FALSE))
+  else {
+    res <- .h2o.fromJSON(jsonlite::fromJSON(rawREST,simplifyDataFrame=FALSE, bigint_as_char=TRUE))
+    resToCompare <- .h2o.fromJSON(jsonlite::fromJSON(rawREST,simplifyDataFrame=FALSE))
```
michalkurka: I wouldn't do this - some of the json structures can be very large and this doubles the parsing time and memory requirements

we can make this optional based on user setting, eg.:

    checkjson <- getOption("h2o.warning.on.string.conversion", FALSE)

h2o-r/tests/testdir_jira/runit_pubdev_6262_long_seed_to_string.R
michalkurka: yes, that looks good, please use the new function here in the test

h2o-r/h2o-package/R/models.R
```diff
@@ -258,19 +266,34 @@ h2o.getFutureModel <- function(object,verbose=FALSE) {
...
-        e <- paste0(e, " \"", fact, "\",")
-      e <- paste(e, "but got", paramValue)
+      type <-  "character"
+      if (name == "seed") {
+        if(is.character(paramValue) && !.is.int64(paramValue))
+          e <- paste0("\"seed\" must be of type long or string long, but got a string which cannot be converted to long.\n")
+        else if(is.numeric(paramValue)){
+          if(!.is.int64(paramValue)){
+            e <- paste0("\"seed\" must be of type long or string long, but got a number which cannot be converted to long.\n")
+          } else if(nchar(toString(paramValue)) > 16) {
```
michalkurka: I don't think this works:

    > paramValue <- 100000000000000000
    > toString(paramValue)
    [1] "1e+17"

I tried this piece of code with the given paramValue and no warning was produced

PR #3338, opened 2019-03-05
PUBDEV-6335: GBM monotonicity constraints only work for gaussian distrib

michalkurka:


angela0xdata:
@michalkurka what about XGBoost?

michalkurka:
XGBoost is not restricted like that

angela0xdata:
okay, thanks for the info!

Mathieu-Jacob:
Can you explain why it is restricted to gaussian only? Is it planned to deploy it for other distributions also?

Mathieu-Jacob:
I came across this link stating monotonic constraints are for regression only: https://0xdata.atlassian.net/browse/PUBDEV-6335

Is the restriction on regression only (which would allow other distributions like tweedie)? or on the distribution?

michalkurka:
It is only allowed for Gaussian in this PR, Bernoulli is implemented in another PR

The reason why it is done this way is that it needs custom implementation for each distribution. We can add other later but at this point in time only Gaussian and Bernoulli will be supported

PR #3339, opened 2019-03-08
PUBDEV-5809: AutoML expose max_runtime_per_model property

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-5809

also see PR https://github.com/h2oai/h2o-3/pull/3218 for core changes required to ensure that each individual model will respect this time constraint, especially when doing cross-validation.

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -167,7 +167,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
   public static void startAutoML(AutoML aml) {
     // Currently AutoML can only run one job at a time
     if (aml.job == null || !aml.job.isRunning()) {
-      H2OJob j = new H2OJob(aml, aml._key, aml.timeRemainingMs());
+      H2OJob j = new H2OJob(aml, aml._key, aml.runCountdown.remainingTime());
       aml.job = j._job;
       j.start(aml.workAllocations.remainingWork());
       DKV.put(aml);
@@ -414,10 +414,6 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     if (buildSpec.build_control.stopping_criteria.max_models() == 0)
       return Integer.MAX_VALUE;
     return buildSpec.build_control.stopping_criteria.max_models() - modelCount.get();
-  }
-
-  private boolean timingOut() {
-    return timeRemainingMs() <= 0;
   }
 
   @Override
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -339,7 +339,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
   @Override
   public void run() {
     runCountdown.start();
-    userFeedback.info(Stage.Workflow, "AutoML build started: " + fullTimestampFormat.format(new Date()));
+    userFeedback.info(Stage.Workflow, "AutoML build started: " + fullTimestampFormat.format(runCountdown.start_time()));
     learn();
     stop();
   }
@@ -352,7 +352,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     jobs = null;
 
     runCountdown.stop();
-    userFeedback.info(Stage.Workflow, "AutoML build stopped: " + fullTimestampFormat.format(new Date()));
+    userFeedback.info(Stage.Workflow, "AutoML build stopped: " + fullTimestampFormat.format(runCountdown.stop_time()));
     userFeedback.info(Stage.Workflow, "AutoML build done: built " + modelCount + " models");
 
     Log.info(userFeedback.toString("User Feedback for AutoML Run " + this._key + ":"));
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -81,7 +81,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     private Work[] allocations = new Work[0];
 
     WorkAllocations allocate(Algo algo, int count, JobType type, int workShare) {
-      if (!canAllocate) throw new IllegalStateException("can't allocate new work");
+      if (!canAllocate) throw new IllegalStateException("Can't allocate new work.");
 
       allocations = ArrayUtils.append(allocations, new Work(algo, count, type, workShare));
       return this;
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -25,6 +25,7 @@ import water.fvec.Frame;
 import water.fvec.Vec;
 import water.nbhm.NonBlockingHashMap;
 import water.util.ArrayUtils;
+import water.util.Countdown;
 import water.util.IcedHashMapGeneric;
 import water.util.Log;
 
@@ -231,7 +232,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
 
   private Date startTime;
   private static Date lastStartTime; // protect against two runs with the same second in the timestamp; be careful about races
-  private long stopTimeMs;
+  private Countdown runCountdown;
   private Job job;                  // the Job object for the build of this AutoML.
 
   private transient List<Job> jobs; // subjobs
@@ -256,6 +257,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     this.startTime = startTime;
     this.buildSpec = buildSpec;
     this.jobs = new ArrayList<>();
+    this.runCountdown = new Countdown(Math.round(1000 * buildSpec.build_control.stopping_criteria.max_runtime_secs()));
 
     try {
       userFeedback = new UserFeedback(this);
@@ -336,7 +338,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
 
   @Override
   public void run() {
-    stopTimeMs = System.currentTimeMillis() + Math.round(1000 * buildSpec.build_control.stopping_criteria.max_runtime_secs());
+    runCountdown.start();
     userFeedback.info(Stage.Workflow, "AutoML build started: " + fullTimestampFormat.format(new Date()));
     learn();
     stop();
@@ -349,6 +351,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     for (Job j : jobs) j.get(); // Hold until they all completely stop.
     jobs = null;
 
+    runCountdown.stop();
     userFeedback.info(Stage.Workflow, "AutoML build stopped: " + fullTimestampFormat.format(new Date()));
     userFeedback.info(Stage.Workflow, "AutoML build done: built " + modelCount + " models");
 
@@ -403,14 +406,8 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     return buildSpec == null ? null : buildSpec.project();
   }
 
-  public long getStopTimeMs() {
-    return stopTimeMs;
-  }
-
   public long timeRemainingMs() {
-    if (getStopTimeMs() < 0) return Long.MAX_VALUE;
-    long remaining = getStopTimeMs() - System.currentTimeMillis();
-    return Math.max(0, remaining);
+    return runCountdown.remainingTime();
   }
 
   public int remainingModels() {
@@ -425,7 +422,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
 
   @Override
   public boolean keepRunning() {
-    return timeRemainingMs() > 0 && remainingModels() > 0;
+    return !runCountdown.timedOut() && remainingModels() > 0;
   }
 
   private boolean isCVEnabled() {
@@ -576,8 +573,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     jobs.add(subJob);
 
     long lastWorkedSoFar = 0;
-    Set<Key<Model>> lastGridModels = new HashSet<>();
-    long lastGridModelCreation = subJob.start_time();
+    long lastGridCount = 0;
 
     while (subJob.isRunning()) {
       if (null != parentJob) {
@@ -585,7 +581,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
           userFeedback.info(stage, "AutoML job cancelled; skipping " + name);
           subJob.stop();
         }
-        if (!ignoreTimeout && timingOut()) {
+        if (!ignoreTimeout && runCountdown.timedOut()) {
           userFeedback.info(stage, "AutoML: out of time; skipping " + name);
           subJob.stop();
         }
@@ -597,22 +593,12 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
       }
 
       if (JobType.HyperparamSearch == work.type) {
-        Grid grid = (Grid)subJob._result.get();
+        Grid<?> grid = (Grid)subJob._result.get();
         int gridCount = grid.getModelCount();
-        if (gridCount > lastGridModels.size()) {
+        if (gridCount > lastGridCount) {
           userFeedback.info(stage, "Built: " + gridCount + " models for search: " + name);
-          long now = System.currentTimeMillis();
-          long durationSinceLastGridModels = now - lastGridModelCreation;
-          lastGridModelCreation = now;
-          Key<Model>[] gridModels = grid.getModelKeys();
-          this.addModels(gridModels);
-          Set<Key<Model>> newGridModels = new HashSet<>(Arrays.asList(gridModels));
-          newGridModels.removeAll(lastGridModels);
-          for (Key<Model> mod : newGridModels) {
-            long buildDuration = millisToSec((double)durationSinceLastGridModels / newGridModels.size());
-            userFeedback.info(stage, mod + " build in "+ buildDuration +"s (approximation)");
-          }
-          lastGridModels.addAll(newGridModels);
+          this.addModels(grid.getModelKeys());
+          lastGridCount = gridCount;
         }
       }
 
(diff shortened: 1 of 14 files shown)


seb-h2o:
@deil87 can you please have a look at this one please? it really needs to get merged, thx

deil87:
@seb-h2o yes, sorry for delay. Going to start it today.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
```diff
@@ -153,4 +189,23 @@ public AutoMLBuildModelsV99() {
   @API(help="The AutoML Job key", direction=API.Direction.OUTPUT)
   public JobV3 job;
 
+
+  @Override
+  public AutoMLBuildSpecV99 fillFromBody(String body) {
+    AutoMLBuildSpecV99 schema = super.fillFromBody(body); //default JSON filling
+
+    // TODO: need to understand why we set stopping tolerance to AUTO iff stopping_criteria is provided without stopping_tolerance
```
deil87: @seb-h2o Have you resolved this TODO?  Judging just by what is written in TODO it looks logical that if we provide `stopping_criteria` and no `stopping_tolerance` has been specified then we should assign it to automatically selected value.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
```diff
@@ -118,16 +119,51 @@ public AutoMLBuildControlV99() {
...
+    @API(help = "Maximum time to spend on each individual model (optional).", direction = API.Direction.INOUT)
+    public double max_runtime_secs_per_model;
+
+    @API(help = "Early stopping based on convergence of stopping_metric. Stop if simple moving average of length k of the stopping_metric does not improve for k:=stopping_rounds scoring events (0 to disable)", level = API.Level.secondary, direction=API.Direction.INOUT)
+    public int stopping_rounds;
+
+    @API(help = "Metric to use for early stopping (AUTO: logloss for classification, deviance for regression)", valuesProvider = StoppingMetricValuesProvider.class, level = API.Level.secondary, direction=API.Direction.INOUT)
+    public ScoreKeeper.StoppingMetric stopping_metric;
+
+    @API(help = "Relative tolerance for metric-based stopping criterion (stop if relative improvement is not at least this much)", level = API.Level.secondary, direction=API.Direction.INOUT)
```
deil87: @seb-h2o here I'm just curious how we decide when level should be set to secondary ?

seb-h2o: didn't notice this `@API` param until now, thanks for pointing at it: I'm afraid it is not used consistently currently:
allowed values are:
`critical` : seems to be a duplicate for the `required` param, and this is the default value... maybe this is the reason why this is ignored, as `required` is doing a better job.
`secondary`: probably relevant for many params.
`expert`: the difference with `secondary` is getting obscure now...
Looks like a not-that-useful param.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
seb-h2o: my interrogation was that not in the `if` but in the `iff`, meaning the `only if` part.
if no `stopping_criteria` is provided at all, then `stopping_tolerance` is not set to `AUTO` but to a default value of `0.001`, which makes those default value difficult to.
To sum up:
```
stopping_criteria = null    ->  stopping_tolerance = 0.001
stopping_criteria = {}       ->  stopping_tolerance = AUTO  // = -1
stopping_criteria = {max_runtime_secs = 300}       ->  stopping_tolerance = AUTO  // = -1
```
You have to agree that it is very confusing, and I'm considering it as a bug indeed.
@ledell are you aware of this?

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
seb-h2o: after short investigation, seems to be used to group params in Flow

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -229,7 +232,7 @@ public AutoMLBuildSpec getBuildSpec() {
 
   private Date startTime;
   private static Date lastStartTime; // protect against two runs with the same second in the timestamp; be careful about races
-  private long stopTimeMs;
+  private Countdown runCountdown;
```
deil87: @seb-h2o runCountdown is slightly confusing as we have two types of countdowns... in terms of number of models and in terms of time that is left. wdyt?

seb-h2o: @deil87 talking about the name? countdown is usually about time, isn't it?

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -594,12 +589,12 @@ private void pollAndUpdateProgress(Stage stage, String name, WorkAllocations.Wor
       }
 
       if (JobType.HyperparamSearch == work.type) {
-        Grid grid = (Grid)subJob._result.get();
+        Grid<?> grid = (Grid)subJob._result.get();
         int gridCount = grid.getModelCount();
-        if (gridCount > gridLastCount) {
-          userFeedback.info(Stage.ModelTraining, "Built: " + gridCount + " models for search: " + name);
+        if (gridCount > lastGridCount) {
```
deil87: @seb-h2o couple of questions for pollAndUpdateProgress method: 
1) Why we are checking for `HyperparamSearch` and `ModelBuild` work types for stragglers and only for `HyperparamSearch` in the `while (subJob.isRunning())` block?
2) Following piece of code is almost identical
```
Grid<?> grid = (Grid) subJob._result.get();
        int gridCount = grid.getModelCount();
        if (gridCount > lastGridCount) {
          userFeedback.info(stage, "Built: " + gridCount + " models for search: " + name);
          this.addModels(grid.getModelKeys());
          lastGridCount = gridCount;
        }
```
why we call `subJob._result.get()` instead of just `subJob.get()` ?  Aren't we checking the same case twice here? If it is for the same case but we need to check it at different moments then we need to move it to a method and call it twice. 
3) if `gridCount` is total number of grid entries then `gridCount > lastGridCount` will always be true except for the case when they are equal. And why don't we want to add model and feedback for last grid count?
4) if I got naming correctly for `lastGridCount` then maybe we can use `lastGridEntry'? wdyt?

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
deil87: @seb-h2o yes, about naming. Probably you are right. Even thought definition is " the act of counting backwards to zero" I found only examples that are related to time. I was confused because of the CountDownLatch class in Java.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: @deil87 
1. from my understanding, `ModelBuild` build only one model, so we don't need to pick it live, whereas for grids it makes sense (live progress is only available in Flow today, but this may/will change). Collecting stragglers also makes sense for grids due to the 1s sleep: after this sleep the last model may be completed and the job stopped, which will escape the while loop, so you need to check for this last ones....
2. `subJob.get()` blocks until completion, whereas `subJob._result.get()` is a way to peep into the result, which makes sense inside the loop as we want to see live updates.
3. correct for the first part. The "feedback" can surely be improved here and in various places, what do you mean by "add model and feedback for last grid count"? Currently, it's adding model and feedback each time there's one or more new models since last update, if you were testing for '=', then the "feedback" would be printed after every 1s sleep...
4. not sure I get your point here, entry would mean it stores a model, I think count makes sense here.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
seb-h2o: Note that this is not an issue with Py/R/Flow clients as stopping_criteria can't be null there, therefore we get `stopping_tolerance = AUTO` if not set.
But for those using Java or REST API directly, this is inconsistent.
For Sparkling Water, there may be the same inconsistency.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: strictly speaking, could be renamed to `CountdownTimer`, but can't do this here as it requires changes in core module :)

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
deil87: @seb-h2o I see now, for me it had equally "not provided" meaning for `null` and `{}` and any other things that denote non-existence/emptiness. Thank you for clarification.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
deil87: @seb-h2o this is actually what I'd initially thought about these levels. Then back to the question how to decide which one to use. Should we find a guideline for that?

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
deil87: @seb-h2o ahah :)

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
deil87: @seb-h2o 3) I see now. Comment for `grid.getModelCount();` is confusing. I thought it is a total number of entries in the grid, but it looks like it is number of (evaluated+launched) entries so far. 
4) Then of course `lastGridEntry` is not appropriate here. But still I think `lastGridCount` could be improved. It could be read as `count from last grid in the sequence of grids`. I would suggest something that convey information that it is about models that have been built...even method that you are calling says `grid.getModelCount()` so it is more about modelsCount than gridCount... for example for gridCount we can use `evaluatedModelsCount' or  `countOfBuiltModelsInGrid` or `evaluatedModelsCountInGrid`. For `lastGridEntry` - same but with `last` prefix.   Longer but more clear.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -55,13 +59,95 @@ public AutoMLBuildControl() {
...
+    public AutoMLStoppingCriteria() {
+      super();
+      _searchCriteria = new RandomDiscreteValueSearchCriteria();
+    }
+
+    public double max_runtime_secs_per_model() {
+      return max_runtime_secs_per_model;
+    }
+
+    public void set_max_runtime_secs_per_model(double max_runtime_secs_per_model) {
```
deil87: @seb-h2o minor, underscore instead of `this`.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -55,13 +59,95 @@ public AutoMLBuildControl() {
...
+
+    public void set_stopping_tolerance(double stopping_tolerance) {
+      _searchCriteria.set_stopping_tolerance(stopping_tolerance);
+    }
+
+    public void set_default_stopping_tolerance_for_frame(Frame frame) {
+      _searchCriteria.set_default_stopping_tolerance_for_frame(frame);
+    }
+
+    public static double default_stopping_tolerance_for_frame(Frame frame) {
```
deil87: @seb-h2o I don't see usages and it is static. Maybe it is left after testing something.

h2o-automl/src/test/java/ai/h2o/automl/AutoMLTest.java
```diff
@@ -192,6 +192,58 @@
...
+    try {
+      AutoMLBuildSpec autoMLBuildSpec = new AutoMLBuildSpec();
+//      fr = parse_test_file("./smalldata/prostate/prostate_complete.csv"); //using slightly larger dataset to make this test useful
+//      autoMLBuildSpec.input_spec.response_column = "CAPSULE";
+      fr = parse_test_file("./smalldata/diabetes/diabetes_text_train.csv"); //using slightly larger dataset to make this test useful
+      autoMLBuildSpec.input_spec.response_column = "diabetesMed";
+      autoMLBuildSpec.input_spec.training_frame = fr._key;
+
+      int max_runtime_secs_per_model = 10;
+      autoMLBuildSpec.build_models.exclude_algos = new Algo[] {Algo.DRF};
```
deil87: @seb-h2o why are we excluding DRF here?

h2o-automl/src/test/java/ai/h2o/automl/AutoMLTest.java
```diff
@@ -192,6 +192,58 @@
...
+          Log.info(key + " added build duration (s): "+ duration / 1e3);
+        }
+        duration = model._output._total_run_time / 1e3;
+        Log.info(key + " model total build duration (s): "+ duration);
+        assertTrue(key + " took longer than required: "+ duration,
+            duration - max_runtime_secs_per_model < tolerance);
+        model.deleteCrossValidationModels();
+      }
+
+      // no assertion, we just want to check leaked keys
```
deil87: @seb-h2o What for we are summing up `duration` here? I would expect assertion here that training for each model  took predefined time +- generously added 2s or at least whole autoML process finished within expected timebox.

h2o-core/src/main/java/water/util/Countdown.java
```diff
@@ -32,11 +32,11 @@ public Countdown(long _time_limit_millis, boolean start) {
     this(_time_limit_millis);
     if (start) start();
   }
-  
+
   public Date start_time() {
     return new Date(_start_time);
   }
-  
+
```
deil87: @seb-h2o it is not your changes but shouldn't we use here camel case for methods? Inconsistency within class... for example `fromSeconds` vs `start_time`

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: you're getting picky here, it's a local variable whose scope is limited to the loop ;)
but yeah, `lastGridModelsCount` is better... I was just reflecting the existing local variable named `gridCount`...

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
seb-h2o: right, I added the underscores to this class in a previous PR, forgot the new one during the merge

h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
```diff
@@ -260,6 +261,24 @@ def test_automl_stops_after_max_runtime_secs():
...
+    models_count = {}
+    for max_runtime_secs_per_model in [0, 3, max_runtime_secs]:
+        aml = H2OAutoML(project_name="py_aml_max_runtime_secs_per_model_{}".format(max_runtime_secs_per_model), seed=1,
+                        max_runtime_secs_per_model=max_runtime_secs_per_model,
+                        max_runtime_secs=max_runtime_secs)
+        aml.train(y=ds['target'], training_frame=ds['train'])
+        models_count[max_runtime_secs_per_model] = len(aml.leaderboard)
+        # print(aml.leaderboard)
+    # there may be one model difference as reproducibility is not perfectly guaranteed in time-bound runs
+    assert abs(models_count[0] - models_count[max_runtime_secs]) <= 1
```
deil87: @seb-h2o it means that in `models_count[max_runtime_secs]` case we might not get a model in the leaderboard? Why then for the `models_count[3]` case we are expecting one model to be built? `models_count[max_runtime_secs]` has more time then `models_count[3]` and intuitively I would expect it to deliver more models.

h2o-r/h2o-package/R/automl.R
```diff
@@ -189,6 +190,9 @@ h2o.automl <- function(x, y, training_frame,
 
   # Update build_control list with top level build control args
   build_control <- list(stopping_criteria = list(max_runtime_secs = max_runtime_secs))
+  if (!is.null(max_runtime_secs_per_model)) {
+      build_control$stopping_criteria$max_runtime_secs_per_model <- max_runtime_secs_per_model
```
deil87: @seb-h2o so we can use two notations interchangeably `self.build_control["stopping_criteria"]["max_runtime_secs_per_model"]` and  `build_control$stopping_criteria$max_runtime_secs_per_model` ?

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
deil87: @seb-h2o regarding limited scope. If this was in the dedicated method with the naming that reflects what we are doing here...then yes. But this method is 82 lines long and name of the method does not say anything about entities we are polling and tracking. Sorry for being picky... but it is for good :)

h2o-r/h2o-package/R/automl.R
seb-h2o: yep

h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
seb-h2o: there's no such assumption.
Just testing here that when setting `max_runtime_secs_per_model` to 0 (ie. disabled) or to `max_runtime_secs` (full time), then we get same amount of models (with a tolerance of 1).
But when setting it to 3s, then we get much less models, than when it's disabled (=0).

h2o-core/src/main/java/water/util/Countdown.java
seb-h2o: right right right, will do a small countdown renaming in a small future PR

h2o-automl/src/test/java/ai/h2o/automl/AutoMLTest.java
seb-h2o: yes, only the last assertion is useful now for this test.
Was using the duration on CVs to investigate timing issues at some point, this looks confusing now, as well as those logs. Will clean up.

h2o-automl/src/test/java/ai/h2o/automl/AutoMLTest.java
seb-h2o: another investigation leftover: removing.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: it's also legacy code: anyway, now that I need to make another push, will rename this

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
seb-h2o: good catch, it's usage probably got lost during a conflict resolution -> fixed

seb-h2o:
@deil87 addressed test clean up + naming issues

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -615,23 +610,23 @@ private void pollAndUpdateProgress(Stage stage, String name, WorkAllocations.Wor
...
       } else if (subJob.get() == null) {
-        userFeedback.info(stage, name + " cancelled");
+        userFeedback.warn(stage, name + " cancelled");
       } else {
-        Grid grid = (Grid) subJob.get();
+        Grid<?> grid = (Grid) subJob.get();
         int gridCount = grid.getModelCount();
-        if (gridCount > gridLastCount) {
-          userFeedback.info(Stage.ModelTraining, "Built: " + gridCount + " models for search: " + name);
+        if (gridCount > lastTotalGridModelsBuilt) {
```
deil87: @seb-h2o gridCount -> totalGridModelsBuilt

seb-h2o: you see? you ask to rename legacy code and now we start rechecking everything.
will rename, but can you still please approve? usage is to approve as soon as there's no major concern/unanswered question, and trust author for the rest, thx.

deil87: @seb-h2o I don't agree with suggested attitude. Once your agreed to change legacy code then let's do it consistently.

PR #3342, opened 2019-03-09
PUBDEV-6347 - Update S3 documentation for 3.22.x

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6347

The examples are re-formatted from my blogpost: https://www.pavel.cool/h2o-3/h2o-s3-credentials-change/

### DIFF
--- a/h2o-docs/src/product/aws.rst
+++ b/h2o-docs/src/product/aws.rst
@@ -21,7 +21,7 @@ AWS Standalone Instance
 -----------------------
 
 When running H2O in standalone mode using the simple Java launch
-command, we can pass in the S3 credentials in two ways.
+command, we can pass in the S3 credentials in three ways.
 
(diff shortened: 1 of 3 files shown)


angela0xdata:
@Pscheidl I made a couple of changes to the ec2-and-s3.rst file.

Also, the /product/aws.rst file is actually ignored, and I’m not sure why it’s still in here. (Maybe there was a reason at one point? ... I can't recall... But it’s not needed anymore.) As you discovered, it was replaced by the /cloud-integration/ec2-and-s3.rst file. 

And I see you updated the how-to markdown file. It wasn’t necessary, though. There is a note at the top of the file saying that the file is no longer maintained. But then it doesn’t hurt anything to update it.

Pscheidl:
@angela0xdata Thank you !

PR #3343, opened 2019-03-10
PUBDEV-6334 update sort javadoc

wendycwong:
The work in this PR completes the work  requested in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6334?filter=-1

I corrected the error in Javadoc for Frame.java sort function.  Thanks to Andrey for catching it.

The sorting function itself is implemented in the rapids framework and that is why the sort tests are under the rapids folder as well.

The whole rapids design came about before I joined H2O so I have no idea why things are organized the way it is.

For anything that I have implemented, I made sure to add thorough tests in Java, R and Python clients. I have done the following for sort:
1. enable sorting of doubles;
2. add sorting directions support;
3. allow string columns in the frame as long as we do not sort with it.
4. fixed memory issue with sort when there are many columns of duplicated sort values.

### DIFF
--- a/h2o-core/src/main/java/water/fvec/Frame.java
+++ b/h2o-core/src/main/java/water/fvec/Frame.java
@@ -1612,9 +1612,9 @@ public class Frame extends Lockable<Frame> {
   @Override public Class<KeyV3.FrameKeyV3> makeSchema() { return KeyV3.FrameKeyV3.class; }
 
   /** Sort rows of a frame, using the set of columns as keys.  User can specify sorting direction for each sorting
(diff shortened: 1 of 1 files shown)


deil87:
@wendycwong  Thank you!   Yep.... I can see now that there are tests in Py/R `pyunit_pubdev_4870_sort_bug_pubdev_4404_desc.py` and `runit_pubdev_4870_sort_bug_pubdev4404_desc.R` that are checking mixed direction for sorting. Unfortunately tests in Py and R are not organised into a nice folder hierarchy so it is difficult to find corresponding tests. So... looks good to me!

PR #3348, opened 2019-03-12
[PUBDEV-6269] Test for kerberized hive + kerberos test refactor

honzasterba:
remove KRB testing from hadoop-smoke-job, only test these cases on some hadoop builds
add tests for hadoop driver on kerberos


honzasterba:
ready for review, tested here:
http://mr-0xc1:8080/job/h2o-3-kerberos-smoke-pipeline/job/honza%2FPUBDEV-6269/12/

PR #3349, opened 2019-03-12
PUBDEV-6365: Adds support for monotonicity constraint for binomial GBM models

michalkurka:


michalkurka:
@maurever this is WIP for now - definitely needs some polishing

h2o-algos/src/main/java/hex/tree/DHistogram.java
```diff
@@ -408,7 +421,7 @@ public void incr0( int b, double y, double w ) {
    * @param hi  upper bound on index into rows array to be processed by this call (exclusive)
    * @param lo  lower bound on index into rows array to be processed by this call (inclusive)
    */
-  public void updateHisto(double[] ws, double[] cs, double[] ys, int [] rows, int hi, int lo){
+  void updateHisto(double[] ws, double resp[], double[] cs, double[] ys, int [] rows, int hi, int lo){
```
maurever: Please add doc. :) It is confusing to have `ys` named as response and add new value which is also named similar to response.

michalkurka: Sure, will do - thank you.

PR #3350, opened 2019-03-12
PUBDEV-6208: SE Model deletion can lead to removal of shared vecs

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6208

Not sure it covers the specific use-case mentioned in the ticket, but I was able to reproduce a similar behaviour when deleting a SE after AutoML run in Flow.
Note that AutoML has always `keep_level_one_frame=true`, which is the reason why this issue occurs more often there.

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -169,8 +169,8 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
     if (keepLevelOneFrame) {
       levelOneFrame.write_lock(_job);
       levelOneFrame = levelOneFrame.deepCopy(levelOneFrame._key.toString());
+      levelOneFrame.update(_job);
       levelOneFrame.unlock(_job);
-      DKV.put(levelOneFrame);
     }
     return levelOneFrame;
   }
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -167,10 +167,11 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
     boolean keepLevelOneFrame = isTraining && _parms._keep_levelone_frame;
     Frame levelOneFrame = prepareLevelOneFrame(levelOneKey, baseModels.toArray(new Model[0]), baseModelPredictions.toArray(new Frame[0]), actuals);
     if (keepLevelOneFrame) {
-      levelOneFrame.write_lock(_job);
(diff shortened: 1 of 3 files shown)


h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -102,46 +104,48 @@ private Frame prepareLevelOneFrame(String levelOneKey, Model[] baseModels, Frame
...
+
+      Frame old = DKV.getGet(levelOneFrame._key);
+      if (old != null && old instanceof Frame) {
+        Frame oldFrame = (Frame) old;
+        // Remove ALL the columns so we don't delete them in remove_impl.  Their
+        // lifetime is controlled by their model.
+        oldFrame.removeAll();
+        oldFrame.write_lock(_job);
+        oldFrame.update(_job);
+        oldFrame.unlock(_job);
```
seb-h2o: sorry for the noise: there was a block with wrong indentation here.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -160,8 +164,15 @@ private Frame prepareLevelOneFrame(String levelOneKey, Key<Model>[] baseModelKey
       baseModels.add(aModel);
       baseModelPredictions.add(predictions);
     }
-
-    return prepareLevelOneFrame(levelOneKey, baseModels.toArray(new Model[0]), baseModelPredictions.toArray(new Frame[0]), actuals);
+    boolean keepLevelOneFrame = isTraining && _parms._keep_levelone_frame;
+    Frame levelOneFrame = prepareLevelOneFrame(levelOneKey, baseModels.toArray(new Model[0]), baseModelPredictions.toArray(new Frame[0]), actuals);
+    if (keepLevelOneFrame) {
+      levelOneFrame.write_lock(_job);
+      levelOneFrame = levelOneFrame.deepCopy(levelOneFrame._key.toString());
```
michalkurka: Are we sure that when `isTraining` is true no new Vec are being calculated? Otherwise, we would leak data if we made a defensive copy of Vecs that no-one else has a reference to.

seb-h2o: @michalkurka  that's why I also tested it with blending (for which there are new Vec predictions computed).
And that's why the copy is done only if, on top of this, user asked to keep the level one frame: as soon as he makes this choice, the Vecs lifecycle depends on the model itself. So we don't leak keys, they're removed with the model.

seb-h2o:
@michalkurka : I will, I started to fix it in `rel-xu` first but I had issues with SE testing there. Now that my fix is simpler, I'll try again, thx!

seb-h2o:
@michalkurka please see https://github.com/h2oai/h2o-3/pull/3355 for the fix port to `rel-xu`

PR #3351, opened 2019-03-12
PUBDEV-6359: AutoML, ensure that no additional model get trained after job cancellation

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6359


seb-h2o:
@ledell @deil87 tiny one that was forgotten: please spend max 2 min on this

PR #3361, opened 2019-03-15
[PUBDEV-6363] remove support for getting hive config from hive-site.xml

honzasterba:
[PUBDEV-6363](https://0xdata.atlassian.net/browse/PUBDEV-6363)

- it never really worked for hiveHost (the config value used does something else)


h2o-mapreduce-generic/src/main/java/water/hadoop/HiveTokenGenerator.java
```diff
@@ -35,26 +31,23 @@ public HiveOptions(String hiveHost, String hivePrincipal) {
       _host = hiveHost;
       _principal = hivePrincipal;
     }
+
+    private static boolean isPresent(String value) {
+      return value != null && !value.isEmpty();
```
Pscheidl: To trim or not to trim ? Not sure if whitespaces are welcome here, just an idea.

1rs9jw

PR #3363, opened 2019-03-16
PUBDEV-6056: TreeSHAP for GBM Trees

michalkurka:
@Pscheidl this is not exposed to the users just yet - only internal implementation at this point.

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
@@ -1,5 +1,6 @@
 package hex.genmodel.algos.tree;
 
+import ai.h2o.algos.tree.INode;
 import hex.genmodel.tools.PrintMojo;
 import hex.genmodel.utils.GenmodelBitSet;
 
@@ -13,7 +14,8 @@ import java.util.Objects;
  * Node in a tree.
  * A node (optionally) contains left and right edges to the left and right child nodes.
  */
-public class SharedTreeNode {
+public class SharedTreeNode implements INode<double[]> {
+  final int internalId; // internal tree id (that links the node back to the array of nodes of the tree) - don't confuse with user-facing nodeNumber!
   final SharedTreeNode parent;
   final int subgraphNumber;
   int nodeNumber;
@@ -44,7 +46,8 @@ public class SharedTreeNode {
    * @param sn Tree number
    * @param d Node depth within the tree
    */
-  SharedTreeNode(SharedTreeNode p, int sn, int d) {
+  SharedTreeNode(int id, SharedTreeNode p, int sn, int d) {
+    internalId = id;
     parent = p;
     subgraphNumber = sn;
     depth = d;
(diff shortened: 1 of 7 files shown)


h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
```diff
@@ -13,7 +15,8 @@
  * Node in a tree.
  * A node (optionally) contains left and right edges to the left and right child nodes.
  */
-public class SharedTreeNode {
+public class SharedTreeNode implements INode<double[]>, INodeStat {
+  final int internalId; // internal tree id (that links the node back to the array of nodes of the tree) - don't confuse with user-facing nodeNumber!
```
Pscheidl: Is the array sequantially numberes ? If yes, this could be exposed via Tree API ?

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeSubgraph.java
```diff
@@ -59,7 +60,7 @@ public void setFontSize(int fontsize) {
    * @return The new child node
    */
   public SharedTreeNode makeLeftChildNode(SharedTreeNode parent) {
-    SharedTreeNode child = new SharedTreeNode(parent, subgraphNumber, parent.getDepth() + 1);
+    SharedTreeNode child = new SharedTreeNode(nodesArray.size(), parent, subgraphNumber, parent.getDepth() + 1);
```
Pscheidl: This does not make sense to me. Node's ID is length of the array with nodes ? Does the array grow, so the node ID always corresponds to the array size ?

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeSubgraph.java
```diff
@@ -105,6 +106,15 @@ public SharedTreeNode walkNodes(final String path) {
     return n;
   }
 
+  public float scoreTree(double[] data) {
+    SharedTreeNode n = rootNode;
```
Pscheidl: Elegant solution with respect to current encapsulation.

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeSubgraph.java
```diff
@@ -105,6 +106,15 @@ public SharedTreeNode walkNodes(final String path) {
     return n;
   }
 
+  public float scoreTree(double[] data) {
```
Pscheidl: Elegant solution with respect to current encapsulation.

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeSubgraph.java
michalkurka: @Pscheidl the arraylist grows, adding a node to the list is thus equivalent to incrementing a node id

PR #3367, opened 2019-03-18
PUBDEV-6350: fixed rendering of multiple Confusion Matrices when using Py and R clients

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6350

### DIFF
--- a/h2o-py/h2o/model/metrics_base.py
+++ b/h2o-py/h2o/model/metrics_base.py
@@ -573,18 +573,23 @@ class H2OBinomialModelMetrics(MetricsBase):
         return self._metric_json["thresholds_and_metric_scores"]["tpr"]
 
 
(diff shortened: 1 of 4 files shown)


seb-h2o:
will update, looks like there is the same bug with R client

seb-h2o:
@ledell , @deil87 : this is a small one, and the build is fully green. It's rare enough to be worth a quick review, thx.

seb-h2o:
anyone interested? @deil87 as a creator? @ledell for the R part? thank you....

ledell:
Does anyone know why we call the list of metrics, `max_metrics`?  Max of what?

deil87:
@ledell this is because we are looking for a threshold that maximises particular metric. So `max_metrics` is a list of metrics wrt which we are able to select optimal thresholds.

PR #3370, opened 2019-03-19
PUBDEV-6346: Enable R client to call groupby with multiple columns. 

wendycwong:
This PR fixes the bug in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6346?filter=-1

Fixed frame.R where it was not able to include multiple columns.

Fixed bug with duplicated columns return when specifying multiple groupby actions.

Fixed bug when user does not provide any action columns.  In this case, we will return actions performed on all elligible actions per proposal from Joe Granados.

Added a Runit test to verify that my fix is working.

### DIFF
--- a/h2o-r/h2o-package/R/frame.R
+++ b/h2o-r/h2o-package/R/frame.R
@@ -226,7 +226,6 @@ pfr <- function(x) { chk.H2OFrame(x); .pfr(x) }
       # Convert to data.frame
       L <- lapply(res$columns, function(c) {
         row <- if( c$type!="string" && c$type!="uuid" )  c$data  else  c$string_data
-        #if( length(row)!=res$row_count ) browser()
         stopifnot(length(row)==res$row_count) # No short columns
         row
       })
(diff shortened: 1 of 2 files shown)


wendycwong:
Joe:

Good point, I will check what happens with empty columns and add the RACE column as well.

wendycwong:
With this command:
gp_sd <- h2o.group_by(data = df.hex, by = c("RACE") ,sd(c("VOL", "AGE", "PSA", "RACE")), mean(c("VOL", "AGE", "PSA", "RACE")))

You will get:
  RACE sdev_VOL  sdev_AGE sdev_PSA sdev_RACE mean_VOL mean_AGE mean_PSA mean_RACE
1    0 10.84497 17.214335 31.68159         0    12.06667 62.66667 34.63333                     0
2    1 17.98231  6.461226 18.71270         0     15.38185 66.16129 14.26475                         1
3    2 21.78136  6.078064 27.19019         0    20.20833 65.16667 24.64167                       2

Which makes sense.

wendycwong:
However, with this one:

gp_sd_1 <- h2o.group_by(data = df.hex, by = c("RACE") ,sd(), mean())

You will get:

Error in agg[[2]] : subscript out of bounds

Which also makes sense.

g-eoj:
The behavior of `gp_sd_1 <- h2o.group_by(data = df.hex, by = c("RACE") ,sd(), mean())` is different from the docs and the python api. Not sure how import that is.

wendycwong:
However, the python client will return the groupby action performed on all the remaining columns.  I think we better make R and Python perform the same.  Let me add it in.

PR #3371, opened 2019-03-20
PUBDEV-6362: Replaced "h2o cloud" with "h2o cluster" in end-user docs

angela0xdata:
Updated User Guide, booklets, R, and Python docs.
Driveby fix: Removed FAQ indicating that Java 9 isn't supported. Fixed combining columns example.


ledell:
@angela0xdata I know this is already merged, but I reviewed it and it's perfect.  Thanks!

PR #3374, opened 2019-03-20
PUBDEV-6319 astgroup inconsistent

wendycwong:
This PR fixes the race condition caught by Andrey Spiridonov in this JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6319?filter=-1

I fixed the race condition by giving each thread its own IcedHashMap instead of sharing one per node.

I copied over Andrey Spiridonov's groupby test to make sure the fix actually works.

### DIFF
--- a/h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
@@ -514,7 +514,7 @@ public class AstGroup extends AstPrimitive {
   // Main worker MRTask.  Makes 1 pass over the data, and accumulates both all
   // groups and all aggregates
   public static class GBTask extends MRTask<GBTask> {
-    IcedHashMap<G, String> _gss; // each thread will have its own IcedHashMap
+    IcedHashMap<G, String> _gss = new IcedHashMap<>(); // each thread will have its own IcedHashMap
     private final int[] _gbCols; // Columns used to define group
     private final AGG[] _aggs;   // Aggregate descriptions
     private final int _medianCounts;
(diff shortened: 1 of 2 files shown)


michalkurka:
@wendycwong, where was the race condition happening?

wendycwong:
Michal:  The race condition happens when all threads share the _gss and try to do the reduce on it: line 548 and line 555.

wendycwong:
Thank you Andrey for CATCHING it!  You rock!

wendycwong:
Speed test with AstGroup with and without Wendy fix.  I take a 20G airlines dataset and choose one column as the group_by class and perform mean, min, max, mean, var, sd, ss, sum and median on all the rest of the columns.  To account for the various loading, I ran another experiment with the AstGroup.  Please see the results with raw run time and then with adjusted run time. 

image

wendycwong:
Michal:  I expected them to have similar performance because my fix did not add much to the operation and the memory either.  The small experiment supported my expectation for once!  Please let me know what you think.  Muchas gracias.

deil87:
@wendycwong closing original PR https://github.com/h2oai/h2o-3/pull/3322

deil87:
@wendycwong Is this PR going to make it into upcoming release? I set this PR as a blocker for target encoding integration into AutoML task... as I'm hitting those race conditions quite often when search over TE hyperparameters is happening.

deil87:
@wendycwong I've pushed a commit with some idea. Feel free to revert it.  Regarding memory consumption increase caused by your fix... I think it is natural to your fix as you start to store a reference to _gcc from outside of the map... and probably instances was not properly garbage collected. I've decided to try to use one hashmap per node..... but run into race condition especially for frequently updated groups. I added `synchronized` and it fixed race condition... but it might slow down overall performance.  Ideally we need to understand what within synchronised block causing race. I'm not sure but in the console logs I could even see situation when reduce start to execute before all the `map` calls are finished. I know that some MR frameworks has control over this (wait for all maps mode)...but in our case it is not clear for me what options we have yet. From high level perspective I suspect that it might be related to the fact that we first get something `_gss.getk(gWork)` and then update `_aggg[i].atomic..` but in the meantime of this two operations other threads can change something. Please let me know what you think.

deil87:
@michalkurka It should never be merged ? I think we just need to find appropriate way to fix the issue without memory implications. That is what i'm trying to do.

michalkurka:
@deil87 thanks for looking into it; it should eventually be merged! it just shouldn't be merged in the current state

both Wendy's and your solution have perf implications - we need to find and fix the underlying issues

there is currently a support ticket open on memory issues with AstGroups on a huge dataset - previous Wendy's solution would make it worse - that is why I marked as "Changes Requested"

deil87:
@michalkurka Sure. I did not push a solution... just shared my findings and thoughts. If @wendycwong has a benchmark then I would be curious to know how my changes affect performance. Maybe it improves memory consumption.

PR #3377, opened 2019-03-22
Standardisation investigation

deil87:
I was looking for a way to standardise frames in Java, found DataInfo class but there is no usages for method `standardise`. What I found out is that `DataInfo._adaptedFrame` does not contain data with applied transformations(e.g. standardisation). We need to use `extractDenseRow()` method to get it.

But what is strange is that most of the tests that look like they have to depend on standardised parameter pass regardless of whether we enabled it or not. 

PS don't pay attention to the name of the branch. It is probably working but tests are not checking this

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -1127,8 +1127,10 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
           DataInfo activeData = _state.activeData();
           double [] beta_nostd = activeData.denormalizeBeta(beta);
           DataInfo.TransformType transform = activeData._predictor_transform;
+          // TODO Why are we disabling Transform in the block where we want to standardize?
           activeData.setPredictorTransform(DataInfo.TransformType.NONE);
           Gram g = new GLMIterationTask(_job._key,activeData,new GLMWeightsFun(_parms),beta_nostd).doAll(activeData._adaptedFrame)._gram;
+          // TODO enabling it back...see hotfix comments to the commit affd062d9d842a4f5a7d7fa9908107ef7b9e0e34
           activeData.setPredictorTransform(transform); // just in case, restore the trasnform
           g.mul(_parms._obj_reg);
           chol = g.cholesky(null);
--- a/h2o-algos/src/test/java/hex/DataInfoTestAdapt.java
+++ b/h2o-algos/src/test/java/hex/DataInfoTestAdapt.java
@@ -2,8 +2,10 @@ package hex;
 
 import hex.glm.GLMModel;
 import hex.splitframe.ShuffleSplitFrame;
+import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Test;
+import water.DKV;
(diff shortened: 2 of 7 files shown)


h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1127,8 +1127,10 @@ else if(_parms._family == Family.gaussian && _parms._link == Link.identity)
           DataInfo activeData = _state.activeData();
           double [] beta_nostd = activeData.denormalizeBeta(beta);
           DataInfo.TransformType transform = activeData._predictor_transform;
+          // TODO Why are we disabling Transform in the block where we want to standardize?
           activeData.setPredictorTransform(DataInfo.TransformType.NONE);
```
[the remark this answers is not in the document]
deil87: @wendycwong Thank you Wendy, I got it. I just wanted to point out that tests that are checking some scenarios that are supposed to use standardisation .... actually are passing even with disabled standardisation.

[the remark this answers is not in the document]
wendycwong: Can you point me to a test?  This is not good.

h2o-algos/src/test/java/hex/DataInfoTestAdapt.java
```diff
@@ -26,12 +28,21 @@
     Model.InteractionSpec interactions = Model.InteractionSpec.allPairwise(new String[]{"class", "sepal_len"});
 
     boolean useAll=false;
-    boolean standardize=false;  // golden frame is standardized before splitting, while frame we want to check would be standardized post-split (not exactly what we want!)
+    boolean standardize=true;  // golden frame is standardized before splitting, while frame we want to check would be standardized post-split (not exactly what we want!)
```
deil87: @wendycwong changing  this did not break a test

h2o-algos/src/test/java/hex/glm/GLMBasicTestRegression.java
```diff
@@ -145,7 +145,7 @@ public void testSparse() {
     parms._train = _weighted._key;
     parms._ignored_columns = new String[]{_weighted.name(0)};
     parms._response_column = _weighted.name(1);
-    parms._standardize = true;
+    parms._standardize = false;
```
deil87: @wendycwong changing  this did not break a test. Actually I was changing  as many standardize properties as possible without breaking tests....to show how strange it is.

h2o-algos/src/test/java/hex/glm/GLMBasicTestRegression.java
```diff
@@ -762,10 +762,12 @@ public void testPValuesGaussian(){
 //    VOL         -0.18129997  0.1620383 -1.11887108 2.631951e-01
 //    GLEASON      0.91750972  0.1963285  4.67333842 2.963429e-06
 
-    params._standardize = true;
+    //TODO Comment above says that following section is about STANDARDIZED case but when we disable it test still passes. Looks like tolerance is not small enough
```
deil87: @wendycwong `testPValuesGaussian` is a bright example..... note comments 1) Standardised 2)  standardised

PR #3380, opened 2019-03-25
PUBDEV-6367 - Overridable non-data line markers

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6367

By default, some parsers are skipping lines beginning with some characters. Those defaults are implemented per each parser. The main point of this PR is to fulfill all the following conditions:

1. Make the defaults overridable,
1. Preserve the original behavior(defaults) when non-data line markers are not specified.

As a positive side-effect of this API, I made `CsvParser` and `ArffParser` use the new generic methods passing the whole `ParseSetup` object (reference to it of course), instead of pin-pointing only the necessary stuff. This old method has even been deprecated by @michalkurka some time ago. :)

### DIFF
--- a/h2o-core/src/main/java/water/parser/CsvParser.java
+++ b/h2o-core/src/main/java/water/parser/CsvParser.java
@@ -20,15 +20,16 @@ public class CsvParser extends Parser {
   private static final int HAS_HEADER = ParseSetup.HAS_HEADER;
   private static final byte[] NON_DATA_LINE_MARKERS_DEFAULT = {'#'};
 
+  private final byte[] _nonDataLineMarkers; 
+
   CsvParser( ParseSetup ps, Key jobKey ) {
+    this(ps, NON_DATA_LINE_MARKERS_DEFAULT, jobKey);
+  }
+
+  CsvParser(ParseSetup ps, byte[] defaultNonDataLineMarkers, Key jobKey) {
     super(ps, jobKey);
+    _nonDataLineMarkers = ps._nonDataLineMarkers != null ? ps._nonDataLineMarkers : defaultNonDataLineMarkers;
   }
-
-  protected byte[] nonDataLineMarkersDefault() {
-    return NON_DATA_LINE_MARKERS_DEFAULT;
-  }
-
-
 
   // Parse this one Chunk (in parallel with other Chunks)
   @SuppressWarnings("fallthrough")
@@ -67,7 +68,7 @@ public class CsvParser extends Parser {
     byte c = bits[offset];
     // skip comments for the first chunk (or if not a chunk)
     if( cidx == 0 ) {
-      while (ArrayUtils.contains(_setup._nonDataLineMarkers, c) || isEOL(c)) {
+      while (ArrayUtils.contains(_nonDataLineMarkers, c) || isEOL(c)) {
         while ((offset < bits.length) && (bits[offset] != CHAR_CR) && (bits[offset  ] != CHAR_LF)) {
 //          System.out.print(String.format("%c",bits[offset]));
           ++offset;
@@ -234,7 +235,7 @@ MAIN_LOOP:
               state = EXPECT_COND_LF;
             break;
           }
-          if (ArrayUtils.contains(_setup._nonDataLineMarkers, c)) {
+          if (ArrayUtils.contains(_nonDataLineMarkers, c)) {
             state = SKIP_LINE;
             break;
           }
@@ -527,7 +528,7 @@ MAIN_LOOP:
 
   @Override protected int fileHasHeader(byte[] bits, ParseSetup ps) {
     boolean hasHdr = true;
-    String[] lines = getFirstLines(bits, ps._single_quotes, ps._nonDataLineMarkers);
+    String[] lines = getFirstLines(bits, ps._single_quotes, _nonDataLineMarkers);
     if (lines != null && lines.length > 0) {
       String[] firstLine = determineTokens(lines[0], _setup._separator, _setup._single_quotes);
       if (_setup._column_names != null) {
@@ -687,7 +688,8 @@ MAIN_LOOP:
    */
   static ParseSetup guessSetup(byte[] bits, byte sep, int ncols, boolean singleQuotes, int checkHeader,
                                String[] columnNames, byte[] columnTypes, String[][] naStrings, byte[] nonDataLineMarkers) {
-    if(nonDataLineMarkers == null) nonDataLineMarkers = NON_DATA_LINE_MARKERS_DEFAULT;
+    if (nonDataLineMarkers == null)
+      nonDataLineMarkers = NON_DATA_LINE_MARKERS_DEFAULT;
     int lastNewline = bits.length-1;
     while(lastNewline > 0 && !CsvParser.isEOL(bits[lastNewline]))lastNewline--;
     if(lastNewline > 0) bits = Arrays.copyOf(bits,lastNewline+1);
--- a/h2o-core/src/main/java/water/parser/CsvParser.java
+++ b/h2o-core/src/main/java/water/parser/CsvParser.java
@@ -22,7 +22,6 @@ public class CsvParser extends Parser {
 
   CsvParser( ParseSetup ps, Key jobKey ) {
     super(ps, jobKey);
-    if(_setup._nonDataLineMarkers == null) _setup._nonDataLineMarkers = nonDataLineMarkersDefault();
   }
 
   protected byte[] nonDataLineMarkersDefault() {
--- a/h2o-core/src/main/java/water/parser/CsvParser.java
+++ b/h2o-core/src/main/java/water/parser/CsvParser.java
@@ -18,13 +18,18 @@ public class CsvParser extends Parser {
   private static final int NO_HEADER = ParseSetup.NO_HEADER;
   private static final int GUESS_HEADER = ParseSetup.GUESS_HEADER;
   private static final int HAS_HEADER = ParseSetup.HAS_HEADER;
-  private static final byte[] NON_DATA_LINE_MARKERS = {'#'};
+  private static final byte[] NON_DATA_LINE_MARKERS_DEFAULT = {'#'};
 
-  CsvParser( ParseSetup ps, Key jobKey ) { super(ps, jobKey); }
-
-  protected byte[] nonDataLineMarkers() {
-    return NON_DATA_LINE_MARKERS;
+  CsvParser( ParseSetup ps, Key jobKey ) {
+    super(ps, jobKey);
+    if(_setup._nonDataLineMarkers == null) _setup._nonDataLineMarkers = nonDataLineMarkersDefault();
   }
+
+  protected byte[] nonDataLineMarkersDefault() {
+    return NON_DATA_LINE_MARKERS_DEFAULT;
+  }
+
+
 
   // Parse this one Chunk (in parallel with other Chunks)
   @SuppressWarnings("fallthrough")
@@ -62,9 +67,8 @@ public class CsvParser extends Parser {
     int colIdx = 0; // count each actual column in the dataset including the skipped columns
     byte c = bits[offset];
     // skip comments for the first chunk (or if not a chunk)
-    byte[] nonDataLineMarkers = nonDataLineMarkers();
     if( cidx == 0 ) {
-      while (ArrayUtils.contains(nonDataLineMarkers, c) || isEOL(c)) {
+      while (ArrayUtils.contains(_setup._nonDataLineMarkers, c) || isEOL(c)) {
         while ((offset < bits.length) && (bits[offset] != CHAR_CR) && (bits[offset  ] != CHAR_LF)) {
 //          System.out.print(String.format("%c",bits[offset]));
           ++offset;
@@ -231,7 +235,7 @@ MAIN_LOOP:
               state = EXPECT_COND_LF;
             break;
           }
-          if (ArrayUtils.contains(nonDataLineMarkers, c)) {
+          if (ArrayUtils.contains(_setup._nonDataLineMarkers, c)) {
             state = SKIP_LINE;
             break;
           }
@@ -524,7 +528,7 @@ MAIN_LOOP:
 
   @Override protected int fileHasHeader(byte[] bits, ParseSetup ps) {
     boolean hasHdr = true;
-    String[] lines = getFirstLines(bits, ps._single_quotes, nonDataLineMarkers());
+    String[] lines = getFirstLines(bits, ps._single_quotes, ps._nonDataLineMarkers);
     if (lines != null && lines.length > 0) {
       String[] firstLine = determineTokens(lines[0], _setup._separator, _setup._single_quotes);
       if (_setup._column_names != null) {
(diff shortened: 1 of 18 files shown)


h2o-core/src/main/java/water/parser/ParseSetup.java
```diff
@@ -166,15 +172,36 @@ public ParseSetup(ParseSetupV3 ps) {
...
+        domains, naStrings, data, new ParseWriter.ParseErr[0], FileVec.DFLT_CHUNK_SIZE, nonDataLineMarkers);
   }
+
+  /**
+   * Create a ParseSetup with all parameters except chunk size.
+   *
+   * Typically used by file type parsers for returning final valid results
+   * _chunk_size will be set later using results from all files.
+   */
+  public ParseSetup(ParserInfo parseType, byte sep, boolean singleQuotes, int checkHeader,
```
michalkurka: this is becoming a constructor hell :)

h2o-py/h2o/h2o.py
```diff
@@ -591,6 +593,7 @@ def parse_setup(raw_frames, destination_frame=None, header=0, separator=None, co
     :param na_strings: A list of strings, or a list of lists of strings (one list per column), or a dictionary
         of column names to strings which are to be interpreted as missing values.
     :param skipped_columns: an integer lists of column indices to skip and not parsed into the final frame from the import file.
+    :param non_data_line_markers: A string containing custom characters to be treated as non-data line markers
```
michalkurka: this is confusing, isn't it? 

I would expect that `non_data_line_markers = None` means there are no such markers, not that the default will be applied

h2o-core/src/main/java/water/parser/CsvParser.java
```diff
@@ -18,14 +18,19 @@
...
 
-  CsvParser( ParseSetup ps, Key jobKey ) { super(ps, jobKey); }
+  CsvParser( ParseSetup ps, Key jobKey ) {
+    super(ps, jobKey);
+    if(_setup._nonDataLineMarkers == null) _setup._nonDataLineMarkers = nonDataLineMarkersDefault();
+  }
 
-  protected byte[] nonDataLineMarkers() {
-    return NON_DATA_LINE_MARKERS;
+  protected byte[] nonDataLineMarkersDefault() {
```
honzasterba: this method is useless I think

h2o-core/src/main/java/water/parser/ParseSetup.java
honzasterba: builder?

h2o-py/h2o/h2o.py
honzasterba: it also may be confusing what a non-data-line-marker is, I would suggest rewording, to something like:

If a line in imported file starts with any character in given string it will be imported. Empty string means all lines are imported, None means that default behaviour for given format will be used.

Pscheidl: Yes, it is confusing, but as someone hard-coded these values earlier, it is harder to change it right now. So, should we introduce a new default being handed over ? Some special combination of chars, like `DEFAULT` or something like that ? Seems also confusing to the users.

Best solution would be to drop the default behavior.

h2o-core/src/main/java/water/parser/CsvParser.java
Pscheidl: It is not. This behavior is overridden in `ARFFParser`.

Pscheidl: Unfortunately, it is not. The behavior is overridden in `ARFFParser`, which calls CSVParser's constructor in it's own constructor. Each class has different defaults to provide.

h2o-py/h2o/h2o.py
Pscheidl: One solution to keep the old behavior and to make it less cofusing is to named the parameter `custom_non_data_line_markers`.

Pscheidl: See commit: https://github.com/h2oai/h2o-3/pull/3380/commits/827331e88c90f0317ec303b23a80aaf5fa9a6470

h2o-core/src/main/java/water/parser/ParseSetup.java
Pscheidl: Thanks for pointing it out. Turns out the constructor injection can be "eased" by deleting some of the older ones. Those were not marked as unused code, but the ones that remained after commit https://github.com/h2oai/h2o-3/pull/3380/commits/0e610d07febddbc27082c31202d9bbbc4ddcbe7b are completely OK.

Pscheidl:
@michalkurka @honzasterba Please re-review

h2o-py/h2o/h2o.py
```diff
@@ -400,6 +400,7 @@ def import_file(path=None, destination_frame=None, parse=True, header=0, sep=Non
     :param pattern: Character string containing a regular expression to match file(s) in the folder if `path` is a
         directory.
     :param skipped_columns: an integer list of column indices to skip and not parsed into the final frame from the import file.
+    :param non_data_line_markers: If a line in imported file starts with any character in given string it will be imported. Empty string means all lines are imported, None means that default behaviour for given format will be used
```
honzasterba: I think this is not correct, correct is:

If a line in imported file starts with any character in given string it will NOT be imported.

h2o-py/h2o/h2o.py
```diff
@@ -555,7 +557,7 @@ def import_sql_select(connection_url, select_query, username, password, optimize
 
 
 def parse_setup(raw_frames, destination_frame=None, header=0, separator=None, column_names=None,
-                column_types=None, na_strings=None, skipped_columns=None):
+                column_types=None, na_strings=None, skipped_columns=None, custom_non_data_line_markers = None):
```
honzasterba: does not match formatting of previous args

h2o-py/h2o/h2o.py
```diff
@@ -591,6 +593,7 @@ def parse_setup(raw_frames, destination_frame=None, header=0, separator=None, co
     :param na_strings: A list of strings, or a list of lists of strings (one list per column), or a dictionary
         of column names to strings which are to be interpreted as missing values.
     :param skipped_columns: an integer lists of column indices to skip and not parsed into the final frame from the import file.
+    :param custom_non_data_line_markers: If a line in imported file starts with any character in given string it will be imported. Empty string means all lines are imported, None means that default behaviour for given format will be used
```
honzasterba: same as above

h2o-r/h2o-package/R/import.R
```diff
@@ -55,6 +55,7 @@
 #' @param decrypt_tool (Optional) Specify a Decryption Tool (key-reference
 #'        acquired by calling \link{h2o.decryptionSetup}.
 #' @param skipped_columns a list of column indices to be skipped during parsing.
+#' @param custom_non_data_line_markers (Optional) If a line in imported file starts with any character in given string it will be imported. Empty string means all lines are imported, None means that default behaviour for given format will be used
```
honzasterba: same as above + R does not have None but NULL

h2o-r/h2o-package/R/import.R
```diff
@@ -121,7 +125,8 @@ h2o.importFolder <- function(path, pattern = "", destination_frame = "", parse =
 if(parse) {
     srcKey <- res$destination_frames
     return( h2o.parseRaw(data=.newH2OFrame(op="ImportFolder",id=srcKey,-1,-1),pattern=pattern, destination_frame=destination_frame,
-            header=header, sep=sep, col.names=col.names, col.types=col.types, na.strings=na.strings, decrypt_tool=decrypt_tool, skipped_columns=skipped_columns) )
+            header=header, sep=sep, col.names=col.names, col.types=col.types, na.strings=na.strings, decrypt_tool=decrypt_tool,
+             skipped_columns=skipped_columns, custom_non_data_line_markers = custom_non_data_line_markers) )
```
honzasterba: whitespace

h2o-r/h2o-package/R/parse.R
```diff
@@ -30,22 +30,25 @@
 #'        acquired by calling \link{h2o.decryptionSetup}.
 #' @param chunk_size size of chunk of (input) data in bytes
 #' @param skipped_columns a list of column indices to be excluded from parsing
+#' @param custom_non_data_line_markers (Optional) If a line in imported file starts with any character in given string it will be imported. Empty string means all lines are imported, None means that default behaviour for given format will be used
```
honzasterba: same as above

h2o-r/h2o-package/R/parse.R
```diff
@@ -140,6 +147,8 @@ h2o.parseSetup <- function(data, pattern="", destination_frame = "", header = NA
   # begin the setup
   # setup the parse parameters here
   parseSetup.params <- list()
+  
+  if(!is.null(custom_non_data_line_markers)) {parseSetup.params$custom_non_data_line_markers = custom_non_data_line_markers}
```
honzasterba: unnecessary {}, I would either remove the {} or put the statement on a new line

h2o-py/h2o/h2o.py
Pscheidl: Sorry, I copy&pasted your proposal without reading it.

h2o-py/h2o/h2o.py
Pscheidl: Resolved.

h2o-py/h2o/h2o.py
Pscheidl: Resolved.

h2o-r/h2o-package/R/import.R
Pscheidl: Resolved.

h2o-r/h2o-package/R/import.R
Pscheidl: Removed

h2o-r/h2o-package/R/parse.R
Pscheidl: Resolved, added 'not' plus 'null'

h2o-r/h2o-package/R/parse.R
Pscheidl: Style guidelines do not forbid it (https://style.tidyverse.org/syntax.html#indenting - section 2.4.1) . However, the lines was **too long**, so I chose the s econd option.

Pscheidl:
@honzasterba Thanks for the comments again. Is it worthy merging now ?

Gandalf

h2o-r/h2o-package/R/parse.R
```diff
@@ -67,6 +70,9 @@ h2o.parseRaw <- function(data, pattern="", destination_frame = "", header=NA, se
             decrypt_tool = .decrypt_tool_id(parse.params$decrypt_tool),
             skipped_columns = paste0("[", paste(parse.params$skipped_columns, collapse=','), "]")
             )
+   if(!is.null(custom_non_data_line_markers)){
+    parse.params <- append(parse.params,list(custom_non_data_line_markers = custom_non_data_line_markers))
```
honzasterba: whitespace, we use 2 space indent

h2o-r/h2o-package/R/parse.R
```diff
@@ -140,6 +147,10 @@ h2o.parseSetup <- function(data, pattern="", destination_frame = "", header = NA
   # begin the setup
   # setup the parse parameters here
   parseSetup.params <- list()
+  
+  if(!is.null(custom_non_data_line_markers)) {
+   parseSetup.params$custom_non_data_line_markers = custom_non_data_line_markers
```
honzasterba: whitespace, we use 2 space indent

h2o-r/h2o-package/R/parse.R
```diff
@@ -140,6 +147,10 @@ h2o.parseSetup <- function(data, pattern="", destination_frame = "", header = NA
...
+   }
```
honzasterba: whitespace

h2o-r/h2o-package/R/parse.R
Pscheidl: This place has +2 spaces indentation. I see no problem here.

h2o-r/h2o-package/R/parse.R
```diff
@@ -127,7 +133,8 @@ h2o.parseRaw <- function(data, pattern="", destination_frame = "", header=NA, se
 #' @seealso \link{h2o.parseRaw}
 #' @export
 h2o.parseSetup <- function(data, pattern="", destination_frame = "", header = NA, sep = "", col.names = NULL, col.types = NULL,
-                           na.strings = NULL, parse_type = NULL, chunk_size = NULL, decrypt_tool = NULL, skipped_columns=NULL) {
+                           na.strings = NULL, parse_type = NULL, chunk_size = NULL, decrypt_tool = NULL, skipped_columns=NULL,
+                           custom_non_data_line_markers=NULL) {
```
Pscheidl: This function has spaces around `=` operator. Some functions have it, some don't. The style is not really unified. To be consistent with function's header, let's change it.

h2o-r/h2o-package/R/parse.R
Pscheidl: Re-formatted in R studio. For some reason. GitHub won't show the change 

Snímek z 2019-03-27 14-35-41

Pscheidl: I have no idea why the change is not shown. It has proper indentation after the re-format. Everything else is ok. Everything is pushed.



I even broken it intentionally in https://github.com/h2oai/h2o-3/pull/3380/commits/82c557a520572cbc98a0e238aeb7fc06cb627f25 and then fixed it back in https://github.com/h2oai/h2o-3/pull/3380/commits/cd6352c108608c7c5a9ea48f32106709612fba54 - no luck.

Pscheidl: Great, now it is shown correctly even here, not sure what was wrong with github.

Pscheidl: Here, it looks like `  if(!is.null(custom_non_data_line_markers)){` has three spaces before it, in reality it only has two. See the commit links ^^^

h2o-py/h2o/h2o.py
```diff
@@ -591,6 +593,7 @@ def parse_setup(raw_frames, destination_frame=None, header=0, separator=None, co
     :param na_strings: A list of strings, or a list of lists of strings (one list per column), or a dictionary
         of column names to strings which are to be interpreted as missing values.
     :param skipped_columns: an integer lists of column indices to skip and not parsed into the final frame from the import file.
+    :param custom_non_data_line_markers: If a line in imported file starts with any character in given string it will NOT be imported. Empty string means all lines are imported, None means that default behaviour for given format will be used
```
ledell:
```suggestion
    :param custom_non_data_line_markers: If a line in an imported file starts with any character in the given string, it will be skipped and NOT be imported. An empty string means all lines are imported, None means that default behavior for given format will be used.
```

h2o-r/h2o-package/R/parse.R
```diff
@@ -30,22 +30,25 @@
 #'        acquired by calling \link{h2o.decryptionSetup}.
 #' @param chunk_size size of chunk of (input) data in bytes
 #' @param skipped_columns a list of column indices to be excluded from parsing
+#' @param custom_non_data_line_markers (Optional) If a line in imported file starts with any character in given string it will NOT be imported. Empty string means all lines are imported, NULL means that default behaviour for given format will be used
```
ledell:
```suggestion
#' @param custom_non_data_line_markers (Optional) If a line in an imported file starts with any character in the given string, it will be skipped and NOT be imported. An empty string means all lines are imported, NULL means that default behavior for given format will be used.
```

h2o-py/h2o/h2o.py
```diff
@@ -400,6 +400,7 @@ def import_file(path=None, destination_frame=None, parse=True, header=0, sep=Non
     :param pattern: Character string containing a regular expression to match file(s) in the folder if `path` is a
         directory.
     :param skipped_columns: an integer list of column indices to skip and not parsed into the final frame from the import file.
+    :param custom_non_data_line_markers: If a line in imported file starts with any character in given string it will NOT be imported. Empty string means all lines are imported, None means that default behaviour for given format will be used
```
ledell:
```suggestion
    :param custom_non_data_line_markers: If a line in an imported file starts with any character in the given string, it will be skipped and NOT be imported. An empty string means all lines are imported, None means that default behavior for given format will be used.
```

PR #3381, opened 2019-03-25
PUBDEV-6374: Allow users to define custom HTTP headers

michalkurka:
Example:

java -jarh2o.jar -add_http_header 'Strict-Transport-Security' 'max-age=3600' -jks ../mykeystore.jks


michalkurka:
@jakubhava how can we make it accessible to SW users?

jakubhava:
We can expose it via `spark.driver.extraJavaOptions=-Dai.h2o..` and  `spark.driver.extraJavaOptions=-Dai.h2o..`

PR #3382, opened 2019-03-26
PUBDEV-6376: StackedEnsemble prediction fails when applied to a test dataset without response column

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6376

ensured that GLM respect computeMetrics parameter when it doesn't have direct access to original frame (like in SE scoring)

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLMModel.java
+++ b/h2o-algos/src/main/java/hex/glm/GLMModel.java
@@ -1431,10 +1431,10 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
       adaptFrm.remove(responseId);
     }
     // Build up the names & domains.
-    final boolean computeMetrix = computeMetrics && (adaptFrm.vec(_output.responseName()) != null && !adaptFrm.vec(_output.responseName()).isBad());
-    String [] domain = _output.nclasses()<=1 ? null : !computeMetrix ? _output._domains[_output._domains.length-1] : adaptFrm.lastVec().domain();
+    final boolean detectedComputeMetrics = computeMetrics && (adaptFrm.vec(_output.responseName()) != null && !adaptFrm.vec(_output.responseName()).isBad());
+    String [] domain = _output.nclasses()<=1 ? null : !detectedComputeMetrics ? _output._domains[_output._domains.length-1] : adaptFrm.lastVec().domain();
     // Score the dataset, building the class distribution & predictions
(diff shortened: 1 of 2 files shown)


seb-h2o:
@michalkurka please have a look: tiny fix but very important! thx

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1452,7 +1452,7 @@ protected Frame predictScoreImpl(Frame fr, Frame adaptFrm, String destination_ke
     GLMScore gs = makeScoringTask(adaptFrm,true,j);// doAll(names.length,Vec.T_NUM,adaptFrm);
     assert gs._dinfo._valid:"_valid flag should be set on data info when doing scoring";
     gs.doAll(names.length,Vec.T_NUM,gs._dinfo._adaptedFrame);
-    if (gs._computeMetrics)
+    if (computeMetrics && gs._computeMetrics)
```
michalkurka: It seems like a better fix would be to propagate the flag to `makeScoringTask` - because GLMScore will actually run `_mb.perRow` to prepare the metadata for the metrics

seb-h2o: @michalkurka:  yes, wanted to minimize changes but this would be more robust. 
Will propagate then.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1424,17 +1424,17 @@ public void generate(JCodeSB out) {
 
 
 
-  private GLMScore makeScoringTask(Frame adaptFrm, boolean generatePredictions, Job j){
+  private GLMScore makeScoringTask(Frame adaptFrm, boolean generatePredictions, Job j, boolean computeMetrics){
```
michalkurka: if we call the parameter `requestMetrics` we won't have the name clash

michalkurka:
I think the GLM change makes sense. However, why does GLM thinks it does have a response when SE doesn't?

seb-h2o:
for documentation purpose, answering @michalkurka question as we had a separate discussion about GLM behaviour:
GLM was not considering the adapted response column as "Bad", if it were in this state then it would have properly skipped metrics computation.
However, this is still a separate issue that is fixed by https://0xdata.atlassian.net/browse/PUBDEV-6379.
cf. also https://github.com/h2oai/h2o-3/pull/3383

michalkurka:
underlying cause fixed in https://github.com/h2oai/h2o-3/pull/3383

we shouldn't need the urgently anymore

fix makes sense, I would just clean-up the naming computeMetrics vs computeMetrix

PR #3274, opened 2019-02-01
PUBDEV-6136 - Mojo delegating generic model

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6136

PR #3384, opened 2019-03-26
PUBDEV-6378 - Generic model release notes (PUBDEV-6136)

Pscheidl:
### Release notes PR
JIRA: [https://0xdata.atlassian.net/browse/PUBDEV-6378](https://0xdata.atlassian.net/browse/PUBDEV-6378)

### PR that is actually being documented
JIRA: [PUBDEV-6136](https://0xdata.atlassian.net/browse/PUBDEV-6136)
GitHub PullRequest: https://github.com/h2oai/h2o-3/pull/3274

### DIFF
--- a/h2o-docs/src/product/data-science/generic.rst
+++ b/h2o-docs/src/product/data-science/generic.rst
@@ -41,11 +41,11 @@ Importing in R or Python
    .. code-block:: python
 
     data = h2o.import_file(path='training_dataset.csv')
-    model = H2OGeneralizedLinearEstimator()
-    model.train(x = ["Some column", "Another column"], y = "response", training_frame=data)
+    original_model = H2OGeneralizedLinearEstimator()
+    original_model.train(x = ["Some column", "Another column"], y = "response", training_frame=data)
 
     path = '/path/to/model/directory/model.zip'
-    model.download_mojo(path)
+    original_model.download_mojo(path)
 
     generic_model = H2OGenericEstimator.from_file(path)
     new_observations = h2o.import_file(path='new_observations.csv')
@@ -86,11 +86,11 @@ every time a Generic model is created.
    .. code-block:: python
 
     data = h2o.import_file(path='training_dataset.csv')
-    model = H2OGeneralizedLinearEstimator()
-    model.train(x = ["Some column", "Another column"], y = "response", training_frame=data)
+    original_model = H2OGeneralizedLinearEstimator()
+    original_model.train(x = ["Some column", "Another column"], y = "response", training_frame=data)
 
     path = '/path/to/model/directory/model.zip'
-    model.download_mojo(path)
+    original_model.download_mojo(path)
     
     imported_mojo_key = h2o.lazy_import(file)
     generic_model = H2OGenericEstimator(model_key = get_frame(model_key[0]))
--- a/h2o-docs/src/product/data-science/generic.rst
+++ b/h2o-docs/src/product/data-science/generic.rst
@@ -28,27 +28,27 @@ embraces it for the purpose of scoring. Information output about the model may b
 
     data <- h2o.importFile(path = 'training_dataset.csv')
     cols <- c("Some column", "Another column")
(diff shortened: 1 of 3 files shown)


h2o-docs/src/product/data-science/generic.rst
```diff
@@ -0,0 +1,92 @@
...
+
+    data <- h2o.importFile(path = 'training_dataset.csv')
+    cols <- c("Some column", "Another column")
+    model.original <- h2o.glm(x=cols, y = "response", training_frame = data)    
+
+    path <- "/path/to/model/directory"
+    mojo.destination <- h2o.download_mojo(model = model.original, path = path)
+    
+    # Only import or upload MOJO model data, do not initialize the generic model yet
+    imported_mojo_key <- h2o.importFile(mojo.destination, parse = FALSE)
```
michalkurka: Would be nice to combine the next 2 lines into a single convenient function. The users shouldn't need to worry about parse = FALSE

+ the mojo should be deleted from memory when the model gets deleted

h2o-docs/src/product/data-science/generic.rst
```diff
@@ -0,0 +1,92 @@
...
+   .. code-block:: python
+
+    data = h2o.import_file(path='training_dataset.csv')
+    glm = H2OGeneralizedLinearEstimator()
+    glm.train(x = ["Some column", "Another column"], y = "response", training_frame=data)
+
+    path = '/path/to/model/directory/model.zip'
+    drf.download_mojo(path)
+    
+    imported_mojo_key = h2o.lazy_import(file)
```
michalkurka: same as above - make it easy for the users

h2o-docs/src/product/data-science/generic.rst
Pscheidl: There is such convenience function. This is only the advanced example when the MOJO already is in-memory for ANY reason. Look above.

h2o-docs/src/product/data-science/generic.rst
Pscheidl: Same as above, read the previous part of the documentation. The convenience function comes first, then this.

h2o-docs/src/product/data-science/generic.rst
Pscheidl: See the `Importing a Generic Model` section.

h2o-docs/src/product/data-science/generic.rst
Pscheidl: See the `Importing a Generic Model` section

h2o-docs/src/product/data-science/generic.rst
```diff
@@ -0,0 +1,98 @@
...
+.. example-code::
+   .. code-block:: r
+
+    data <- h2o.importFile(path = 'training_dataset.csv')
+    cols <- c("Some column", "Another column")
+    model.original <- h2o.glm(x=cols, y = "response", training_frame = data)    
+
+    path <- "/path/to/model/directory"
+    mojo.destination <- h2o.download_mojo(model = model.original, path = path)
+    model.generic <- h2o.genericModel(mojo.destination)
```
ledell: My only comment would be to not use dots in object names (this goes against the [tidyverse style guide](https://style.tidyverse.org/), so I'd change this to `model_original`, `mojo_destination` and `model_generic`.

ledell: Also ideally we would use the same object names in the R and Python examples (e.g. `glm` and `model` in Python vs `model_original` and `model_generic` in R) when we do side-by-side examples in the docs.

Pscheidl: Resolved. Dots removed, variable names aligned.

Pscheidl:
All comments addressed.

h2o-docs/src/product/data-science/generic.rst
michalkurka: ok, thanks for explaining

PR #3385, opened 2019-03-26
PUBDEV-6357: Categorical value out of bounds error when calling model using Python

michalkurka:
https://0xdata.atlassian.net/browse/PUBDEV-6357


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -273,7 +270,7 @@ private static int getNzCount(ZeroWeight zw, List<SparseItem> sparseItems, int n
      * @return DMatrix
      * @throws XGBoostError
      */
-    public static DMatrix convertChunksToDMatrix(Key<DataInfo> dataInfoKey,
+    public static DMatrix convertChunksToDMatrix(DataInfo di,
```
Pscheidl: You're probably missing `assert di != null;` to be consistent with `convertFrameToDMatrix` ^^^.

I'd even consider checking this all the time, e.g. Objects.requireNotNull.

PR #3390, opened 2019-03-27
[PUBDEV-6358] drop correct temp table on JDBC import

honzasterba:


honzasterba:
discovered by failing hadoop kerberos pipeline

PR #3393, opened 2019-03-28
PUBDEV-6382: Extend Py API for TargetEncoder with convenient method for getting encoding maps

deil87:
It will make it easier to access encoding maps on the client side


deil87:
WIP

PR #3394, opened 2019-03-28
PUBDEV-6383: SMBO support for target encoding hyperparameters search

deil87:
This is a copy of the branch from following PR https://github.com/h2oai/h2o-3/pull/3183 .
I will proceed here with SMBO support as we will go with random grid search for first version of integration of TE into AutoML.


h2o-algos/src/main/java/hex/tree/drf/DRF.java
```diff
@@ -155,8 +163,23 @@ private void initTreeMeasurements() {
...
+      boolean collectConfidenceData = true; // TODO parameter
+
+      String id_for_confidence_metrics = "id_for_confidence_metrics";
+      if(collectConfidenceData && _train.find(id_for_confidence_metrics) == -1) {
+        _train.add(id_for_confidence_metrics, Vec.makeCon(Vec.newKey(), idx));
+      }
+      int indexOfIdColumn = _train.find(id_for_confidence_metrics);
+      CollectPreds cp = new CollectPreds(ktrees,leafs,_model.defaultThreshold(), indexOfIdColumn).doAll(_train,_parms._build_tree_one_node);
 
+      IcedHashMap<Integer, Double[]> predictionsMap = cp._predictionsMap;
```
michalkurka: Why do we need Double[] array? Can we use a primitive array instead?

deil87: @michalkurka At the beginning I was using different types for different parameters and my Map was storing values as Object[]. Later when I added Surrogate Model I needed to convert my grid  into Frame and then be able to extract rows from this Frame....and there are only `.atd(index)`, `.at8(index)` or something like this which return only doubles or long. So I decided to use single representation with doubles. Yes with current version probably there is no point in having Doubles.

PR #3396, opened 2019-03-28
PUBDEV-4659 fix predict on test data

maurever:
Fix the predict method on test data without the response column.


h2o-algos/src/main/java/hex/DataInfo.java
```diff
@@ -521,7 +521,7 @@ public DataInfo filterExpandedColumns(int [] cols){
       ignoredCols[ignoredCnt++] = k+_cats;
     Frame f = new Frame(_adaptedFrame.names().clone(),_adaptedFrame.vecs().clone());
     if(ignoredCnt > 0) f.remove(Arrays.copyOf(ignoredCols,ignoredCnt));
-    assert catLvls.length < f.numCols():"cats = " + catLvls.length + " numcols = " + f.numCols();
+    assert catLvls.length <= f.numCols():"cats = " + catLvls.length + " numcols = " + f.numCols();
```
michalkurka: This makes sense but it would be good to add a small unit test to check it.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1204,7 +1206,6 @@ else if (isOffset)
         else if (isWeights && computeMetrics) {
           if (expensive) {
             vec = test.anyVec().makeCon(1);
-            toDelete.put(vec._key, "adapted missing vectors");
```
michalkurka: this looks suspicious, why is it okay to not delete the key?

h2o-r/tests/testdir_jira/runit_pubdev_4659.R
```diff
@@ -0,0 +1,42 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../scripts/h2o-r-test-setup.R")
+
+
+test_4659 <- function() {
+    prostate = read.csv(locate("smalldata/logreg/prostate.csv"))
+    prostate$our_factor<-as.factor(paste0("Q",c(rep(c(1:380),1))))
+
+    prostate<-as.h2o(prostate)
+    prostate$weight<-1
```
michalkurka: random weights would be better (future-proof)

h2o-r/tests/testdir_jira/runit_pubdev_4659.R
```diff
@@ -0,0 +1,42 @@
...
+    
+    prostate_train<-prostate[1:300,]
+    prostate_test<-prostate[301:380,]
+
+    # delete response variable from test data
+    prostate_test<-prostate_test[,-3]
+    
+    print("GLM 1")
+    model<-h2o.glm(y = "AGE", x = c("our_factor"), training_frame = prostate_train,offset_column="weight")
```
michalkurka: it is better not to reuse weights for offset column

offset column is confusing to people, we shouldn't publish examples that put weight column into offset

h2o-core/src/main/java/hex/Model.java
maurever: I accidentally deleted it. Fortunately, we have reviews. :)

h2o-r/tests/testdir_jira/runit_pubdev_4659.R
maurever: Yes, that is true, but for the test, it is important the presence of the column, it does not matter what it contains.

h2o-r/tests/testdir_jira/runit_pubdev_4659.R
maurever: It is also true; I will change it. I only copied paste the test case from the JIRA without thinking about it.

h2o-algos/src/main/java/hex/DataInfo.java
maurever: There is the test case (the second one) which test it in runit_pubdev_4659.R. It is the case where there is only one column of data and no response column in test data (no weights, offset column, etc.).

PR #3400, opened 2019-03-28
PUBDEV-6377 - Filling of sparse DMatrix data

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6377

### DIFF
--- a/h2o-core/src/test/java/water/TestUtil.java
+++ b/h2o-core/src/test/java/water/TestUtil.java
@@ -1050,18 +1050,7 @@ public class TestUtil extends Iced {
       final Vec f1Vec = f1.vec(vecNum);
       final Vec f2Vec = f2.vec(vecNum);
 
-      if (f1Vec.length() != f2Vec.length()) {
-        throw new IllegalStateException(String.format("Length of vectors in column %o is not equal: {%o,%o}",
-                vecNum, f1Vec.length(), f2Vec.length()));
-      }
-
-      for (int i = 0; i < f1Vec.length(); i++) {
-        if (f1Vec.at(i) != f2Vec.at(i)) {
-          throw new IllegalStateException(String.format("Not equal values found in column %o: {%f, %f}",
-                  vecNum,f1Vec.at(i), f2Vec.at(i)));
-        }
-      }
-
+      assertVecEquals(f1Vec, f2Vec, 0);
     }
 
     return true;
--- a/h2o-core/src/test/java/water/TestUtil.java
+++ b/h2o-core/src/test/java/water/TestUtil.java
@@ -1027,5 +1027,44 @@ public class TestUtil extends Iced {
     }
     return result;
   }
+
+  /**
+   * @param f1 Frame to be comapared, not null
+   * @param f2 Frame to be compared, not null
+   * @return True if matrices are precisely the same - number of columns, rows & values at each cell.
+   * @throws IllegalStateException If any inequalities are found
+   */
+  public static boolean compareFrames(final Frame f1, final Frame f2) throws IllegalStateException {
+    Objects.requireNonNull(f1);
+    Objects.requireNonNull(f2);
+
+    if (f1.numCols() != f2.numCols())
+      throw new IllegalStateException(String.format("Number of columns is not the same: {%o, %o}",
+              f1.numCols(), f2.numCols()));
+    if (f1.numRows() != f2.numRows())
+      throw new IllegalStateException(String.format("Number of rows is not the same: {%o, %o}",
+              f1.numRows(), f2.numRows()));
+
+    for (int vecNum = 0; vecNum < f1.numCols(); vecNum++) {
+
+      final Vec f1Vec = f1.vec(vecNum);
+      final Vec f2Vec = f2.vec(vecNum);
+
+      if (f1Vec.length() != f2Vec.length()) {
+        throw new IllegalStateException(String.format("Length of vectors in column %o is not equal: {%o,%o}",
+                vecNum, f1Vec.length(), f2Vec.length()));
+      }
+
+      for (int i = 0; i < f1Vec.length(); i++) {
+        if (f1Vec.at(i) != f2Vec.at(i)) {
+          throw new IllegalStateException(String.format("Not equal values found in column %o: {%f, %f}",
+                  vecNum,f1Vec.at(i), f2Vec.at(i)));
+        }
+      }
+
+    }
+
+    return true;
+  }
   
 }
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
@@ -96,7 +96,6 @@ public class XGBoostUtils {
         if (weightVector != null) {
             weights = malloc4f(nRows);
         }
-        sparse = true;
         if (sparse) {
             Log.debug("Treating matrix as sparse.");
             // 1 0 2 0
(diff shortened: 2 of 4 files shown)


Pscheidl:
WIP.

Current status causes `malloc(): memory corruption`. ARRAY_SIZE is temporarily lowered to 10 to try out all the functionality. There should've never been problem with row indices in CSR, as the number of rows never exceeded length of first row of the two-dim array  there (the two-dim array in DMatrix for row indices is pointless).

However, it seems like the data[][] matrix was never able to reach second row, which is t**erribly bad mistake**.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -96,7 +96,7 @@ public static DMatrix convertFrameToDMatrix(DataInfo di,
         if (weightVector != null) {
             weights = malloc4f(nRows);
         }
-
+        sparse = true;
```
Pscheidl: WIP. Helped me debug it.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -534,65 +537,92 @@ private static int initalizeFromChunkIds(Frame f, int[] chunks, Vec.Reader[] vec
...
                     if (val != 0) {
-                        data[currentRow][currentCol] = val;
-                        colIndex[currentRow][currentCol++] = di._catOffsets[di._catOffsets.length - 1] + j;
+                        data[rowPointer][colPointer] = val;
+                        colIndex[rowPointer][colPointer++] = di._catOffsets[di._catOffsets.length - 1] + j;
                         nonZeroCount++;
                     }
                 }
-
-                rowHeaders[0][++actualRows] = nonZeroCount;
```
Pscheidl: This was usually sufficient, as the number of rows was never larger than the first dimension of this arrray.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -534,65 +537,92 @@ private static int initalizeFromChunkIds(Frame f, int[] chunks, Vec.Reader[] vec
...
 
                 for (int j = 0; j < di._cats; ++j) {
-                    data[currentRow][currentCol] = 1; //one-hot encoding
+                    if(colPointer == ARRAY_MAX){
+                        colPointer = 0;
+                        rowPointer++;
+                    }
+                    data[rowPointer][colPointer] = 1; //one-hot encoding
                     if (vecs[j].isNA(i)) {
-                        colIndex[currentRow][currentCol++] = di.getCategoricalId(j, Float.NaN);
```
Pscheidl: currentRow was ALWAYS 0. Always, always, always. Big mistake.

Pscheidl:
At this point, there is one test on dense data (intentionally), force to be treated as sparse matrix.  Actually, there is not a single zero in the dataset (cars used).

Used it as an "integration" test for current status, as I do think currently the indices are filled correctly and the data matrix is ok. For some reason, the native part fails and indicates one row less than it should be. Scoring is also broken.

Either my new way of filling CSR structures is wrong, or there is something wrong in other places as well.

_In this phase, let's skip the fact there is a lot of inherited code duplication and the whole thing could be nicer. Functionality first._

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -522,74 +521,132 @@ private static int initalizeFromChunkIds(Frame f, int[] chunks, Vec.Reader[] vec
...
+            boolean foundNonZero = false;
 
             for (int j = 0; j < di._cats; ++j) {
-                data[currentRow][currentCol] = 1; //one-hot encoding
+                if(currentCol == SPARSE_MATRIX_DIM){
+                    currentCol = 0;
+                    rowPointer++;
+                }
+                data[rowPointer][currentCol] = 1; //one-hot encoding
+                if(!foundNonZero){
```
Pscheidl: The FIRST non-zero character belongs to these row-related indices (pointing to the index of the non-zero element just discovered in the array).

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -522,74 +521,132 @@ private static int initalizeFromChunkIds(Frame f, int[] chunks, Vec.Reader[] vec
...
+                    foundNonZero = true;
+                    for (int k = lastNonZeroRow; k < actualRows; k++) {
+                        rowHeaders[rowHeaderRowPointer][rowHeaderColPointer++] = nonZeroCount;
```
Pscheidl: Multiple lines may contain zeroes. The value of the very next non-zero element must be filled for all the previous lines with all zeroes.

Pscheidl:
https://imada.sdu.dk/~marco/DM545/Resources/Ipython/Sparse.php

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostUtilsTest.java
```diff
@@ -46,6 +69,116 @@ public void parseFeatureScores() throws IOException, ParseException {
...
+      Scope.track(testFrame);
+
+      testCSRPredictions(frame, response, testFrame );
+    } finally {
+      Scope.exit();
+    }
+  }
+
+  @Test
+  public void testCSRPredictionComparison_synthetic_sparse() {
```
Pscheidl: This test will currently fail, predictions of dense and sparse are not the same.

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -1027,5 +1027,44 @@ protected static RowData toRowData(Frame fr, String[] columns, long row) {
...
+      final Vec f1Vec = f1.vec(vecNum);
+      final Vec f2Vec = f2.vec(vecNum);
+
+      if (f1Vec.length() != f2Vec.length()) {
+        throw new IllegalStateException(String.format("Length of vectors in column %o is not equal: {%o,%o}",
+                vecNum, f1Vec.length(), f2Vec.length()));
+      }
+
+      for (int i = 0; i < f1Vec.length(); i++) {
+        if (f1Vec.at(i) != f2Vec.at(i)) {
```
michalkurka: there is already a function that compares Vecs, please re-use it, this might have issues with NAs also

Pscheidl: True. However, this will be deleted. These tests are meaningless, as native sparse may produce different results than dense in some cases.

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostUtilsTest.java
```diff
@@ -56,5 +174,338 @@ public void parseFeatureScores() throws IOException, ParseException {
...
+
+  @RunWith(Parameterized.class)
+  public static final class XGBoostSparseMatrixTest extends XGBoostUtilsTest {
+
+    private static final int MAX_ARR_SIZE = Integer.MAX_VALUE - 10;
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+      return Arrays.asList(new Object[][]{
+              {30, 10, MAX_ARR_SIZE},
```
Pscheidl: Goal is to test native predictions trained on matrix with various layout (effectively 1D, 2D) of internal memory representation. Same applies for H2O DMatrix construction. Predictions should be the same each time.

Pscheidl:
This PR is being closed because:

1. New PR against rel-yates is being opened (can be ofc. changed without closing)
2. We need benchmarks running before this is merged and there is a bug in our pipeline, preventing us from running it against branches with a slash in their name.

PR #3405, opened 2019-03-28
PUBDEV-6386: XGBoost MOJO should use Java predictor by default

michalkurka:
Java predict is now proven and should be used as the default implementation for MOJO scoring. This also gives people access to TreeSHAP by default.

### DIFF
--- a/h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/DefaultMojoImplTest.java
+++ b/h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/DefaultMojoImplTest.java
@@ -1,0 +1,103 @@
+package hex.tree.xgboost;
+
+import hex.ModelMetricsRegression;
+import hex.SplitFrame;
+import hex.genmodel.MojoModel;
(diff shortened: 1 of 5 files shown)


h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/DefaultMojoImplTest.java
```diff
@@ -0,0 +1,103 @@
...
+      MojoModel mojo = model.toMojo();
+      assertEquals(XGBoostNativeMojoModel.class.getName(), mojo.getClass().getName());
+    } finally {
+      Scope.exit();
+    }
+  }
+
+  /* FIXME: PUBDEV-6387: gblinear causes a segmenation fault
+  @Test
+  public void testGBLinear() throws IOException {
```
Pscheidl: Delete it or uncomment it ?

michalkurka: This should stay, it cannot be uncommented until PUBDEV-6387 gets fixed.

PR #3407, opened 2019-03-29
[BUILD] fix hadoop tests on some hdp2.x builds

honzasterba:
- on 2.2 issues with access righhts, disable hive impersonation everywhere
- on 2.3,4,5 the classpath for hive jdbc is missing xerces impl add it manually

### DIFF
--- a/docker/hadoop/common/conf-hive-kerberos/hive-site.xml
+++ b/docker/hadoop/common/conf-hive-kerberos/hive-site.xml
@@ -45,6 +45,10 @@
         <name>hive.users.in.admin.role</name>
         <value>hdfs,hive</value>
     </property>
+    <property>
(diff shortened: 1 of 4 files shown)


docker/hadoop/common/conf-hive-kerberos/hive-site.xml
```diff
@@ -45,6 +45,10 @@
         <name>hive.users.in.admin.role</name>
         <value>hdfs,hive</value>
     </property>
+    <property>
```
michalkurka: does this do anything? It just seems to be moved

honzasterba: yes, just moved to be in the same place for both config files

PR #3409, opened 2019-03-29
PUBDEV-6040: Updates to Target Encoding Topic

angela0xdata:
- Update R examples to reflect the new API.
- Added Python examples. Note that the Python examples are faillng, though.


h2o-docs/src/product/data-munging/target-encoding.rst
```diff
@@ -146,136 +216,144 @@ Now we will perform target encoding on ``addr_state`` to see if this representat
...
-    ## 1         AK    0               0.1212239
-    ## 2         AK    0               0.1212239
-    ## 3         AK    0               0.1212239
-    ## 4         AK    0               0.1212239
-    ## 5         AK    0               0.1212239
-    ## 6         AK    0               0.1212239
+   .. code-block:: python
 
-For our testing data, we will use the parameters:
+    ext_train = TargetEncoder(x=predictor, y=response, target_encode_map=te_map,
```
deil87: @angela0xdata please read the tests which I have sent you names of before carefully. `titanic_with_te_kfoldstrategy` test in `/h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_automl_gbm_target_encoding.py` might help you the most.  
Pay attention to parameters that you are passing and to what objects. You should not pass 'target_encode_map=te_map,' to `TargetEncoder` object. Also when you do `transform` you should specify not only frame you are going to apply target encoding to.

angela0xdata: @deil87 QQ. Why does R use a target_encode_map but Python does not?

angela0xdata: @deil87 NVM... I see now.

deil87: @angela0xdata PR has WIP label. Is it ready for review?

angela0xdata: yes, i've requested a review. I'm still getting errors with my Python examples. I also get errors when I run this python unit test: https://github.com/h2oai/h2o-3/blob/78dbc189c3d6273d8f7b8a61a0199eb5c7c75d5a/h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_automl_gbm_target_encoding.py

deil87: @angela0xdata tests in `pyunit_automl_gbm_target_encoding.py` should be fine as Mr. Jenkins check them. It might mean that something is wrong with your local setup. Maybe you are using old version of the backend. Can you share errors that you are getting in Slack?

angela0xdata: @deil87 I was trying to run that test as a user would (so like a demo), not as Jenkins would, so I tried to make some adjustments. But I'm no expert.

h2o-docs/src/product/data-munging/target-encoding.rst
```diff
@@ -146,136 +216,152 @@ Now we will perform target encoding on ``addr_state`` to see if this representat
...
+   .. code-block:: r
+
+    ext_test <- h2o.target_encode_transform(test, x = list("addr_state"), y = response,
+                                            target_encode_map = te_map, holdout_type = "none",
+                                            fold_column = "fold", noise = 0,
+                                            blended_avg = FALSE, seed=1234)
+
+   .. code-block:: python
+
+    ext_test = TargetEncoder(x=predictor, y=response, target_encode_map=te_map,
```
deil87: @angela0xdata We should not pass `target_encode_map=te_map` and `holdout_type="none"` to TargetEncoder class. TargetEncoder creates `te_map` and holdout_type is needed only during application of the encoding map when we call `.transform()`

deil87:
@angela0xdata I have added python demo tests for bad_loan dataset.

h2o-docs/src/product/data-munging/target-encoding.rst
angela0xdata: Thanks for the test. It's helpful (though there are some references to Titanic in it), and I made  a small change, but I'm not able to run that test locally. 
When running the doc example, the following still results in an error.

```
te_map = TargetEncoder(x=predictor, y=response, fold_column="fold", blended_avg= True, inflection_point = 3, smoothing = 1)
te_map.fit(train)
encoded_train = te_map.transform(frame=train, holdout_type="kfold", seed=1234)
```

Error:

```
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/h2o/targetencoder.py", line 108, in transform
    assert self._encodingMap.map_keys['string'] == self._teColumns
AssertionError
```

h2o-docs/src/product/data-munging/target-encoding.rst
```diff
@@ -337,9 +338,13 @@ We do not need to apply any of the overfitting prevention techniques because our
 
    .. code-block:: python
 
-    ext_test = TargetEncoder(x=predictor, y=response,
-                             fold_column="fold", holdout_type="none", 
-                             blended_avg=False, noise=0, seed=1234)
+    # Add a fold column to the test set
+    fold = test.kfold_column(n_folds=5, seed=1234)
```
deil87: @angela0xdata there is no need to have fold column when we want to apply target encoding to test data. You need to create encoding map based on training data(or holdout data that model are not going to be trained on) and then just apply it to test frame. Test frame should not have fold column at all.

deil87: we just need to specify `fold_column` so that we can group encodings in our encoding map as they are split into folds in case when we generated map with folds for `kfold` application strategy.

h2o-docs/src/product/data-munging/target-encoding.rst
```diff
@@ -146,136 +216,151 @@ Now we will perform target encoding on ``addr_state`` to see if this representat
...
-    ## 2         AK               0.1521739
-    ## 3         AK               0.1521739
-    ## 4         AK               0.1521739
-    ## 5         AK               0.1521739
-    ## 6         AK               0.1521739
 
 Train Model with KFold Target Encoding
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Train a new model, this time replacing the ``addr_state`` with the ``TargetEncode_addr_state``.
```
deil87: @angela0xdata this is a prefix from old R version of target encoding. I'm using `addr_state_te` naming.

h2o-docs/src/product/data-munging/target-encoding.rst
```diff
@@ -146,136 +216,151 @@ Now we will perform target encoding on ``addr_state`` to see if this representat
...
-                                        fold_column = "fold",
-                                        blended_avg = FALSE, noise_level = 0)
+**Apply Target Encoding to Testing Dataset**
 
-    head(ext_test[c("addr_state", "TargetEncode_addr_state")])
+We do not need to apply any of the overfitting prevention techniques because our target encoding map was created on the training data, not the testing data.
 
-::
+-  ``holdout_type="none"``
+-  ``blended_avg=FALSE``
```
deil87: @angela0xdata blending is not about overfitting prevention. It is about imperfectness of our data. When we don't have enough data for a particular categorical level we want to use prior (means ignoring levels of the categorical variable) mean and blend it with what we know from posterior mean.

h2o-docs/src/product/data-munging/target-encoding.rst
deil87: @angela0xdata I see the problem. You should pass columns to encode as a list.  
```
teColumns =[ "addr_state"] <----- list here
targetEncoder = TargetEncoder(x= teColumns, y= targetColumnName, fold_column=foldColumnName, blended_avg= True, inflection_point = 3, smoothing = 1)
```
we might want to improve that by accepting just string to make it more robust.

deil87: created a Jira for that https://0xdata.atlassian.net/browse/PUBDEV-6432

deil87: @angela0xdata Actually this is a bug as I already added support for passing argument as string. But did not check in the test that calling `transform` method is working fine.

h2o-docs/src/product/data-munging/target-encoding.rst
angela0xdata: @deil87 okay, I'll try this all again and let you know how this goes. But just an FYI, you're welcome to check out the branch and make edits directly in the file. Again, I'm not an expert. I was just trying to make the Python version work similarly to R.

deil87: @angela0xdata That is actually what I'm doing :) Please take a look at changes that I'have pushed. I know that this task might be a painful one for you... I'm trying to help.

h2o-docs/src/product/data-munging/target-encoding.rst
deil87: @angela0xdata this PR will fix the error that you have been experiencing https://github.com/h2oai/h2o-3/pull/3440. Or just pass predictor as a list.

angela0xdata:
@deil87 Everything is working now! And R and Python examples return the same values.
Thank you so much for your help, additions, and patience.

h2o-docs/src/product/data-munging/target-encoding.rst
angela0xdata: Thanks! I'm just passing it as a list now.

h2o-docs/src/product/data-munging/target-encoding.rst
angela0xdata: Oh... yeah, I meant that I appreciated your updates, and in the future, you're always welcome to again :) ... sorry, that was pre-coffee.

PR #3411, opened 2019-04-01
[PUBDEV-6390] h2o-hive is not published to maven central

jakubhava:


honzasterba:
I guess we'll want this in rel-yates as well

jakubhava:
Also tested local publish of the project, h2o-hive is correctly there

PR #3419, opened 2019-04-04
PUBDEV-6371 bug in KMeans flow example

maurever:


maurever:
Ok, I did it and rechecked all example. It should be ok now.

laurendiperna:
thanks so much @maurever !

PR #3422, opened 2019-04-05
PUBDEV-6395: Fix ArrayIndexOutOfBounds in AUCBuilder

michalkurka:


h2o-core/src/test/java/hex/AUCBuilderTest.java
```diff
@@ -80,4 +80,17 @@ public void testPubDev6399ReduceDoesntUsePreviousKnownSSX() {
     assertArrayEquals(expected, ths, 0);
   }
 
+  @Test
+  public void testLargeWeights() {
+    AUC2.AUCBuilder ab = new AUC2.AUCBuilder(2);
+
+    double w = Double.MAX_VALUE;
+    ab.perRow(0, 1, w);
+    ab.perRow( 0.3, 1, w);
```
honzasterba: whitespace

h2o-core/src/test/java/hex/AUCBuilderTest.java
```diff
@@ -80,4 +80,17 @@ public void testPubDev6399ReduceDoesntUsePreviousKnownSSX() {
     assertArrayEquals(expected, ths, 0);
   }
 
+  @Test
+  public void testLargeWeights() {
+    AUC2.AUCBuilder ab = new AUC2.AUCBuilder(2);
```
honzasterba: does this test cover all of the added branches above? do we need one that does?

PR #3429, opened 2019-04-09
PUBDEV-6377 - Filling of sparse DMatrix data

Pscheidl:
[JIRA - PUBDEV-6377](https://0xdata.atlassian.net/browse/PUBDEV-6377)

Please also visit [Benchmarks](http://mr-0xc1:8080/view/H2O-3/job/h2o-3-benchmark-pipeline/job/pavel_pubdev-6377/) to see the performance impact of these changes.

### DIFF
--- a/h2o-core/src/test/java/water/TestUtil.java
+++ b/h2o-core/src/test/java/water/TestUtil.java
@@ -1050,18 +1050,7 @@ public class TestUtil extends Iced {
       final Vec f1Vec = f1.vec(vecNum);
       final Vec f2Vec = f2.vec(vecNum);
 
-      if (f1Vec.length() != f2Vec.length()) {
-        throw new IllegalStateException(String.format("Length of vectors in column %o is not equal: {%o,%o}",
-                vecNum, f1Vec.length(), f2Vec.length()));
-      }
-
-      for (int i = 0; i < f1Vec.length(); i++) {
-        if (f1Vec.at(i) != f2Vec.at(i)) {
-          throw new IllegalStateException(String.format("Not equal values found in column %o: {%f, %f}",
-                  vecNum,f1Vec.at(i), f2Vec.at(i)));
-        }
-      }
-
+      assertVecEquals(f1Vec, f2Vec, 0);
     }
 
     return true;
--- a/h2o-core/src/test/java/water/TestUtil.java
+++ b/h2o-core/src/test/java/water/TestUtil.java
@@ -1027,5 +1027,44 @@ public class TestUtil extends Iced {
     }
     return result;
   }
+
+  /**
+   * @param f1 Frame to be comapared, not null
+   * @param f2 Frame to be compared, not null
+   * @return True if matrices are precisely the same - number of columns, rows & values at each cell.
+   * @throws IllegalStateException If any inequalities are found
+   */
+  public static boolean compareFrames(final Frame f1, final Frame f2) throws IllegalStateException {
+    Objects.requireNonNull(f1);
+    Objects.requireNonNull(f2);
+
+    if (f1.numCols() != f2.numCols())
+      throw new IllegalStateException(String.format("Number of columns is not the same: {%o, %o}",
+              f1.numCols(), f2.numCols()));
+    if (f1.numRows() != f2.numRows())
+      throw new IllegalStateException(String.format("Number of rows is not the same: {%o, %o}",
+              f1.numRows(), f2.numRows()));
+
+    for (int vecNum = 0; vecNum < f1.numCols(); vecNum++) {
+
+      final Vec f1Vec = f1.vec(vecNum);
+      final Vec f2Vec = f2.vec(vecNum);
+
+      if (f1Vec.length() != f2Vec.length()) {
+        throw new IllegalStateException(String.format("Length of vectors in column %o is not equal: {%o,%o}",
+                vecNum, f1Vec.length(), f2Vec.length()));
+      }
+
+      for (int i = 0; i < f1Vec.length(); i++) {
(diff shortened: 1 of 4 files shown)


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -522,76 +521,137 @@ private static int initalizeFromChunkIds(Frame f, int[] chunks, Vec.Reader[] vec
...
                     } else {
-                        colIndex[currentRow][currentCol++] = di.getCategoricalId(j, vecs[j].at8(i));
+                        colIndex[rowPointer][currentCol++] = di.getCategoricalId(j, vecs[j].at8(i));
                     }
                     nonZeroCount++;
                 }
 
                 for (int j = 0; j < di._nums; ++j) {
+                    if(currentCol == SPARSE_MATRIX_DIM){
+                        currentCol = 0;
```
Pscheidl: I do believe a single check at each round will in the end (after all optimizations kick in) be a little bit faster than calculating the indices each time.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -522,76 +521,137 @@ private static int initalizeFromChunkIds(Frame f, int[] chunks, Vec.Reader[] vec
...
+                        rowPointer++;
+                    }
+                    data[rowPointer][currentCol] = 1; //one-hot encoding
+                    
+                    if(!foundNonZero){
+                        foundNonZero = true;
+                        for (int k = lastNonZeroRow; k < actualRows; k++) {
+                        rowHeaders[rowHeaderRowPointer][rowHeaderColPointer++] = nonZeroCount;
```
michalkurka: This doesn't seem correct to me. IMHO every time you increment a pointer you need to make sure there is room in the array.

Is there a mechanism that prevents you from running out of space in `rowHeaders[rowHeaderRowPointer]`

Pscheidl: There is a prevention mechanism. It is actually so simple it is easy to overlook. 

This particular "row header" (as we call it) has one purpose: to mark the first non-zero element each line begins with. As it required for this to happen exactly once per each row (if a row is empty, last non-zero is re-used, there is this special handling a little bit further in the code), the increment only happens once. Therefore, there is no need to check it more than once per each line.

This implies the actual check for `rowHeaderColPointer` overflow is NOT done in the ` for (int j = 0; j < di._cats; ++j) ` loop, but rather directly in the embedding loop `for(long i = f.anyVec().espc()[chunk]; i < f.anyVec().espc()[chunk+1]; i++)`. It is done right after weights are checked not to be null and the line verified for not to be skipped.

This piece of code does it:
```java
                if(rowHeaderColPointer == SPARSE_MATRIX_DIM){
                // enlarge final data arrays by 2x if needed
                    rowHeaderColPointer = 0;
                    rowHeaderRowPointer++;
                }
                boolean foundNonZero = false;
```

As seen in the example, there is also `foundNonZero` variable defined and explicitely set to false. This variable makes it possible to find out if there was a non-zero element found already on this row (as the outer loop is naturally traversing the matrix row-wise).

And as you can see a little bit further on line 550:

```java
                    if(!foundNonZero){
                        foundNonZero = true;
....
```

The non-zero element is only written into row headers if there was none discovered on this line before.

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostUtilsTest.java
```diff
@@ -56,5 +174,338 @@ public void parseFeatureScores() throws IOException, ParseException {
...
+
+  private static CsrLayout createLayout(long size, int maxArrLen) {
+    CsrLayout l = new CsrLayout();
+    l._numRegRows = (int) (size / maxArrLen);
+    l._regRowLen = maxArrLen;
+    l._lastRowLen = (int) (size - ((long) l._numRegRows * l._regRowLen)); // allow empty last row (easier and it shouldn't matter)
+    return l;
+  }
+
+  private static class CsrLayout {
```
Pscheidl: Borrowed from @michalkurka - Thank you !

wendycwong:
Pavel:

It looks like the major change you made was the way you increment the row indices and this resolves the index out of bound problem?  Amazing.


Wendy

Pscheidl:
@wendycwong Yes, unfortunately the CSR algorithm was implemented badly - there was no translation between 1D space and 2D internal representation.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
michalkurka: Thanks!

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -1027,5 +1027,33 @@ protected static RowData toRowData(Frame fr, String[] columns, long row) {
...
     return result;
   }
+
+  /**
+   * @param f1 Frame to be comapared, not null
+   * @param f2 Frame to be compared, not null
+   * @return True if matrices are precisely the same - number of columns, rows & values at each cell.
+   * @throws IllegalStateException If any inequalities are found
+   */
+  public static boolean compareFrames(final Frame f1, final Frame f2) throws IllegalStateException {
```
michalkurka: Nice!

PR #2724, opened 2018-08-09
PUBDEV-5745, PUBDEV-5746 keep_cross_validation_models support for all models

seb-h2o:
This PR resolves 2 tickets as they are strongly connected:
- https://0xdata.atlassian.net/browse/PUBDEV-5745
- https://0xdata.atlassian.net/browse/PUBDEV-5746

- To offer the possibility to clean cv models during Grid Search, keep_cross_validation_models parameter is now exposed to all models.
- To improve support in AutoML, and ensure that CV models are cleaned progressively, the keep_cross_validation_models automl param is passed to all models as well as to Stacked Ensemble metalearner.
- Changed the defaults for keep_cv_models and keep_cv_predictions in R+Py clients, as they differed from defaults in Java/REST API. Also changed those defaults for AutoML to align with other models.

PR #3430, opened 2019-04-09
PUBDEV-6415: keep_cross_validation_predictions parameter is not working

deil87:
This PR contains demo test that shows that `keep_cross_validation_predictions` is ignored.

### DIFF
--- a/h2o-automl/src/test/java/ai/h2o/automl/AutoMLKeepCVPreds.java
+++ b/h2o-automl/src/test/java/ai/h2o/automl/AutoMLKeepCVPreds.java
@@ -1,0 +1,55 @@
+package ai.h2o.automl;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+import water.Key;
+import water.fvec.Frame;
+
+import static org.junit.Assert.assertTrue;
+
+public class AutoMLKeepCVPreds extends water.TestUtil {
+
+  @BeforeClass
+  public static void setup() {
+    stall_till_cloudsize(1);
+  }
+
+  long autoMLSeed = 2345;
+
+  int numberOfModelsToCompareWith = 1;
+  
+  Algo[] excludeAlgos = {Algo.DeepLearning, Algo.DRF, Algo.GLM /*Algo.XGBoost */, Algo.GBM, Algo.StackedEnsemble};
+//  Algo[] excludeAlgos = {/*Algo.DeepLearning,*/ Algo.DRF, Algo.GLM , Algo.XGBoost , Algo.GBM, Algo.StackedEnsemble};
+
+  @Test
+  public void keep_cv_predictions_parameter_is_being_ignored_test() {
+    AutoML aml = null;
+    Frame fr = null;
+
+    String responseColumnName = "survived";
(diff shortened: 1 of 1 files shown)


deil87:
@seb-h2o If I understood your correctly we are overriding `keep_cross_validation_predictions` to `true` to be able to build SE. Maybe we should then remove `keep_cross_validation_predictions` parameter from the list of AutoML parameters and rely on whether we are going to build SE or not?

deil87:
@ledell you probably posted comment to a wrong PR. Or maybe I don't get it. Your comments is more about getting CV predictions from SE.

ledell:
@deil87 It's related because in your code you are trying to get the CV preds from the leader model which is a Stacked Ensemble.

deil87:
@ledell please take a look at `exclude_algos` parameter ....Stacked Ensemble was excluded. Test fails as xgboost keeps cv predictions regardless of the fact that parameter `keep_cross_validation_predictions` is set to `false`

ledell:
@deil87 Ah, ok - I missed that!  Thanks.

deil87:
@seb-h2o Could you please take a look one more time at this one? at least read through comments. It is xgboost model that keeps cv predictions not a stacked ensemble.

h2o-automl/src/test/java/ai/h2o/automl/AutoMLKeepCVPreds.java
```diff
@@ -0,0 +1,55 @@
...
+    autoMLBuildSpec.build_control.keep_cross_validation_models = false;
+    autoMLBuildSpec.build_control.keep_cross_validation_predictions = false;
+
+    long start1 = System.currentTimeMillis();
+    aml = AutoML.startAutoML(autoMLBuildSpec);
+    aml.get();
+
+    Key[] cross_validation_predictions = aml.leader()._output._cross_validation_predictions;
+    assertTrue(cross_validation_predictions == null);
+
```
seb-h2o: @deil87 @ledell 
this passes the test:
```
    assertNotNull(cross_validation_predictions);
    assertEquals(autoMLBuildSpec.build_control.nfolds, cross_validation_predictions.length);
    for (Key k : cross_validation_predictions) {
      assertNull(k.get());
    }
```

At some point, I had the intention to nullify this array (same for `cross_validation_models`) as it wasn't done before anyway, but after a discussion with @michalkurka we decided to leave it as it was: delete the keys from DKV but leave the output as an array with the keys, even if they don't reference any object anymore: cf. https://github.com/h2oai/h2o-3/pull/2724#discussion_r210131092

deil87: @seb-h2o thanks for the reference to discussion. Do we have inconsistency now? because I believe I have seen that some models return null if parameter is set to false.

seb-h2o: @deil87 it can probably be null if CV was disabled, but should be consistently an `array of dereferenced keys` otherwise

PR #3431, opened 2019-04-10
Add six to Python setuptools install requirements

swt2c:
Currently, if you install h2o using pip in an empty virtual environment,
'import h2o' will fail due to lack of the six library.  This resolves the
issue.

### DIFF
--- a/h2o-py/setup.py
+++ b/h2o-py/setup.py
@@ -80,7 +80,13 @@ setup(
     ]},
 
     # run-time dependencies
-    install_requires=["requests", "tabulate", "future", "colorama>=0.3.8"],
+    install_requires=[
+        "requests",
+        "tabulate",
+        "future",
+        "colorama>=0.3.8",
+        "six",
+    ],
 
     # optional dependencies
     extras_require={


michalkurka:
@swt2c thank you for this PR; we will look into it!

michalkurka:
@swt2c looks like we don't actually need to depend on the `six` package, we already have internal functions that handle the same functionality

I switched to using these internal functions instead of six in this PR: https://github.com/h2oai/h2o-3/pull/3435 - please take a look

swt2c:
@michalkurka yeah that solution is fine too!

Just of out curiosity, why didn't any of the CI stuff fire for this PR?

michalkurka:
@swt2c that is a good question. Our docker images do have `six` installed, it was specified in the test requirements, it was also in the conda's meta.yaml specification.

We do have a sanity check that the package can be installed from conda in a brand new environment and then used but we are currently missing such a test for pip. We will be adding such a test to prevent an issue like this in the future.

bilcus:
@michalkurka I think @swt2c was wondering why his PR did not get built/tested by CI at all.

michalkurka:
oh :) I see - that was wishful thinking on my side - I was just asking that question myself - why did we not catch the issue?

yeah, it won't run them because it is a PR from a fork; we don't allow running tests from forks automatically - someone has to first review the PR and approve running the tests otherwise people could just mess with our infrastructure and mine bitcoins or whatever :)

swt2c:
Closing this in favor of #3435.  Thanks!

PR #3432, opened 2019-04-10
PUBDEV-6393 - Rapids as.factor operation converts non-ASCII strings to sanitized forms

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6393

My assumption was that the behavior of `asFactor` should be the same as in `Categorical` class. Rules for conversion to categoricals should be the same.

### DIFF
--- a/h2o-core/src/main/java/water/util/VecUtils.java
+++ b/h2o-core/src/main/java/water/util/VecUtils.java
@@ -84,7 +84,12 @@ public class VecUtils {
             nc.addNA();
           } else {
             c.atStr(bs, row);
-            nc.addNum(lookupTable.get(bs.toSanitizedString()), 0);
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -623,7 +627,12 @@ public void map(Chunk c) {
       for (int i = 0; i < c.len(); i++) {
         if (!c.isNA(i)) {
           c.atStr(bs, i);
-          _uniques.put(bs.toSanitizedString(), _placeHolder);
+          final String strRepresentation = bs.toString();
+          if (strRepresentation.contains("\uFFFD")) {
```
Pscheidl: My assumption was that this behavior should be the same as in `Categorical` class.

PR #3433, opened 2019-04-10
[XGBOOST] upgrade to latest master

honzasterba:


michalkurka:
This is great but before merging can we also run the full nightly on it?

honzasterba:
nightly done and looks good

PR #3435, opened 2019-04-11
PUBDEV-6421: Get rid off `six` library from the python client

michalkurka:
Same can be accomplished using our own functions


michalkurka:
@bilcus thanks, I removed it from conda's meta.yaml and welcome.rst

I think it should definitely stay in `h2o-py/test-requirements.txt` because six is a transitive dependency of `pyopenssl` and in tests we want to have predictable versions.

Not sure about the requirements.txt - @honzasterba - wdyt?

PR #3438, opened 2019-04-15
More precise JavaDoc for compareFrames method in TestUtil class

Pscheidl:
A simple pull request making the documentation of TestUtil method I created earlier more precise. The contract was not defined that precisely :)

### DIFF
--- a/h2o-core/src/test/java/water/TestUtil.java
+++ b/h2o-core/src/test/java/water/TestUtil.java
@@ -1031,7 +1031,7 @@ public class TestUtil extends Iced {
   /**
    *
    * Compares two frames. Two frames are equal if and only if they contain the same number of columns, rows,
-   * and value at each cell (coordinate) is the same. Column names are ignored, as well as chunks sizes and all other
+   * and values at each cell (coordinate) are the same. Column names are ignored, as well as chunks sizes and all other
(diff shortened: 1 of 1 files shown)


h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -1029,7 +1029,12 @@ protected static RowData toRowData(Frame fr, String[] columns, long row) {
   }
 
   /**
-   * @param f1 Frame to be comapared, not null
+   *
+   * Compares two frames. Two frames are equal if and only if they contain the same number of columns, rows
```
maurever:
```suggestion
   * Compares two frames. Two frames are equal if and only if they contain the same number of columns, rows,
```

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -1029,7 +1029,12 @@ protected static RowData toRowData(Frame fr, String[] columns, long row) {
...
+   * and value at each cell (coordinate) is the same. Column names are ignored, as well as chunks sizes and all other
```
maurever:
```suggestion
   * and values at each cell (coordinate) are the same. Column names are ignored, as well as chunks sizes and all other
```

PR #3440, opened 2019-04-16
[PUBDEV-6432] Added Py API support for single column name or column index for target encoding.

deil87:
Before there was a support for :
```
teColumns = ["home.dest"]
    targetEncoder = TargetEncoder(x= teColumns, y= targetColumnName,blended_avg= True, inflection_point = 3, smoothing = 1)
```
with this PR it will be possible to pass column as string:
```
teColumns = "home.dest"
    targetEncoder = TargetEncoder(x= teColumns, y= targetColumnName, blended_avg= True, inflection_point = 3, smoothing = 1)
```
or as index of the column:
```
teColumns = 13
    targetEncoder = TargetEncoder(x= teColumns, y= targetColumnName, blended_avg= True, inflection_point = 3, smoothing = 1)
```

### DIFF
--- a/h2o-py/h2o/targetencoder.py
+++ b/h2o-py/h2o/targetencoder.py
@@ -58,6 +58,9 @@ class TargetEncoder(object):
             threshold between the posterior and the prior probability.
 
         """
+
+        if(type(x) == str or type(x) == int):
+            x = [x]
         self._teColumns = x
         self._responseColumnName = y
         self._foldColumnName = fold_column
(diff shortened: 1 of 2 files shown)


ledell:
@deil87 Does this leave the support for a list or remove the list support?

deil87:
@ledell No, actually there was such a logic but my tests were checking only `fit` method and not `transform` - so it could be seen as a bug. Also added test for a single column index. Just noticed that I probably created a branch from Angela's branch.... will fix it later.

PR #3444, opened 2019-04-17
PUBDEV-6394 - Rapids replace operations on categoricals do not respect reduced categorical domains

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6394

When applying replacement on parts of categorical string, a duplicate categorical levels might be created, effectively reducing the cardinality of the set.

In this PR, the proposed way to deal with it is to change the categorical mapping (only if cardinality changes !) to the new domain by gathering the categorical ordinals which now map to the same index in string domain levels array.

For example, if the following frame named `originalFrame`, containing one column only with five categoricals and three categorical levels ( `["Ｘ県 Ａ市", "Ｘ県 Ｂ市", "Ｙ県 Ｃ市"]` ): 

```plain
location
-----------
Ｘ県 Ａ市
Ｘ県 Ｂ市
Ｘ県 Ｂ市
Ｙ県 Ｃ市
Ｙ県 Ｃ市
```

and everything after the space was removed/replaced by calling `originalFrame['location'].gsub(' .*', '')`. The result should be

```plain
```plain
location
-----------
Ｘ県
Ｘ県
Ｘ県
Ｙ県
Ｙ県
```

and there should be only two categorical levels in the domain: `["Ｘ県", "Ｙ県"]`.

Created two JUnit tests and a kind-of duplicate Python test (really quick though) to demonstrate the result on what the reported provided us.

Note: `AstStrSplit`  has it's own way to deal with changes in categorical values already coded inside.

### DIFF
--- a/h2o-core/src/main/java/water/fvec/Vec.java
+++ b/h2o-core/src/main/java/water/fvec/Vec.java
@@ -469,7 +469,22 @@ public class Vec extends Keyed<Vec> {
     return v;
   }
 
-  public Vec remapDomain(final String[] newDomainValues) {
+  /**
+   * Remaps vec's current domain levels to a new set of values. The cardinality of the new set of domain values might be
+   * less than or equal to the cardinality of current domain values. The new domain set is automatically extracted from
+   * the given mapping array.
+   * <p>
+   * Changes are made to this very vector, no copying is done. If you need the original vector to remain unmodified,
+   * please make sure to copy it first.
+   *
+   * @param newDomainValues An array of new domain values. For each old domain value, there must be a new value in
+   *                        this array. The value at each index of newDomainValues array represents the new mapping for
+   *                        this very index. May not be null.
+   * @throws UnsupportedOperationException When invoked on non-categorical vector
+   * @throws IllegalArgumentException      Length of newDomainValues must be equal to length of current domain values of
+   *                                       this vector
+   */
+  public void remapDomain(final String[] newDomainValues) throws UnsupportedOperationException, IllegalArgumentException {
     // Sanity checks
     Objects.requireNonNull(newDomainValues);
     if (_domain == null)
@@ -506,11 +521,9 @@ public class Vec extends Keyed<Vec> {
       reducedDomainIdx++;
     }
 
-    final Vec copy = makeCopy();
     new RemapDomainTask(indicesMap)
-            .doAll(copy);
-    copy.setMeta(Vec.T_CAT, reducedDomain);
-    return copy;
+            .doAll(this);
+    this.setMeta(Vec.T_CAT, reducedDomain);
   }
 
   /**
--- a/h2o-core/src/main/java/water/fvec/Vec.java
+++ b/h2o-core/src/main/java/water/fvec/Vec.java
@@ -5,9 +5,7 @@ import water.nbhm.NonBlockingHashMap;
 import water.parser.BufferedString;
 import water.util.*;
 
-import java.util.Arrays;
-import java.util.Random;
-import java.util.UUID;
+import java.util.*;
 
 /** A distributed vector/array/column of uniform data.
  *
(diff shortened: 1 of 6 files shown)


h2o-core/src/main/java/water/rapids/ast/prims/string/AstReplaceAll.java
```diff
@@ -65,16 +66,29 @@ public Val apply(Env env, Env.StackHelp stk, AstRoot asts[]) {
   }
 
   private Vec replaceAllCategoricalCol(Vec vec, String pattern, String replacement, boolean ignoreCase) {
+    final Pattern compiledPattern = Pattern.compile(pattern); // Compile the pattern once before replacement
     String[] doms = vec.domain().clone();
-    for (int i = 0; i < doms.length; ++i)
+    Set<String> newDomainSet = new HashSet<>(); // The pattern might create multiple domains with the same name
```
Pscheidl: Might get big, but is fast to search for duplicated. Not sure if there is better alternative without sacrificing speed.

h2o-core/src/main/java/water/rapids/ast/prims/string/AstReplaceAll.java
```diff
@@ -65,16 +66,29 @@ public Val apply(Env env, Env.StackHelp stk, AstRoot asts[]) {
   }
 
   private Vec replaceAllCategoricalCol(Vec vec, String pattern, String replacement, boolean ignoreCase) {
+    final Pattern compiledPattern = Pattern.compile(pattern); // Compile the pattern once before replacement
```
Pscheidl: This is actually a speedup, as the previous version had to effectively call Pattern.compile every time with String.replaceAll. This change is present four times - once for string vectors and once for categorical vectors branch in both AstReplaceAll and AstReplaceFirst - four times totally.

h2o-core/src/main/java/water/fvec/Vec.java
```diff
@@ -471,6 +469,83 @@ public Vec makeCopy(String[] domain, byte type) {
...
+   * please make sure to copy it first.
+   *
+   * @param newDomainValues An array of new domain values. For each old domain value, there must be a new value in
+   *                        this array. The value at each index of newDomainValues array represents the new mapping for
+   *                        this very index. May not be null.
+   * @throws UnsupportedOperationException When invoked on non-categorical vector
+   * @throws IllegalArgumentException      Length of newDomainValues must be equal to length of current domain values of
+   *                                       this vector
+   */
+  public void remapDomain(final String[] newDomainValues) throws UnsupportedOperationException, IllegalArgumentException {
```
michalkurka: IMHO this is a too high-level function to be included in Vec, I would move it out of this class. 

Take a look at function `setDomain` - this function just changes the domain on the live object but doesn't reflect the changes in DKV.

This function mutates the Vec's data doesn't make sure the Vec's metadata is also updated. That is why I think it doesn't belong to the Vec class.

h2o-core/src/main/java/water/rapids/ast/prims/string/AstReplaceAll.java
```diff
@@ -65,16 +66,29 @@ public Val apply(Env env, Env.StackHelp stk, AstRoot asts[]) {
...
+    }
 
-    return vec.makeCopy(doms);
+    if (newDomainSet.size() == doms.length) {
+      // Avoid remapping if cardinality is the same
+      newDomainSet = null;
+      return vec.makeCopy(doms);
+    } else {
+      newDomainSet = null;
+      final Vec copy = vec.makeCopy();
```
michalkurka: Hmm, this creates a deep copy of the data just to overwrite it with something else.

I think it would be nicer if the RemapDomainTask simply produced a new vector (instead of mutating an existing duplicate version).

h2o-core/src/main/java/water/rapids/ast/prims/string/AstReplaceAll.java
michalkurka: IMHO not an issue.

michalkurka:
@Pscheidl good job! Some improvements suggested

michalkurka:
BTW Why is this marked as "breaking api change"?

Pscheidl:
> BTW Why is this marked as "breaking api change"?

By mistake. Sorry.

h2o-core/src/main/java/water/rapids/ast/prims/string/AstReplaceAll.java
Pscheidl: Since the old logic was mutating state of a deep copy of the original vec, this approach made sense. 

What doesn't make sense is to create a deep copy & then mutate it, making this approach not justifiable as well.

h2o-core/src/main/java/water/fvec/Vec.java
Pscheidl: At the end of the function
```
this.setMeta(Vec.T_CAT, reducedDomain);
```

Pscheidl: Absolutely right. Addressed.

h2o-core/src/main/java/water/rapids/ast/prims/string/AstReplaceAll.java
Pscheidl: Correct. Removed. Addressed this problem. The MR task itself now makes the task distributed uses NewChunk to pump data into a newly created Vec. No more deep copying and then "iterating" again, resulting in O(2n) on this level.

Pscheidl:
@michalkurka Addressed all your comments.

Pscheidl:
Merging, considering test failures unrelated:

```
water.util.DistributedException: DistributedException from /127.0.0.1:48004: 'javassist.CannotCompileException: by java.lang.IllegalAccessException: class javassist.util.proxy.DefineClassHelper$JavaOther cannot access a member of class java.lang.ClassLoader (in module java.base) with modifiers "protected final"'
```

PR #3445, opened 2019-04-18
Removed duplicate stack trace log in TCPReceiverThread

Pscheidl:
Discovered while investigating: [rel-yates build 17](http://mr-0xc1:8080/view/H2O-3/job/h2o-3-nightly-pipeline/job/rel-yates/17/testReport/junit/water/TCPReceiverThreadTest/Java_11_JUnit___testDontShutdownOnGarbageRequests/)

### DIFF
--- a/h2o-core/src/main/java/water/TCPReceiverThread.java
+++ b/h2o-core/src/main/java/water/TCPReceiverThread.java
@@ -123,7 +123,6 @@ public class TCPReceiverThread extends Thread {
         if (wrappedSocket != null) {
           try { wrappedSocket.close(); } catch (Exception e2) { Log.trace(e2); }
         }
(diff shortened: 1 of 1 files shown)


h2o-core/src/main/java/water/TCPReceiverThread.java
```diff
@@ -123,7 +123,6 @@ public void run() {
         if (wrappedSocket != null) {
           try { wrappedSocket.close(); } catch (Exception e2) { Log.trace(e2); }
         }
-        e.printStackTrace();
```
Pscheidl: Logger twice.

michalkurka: Log.err won't log the stacktrace

PR #3446, opened 2019-04-18
[PUBDEV-6431] fix exception for job list with replaced keys

honzasterba:
- when listing jobs use the actual value type for job result rather than the one stored in the job
- this fixes exception case when one job result is overwritten by another job result of different type

### DIFF
--- a/h2o-core/src/main/java/water/api/schemas3/JobV3.java
+++ b/h2o-core/src/main/java/water/api/schemas3/JobV3.java
@@ -3,6 +3,8 @@ package water.api.schemas3;
 import water.*;
 import water.api.API;
 import water.api.schemas3.KeyV3.JobKeyV3;
+import water.exceptions.H2OIllegalArgumentException;
+import water.util.Log;
 import water.util.PojoUtils;
 
 import java.io.PrintWriter;
@@ -91,8 +93,17 @@ public class JobV3 extends SchemaV3<Job, JobV3> {
     msec = job.msec();
     ready_for_view = job.readyForView();
 
-    Keyed dest_type = (Keyed)TypeMap.theFreezable(job._typeid);
-    dest = job._result == null ? null : KeyV3.make(dest_type.makeSchema(),job._result);
+    Keyed dest_type;
+    Value value = null;
+    if (job._result != null) {
+      value = DKV.get(job._result);
+    }
+    if (value != null) {
+      dest_type = (Keyed) TypeMap.theFreezable(value.type());
+    } else {
+      dest_type = (Keyed) TypeMap.theFreezable(job._typeid);
+    }
+    dest = job._result == null ? null : KeyV3.make(dest_type.makeSchema(), job._result);
     return this;
   }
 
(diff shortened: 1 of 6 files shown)


h2o-r/h2o-package/R/communication.R
```diff
@@ -749,6 +749,40 @@ h2o.clusterInfo <- function() {
...
+      'Key<AutoML>' = 'Auto Model',
+      'Key<ScalaCodeResult>' = 'Scala Code Execution',
+      'Key<KeyedVoid>' = 'Void',
+      'Unknown'
+    )
+}
+
+#' Return list of jobs performed by the H2O cluster
+#' @export
+h2o.listJobs <- function() {
```
michalkurka: the current convention is to use underscore notation, eg. h2o.list_jobs

h2o-r/h2o-package/R/communication.R
```diff
@@ -749,6 +749,40 @@ h2o.clusterInfo <- function() {
   }
 }
 
+.h2o.translateJobType <- function(type) {
```
michalkurka: Can you please add the same function to python as well, so that we have a parity?

h2o-core/src/main/java/water/api/schemas3/JobV3.java
```diff
@@ -91,8 +93,14 @@ public JobV3() {}
     msec = job.msec();
     ready_for_view = job.readyForView();
 
-    Keyed dest_type = (Keyed)TypeMap.theFreezable(job._typeid);
-    dest = job._result == null ? null : KeyV3.make(dest_type.makeSchema(),job._result);
+    Keyed dest_type;
+    Value value = DKV.get(job._result);
```
michalkurka: IMHO this is not the best place to fix this. We are keeping stale info in the DKV and overriding it here.

Instead of that, we could deprecate the old information when a new Job is started with a result_key that overwrites an existing Job result.

Take a look here: https://github.com/h2oai/h2o-3/blob/master/h2o-core/src/main/java/water/Job.java#L286
- here we know we are running a new job that will overwrite the results of another job. This is where we know the old results will no longer be relevant, I think the older job shouldn't point to the new one anymore.

honzasterba: i am not sure I understand what you mean, I tried something like this:

```java
    // 3 - Update job list
    final Key jobkey = _key;
    final Key jobResultKey = _result;
    new TAtomic<JobList>() {
      @Override public JobList atomic(JobList old) {
        if( old == null ) old = new JobList();
        Key[] jobs = old._jobs;
        for (Key<Job> jk : old._jobs) {
          Job j = DKV.get(jk).get();
          if (j._result.equals(jobResultKey)) {
            j._result = null;
          }
        }
        old._jobs = Arrays.copyOf(jobs, jobs.length + 1);
        old._jobs[jobs.length] = jobkey;
        return old;
      }
    }.invoke(LIST);
```

But Job#_result is final, so we need to duplicate and replace the old job instance when the result is owerwritten.

michalkurka: Yes, this is what I was thinking. However, fetching from DKV in an atomic update is not a good idea. I think your original approach is fine. Thanks for trying it out

PR #3447, opened 2019-04-18
fixed md5 generation for h2o.jar for cran

abal5:
cheers


michalkurka:
@abal5 thank you,  maybe this should also go to `rel-yates`?

PR #3448, opened 2019-04-19
Incomplete Cholesky Factorization

michalkurka:
Implementation of row-based Incomplete Cholesky Factorization based on paper https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/34638.pdf

This is a pre-processing step for the future implementation of distributed SVM

### DIFF
--- a/h2o-algos/src/main/java/hex/DataInfo.java
+++ b/h2o-algos/src/main/java/hex/DataInfo.java
@@ -900,6 +900,29 @@ public class DataInfo extends Keyed<DataInfo> {
         res += vec[vec.length-1];
       return res;
     }
+    public final double innerProduct(DataInfo.Row row) {
+      assert !_intercept;
+      assert numIds == null;
+
+      double res = 0;
+      for (int i = 0; i < nBins; ++i)
+        if (binIds[i] == row.binIds[i])
+          res += 1;
+      for (int i = 0; i < numVals.length; ++i)
+        res += numVals[i] * row.numVals[i];
+
+      return res;
+    }
+    public final double twoNormSq() {
+      assert !_intercept;
+      assert numIds == null;
+
+      double res = nBins;
+      for (double v : numVals)
+        res += v * v;
+
+      return res;
+    }
 
     public double[] expandCats() {
       if(isSparse() || _responses > 0) throw H2O.unimpl();
--- a/h2o-algos/src/main/java/hex/svm/IncompleteCholeskyFactorization.java
+++ b/h2o-algos/src/main/java/hex/svm/IncompleteCholeskyFactorization.java
@@ -1,0 +1,278 @@
+package hex.svm;
+
+import hex.DataInfo;
+import water.*;
+import water.fvec.Chunk;
+import water.fvec.Frame;
+import water.fvec.NewChunk;
+import water.fvec.Vec;
+import water.util.Log;
+
+/**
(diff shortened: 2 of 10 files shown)


h2o-algos/src/main/java/hex/DataInfo.java
```diff
@@ -900,6 +900,29 @@ public final double innerProduct(double [] vec, boolean icptFirst) {
...
+    public final double twoNormSq() {
+      assert !_intercept;
+      assert numIds == null;
+
+      double res = nBins;
+      for (double v : numVals)
+        res += v * v;
+
+      return res;
+    }
```
wendycwong: Done review.

h2o-algos/src/main/java/hex/svm/Kernel.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Override
+  public double calcKernelWithLabel(DataInfo.Row a, DataInfo.Row b) {
+    if ((int) a.response[0] != (int) b.response[0]) {
+      return -calcKernel(a, b);
+    } else {
+      return calcKernel(a, b);
+    }
+  }
+
```
wendycwong: Nice, we can then add polynomial kernel and whatever kernel that people want to use!

h2o-algos/src/test/java/hex/svm/IncompleteCholeskyFactorizationTest.java
```diff
@@ -0,0 +1,37 @@
...
+    } finally {
+      expected.delete();
+      if (fr != null)
+        fr.delete();
+      if (icf != null)
+        icf.delete();
+    }
+  }
+
+}
```
wendycwong: Done review.

h2o-algos/src/main/java/hex/svm/IncompleteCholeskyFactorization.java
```diff
@@ -0,0 +1,278 @@
...
+      Chunk two_norm_sq = cs[cs.length - 1];
+
+      for (int r = 0; r < cs[0]._len; r++) {
+        _dinfo.extractDenseRow(cs, r, row);
+        final double tns = row.twoNormSq();
+        row.response[1] = tns;
+        two_norm_sq.set(r, tns);
+        double diag1val = _kernel.calcKernel(row, row);
+        nc.addNum(diag1val);
+      }
```
wendycwong: Calculate and set the diagonal values of Q

wendycwong: It looks like the content of diag1 is set once and never changed.  However, during each iteration, the rows that are not in the set of pivots are changed during each iteration until either we hit the rank or that row is chosen as the one with the highest value.  Not sure this is accurate....

wendycwong:
Michal:

It looks good except I have one question.


Wendy

h2o-algos/src/main/java/hex/svm/IncompleteCholeskyFactorization.java
michalkurka: Good observation Wendy!

Vector `v` (using notation of the paper) is made of two components: `diag1` and `diag2`. The implementation only updates `diag2`.

When we are finding the pivot we have take that into account: https://github.com/h2oai/h2o-3/pull/3448/files#diff-da32ad2a741144bea660b1709b30abacR187

PR #3450, opened 2019-04-19
PUBDEV-6439 - Handling of non-ascii character in Py2's CSV reader & in one of the tests

Pscheidl:
https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6439

As a part of PUBDEV-6439, we introduced a Python test. The test was feeding non-ascii chars to H2O and expected them to be returned back. The test calls `    def categories(self):` function on an `H2OFrame`, which in return calls `return self.levels()[0]`. Then, the `levels()` function calls yet another function named `    def as_data_frame(self, use_pandas=True, header=True):`. 

If Pandas are used, everything is ok, as Pandas use their own ways to handle non-ascii characters. However, in this case, pandas are never used and line `1302` is reached. The line looks like this:

```python
frame = [row for row in csv.reader(StringIO(self.get_frame_data()))]
```

In Python 2, the csv.reader will break, as it can not read non-ascii chars. In Python 3, it works. I like the following summary of differences between Python 2 and Python 3 strings, besides original PEPs ([source](http://lucumr.pocoo.org/2014/1/5/unicode-in-2-and-3/)):

```
The main difference between Python 2 and Python 3 is the basic types that exist to deal with texts and bytes. On Python 3 we have one text type: str which holds Unicode data and two byte types bytes and bytearray.

On the other hand on Python 2 we have two text types: str which for all intents and purposes is limited to ASCII + some undefined data above the 7 bit range, unicode which is equivalent to the Python 3 str type and one byte type bytearray which it inherited from Python 3.
``` 

## Resolution

There are multiple ways to handle this. Official guide [provides two methods](https://docs.python.org/2/library/csv.html#examples) doing the encoding/decoding. However, this has two major flaws. It is not Python-3 compatible, as the csv.reader expects different input in both versions. Secondly, it does not solve other issues like [bpo-12178](https://bugs.python.org/issue12178).

Oh, and one more thing - the `unicode()` method it uses is now removed from Python 2 and 3. 

Overcoming all the troubles with python2 vs python3 csv parser encoding handling seems to be a time-consuming task. Luckily, the problem is so common it is addressed by the **CSV23** module: https://pypi.org/project/csv23/ - a thin wrapper over the built-in csv functionality.

In case of Python 3, it uses basically uses the same routing we are using right now, producing the same output. As for Python 2, it escapes the characters, as Python 2 limits itself to `unicode` type by default and we're working with it everywhere. The result is verified to be the same, only in Python 2 the result might come in an encoded form if there are non-ascii chars in the cell.


h2o-py/h2o/frame.py
```diff
@@ -1299,7 +1299,7 @@ def as_data_frame(self, use_pandas=True, header=True):
         if can_use_pandas() and use_pandas:
             import pandas
             return pandas.read_csv(StringIO(self.get_frame_data()), low_memory=False, skip_blank_lines=False)
```
Pscheidl: Pandas provide their own built-in types to handle non-ascii chars. This branch is ok.

h2o-py/tests/testdir_jira/pyunit_pubdev_6393.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
```
Pscheidl: There are non-ascii characters in the test.

michalkurka:
@Pscheidl great investigation, can you please add a unit test that shows where `as_data_frame` fails?

Pscheidl:
@michalkurka I've added an explicit test just for that. Also testing the parsed values are the same.

Handling this problem is not too hard, but it's not 10 lines of code either. There is a great library named [CSV23](https://github.com/xflr6/csv23/tree/master/csv23 ) I looked at, checked and borrowed their code, as they already handled all the problems that might rise with different setups under different circumstances. As we do not need their `write` module, I removed that.

The reason not to include this library directly is simple: H2O tries to avoid any new pip dependencies right now. And this code is pretty small in general, few files with tens of lines of code in each. Original authors were of course given credit.

h2o-py/setup.py
```diff
@@ -29,7 +29,7 @@
 
     description='H2O, Fast Scalable Machine Learning, for python ',
     long_description=long_description,
-
+To cos po
```
honzasterba: left by accident?

Pscheidl: Yes, sorry.

Pscheidl:
The escaping only occurs if necessary:

```python
    data = [['C1'],
             ['Ｘ県 Ａ市 '], # First observation contains replacement character for unknown char
             ['Franta jede na kole']]
```

**Output:**

```
Parse progress: |█████████████████████████████████████████████████████████| 100%
[u'Franta jede na kole', u'\uff38\u770c \uff21\u5e02 ']
```

PR #3451, opened 2019-04-22
PUBDEV-5969 updated hist() plot method  to be compatible with matplot…

laurendiperna:
…lib 3.x and 2.x

### DIFF
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -2536,7 +2536,12 @@ class H2OFrame(object):
             plt.xlabel(self.names[0])
             plt.ylabel("Frequency")
             plt.title("Histogram of %s" % self.names[0])
-            plt.bar(left=lefts, width=widths, height=counts, bottom=0)
+
+            try:
+                plt.bar(left=lefts, width=widths, height=counts, bottom=0)
+            except TypeError:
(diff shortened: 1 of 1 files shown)


h2o-py/h2o/frame.py
```diff
@@ -2536,7 +2536,12 @@ def hist(self, breaks="sturges", plot=True, **kwargs):
             plt.xlabel(self.names[0])
             plt.ylabel("Frequency")
             plt.title("Histogram of %s" % self.names[0])
-            plt.bar(left=lefts, width=widths, height=counts, bottom=0)
+
+            try:
```
michalkurka: Can we make the decision based on `matplotlib.__version__` instead? IMHO that is a bit more obvious (and future proof).

laurendiperna: good point! I updated the code to reflect the version of matplotlib. To avoid the deprecation warning for the `left` argument in `plt.bar()` the check kicks in for matplotlib 2.1.0 or later. Here is where matplotlib notes the [start of the deprecation warning](https://matplotlib.org/2.2.3/api/api_changes.html#change-to-signatures-of-bar-barh). Here is where they [note the removal](https://matplotlib.org/api/api_changes.html#bar-barh-no-longer-accepts-left-bottom-as-first-named-argument).

michalkurka:
Thank you @laurendiperna!

PR #3452, opened 2019-04-22
PUBDEV-6338: In docs, add R example for slicing by date

angela0xdata:
I'm not sure this is working correctly, though. (Or I did something wrong.) I imagine that my example would remove one row, but it removed more.


angela0xdata:
@ledell I don't know of a time series dataset that's available on S3. I couldn't find one in my smalldata folder.

ledell:
@angela0xdata Ok, no problem.  We can ask @abal5 to upload this dataset to S3.

PR #3455, opened 2019-04-23
[PUBDEV-6428] udpate xgboost to latest stable

honzasterba:


build.gradle
```diff
@@ -209,6 +209,10 @@ subprojects {
         if (project.hasProperty("enableMavenLocal")) { // useful for development
             mavenLocal()
         }
+        maven {
```
michalkurka: Should be removed before merged

PR #3457, opened 2019-04-24
[PUBDEV-5937] Log errors rather than throwing them in XGBoostUpdater.…

honzasterba:
…run method

- also add an uncaught exception handler

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostUpdater.java
+++ b/h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostUpdater.java
@@ -52,12 +52,14 @@ public class XGBoostUpdater extends Thread {
     } catch (InterruptedException e) {
       XGBoostUpdater self = updaters.get(_modelKey);
       if (self != null) {
+        Log.err("Updater thread was interrupted while it was still registered, name=" + self.getName());
         Log.err(e);
-        throw new IllegalStateException("Updater thread was interrupted while it was still registered, name=" + self.getName());
+      } else {
+        Log.debug("Updater thread interrupted.", e);
       }
     } catch (XGBoostError e) {
+      Log.err("XGBoost training iteration failed");
       Log.err(e);
-      throw new IllegalStateException("XGBoost training iteration failed", e);
     } finally {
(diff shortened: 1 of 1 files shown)


h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostUpdater.java
```diff
@@ -37,6 +37,12 @@ private XGBoostUpdater(Key<XGBoostModel> modelKey, DMatrix trainMat, BoosterParm
     _rabitEnv = rabitEnv;
     _in = new SynchronousQueue<>();
     _out = new SynchronousQueue<>();
+    setUncaughtExceptionHandler(new UncaughtExceptionHandler() {
```
michalkurka: This is the exact same fix I did, isn't it? :)

The only difference is that I am calling it after the updater is instantiated because calling an overridable method in a constructor is generally not a good practice.

honzasterba: I am not sure I follow, I have searched if setUncaughtExceptionHandler is used anywhere else and did not find it

I have replaced throwing the exception with logging since it does not appear anywhere and If there are any undeclared exceptions thrown the handler will log them as well

michalkurka: I didn't merge it to rel-yates, only to master: https://github.com/h2oai/h2o-3/pull/3456/files#diff-60c0595923fbd4a973dfececd24bd72dR195

sorry, I will push the commit there too - you can change it to your liking

honzasterba:
@michalkurka I refined the solution a little bit

h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/XGBoostUpdater.java
michalkurka: I think the handler is still set 2x

honzasterba: true, fixed

PR #3459, opened 2019-04-25
[ITA-374] Add Java 12 Docker images.

michal-raska:


michal-raska:
Then changeset is somewhat bigger, I had to modify the Java 8 installation, since it's no longer available

michal-raska:
@michalkurka based on our discussion, this PR goes to master only

PR #3461, opened 2019-04-25
PUBDEV-6435 - MOJO import convenience methods

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6435

Goals of this PR: 
- Offer convenience methods to import MOJO and create a Generic model out of it
- Make upload from local filesystem available
- Print details about the imported model (there might not be much available right now, if anything, being solved in other PRs)
- Python & R
- Write tests


Pscheidl:
Test failures unrelated (re-running).

ledell:
@Pscheidl I think this is good.  It departs from our old/existing convention in h2o-r to use camel case for import related functions, but I think it's still best to move all new functions to Tidyverse style guide (just as you did).  At least half of the `h2o.import*` functions are using the new style at this point.

Screen Shot 2019-04-27 at 9 38 51 AM

PR #3462, opened 2019-04-25
PUBDEV-6408- CSV - ArrayIndexOutOfBounds when quotes are parsed

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6408

It happends during STRING_END state while the parser had to advance to a next chunk.


h2o-core/src/main/java/water/parser/CsvParser.java
```diff
@@ -130,26 +130,17 @@
...
-            if (quoteCount>1) {
-              str.addChar();
-              quoteCount--;
-            }
-
             state = COND_QUOTE;
             continue MAIN_LOOP;
           }
-          if ((!isEOL(c) || quoteCount == 1) && (c != CHAR_SEPARATOR)) {
+          if ((!isEOL(c) && c != CHAR_SEPARATOR) || quoteCount == 1) {
```
Pscheidl: Fixes another boung I just found. There is no reason to add characters not fitting this rule in another step. All quotes are now handled at the beginning of STRING state and will not enter this state unless quoteCount is 0 or 1. It can now never be > 1.

h2o-core/src/main/java/water/parser/CsvParser.java
```diff
@@ -495,9 +481,13 @@
         if( !firstChunk || bits1 == null ) { // No more data available or allowed
           // If we are mid-parse of something, act like we saw a LF to end the
           // current token.
-          if(c == CHAR_LF || c == CHAR_CR) quoteCount++;
```
Pscheidl: The idea here was simple: If the parse hit the end of a chunk and there was no more data available, increment quotecount by one.

This is a nonsense and does not fit current logic. This is first part of the fix. The new condition below finalizes the fix.

h2o-core/src/main/java/water/parser/CsvParser.java
```diff
@@ -495,9 +481,13 @@
...
           if ((state != EXPECT_COND_LF) && (state != POSSIBLE_EMPTY_LINE)) {
-            c = CHAR_LF;  continue; // MAIN_LOOP;
+            c = CHAR_LF;
+            if(state == STRING) {
```
Pscheidl: Once the parser truly reaches the end, just set quoteCount to zero, as it does not matter anymore and set state to STRING_END. Only for String.

h2o-core/src/test/java/water/parser/CsvParserTest.java
```diff
@@ -317,4 +318,31 @@ public void nonLineMarkers_modified(){
...
+    parseSetup._separator = ',';
+    parseSetup._column_types = new byte[]{Vec.T_STR, Vec.T_STR};
+    parseSetup._column_names = new String[]{"C1"};
+    parseSetup._number_columns = 1;
+    parseSetup._single_quotes = false;
+    parseSetup._nonDataLineMarkers = new byte[0];
+    CsvParser csvParser = new CsvParser(parseSetup, null);
+    
+      
+    final Parser.StreamData data = new Parser.StreamData(new ByteArrayInputStream("\"abcde,".getBytes()), 6);
```
Pscheidl: Advancement to next chunk is necessary to reproduce, thus StreamData is used.

Pscheidl:
The original issue is fixed, however tests do discover failures due to landing the beginning of a chunk `cidx > 0` inside quotes and possibly other troubles. 

What needs to be done is: https://github.com/Rdatatable/data.table/blob/master/src/fread.c#L346

Shall we resolve it as a separate issue or in this PR @michalkurka ? Could take some time.

Pscheidl:
After discussion with @michalkurka, the other issues related to chunking will be solved in a separate PR.

Pscheidl:
@honzasterba That's what GitHub's "squash and merge" button is for. For reviewing purpose, it is always better to have the possibility of seeing the incremental changes if required.

Pscheidl:
We already have a JIRA for the other issue: https://0xdata.atlassian.net/browse/PUBDEV-5947

PR #3465, opened 2019-04-27
Explicitly re-map domains in prediction frames if necessary (PUBDEV-6449)

michalkurka:
Instead of using CategoricalWrappedVec: update is more expensive than
wrapping a Vec but the benefit is that it is straightforward to work further
with a materialized Vec. Modifying a CategoricalWrappedVec might not always
work.

### DIFF
--- a/h2o-algos/src/main/java/hex/generic/Generic.java
+++ b/h2o-algos/src/main/java/hex/generic/Generic.java
@@ -27,15 +27,11 @@ public class Generic extends ModelBuilder<GenericModel, GenericModelParameters, 
 
     public Generic(GenericModelParameters genericParameters){
         super(genericParameters);
+        init(false);
     }
 
     public Generic(boolean startup_once) {
         super(new GenericModelParameters(), startup_once);
-    }
-
-    @Override
-    public void init(boolean expensive) {
-        super.init(expensive);
     }
 
     @Override
--- a/h2o-core/src/main/java/water/fvec/CategoricalWrappedVec.java
+++ b/h2o-core/src/main/java/water/fvec/CategoricalWrappedVec.java
@@ -3,6 +3,7 @@ package water.fvec;
 import water.AutoBuffer;
 import water.Key;
 import water.DKV;
+import water.MRTask;
 import water.util.ArrayUtils;
 
 import java.util.Arrays;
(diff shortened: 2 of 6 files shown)


h2o-core/src/main/java/water/fvec/CategoricalWrappedVec.java
```diff
@@ -30,7 +30,11 @@
 
   /** Main constructor: convert from one categorical to another */
   public CategoricalWrappedVec(Key key, int rowLayout, String[] toDomain, Key masterVecKey) {
```
Pscheidl: Note: `doCopy` method should look like this ?

```java
  @Override
  public Vec doCopy() {
    return new CategoricalWrappedVec(group().addVec(),_rowLayout, domain(), _masterVecKey, _cascadeDelete);
  }
```

Pscheidl:
I know this is a WIP, but I gave it a try and found two places of improvement:

- doCopy method of vector wrapper should copy cascade settings (to cascade or not to cascade)
- Once the frame is deleted, Vec's custom removal logic is never called.

I've created a patch with proposed changes. Not polished in terms of speed, just a pointer to the exact locations I have in mind.


```diff
Index: h2o-core/src/main/java/water/fvec/Vec.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- h2o-core/src/main/java/water/fvec/Vec.java	(revision 24376a584d83d481932ac08ac65ec2ffa5df7a31)
+++ h2o-core/src/main/java/water/fvec/Vec.java	(date 1556344339000)
@@ -1337,7 +1337,10 @@
     // Remove Vecs everywhere first - important! (this should make simultaneously running Rollups to fail)
     new MRTask() {
       @Override public void setupLocal() {
-        for( Key k : keys ) if( k != null ) Vec.bulk_remove_vec(k, ncs);
+        for( Key k : keys ) if( k != null ) {
+          k.remove();
+          Vec.bulk_remove_vec(k, ncs);
+        }
       }
     }.doAllNodes();
     // Remove RollupStats
Index: h2o-core/src/main/java/water/fvec/CategoricalWrappedVec.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- h2o-core/src/main/java/water/fvec/CategoricalWrappedVec.java	(revision 24376a584d83d481932ac08ac65ec2ffa5df7a31)
+++ h2o-core/src/main/java/water/fvec/CategoricalWrappedVec.java	(date 1556343164000)
@@ -142,7 +142,7 @@
 
   @Override
   public Vec doCopy() {
-    return new CategoricalWrappedVec(group().addVec(),_rowLayout, domain(), _masterVecKey);
+    return new CategoricalWrappedVec(group().addVec(),_rowLayout, domain(), _masterVecKey, _cascadeDelete);
   }
 
   public static class CategoricalWrappedChunk extends Chunk {
```

michalkurka:
@Pscheidl simpler solution implemented, please review

Pscheidl:
Fixes https://0xdata.atlassian.net/browse/PUBDEV-6449. I allowed myself to link this PR to the JIRA.

h2o-algos/src/main/java/hex/generic/Generic.java
```diff
@@ -27,17 +27,13 @@
 
     public Generic(GenericModelParameters genericParameters){
         super(genericParameters);
+        init(false);
```
Pscheidl: Thank you, great to delete the leftovers.

PR #3470, opened 2019-05-01
SVM: Implement Primal-Dual Interior Point Method

michalkurka:


h2o-algos/src/main/java/hex/svm/LLMatrix.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+
+  private int dim() {
+    return _dim;
+  }
+
+  final double get(int x, int y) {
+    return _data[y][x-y];
+  }
```
wendycwong: Michal:

x is the row index, y is the column index, right?

h2o-algos/src/main/java/hex/svm/LLMatrix.java
```diff
@@ -0,0 +1,87 @@
...
+
+  final void set(int x, int y, double value) { 
+    _data[y][x-y] = value;
+  }
+
+  final void addUnitMat() {
+    for (double[] col : _data)
+      col[0] += 1;
+  }
```
wendycwong: Michal:

This adds 1 to the diagonal element?

h2o-algos/src/main/java/hex/svm/LLMatrix.java
```diff
@@ -0,0 +1,87 @@
...
+  private final double[][] _data;
+
+  LLMatrix(int dim) {
+    _dim = dim;
+    _data = new double[dim][];
+    for (int i = 0; i < dim; i++) {
+      _data[i] = new double[dim - i];
+    }
+  }
+
```
wendycwong: If I assume for the double array data, the first index is row index, the second index is column index, then the data is a upper triangular matrix, right?

h2o-algos/src/main/java/hex/svm/PrimalDualIPM.java
```diff
@@ -0,0 +1,502 @@
...
+      _la = cs[5];
+      _dla = cs[6];
+      _tlx = cs[7];
+      _tux = cs[8];
+      _xilx = cs[9];
+      _laux = cs[10];
+      _d = cs[11];
+      _dx = cs[12];
+      map();
+    }
```
wendycwong: Michalk:

I know what most of your variable names mean except the following:

_tlx, _tux, _xilx, _laux.  Can you tell me what they are?

h2o-algos/src/main/java/hex/svm/LLMatrix.java
michalkurka: Yes, adds 1 to the diagonal.

h2o-algos/src/main/java/hex/svm/PrimalDualIPM.java
michalkurka: These temporary vectors and they are not referenced in the paper, I took them from the reference implementation:

    //     tlx = \frac{1}{t \alpha}
    //     tux = \frac{1}{t (C - \alpha)}
    //     xilx = \frac{\xi}{\alpha}
    //     laux = \frac{\lambda}{C - \alpha}
    //     D^(-1) = \diag(\frac{\xi}{\alpha} + \frac{\lambda}{C - \alpha})

wendycwong: I finally got it after going through your UpdateVarsTask!

michalkurka:
@wendycwong thanks for the review! much appreciated

PR #3472, opened 2019-05-02
PUBDEV-6416 - Grid search appears to use old parameters

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6416

In the JIRA, it is assumed grid search somehow re-uses original parameters and fails model training even after invoked for a second times with different search space defined. This is not the case. The root cause is simple - there are potentially n > 1 models trained during grid search. The algorithm has to have means to store the errors encountered during training, if any occur.

Once an attempt to add more models to the grid by invoking the search one more time, the previously encountered failures are still present in our Grid object in DKV (as the Grid is re-used). Releasing old errors once new grid search is triggered and starting from scratch seems to be the simplest solution. 

The previous state caused our client to print old error messages, even if unrelated to the current hyperparameter space search. The search was succesfull, only old errors were printed, giving the user a perfect impression of a failed grid search.

There are two tests. JUnit test is obvious and tests no failures are propagated. The second Python integration test does basically the some, only makes sure the error is gone once search is invoked second time. 

I also double checked the error is gone once acceptable hyperparams are provided in Jupyter notebook provided by our customer. Published the screenshot in the support ticket.


h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -98,6 +98,8 @@ private GridSearch(Key<Grid> gkey, HyperSpaceWalker<MP, ?> hyperSpaceWalker) {
       if (! (keyed instanceof Grid))
         throw new H2OIllegalArgumentException("Name conflict: tried to create a Grid using the ID of a non-Grid object that's already in H2O: " + _job._result + "; it is a: " + keyed.getClass());
       grid = (Grid) keyed;
+      // Old failures are no longer needed once new hyperspace search is issued by the user.
+      if ((grid.getFailureCount() > 0)) grid.initFailureFields();
```
michalkurka: Hmm, is this a good idea? IMHO should keep the failures of the models that are part of this grid but discard them for the ones that are not part of this grid anymore, no?

michalkurka:
This is a simple and elegant solution but I am not yet fully convinced :)

I think if there is a way to preserve the failures of the models that were actually part of the grid - I would prefer that. We just need to discard the failures that are no longer part of the grid.

But this solution makes sense too. I am just concerned that users will change an unrelated hyperparameter and won't see that they actually didn't fix the issue.

Pscheidl:
- Store failures related to models that made it into the grid separately from those who did not make it.
- If `putModel` method is not called in`GridSearch.buildModel()` due to a model build failure, the errors are stored under `Grid.NO_MODEL_FAILURES_KEY`.
- Failures not related to any model in the grid (stored under `Grid.NO_MODEL_FAILURES_KEY`) are removed every time a new hyperspace walk is invoked by the user. This behavior can be safely modified in the future.
- Errors related to in-grid models are easily separable per model.
- Immutable list of errors `SearchFailure` is returned - its lifetime is not bound to the grid (considered an advantage). Also prevents mutating Grid's failure from the outer world.

I was unable to re-use the generic parameter `MP` of the `Grid` class, as non-static class `SearchFailure` would introduce a cyclic dependency and serialization would fail.

PR #3473, opened 2019-05-02
PUBDEV-6466: Investigation on Vec's setDomain

deil87:
I think that this is a bug because Vec is supposed to be distributed and if we have method `setDomain` it means that this info should be propagated to all the nodes.

Another thing is that we don't have Java unit tests for this functionality.

But I can imagine that we kind of don't wan't to provide access to the domains from the `map` methods of MRTasks and don't care about how domain is being propagated. I run into this when I wanted to create keys for BroadcastJoin... I was using factor values. I might probably use level values(numbers) as a workaround. @michalkurka Just wondering if there is such an assumption that we can't call `some_chunk.vec().domain()` or `some_chunk.vec().factor()` from within `map`?

### DIFF
--- a/h2o-automl/src/test/java/ai/h2o/automl/targetencoding/TargetEncodingDistributedTest.java
+++ b/h2o-automl/src/test/java/ai/h2o/automl/targetencoding/TargetEncodingDistributedTest.java
@@ -1,0 +1,107 @@
+package ai.h2o.automl.targetencoding;
+
+import org.junit.After;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import water.MRTask;
+import water.TestUtil;
+import water.fvec.Chunk;
+import water.fvec.Frame;
+import water.fvec.TestFrameBuilder;
+import water.fvec.Vec;
+
+import java.util.Arrays;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class TargetEncodingDistributedTest extends TestUtil {
+
+
+  @BeforeClass
+  public static void setup() {
+    stall_till_cloudsize(2);
+  }
+
+  private Frame fr = null;
+
+
+  @Test
+  public void imputeNAsForColumnTest() {
+    String teColumnName = "ColA";
+    String targetColumnName = "ColB";
+    fr = new TestFrameBuilder()
+            .withName("testFrame")
+            .withColNames(teColumnName, targetColumnName)
+            .withVecTypes(Vec.T_CAT, Vec.T_CAT)
+            .withDataForCol(0, ar("a", "b", null, null, null))
+            .withDataForCol(1, ar("2", "6", "6", "2", "6"))
+            .withChunkLayout(3,2)
+            .build();
+
+    String nullStr = null;
+    fr.vec(0).set(2, nullStr);
+
+    String[] teColumns = {""};
+    TargetEncoder tec = new TargetEncoder(teColumns);
+
+    assertTrue(fr.vec("ColA").isCategorical());
+    assertEquals(2, fr.vec("ColA").cardinality());
+
+    Frame res = tec.imputeNAsForColumn(fr, "ColA", "ColA_NA");
+
+    new MRTask() {
+      @Override
+      public void map(Chunk[] cs) {
+        long levelValue = cs[0].at8(0);
+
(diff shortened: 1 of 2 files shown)


h2o-automl/src/test/java/ai/h2o/automl/targetencoding/TargetEncodingDistributedTest.java
```diff
@@ -0,0 +1,107 @@
...
+    fr = new TestFrameBuilder()
+            .withName("testFrame")
+            .withColNames(teColumnName)
+            .withVecTypes(Vec.T_CAT)
+            .withDataForCol(0, ar("a","a","b", "b", "b"))
+            .withChunkLayout(3,2)
+            .build();
+
+    final String[] domain = {"a", "b", "imputed_cat"};
+    fr.vec(teColumnName).setDomain(domain);
```
michalkurka: You are modifying just the local copy. In order to make the change available to the MRTask, you need to propagate the change to DKV. A good pattern is to acquire a `lock`, modify the locked object and then call `update` (which will automatically install it in DKV) and release the lock.

deil87: @michalkurka I see, thank you!  I would just expect this to be done inside `setDomain` automatically. Should not we do this?

michalkurka:
> Just wondering if there is such an assumption that we can't call some_chunk.vec().domain() or some_chunk.vec().factor() from within map

There is no such assumption, calling `chunk.vec().domain()` is a legal use and it will work as long as the Vec is correctly installed in DKV prior calling the MRTask

You can also call `chunk.vec().factor(n)` however, that is less common as in general, you might be fetching data from a remote node. But that is sometimes justified as well.

h2o-automl/src/test/java/ai/h2o/automl/targetencoding/TargetEncodingDistributedTest.java
```diff
@@ -0,0 +1,107 @@
...
+    new MRTask() {
+      @Override
+      public void map(Chunk[] cs) {
+        long levelValue = cs[0].at8(0);
+
+        System.out.println("Level value: " + levelValue);
+        System.out.println("Domain: " + Arrays.toString(cs[0].vec().domain()));
+        int length = cs[0].vec().domain().length;
+        System.out.println("Domain length: " + length);
+        cs[0].vec().factor(2); // should throw exception on one of the nodes
```
michalkurka: This fails because the Vec is not properly updated in DKV in function `imputeNAsForColumn`

      ...
      System.arraycopy(oldDomain, 0, newDomain, 0, oldDomain.length);
      newDomain[indexForNACategory] = strToImpute;
      currentVec.setDomain(newDomain);
      return data;

h2o-automl/src/test/java/ai/h2o/automl/targetencoding/TargetEncodingDistributedTest.java
michalkurka: I don't think it can really be done automatically, most of the time setDomain has to be done with conjunction of updating the chunks as well. You don't want to end up with an inconsistent Vec that just has updated domain but corrupted chunks. That is why locking is a good idea for updates of Frames/Vec that you know someone else might be referencing (if you are the creator of the Vec and no one has it - it is an overkill).

deil87: @michalkurka It looks like we don't have ability to lock on Vec level. But Frame does have it. There are not that many usages in the code.... I'm trying something like this ->
```
Lockable lock = fr.write_lock();
    fr.vec(teColumnName).setDomain(domain);
    fr.update();
    lock.unlock()
```
and it does not work. Could you please point out to an example?

michalkurka: :) you are right! I was talking about Frames but answering about Vec - sorry for confustion, then you just need to do DKV.put on the modified Vec. For modifying Frames - your code snippet is correct.

deil87: @michalkurka It works!

PR #3476, opened 2019-05-03
XGBoost: Avoid repeated detection of (default) GPU

michalkurka:
For the most common use case - backend=auto and gpu_id=0 (default)
If we find out GPU 0 is not present blacklist it and never try to detect
its availablity again.


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -563,6 +563,7 @@ final void updateBooster() {
     }
   }
 
+  private static volatile boolean DEFAULT_GPU_BLACKLISTED = false;
```
honzasterba: the name of the variable implies it is a constant, when in fact it is not

PR #3477, opened 2019-05-05
Failing Iced serialization/writing into autobuffer

Pscheidl:
In case of a non-static inner class, the serialization runs in cycles, resulting in stack overflow.

Same effect is achieved with `Serializable`, only the serialization goes through different route. Means the problem probably lies somewhere else.

The cycle makes sense - the instance of non-static inner class needs the outer class instance. However, the outer class contains a field with the inner class. This creates the infinite loop.


Pscheidl:
Creates a cycle in the dependency graph, while the dependency graph must be a tree. Static class required in this case.

PR #3478, opened 2019-05-06
PUBDEV-6471 - XGBoost Error: Some trailing characters could not be parsed

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6471

My proposal for a fix. I remember fixing this issue long time ago, long before this method was introduced (October last year is what git says). Looks like this new piece of code broke it again.

### DIFF
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -1612,6 +1612,7 @@ final public class H2O {
     Log.info("OS version: "+System.getProperty("os.name")+" "+System.getProperty("os.version")+" ("+System.getProperty("os.arch")+")");
     long totalMemory = OSUtils.getTotalPhysicalMemory();
     Log.info ("Machine physical memory: " + (totalMemory==-1 ? "NA" : PrettyPrint.bytes(totalMemory)));
+    Log.info("Machine locale: " + Locale.getDefault());
   }
 
   /** Initializes the local node and the local cloud with itself as the only member. */
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
@@ -8,6 +8,7 @@ import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
 /**
@@ -41,7 +42,7 @@ public class BoosterParms extends Iced<BoosterParms> {
    */
   private static Map<String, Object> localizeDecimalParams(final Map<String, Object> params) {
     Map<String, Object> localized = new HashMap<>(params.size());
-    final NumberFormat localizedNumberFormatter = DecimalFormat.getNumberInstance();
+    final NumberFormat localizedNumberFormatter = DecimalFormat.getNumberInstance(Locale.ENGLISH);
     for (String key : params.keySet()) {
       final Object value = params.get(key);
       final Object newValue;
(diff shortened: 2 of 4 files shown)


h2o-core/src/main/java/water/H2O.java
```diff
@@ -1612,6 +1612,7 @@ private static void printAndLogVersion(String[] arguments) {
     Log.info("OS version: "+System.getProperty("os.name")+" "+System.getProperty("os.version")+" ("+System.getProperty("os.arch")+")");
     long totalMemory = OSUtils.getTotalPhysicalMemory();
     Log.info ("Machine physical memory: " + (totalMemory==-1 ? "NA" : PrettyPrint.bytes(totalMemory)));
+    Log.info("Machine locale: " + Locale.getDefault());
```
michalkurka: Nice, thank you!

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
```diff
@@ -41,7 +42,7 @@ public static BoosterParms fromMap(Map<String, Object> map) {
    */
   private static Map<String, Object> localizeDecimalParams(final Map<String, Object> params) {
     Map<String, Object> localized = new HashMap<>(params.size());
-    final NumberFormat localizedNumberFormatter = DecimalFormat.getNumberInstance();
+    final NumberFormat localizedNumberFormatter = DecimalFormat.getNumberInstance(Locale.ENGLISH);
```
michalkurka: the issue here is that all tests will pass regardless if this is correct or wrong because all the test servers are already running US_EN locale

Pscheidl: The test explicitely changes locale just because of that. For a tiny while.

michalkurka: Can we verify it also affect the native code? And not just java?

Pscheidl: I verified it manually by setting locale to english, undo the changes in this PR and ran this test. It failed as well.

Pscheidl: The setLocale call only affects the JVM. I thought you are asking about different combinations of Locale and formatter set to English locale explicitly.

michalkurka: yes, I am not sure this test proves that xgb will work fine when the JVM is started on a different locale

Pscheidl:
The original issue was the opposite. XGBoost was locale sensitive, so we had to do localization in H2O. After I've tried it out after @honzasterba 's XGBoost update, the behavior of XGBoost has changed. It is my fault I did not add test for it earlier, so it slipped undetected. There is a test now.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
Pscheidl: It does. The default, non-arg method looks like this:

```
    public final static NumberFormat getNumberInstance() {
        return getInstance(Locale.getDefault(Locale.Category.FORMAT), NUMBERSTYLE);
    }
```

And we're setting the locale temporarily by using the following method in Locale class:

```
    public static synchronized void setDefault(Locale newLocale) {
        setDefault(Category.DISPLAY, newLocale);
        setDefault(Category.FORMAT, newLocale);
        defaultLocale = newLocale;
    }
```

We should be fine. Setting just FORMAT could be theoretically sufficient, but I see no point in having two different settings on one JVM at one time. And of course, if there are errors and those are printed, the output could be misleading.

Pscheidl: It doesn't test the whole machine with native XGBoost running on different locale. Of course.

PR #3480, opened 2019-05-06
We need to update the reference to a possibly stale Vec in Vec#chunkForCIdx

michalkurka:
The assumption of `Vec#chunkForCIdx` was that Vec is set only once at the beginning and then it never changes. This works fine in most of the cases but breaks if you mutate the Vec on a different node which is not the home node of the Vec and send an MRTask that accesses Chunk#_vec.

This is WIP because any call to chunkForChunkIdx will now update the Vec even when it is a temporary change that won't be propagated to DKV.

cc: @deil87

### DIFF
--- a/h2o-core/src/test/java/water/fvec/VecTest.java
+++ b/h2o-core/src/test/java/water/fvec/VecTest.java
@@ -1,16 +1,14 @@
 package water.fvec;
 
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.Test;
-import water.Futures;
-import water.MRTask;
-import water.Scope;
-import water.TestUtil;
+import water.*;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 import static water.fvec.Vec.makeCon;
+import static water.fvec.Vec.makeConN;
 import static water.fvec.Vec.makeSeq;
 
 /** This test tests stability of Vec API. */
@@ -148,4 +146,50 @@ public class VecTest extends TestUtil {
     source.remove(new Futures()).blockForPending();
     con.remove(new Futures()).blockForPending();
   }
+
+  @Test public void testChunkForChunkIdxAfterVecUpdate() {
+    try {
+      Scope.enter();
+      Vec v1 = Scope.track(makeSeq(2 * FileVec.DFLT_CHUNK_SIZE, false));
+      Vec v2 = Scope.track(makeSeq(2 * FileVec.DFLT_CHUNK_SIZE, false));
+      Chunk c0 = v1.chunkForChunkIdx(0);
+      c0._vec = v2; // inject any vec into the cached POJO, this simulates a stale Vec reference
+      Chunk c0c = v1.chunkForChunkIdx(0);
+      assertSame(c0, c0c);
+      // the vec reference was updated in the existing live object
+      assertSame(v1, c0c._vec);
+    } finally {
+      Scope.exit();
+    }
+  }
+
+  @Test public void testChunkForChunkIdxMRTask() {
+    Assume.assumeTrue(H2O.getCloudSize() > 0);
+    try {
+      Scope.enter();
+      Vec v = Scope.track(makeConN((long) 1e6, 16));
+      // 1. Run an arbitrary MRTask to populate the POJO Chunk caches
+      new MRTask() {
+        @Override
+        public void map(Chunk c) {
+          if (c.vec().get_type() != Vec.T_NUM)
+            throw new IllegalStateException("Expected a numeric Vec");
+        }
+      }.doAll(v);
+      // 2. Install updated Vec in DKV
+      v._type = Vec.T_TIME;
+      DKV.put(v);
+      // 3. Run another MRTask, we should get updated Vec on all nodes not the stale one
+      new MRTask() {
+        @Override
+        public void map(Chunk c) {
+          if (c.vec().get_type() != Vec.T_TIME)
+            throw new IllegalStateException("Expected a time Vec");
+        }
+      }.doAll(v);
+    } finally {
+      Scope.exit();
+    }
+  }
+  
 }
--- a/h2o-core/src/test/java/water/fvec/VecTest.java
+++ b/h2o-core/src/test/java/water/fvec/VecTest.java
@@ -5,6 +5,7 @@ import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import water.*;
+import water.util.ReflectionUtils;
 
 import static org.junit.Assert.*;
 import static water.fvec.Vec.makeCon;
@@ -164,7 +165,7 @@ public class VecTest extends TestUtil {
   }
 
   @Test public void testChunkForChunkIdxMRTask() {
-    Assume.assumeTrue(H2O.getCloudSize() > 0);
+    Assume.assumeTrue(H2O.getCloudSize() > 1);
     try {
       Scope.enter();
       Vec v = Scope.track(makeConN((long) 1e6, 16));
@@ -179,7 +180,9 @@ public class VecTest extends TestUtil {
       // 2. Install updated Vec in DKV
       v._type = Vec.T_TIME;
       DKV.put(v);
-      // 3. Run another MRTask, we should get updated Vec on all nodes not the stale one
+      // 3. Just for fun - show the state of DKV after the update (not needed for the test)
+      printStoreInfo(v._key);
+      // 4. Run another MRTask, we should get updated Vec on all nodes not the stale one
       new MRTask() {
         @Override
         public void map(Chunk c) {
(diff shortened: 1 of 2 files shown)


deil87:
Just to add an extra context to the task. Here is the commit with test that was failing in a distributed scenario https://github.com/h2oai/h2o-3/commit/6d3b27a1023dd219c54c91da2e406598252c5adc

deil87:
@michalkurka are we waiting for the tests that will show that impact of the changes is not causing serious performance issues?

deil87:
@michalkurka maybe you are aware of it but just as a quick feedback.... aside from the possible performance implications you are worried about I have tried to apply these changes and they unfortunately do not fix my test

michalkurka:
@deil87 thanks for catching that - I didn't push the correct fix (facepalm)

I am not worried as much about performance as about changing the semantics of the method. This can create various race conditions if the API is not used properly.

h2o-core/src/test/java/water/fvec/VecTest.java
```diff
@@ -148,4 +146,50 @@ public void map(Chunk c) {
...
+  @Test public void testChunkForChunkIdxAfterVecUpdate() {
+    try {
+      Scope.enter();
+      Vec v1 = Scope.track(makeSeq(2 * FileVec.DFLT_CHUNK_SIZE, false));
+      Vec v2 = Scope.track(makeSeq(2 * FileVec.DFLT_CHUNK_SIZE, false));
+      Chunk c0 = v1.chunkForChunkIdx(0);
+      c0._vec = v2; // inject any vec into the cached POJO, this simulates a stale Vec reference
+      Chunk c0c = v1.chunkForChunkIdx(0);
+      assertSame(c0, c0c);
+      // the vec reference was updated in the existing live object
```
deil87: @michalkurka so is it expected behaviour that we change ref in `c0._vec` from `v2` to `v1` by just calling `v1.chunkForChunkIdx`? Is it the change of semantics that you were worried about?

deil87:
@michalkurka Thanks! Looks like it is fixing an issue. Tests are fine. What else do we need to merge it?

h2o-core/src/test/java/water/fvec/VecTest.java
deil87: @michalkurka I maybe understand the problem that you described in the description. If we mutate not on the home node and send MRTask to `home` the following condition will return true `v == this` and it will return `c` - a stale version. But how do we define `home`? isn't it just the node from where we call MRTask? I'm asking because I wonder how to run into the case that you have described....how to call MRTask not from `home`?

michalkurka:
@deil87 I suggest to not to make this PR a pre-requisite of your work. Please use a workaround until this fix is available.

deil87:
@michalkurka thanks for the suggestion,  will use workaround for the time being.

h2o-core/src/test/java/water/fvec/VecTest.java
michalkurka: Home nodes are determined by where the underlying key that references the Vec is homed.

michalkurka:
Looks good to be merged to me. Objections?

h2o-core/src/test/java/water/fvec/VecTest.java
deil87: @michalkurka But I don't see any logic that is dependant on keys ( that could probably distinguish home from not_home case ). Do we have code like `isHomeKey(key): Boolean`?
 I can only see that we have `v == this` condition that is checking addresses of the objects and just for local copy we can say that `it was already filled in`.  I was trying to understand your concern expressed in the issue's description but still I'm not sure that I can say that I understand how we addressed this concern. Lack of knowledge on my side.

michalkurka:
@Pscheidl please help review

h2o-core/src/test/java/water/fvec/VecTest.java
```diff
@@ -148,4 +146,50 @@ public void map(Chunk c) {
...
+      assertSame(v1, c0c._vec);
+    } finally {
+      Scope.exit();
+    }
+  }
+
+  @Test public void testChunkForChunkIdxMRTask() {
+    Assume.assumeTrue(H2O.getCloudSize() > 0);
```
Pscheidl: This is always true. `> 1` perhaps ?

Pscheidl:
@michalkurka I was actually studying both PRs (this and Andreys) for quite some time today. I wanted to understand the use cases.

First, I do think this is avoidable, as you say. Andrey has his own solution in his PR on how to overcome this behavior. I have a simple one as well: https://github.com/h2oai/h2o-3/commit/7da3e42c694d4d59d284126d1f0ef186d6c84b3f

What you've implemented follows the same pattern with which the previous code in `Vec.chunkForIdX()` was created. The outdated, unsynchronized object replaced by other Vec after updating the frame may still effectively be used, as it holds reference to not yet outdated Chunks. Avoiding this issue in an environment when the two sides both hold a reference to its counterpart and one side's link changes is a little bit more challenging. There are multiple ways to address this, one of them is never to allow clients/users of the API to reach such state. I guess database normal forms are good example.

In this case, it can be fixed in at least two ways:
- Leave the users with current state and let them handle the situations  by themselves - by handling the references correctly. The users should be responsible for refreshing the Vec object.
- if the first point is not applicable, then the original objects should be modified, instead of creating a new object leading to a shared state, which however contains a backwards reference that is always invalid in one of the cases. I guess the update logic could be improved by a strategy pattern on every `Lockable`:

```
  @Override
  public Frame update() {
    final Lockable lockable = this.write_lock();
    for (Vec vec : vecs()){
      // Update the vecs using MRTask here
    }
    
    // Update the rest
    
    lockable.unlock();
    return this;
  }
```

For vecs, it could internally use something like that: https://github.com/h2oai/h2o-3/commit/7da3e42c694d4d59d284126d1f0ef186d6c84b3f

It could even detect changes prior force-updating stuff, which what we're not capable of right now I believe.

This PR fixes the "problem", but the problem can be always avoided. This piece of code creates a **huge precedence** that a client can rely on a piece of shared state being mutated. In fact, it can not be relied on:

```
Chunk original = originalVec.chunkForIdX(0);
Chunk sameChunkFromDifferentVec = modifiedVec.chunkForIdX(0);
assert original == sameChunkFromDifferentVec // True, same chunk objects

assert original._vec == modifiedVec // True - points to new chunk after the change
original._vec = originalVec;

Chunk cAfterChange = modifiedVec.chunkForIdX(0);
assert cAfterChange._vec == modifiedVec; // Should be true, the chunkForIdX did set this

// Do some work here

assert cAfterChange._vec == modifiedVec; // Might actually fail some time later, as the chunk might have been accessed from another vector somewhere else during an asynchronous task or even during the work done in the meantime. Its internal _vec variable points to the last Vector this chunk was accessed by
```

Mutating the state might actually cause data races when used incorrectly in future.

WDYT ?

michalkurka:
@Pscheidl thank you for the analysis, I will answer later in more detail but for now, I wanted to point out that the commit you are referencing https://github.com/h2oai/h2o-3/commit/7da3e42c694d4d59d284126d1f0ef186d6c84b3f doesn't actually fix the problem.

The issue is that Chunk POJO has a live reference to a Vec POJO keyed under key K. When the MRTask is submitted over a Frame that has a Vec with key K, the Vec object is actually the right one (it corresponds to the master object under key K), there is no issue there. The problem is that the cached Chunk POJO has a reference to a stale Vec (that is not referenced anywhere except for the Chunk).

You have no guarantee that the Vec you are updating will actually be in the cache on the node in both Chunk and in the DKV cache.

What your PR does is that it forces to get a cached Vec onto the node and if there was one it updates the existing one or creates a new one. However, there is no guarantee you are updating the same instance that is cached in the Chunk's POJO. They can be different and they will be different.

This commit might even work for @deil87's PR but it is _not correct_ - I wanted to be clear about that.

h2o-core/src/test/java/water/fvec/VecTest.java
```diff
@@ -148,4 +146,50 @@ public void map(Chunk c) {
...
+    try {
+      Scope.enter();
+      Vec v = Scope.track(makeConN((long) 1e6, 16));
+      // 1. Run an arbitrary MRTask to populate the POJO Chunk caches
```
deil87: @michalkurka What I'm trying to understand is what are you referring to by saying "POJO Chunk caches". Is it just Chunk type objects in the heaps of the Nodes where MRTask is being executed? And what you are saying is that by updating newly retrieved from DKV Vec's instances on the remote Node could be not the ones that will be used in second MRTask ( stage 3 in you test below)?

h2o-core/src/test/java/water/fvec/VecTest.java
```diff
@@ -148,4 +146,50 @@ public void map(Chunk c) {
...
+      v._type = Vec.T_TIME;
+      DKV.put(v);
+      // 3. Run another MRTask, we should get updated Vec on all nodes not the stale one
+      new MRTask() {
+        @Override
+        public void map(Chunk c) {
+          if (c.vec().get_type() != Vec.T_TIME)
+            throw new IllegalStateException("Expected a time Vec");
+        }
+      }.doAll(v);
```
deil87: @michalkurka Where will this `v` is going to come from when the MRTask is deserialised on the remote node for execution?

h2o-core/src/test/java/water/fvec/VecTest.java
michalkurka: @deil87 What you actually put in DKV is Value objects (think of them as wrappers), not the POJOs directly. Value object has a raw byte array (iced object) and a live object POJO, this is what I refer to as a cache. The live POJO of a Chunk can be discarded to release memory if the node is under memory pressure. New POJO will be re-instantiated from the raw bytes when needed again.

h2o-core/src/test/java/water/fvec/VecTest.java
michalkurka: Yes!!! That was my intention, will fix. Thanks for noticing. The test doesn't test much on a single node :)

h2o-core/src/test/java/water/fvec/VecTest.java
deil87: @michalkurka 
> You have no guarantee that the Vec you are updating will actually be in the cache on the node in both Chunk and in the DKV cache.

So Chunk Value object stores live Chunk POJO in it and DKV stores Vec Value object with both byte array and live Vec POJO.  When we update Vec in DKV it will only update Vec Value object (and probably discard live Vec's POJO) but there is no logic that will go to Chunk Value object and update/discard live chunk's POJO which stores reference to old live Vec's POJO, right?

h2o-core/src/test/java/water/fvec/VecTest.java
michalkurka: When MRTask is submitted, only keys are actually sent to the remote nodes. The values of keys (Vecs) are then fetched if they are not already present on the remote.

`DKV.put` on line 181 invalidates local replicas for the `v._key` - this means some nodes will have to get the Vec from a remote. Either way, the MRTask and the nodes will have a correct (=cluster consistent) instance of the Vec. However, none of the invalidation is applied to POJOs of Vecs that referenced in Chunks.

deil87: @michalkurka thanks, it looks like my understanding described  in https://github.com/h2oai/h2o-3/pull/3480/files#r288684152 comment is correct.

h2o-core/src/test/java/water/fvec/VecTest.java
michalkurka: > but there is no logic that will go to Chunk Value object and update/discard live chunk's POJO which stores reference to old live Vec's POJO, right?

Yes, that is absolutely correct.

When you update a Vec in DKV (by DKV.put for example) you will update the live object (POJO) on the node that made the update, you will update the bytes on both node who made the update and on the home node. On the home node, the POJO will be invalidated, on other nodes will just invalidate their entry in local H2O.STORE.

After the update, any node that runs DKV.getGet(vec key) will get a correct updated Vec. So far so good.

But none of this involves Chunks... we were only concerned with the content of H2O.STORE, we never touched any Chunks in the process. Which means the cached Vecs in the live chunk POJOs can have a stale instance of the Vec that is not actually reachable from anywhere else.

Because you can only get to a Chunk instance from `Vec#chunkForChunkIdx` this a good place to fix it. There are alternative solutions as well... but I find this approach reasonable and it has no performance harm.

h2o-core/src/test/java/water/fvec/VecTest.java
michalkurka: Yep, it is correct :)

h2o-core/src/test/java/water/fvec/VecTest.java
Pscheidl: Cascades !

Pscheidl: Update Cascades ?! Leveraging the strategy pattern seems to be the natural choice during update operations ?

PR #3481, opened 2019-05-07
PUBDEV-6440 - Unable to parse files from s3 [using s3's directory url with s3 protocol]

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6440

In **S3**, everything is an object. Including folders. Which means H2O was trying to parse folders, as these were not filtered out.

- Filtering folders based on `/` postfix.
- Current logic also walks all the subfolders, This behavior has not been changed.
- Pattern was not taken into account when `importFiles` was invoked. This has been fixed, as users might need to filter the files in the subdirectories (for example).
- Flow tested

FireShot Capture 001 - H2O Flow - 192 168 59 179

Note: This PR is only about S3. S3A has different handling of everything.

### DIFF
--- a/h2o-persist-s3/src/main/java/water/persist/PersistS3.java
+++ b/h2o-persist-s3/src/main/java/water/persist/PersistS3.java
@@ -180,7 +180,8 @@ public final class PersistS3 extends Persist {
   }
 
 
-  private static void processListing(ObjectListing listing, final String pattern, ArrayList<String> succ, ArrayList<String> fail, boolean doImport) {
+  private static void processListing(ObjectListing listing, String pattern, ArrayList<String> succ, ArrayList<String> fail, boolean doImport) {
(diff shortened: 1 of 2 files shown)


h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -488,8 +490,8 @@ protected String wrapKey(String s) {
   }
 
 
-  Cache _bucketCache = new Cache();
-  HashMap<String, KeyCache> _keyCaches = new HashMap<>();
+  static volatile Cache _bucketCache = new Cache();
```
Pscheidl: This was here for eternity and was accidentally removed for no reason (probably as a leftover) in of my PRs and nobody noticed. Returned it back.

Pscheidl:
Test failure unrelated: `runit_PUBDEV_5705_drop_columns_parser_arff_large.R`

PR #3483, opened 2019-05-07
PUBDEV-6476: Experimenting with Rabit.init in order to speed-up scoring

michalkurka:
@honzasterba the first commit can go in, the second one is just experimental

This PR shows that killing all calls to Rabit.init/shutdown significantly speeds-up scoring and brings it to performance pre-xgboost upgrade. I think it is conclusive that this is the cause of the slowdown reported in PUBDEV-6476.


honzasterba:
I have finalized the changes here: https://github.com/h2oai/h2o-3/pull/3489

michalkurka:
Not needed anymore

PR #3487, opened 2019-05-08
PUBDEV-6476: XGBoost - better (faster) way of extracting booster bytes

michalkurka:
Avoid using Rabit.init/shutdown - it is causing slowdown with new XGB


michalkurka:
this is the non-exerimental part of https://github.com/h2oai/h2o-3/pull/3486

PR #3488, opened 2019-05-09
[PUBDEV-6108] expose coxPh to python

honzasterba:
any suggestions on how to improve the API are welcome


h2o-py/tests/testdir_algos/coxph/pyunit_coxph_smoke.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import print_function
+import sys
+sys.path.insert(1,"../../../")
+import h2o
+from tests import pyunit_utils
+from h2o.estimators.coxph import H2OCoxProportionalHazardsEstimator
```
Pscheidl: I did a small check if this is correctly exposed in `h2o.estimators` as the rest of the algos. It is. So just doing `from h2o.estimators import H2OCoxProportionalHazardsEstimator` works. :heavy_check_mark:

michalkurka:
@honzasterba looks good - I think this is great

How does the model summary look like in comparison to R? I think we should get feedback from users. Perhaps port the demo we have for R to Python as well to see if we have the main features?

PR #3489, opened 2019-05-09
PUBDEV-6476 - Finish cleanup of Rabit.init calls

honzasterba:


honzasterba:
bechmarked on mac locally and we are back to pre-upgrade numbers for train times

PR #3493, opened 2019-05-09
PUBDEV-4076 custom distribution proof of concept

maurever:
Implement possibility to use custom loss function (distribution) in GBM. 

JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-4076


h2o-core/src/main/java/water/udf/CDistributionFunc.java
```diff
@@ -0,0 +1,86 @@
...
+    public String linkInvString(String f);
+
+    /**
+     * Deviance of given distribution function at predicted value f
+     * @param w observation weight
+     * @param y (actual) response
+     * @param f (predicted) response in original response space (including offset)
+     * @return deviance
+     */
+    public double deviance(double w, double y, double f);
```
michalkurka: only needed for metrics, imho

h2o-core/src/main/java/water/udf/CDistributionFunc.java
```diff
@@ -0,0 +1,86 @@
+package water.udf;
+
+import java.io.Serializable;
+
+/**
+ * Custom Distribution Function
+ */
+public interface CDistributionFunc extends CFunc, Serializable {
```
michalkurka: Might make sense to split into 3-4 logical classes.

h2o-core/src/main/java/water/udf/CDistributionFunc.java
maurever: Yes, it is true, however, how do you want to calculate deviance if a user defined a custom distribution only for the gradient calculation?

h2o-core/src/main/java/water/udf/CDistributionFunc.java
maurever: Yeah, I'm still working on the structure...

h2o-core/src/main/java/water/udf/CDistributionFunc.java
michalkurka: That is a good point. My thinking was that it is only used for metrics for regression models, so maybe it is one thing we could make optional/move to another class. But I think that is too early to decide. Please continue with your POC and don't worry - I am just thinking out loud.

maurever:
The custom distribution approach is implemented. What I did:
- custom distribution loading and using
- MOJO and POJO support
- basic doc
- JUnit and pyunit tests

It is the first step to customize loss function. However, the way how to use it now could not be so clear for users. The next steps are:
- predefine type of the most used distributions which user can reuse and edit only important methods (for example for loss calculation only gradient and gamma is necessary)
- improve doc - edit methods description to be clear
- add an example on real data, how the customization can help

WDYT @michalkurka?

h2o-core/src/main/java/hex/DistributionFactory.java
```diff
@@ -439,30 +421,116 @@ public QuantileDistribution(Model.Parameters params){
...
+        super(params);
+        customDistribution = new CustomDistributionWrapper(CFuncRef.from(params._custom_distribution_func));
+        assert customDistribution != null;
+        assert customDistribution.getFunc() != null;
+        this.setLinkFunction(LinkFunctionFactory.getLinkFunction(customDistribution.getFunc().link()));
+    }
+    
+    public static CustomDistribution getCustomDistribution(Model.Parameters params){
+        if(inst == null){
+            System.out.println("new instance");
```
michalkurka: Println

h2o-core/src/main/java/hex/DistributionFactory.java
```diff
@@ -439,30 +421,116 @@ public QuantileDistribution(Model.Parameters params){
...
+    
+    private CustomDistributionWrapper customDistribution;
+    private static CustomDistribution inst;
+    
+    private CustomDistribution(Model.Parameters params){
+        super(params);
+        customDistribution = new CustomDistributionWrapper(CFuncRef.from(params._custom_distribution_func));
+        assert customDistribution != null;
+        assert customDistribution.getFunc() != null;
+        this.setLinkFunction(LinkFunctionFactory.getLinkFunction(customDistribution.getFunc().link()));
```
michalkurka: calling overridable methods from a constructor is a bad practice because of descendants of this class can override the method and assume the class is fully constructed when the method is called

h2o-core/src/main/java/hex/DistributionFactory.java
```diff
@@ -439,30 +421,116 @@ public QuantileDistribution(Model.Parameters params){
...
+    public double negHalfGradient(double y, double f) { return y > f ? 0.5 * quantileAlpha : 0.5 * (quantileAlpha - 1); }
+}
+
+/**
+ * Custom distribution class to customized loss and prediction calculation.
+ * Currently supported only for GBM algorithm.
+ */
+class CustomDistribution extends Distribution {
+    
+    private CustomDistributionWrapper customDistribution;
```
michalkurka: breaks h2o naming conventions - fields should start with `_`

this applies to the whole PR - please keep it consistent with the existing code base

h2o-core/src/main/java/hex/Model.java
```diff
@@ -873,10 +879,9 @@ public Model(Key<M> selfKey, P parms, O output) {
     _output = output;  // Output won't be set if we're assert output != null;
     if (_output != null)
       _output.startClock();
-    _dist = isSupervised() && _output.nclasses() == 1 ? DistributionFactory.getDistribution(_parms) : null;
+    _dist = isSupervised() && _output.nclasses() == 1 ? DistributionFactory.getDistribution(parms) : null;
```
michalkurka: why this change? once the field is set, we should keep working just with the field reference - if anyone clones or modifies the object it breaks your code in a bad way

h2o-core/src/main/java/hex/ModelMetricsRegression.java
```diff
@@ -41,7 +41,11 @@ public static ModelMetricsRegression getFromDKV(Model model, Frame frame) {
   public String toString() {
     StringBuilder sb = new StringBuilder();
     sb.append(super.toString());
-    sb.append(" mean residual deviance: " + (float)_mean_residual_deviance + "\n");
+    if(!Double.isNaN(_mean_residual_deviance)) {
```
michalkurka: IMHO we should print "NA" instead of skipping the field - this is an use

h2o-core/src/main/java/hex/ModelMetricsRegression.java
```diff
@@ -118,10 +122,16 @@ public MetricBuilderRegression(Distribution dist) {
       _abserror += w*Math.abs(err);
       _rmslerror += w*err_msle;
       assert !Double.isNaN(_sumsqe);
-      if (m != null && m._parms._distribution != DistributionFamily.huber)
-        _sumdeviance += m.deviance(w, yact[0], ds[0]);
-      else if (_dist!=null)
-        _sumdeviance += _dist.deviance(w, yact[0], ds[0]);
+      
+      // Deviance method is not support in custom distribution
```
michalkurka: typo, "not supported"

h2o-core/src/main/java/hex/ModelMetricsRegression.java
```diff
@@ -158,8 +168,10 @@ public ModelMetrics makeModelMetrics(Model m, Frame f, Frame adaptedFrame, Frame
 
           meanResDeviance = new MeanResidualDeviance(_dist, preds.anyVec(), actual, weight).exec().meanResidualDeviance;
         }
-      } else {
+      } else if((m != null && m._parms._distribution != DistributionFamily.custom) || (_dist != null && _dist.distribution != DistributionFamily.custom) ) {
```
michalkurka: can you please explain why can we not really on a single source of the family parameter?

michalkurka:
@maurever great work!

I especially appreciate the test coverage and thorough documentation.

I would recommend fixing the naming conventions (field names start with `_`), I also posted smaller comments.

h2o-core/src/main/java/hex/DistributionFactory.java
maurever: Yes, sure. That is absolutely true.

h2o-core/src/main/java/hex/Model.java
maurever: It was a mistake when I tried something to change here. Good catch.

h2o-core/src/main/java/hex/ModelMetricsRegression.java
maurever: Now, it says `mean residual deviance: NaN`,  if it is not correct, I can change it to  `mean residual deviance: N/A`.

h2o-core/src/main/java/hex/ModelMetricsRegression.java
maurever: Because `Model m` can be null and then it throws null pointer exception.

h2o-core/src/main/java/hex/DistributionFactory.java
maurever: Ok, I will change it.

maurever:
@angela0xdata, can you please review the doc parts? Thank you. :)

angela0xdata:
Made a couple of minor changes so that this new function is included in the Parameters Appendix and in the list of GBM parameters.

maurever:
Thank you @angela0xdata!

PR #3498, opened 2019-05-10
PUBDEV-6162 - Dynamic change of S3A credentials

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6162

Stripping/redacting and re-inserting in-url credentials is dangerous and there are plans not to support it in future (https://issues.apache.org/jira/browse/HADOOP-14439), as it is a security threat. Also, the behavior may vastly change across different versions (some leave the in-url credentials once objects are returned, some do not).

A good way to do it is to introduce our own dynamic credentials provider, taking the credentials from DKV and providing users with functionality of setting the credentials. Just like we did for S3.


h2o-r/h2o-package/R/import.R
```diff
@@ -367,5 +367,27 @@ h2o.set_s3_credentials <- function(secretKeyId, secretAccessKey){
...
+#'
+#' Sets AWS credentials for Hadoop S3A dynamically
+#'
+#' Providing incorrect credentials results in an error during the first related S3A import call.
+#'
+#' @param accessKeyId Amazon S3 Secret Key ID (provided by Amazon, corresponds to fs.s3.awsAccessKeyId)
+#' @param secretAccessKey Amazon S3 Secret Access Key (provided by Amazon, corresponds to fs.s3.awsSecretAccessKey)
+#' 
+#' @export
+h2o.set_s3a_credentials <- function(accessKeyId, secretAccessKey){
```
Pscheidl: according to tidyverse, the args should probably be snake case, but the S3 function above is camel case, so I wanted to keep those two together.

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -79,6 +76,7 @@
         Log.debug("Cannot find HADOOP_CONF_DIR or YARN_CONF_DIR - default HDFS properties are NOT loaded!");
       }
     }
+    conf.set("fs.s3a.aws.credentials.provider", "water.persist.H2ODynamicS3ACredentialsProvider,com.amazonaws.auth.EnvironmentVariableCredentialsProvider,com.amazonaws.auth.InstanceProfileCredentialsProvider,org.apache.hadoop.fs.s3a.AnonymousAWSCredentialsProvider");
```
Pscheidl: source: https://hadoop.apache.org/docs/r3.0.3/hadoop-aws/tools/hadoop-aws/index.html chapter ` Changing Authentication Providers`

michalkurka:
Looks like duplicated code for S3 credentials, can we just have one endpoint that would serve both? AWS credentials can be in the core to avoid code duplication like this.

Pscheidl:
@michalkurka As commented above, the "duplication" was a choice. Will unify.

Pscheidl:
Now, there is one API to rule them all. Works for both S3 and S3A. Created a new common module to avoid putting it into core.

Pscheidl:
Test failures unrelated, re-running one more time.

build.gradle
```diff
@@ -111,6 +111,7 @@ ext {
       project(':h2o-jetty-9'),
       project(':h2o-ext-jython-cfunc'),
       project(':h2o-hive'),
+      project(':h2o-persist-s3-common'),
```
michalkurka: should be in published projects as well, it can also be just in h2o-core - it will never be excluded from any assembly

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -79,6 +75,7 @@
...
```
michalkurka: But what if the users have their own configuration? This would overwrite it

Pscheidl: What is meant by "their own configuration" ? 

There is:
- H2ODynamicS3ACredentialsProvider (ours)
- EnvironmentVariableCredentialsProvider
- InstanceProfileCredentialsProvider
- AnonymousAWSCredentialsProvider

Is there any other missing ?

build.gradle
Pscheidl: I made it part of core. Finally makes sense to me. You were right from the beginning.

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
michalkurka: The user can specify anything (including their custom provider) in Hadoop configuration for the property "fs.s3a.aws.credentials.provider". Different Hadoop distributions can have different default providers, I could imagine eg. EMR will have very different providers.

My point was that we need to respect what the user has specified in their Hadoop configuration and avoid overwriting something unintentionally.

Pscheidl: For other reviewers: I'll try to find out if simply detecting this value is set to anything at runtime and then avoiding overriding it with our provider chain is enough, or there is any caveeat to it.

Pscheidl: @michalkurka 

Once the configuration file (in this simple case the default one `etc/hadoop/core-site.xml`) is modified to look like this:

```xml
<configuration>
	<property>
		<name>fs.s3a.aws.credentials.provider</name>
		<value>
			org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider,
			com.amazonaws.auth.EnvironmentVariableCredentialsProvider,
			com.amazonaws.auth.InstanceProfileCredentialsProvider
		</value>
	</property>
</configuration>
```

The value is detectable at the time the CONF object is constructed. Which means we can detect changes to this value and not modify it if user has previously set it to a different value in the configuration file. Let me demonstrate it simply with a screenshot:

Snímek z 2019-05-19 13-11-39

PR #3499, opened 2019-05-10
PUBDEV-6491: Fix NPE in HostnameGuesser and improve error reporting

michalkurka:


h2o-core/src/main/java/water/init/HostnameGuesser.java
```diff
@@ -25,9 +25,7 @@ public static InetAddress findInetAddressForSelf(String ip, String network) thro
     }
 
     ArrayList<CIDRBlock> networkList = calcArrayList(network);
-    if (networkList == null) {
-      throw new HostnameGuessingException("No network found! Exiting.");
-    }
+    assert networkList != null;
```
jakubhava: Just small comment: Assertions can be disabled on production machine, is this safe?

michalkurka: I would think so, the method never returns `null` anymore and there is a test. This is just to state the intent and make the invariant clear here, not necesesarily to catch anything

jakubhava: Would it be better to document output of `calcArrayList` and mention that it never returns null? 

Either with annotation or at least regular comment. I find this use of assert a bit of confusing.

michalkurka: @jakubhava thanks, what kind of annotation do you have in mind? Can you point me to an example, please?

jakubhava: `@NotNull`

michalkurka: Can you be more specific? I did my research yesterday but this whole annotation business seems to be a big mess to me. Different libraries providing different annotations NotNull vs NonNull and dormant JSRs.

This is a good source of info: https://stackoverflow.com/questions/4963300/which-notnull-java-annotation-should-i-use/4963565#4963565

jakubhava: It's not critical, I will leave it up on you. 

But my personal opinion is that small comment or annotation is better then using `assert` like this.

michalkurka: Seems like that one is only part of Java EE though. I will remove the assert. Thank you

PR #3501, opened 2019-05-13
[Experiment] Modified SE calculating in GBM/DRF for PUBDEV-6495

michalkurka:


h2o-algos/src/main/java/hex/tree/DTree.java
```diff
@@ -923,10 +925,9 @@ private static Split findBestSplitPoint(DHistogram hs, int col, double min_rows,
     double best_seR=Double.MAX_VALUE;   // squared error for right side of the best split (so far)
     DHistogram.NASplitDir nasplit = DHistogram.NASplitDir.None;
 
-    // squared error of all non-NAs
-    double seNonNA = wYYhi[0] - wYhi[0]* wYhi[0]/ whi[0]; // Squared Error with no split
-    if (seNonNA < 0) seNonNA = 0;
-    double seBefore = seNonNA;
+
+    // SE with no split
+    final double seBefore = (wYYhi[0] + wYYNA) - (wYhi[0] + wYNA) * (wYhi[0] + wYNA) / (whi[0] + wNA);
```
maurever: If I understand correctly: the bug was in total SE there was not SE with NA included? Interesting.

PR #3502, opened 2019-05-15
[Collaboration] TE blogpost

deil87:
Purpose of PR is to make it easier for a reviewers to add their comments.

TODO:
 - consider to add `holdout_type`= none  scenario explanation


meganjkurka:
@deil87 What is the goal of the blogpost?  Is it so that customers understand why target encoding can be helpful and how to use it?

deil87:
@meganjkurka Probably I need to extend `Purpose` section in the blogpost if you are asking :) But basically yes, I wanted to share what I think might be useful for users. People (Marek, Martin, Angela, one user from Moscow) were asking me questions and I decided to try to explain it little bit deeper. But overall it is more about what it is and how to use it (with our API).  I'm not trying to explain here WHY it might be useful. Did I answer your question?

meganjkurka:
It sounds like the questions you were getting were from people at H2O.  The questions I have heard from customers are around why would one want to use it and our recommended best practices.  They are interested from a practical perspective - i.e. how does this improve my lift and why.  

I would also recommend that if this is a blog post it should be kept simple with intuitive explanations.  Information about formulas or naming conventions probably make more sense in a formal paper.

deil87:
@meganjkurka This is close to the true that it is mostly H2O people. Those are people who wanted to apply it and also to understand how it is working. I think that you can't answer WHY something is needed or working without first answering what it is doing under the hood. For me it is completely understandable why **customers** asking _why would anyone want to use it_ - this is the most high level question and people usually ask that kind of questions when they hear about something for the very first time. I actually answered question WHY in the document but briefly. And I think these brief answers might satisfy 90% of the customers. Another simple answer for them could be that it just improves performance of the model. And then they will start to wonder how to try it and what are the best practices. 
Regarding best practices.... this is a serious step. I would not dare to state anything without comprehensive research. We can set defaults to a something that we believe is giving us best results but it is not completely the same as saying publicly that some let's say strategies are better than other. 
And we had Target Encoding before ... did we provide best practices in our docs or blogs for it?

To summarise: this blogpost not for those who is wondering why to use TE, but for those who want to understand what is it, what we should be aware of in general ( like data leakage) and how to use it correctly in terms of our API.

deil87:
@meganjkurka Thank you, I got you point about simplicity of the blog post. But I have seen different blogposts which were of a different complexity and length. I think that with words `in Depth` in the title I'm eligible to get deeper into details. I agree that I might have failed somewhere in finding right words - but this is what review process is for. This is an art to be able to write about something complex in a simple way.

deil87:
@meganjkurka maybe if we need simpler version than we should consider two different blog posts?

meganjkurka:
We do have best practices in our documentation and tutorials - we tell them about the options but only show an example of the steps that are recommended.  

As I mentioned, I think this paper may be helpful to internal h2o users that are working on stuff like Sparkling Water API but will not be helpful to a general external audience.  

My recommendations are:
* explain why this can be helpful
* keep the explanations intuitive
* include an example of using target encoding with our best practices (where best practices could mean try each configuration and see which does best)

michalkurka:
@deil87 Good idea about the second blog post

It would be good to have a blog post/tutorial with a simple explanation who just want to get some results fast with H2O, point out where is the H2O advantage and then provide the second one with an in-depth explanation.

PR #3503, opened 2019-05-15
PUBDEV-6476: Fix slow XGBoost scoring on OS X

michalkurka:
This PR incorporates XGBoost version with a patched Rabit module: https://github.com/h2oai/xgboost/pull/71 (this patch improves performance H2OXGBoost on OS X)


michalkurka:
@honzasterba FIY

PR #3504, opened 2019-05-16
PUBDEV-6503: Allow specifying an external IP for h2odriver callback

michalkurka:
Use case: h2odriver running in a docker contain or somehow isolated from data nodes (data nodes cannot reach the driver node), we need to give an outside IP address to the H2O nodes to use for a callback instead of the one detected in the container/isolated environment.

This would be used in Cloudera Data Science Workbench.

The user run starts the h2odriver in a container and specifies a known external IP for the callback using the `-extdriverif` parameter. It is expected the port doesn't have to be re-mapped like this because the outside port would be known and the container port can be chosen to be the same.

Example (a bit contrived, only for demo purposes):

1. Run ssh tunnel forwarding 172.16.2.204:40000 to 172.16.2.181:40000: `michalk@172.16.2.204:~$ ssh -L \*:40000:localhost:40000 michalk@172.16.2.181`
2. On 172.16.2.181 run h2odriver with mapper->driver callback routed through 172.16.2.204:
    `michalk@172.16.2.181:~$ hadoop jar h2odriver.jar -extdriverif 172.16.2.204 -driverif 127.0.0.1 -driverport 40000 -n 2 -mapperXmx 4g`

Driver prints:

    Using mapper->driver callback IP address and port: 172.16.2.204:40000 (internal callback address: 127.0.0.1:40000)

### DIFF
--- a/h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
+++ b/h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
@@ -104,7 +104,8 @@ public class h2odriver extends Configured implements Tool {
   static String mapperXmx = null;
   static int extraMemPercent = -1;            // Between 0 and 10, typically.  Cannot be negative.
   static String mapperPermSize = null;
-  static String driverCallbackIp = null;
+  static String driverCallbackBindIp = null;
+  static String driverCallbackPublicIp = null;
   static int driverCallbackPort = 0;          // By default, let the system pick the port.
   static PortRange driverCallbackPortRange = null;
   static String network = null;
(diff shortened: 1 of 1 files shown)


h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
```diff
@@ -856,9 +857,13 @@ else if (s.equals("-mapperPermSize")) {
         i++; if (i >= args.length) { usage(); }
         mapperPermSize = args[i];
       }
+      else if (s.equals("-extdriverif")) {
```
Pscheidl: Please document this in the `usage()` method.

michalkurka:
@pavel not all parameters are documented, only the ones that are used frequently are.

Do you still think this should be in the usage output?

PR #3505, opened 2019-05-16
Introduced FeatureTransformer and corresponding genmodel which in particular can be used for storing target encoding map

deil87:
FeatureTransformerWriter_v2

Note: PR is based on MOJO support PR PUBDEV-6255. TojoTransformer should be probably renamed and turn into generic class that will be supporting all future loaded transformers.

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
@@ -341,7 +341,7 @@ public class TargetEncoder {
           if (num.isNA(i) || den.isNA(i))
             encodings.setNA(i);
           else if (den.at8(i) == 0) {
-            Log.info("Denominator is zero. Imputing with _priorMean = " + _priorMean);
+            Log.info("Denominator is zero for column index = " + _encodingsIdx + ". Imputing with _priorMean = " + _priorMean);
             encodings.set(i, _priorMean);
           } else {
             double numberOfRowsInCurrentCategory = den.atd(i);
--- a/h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
@@ -221,7 +221,7 @@ public class TargetEncoder {
         return true;
     }
 
-    Frame groupByTEColumnAndAggregate(Frame data, int teColumnIndex) {
+    public static Frame groupByTEColumnAndAggregate(Frame data, int teColumnIndex) {
       int numeratorColumnIndex = data.find("numerator");
       int denominatorColumnIndex = data.find("denominator");
       AstGroup.AGG[] aggs = new AstGroup.AGG[2];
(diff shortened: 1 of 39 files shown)


michalkurka:
2 questions:

1. Why is TE on Model? Shouldn't it be independent of a Model object?

2. Lots of duplicated because of the new interface. I think you "FeatureTransformer" is actually very similar to "Mojo" or "Model" in many aspects. Also, each Model can be seen as something that generates features for a dependent model. IMHO your feature fits the functionality of GenModel because `score0` can be used to return the encoded value. I am not sure if all this duplication is needed. People might want a MOJO just for Target Encoder, that has value on its own. MOJO is an adopted interface and we need to make sure there is really a good reason for "TOJO". Big argument in favor of MOJO is that you can actually chain it and create MOJO pipelines using https://github.com/h2oai/h2o-3/blob/master/h2o-genmodel/src/main/java/hex/genmodel/MojoPipelineBuilder.java#L14. My question is: is it necessary to have a parallel infrastructure in the code for "TOJO"? 

cc: @tomkraljevic

tomkraljevic:
I haven't thought about this much, but I don't think I would invent a new TOJO.
To me, isn't target just turning a RowData into another RowData?

I would start top-down with what you want the end-user code to look like, get that code reviewed, and then go from there.

Something like this (this is pseudocode, not real code!) would make sense to me:

```
EasyPredictModelWrapper te = new EasyPredictModelWrapper(MojoModel.load("TE_model_R_nnn.zip"));

EasyPredictModelWrapper gbm = new EasyPredictModelWrapper(MojoModel.load("GBM_model_R_1475248925871_74.zip"));

while (have rows) {
    RowData rawRow = blah;
    RowData teRow = te.transform(rawRow);
    BinomialModelPrediction p = gbm.predictBinomial(teRow);
    // use p
    // ...
}
```

deil87:
@michalkurka I created this PR based on the Mojo support for TE because I knew I would be reusing some code and tests from there. I should have probably removed what is not needed anymore.
1) I think the answer to this question is above. I need to remove previous approach
2) There is no duplicated code in the new approach. FeatureTransformer of course is similar but it is not the same so from the OOP point of view it should be separated. For example in spark they don't reuse Models for Transformers.  
It is always possible to stretch concepts to make them look alike.... from Category Theory  point of view everything could be seen as just mapping from one state to another. I'm personally agains thins like using `score0` (which is already something that is not very intuitive because of the `0`) for passing encodings. It will contribute to the developers debt and will make it confusing.
 Another point is that some transformers can produce more than one column (OHO). Btw I'm not sure but we probably do OHO in some models.... how do we represent it in mojo? do we have pipeline stage for this in SW?
If PojoPipelineBuilder is something really useful I think it could be also reused and generalised into two hierarchies later. 
Question about having parallel infrastructure sounds like ` is it necessary to have different classes for Models because they all basically do the same`. 

If Mojo is what concerns you than maybe we still can reuse this genmodel part. But we still can separate Models from transformers in h2o-core. But I would not recommend it as there are lots of methods that are really Model related....like storing domains, algos, thresholds, distributions e.t.c. In case we decide to reuse Mojo we will either add for 99% of the logic `if_else` blocks.

deil87:
I understand that MOJO is almost our marketing brand and inventing TOJO could be something that will require to change a lot of slides and "t-shirts" :) I would still try to split... not by duplication but buy reusing common methods.

deil87:
@tomkraljevic Yes it should look like this. But let me point out to an interesting thing in your pseudocode -> `te.transform` - there is no such a method as `transform` on the models. Of course we can extend *easily* our EasyPredictModelWrapper to support `transformer models` but we at least will have to change its name :)

michalkurka:
word2vec has `transform0`, `score0` on GLRM mojo is actually `transform`.

My point was that models do transformations as well. It is fine to a have hierarchy of classes (I think that would be useful and we clean-up some things) but they should have a common parent and should be loadable by MojoLoader.

deil87:
@michalkurka word2vec is a transformer as well(in spark at least) and I don't know what GLRM is doing yet... but for me it really looks like things started to get mixed up from some point. Of course every design decision could be memorised but it will not be self-explanatory for new joiners. Maybe `word2vec` could be moved later under Transformers as well.
I agree about MojoLoader that it is worth to try it....and I **was** trying. Give me some time... maybe there is a decent solution for reusing Mojo.

deil87:
Going to close this PR. Decided to follow existing paths to add transformers (as models)

PR #3508, opened 2019-05-18
XGBoost: Clean-up + speed-up of Sparse Matrix allocation

michalkurka:
@Pscheidl I noticed we spend a significant amount of time calculating dimensions of the CSR sparse matrix (see attached screenshot - 4 minutes no progress and the code is just calculating the matrix size) on larger dataset. Only single thread is doing something.

This PR makes the calculation parallel.

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
@@ -10,6 +10,7 @@ import water.MrFun;
 import water.fvec.Chunk;
 import water.fvec.Frame;
 import water.fvec.Vec;
+import water.util.ArrayUtils;
 import water.util.Log;
 import water.util.VecUtils;
 
@@ -347,6 +348,10 @@ public class XGBoostUtils {
                                DataInfo di, float[] resp, float[] weights)
         throws XGBoostError {
 
+
+        SparseMatrixDimensions sparseMatrixDimensions = calculateCSRMatrixDimensions(f, chunksIds, weightsVec, di);
+        SparseMatrix sparseMatrix = allocateCSRMatrix(sparseMatrixDimensions);
+
         Vec.Reader[] vecs = new Vec.Reader[f.numCols()];
         for (int i = 0; i < vecs.length; ++i) {
             vecs[i] = f.vec(i).new Reader();
@@ -354,13 +359,11 @@ public class XGBoostUtils {
         Vec.Reader weightsReader = (weightsVec != null) ? weightsVec.new Reader() : null;
         Vec.Reader responseReader = responseVec.new Reader();
 
-        SparseMatrixDimensions sparseMatrixDimensions = calculateCSRMatrixDimensions(f, chunksIds, vecs, weightsReader, di);
-        SparseMatrix sparseMatrix = allocateCSRMatrix(sparseMatrixDimensions);
-
         int actualRows = initalizeFromChunkIds(
                 f, chunksIds, vecs, weightsReader,
                 di, sparseMatrix._rowHeaders, sparseMatrix._sparseData, sparseMatrix._colIndices,
                 responseReader, resp, weights);
+
         return toDMatrix(sparseMatrix, sparseMatrixDimensions, actualRows, di);
     }
 
@@ -623,29 +626,74 @@ public class XGBoostUtils {
         return new SparseMatrixDimensions(nonZeroElementsCount, ++rowIndicesCount);
     }
 
-    protected static SparseMatrixDimensions calculateCSRMatrixDimensions(Frame f, int[] chunks, Vec.Reader[] vecs, Vec.Reader w, DataInfo di) {
-        long nonZeroElementsCount = 0;
-        long rowIndicesCount = 0;
+    static SparseMatrixDimensions calculateCSRMatrixDimensions(Frame f, int[] chunkIds, Vec w, DataInfo di) {
+        CalculateCSRMatrixDimensionsMrFun fun = new CalculateCSRMatrixDimensionsMrFun(f, di, w, chunkIds);
+        H2O.submitTask(new LocalMR(fun, chunkIds.length)).join();
+        
+        return new SparseMatrixDimensions(ArrayUtils.sum(fun._nonZeroElementsCounts), ArrayUtils.sum(fun._rowIndicesCounts) + 1);
+    }
 
-        for (Integer chunk : chunks) {
-            for (long i = f.anyVec().espc()[chunk]; i < f.anyVec().espc()[chunk + 1]; i++) {
-                if (w != null && w.at(i) == 0) continue;
-                rowIndicesCount++;
+    private static class CalculateCSRMatrixDimensionsMrFun extends MrFun<CalculateCSRMatrixDimensionsMrFun> {
+        private Frame _f;
+        private DataInfo _di;
+        private Vec _w;
+        private int[] _chunkIds;
 
-                nonZeroElementsCount+= di._cats;
+        // OUT
+        private long[] _rowIndicesCounts;
+        private long[] _nonZeroElementsCounts;
 
-                for (int j = 0; j < di._nums; ++j) {
-                    double val = vecs[di._cats + j].at(i);
-                    if (val != 0) {
-                        nonZeroElementsCount++;
+        CalculateCSRMatrixDimensionsMrFun(Frame f, DataInfo di, Vec w, int[] chunkIds) {
+            _f = f;
+            _di = di;
+            _w = w;
+            _chunkIds = chunkIds;
+            _rowIndicesCounts = new long[chunkIds.length];
+            _nonZeroElementsCounts = new long[chunkIds.length];
+        }
+
+        @Override
+        protected void map(int i) {
+            final int cidx = _chunkIds[i];
+
+            long rowIndicesCount = 0;
+            long nonZeroElementsCount = 0;
+
+            if (_di._nums == 0) {
+                if (_w == null) {
+                    // no weights and only categoricals => sizing is trivial
+                    rowIndicesCount = _f.anyVec().chunkForChunkIdx(cidx)._len;
+                    nonZeroElementsCount = rowIndicesCount * _di._cats;
+                } else {
+                    Chunk ws = _w.chunkForChunkIdx(cidx);
+                    for (int r = 0; r < ws._len; r++)
+                        if (ws.atd(r) != 0) {
+                            rowIndicesCount++;
+                            nonZeroElementsCount += _di._cats; 
+                        }
+                }
+            } else {
+                Chunk[] cs = new Chunk[_di._nums];
+                for (int c = 0; c < cs.length; c++) {
+                    cs[c] = _f.vec(_di._cats + c).chunkForChunkIdx(cidx);
+                }
+                Chunk ws = _w != null ? _w.chunkForChunkIdx(cidx) : null;
+                for (int r = 0; r < cs[0]._len; r++) {
+                    if (ws != null && ws.atd(r) == 0)
+                        continue;
+                    rowIndicesCount++;
+                    nonZeroElementsCount += _di._cats;
+                    for (int j = 0; j < _di._nums; j++) {
+                        if (cs[j].atd(r) != 0) {
+                            nonZeroElementsCount++;
+                        }
                     }
                 }
             }
+            _rowIndicesCounts[i] = rowIndicesCount;
+            _nonZeroElementsCounts[i] = nonZeroElementsCount;
         }
-
-        return new SparseMatrixDimensions(nonZeroElementsCount, ++rowIndicesCount);
     }
-
 
     /**
      * Dimensions of a Sparse Matrix
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
@@ -22,12 +22,6 @@ import static water.MemoryManager.malloc8;
 
 public class XGBoostUtils {
 
-    /**
-     * Arbitrary chosen initial size of array allocated for XGBoost's purpose.
-     * Used in case of sparse matrices.
-     */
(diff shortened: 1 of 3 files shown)


michalkurka:
Screen Shot 2019-05-16 at 4 36 37 PM

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -872,30 +626,75 @@ protected static SparseMatrixDimensions calculateCSRMatrixDimensions(Chunk[] chu
...
+
+        @Override
+        protected void map(int i) {
+            final int cidx = _chunkIds[i];
+
+            long rowIndicesCount = 0;
+            long nonZeroElementsCount = 0;
+
+            if (_di._nums == 0) {
+                if (_w == null) {
```
Pscheidl: Good speedup, previously known issue.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -872,30 +626,75 @@ protected static SparseMatrixDimensions calculateCSRMatrixDimensions(Chunk[] chu
...
+                Chunk[] cs = new Chunk[_di._nums];
+                for (int c = 0; c < cs.length; c++) {
+                    cs[c] = _f.vec(_di._cats + c).chunkForChunkIdx(cidx);
+                }
+                Chunk ws = _w != null ? _w.chunkForChunkIdx(cidx) : null;
+                for (int r = 0; r < cs[0]._len; r++) {
+                    if (ws != null && ws.atd(r) == 0)
+                        continue;
+                    rowIndicesCount++;
+                    nonZeroElementsCount += _di._cats;
```
Pscheidl: It could also be possible to do `nonZeroElementsCount += rowIndicesCount * _di._cats;` at the end to avoid addition per each row. Tiny speedup.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -872,30 +626,75 @@ protected static SparseMatrixDimensions calculateCSRMatrixDimensions(Chunk[] chu
...
     }
 
-    protected static SparseMatrixDimensions calculateCSRMatrixDimensions(Frame f, int[] chunks, Vec.Reader[] vecs, Vec.Reader w, DataInfo di) {
-        long nonZeroElementsCount = 0;
-        long rowIndicesCount = 0;
+    static SparseMatrixDimensions calculateCSRMatrixDimensions(Frame f, int[] chunkIds, Vec w, DataInfo di) {
+        CalculateCSRMatrixDimensionsMrFun fun = new CalculateCSRMatrixDimensionsMrFun(f, di, w, chunkIds);
+        H2O.submitTask(new LocalMR(fun, chunkIds.length)).join();
+        
+        return new SparseMatrixDimensions(ArrayUtils.sum(fun._nonZeroElementsCounts), ArrayUtils.sum(fun._rowIndicesCounts) + 1);
```
Pscheidl: The reason to do the sum here and not in the reduce phase of the `CalculateCSRMatrixDimensionsMrFun` is speed, right ? `ArrayUtils.sum` is just one method invocation, but `reduce` is many -> creating many contexts and destroying them quickly. For high number of chunks, this could get faster.

Maybe I'm getting into premature optimization, but for higher number of chunks, the number of additions done in a single thread might be as big as `Integer.MAX_VALUE - k`, where k is a small constant (typically `8`) - basically the MAX_ARRAY_SIZE allowed by the platform. Still, this operation is trivial `O(n)`, with parallelization we could achieve `O(n/t + t) + k`, where `t` is the number of threads and `k` is a simplification constant representing overhead used for thread management. This divide & conquer approach could theoretically bring us better benchmark results on large data sets.

michalkurka: I will explain the (2) first because it will make (1) easier to justify.

ad 2) This particular function is called during model training. Every model training starts with making sure the chunk layout of the frame is not hurtful to per-chunk parallelization, this basically means there are enough chunks for available number of CPU cores and the chunks are not too small. Ideal layout will have something like 4x #cores chunks per Vec. This means number of local chunks per Vec is low in the optimal case (hundreds) but there is enough work for each core to do something for most of the time of the execution of the "local MRTask". The amount of work is also comparable - the ideal distribution will have the same number of rows per chunk. This makes per-chunk parallelization work well in H2O because all CPU cores all working for most of the time. If the per-chunk map() is optimal then the whole utilization of the CPU is also optimal.

The length of array `fun._nonZeroElementsCounts` is the same as number of local chunks. Because this is executed in model training we know we can rely on almost optimal chunk layout and we are dealing with just hundreds of numbers to calculate the sum. Calculating the sum is thus trivial in terms of execution time compared to what is done in the LocalMR task.

ad 1) I could use reduce() but I chose not to because of 2 reasons:

a) it is easier to see in debugger what is going on and because of (1) we don't have any performance benefit doing it either way.

b) the next step of this PR is actually to *populate the matrix itself in parallel*. This could be done if keep the pointers (the 2 arrays) and then use them as indices to the emulated infinite array (the 2D structure). This is a similar approach to how we build the BigDenseMatrix, only difference is that we need to calculate the pointers in a different way. That is why I am keeping the pointers unaggregated - I can use them later for to accomplish this goal. And it doesn't hurt performance because the sizes of the metadata arrays are too small to matter.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
michalkurka: Thanks, I will change it.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
Pscheidl: Great, makes sense. Thank you !

Pscheidl:
As I was walking through the code, I found a place for a tiny speedup (I guess we want to squeeze out everything).

```patch
Index: h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
IDEA additional info:
<+>UTF-8
--- h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java	(revision 2e7df33a1cecc30c462e8b6bec546352ac99ec90)
+++ h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java	(date 1558251154000)
@@ -750,17 +750,20 @@
     }
 
     public static FeatureProperties assembleFeatureNames(final DataInfo di) {
-        String[] coefnames = di.coefNames();
+        final String[] coefnames = di.coefNames();
         assert (coefnames.length == di.fullN());
         int numCatCols = di._catOffsets[di._catOffsets.length - 1];
 
-        String[] featureNames = new String[di.fullN()];
+        final String[] featureNames = new String[di.fullN()];
         boolean[] oneHotEncoded = new boolean[di.fullN()];
-        for (int i = 0; i < di.fullN(); ++i) {
+        
+        for (int i = 0; i < numCatCols; i++) {
+            oneHotEncoded[i] = true;
             featureNames[i] = coefnames[i];
-            if (i < numCatCols) {
-                oneHotEncoded[i] = true;
-            }
+        }
+
+        for (int i = numCatCols; i < di.fullN(); ++i) {
+            featureNames[i] = coefnames[i];
         }
         return new FeatureProperties(featureNames, oneHotEncoded);
     }
```

Pscheidl:
The class `FeatureProperties` may be marked as private

```patch
Index: h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
IDEA additional info:
<+>UTF-8
--- h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java	(revision 2e7df33a1cecc30c462e8b6bec546352ac99ec90)
+++ h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java	(date 1558251542000)
@@ -750,7 +750,7 @@
     }
 
     public static FeatureProperties assembleFeatureNames(final DataInfo di) {
-        String[] coefnames = di.coefNames();
+        final String[] coefnames = di.coefNames();
         assert (coefnames.length == di.fullN());
         int numCatCols = di._catOffsets[di._catOffsets.length - 1];
 
@@ -765,7 +765,7 @@
         return new FeatureProperties(featureNames, oneHotEncoded);
     }
 
-    static class FeatureProperties {
+    private static class FeatureProperties {
         public String[] _names;
         public boolean[] _oneHotEncoded;
 ```

**Comment by Pscheidl** [comment]
Visibility of calculateCsrMatrixDimension may be protected, as its per-chunk counterpart:

```patch
Index: h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
IDEA additional info:
<+>UTF-8
--- h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java	(revision 2e7df33a1cecc30c462e8b6bec546352ac99ec90)
+++ h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java	(date 1558251739000)
@@ -626,7 +626,7 @@
         return new SparseMatrixDimensions(nonZeroElementsCount, ++rowIndicesCount);
     }
 
-    static SparseMatrixDimensions calculateCSRMatrixDimensions(Frame f, int[] chunkIds, Vec w, DataInfo di) {
+    protected static SparseMatrixDimensions calculateCSRMatrixDimensions(Frame f, int[] chunkIds, Vec w, DataInfo di) {
         CalculateCSRMatrixDimensionsMrFun fun = new CalculateCSRMatrixDimensionsMrFun(f, di, w, chunkIds);
         H2O.submitTask(new LocalMR(fun, chunkIds.length)).join();
         
```

Pscheidl:
Class `SparseItem` is obsolete

```patch
Index: h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
IDEA additional info:
<+>UTF-8
--- h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java	(revision 2e7df33a1cecc30c462e8b6bec546352ac99ec90)
+++ h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java	(date 1558251896000)
@@ -545,12 +545,7 @@
         }
         return actualRows;
     }
-
-    static class SparseItem {
-        int pos;
-        double val;
-    }
-
+    
     /**
      * Creates a {@link SparseMatrix} object with pre-instantiated backing arrays for row-oriented compression schema (CSR).
      * All backing arrays are allocated using MemoryManager.
```

Pscheidl:
Maybe we can avoid checking target 2D array boundaries with non-zero value observed in the row. The resulting loop is not really pretty and it requires and outer loop to return to, but this way, we may avoid the boundaries checks.

Usually, the target matrix row boundary is `Integer.MAX_VALUE - 8`, or in other words the maximum array size in Java. Or in yet another words, the length of target 2D matrix row. Typically, our row is smaller. For bigger matrices (big is meant column-wise, the number of columns might be really high even for smaller datasets due to one-hot encoding, typical example is `airlines`), a limit is calculated. Then, the for loop does no boundary checking. 

This approach only has one flaw - as it is approaching the end of target matrix row, the loop size shrinks.

I'm aware this section will be reworked and the matrix is going to be **populated in parallel**. That is the reason the patch below only demonstrates this approach on one loop - it is a proposal to be optionally used, if necessary. It basically shows how **current code** could be improved. And it might also be useful in future, after the adjustmens are in place.

```patch
Index: h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
IDEA additional info:
<+>UTF-8
--- h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java	(revision 2e7df33a1cecc30c462e8b6bec546352ac99ec90)
+++ h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java	(date 1558255718000)
@@ -514,24 +514,37 @@
                 }
                 nonZeroCount++;
             }
-            for (int j = 0; j < di._nums; ++j) {
-                if(currentCol == SPARSE_MATRIX_DIM){
+            
+            int looped = 0;
+            outer:
+            while(true) {
+                if (currentCol == SPARSE_MATRIX_DIM) {
                     currentCol = 0;
                     rowPointer++;
                 }
-                float val = (float) chunks[di._cats + j].atd(i);
-                if (val != 0) {
-                    data[rowPointer][currentCol] = val;
-                    colIndex[rowPointer][currentCol++] = di._catOffsets[di._catOffsets.length - 1] + j;
-                    if(!foundNonZero){
-                        foundNonZero = true;
-                        for (int k = lastNonZeroRow; k < actualRows; k++) {
-                            rowHeaders[rowHeaderRowPointer][rowHeaderColPointer++] = nonZeroCount;
-                        }
-                        lastNonZeroRow = actualRows;
-                    }
-                    nonZeroCount++;
+                
+                final int targetRowLim = SPARSE_MATRIX_DIM - currentCol;
+                final int increment = Math.min(targetRowLim, di._nums - looped);
+                int loopLim = looped + increment;
+                
+                for (int j = looped; j < loopLim; ++j) {
+                    float val = (float) chunks[di._cats + j].atd(i);
+                    if (val != 0) {
+                        data[rowPointer][currentCol] = val;
+                        colIndex[rowPointer][currentCol++] = di._catOffsets[di._catOffsets.length - 1] + j;
+                        if (!foundNonZero) {
+                            foundNonZero = true;
+                            for (int k = lastNonZeroRow; k < actualRows; k++) {
+                                rowHeaders[rowHeaderRowPointer][rowHeaderColPointer++] = nonZeroCount;
+                            }
+                            lastNonZeroRow = actualRows;
+                        }
+                        nonZeroCount++;
+                    }
                 }
+                looped += increment;
+                if (looped < di._nums) continue outer;
+                break;
             }
             
             rwRow = setResponseAndWeight(chunks, respIdx, weight, resp, weights, rwRow, i);
```

PR #3510, opened 2019-05-20
[PUBDEV-5937] make java predictor default for XGBoost

honzasterba:
- add new property xgboost.predict.native.enable
- move logic for determining predictor impl into separate file
- cleanup a little bit of dead code from xgboostutils

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
@@ -4,17 +4,19 @@ import biz.k11i.xgboost.Predictor;
 import biz.k11i.xgboost.gbm.GBTree;
 import biz.k11i.xgboost.gbm.GradBooster;
 import biz.k11i.xgboost.tree.RegTree;
-import biz.k11i.xgboost.tree.RegTreeImpl;
 import biz.k11i.xgboost.tree.RegTreeNode;
-import biz.k11i.xgboost.tree.TreeSHAPHelper;
 import biz.k11i.xgboost.util.FVec;
 import hex.*;
 import hex.genmodel.GenModel;
-import hex.genmodel.algos.tree.*;
+import hex.genmodel.algos.tree.SharedTreeGraph;
+import hex.genmodel.algos.tree.SharedTreeGraphConverter;
+import hex.genmodel.algos.tree.SharedTreeNode;
+import hex.genmodel.algos.tree.SharedTreeSubgraph;
 import hex.genmodel.algos.xgboost.XGBoostJavaMojoModel;
 import hex.genmodel.algos.xgboost.XGBoostMojoModel;
 import hex.genmodel.algos.xgboost.XGBoostNativeMojoModel;
 import hex.genmodel.utils.DistributionFamily;
+import hex.tree.xgboost.util.PredictConfiguration;
 import ml.dmlc.xgboost4j.java.*;
 import water.*;
 import water.fvec.Chunk;
@@ -23,14 +25,18 @@ import water.fvec.NewChunk;
 import water.fvec.Vec;
 import water.util.ArrayUtils;
 import water.util.Log;
-import hex.ModelMetrics;
 
 import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
-import java.util.*;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 
-import static hex.tree.xgboost.XGBoost.makeDataInfo;
 import static hex.genmodel.algos.xgboost.XGBoostMojoModel.ObjectiveType;
+import static hex.tree.xgboost.XGBoost.makeDataInfo;
+import static water.H2O.OptArgs.SYSTEM_PROP_PREFIX;
 
 public class XGBoostModel extends Model<XGBoostModel, XGBoostModel.XGBoostParameters, XGBoostOutput> 
         implements SharedTreeGraphConverter, Model.Contributions {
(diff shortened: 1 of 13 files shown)


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -192,6 +198,30 @@ public XGBoostModel(Key<XGBoostModel> selfKey, XGBoostParameters parms, XGBoostO
...
+      FileOutputStream os = new FileOutputStream(fmFile);
+      os.write(this.model_info._featureMap.getBytes());
+      os.close();
+      String fmFilePath = fmFile.getAbsolutePath();
+      String[] d = b.getModelDump(fmFilePath, true, format);
+      for (String l : d) {
+        System.out.println(l);
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
```
Pscheidl: I'd `Log.err` it.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -295,12 +286,12 @@ public static DMatrix convertChunksToDMatrix(DataInfo di,
                 // truly sparse matrix - no categoricals
                 // collect all nonzeros column by column (in parallel), then stitch together into final data structures
                 if (csc) {
-                    trainMat = csc(chunks, weight, nRows, di, resp, weights);
+                    trainMat = csc(chunks, -1, nRows, di, resp, weights);
```
Pscheidl: This method is currently preds-only, but I would avoid these changes. Is narrowing this method's scope what we really want ? If yes, when does the method have correct name ? Leaving the method with more general scope is one way to go.

Otherwise, we can simply discard `float[] weights = null;` on line 277 and pass null explicitly on each line, as this directly shows that weights are always null.

honzasterba: this change does nothing to the method scope, given functionality was never used
cleaned up even furhter

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
honzasterba: good point, updated

michalkurka:
@honzasterba thanks - fiy, I am making changes in XGBoostUtils here: https://github.com/h2oai/h2o-3/pull/3508

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostTest.java
```diff
@@ -61,8 +62,8 @@
   
   @Before
   public void setupMojoJavaScoring() {
-    System.setProperty("sys.ai.h2o.xgboost.scoring.java.enable", confMojoJavaScoring); // mojo scoring
-    System.setProperty("sys.ai.h2o.xgboost.predict.java.enable", confJavaPredict); // in-h2o predict
+    System.setProperty(XGBoostMojoReader.SCORE_JAVA_PROP, confMojoJavaScoring); // mojo scoring
```
michalkurka: I usually avoid doing this, it is better to have the constants hard-coded in tests. An unintended edit in the constant will break functionality for users but tests will stay green.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
Pscheidl: By scope, I meant how general the algorithm represented by the method is.

I am not a code junkie. I agree this functionality was unused and can be added on demand in future easily. No big deal. The decision is yours. I'm ok with it as long as it's correct, and it is.

michalkurka:
This is a good change, thanks! We need to check the impact on the benchmarks as well..

michalkurka:
@honzasterba we should probably also remove the warning in PredictorFactory - we will use UnsafeRegTrees on Intel architectures by default. There shouldn't be a warning in the default path.

honzasterba:
@michalkurka bechmarks show no change in performace, read to merge

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -165,32 +165,22 @@ private static int setResponseAndWeight(Vec.Reader w, float[] resp, float[] weig
     /**
      * convert a set of H2O chunks (representing a part of a vector) to a sparse DMatrix
      * @param response name of the response column
-     * @param weight name of the weight column
-     * @param fold name of the fold assignment column
      * @return DMatrix
      * @throws XGBoostError
      */
     public static DMatrix convertChunksToDMatrix(DataInfo di,
```
michalkurka: OMG, does this mean we were not generating predictions for 0 weighted rows before?

h2o-extensions/xgboost/src/main/java/ml/dmlc/xgboost4j/java/PredictorFactory.java
```diff
@@ -5,23 +5,26 @@
...
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteOrder;
 
 public class PredictorFactory {
-
+  
+  static {
+    XGBoostJavaObjFunRegistration.register();
```
michalkurka: Perfect, thanks!

michalkurka:
@honzasterba do you have a comparison before-after for the scoring?

otherwise looks good to me - but we should have the comparison somewhere available

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
honzasterba: since the argument was never used so no

honzasterba:
comparison:

Screenshot 2019-05-22 at 19 55 03

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
michalkurka: ok, phew

PR #3511, opened 2019-05-20
PUBDEV-6488  h2o.getModelTree failing

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6488

There are three commits in this PR, each representing one improvement.

- https://github.com/h2oai/h2o-3/pull/3511/commits/950494734d72c0de4d1ccbd51b21530153817119 Fixes prediction printing in R (Python is OK)
- https://github.com/h2oai/h2o-3/pull/3511/commits/927aa8ce28ae536d5079a1d8342b5abedcf2f7cf Randomized tests for Tree API
- https://github.com/h2oai/h2o-3/pull/3511/commits/a2aa61a9efa3aee4df7b1c80ea6e44e839abdaf0 Protection against NA only arrays being evaluated as logical in R

The last commit unfortunately has no direct test, as we were unable to reproduce. It was the original issue, @meganjkurka managed to force XGBoost to produce a tree with split features all being NA. But we don't have the code to reproduce it. In future, we're shielded against that.

### DIFF
--- a/h2o-r/tests/testdir_jira/runit_pubdev_6488.R
+++ b/h2o-r/tests/testdir_jira/runit_pubdev_6488.R
@@ -1,0 +1,22 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../scripts/h2o-r-test-setup.R")
+
+
+
+test.tree.fetch <- function() {
+  
+  data <- as.h2o(data.frame(
+    x1 = c(1, NA, 2),
+    x2 = c(2, NA, 1),
+    y = c(1, 2,3)
+  ))
+  
+  # A simple model to test prediction the correctly printed by delegating to print.H2ONode.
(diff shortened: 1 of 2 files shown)


h2o-r/tests/testdir_jira/runit_pubdev_6488.R
```diff
@@ -0,0 +1,137 @@
...
+  response <- sample(0:1,nrows,replace = TRUE)
+  domain <- unique(response)
+  data <- data.frame(
+    x1 = rnorm(nrows),
+    x2 = rbinom(nrows, 1, 0.5),
+    x3 = rbinom(nrows, 10, 0.5),
+    response = response # Binomial response
+  )
+  
+  data <- as.h2o(data)
```
maurever: Do we expect for some random data set it will fail? If yes, I would add seed to be able to reproduce a failure.

Pscheidl: Random printed seed could be a way.

PR #3512, opened 2019-05-20
[PUBDEV-6108] improve python coxph model summary

honzasterba:


h2o-py/h2o/model/coxph.py
```diff
@@ -12,3 +12,13 @@ def formula(self):
     def coefficients_table(self):
         """Coefficients table."""
         return self._model_json["output"]["coefficients_table"]
+
+    def summary(self):
```
maurever: Is it ok, it overloads the `summary` method from the `ModelBase`? 
If there is nothing in `output["model_summary"]`, it is ok. :+1:

PR #3515, opened 2019-05-20
PUBDEV-6454: Fixed Tweedie Regression and Standard Error Calculation

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6457?filter=-1

1. Fixed Hessian/gradient calculation for Tweedie family and added Junit tests to verify the hessian/gradient fix.
2. Add Runit to run Tweedie regression on prostrate data.  Both H2O and R models are built.  Their coefficients are compared and found to be within tolerance.
3. However, the standard error of coefficients only agree between the two model if variance_power=1 and link_power = 0.  They do not agree with the other cases.

I used the standard error calculation using Pearson chi-squared statistic to estimate the dispersion factor.  You can find the standard error calculation on page 28 of this doc: http://statmath.wu.ac.at/courses/heather_turner/glmCourse_001.pdf.  Estimation of the dispersion factor can be found on page 3 of this doc: http://people.stat.sfu.ca/~raltman/stat402/402L25.pdf

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -1083,16 +1083,17 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       for (Vec v : newVecs) v.remove();
       _state.updateState(beta,objold);
     }
+
     private void fitModel() {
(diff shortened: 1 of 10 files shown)


wendycwong:
changed branch name to make sure I can update it.

PR #3516, opened 2019-05-20
PUBDEV-6457: Fixed Tweedie hessian/gradient and standard error calculation

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6457?filter=-1 and https://0xdata.atlassian.net/browse/PUBDEV-6472?filter=-1

1. Fixed Hessian/gradient calculation for Tweedie family and added Junit tests to verify the hessian/gradient fix.
2. Add Runit to run Tweedie regression on prostrate data. Both H2O and R models are built. Their coefficients are compared and found to be within tolerance.
However, the standard error of coefficients only agree between the two model if variance_power=1 and link_power = 0. They do not agree with the other cases.
3. I used the standard error calculation using Pearson chi-squared statistic to estimate the dispersion factor. You can find the standard error calculation on page 28 of this doc: http://statmath.wu.ac.at/courses/heather_turner/glmCourse_001.pdf. Estimation of the dispersion factor can be found on page 3 of this doc: http://people.stat.sfu.ca/~raltman/stat402/402L25.pdf
4. A junit test is added to verify the dispersion parameter estimation is correct.

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -1119,7 +1119,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
         if(_parms._family != Family.binomial && _parms._family != Family.poisson) {
           seEst = true;
           ComputeSETsk ct = new ComputeSETsk(null, _state.activeData(), _job._key, beta, _parms).doAll(_state.activeData()._adaptedFrame);
-          se = ct._sumsqe / (_nobs - 1 - _state.activeData().fullN());
+          se = ct._sumsqe / (_nobs - 1 - _state.activeData().fullN());  // this is the dispersion parameter estimate
         }
         double [] zvalues = MemoryManager.malloc8d(_state.activeData().fullN()+1);
         Cholesky chol = _chol;
@@ -1153,10 +1153,14 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       Frame train = DKV.<Frame>getGet(_parms._train);
       _model.score(train).delete();
       ModelMetrics mtrain = ModelMetrics.getFromDKV(_model, train); // updated by model.scoreAndUpdateModel
-      _model._output._training_metrics = mtrain;
       long t2 = System.currentTimeMillis();
+      if (!(mtrain==null)) {
+        _model._output._training_metrics = mtrain;
+        Log.info(LogMsg(mtrain.toString()));
+      } else {
+        Log.info(LogMsg("ModelMetrics mtrain is null"));
+      }
       Log.info(LogMsg("Training metrics computed in " + (t2-t1) + "ms"));
-      Log.info(LogMsg(mtrain.toString()));
       if(_valid != null) {
         Frame valid = DKV.<Frame>getGet(_parms._valid);
         _model.score(valid).delete();
--- a/h2o-algos/src/main/java/hex/glm/GLMModel.java
+++ b/h2o-algos/src/main/java/hex/glm/GLMModel.java
@@ -28,6 +28,8 @@ import java.util.NoSuchElementException;
  */
 public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLMOutput> {
 
+  final static public double _EPS = 1e-6;
+  final static public double _OneOEPS = 1e6;
   public GLMModel(Key selfKey, GLMParameters parms, GLM job, double [] ymu, double ySigma, double lambda_max, long nobs) {
     super(selfKey, parms, job == null?new GLMOutput():new GLMOutput(job));
     // modelKey, parms, null, Double.NaN, Double.NaN, Double.NaN, -1
@@ -138,10 +140,8 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
   public double deviance(double w, double y, double f) {
     if (w == 0) {
       return 0;
-    } else if (w == 1) {
-      return _parms.deviance(y, f);
     } else {
-      return Double.NaN; //TODO: add deviance(w, y, f)
+      return w*_parms.deviance(y,f);
     }
   }
 
@@ -450,7 +450,7 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
         case logit:
 //        case multinomial:
           double div = (x * (1 - x));
-          if(div < 1e-6) return 1e6; // avoid numerical instability
+          if(div < _EPS) return _OneOEPS; // avoid numerical instability
           return 1.0 / div;
         case identity:
           return 1;
@@ -461,7 +461,7 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
         case ologlog:
           double oneMx = 1.0-x;
           double divsor = -1.0*oneMx*Math.log(oneMx);
-          return (divsor<1e-6)?1e6:(1.0/divsor);
+          return (divsor<_EPS)?_OneOEPS:(1.0/divsor);
         case tweedie:
 //          double res = _tweedie_link_power == 0
 //            ?Math.max(2e-16,Math.exp(x))
@@ -539,8 +539,16 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
     final Link _link;
     final double _var_power;
     final double _link_power;
+    final double _oneOoneMinusVarPower;
+    final double _oneOtwoMinusVarPower;
+    final double _oneMinusVarPower;
+    final double _twoMinusVarPower;
+    final double _oneOLinkPower;
+    final double _oneOLinkPowerSquare;
     double _theta;  // used by negative binomial, 0 < _theta <= 1
     double _invTheta;
+    double _oneOeta;
+    double _oneOetaSquare;
 
     final NormalDistribution _dprobit = new NormalDistribution(0,1);  // get the normal distribution
     
@@ -552,6 +560,12 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
       _link = link;
       _var_power = var_power;
       _link_power = link_power;
+      _oneMinusVarPower = 1-_var_power;
+      _twoMinusVarPower = 2-_var_power;
+      _oneOoneMinusVarPower = _var_power==1?1:1.0/(1-_var_power);
+      _oneOtwoMinusVarPower = _var_power==2?1:1.0/(2-_var_power);
+      _oneOLinkPower = 1.0/_link_power;
+      _oneOLinkPowerSquare = _oneOLinkPower*_oneOLinkPower;
       _theta = theta;
       _invTheta = 1/theta;
     }
@@ -596,9 +610,8 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
         case inverse:
           double xx = (x < 0) ? Math.min(-1e-5, x) : Math.max(1e-5, x);
           return -1 / (xx * xx);
-//        case tweedie:
-//          double vp = (1. - _tweedie_link_power) / _tweedie_link_power;
-//          return (1/ _tweedie_link_power) * Math.pow(x, vp);
+        case tweedie:
+          return _link_power==0?Math.max(x, Double.MIN_NORMAL):x*_oneOLinkPower*_oneOeta;
         default:
           throw new RuntimeException("unexpected link function id  " + this);
       }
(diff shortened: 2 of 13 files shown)


h2o-algos/src/main/java/hex/glm/GLMTask.java
```diff
@@ -2036,12 +2043,17 @@ protected void processRow(Row r) {
         w = _glmw.w;
       }
       double eta = r.innerProduct(_betaNew) + _sparseOffsetNew;
-//      double mu = _parms.linkInv(eta);
-      _sumsqe += w*(eta - z)*(eta - z);
+      double mu = _glmf.linkInv(eta);
+      _sumsqe += _glmf._family.equals(Family.tweedie)?
+              r.weight*((r.response(0)-mu)*(r.response(0)-mu))/Math.pow(mu, _glmf._var_power):
+              w*(eta - z)*(eta - z);
+
```
wendycwong: implementing the dispersion factor calculation for tweedie according to paper http://people.stat.sfu.ca/~raltman/stat402/402L25.pdf

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -28,6 +28,8 @@
  */
 public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLMOutput> {
 
+  final static public double _EPS = 1e-6;
+  final static public double _OneOEPS = 1e6;
   public GLMModel(Key selfKey, GLMParameters parms, GLM job, double [] ymu, double ySigma, double lambda_max, long nobs) {
```
wendycwong: I am sick and tired of setting a value to 1e-6 if it is 0.  Hence, I use a constant EPS to replace the 1e-6.  Doing it this way, we can change it to bigger or smaller at one place.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -552,6 +562,12 @@ public GLMWeightsFun(Family fam, Link link, double var_power, double link_power,
       _link = link;
       _var_power = var_power;
       _link_power = link_power;
+      _oneMinusVarPower = 1-_var_power;
+      _twoMinusVarPower = 2-_var_power;
+      _oneOoneMinusVarPower = _var_power==1?1:1.0/(1-_var_power);
+      _oneOtwoMinusVarPower = _var_power==2?1:1.0/(2-_var_power);
+      _oneOLinkPower = 1.0/_link_power;
+      _oneOLinkPowerSquare = _oneOLinkPower*_oneOLinkPower;
       _theta = theta;
```
wendycwong: This is purely for speed up as we need to do this for every row.  I just calculate them once and use them for every row afterwards.  Probably just some small speedup unless the dataset is huge.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -596,9 +612,8 @@ public final double linkInvDeriv(double x) {
         case inverse:
           double xx = (x < 0) ? Math.min(-1e-5, x) : Math.max(1e-5, x);
           return -1 / (xx * xx);
-//        case tweedie:
-//          double vp = (1. - _tweedie_link_power) / _tweedie_link_power;
-//          return (1/ _tweedie_link_power) * Math.pow(x, vp);
+        case tweedie:
+          return _link_power==0?Math.max(x, Double.MIN_NORMAL):x*_oneOLinkPower*_oneOeta;
```
wendycwong: The link function depends on the link power q  as follows:

image

I am just implementing this.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -611,6 +626,8 @@ public final double linkInvDeriv2(double x) {
           return 0;
         case log:
           return Math.max(x, Double.MIN_NORMAL);
+        case tweedie:
+          return _link_power==0?Math.max(x, Double.MIN_NORMAL):x*_oneOLinkPower*(_oneOLinkPower-1)*_oneOetaSquare;
         default:
```
wendycwong: The link function depends on the link power q as follows:

image

I am just implementing this.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -713,13 +730,16 @@ public final double deviance(double yr, double ym){
...
+          if (_var_power==1) {
+            val = yr*Math.log(y1/ym)-(yr-ym);
+          } else if (_var_power==2) {
+            val = yr*(1/ym-1/y1)-Math.log(y1/ym);
+          } else {
+            val = yr*_oneOoneMinusVarPower*(Math.pow(yr,_oneMinusVarPower)-Math.pow(ym, _oneMinusVarPower))-
+                    (Math.pow(yr,_twoMinusVarPower)-Math.pow(ym, _twoMinusVarPower))*_oneOtwoMinusVarPower;
+          }
+          return 2 * val;
         default:
```
wendycwong: I am calculating the deviance according to this formula:

image

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -785,8 +805,11 @@ public final double likelihood(double yr, double ym) {
...
           if (yr == 0) return -2;
           return -2 * (Math.log(yr / ym) - (yr - ym) / ym);
-        case tweedie:
-          return deviance(yr, ym); //fixme: not really correct, not sure what the likelihood is right now
+        case tweedie: // we ignore the a(y,phi,p) term in the likelihood calculation here since we are not optimizing over them
+          double temp = 0;
+          temp += _var_power==2?Math.log(ym):Math.pow(ym,_twoMinusVarPower)*_oneOtwoMinusVarPower;
+          temp -= yr*(_var_power==1?Math.log(ym):Math.pow(ym,_oneMinusVarPower)*_oneOoneMinusVarPower);
+          return temp; // ignored the a(y,phi,p) term as it is a constant for us
         default:
```
wendycwong: I am calculating the deviance according to this:

image

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -814,6 +837,22 @@ public GLMWeights computeWeights(double y, double eta, double off, double w, GLM
...
+        double xmuPowMP = Math.pow(x.mu, -_var_power);
+        x.w = _var_power==1?(y*oneOxmuSquare*diffOneSquare-(y*oneOxmu-1)*linkInvDeriv2(x.mu)):(_var_power==2?
+                ((oneOxmu-y*xmuPowMP)*linkInvDeriv2(x.mu)+
+                        (y*2*Math.pow(x.mu, -3)-oneOxmuSquare)*diffOneSquare)
+                :(_var_power*y*Math.pow(x.mu, -_var_power-1)+_oneMinusVarPower*xmuPowMP)*diffOneSquare-
+                (y*xmuPowMP-Math.pow(x.mu, _oneMinusVarPower))*linkInvDeriv2(x.mu));
+        x.z = etaOff+(_var_power==1?(y*oneOxmu-1):
+                (_var_power==2?(y*Math.pow(x.mu, -_var_power)-oneOxmu):
+                        (y*Math.pow(x.mu, -_var_power)-Math.pow(x.mu, _oneMinusVarPower))))*linkInvDeriv(x.mu)/x.w;
+        x.w *= w;
```
wendycwong: I am taking the second derivative of the negative likelihood function here.

h2o-algos/src/main/java/hex/glm/GLMTask.java
```diff
@@ -690,8 +691,16 @@ public GLMGenericGradientTask(Key jobKey, DataInfo dinfo, GLMParameters parms, d
...
+          if (_glmf._family.equals(Family.tweedie)) {
+            _glmf._oneOeta = 1.0/(es[i]==0?hex.glm.GLMModel._EPS:es[i]);
+            _glmf._oneOetaSquare = _glmf._oneOeta*_glmf._oneOeta;
+            es[i] = ws[i]*_glmf.linkInvDeriv(mu)*(_glmf._var_power==1?(ys[i]/mu-1):
+                    (_glmf._var_power==2?(ys[i]*Math.pow(mu, -_glmf._var_power)-1/mu):
+                            (ys[i]*Math.pow(mu, -_glmf._var_power)-Math.pow(mu, _glmf._oneMinusVarPower))));
+          } else {
+            es[i] = ws[i] * (mu - ys[i]) / (var * _glmf.linkDeriv(mu));
+          }
         }
```
wendycwong: I am taking the derivate of the likelihood and calculating the gradient here.

h2o-algos/src/main/java/hex/gram/Gram.java
```diff
@@ -176,7 +176,7 @@ private final void updateZ(final double [] gamma, final double [][] Z, int j){
    * When doing that, we need to replace dot(xi,xk) with dot(xi,zk) for all i.
    * There are two cases,
    *   1)  dot(xk,xk) -> dot(zk,zk)
-   *       dot(xk - sum(gamma_l*x_l,xk - sum(gamma_l*x_l)
+   *       dot(xk - sum(gamma_l*x_l),xk - sum(gamma_l*x_l))
    *       = dot(xk,xk) - 2* sum(gamma_l*dot(x_i,x_k) + sum(gamma_l*sum(gamma_k*dot(x_l,x_k)))
    *       (can be simplified using the fact that dot(zi,zj) == 0 for i != j
    *   2)  dot(xi,xk) -> dot(xi,zk) for i != k
```
wendycwong: Only fixing the comments here.

h2o-algos/src/main/java/hex/glm/GLMModel.java
Pscheidl: No need to re-calculate it per each row, when it can be done just once :)

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -151,6 +151,23 @@ public static void checkLeakedKeys() {
     _initial_keycnt = H2O.store_size();
   }
 
+  public static void checkArrays(double[] expected, double[] actual, double threshold) {
```
Pscheidl: Why not use `assertArrayEquals(double[], double[], double delta)` from JUnit ?

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -151,6 +151,23 @@ public static void checkLeakedKeys() {
...
+    for(int i = 0; i < actual.length; i++) {
+      if (!Double.isNaN(expected[i]) && !Double.isNaN(actual[i])) // only compare when both are not NaN
```
Pscheidl: We might use `Double.compare` right away.

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -151,6 +151,23 @@ public static void checkLeakedKeys() {
...
+        assertEquals(expected[i], actual[i], threshold*Math.max(Math.abs(expected[i]),Math.abs(actual[i])));
```
Pscheidl: I don't understand `threshold*Math.max(Math.abs(expected[i]),Math.abs(actual[i]))`

Do you intend to make the actual resulting threshold to be a fraction of the greater of the two compared numbers ? Couldn't this result in big inaccuracies ?

wendycwong: Pavel:  

Yes, you are correct in the absolute sense.  However, the difference caused by numerical instability scaled with the magnitude of the numbers.  In order to get rid of the effect of this magnitude scaling, I set the tolerance to scale with the number magnitude. 

However, I think I should use the Math.min instead of Math.max in this case.

h2o-core/src/test/java/water/TestUtil.java
wendycwong: Nope, I need to skip over NaNs and also, I need to change the threshold per element due to the fact that I want to scale the tolerance according to the element magnitudes.

h2o-core/src/test/java/water/TestUtil.java
wendycwong: I tried to use Double.compare but there is no threshold value I can set, am I calling it the right way?

Pscheidl: Of course you are, wrote this comment before I noticed the threshold and forgot to delete it, sorry.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1119,7 +1119,7 @@ else if(_parms._family == Family.gaussian && _parms._link == Link.identity)
         if(_parms._family != Family.binomial && _parms._family != Family.poisson) {
           seEst = true;
           ComputeSETsk ct = new ComputeSETsk(null, _state.activeData(), _job._key, beta, _parms).doAll(_state.activeData()._adaptedFrame);
-          se = ct._sumsqe / (_nobs - 1 - _state.activeData().fullN());
+          se = ct._sumsqe / (_nobs - 1 - _state.activeData().fullN());  // this is the dispersion parameter estimate
         }
         double [] zvalues = MemoryManager.malloc8d(_state.activeData().fullN()+1);
```
wendycwong: I used the dispersion parameter estimate like this from the doc:

image

Pscheidl:
@wendycwong You may get inspired by this PR a do something similar with link function in GLM.

PR #3518, opened 2019-05-21
PUBDEV-6512: Fix and reenable AutoML Java UT in multinode env

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6512

ensure that Java UTs take into consideration the fact that algos are enabled or not


deil87:
@seb-h2o Tests are green! LGTM

PR #3525, opened 2019-05-22
PUBDEV-6479: categorical encoding mismatch when building ensemble

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6479

When building a SE with base models having a specific categorical encoding, SE model is built correctly but SE fails on predict.

fix proposal:
- expose `categorical_encoding` for SE.
- let the client set the categorical encoding explicitly on SE if s·he did it on the base models.

e.g.:
```
estRF = H2ORF(keep_cross_validation_predictions = True, 
                           categorical_encoding="LabelEncoder")

ensemble = H2OStackedEnsembleEstimator(base_models=[estRF, estRF], 
                         categorical_encoding="LabelEncoder"))
```


seb-h2o:
@michalkurka : please check proposal, writing UT now

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -108,6 +108,24 @@ public void initMetalearnerParams(Metalearner.Algorithm algo) {
...
+            _output._names,
+            _output._domains,
+            _output._metalearner._parms,
+            expensive,
+            computeMetrics,
+            _output.interactionBuilder(),
+            getToEigenVec(),
+            _toDelete,
+            false
+    );
```
seb-h2o: I could probably use as well:
```
    return _output._metalearner.adaptTestForTrain(test, expensive, computeMetrics);
```
trying it and checking the possible impact of that change

seb-h2o: actually, no, we can't use this simpler form, as the metalearner is trained on a totally different dataset...
even wondering if passing `_output._metalearner._parms` is such a good idea:
- `metalearner._parms` should be used to deal with the `level_one_frame` only, not the training/predictions sets.

seb-h2o: cancelled this approach as it was mixing 2 set of params (SE params and metalearner params) that should stay unrelated most of the time, especially when they're use to manipulate the input data.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -209,7 +209,17 @@ public void setTrain(Key<Frame> train) {
       AUTO, Random, Modulo, Stratified
     }
     public enum CategoricalEncodingScheme {
-      AUTO(false), OneHotInternal(false), OneHotExplicit(false), Enum(false), Binary(false), Eigen(false), LabelEncoder(false), SortByResponse(true), EnumLimited(false);
```
michalkurka: Can you please confirm this was just reformatted or is there also another change?

michalkurka:
@seb-h2o I don't know if I understand the fix... conceptually SE model doesn't need to do any encoding because it never acts on the raw data (as of now) it should just let the downstream models adapt the frames in any way they need to independently and then apply the super learner.

h2o-core/src/main/java/hex/Model.java
seb-h2o: just reformatting

seb-h2o:
@michalkurka 
> SE model doesn't need to do any encoding because it never acts on the raw data
it's true during training, but not fully for scoring/predictions: there's one call to `adaptTestForTrain` that also applies for SE.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -371,13 +371,17 @@ void checkAndInheritModelProperties() {
         // !beenHere: this is the first base_model
         this.modelCategory = aModel._output.getModelCategory();
         this._dist = new Distribution(distributionFamily(aModel));
-        _output._domains = Arrays.copyOf(aModel._output._domains, aModel._output._domains.length);
-
         // TODO: set _parms._train to aModel._parms.train()
 
-        _output.setNames(aModel._output._names, aModel._output._column_types);
-        this.names = new NonBlockingHashSet<>();
-        this.names.addAll(Arrays.asList(aModel._output._names));
```
seb-h2o: it seems better to rely on `train` domains/names than on the output domains provided by the first base model.
If this one encoded the categoricals, then the `output.domains` keeps only the target, as there's no categoricals anymore...

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1157,10 +1174,13 @@ public double lift_top_group() {
...
-        parms._categorical_encoding == Parameters.CategoricalEncodingScheme.OneHotExplicit ||
-        parms._categorical_encoding == Parameters.CategoricalEncodingScheme.Eigen ||
-        parms._categorical_encoding == Parameters.CategoricalEncodingScheme.Binary;
+    // keep in sync with FrameUtils.categoricalEncoder: as soon as a categorical column has been encoded, we should check here.
+    final boolean checkCategoricals = parms._categorical_encoding == Parameters.CategoricalEncodingScheme.Binary
+            || parms._categorical_encoding ==Parameters.CategoricalEncodingScheme.LabelEncoder
+            || parms._categorical_encoding == Parameters.CategoricalEncodingScheme.Eigen
+//            || parms._categorical_encoding == Parameters.CategoricalEncodingScheme.EnumLimited
+            || parms._categorical_encoding == Parameters.CategoricalEncodingScheme.OneHotExplicit
+            ;
```
seb-h2o: including "LabelEncoder" here allows proper conversion below L1260

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1251,12 +1271,12 @@ else if (isWeights && computeMetrics) {
...
           }
         } else if(vec.isCategorical()) {
-          if (parms._categorical_encoding == Parameters.CategoricalEncodingScheme.LabelEncoder) {
+          try {
             Vec evec = vec.toNumericVec();
             toDelete.put(evec._key, "label encoded vec");
             vec = evec;
-          } else {
-            throw new IllegalArgumentException("Test/Validation dataset has categorical column '" + names[i] + "' which is real-valued in the training data");
+          } catch(Exception e) {
```
seb-h2o: suggesting to be more lenient here as the conversion above is actually half-backed imo.
It works only if the domains from the test frame are indexed exactly the same way as the training frame (which provides the domains through `_output.domains`).
Although, it can easily not be the case if test frame is a feature subset of training features for example.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1333,8 +1333,10 @@ else if (_parms._weights_column != null && _weights != null && !_weights.isBinar
       if (restructured)
         _train.restructure(_train.names(), vecs);
     }
-    assert (!expensive || _valid==null || Arrays.equals(_train._names, _valid._names) || _parms._categorical_encoding == Model.Parameters.CategoricalEncodingScheme.Binary);
-    if (_valid!=null && !Arrays.equals(_train._names, _valid._names) && _parms._categorical_encoding == Model.Parameters.CategoricalEncodingScheme.Binary) {
+    boolean names_may_differ = _parms._categorical_encoding == Model.Parameters.CategoricalEncodingScheme.Binary;
+    boolean names_differ = _valid !=null && !Arrays.equals(_train._names, _valid._names);
+    assert (!expensive || names_may_differ || !names_differ);
```
seb-h2o: this currently fails if I add EnumLimited to the `checkCategoricals` in adaptTestToTrain due to double encoding (enum limited encoding is not idempotent...)

h2o-core/src/main/java/hex/Model.java
michalkurka: Not sure if we are on the same page - can you explain a bit more?

IMHO `toNumericVec` on a categorical column will try to interpret the domain values as integer numbers according to these rules:

   * If the first value in the domain of the src Vec is a stringified ints,
   * then it will use those ints. Otherwise, it will use the raw enumeration level mapping.

This behavior honestly doesn't make much sense to me right now and can be outright dangerous as it all depends just on the first value of the domain.

We get a test frame with a Vec that is categorical instead of numerical. IMO this has nothing to do with categorical encoding. The Model output remembers the original domain of the training frame (before label encoding...) and this means the original training Vec simply was numerical not categorical.

Am I missing something?

michalkurka: Looks like LabelEncoder is broken: https://github.com/h2oai/h2o-3/pull/3527

seb-h2o: @michalkurka the test behaves as expected with my changes here.
It's not as much that `H2O forgets the original domain of LabelEncoded columns`: this is actually not the case, there are available in `_output._origDomains` whatever the encoding.
However, those origDomains are currently ignored in `adaptTestForTrain` for `LabelEncoder`, hence the reason I'm adding it to the `checkCategoricals` list, and also why I want to have it working with `EnumLimited` as well.

I have also good reasons to think that the absence of those 2 encodings there is an omission: those have been created (in FrameUtils) several months after `Binary`, `Eigen`... and given the complete decoupling between this logic and their definition in `FrameUtils`, it would not be surprising that author forgot to add them to the list here: I'll had a new method to the enum to avoid this in the future.

seb-h2o: regarding the call to `toNumericVec`: it is a fallback mechanism. It was called much more frequently before my changes (but only if encoding was specified to LabelEncoder).
Will check various use-cases to figure out if it shouldn't be simply removed.

michalkurka: Perfect, that makes sense - thank you.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1251,13 +1280,7 @@ else if (isWeights && computeMetrics) {
...
           }
         } else if(vec.isCategorical()) {
-          if (parms._categorical_encoding == Parameters.CategoricalEncodingScheme.LabelEncoder) {
-            Vec evec = vec.toNumericVec();
-            toDelete.put(evec._key, "label encoded vec");
-            vec = evec;
-          } else {
-            throw new IllegalArgumentException("Test/Validation dataset has categorical column '" + names[i] + "' which is real-valued in the training data");
-          }
+          throw new IllegalArgumentException("Test/Validation dataset has categorical column '" + names[i] + "' which is real-valued in the training data");
```
seb-h2o: @michalk: decided to completely remove this fallback as it should not be necessary anymore, all conversion should be handled explicitly with logic above.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1157,24 +1174,36 @@ public double lift_top_group() {
...
+    ).indexOf(parms._categorical_encoding) >= 0;
 
     // test frame matches the user-given frame (before categorical encoding, if applicable)
     if( checkCategoricals && origNames != null ) {
-      boolean match = Arrays.equals(origNames, test._names);
+      boolean match = Arrays.equals(origNames, test.names());
       if (!match) {
-        match = true;
         // In case the test set has extra columns not in the training set - check that all original pre-encoding columns are available in the test set
         // We could be lenient here and fill missing columns with NA, but then it gets difficult to decide whether this frame is pre/post encoding, if a certain fraction of columns mismatch...
```
seb-h2o: @michalkurka : do you understand this comment?
I don't understand why we can't be lenient indeed...  we should be able to make predictions on a test frame having only a subset of predictors columns (origNames), I don't see why we require them to be all present in the test frame.
any idea?

seb-h2o:
due to docker image issues, could check that Jenkins tests are passing by upgrading the image version: http://mr-0xc1:8080/job/h2o-3-pipeline/job/PR-3525/10/

however had to rollback this image version afterwards as it breaks some non-related coxph R tests.
=> merging

PR #3527, opened 2019-05-22
PUBDEV-6521: LabelEncoder doesn't work when test frame has a different domain

michalkurka:
Demonstrates that label-encoding can lead to model mispredictions on test data that have a different (=incomplete) domain.


michalkurka:
@seb-h2o already has a test for this

PR #3530, opened 2019-05-23
PUBDEV-6458 - PrintMOJO categorical levels printing with numerically represented split (PrintMojo part)

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6458

This is only a first part, touching only `PrintMojo` class. Tree API changes are going to be delivered in a separate PR.

## Demonstration
On a simple tree-based model created with `model <- h2o.isolationForest(x = c("C5"), training_frame = data, ntrees = 10)`
### Old behavior
old

### New behavior
new


h2o-genmodel/src/main/java/hex/genmodel/algos/gbm/GbmMojoModel.java
```diff
@@ -14,7 +14,7 @@
 /**
  * "Gradient Boosting Machine" MojoModel
  */
-public final class GbmMojoModel extends SharedTreeMojoModel implements SharedTreeGraphConverter, PredictContributionsFactory {
```
Pscheidl: In vain here, removed. SharedTreeModel already implements it.

h2o-genmodel/src/main/java/hex/genmodel/algos/drf/DrfMojoModel.java
```diff
@@ -9,7 +9,7 @@
 /**
  * "Distributed Random Forest" MojoModel
  */
-public final class DrfMojoModel extends SharedTreeMojoModel implements SharedTreeGraphConverter {
```
Pscheidl: In vain here, removed. SharedTreeModel already implements it.

michalkurka:
@Pscheidl this is great and it will certainly get the job done

I had a small idea :)

What if we modify this function instead:

    public SharedTreeGraph _computeGraph(int treeToPrint)

SharedTreeGraph would be either returned in the original form or based on a provided option with numerical splits on categorical columns remapped. Tree rendering logic would then stay the same - the advantage would be that we could use it also in TreeAPI the same way.

After this change, the TreeAPI will show trees that won't be the same as the visualized trees.

This is an idea for the future, doesn't have to be done in this PR - not at all, just think about if that direction would make sense to you.

Pscheidl:
@michalkurka What you propose is by far the best option. I was not sure we want to change the contract (we may keep both, but the default contract will change for all the clients of SharedTreeGraph).

If this is ok, I'll have a look if it is doable this way.

Pscheidl:
@maurever This is out of this PRs scope. However, Tree API is going to give you text representation of the nodes with detailed splits if required.

Pscheidl:
- I am working on a new PR, where SharedTreeModel is modified directly.
- This PR is wrong. It does not calculate missing categorical values and this would be really hard to do (parents need to be visited just like SharedTreeNode does it when calculating inherited levels.

PR #3531, opened 2019-05-23
PUBDEV-6508: Added validation of parameters for R Target Encoding API

deil87:
This PR is supposed to improve user experience of R Target Encoding API by validation of parameters and providing meaningful errors and warnings

### DIFF
--- a/h2o-r/h2o-package/R/targetencoder.R
+++ b/h2o-r/h2o-package/R/targetencoder.R
@@ -104,6 +104,36 @@ h2o.target_encode_transform <- function(frame, x, y, target_encode_map, holdout_
         stop(paste0("`holdout_type` must be one of the following: kfold, loo, none but got "
         , holdout_type))
     }
+    if (holdout_type %in% c("kfold")){
+        if(missing(fold_column)) {
+            stop("If `holdout_type` is set to `kfold` then it is required to provide `fold_column` parameter")
+        }
+        
+        #Checking that encoding map was created with specified fold column
+        frameKeys <- attr(target_encode_map, "frames")
+        emFrameKeys <- lapply(frameKeys, function(x) x$key$name )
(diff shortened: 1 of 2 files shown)


michalkurka:
@deil87 this is a good change and definitely a step in the right direction.

Do you think you can also add some backend validation? We shouldn't see NPEs coming from Java backend.

deil87:
@michalkurka There is some validation on the backend. Messages from those exceptions are being nicely propagated to client though they contain Java-style parameters' names.  I actually think that anyway validation should be done on the client side (R/Py) whenever it is possible so that we don't even bother backend. I can improve backend validation as well but maybe It would be better to do it in a separate PR, wdyt? Python's validation could be also improved to match R's one.

michalkurka:
sure, a new pr makes sense - I was referring to this particular exception:

```
ERROR: Unexpected HTTP Status code: 400 Bad Request (url = http://localhost:54321/99/Rapids)

java.lang.IllegalArgumentException
 [1] "java.lang.IllegalArgumentException: Name lookup of 'NULL' failed"                                                   
 [2] "    water.rapids.Env.lookup(Env.java:405)"                                                                          
 [3] "    water.rapids.ast.params.AstId.exec(AstId.java:33)"                                                              
 [4] "    water.rapids.ast.prims.mungers.AstTargetEncoderTransform.getInflectionPoint(AstTargetEncoderTransform.java:154)"
```

meganjkurka:
@deil87 I think the target encoding transform function should work with the default parameters.  Right now if I leave the optional arguments with their default settings I will get an error.

deil87:
Thank you @meganjkurka ! I know about this. Fixing this would require semantic change of the method and it should go to a separate PR. After our discussion in Slack I'm waiting for default parameters for hyper parameters. I pointed our that default values are different in DAI and in your previous R implementation so could you please investigate a little bit which version of default parameters is correct?

PR #3533, opened 2019-05-24
PUBDEV-6525: Set default values for R's API of Target Encoding

deil87:
For now I've set default values to the ones that DAI is using but I'm waiting for an update from @meganjkurka about the values.

@angela0xdata this PR if merged will require some changes in docs


zainhaq-h2o:
Default values added in pull request: https://github.com/h2oai/h2o-3/pull/3589

PR #3534, opened 2019-05-24
PUBDEV-6458 - Categorical splits represented as numerical (SharedTree solution)

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6458

As discovered in : https://github.com/h2oai/h2o-3/pull/3530

- It is best to fix it once and for all
- There is a strong suspicion our inherited categorical levels calculations are wrong.

### Categorical representation

Hide internal categorical splits represented in a numerical way.

tree_representation

### Internal representation
Print the numerical representation of the categorical split, but keep the categorical levels calculated for **each edge printed as well**, for better verifiability and readability.

tree_representation_internal

### DIFF
--- a/h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
+++ b/h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
@@ -1,0 +1,198 @@
+package hex.tree;
+
+import hex.genmodel.tools.PrintMojo;
+import hex.tree.isofor.IsolationForest;
+import hex.tree.isofor.IsolationForestModel;
+import org.apache.commons.io.FileUtils;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import water.Scope;
+import water.TestUtil;
+import water.fvec.Frame;
+
+import java.io.File;
+import java.io.IOException;
+import java.security.Permission;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.junit.Assert.*;
+
+public class PrintMojoTreeTest {
+
+  @Rule
+  public TemporaryFolder folder = new TemporaryFolder();
+
+  private SecurityManager originalSecurityManager;
+
+  @Before
+  public void setUp() throws Exception {
+    TestUtil.stall_till_cloudsize(1);
+    originalSecurityManager = System.getSecurityManager();
+    System.setSecurityManager(new PreventExitSecurityManager());
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    System.setSecurityManager(originalSecurityManager);
+  }
+
+  @Test
+  public void testMojoCategoricalPrint() throws IOException {
+    try {
+      Scope.enter();
+      Frame train = Scope.track(TestUtil.parse_test_file("smalldata/iris/iris.csv"));
+
+      IsolationForestModel.IsolationForestParameters p = new IsolationForestModel.IsolationForestParameters();
+      p._train = train._key;
+      p._ignored_columns = new String[]{"C1", "C2", "C3", "C4"};
+      p._seed = 0xFEED;
+      p._ntrees = 1;
+
+      IsolationForestModel model = new IsolationForest(p).trainModel().get();
+      final File modelFile = folder.newFile();
+      model.exportMojo(modelFile.getAbsolutePath(), true);
+
+      final File treeOutput = folder.newFile();
+      try {
+        PrintMojo.main(new String[]{"--input", modelFile.getAbsolutePath(), "--output", treeOutput.getAbsolutePath()});
+        fail("Expected PrintMojo to call System.exit()");
+      } catch (PreventedExitException e) {
+      }
+
+      final String treeDotz = FileUtils.readFileToString(treeOutput);
+      System.out.println(treeDotz);
+      assertFalse(treeDotz.isEmpty());
+
+
+      final Pattern labelPattern = Pattern.compile("label{1}=\\\"(.*?)\\\"");
+      final Pattern labelContentPattern = Pattern.compile(".*[<>=].*"); // Contains < or > or =
+      final Matcher matcher = labelPattern.matcher(treeDotz);
+
+      assertEquals(1, matcher.groupCount());
+
+      while (matcher.find()){
+        assertFalse(labelContentPattern.matcher(matcher.group(1)).matches());
+      }
+
+    } finally {
+      Scope.exit();
+    }
+  }
+
+  @Test
+  public void testMojoCategoricalPrint_limitedLevels() throws IOException {
+    try {
+      Scope.enter();
+      Frame train = Scope.track(TestUtil.parse_test_file("smalldata/iris/iris.csv"));
+
+      IsolationForestModel.IsolationForestParameters p = new IsolationForestModel.IsolationForestParameters();
+      p._train = train._key;
+      p._ignored_columns = new String[]{"C1", "C2", "C3", "C4"};
+      p._seed = 0xFEED;
+      p._ntrees = 1;
+
+      IsolationForestModel model = new IsolationForest(p).trainModel().get();
+      final File modelFile = folder.newFile();
+      model.exportMojo(modelFile.getAbsolutePath(), true);
+
+      final File treeOutput = folder.newFile();
+      try {
+        PrintMojo.main(new String[]{"--input", modelFile.getAbsolutePath(), "--output", treeOutput.getAbsolutePath(), "--levels", "1"});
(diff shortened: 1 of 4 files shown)


h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeMojoModel.java
```diff
@@ -337,9 +339,18 @@ private static void computeTreeGraph(SharedTreeSubgraph sg, SharedTreeNode node,
         if (!naVsRest) {
             // Extract value or group to split on
             if (equal == 0) {
+              float splitVal = ab.get4f();
+
+              if (domains[colId] != null) {
+                GenmodelBitSet bs = new GenmodelBitSet(0);
+                final ByteBuffer byteBuffer = ByteBuffer.allocate(4).order(ByteOrder.nativeOrder())
```
Pscheidl: This is just an early stage attempt, detailed description of how these bytes should look like is in `CompressedTree`.

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
```diff
@@ -0,0 +1,198 @@
...
+      try {
+        PrintMojo.main(new String[]{"--input", modelFile.getAbsolutePath(), "--output", treeOutput.getAbsolutePath(), "--levels", "1"});
+        fail("Expected PrintMojo to call System.exit()");
+      } catch (PreventedExitException e) {
+      }
+
+      final String treeDotz = FileUtils.readFileToString(treeOutput);
+      assertFalse(treeDotz.isEmpty());
+
+      assertTrue(treeDotz.contains("\"SG_0_Node_0\" -> \"SG_0_Node_1\" [fontsize=14, label=\"[NA]\\n2 levels\\n\"]\n" +
```
Pscheidl: Hmmm, doesn't this part rely too much on the output ? The implementation can change. However, this is the most precise control.

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
```diff
@@ -177,11 +177,15 @@ private BitSet calculateChildInclusiveLevels(boolean includeAllLevels,
...
     BitSet childInclusiveLevels = new BitSet();
 
+    int splitPoint = Float.isNaN(splitValue) ? -1 : (int) Math.ceil(splitValue);
+
     for (int i = 0; i < domainValues.length; i++) {
       // Calculate whether this level should flow into this child node.
       boolean includeThisLevel = false;
       {
-        if (discardAllLevels) {
+        if (splitPoint != -1) {
```
michalkurka: Can we just work with the original `splitValue` here?

I think it would be a bit clearer if we just compared the index to the original splitValue, wdyt?

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -68,6 +71,10 @@ private static void usage() {
     System.out.println("");
     System.out.println("    --fontsize | -f    Set font sizes of strings.");
     System.out.println("");
+    System.out.println("    --raw    Print raw graph representation to standard output.");
```
michalkurka: What is "raw"?

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -68,6 +71,10 @@ private static void usage() {
...
+    System.out.println("");
+    System.out.println("    --internal    Internal H2O representation of the graph (splits etc.) is used for generating the GRAPHVIZ format.");
```
michalkurka: What does adding `internal` mean to for the output?

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
Pscheidl: Raw is a mode that has been there for a long time, added by @tomkraljevic . I've only added it to the documentation. It basically prints the graph in a more human-readable format. Examle:

```
Graph

    ----- Tree 0 -----
    Nodes
        Node 0
            weight:      150.0
            depth:       0
            colId:       0
            colName:     C5
            leftward:    true
            naVsRest:    false
            splitVal:    1.5
            isBitset:    true
            predValue:   0.3649551
            squaredErr:  11.1307125
            leftChild:   Node 1
            rightChild:  Node 4
        Node 4
            weight:      50.0
            depth:       1
            colId:       0
            colName:     
            leftward:    false
            naVsRest:    false
            splitVal:    NaN
            isBitset:    false
            predValue:   0.0
            squaredErr:  4.1470647
            leftChild:   
            rightChild:  
        Node 1
            weight:      100.0
            depth:       1
            colId:       0
            colName:     C5
            leftward:    false
            naVsRest:    false
            splitVal:    NaN
            isBitset:    true
            predValue:   0.54743266
            squaredErr:  6.983648
            leftChild:   Node 5
            rightChild:  Node 6
        Node 6
            weight:      50.0
            depth:       2
            colId:       0
            colName:     
            leftward:    false
            naVsRest:    false
            splitVal:    NaN
            isBitset:    false
            predValue:   1.0
            squaredErr:  3.7480965
            leftChild:   
            rightChild:  
        Node 5
            weight:      50.0
            depth:       2
            colId:       0
            colName:     
            leftward:    false
            naVsRest:    false
            splitVal:    NaN
            isBitset:    false
            predValue:   1.0
            squaredErr:  3.2347205
            leftChild:   
            rightChild:  

    Edges
        Node 0 ---left---> Node 1
        Node 1 ---left---> Node 5
        Node 1 ---right--> Node 6
        Node 0 ---right--> Node 4
```

Under this output, there is the graphviz format as well. I'll document it better.

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
Pscheidl: Internal H2O representation of the graph. Currently, internal representations of the splits. It will print out numerical representation of categorical splits too, together with categorical level labels on the edges.

You can see an example of the behavior in this PR's description (it is up to date).

Pscheidl: This is the functionality you requested. I'll push a better documentation.

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
Pscheidl: It is correct that the `splitPoint != -1` check can be dropped, as any comparison of a Float compared to NaN yields false.

Pscheidl: Resolved in commit https://github.com/h2oai/h2o-3/pull/3534/commits/a82f14048dd9fb0078b555443d787b38522c513f

The commit message says:  `Simplified calculating child inclusive levels with numerically represented categorical split. The regimes discardAll and includeAll are now checked first.`

Thanks to you, I found a potential bug. There are two additional modes:

- include all categoricals
- discard all categoricals

Those in my opinion need to be checked first, as they could be potentially present with both numerical and bitset representation of the split. So I moved those a little bit "more up".

When all categoricals are not included nor discarded, a third check takes place: `else if (!Float.isNaN(splitValue)) ` - if the node has numerically represented categorical split, the inclusive categorical is calculated as `includeThisLevel = Math.ceil(splitValue) > i ^ nodeBitsetDoesContain;`.

If the split is not represented by a numerical split but rather by a `BitSet` (`else if (bs.isInRange(i) && bs.contains(i) == nodeBitsetDoesContain)`), original logic `includeThisLevel = calculateIncludeThisLevel(inheritedInclusiveLevels, i);` is used.

Pscheidl: I also got rid of the Math.ceil operation by inverting the problem from `Math.ceil(splitValue) > i ^ nodeBitsetDoesContain` to `splitValue < i ^ !nodeBitsetDoesContain;`

Should be faster, as the ceil operation is costly, the comparison operation remains and inverting a bit is cheaper.

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
michalkurka: Yeah, I know that :) But the users won't know that from the description. I would either not expose the output (that would be my preference) or document what it does.

Pscheidl: Removed.

PR #3536, opened 2019-05-24
PUBDEV-6523: Prediction fails on a subset of training features when using categorical_encoding

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6523

When using categorical encoding, `adaptTestToTrain` requires that all training features are present in the test frame, which is overly strict as it is not required by default (without encoding).


seb-h2o:
@michalkurka please have a look, feel free to include any expert of this part.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1277,8 +1281,12 @@ else if (isWeights && computeMetrics) {
       }
       vvecs[i] = vec;
     }
+
+    if( good == convNaN )
+      throw new IllegalArgumentException("Test/Validation dataset has no columns in common with the training set");
+
```
seb-h2o: fail early, otherwise another weaker condition may return earlier

PR #3540, opened 2019-05-28
PUBDEV-6532 - XGBoost should not use grouping in parameter localization

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6532

XGBoost formats non-integer numbers (doubles, floats) using Locale.ENGLISH to ensure a decimal point "." is used instead of a comma ",".

This locale setting groups large numbers by thousands and splits the groups with ",", which is unparseable to XGBoost.

https://github.com/dmlc/dmlc-core/blob/3943914eed66470bd010df581e29e4dca4f7df6f/include/dmlc/parameter.h#L1011

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
@@ -1,15 +1,11 @@
 package hex.tree.xgboost;
 
-import sun.misc.FloatingDecimal;
 import water.Iced;
 import water.util.IcedHashMapGeneric;
 import water.util.TwoDimTable;
 
-import java.text.DecimalFormat;
-import java.text.NumberFormat;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Locale;
 import java.util.Map;
 
 /**
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
@@ -1,5 +1,6 @@
 package hex.tree.xgboost;
 
+import sun.misc.FloatingDecimal;
 import water.Iced;
 import water.util.IcedHashMapGeneric;
 import water.util.TwoDimTable;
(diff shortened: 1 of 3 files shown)


h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostParamsTest.java
```diff
@@ -35,6 +35,7 @@ public void parametersLocaleIndependent() {
       XGBoostModel.XGBoostParameters parms = new XGBoostModel.XGBoostParameters();
       parms._ntrees = 1;
       parms._max_depth = 5;
+      parms._min_rows = 5000000;
```
Pscheidl: translates to _min_child_weight, watch out :)

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
```diff
@@ -43,6 +43,7 @@ public static BoosterParms fromMap(Map<String, Object> map) {
   private static Map<String, Object> localizeDecimalParams(final Map<String, Object> params) {
     Map<String, Object> localized = new HashMap<>(params.size());
     final NumberFormat localizedNumberFormatter = DecimalFormat.getNumberInstance(Locale.ENGLISH);
```
honzasterba: should we really use localized number format instead of some IEEE standard number format?

Pscheidl: Which IEEE standard do you have in mind ? The point here is to make sure decimal point is unified.

honzasterba: hmm, good question. then should we just use new DecimalFormat("0.0;-0.0") if English is really not what we want?

Pscheidl: That would have the same effect indeed.

Pscheidl: Won't work. Try it out. It will still use the default locale.

michalkurka: Why is Double.toString() not enough? :)

Pscheidl: It's the same. I'll do it for you.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
```diff
@@ -1,5 +1,6 @@
 package hex.tree.xgboost;
 
+import sun.misc.FloatingDecimal;
```
michalkurka: sun.misc?

Pscheidl: Unused. Gone together with others. Thanks.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/BoosterParms.java
Pscheidl: This is already resolved.

PR #3543, opened 2019-05-28
PUBDEV-6501: Disable examples in R package

ledell:
Changes all the `\donttest` to `\dontrun` in the R examples so we don't run into issues with CRAN servers.
https://0xdata.atlassian.net/browse/PUBDEV-6501


michalkurka:
auto merged to master - not needed anymore

PR #3546, opened 2019-05-29
PUBDEV-6134: Adding SVM as a Model

michalkurka:
@wendycwong still WIP; but more or less in the shape I want it to be

This PR is not the last one, missing features and missing py/r client integration will come later.


h2o-algos/src/main/java/hex/FrameTask.java
```diff
@@ -187,13 +189,18 @@ protected void chunkDone(long n){}
         }
         assert(r >= 0 && r<=nrows);
 
+        seed = offset + rep * nrows + r;
+        if (skipRow(seed)) {
+          num_skipped_rows++;
+          continue;
+        }
+        row = _dinfo.newDenseRow(); // FIXME
```
michalkurka: Working on a fix; will be addressed in this PR

PR #3548, opened 2019-05-31
PUBDEV-6432 Added support for single te column as string or index in Py TE API

deil87:
This PR is a cherry-pick from a [PR](https://github.com/h2oai/h2o-3/pull/3440) that was already merged into master.


deil87:
@michalkurka what about merging this?

PR #3550, opened 2019-05-31
[PUBDEV-6533] FLOW - fix NaNs and undefined in cluster status

honzasterba:
- put back some of previously removed heartbeat values

### DIFF
--- a/h2o-core/src/main/java/water/api/schemas3/CloudV3.java
+++ b/h2o-core/src/main/java/water/api/schemas3/CloudV3.java
@@ -134,6 +134,8 @@ public class CloudV3 extends RequestSchemaV3<Iced, CloudV3> {
 
     @API(help="Data on Node memory", direction=API.Direction.OUTPUT)
     public long mem_value_size;
+    @API(help="Data on Node (memory or disk)", direction=API.Direction.OUTPUT)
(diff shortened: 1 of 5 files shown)


h2o-core/src/main/java/water/api/schemas3/CloudV3.java
```diff
@@ -134,6 +134,8 @@ public NodeV3() {}
 
     @API(help="Data on Node memory", direction=API.Direction.OUTPUT)
     public long mem_value_size;
+    @API(help="Data on Node (memory or disk)", direction=API.Direction.OUTPUT)
```
michalkurka: I am not sure about the "memory or disk" comment. H2O doesn't swap anything on disk (anymore).

It is possible that some Value is not yet loaded in memory because - that is a case of a FileVec - the bytes are loaded on demand. The source of the bytes can be a local disk, S3, HTTP source, HDFS, GCS.

honzasterba: H2O doesn't swap anything on disk (anymore).

Aha! that makes a couple of things in the UI obsolete and confirms my suspicion. I think that means that this whole PR will be canceled and changes will be just in flow UI

PR #3551, opened 2019-05-31
[PUBDEV-6528] do not test xgboost on ubuntu 14 GPU

honzasterba:
- it is no longer supported


michalkurka:
@honzasterba why is it no longer supported?

@angela0xdata we would need to fix the documentation if this gets merged

honzasterba:
@michalkurka we discussed this with @michal-raska -  GPUv4 requires OMPv4 which is not supported on Ubuntu 14. this not a new thing and just surfaced now because we fixed all the other issues with xgb smoke tests

PR #3554, opened 2019-05-31
PUBDEV-6515 - MOJO loading fails when categorical values contain a newline character

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6515

The problem is simple, we're reading the input by lines. And if a categorical value has a new line at the end, it will be parsed as a new empty categorical levels.

The solution is simple: one categorical per line, with harmful characters escaped (new line chars).

### DIFF
--- a/h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
+++ b/h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
@@ -11,7 +11,6 @@ import water.fvec.Frame;
 import water.fvec.TestFrameBuilder;
 import water.fvec.Vec;
 
-import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
--- a/h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
+++ b/h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
@@ -116,6 +116,8 @@ public class MojoIntegrationTest {
   /**
    * Custom Mojo writer with categorical levels serialized "the old way" - one categorical per row.
    * This approach made it impossible to represent multi-level categoricals.
+   * 
+   * No need to override `writekv("multiline_categoricals", true);` to false, as the DFA should be able to recognize a line not beginning with a quote.
    */
   private static final class OldDomainSerializationGBMMojoWriter extends GbmMojoWriter {
 
--- a/h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
+++ b/h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
@@ -31,8 +31,8 @@ public class MojoIntegrationTest {
     try {
       Scope.enter();
 
-      final Frame trainingFrame = new TestFrameBuilder().withDataForCol(0, new String[]{"L1\n \"\"L1", "L2\n\r", "\nL3\n", "L4\r", "\\"})
-              .withDataForCol(1, new String[]{"\"R1", "R2", "R3", "", "   "})
+      final Frame trainingFrame = new TestFrameBuilder().withDataForCol(0, new String[]{"L1\n \"\"L1", "L2\n\r", "\nL3\n", "L4\r", "\\", "L6\f", "\rL7\b"})
+              .withDataForCol(1, new String[]{"\"R1", "R2", "R3", "", "   ", "R6\t", "R7\b"})
               .withUniformVecTypes(2, Vec.T_CAT)
               .build();
 
--- a/h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
+++ b/h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
@@ -51,9 +51,7 @@ public class MojoIntegrationTest {
       final GBMModel gbmModel = gbm.trainModel().get();
       assertNotNull(gbmModel);
 
-      final ByteArrayOutputStream originalModelMojo = new ByteArrayOutputStream();
       final File originalModelMojoFile = File.createTempFile("mojo", "zip");
-      gbmModel.getMojo().writeTo(originalModelMojo);
       gbmModel.getMojo().writeTo(new FileOutputStream(originalModelMojoFile));
 
       MojoModel mojoModel = MojoModel.load(originalModelMojoFile.getAbsolutePath());
@@ -94,9 +92,7 @@ public class MojoIntegrationTest {
       final GBMModel gbmModel = gbm.trainModel().get();
       assertNotNull(gbmModel);
 
-      final ByteArrayOutputStream originalModelMojo = new ByteArrayOutputStream();
       final File originalModelMojoFile = File.createTempFile("mojo", "zip");
-      gbmModel.getMojo().writeTo(originalModelMojo);
       new OldDomainSerializationGBMMojoWriter(gbmModel).writeTo(new FileOutputStream(originalModelMojoFile));
 
 
--- a/h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
+++ b/h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
@@ -1,0 +1,146 @@
+package hex.mojo;
+
+import hex.genmodel.MojoModel;
+import hex.tree.gbm.GBM;
+import hex.tree.gbm.GBMModel;
+import hex.tree.gbm.GbmMojoWriter;
+import org.junit.Before;
+import org.junit.Test;
+import water.*;
+import water.fvec.Frame;
+import water.fvec.TestFrameBuilder;
+import water.fvec.Vec;
+
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+
+import static org.junit.Assert.*;
+
+public class MojoIntegrationTest {
+
+  @Before
+  public void setUp() throws Exception {
+    TestUtil.stall_till_cloudsize(1);
+  }
+
+
+  @Test
+  public void testMojo_MultilineCategoricals() throws IOException {
+    try {
+      Scope.enter();
+
+      final Frame trainingFrame = new TestFrameBuilder().withDataForCol(0, new String[]{"L1\n \"\"L1", "L2\n\r", "\nL3\n", "L4\r", "\\"})
+              .withDataForCol(1, new String[]{"\"R1", "R2", "R3", "", "   "})
+              .withUniformVecTypes(2, Vec.T_CAT)
+              .build();
+
+      Scope.track(trainingFrame);
+
+      GBMModel.GBMParameters parameters = new GBMModel.GBMParameters();
+      parameters._train = trainingFrame._key;
+      parameters._response_column = trainingFrame._names[trainingFrame.numCols()-1];
+      parameters._seed = 0XFEED;
+      parameters._max_depth = 1;
+      parameters._ntrees = 1;
+      parameters._min_rows = 1;
+      
+      
+      GBM gbm = new GBM(parameters);
+      final GBMModel gbmModel = gbm.trainModel().get();
+      assertNotNull(gbmModel);
+
+      final ByteArrayOutputStream originalModelMojo = new ByteArrayOutputStream();
+      final File originalModelMojoFile = File.createTempFile("mojo", "zip");
+      gbmModel.getMojo().writeTo(originalModelMojo);
+      gbmModel.getMojo().writeTo(new FileOutputStream(originalModelMojoFile));
+
+      MojoModel mojoModel = MojoModel.load(originalModelMojoFile.getAbsolutePath());
+      assertNotNull(mojoModel);
+      
+      for (int i = 0; i < gbmModel._output._domains.length; i++) {
+        assertArrayEquals(gbmModel._output._domains[i], mojoModel._domains[i]);
+      }
+
+
+    } finally {
+      Scope.exit();
+    }
+  }
+
+  @Test
+  public void testMojo_parseUnquoted() throws IOException {
+    try {
+      Scope.enter();
+
+      final Frame trainingFrame = new TestFrameBuilder().withDataForCol(0, new String[]{"LVL1", " LVL2  ", "LVL 3 ", " LVL4 "}) // Third with space in the middle
+              .withDataForCol(1, new String[]{"RLVL1", "RLVL2", "", "RLVL4"}).
+                      withUniformVecTypes(2, Vec.T_CAT)
+              .build();
+
+      Scope.track(trainingFrame);
+
+      GBMModel.GBMParameters parameters = new GBMModel.GBMParameters();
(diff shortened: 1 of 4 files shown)


h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
```diff
@@ -251,6 +252,7 @@ else if (line.equals("[columns]")) {
     return info;
   }
 
+  private Pattern CRLF_ONLY_LINE = Pattern.compile("[\r\n]*");
```
honzasterba: this will also match an empty string, if that is expected it should be part of the test

Pscheidl:
This PR has a WIP on it and the description says it all. It is not presented as a fix for the problem, but as a way to reproduce and an exact pointer to where the problem lies.

In the description, I ask about various ways of solving it. One of them is solving it "in the writer", as you mention. So I guess you vote for fixing it on the writer.

Introducing a structure and a schema (e.g. XML, example chosen randomly) would fix the problem for MOJOs generated in future. Keeping support for the current format would also be perfectly possible. However:

- Fixing it only on the writer might or might not help existing users with existing MOJOs (even though this was reported internally, DAI probably has some existing MOJOs and they'd need to upgrade Genmodel). But there might also be some users affect, even in future. For our customers it is generally slow.
- It makes no sense to do it if we decide we only want one-liner categoricals. 

The debate now is on what should we do. I'd propose introducing a schema to the files with categorical levels, making it a lot easier to parse and any values might be included.

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
Pscheidl: Yes, this is intended and it will for sure be part of the test. Thanks for noticing. Once we know this is the way to go, I'll finish the PR. Sorry for not making it obvious.

michalkurka:
> Do we want to enable multi-line categoricals in MOJO ?

MOJO needs to be able to represent whatever H2O is able to represent. Otherwise, we will run into a similar issue again.

Pscheidl:
@michalkurka Thanks. Will do.

Pscheidl:
I've implemented a simple way to achieve multi-line categorical levels. Each level is from now on **enclosed** in quotes. Inside the quotes, there can be anything.

- Parsing stays `O(n)`, as there is a state machine taking care of it
- Backwards compatible - the DFA is able to recognize lines not starting with quotes and treat them the old way - line terminator (one of {\n,\r} - see code for details) means the end of the categorical level.
- Testing the old way and the new way

This has one flaw. If **old categorical contains a quote at the beginning**, then they'll be treated as new ones. I think we can not risk it. One way to solve it is to raise the MOJO versions or introduce new internal structure versioning. The DFA will the use the proper strategy for both and it can not make a mistake anymore. Before I make this final touch, I'd like to know your opinion again.

Thank you.

Pscheidl:
Also small self-correction : Introduction of schema would not solve the problem 100%, as there could be literally any characters in the old MOJOs. Probabilistically, it is doable, but it's not a flawless fix. According to the infinite monkey theorem, if there was a monkey generating datasets, training models & exporting them as MOJO with old GenModel and loading them with new GenModel, it would surely find a variation of characters matching the schema.

Pscheidl:
Added a new flag inside the MOJO description to indicate the domains do support multi-line categoricals. Otherwise the state machine will never trigger the quoted regime and will consider only one categorical level per line.

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
```diff
@@ -263,26 +291,95 @@ else if (line.equals("[columns]")) {
...
-        int id = 0;  // domain elements counter
+
+
+      try (BufferedReader br = _reader.getTextFile("domains/" + domfile)) {
+        States state = States.NEW_CATEGORICAL;
+        StringBuilder stringBuilder = new StringBuilder();
+        int quoteCount = 0;
+        int domainIdx = 0;
+        int c;
+        parsingLoop:
```
honzasterba: I believe labels and controls statements are a really big code-smell and should not be used.

h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
```diff
@@ -0,0 +1,148 @@
...
+      parameters._min_rows = 1;
+      
+      
+      GBM gbm = new GBM(parameters);
+      final GBMModel gbmModel = gbm.trainModel().get();
+      assertNotNull(gbmModel);
+
+      final ByteArrayOutputStream originalModelMojo = new ByteArrayOutputStream();
+      final File originalModelMojoFile = File.createTempFile("mojo", "zip");
+      gbmModel.getMojo().writeTo(originalModelMojo);
```
honzasterba: not used

h2o-algos/src/test/java/hex/mojo/MojoIntegrationTest.java
```diff
@@ -0,0 +1,148 @@
...
+      parameters._min_rows = 1;
+
+
+      GBM gbm = new GBM(parameters);
+      final GBMModel gbmModel = gbm.trainModel().get();
+      assertNotNull(gbmModel);
+
+      final ByteArrayOutputStream originalModelMojo = new ByteArrayOutputStream();
+      final File originalModelMojoFile = File.createTempFile("mojo", "zip");
+      gbmModel.getMojo().writeTo(originalModelMojo);
```
honzasterba: not used

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
```diff
@@ -233,16 +235,21 @@ private void writeModelInfo() throws IOException {
     finishWritingTextFile();
   }
 
+  private static final Pattern QUOTE_PATTERN = Pattern.compile("[\"]{1}"); // Escape quotes
```
honzasterba: {1} is a no-op a well as [] with just one member, it seems like only replaceAll("\"", "\"\"") would be enough.

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
```diff
@@ -233,16 +235,21 @@ private void writeModelInfo() throws IOException {
...
     int domIndex = 0;
     for (String[] domain : model.scoringDomains()) {
       if (domain == null) continue;
       startWritingTextFile(String.format("domains/d%03d.txt", domIndex++));
       for (String category : domain) {
-        writeln(category.replaceAll("\n", "\\n"));  // replace newlines with "\n" escape sequences
+        StringBuilder stringBuilder = new StringBuilder("\"");
+        stringBuilder.append(QUOTE_PATTERN.matcher(category).replaceAll(QUOTE_ESCAPING_REPLACEMENT));
+        stringBuilder.append("\"");
+        writeln(stringBuilder.toString());  // replace newlines with "\n" escape sequences
```
honzasterba: comment does not make sense

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
```diff
@@ -184,6 +185,7 @@ private void addCommonModelInfo() throws IOException {
     writekv("prior_class_distrib", Arrays.toString(model.priorClassDist()));
     writekv("model_class_distrib", Arrays.toString(model.modelClassDist()));
     writekv("timestamp", model.timestamp());
+    writekv("multiline_categoricals", true);
```
Pscheidl: Here is the flag.

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
Pscheidl: String.replaceAll does Pattern.compile all the time, so I'll avoid that. But yes, it can be simplified.

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
Pscheidl: Do you have any specific way of re-implementing this in mind ? It could be splitted into more object-oriented way, but then I'm afraid:

- Lots of new frames could potentially be created on the stack (or at least we'll be at the mercy of the VM to do the inlining)
- The code could in a way be more spaghetti. Lots of methods and the state machine would not be in one place.

I supposed the code style can be adjusted a little bit at least - for example `break states;` can be replaced with more clear `continue parsingLoop` and the `parsingLoop` itself might be ranemed to be more descriptive as well.

I have no intention on dismissing your comment, but there are more way to do it. We might even use a pre-built state machine lib, which is unwanted in GenModel. Implementing state machine on our own could be beneficial, e.g. the next CSV parser could utilize it as well. I'm just unsure about the performance. Well tweaked FSA doing only the few steps necessary is a guarantee of speed.

Please provide some examples on what would you ideally want to do. I'll also try to think about it.

Pscheidl: I'm going to discard this whole block in favor of character escaping. So we'll get rid of the handcrafted state machine.

Pscheidl:
Please re- review.

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
```diff
@@ -233,16 +235,18 @@ private void writeModelInfo() throws IOException {
     finishWritingTextFile();
   }
 
+  private static final Pattern QUOTE_PATTERN = Pattern.compile("[\"]{1}"); // Escape quotes
```
michalkurka: Not needed anymore

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
```diff
@@ -252,6 +256,7 @@ else if (line.equals("[columns]")) {
   }
 
   private String[][] parseModelDomains(int n_columns) throws IOException {
+    final boolean multilineCategoricals = existsInKV("multiline_categoricals") && (boolean) readkv("multiline_categoricals"); // The key might not exist in older MOJOs
```
michalkurka: This can cause NPE if the key is present but the corresponding value is null.

Safe way to do the same thing:

    Boolean.TRUE.equals(readkv("multiline_categoricals"))

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
michalkurka: How about "escape_domain_values"? It would be good to capture how are we actually treating the new lines in the name of the property.

michalkurka:
Looks good, minor suggestions added

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
Pscheidl: Why not ? Renamed.

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
Pscheidl: Thanks for catching it for me. Deleting existsInKVMethod (unused now).

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
Pscheidl: Removed. As well as the line down below this.

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
```diff
@@ -263,20 +264,23 @@ else if (line.equals("[columns]")) {
...
-      BufferedReader br = _reader.getTextFile("domains/" + domfile);
-      try {
+
+      try (BufferedReader br = _reader.getTextFile("domains/" + domfile)) {
         String line;
         int id = 0;  // domain elements counter
         while (true) {
-          line = br.readLine();
-          if (line == null) break;
+          final String str = br.readLine();
```
honzasterba: I would do
        while ((line = br.readLine()) != null) {
istead of nullcheck and break

Pscheidl: Done in https://github.com/h2oai/h2o-3/pull/3554/commits/3fac9b9ebb246830fa8003f81dc4562ba8988ea1.

Your comment actually led me to another fault: The original code uses "line" variable and I added a "str" variable in a local POC and never got rid of that. So now we properly have only one variable named `line` that is escaped if required.

honzasterba: interesting,  how come the tests passed?

honzasterba: ...because it was not a bug, just a code cleanup issue. got it :)

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
```diff
@@ -263,20 +264,19 @@ else if (line.equals("[columns]")) {
...
-      BufferedReader br = _reader.getTextFile("domains/" + domfile);
-      try {
+
+      try (BufferedReader br = _reader.getTextFile("domains/" + domfile)) {
         String line;
         int id = 0;  // domain elements counter
-        while (true) {
-          line = br.readLine();
+        while ((line = br.readLine()) != null) {
           if (line == null) break;
```
honzasterba: unnecessry check.

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
Pscheidl: Because the `line` variable was always assigned correctly. First, the `str` variable was initialized by calling `final String str = br.readLine();` and then either:

1. Escaped and the resulting value assigned to `line` variable
1. Not escaped and just assigned to `line` variable

And at the end `domain[id++] = line;` was done. I just got rid of `str`.

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
Pscheidl: Forgot it there. Removed it in https://github.com/h2oai/h2o-3/pull/3554/commits/eaa08fd3f1171dd7c6d492767141404a19743ef0. Thanks.

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
```diff
@@ -8,6 +8,7 @@
 import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.regex.Pattern;
```
michalkurka: Do we still need this import?

Pscheidl: No, removed.

PR #3555, opened 2019-05-31
PUBDEV-6520: Expose RegexTokenizer to users of Java API

michalkurka:


h2o-core/src/test/java/hex/RegexTokenizerTest.java
```diff
@@ -0,0 +1,65 @@
...
+    Frame f = new Frame(Key.<Frame>make());
+    f.add("C1", svec("a b c", null, "1 2", "cowabunga"));
+    f.add("C1", svec(null, null, "A", null));
+    DKV.put(f);
+    _tracker.track(f);
+
+    Frame tokenized = new RegexTokenizer(" ").transform(f);
+    _tracker.track(tokenized);
+
+    Vec expected = svec("a", "b", "c", null, null, "1", "2", "A", null, "cowabunga", null);
```
mn-mikke: WDYT about ignoring empty rows:
```svec("a", "b", "c", null, "1", "2", "A", null, "cowabunga", null)```

File: h2o-core/src/main/java/hex/RegexTokenizer.java [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+ * final RegexTokenizer tokenizer = new RegexTokenizer.Builder()
+ *   .setRegex("[,;]")
+ *   .setMinLength(2)
+ *   .setToLowercase(true)
+ *   .create();
+ * final Frame tokens = tokenizer.transform(inputFrame);
+ * }
+ * </pre>
+ */
+public class RegexTokenizer extends MRTask<RegexTokenizer> {
```
mn-mikke: WDYT about exposing a property that would enable to specify what columns will be used for tokenization? Couldn't  a frame contain also other features that are not related to word2vec?

h2o-core/src/main/java/hex/RegexTokenizer.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private RegexTokenizer(String regex, boolean toLowercase, int minLength) {
+    _regex = regex;
+    _toLowercase = toLowercase;
+    _minLength = minLength;
+  }
+
+  @Override
+  public void map(Chunk[] cs, NewChunk nc) {
```
mn-mikke: It would be nice to use some class instead of ```Chunk[]``` in future. I needed a help from @jakubhava to understand that this describes one partition across multiple columns.

jakubhava: Just clarification -> it describes one partition in a vector, but there is ( at least as I see it) unwritten invariant that says that for example, all the first chunks across different vectors have to have same size

h2o-algos/src/main/java/hex/word2vec/Word2Vec.java
```diff
@@ -111,4 +113,26 @@ private void convertToModel(Frame preTrained, Word2VecModel model) {
...
+  public static Job<Word2VecModel> fromPretrainedModel(Frame model) {
+    if (model != null || model.numCols() < 2) {
+      throw new IllegalArgumentException("Frame representing an external word2vec needs to have at least 2 columns.");
+    }
+    if (model.vec(0).get_type() != Vec.T_STR) {
+      throw new IllegalArgumentException("First column is expected to contain the dictionary words and be represented as String, " +
+              "instead got " + model.vec(0).get_type_str());
+    }
+    for (int i = 1; i < model.numCols(); i++) {
+      if (model.vec(i).get_type() != Vec.T_NUM) {
```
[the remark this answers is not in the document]
michalkurka: Good idea

h2o-core/src/main/java/hex/RegexTokenizer.java
michalkurka: Instead of doing that user can just subset the frame's columns. IMHO that is a better approach from maintainability point of view.

Example:

    final Frame tokens = tokenizer.transform(inputFrame.subframe(new String[]{"a", "b"}));

h2o-core/src/main/java/hex/RegexTokenizer.java
michalkurka: Yes, that is a good idea - having this is also necessary for true sparse support.

h2o-core/src/test/java/hex/RegexTokenizerTest.java
michalkurka: If I ignore the empty rows I lose the mapping back to original sentences. Currently, I can rely on this invariant: the number of NAs in the output is equal to the number of rows. That lets me map the words back using a simple transformation (that H2O is capable of doing).

If I start skipping empty rows I might be not able to map it back so easily (because there is also filtering on minimum lenght - output might be empty even if the source had words).

mn-mikke: It makes sense. Thanks for your explanation!

h2o-core/src/main/java/hex/RegexTokenizer.java
mn-mikke: Ok and other features can be appended to the output of word2vec by some `join` operation, I assume?

michalkurka: IMHO that is not necessary (or possible), you only use the tokenizer for word2vec training (as a direct input to the model builder). If you have multiple algos, then they will have their own inputs. Perhaps this is the reason why this tokenizer is not good for SW integration. However, imho it makes sense in H2O's context.

If you want to use word2vec's transform then you can just run Word2VecModel#transform() and column bind the rest of the features if they are supposed to be used as an input of a dependent model.

mn-mikke: Thanks for clarification!

michalkurka:
Thanks for the review @mn-mikke

PR #3558, opened 2019-06-02
PUBDEV-6538 - Expose retainKeys

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6538


Pscheidl:
https://tools.ietf.org/html/rfc7231#section-4.3.5

Pscheidl:
There were test failures due to an older check for `h2o.removeAll` Python signature written by @wendycwong. Including @wendycwong into this review. The very fix of this test is in the following commit: https://github.com/h2oai/h2o-3/pull/3558/commits/d60ed0bdea154d23b57226e338c73900b9446da0

h2o-core/src/main/java/water/api/RemoveAllHandler.java
```diff
@@ -23,14 +20,47 @@ public RemoveAllV3 remove(int version, RemoveAllV3 u) {
...
     H2O.getPM().getIce().cleanUp();
     H2O.updateNotIdle();
-    Log.info("Finished removing objects");
-    return u;
   }
-}
\ No newline at end of file
+
+  private void retainKeys(final KeyV3[] retained_keys) {
+    Log.info(String.format("Removing all objects, except for %d provided key(s)", retained_keys.length));
```
michalkurka: possible NPE? line 52 permits `retained_keys` to be null

h2o-core/src/main/java/water/api/RemoveAllHandler.java
```diff
@@ -23,14 +20,47 @@ public RemoveAllV3 remove(int version, RemoveAllV3 u) {
...
+    final Key[] retainedKeys;
+    if (retained_keys == null) {
+      retainedKeys = new Key[0];
+    } else {
+      retainedKeys = new Key[retained_keys.length];
+      for (int i = 0; i < retainedKeys.length; i++) {
+        if (retained_keys[i] == null) continue; // Protection against null keys from the client - ignored
```
michalkurka: It would probably be better to avoid passing null to ClearDKVTask

h2o-core/src/main/java/water/api/RemoveAllHandler.java
Pscheidl: This is actually secured on line 24 `if (u.retained_keys != null && u.retained_keys.length != 0) {`.

h2o-core/src/main/java/water/api/RemoveAllHandler.java
Pscheidl: Retaining null results in an effect no-op. But the user won't know. The documentation says nulls are silently ignored.

I'll add a verification mechanism.

Pscheidl: Added. Also please note the DKVCleaner itself does not validate if the key names are actually valid. Non-valid/not existing keys will simply be ignored, as the algorithm iterates keys in DKV and tries to match them against a set with keys to retain.

h2o-core/src/main/java/water/api/RemoveAllHandler.java
michalkurka: ok, I see - I got confused by the naming `retained_keys` vs `retainedKeys`

makes sense, sorry

PR #3560, opened 2019-06-04
[ITA-375] Add Python 3.7+ images.

michal-raska:
https://0xdata.atlassian.net/browse/PUBDEV-5932

Adding support for Py 3.7 and 3.8.

Side changes: when rebuilding images, also had to fix R dependencies. Due to recent release of R4, many R libraries release a new update which may not be compatible with some of our requirements. Full support for R4 is however left to another task (and someone else..).


h2o-py/test-requirements.txt
```diff
@@ -12,13 +12,15 @@ twine==1.10.0
...
+scipy==1.2.2; python_version == '2.7'
+scipy==1.3.0; python_version > '2.7'
+pandas==0.24.2
+git+https://github.com/statsmodels/statsmodels#egg=975813e180c3b727e1d19ed40a54e0e343105dec # use patched version, latest release 0.9.0 is not compatible with other libs
 patsy==0.5.0
 sklearn
-scikit-learn==0.19.1
+scikit-learn==0.20.3; python_version == '2.7'
+scikit-learn==0.21.2; python_version > '2.7'
 six==1.11.0
```
seb-h2o: @michal-raska: these are a lot of library upgrades at once: are they all used by our tests? or demos maybe?
Even if all those upgrades look `minor` according to semver, some of them are major: for example, I'm aware of several backwards incompatibilities for scikit-learn 0.19 to 0.20.
That's why I'm wondering if those are necessary or just good-to-have.

docker/Makefile
```diff
@@ -168,6 +168,36 @@ ifeq ($(shell echo $(PUSH) | tr [:upper:] [:lower:] ),true)
 	docker push harbor.h2o.ai/opsh2oai/h2o-3/dev-python-3.6-jdk-$*$(GPU_SUFFIX):$(VERSION)
 endif
 
+ifneq ($(CI), 1)
+dev-python-3.7-openjdk-7: dev-python-3.7
+endif
+dev-python-3.7-openjdk-7:
```
seb-h2o: is it for the "perverts" who want to use bleeding-edge Python with 8yo Java? :)
Meaning, is it worth providing/supporting this?

seb-h2o:
@michal-raska : looks like many test errors are due to library upgrades, not sure it's related with Py 3.7.
I can help with this though.
Do you plan to add Jenkins stages for Py 3.7 as well? `Py3.7 Init Java 8`, `Py3.7 Medium-large` ? as I don't see any test executed against Py 3.7 for now, right?

h2o-py/test-requirements.txt
michal-raska: I was unable to get the lower versions installed for Python 3.7. This is the minimal set we need to upgrade in order to build the Python 3.7 images.

docker/Makefile
michal-raska: @michalkurka will have to tell if it's required, but there was an issue (I'm not sure now what was the issue though) which was there only if launching H2O backend from Python/R client. I think we want to have all combinations covered, but we can remove.

michal-raska:
> @michal-raska : looks like many test errors are due to library upgrades, not sure it's related with Py 3.7.
> I can help with this though.
> Do you plan to add Jenkins stages for Py 3.7 as well? `Py3.7 Init Java 8`, `Py3.7 Medium-large` ? as I don't see any test executed against Py 3.7 for now, right?

Yes, my initial though was to debug the existing stages and once they are "master green-ish" then add and debug the Py 3.7 stages. Do you want me to add them now?

seb-h2o:
@michal-raska : I removed use of deprecated sklearn APIs and found some library versions that seems to work with any Py version.
Let's see the current build state and then you can add the `Py 3.7` build step.

docker/scripts/java-11-vars.sh
```diff
@@ -2,5 +2,9 @@
 
 JAVA_INSTALL_VERSION_MAJOR='11'
 JAVA_INSTALL_VERSION=3
+MAGIC_B_NUM=12
+JAVA_INSTALL_LINK_HASH=37f5e150db5247ab9333b11c1dddcd30
 JAVA_INSTALL_PATH=/usr/opt/java-${JAVA_INSTALL_VERSION_MAJOR}-${JAVA_INSTALL_VERSION}/
-JDK_FILE=/tmp/jdk-11.0.3_linux-x64_bin.tar.gz
+JDK_VERSION=${JAVA_INSTALL_VERSION_MAJOR}.0.${JAVA_INSTALL_VERSION}
+DOWNLOAD_URL=http://download.oracle.com/otn-pub/java/jdk/${JDK_VERSION}+${MAGIC_B_NUM}/${JAVA_INSTALL_LINK_HASH}/jdk-${JDK_VERSION}_linux-x64_bin.tar.gz
```
honzasterba: we should avoid downloading from oracle if possible, every now and then they make a new release and than this link stops working

docker/Jenkinsfile-build-docker
```diff
@@ -1,7 +1,7 @@
 @Library('test-shared-library') _
 import ai.h2o.ci.buildsummary.StagesSummary
 
-DEFAULT_LABEL = 'mr-0xc10'
+DEFAULT_LABEL = 'mr-0x2'
```
honzasterba, line 4: why? also probably unrelated o this change

docker/Makefile
honzasterba: h2o backend is jdk8+, jdk7 is supported only when MOJO+genmodel

sebhrusen: still raises the question if it makes sense to keep all clients combinations with jdk7: after all, the communication with backend relies only on  REST API, so I personally don't see the benefit of trying all Py versions with jdk7.
I added the new ones for consistency, but I'd be happy to keep only let's py2.7+jdk7 and py3.5+jdk7

docker/Jenkinsfile-build-docker
sebhrusen: hmm, will revert, the change was originally made by @michal-raska in original PR, but may have been a temporary change.

docker/scripts/java-11-vars.sh
sebhrusen: let me reverse this too...

docker/scripts/install_python_version
```diff
@@ -24,20 +24,35 @@ wget -nv https://raw.githubusercontent.com/h2oai/h2o-3/${H2O_BRANCH}/h2o-py/docs
 IFS=',' read -r -a array <<< ${VERSIONS_TO_INSTALL}
 for python_version in "${array[@]}"; do
   echo "###### Installing Python ${python_version} ######"
+  if [[ "${python_version}" =~ ^([0-9])\.([0-9]+).* ]]; then
```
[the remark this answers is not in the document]
sebhrusen: added the if-clause mainly to be able to use bash regex, but yeah, let me add the else clause.

h2o-py/requirements.txt
```diff
@@ -1,26 +1,29 @@
...
-sklearn
-six
-seaborn
-matplotlib
-numpydoc
-mysqlclient
-recommonmark
+
+# [-data science]
+#numpy
```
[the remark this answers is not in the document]
sebhrusen: was doing some cleanup without being sure at first if this should be deleted.
Now that it works, let me delete those commented out requirements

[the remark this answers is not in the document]
sebhrusen: ah no, these are non-mandatory requirements, which are used in `h2o-py` packages, but only for integration.
I think I prefer to keep them in this file: it provides a single reference point listing all the libraries we use and/or support.
I'll add a comment in this file explaining why some of them are commented out.

PR #3562, opened 2019-06-04
[PUBDEV-6534] allow vec type conversion from T_BAD

honzasterba:
- conversion to string is straigt forward as the code already supports it
- conversion to enum done via tostring->toenum

I am sure I am missing something why this is not a good idea (other than toEnum(toString()) being inefficient). Am I?

### DIFF
--- a/h2o-core/src/main/java/water/util/VecUtils.java
+++ b/h2o-core/src/main/java/water/util/VecUtils.java
@@ -38,6 +38,7 @@ public class VecUtils {
       case Vec.T_CAT:
         return src.makeCopy(src.domain());
       case Vec.T_NUM:
+      case Vec.T_BAD:
         return numericToCategorical(src);
       case Vec.T_STR: // PUBDEV-2204
         return stringToCategorical(src);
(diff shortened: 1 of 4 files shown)


h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -41,6 +41,8 @@ public static Vec toCategoricalVec(Vec src) {
         return numericToCategorical(src);
       case Vec.T_STR: // PUBDEV-2204
         return stringToCategorical(src);
+      case Vec.T_BAD:
+        return stringToCategorical(numericToStringVec(src));
```
michalkurka: Why do we need to first convert to string and then to categorical? BAD Vecs are all NA, shouldnt numericToCategorical work just as well?

honzasterba: good point, fixed

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6534

Pscheidl:
Just one thing struck me: Why do we test this core functionality in Python ? Why not, it's not big deal, but I wanted to give it a try as a JUnit test. Then I immediately reminded myself of `TestFrameBuilder` being unable to deal with BAD vecs. So I created a completely unrelated PR with an attempt to make our life easier in the future: https://github.com/h2oai/h2o-3/pull/3565

PR #3563, opened 2019-06-04
PUBDEV-5885: consistent support for frame arg as key in R

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-5885

R client was mostly supporting frame args as key/ID, but failed for ensembles and automl

To REVIEWERS:
many changed files, but mostly generated ones.
Please pay mainly attention to:
- `frame.R`, `gen_R.py` and `automl.R`
- tests


h2o-bindings/bin/gen_R.py
```diff
@@ -176,6 +170,30 @@ def gen_module(schema, algo, module):
         yield "         x <- setdiff(colnames(training_frame), event_column)"
```
seb-h2o: blocks above and below were moved around because of this statement that was using training_frame BEFORE it was actually validated

Pscheidl: This actually makes sense even as a standalone refactor.

michalkurka:
@seb-h2o I think your implementation is good but am not sure if I agree with the jira. IMHO we should really not teach our users to access `model$model_summary$training_frame` like this

    se <- h2o.stackedEnsemble(y = 5, training_frame = aml@leader@model$model_summary$training_frame, base_models = model_ids)

I would much rather see a helper method that would help extract the training frame from a model.

IMHO it makes the API inconsistent. Why do we support keys in model builders and not anywhere else?

@seb-h2o @ledell WDYT?

seb-h2o:
@michalkurka I agree that consistency should be the main concern in this case, but in any case it requires changes as currently:
- R client supports passing frame as keys for all algos (although not for `h2o.predict`, another inconsistency).
- Py client doesn't support frame as keys anywhere: consistent internally, but inconsistent if we compare client APIs.

I also agree that `training_frame = aml@leader@model$model_summary$training_frame` is not very elegant and should not be encouraged, but there are many other use-cases:
```
fr = 'my_frame'
h2o.uploadFile(destination_frame=fr, ...)
...
h2o.gbm(training_frame=fr, ...)
```
not saying that this is a good way to use the API...

personally, from an API perspective, I would prefer to completely remove frame-as-keys support in R client, but this is not possible today as it would introduce backwards incompatibility: so we're left with 2 choices:
- make R client fully consistent and support frames as keys, regardless of its difference with Py client.
- make Py client consistent with R one, to facilitate transition/translation from one to the other.

ledell:
> I would much rather see a helper method that would help extract the training frame from a model.

@michalkurka What are you thinking, something like `h2o.getModelTrainingFrameID()`?

michalkurka:
@seb-h2o 

> personally, from an API perspective, I would prefer to completely remove frame-as-keys support in R client

agreed 👍

> but this is not possible today as it would introduce backwards incompatibility

also agreed 👍

I think the general goal really is to have feature parity between R and Python, however, we shouldn't apply this in cases when dealing with APIs that are designed poorly or designed to overcome language limitations.

My thinking is the following:
1. It is unrealistic to fix all R functions to accept both key and H2OFrame
2. It is inefficient to use key reference in more than one function call, it can actually be expensive to collect Frame metadata from the backend if Frame has many columns, I would consider the given example to be an anti-pattern

    fr = 'my_frame'
    h2o.uploadFile(destination_frame=fr, ...)
    ...
    h2o.gbm(training_frame=fr, ...)
3. We are discussing frame keys, but same can be applied to model keys as well (should h2o.predict support predicting with a model key? imho not)

My conclusion is: I would fix it for Stacked Ensemble but avoid propagating this feature to Python.

I would also add a convenience function so that users can get an instance of H2OFrame out of a model instance. Eg. `h2o.get_model_frame(h2o.get_leader(aml), type = "training")`. This makes a nicer example and can be ported to Python.

cc: @ledell

ledell:
@michalkurka I like your idea of the `type` argument, but i think we might want to rename it to `h2o.get_model_frame_id()` since it's the ID returned, not the frame.

michalkurka:
@ledell the idea is to return a Frame, not and id/key and work with Frame instances instead of Keys

returning a Frame opens the user all the options in H2O, not just model building

ledell:
@michalkurka Ok, the original purpose of the ticket was supporting frame IDs so that's why I thought you meant ID.  In that case, ignore my suggestion!

seb-h2o:
@ledell , @michalkurka : what do we do with this one?
I could revert all Python changes and keep only the R changes if this works for you? Legacy R code is already supporting train frames as keys in most places, at least it'll be consistent there.
wdyt?

michalkurka:
@seb-h2o I vote in favor of removing it from Python

seb-h2o:
@michalkurka : I reverted all Py client changes here.
Moved some of those changes to https://github.com/h2oai/h2o-3/pull/3587
as I ran into few bugs when writing tests for frame validation, so they're fixed in PUBDEV-6576, just removed the support for frame id as param.

seb-h2o:
verified that failing Jenkins tests are not related with changes -> merging

PR #3564, opened 2019-06-04
PUBDEV-6548: Specify certificate alias when starting H2O with SSL

michalkurka:


h2o-webserver-iface/src/main/java/water/webserver/iface/H2OHttpConfig.java
```diff
@@ -28,4 +30,54 @@
 
   public String context_path;
 
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
```
jakubhava: I'm sure that you thought about that, but is there any specific reason to implement the equals method, can't we use the default implementation?

jakubhava:
Once this is in, we will also expose it in a form of spark option in sw https://0xdata.atlassian.net/browse/SW-1325

h2o-webserver-iface/src/main/java/water/webserver/iface/H2OHttpConfig.java
michalkurka: Default implementation? Do you mean the equals from Object? That is just comparing the object reference.

Or do you mean equals generated at runtime by a utility function/library?

This code is generated by IntelliJ, not handwritten.

jakubhava: Cool! What do you think about mentioning that this was generated by IDE?

michalkurka: Sure, I will add!

PR #3565, opened 2019-06-05
PUBDEV-6550 - TestFrameBuilder BAD Vec support

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6550

During the review of https://github.com/h2oai/h2o-3/pull/3562/files , I've discovered the `TestFrameBuilder` class does not allow us to create Frames with BAD vectors. 

This PR does the bare minimum (an all-NaN column inside a frame) and tries to prepare the ground for future PRs.

### DIFF
--- a/h2o-core/src/test/java/water/fvec/TestFrameBuilder.java
+++ b/h2o-core/src/test/java/water/fvec/TestFrameBuilder.java
@@ -316,6 +316,9 @@ public class TestFrameBuilder {
             else
               nchunks[colIdx].addNA();
             break;
+          case Vec.T_BAD:
+            nchunks[colIdx].addNum(numericData.get(colIdx)[i]);
+            break;
           default:
             throw new UnsupportedOperationException("Unsupported Vector type for the builder");
 
@@ -342,6 +345,7 @@ public class TestFrameBuilder {
       switch (vecTypes[i]){
         case Vec.T_TIME:
         case Vec.T_NUM:
+        case Vec.T_BAD:
           if(numericData.get(i)==null){
             numericData.put(i, new double[0]); // init with no data as default
           }
@@ -410,6 +414,13 @@ public class TestFrameBuilder {
             numRows = stringData.get(colIdx).length;
           } else {
             throwIf(numRows != stringData.get(colIdx).length, "Columns have different number of elements");
+          }
+          break;
+        case Vec.T_BAD:
+          final double[] data = numericData.get(colIdx);
+          numRows = data.length;
+          for (int i = 0; i < data.length; i++) {
+            throwIf(!Double.isNaN(data[i]), "All elements in a bad column must be NAs or zeros");
           }
           break;
         default:
--- a/h2o-core/src/test/java/water/fvec/TestFrameBuilderTest.java
+++ b/h2o-core/src/test/java/water/fvec/TestFrameBuilderTest.java
@@ -1,9 +1,13 @@
 package water.fvec;
 
+import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


h2o-core/src/test/java/water/fvec/TestFrameBuilder.java
```diff
@@ -412,6 +416,13 @@ private void checkColumnData() {
             throwIf(numRows != stringData.get(colIdx).length, "Columns have different number of elements");
           }
           break;
+        case Vec.T_BAD:
+          final double[] data = numericData.get(colIdx);
+          numRows = data.length;
+          for (int i = 0; i < data.length; i++) {
+            throwIf(!Double.isNaN(data[i]), "All elements in a bad column must be NAs or zeros");
```
michalkurka: zeros? why would a constant 0 chunk be BAD?

h2o-core/src/test/java/water/fvec/TestFrameBuilderTest.java
```diff
@@ -228,4 +232,33 @@ public void withRandomBinaryDataForColTest(){
...
+  public void testBadVec(){
+    try {
+      Scope.enter();
+
+      Frame frame = new TestFrameBuilder()
+              .withVecTypes(Vec.T_BAD)
+              .withDataForCol(0, ard(Float.NaN, Float.NaN, Float.NaN)) // All NaN column
+              .withName("fr")
+              .build();
+      Scope.track(frame);
```
michalkurka: build() method automatically puts the frame into Scope

h2o-core/src/test/java/water/fvec/TestFrameBuilder.java
Pscheidl: Good point, I even had it in the error message :) Support added in https://github.com/h2oai/h2o-3/pull/3565/commits/2838b9968189eec8bbba95e1c4eb4244969bce60. 

Had to tell NewChunk to consider all 0 chunks as BAD as well.

Pscheidl:
Waiting for tests.

h2o-core/src/test/java/water/fvec/TestFrameBuilder.java
Pscheidl: Correction, addressed in https://github.com/h2oai/h2o-3/commit/58445e683c0a3be2dea1d4d37c97403cba49f580

h2o-core/src/test/java/water/fvec/TestFrameBuilderTest.java
Pscheidl: resolved.

h2o-core/src/test/java/water/fvec/TestFrameBuilder.java
Pscheidl: Makes sense, but `NewChunk` consideres BAD types only a vec full of NAs. Which also makes sense. It's a valid chunk full of constant observations.

```
    if(_naCnt == _len)          // All NAs ==> NA Chunk
      return Vec.T_BAD;
```

Pscheidl: Considering a chunk full of zeroes to be `BAD` breaks most of our tests. Setting it to be BAD only if the VEC is pre-assigned as BAD (and not null) is also possible, but it makes no sense to me. I'd stick with current solution, as we're able to represent BAD chunks in tests and deal with this separately.

michalkurka: @Pscheidl I was referring to your error message "All elements in a bad column must be NAs or zeros"

I don't understand the mention of zeros

PR #3567, opened 2019-06-05
[HOTFIX] PUBDEV-6552: XGBoost - fix parallel population of CSR matrix

michalkurka:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/SparseMatrixFactory.java
```diff
@@ -161,30 +154,37 @@ protected void map(int chunkIdx) {
             NestedArrayPointer rowHeaderPointer = new NestedArrayPointer(rwRow);
             NestedArrayPointer dataPointer = new NestedArrayPointer(nonZeroCount);
 
-            for(long i = _f.anyVec().espc()[chunk]; i < _f.anyVec().espc()[chunk+1]; i++) {
-                if (_w != null && _w.at(i) == 0) continue;
+            Chunk weightChunk = _w != null ? _w.chunkForChunkIdx(chunk) : null;
+            Chunk respChunk = _respVec.chunkForChunkIdx(chunk);
+            Chunk[] featChunks = new Chunk[_vecs.length];
```
honzasterba: I think this array is not necessary we can do _vecs[i].chunkForChunkIdx(chunk) inside every for loop and save a bit fo memory, with same complexity

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/SparseMatrixFactory.java
```diff
@@ -161,30 +154,37 @@ protected void map(int chunkIdx) {
...
+                    float val = (float) featChunks[_di._cats + j].atd(i);
                     if (val != 0) {
                         dataPointer.set(_matrix._sparseData, val);
                         dataPointer.set(_matrix._colIndices, _di._catOffsets[_di._catOffsets.length - 1] + j);
                         dataPointer.increment();
                         nonZeroCount++;
                     }
                 }
-                rwRow = setResponseAndWeight(_w, _resp, _weights, _respVec, rwRow, i);
+                rwRow = setResponseAndWeight(_w != null ? _w.new Reader() : null, _resp, _weights, _respVec.new Reader(), rwRow, i);
```
honzasterba: there is an alternative method that deals with vecs directly, maybe we can use those?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/SparseMatrixFactory.java
michalkurka: I would (strongly) advise against that. The default `chunkForChunkIdx` is not super-trivial by itself (it even updates the last accessed time on on the associated Value in DKV). But the main reason is that this method can be and is overridden in WrappedVec implementations. The overriding implementations are typically much more expensive - they typically create a new instance of Chunk.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/SparseMatrixFactory.java
michalkurka: yes, great point - I suggest doing that in the next step

Pscheidl: Is there something preventing us from instantiating ` _w.new Reader()` once per method ?

`Chunk weightChunk = _w != null ? _w.chunkForChunkIdx(chunk) : null;` remains the same for the whole method, so there is probably no reason to instantiate new reader once per `for(int i = 0; i < respChunk._len; i++) {` iteration.

michalkurka: oh, shoot - I am doing that in a loop, will fix

Pscheidl: Sorry I did not notice earlier, but I guess the same applies to `_respVec.new Reader()`.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/SparseMatrixFactory.java
```diff
@@ -122,22 +115,22 @@ public static int initializeFromChunkIds(
...
+        Vec _respVec;
         float[] _resp; 
         float[] _weights;
         
         // OUT
         int[] _actualRows;
         
         InitializeCSRMatrixFromChunkIdsMrFun(
-            Frame f, int[] chunks, Vec.Reader[] vecs, Vec.Reader w, DataInfo di,
+            Frame f, int[] chunks, Vec[] vecs, Vec w, DataInfo di,
```
Pscheidl: Minor, let's not solve in this PR. I've noticed the vecs ALWAYS come from the frame. I guess handing over the `Frame f` is enough ?

PR #3569, opened 2019-06-06
[PUBDEV-6552] small improvements to xgboost parallel matrix population

honzasterba:
continuation of https://github.com/h2oai/h2o-3/pull/3567/files

checked, no measurable impact on performance

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/MatrixFactoryUtils.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/MatrixFactoryUtils.java
@@ -7,7 +7,7 @@ public class MatrixFactoryUtils {
 
     public static int setResponseAndWeight(Chunk[] chunks, int respIdx, int weightIdx, float[] resp, float[] weights, int j, int i) {
         if (weightIdx != -1) {
-            if(chunks[weightIdx].atd(i) == 0) {
+            if (chunks[weightIdx].atd(i) == 0) {
                 return j;
             }
             weights[j] = (float) chunks[weightIdx].atd(i);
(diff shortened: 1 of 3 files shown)


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/SparseMatrixFactory.java
```diff
@@ -128,15 +127,14 @@ public static int initializeFromChunkIds(
         int[] _actualRows;
         
         InitializeCSRMatrixFromChunkIdsMrFun(
-            Frame f, int[] chunks, Vec[] vecs, Vec w, DataInfo di,
+            Frame f, int[] chunks, Vec w, DataInfo di,
```
Pscheidl: Minor suggestion: Since you're doing some meaningful renaming in this PR, I'd use longer names more meaningful for newcomers here as well.

honzasterba: good idea, fixed

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/MatrixFactoryUtils.java
```diff
@@ -16,14 +16,14 @@ public static int setResponseAndWeight(Chunk[] chunks, int respIdx, int weightId
         return j;
     }
 
-    public static int setResponseAndWeight(Vec.Reader w, float[] resp, float[] weights, Vec.Reader respVec, int j, long i) {
-        if (w != null) {
-            if(w.at(i) == 0) {
+    public static int setResponseAndWeight(Chunk weightChunk, Chunk respChunk, float[] resp, float[] weights, int j, int i) {
```
Pscheidl: Using the chunks directly seems like good improvement to me. No need to create readers.

Just thinking out loud: Could this method's logic be inlined somehow ? It is invoked once per line (we have big tests with hundreds of gigabytes and tens of millions of lines). A guarantee of saving us from copying 4 pointers & 32 + 64bit number once per line could mean some performance gain ? I'd investigate how JVM behaves to this method.

PR #3577, opened 2019-06-08
PUBDEV-6549 Retaining shared keys

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6549

After declining introduction of reference counting, reference storing or garbage collecting, we've introduced a new mechanism to retain shared keys. Every class extending `Keyed` may now optionally override `retain_impl` method - which by default deletes everything. This method receives a `Set` of keys the user wishes to retain. A user may retain only a `Frame` or a `Model`, but this set contains all the keys related to a Frame (vectors + misc stuff, content of `Vec._keys`) or a Model (all frames supplied in parameters).

:warning: The behavior of retaining the model shall be discussed. This PR contains a proposal not to delete Model's internal keys (training and validation frames) once a model is retained, as those might be used in future. If I'm mistaken or this behavior is NOT what we want, this behavior can be overridden easily.

Thank you.


h2o-core/src/main/java/water/DKV.java
```diff
@@ -247,23 +247,78 @@ public ClearDKVTask(Key[] retainedKeys) {
...
+   *                     Should contain user-specified keys to retain in order to extract anything.
+   */
+  private static void extractNestedKeys(final Set<Key> retainedKeys) {
+
+    final Iterator<Key> keysIterator = retainedKeys.iterator(); // Traverse keys provided by the user only.
+    while (keysIterator.hasNext()) {
+      final Key key = keysIterator.next();
+      final Value value = Value.STORE_get(key);
+
+      if (!value.isFrame() && !value.isModel()) {
```
honzasterba: could value be null? could the key be invalid?
looking at the code this sure is possible

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -808,6 +808,47 @@ public Futures postWrite(Futures fs) {
...
+    final List<Key> deletedKeys= new ArrayList<>();
+    for (int i = 0; i < delCandidateKeys.length; i++) {
+      if(!retainedKeys.contains(delCandidateKeys[i])){
+        deletedKeys.add(delCandidateKeys[i]);
+      }
+      
+    }
+    
+    // Bulk dumb local remove - no JMM, no ordering, no safety.
+    Vec.bulk_remove(deletedKeys.toArray(new Key[]{}), v.nChunks());
```
honzasterba: should the frame also remove self?

h2o-core/src/main/java/water/DKV.java
Pscheidl: Correct. I'll fix this. Also, the multiple ifs might in fact all be one if-else statement.

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -808,6 +808,47 @@ public Futures postWrite(Futures fs) {
     return fs;
   }
 
+  @Override
+  protected Futures retain_impl(Futures fs, Set<Key> retainedKeys) {
+    final Key[] delCandidateKeys = _keys;
```
Pscheidl: Just noticed strong code duplication here. I'll remove it.

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: That's what method `remove()` does for removal before calling this `remove_impl()` method. So method `retain` does this as well, before calling `retain_impl()`. Followed the pattern that is already there.

h2o-core/src/main/java/water/DKV.java
Pscheidl: Fixed in https://github.com/h2oai/h2o-3/pull/3577/commits/391a7ef949c4c82321e1d0cf31f1c227a7c5d096

h2o-core/src/main/java/water/Keyed.java
```diff
@@ -14,11 +16,42 @@
   /** Remove this Keyed object, and all subparts; blocking. */
   public final void remove( ) { remove(new Futures()).blockForPending(); }
   /** Remove this Keyed object, and all subparts.  */
+
+  /**
```
michalkurka: My first thought is we shouldn't proliferate this mechanism to this level.

michalkurka:
I would prefer this to be implemented outside of core-level classes. There is some code duplication that I don't think is unavoidable.

There essentially already has a function, that can be used:

    public static void deleteTempFrameAndItsNonSharedVecs(Frame tempFrame, Frame baseFrame) {
      Key[] keys = tempFrame.keys();
      for( int i=0; i<keys.length; i++ )
        if( baseFrame.find(keys[i]) == -1 ) //only delete vecs that aren't shared
          keys[i].remove();
      DKV.remove(tempFrame._key); //delete the frame header
    }

This function is not high-perf but that is fine for now, we can fix improve that. The point is in sharing the logic.

Perhaps we could do this:
1. collect all vecs from Frames we want to keep into a single Frame (that is a convenient wrapper) = sharedVecsFrame
2. call `deleteTempFrameAndItsNonSharedVecs(f, sharedVecsFrame)` on each Frame `f` we don't want to keep

michalkurka:
@pscheidl Your approach works (!!!) but IMHO the use case we are solving is not usual and we should not introduce a very generic mechanism (Keyed class) to solve a problem that is not common.

WDYT?

Pscheidl:
Agreed @michalkurka, I'll move it elsewhere.

Pscheidl:
@michalkurka I'd like to propose a new approach that lies somewhere in the middle. As the original logic in `deleteTempFrameAndItsNonSharedVecs` method used linear search, resulting in at least `O(n^2)` as it was doing the search through all the `baseFrame`'s vectors ....  I re-used my `retain` method, which is now called from `deleteTempFrameAndItsNonSharedVecs`.

My approach creates a set and then checks the matches, resulting in `O(2n)` - better.

Completely removed the changes from `Keyed` - this is undone.

Please see if you like it. This also enables me not to create one master `Frame` to rule them all, containing all the retained vecs.

h2o-algos/src/test/java/hex/ClearDKVTaskTest.java
```diff
@@ -98,6 +97,113 @@ public void testGBM() {
         }
     }
 
+    /**
+     * @author Michal Kurka
```
maurever: Is that still true? ;)

h2o-core/src/main/java/water/Keyed.java
```diff
@@ -3,6 +3,8 @@
 import water.api.schemas3.KeyV3;
 import water.fvec.*;
 
+import java.util.Set;
```
maurever: Unused import?

Pscheidl:
Found a bug. After the model's training frames are added to the list of retained elements, those frames must have their inner keys extracted and saved in the retained elements set as well. Currently we don't test such scenario, and I'll add the fix and a test for it.

Pscheidl:
I guess the test failure might have something to do with values not being available on every node.

h2o-core/src/main/java/water/Keyed.java
```diff
@@ -3,6 +3,8 @@
 import water.api.schemas3.KeyV3;
```
michalkurka: no changes needed, can it be removed from this PR?

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -71,14 +71,14 @@
...
-    for( int i=0; i<keys.length; i++ )
-      if( baseFrame.find(keys[i]) == -1 ) //only delete vecs that aren't shared
-        keys[i].remove();
-    DKV.remove(tempFrame._key); //delete the frame header
+    Set<Key> baseFramekeys = new HashSet<>(baseFrame._keys.length);
+    for (int i = 0; i < baseFrame._keys.length; i++) {
+      baseFramekeys.add(baseFrame._keys[i]);
+    }
+    
+    tempFrame.retain(new Futures(), baseFramekeys);
```
michalkurka: This usually leads to very strange behavior, race condition and failing tests. You do need to wait for completion of Futures _unless_ it is a responsibility of the caller. In this case it is not.

I suggest to leave this method as-is, and use the new method when it proves itself.

michalkurka:
@Pscheidl thanks for the changes! I have a couple of suggestions:

1. I wouldn't modify `deleteTempFrameAndItsNonSharedVecs` - your version has a bug and there are no tests, lets keep this PR laser-focused on one thing.
2. (Minor) please remove unnecessary formatting changes from Keyed.
3. (Opinion) I think your ClearDKVTask has outgrown the DKV class, I suggest to move it to its own class.

h2o-core/src/main/java/water/ExtractNestedKeysTask.java
```diff
@@ -0,0 +1,22 @@
+package water;
+
+public class ExtractNestedKeysTask extends MRTask<ExtractNestedKeysTask> {
```
Pscheidl: Please ignore.

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -0,0 +1,9 @@
+package water;
+
+public class DKVManager {
+  
+  public void retain(final Key[] retainedKeys){
```
Pscheidl: Will be called via a function, this is just a helper during development right now, please ignore. The classes will look nicer in the end.

h2o-core/src/main/java/water/ClearDKVTask.java
```diff
@@ -0,0 +1,106 @@
...
+    final Collection<Value> storeKeys = H2O.STORE.values();
+    Futures removalFutures = new Futures();
+    System.out.println(Arrays.toString(retainedKeys.toArray()));
+    for (final Value value : storeKeys) {
+      if (retainedKeys.contains(value._key)) {
+        continue;
+      }
+      if(value.isNull()) continue;
+
+      if (value.isFrame()) {
```
Pscheidl: Note: Multiple nodes will theoretically call remove operation of one key is on multiple nodes.

h2o-core/src/main/java/water/ClearDKVTask.java
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * @param retainedKeys A {@link Set} of retained keys to insert the extracted {@link Frame} and {@link Model} keys to.
+   *                     Should contain user-specified keys to retain in order to extract anything.
+   */
+  private static void extractNestedKeys(final Set<Key> retainedKeys) {
+
+    final Iterator<Key> keysIterator = retainedKeys.iterator(); // Traverse keys provided by the user only.
+    while (keysIterator.hasNext()) {
+      final Key key = keysIterator.next();
+      final Value value = Value.STORE_get(key);
```
Pscheidl: Not all keys have values locally, which is the main point of H2O's existence. This should really be done otherwise - gather all keys to be retained from all nodes in a separate MR Task and then invoke deletion.

This leads into situations where in multi-node (cluster of 5), some keys do get deleted and some tests will fail.

h2o-algos/src/test/java/hex/ClearDKVTaskTest.java
```diff
@@ -98,8 +97,167 @@ public void testGBM() {
...
+      // Training frame of the model should be retained as well
+      final Value trainVal = DKV.get(trainingFrame._key);
+      assertNotNull(trainVal);
+      assertTrue(trainVal.isFrame());
+      assertFalse(trainVal.isDeleted());
+      assertNotNull(trainVal.get()); // The training frame should be there
+
+      // No shared vecs were deleted from the training frame as the sharedFrame was deleted.
+      for (Key k : trainingFrame.keys()) {
+        assertNotNull(DKV.get(k));
```
Pscheidl: This is an enhancement - actually get the frame's vec keys from DKV and see if they're truly there.

Pscheidl:
I completely re-visited the code.

The original proposal had one issue - nested key extraction could not work only locally. Key extraction required global information - doable in two ways. One was introducing a new `MRTask` and then collecting all the nested keys found, but DKV.get in the loop does the same thing, without modifying `MRTask`. So I chose this way of the warrior. :crossed_swords:  This part is done on the master node and there is no MRTask required.

Once the complete set of keys to be retained is collected, each node receives it and compares it to it's local cache (store). And deletes anything required locally. This part is distributed and represented by `ClearDKVTask` in `DKVManager` class.

Also, I've decided to change the API, as exposing the MRTask itself is unnecessary. There is a more high level method in DKVManager named `retain` doing just that.

michalkurka:
please fix references to local file paths, eg.:

            trainingFrame = parse_test_file("/home/pavel/wrk/h2o-3/smalldata/iris/iris_wheader.csv");

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -0,0 +1,122 @@
...
+    final Iterator<Key> keysIterator = retainedKeys.iterator(); // Traverse keys provided by the user only.
+    while (keysIterator.hasNext()) {
+      final Key key = keysIterator.next();
+
+      final Value value = DKV.get(key);
+
+      if (value == null || value.isNull()) {
+        continue; // The key might not have value cached locally. Do not delete.
+      } else if (!value.isFrame() && !value.isModel()) {
+        Log.info(String.format("Given key %s is of type %d. Not retaining.", key.toString(), value.type()));
```
michalkurka: perhaps a warning would be better?

michalkurka: or should we just fail?

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if (model._parms._valid != null) {
+      retainedKeys.add(model._parms._valid);
+      extractFrameKeys(retainedKeys, model._parms._valid.get());
+    }
+  }
+
+  private static class ClearDKVTask extends MRTask<ClearDKVTask> {
+
+    private final Key[] _retainedKeys; // Only model and frame keys
```
michalkurka: is this true? how about the shared vecs?

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: We might as well fail and the user will know what he or she did wrong.

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: No longer true, old commentary. Updated the commentary. Resolved in https://github.com/h2oai/h2o-3/pull/3577/commits/3786ef7de44574103c72247aba21c48e07126e67

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: Resolved in https://github.com/h2oai/h2o-3/pull/3577/commits/3786ef7de44574103c72247aba21c48e07126e67

Pscheidl: Also added a test https://github.com/h2oai/h2o-3/pull/3577/commits/ee15d8a864bc3ec558cbb10329e6d38cb5aa781f

h2o-algos/src/test/java/hex/DKVManagerTest.java
```diff
@@ -0,0 +1,346 @@
...
+      assertNull(DKV.get(frame._key));
+
+    } finally {
+      if (frame != null) frame.delete();
+    }
+  }
+
+    private static void testRetainFrame(Frame trainingFrame) {
+      DKVManager.retain(new Key[]{trainingFrame._key});
+      assertTrue(H2O.STORE.containsKey(trainingFrame._key));
```
michalkurka: IMHO not entirely correct, please use DKV.get instead

h2o-core/src/main/java/water/MRTask.java
```diff
@@ -240,6 +240,7 @@ private Frame closeFrame(Key key, String[] names, String[][] domains, Futures fs
     }
     return new Frame(key,names,vecs);
   }
+  public void map(){}
```
michalkurka: not needed anymore

michalkurka:
ok, please undo changes in Keyed and MRTask - they are not needed anymore

michalkurka:
otherwise looks good!

h2o-algos/src/test/java/hex/DKVManagerTest.java
Pscheidl: Fixed in https://github.com/h2oai/h2o-3/pull/3577/commits/4fae893626b410e4c67fcdb31261cbb6982b038c, multiple occurances.

h2o-core/src/main/java/water/MRTask.java
Pscheidl: Removed in https://github.com/h2oai/h2o-3/pull/3577/commits/70b82635533f9972649766784ef7d404d459a6ed

PR #3579, opened 2019-06-09
PUBDEV-6556 - S3 tests failing non-deterministically in hadoop pipeline

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6556

## Changes
This test enhancement does two things:
- Uses millisecond precision in file name to avoid collisions
- Before H2O.import is called, it tries to verify the object is present in S3. If not, it will try each 2 seconds, up to 10 times, then it times out with an exception.

Tests are running here: http://mr-0xc1:8080/view/H2O-3/job/h2o-3-hadoop-smoke-pipeline/job/pavel%252Fpubdev-6556/1/

## To-do 
Check S3 storage for file deletion ! (should work after different PR, but must be checked)

### DIFF
--- a/h2o-hadoop-3/tests/python/pyunit_s3_import_export.py
+++ b/h2o-hadoop-3/tests/python/pyunit_s3_import_export.py
@@ -12,7 +12,7 @@ import boto3
 def s3_import_export():
     local_frame = h2o.import_file(path=pyunit_utils.locate("smalldata/logreg/prostate.csv"))
     for scheme in ["s3a"]:  # s3n is deprecated since HDP3/CDH6
-        timestamp = datetime.today().utcnow().strftime("%Y%m%d-%H%M%S")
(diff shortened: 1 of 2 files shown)


h2o-hadoop-3/tests/python/pyunit_s3_import_export.py
```diff
@@ -11,17 +11,27 @@
 
 def s3_import_export():
     local_frame = h2o.import_file(path=pyunit_utils.locate("smalldata/logreg/prostate.csv"))
-    for scheme in ["s3a"]: # s3n is deprecated since HDP3/CDH6
-        timestamp = datetime.today().utcnow().strftime("%Y%m%d-%H%M%S")
+    for scheme in ["s3a"]:  # s3n is deprecated since HDP3/CDH6
```
Pscheidl: PEP 8 space

Pscheidl:
`Failed to extract h2o-3-stash-hadoop.tar.gz` - test failure unrelated, merging.

PR #3582, opened 2019-06-10
PUBDEV-6565: Enable XGBoost Junit tests in multi-node

michalkurka:
This PR introduces a new stage for XGBoost Multinode Junit tests. The purpose of having it separate for now is to have very good visibility that these tests are actually run and to be able to quickly fix remaining issues. I had to ignore 4 tests in XGBoostTest, these tests should be fixed as a part of PUBDEV-6565.

@michal-raska please look into the changes in the makefile, especially those that modify the artifacts to be archived (without these changes the test directories were empty).

### DIFF
--- a/scripts/jenkins/Makefile.jenkins
+++ b/scripts/jenkins/Makefile.jenkins
@@ -262,7 +262,7 @@ test-junit-jenkins:
 	@$(MAKE) -f $(THIS_FILE) test-junit
 
 test-junit:
-	./gradlew test -x :h2o-ext-mojo-pipeline:test -x h2o-automl:test $$ADDITIONAL_GRADLE_OPTS
+	./gradlew test -x h2o-ext-mojo-pipeline:test -x h2o-automl:test -x h2o-ext-xgboost:testMultiNode $$ADDITIONAL_GRADLE_OPTS
 
 test-junit-automl-jenkins:
 	$(call sed_test_scripts)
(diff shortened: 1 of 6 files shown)


michal-raska:
Thanks @michalkurka!

scripts/jenkins/Makefile.jenkins
```diff
@@ -271,26 +273,33 @@ test-junit-automl-jenkins:
...
-	./gradlew test -x h2o-scala_2.11:test -x h2o-scala_2.10:test -x :h2o-ext-mojo-pipeline:test -x h2o-automl:test $$ADDITIONAL_GRADLE_OPTS
+	./gradlew test -x h2o-scala_2.11:test -x h2o-scala_2.10:test -x h2o-ext-mojo-pipeline:test -x h2o-automl:test -x h2o-ext-xgboost:testMultiNode $$ADDITIONAL_GRADLE_OPTS
 
 test-junit-12-jenkins:
 	$(call sed_test_scripts)
 	@$(MAKE) -f $(THIS_FILE) test-junit-12
 
 test-junit-12:
-	./gradlew test -x h2o-scala_2.11:test -x h2o-scala_2.10:test -x :h2o-ext-mojo-pipeline:test -x h2o-automl:test $$ADDITIONAL_GRADLE_OPTS
+	./gradlew test -x h2o-scala_2.11:test -x h2o-scala_2.10:test -x :h2o-ext-mojo-pipeline:test -x h2o-automl:test -x h2o-ext-xgboost:testMultiNode $$ADDITIONAL_GRADLE_OPTS
```
honzasterba: I dont like this insonsistency of once using :h2o-ext-mojo-pipeline and once calling it h2o-ext-mojo-pipeline (its both the same, but inconsistent)

scripts/test-xgb-gpu-smoke.sh
```diff
@@ -21,7 +21,7 @@ function finish {
     set -e
 }
 
-./gradlew h2o-ext-xgboost:test &
+./gradlew h2o-ext-xgboost:test -x :h2o-ext-xgboost:testMultiNode &
```
honzasterba: unnecessary colon
also the exclude of  h2o-ext-xgboost:multi seems unecessary as it depends on h2o-ext-xgboost:test and shoudl be trainsiently excluded no?

scripts/jenkins/Makefile.jenkins
michalkurka: thanks, will fix!

scripts/test-xgb-gpu-smoke.sh
michalkurka: `h2o-ext-xgboost:test` depends on `h2o-ext-xgboost:testSingleNode` and`h2o-ext-xgboost:testMultiNode` and others, i'm excluding multinode because i don't think we need to eun multinode in xgb smoke tests

honzasterba: oh sorry, I misunderstood the line, makes sense

michalkurka:
Got verbal approval by @honzasterba, merging

PR #3586, opened 2019-06-11
[PUBDEV-6573] enable java scoring for xgboost MOJOs by default

honzasterba:


h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostMojoReader.java
```diff
@@ -3,13 +3,10 @@
...
 
 import java.io.IOException;
-import java.util.Arrays;
 
-/**
- */
 public class XGBoostMojoReader extends ModelMojoReader<XGBoostMojoModel> {
   
-  public static final String SCORE_JAVA_PROP = "sys.ai.h2o.xgboost.scoring.java.enable";
+  public static final String SCORE_JAVA_PROP = "sys.ai.h2o.xgboost.scoring.java.enable  ";
```
michalkurka: typo

michalkurka:
Can you confirm this will help when none of the SO can be loaded by the system? It seems to me that we might fail on loading the native library even if we don't get to use it because `XGBoostNativeMojoModel` is a dependency on the classloading level (and classloading will trigger static field initialization).

honzasterba:
I have verified that native lib loading is not triggered until XGBoostNativeMojoModel is instantiated

michalkurka:
thanks

imho, the only issue I see that this is going to the past - older mojos will start scoring with java-predictor; this was not the case before because we were only using java-predictor with when booster was set to `gbtree` (older mojos won't have this property)

michalkurka:
@honzasterba I suggest to target this PR for a major release and add compatibility tests for XGBoost MOJO: https://github.com/h2oai/mojoland before merging

We need to make sure the predictions will be the same for all boosters. `gbtree` is tested thoroughly but the same cannot be said about dart and gblinear. IMHO if we want to flip the default, we do need to extend our test suit first.

honzasterba:
Mojoland updated in a separate PR
Added unit test that verifies that java and native predicions return same results. DART native predictions however return undeterministic results and as such had to be excluded from the unit test

h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostMojoReader.java
```diff
@@ -37,29 +34,21 @@ protected XGBoostMojoModel makeModel(String[] columns, String[][] domains, Strin
     } catch (IOException e) {
       throw new IllegalStateException("MOJO is corrupted: cannot read the serialized Booster", e);
     }
-    Boolean conf = getJavaScoringConfig();
-    if (useJavaScoring(conf)) {
+    boolean javaScoringByDefault = readkv("use_java_scoring_by_default", false);
```
michalkurka: this changes the behavior, old MOJOs that were predicting with java-predictor because they were using gbtree will start predicting with native

honzasterba: good catch, fixed

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/DefaultMojoImplTest.java
```diff
@@ -91,4 +114,27 @@ private static XGBoostModel trainModel(Booster booster) {
     return model;
   }
 
+  /**
+   * Old mojo writer that does not include use_java_scoring_by_default
+   */
+  private static final class OldSerializationXGBMojoWriter extends XGBoostMojoWriter {
```
michalkurka: can we just commit a small old mojo in the repository instead of generating it like this?

h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostMojoReader.java
```diff
@@ -37,29 +34,26 @@ protected XGBoostMojoModel makeModel(String[] columns, String[][] domains, Strin
...
-    String booster = readkv("booster");
-    return "gbtree".equals(booster); // use java scoring for `gbtree` (well tested); `native` for dart & gblinear
-  }
-
-  public static Boolean getJavaScoringConfig() {
+  public static boolean useJavaScoring(Boolean mojoValue, String booster) {
     String javaScoringEnabled = System.getProperty(SCORE_JAVA_PROP);
     if (javaScoringEnabled == null) {
-      return null;
+      if (mojoValue != null) {
```
michalkurka: thanks for the fix

h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostMojoReader.java
```diff
@@ -37,29 +34,26 @@ protected XGBoostMojoModel makeModel(String[] columns, String[][] domains, Strin
...
-    if (conf != null) {
-      // user set the property - respect the decision
-      return conf;
-    }
-    String booster = readkv("booster");
-    return "gbtree".equals(booster); // use java scoring for `gbtree` (well tested); `native` for dart & gblinear
-  }
-
-  public static Boolean getJavaScoringConfig() {
+  public static boolean useJavaScoring(Boolean mojoValue, String booster) {
```
michalkurka: imho we can call it mojoPreferredJavaScoring (or something like that ), I wasn't sure what mojoValue meant when I saw this method

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostPredictImplComparisonTest.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+    
+    private double getDelta() {
+        if ("gbtree".equals(booster)) {
+            return 1e-10;
+        } else if ("dart".equals(booster)) {
+            return 1e-10;
+        } else {
+            return 1e-3;
```
honzasterba: we should know why there is a difference

h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostMojoReader.java
honzasterba: fixed

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/DefaultMojoImplTest.java
honzasterba: I kinda like this approach a little more because the difference in mojo is visible from code

maurever: I also prefer to add old mojo in the repository. The difference is visible from code; however, when someone adds some new parameter, and he/she would like to test it, he/she would have to implement a new object where only one parameter will be different, etc...

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -670,11 +670,22 @@ public static Vec uvec(UUID...rows) {
   public static <T> T[] aro(T ...a) { return a ;}
 
   // ==== Comparing Results ====
+  
+  public static void assertFrameEquals(Frame expected, Frame actual, double delta) {
```
maurever: :+1:

PR #3591, opened 2019-06-13
PUBDEV-6584: Fixing XSS headers, raising errors in Flow

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6584


h2o-core/src/main/java/water/server/ServletUtils.java
```diff
@@ -188,7 +188,7 @@ public static void setCommonResponseHttpHeaders(HttpServletResponse response, fi
     response.setHeader("X-h2o-context-path", sanatizeContextPath(H2O.ARGS.context_path));
     // Security headers
     response.setHeader("X-Frame-Options", "deny");
-    response.setHeader("X-XSS-Protection", "X-XSS-Protection: 1; mode=block");
```
michalkurka: omg, good catch - this was a copy&paste error

PR #3592, opened 2019-06-13
PUBDEV-6112 StackEnsemble verification of fold column presence in training/validation dataset

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6112

## The cause

The exception provided by the reporter in the JIRA can be reproduced by training a GBM model on a dataset with fold column specified (and of course, the CV models kept). The `GBMModel` parameters are:

```java
      GBMModel.GBMParameters parameters = new GBMModel.GBMParameters();
      parameters._train = trainingFrame._key;
      parameters._fold_column = "class";
      parameters._seed = 0xFEED;
      parameters._response_column = "petal_len";
      parameters._ntrees = 1;
      parameters._keep_cross_validation_predictions = true;
```

 Afterwards, the `StackEnsembleModel` is created with the previously mentioned base GBM model inside:

```java
      final StackedEnsembleParameters seParams = new StackedEnsembleParameters();
      seParams._train = partialFrame._key;
      seParams._response_column = "petal_len";
      seParams._metalearner_algorithm = Algorithm.AUTO;
      seParams._base_models = new Key[]{gbmModel._key};
      seParams._seed = 0xFEED;
      seParams._metalearner_fold_column = "class";
```

In `StackEnsemble` class, once line `123` (sic!) is reached:

```java
      if (_model._parms._metalearner_fold_column != null) {
        Vec foldColumn = actuals.vec(_model._parms._metalearner_fold_column);
        levelOneFrame.add(_model._parms._metalearner_fold_column, foldColumn);
      }
```

the property `foldColumn` is filled with null, as the fold column does not exist in the training dataset, but it did exist in the training dataset of the GBM. This might also happen during validation.

## Resolution

This is considered to be user's fault. However, we should prevent nullpointer once the user provides dataset with missing fold column (or renamed). That is why an exception with human-readable error message about fold column missing (while printing the columns available) is thrown.

This is the case of the JIRA reporter, as he's constantly changing the data and chances are the validation data frame he provided does not have the right column (or with a bad name). His example is not runnable (demo only) and we do not have access to the data the reporter used.

## Additional changes

- The StackEnsemble driver's `computeImpl` method is not very fault tolerant. It creates a model at the beginning and does not delete the model if there is an exception inside, effectively creating a key leak. The test constructed tests this as well (as a side effect). A mechanism to remove the model created and to rethrow the exception afterwards has been introduced. See my comments.

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -70,9 +70,7 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
   public void init(boolean expensive) {
     super.init(expensive);
 
-    checkFoldColumnPresent(_parms._metalearner_fold_column, _parms._train != null ? _parms._train.get() : null,
-            _parms._valid != null ? _parms._valid.get() : null,
-            _parms._blending != null ? _parms._blending.get() : null);
+    checkFoldColumnPresent(_parms._metalearner_fold_column, train(), valid(), _parms.blending());
   }
 
   /**
@@ -87,8 +85,8 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
     for (Frame frame : frames) {
       if (frame == null) continue; // No frame provided, no checks required
       if (frame.vec(foldColumnName) == null) {
-        throw new IllegalArgumentException(String.format("Specified fold column '%s' not found in the data frame: '%s'. Available column names are: %s",
-                foldColumnName, frame._key, Arrays.toString(frame.names())));
+        throw new IllegalArgumentException(String.format("Specified fold column '%s' not found in one of the supplied data frames. Available column names are: %s",
+                foldColumnName, Arrays.toString(frame.names())));
       }
     }
   }
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -16,7 +16,6 @@ import water.fvec.Vec;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Objects;
 
 import water.util.ArrayUtils;
 import water.util.Log;
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -121,6 +121,10 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
       // Add metalearner_fold_column to level one frame if it exists
       if (_model._parms._metalearner_fold_column != null) {
         Vec foldColumn = actuals.vec(_model._parms._metalearner_fold_column);
+        if (foldColumn == null) {
+          throw new IllegalArgumentException(String.format("Specified fold column '%s' not found in the data frame: '%s'. Available column names are: %s",
+                  _model._parms._metalearner_fold_column, actuals._key, Arrays.toString(actuals.names())));
+        }
         levelOneFrame.add(_model._parms._metalearner_fold_column, foldColumn);
       }
       // Add response column to level one frame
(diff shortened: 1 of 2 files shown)


h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -214,51 +218,56 @@ public void computeImpl() {
     init(true);
     if (error_count() > 0)
         throw H2OModelBuilderIllegalArgumentException.makeFromBuilder(StackedEnsemble.this);
+      try {
```
Pscheidl: No changes in this block, only the try-catch block has been introduced, thus the code inside is shifted to the right.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -214,51 +218,56 @@ public void computeImpl() {
...
+                  metalearnerSeed
+          );
+          metalearner.compute();
+        } else {
+          throw new H2OIllegalArgumentException("Invalid `metalearner_algorithm`. Passed in " + metalearnerAlgoSpec +
+                  " but must be one of " + Arrays.toString(Metalearner.Algorithm.values()));
+        }
+      } catch (Throwable t){
+        // If there is an exception during model training or validation, remove the model before exiting
+        _model.remove();
```
Pscheidl: Remove the model if there is ANY error/exception inside and then rethrow it. Preventing leaks.

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
```diff
@@ -879,4 +885,56 @@ private void test_SE_scoring_with_missing_response_column(Algorithm algo) {
...
+      assertNotNull(gbmModel);
+      
+
+      final StackedEnsembleParameters seParams = new StackedEnsembleParameters();
+      seParams._train = partialFrame._key;
+      seParams._response_column = "petal_len";
+      seParams._metalearner_algorithm = Algorithm.AUTO;
+      seParams._base_models = new Key[]{gbmModel._key};
+      seParams._seed = 0xFEED;
+      seParams._metalearner_fold_column = "class";
```
seb-h2o: @Pscheidl so to reproduce this use-case, the user needs to explicitly specify a `_metalearner_fold_column` that doesn't exist in the `train` pr `blending` dataset? 
Then, I'd rather suggest to validate it in the same place as we check that it matches base models `fold_column`, that is in `StackedEnsemble.checkAndInheritModelProperties()`: having parameters validation spread around the code base gets quickly confusing.

seb-h2o: Although, changing my mind, may be better in `prepareLevelOneFrame` you're right, as it should be checked for `validation` and `blending` frames as well. Yeah, our parameters validation sucks.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
michalkurka: there should be a try-finally,

this is an example of a correct way of doing that: https://github.com/h2oai/h2o-3/blob/master/h2o-algos/src/main/java/hex/tree/SharedTree.java#L223

please note that the outer `Model model` is not set until we actually acquire the lock, otherwise we would try to unlock a Model that was not locked in the finally block

michalkurka: IMHO the catch block shouldn't be there, no other Drivers have that as @seb-h2o pointed out

there should also not be any parameter validation, that should be done in init phase before the model training is already in progress, I am referring to line 264

seb-h2o: @michalkurka : yeah agree, there are already many `H2OIllegalArgumentException` all over the place in this class: I'd suggest to create a separate refactoring ticket though to make proper parameters validation for SE: `StackedEnsemble` needs to override `init` and do all validation there, and probably call `StackedEnsembleModel#checkAndInheritModelProperties` during this validation.

Pscheidl: I wanted to override init. I'll move the code according to the comments above. Makes sense. I'll do just the quickfix. @seb-h2o Please create the JIRA if you want to, I'll leave the refactoring to the AutoML team.

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
Pscheidl: Moved to init phase. WDYT ?

Pscheidl:
@seb @michalkurka Moved to init phase. Wrote test for each separately. 

Left the try-finally (removed catch block) block in place.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -66,6 +66,28 @@ public boolean haveMojo() {
...
+
+    if (_parms._metalearner_fold_column != null && _parms._valid != null) {
+      final Frame validationFrame = _parms._valid.get();
+      if(validationFrame.vec(_parms._metalearner_fold_column) == null) {
+        throw new IllegalArgumentException(String.format("Specified fold column '%s' not found in the validation data frame: '%s'. Available column names are: %s",
+                _parms._metalearner_fold_column, validationFrame._key, Arrays.toString(validationFrame.names())));
+      }
+    }
+    
+  }
```
seb-h2o: @Pscheidl nice thanks! Can you please also add similar logic for `_parms._blending`?

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
seb-h2o: https://0xdata.atlassian.net/browse/PUBDEV-6588

seb-h2o: @Pscheidl I don't understand the model removal in the finally block: it means it's gonna be removed whatever happens.
Did you mean `unlock` as @michalkurka  suggested?

Pscheidl: @seb-h2o Oh yes ! Of course ! Of course, I never meant to do that. Sorry and thank you for discovering it. Resolved in https://github.com/h2oai/h2o-3/pull/3592/commits/d2f6a14814817829fa52dd82733a2e6fc2e80d30

Pscheidl: @seb-h2o I actually decided to completely revert the changes in compute-impl (try-finally block was the only change). Are you ok with it ? It's not in the scope of this PR and it cause some tests to fail. I'd rather avoid it. My goal is to introduce some validation and leave the rest of the refactorings for later.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -66,6 +66,33 @@ public boolean haveMojo() {
     return true;
   }
 
+  @Override
+  public void init(boolean expensive) {
+    super.init(expensive);
+
+    checkFoldColumnPresent(_parms._metalearner_fold_column, _parms._train != null ? _parms._train.get() : null,
```
michalkurka: This is IMHO not entirely correct. You are not actually validating the frame that the algorithm will use, you are validating the original raw (unadapted) frame.

I think at this point you should be using the `train()`/`valid()` frames otherwise the init fail will pass but algo will fail when it tries to train the metalearner (it will not see any fold column and will fail on validation).

I added a commit with a failing test that demonstrates that.

Pscheidl: Oh, I dod not realize I'm not using adapted frames. :space_invader: So I did fix this for validation & training dataframe. Blending frame is taken as-is from the params, as the blending driver does the same.

You can see the changes in https://github.com/h2oai/h2o-3/pull/3592/commits/7a13e95c7c1d341653727eee10beb8ab9d34cc71

Please note I've deleted the frame key from the error mesasge, as the adapated temporary frames do not have a key and thus there would be null for train/valid.

PR #3594, opened 2019-06-13
PUBDEV-6582: Added Conda installation instructions for download

zainhaq-h2o:
Added Conda installation instructions to the downloads page


michalkurka:
@zainhaq-h2o very cool, thank you!

lets merge and check on nightly builds if it works as expected before closing the corresponing jira

PR #3596, opened 2019-06-13
PUBDEV-6569: AutoML, set `score_tree_interval` for DRF+XRT

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6569

Also added a test verifying that parameters required for reproducibility are set or propagated by default.
When writing this test, discovered that same seed was always reused for hardcoded GBM and XGB, so, as it was not designed to behave this way, made it work by fixing the AutoML part of https://0xdata.atlassian.net/browse/PUBDEV-6385 (ie cloning parameters instead of reusing them).

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -684,7 +684,9 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     builder._parms = parms;
 
     setCommonModelBuilderParams(builder._parms);
+    setStoppingCriteria(builder._parms, defaults, true);
 
+    // override model's max_runtime_secs to ensure that the total max_runtime doesn't exceed expectations
     if (ignoreLimits)
       builder._parms._max_runtime_secs = 0;
     else if (builder._parms._max_runtime_secs == 0)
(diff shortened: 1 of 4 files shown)


ledell:
> same seed was always reused for hardcoded GBM and XGB. 
@seb-h2o  Good catch.  I'll just make a comment here on how you changed it for reference -- rather than re-using the same seed for each GBM (or RF), we increment the seed by 1 (in a predictable fashion so that we still get reproducibility).  Thanks.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -902,31 +899,31 @@ void defaultXGBoosts(boolean emulateLightGBM) {
...
+    commonXGBoostParameters._distribution = getResponseColumn().isBinary() && !(getResponseColumn().isNumeric()) ? DistributionFamily.bernoulli
                     : getResponseColumn().isCategorical() ? DistributionFamily.multinomial
                     : DistributionFamily.AUTO;
 
-    xgBoostParameters._score_tree_interval = 5;
-    xgBoostParameters._stopping_rounds = 5;
-//    xgBoostParameters._stopping_tolerance = Math.min(1e-2, RandomDiscreteValueSearchCriteria.default_stopping_tolerance_for_frame(this.trainingFrame));
+    commonXGBoostParameters._score_tree_interval = 5;
+    commonXGBoostParameters._stopping_rounds = 5;
+//    commonXGBoostParameters._stopping_tolerance = Math.min(1e-2, RandomDiscreteValueSearchCriteria.default_stopping_tolerance_for_frame(this.trainingFrame));
```
deil87: @seb Can we remove these legacy comments?

PR #3597, opened 2019-06-14
PUBDEV-6530: Add an option to export CSV compressed by gzip

michalkurka:
https://0xdata.atlassian.net/browse/PUBDEV-6530

### DIFF
--- a/h2o-core/src/main/java/water/util/CompressionFactory.java
+++ b/h2o-core/src/main/java/water/util/CompressionFactory.java
@@ -1,0 +1,57 @@
+package water.util;
+
+import water.Iced;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.reflect.Constructor;
+import java.util.zip.GZIPOutputStream;
+
+public class CompressionFactory extends Iced<CompressionFactory> {
+
+  private String _name;
+
+  private CompressionFactory(String name) {
+    _name = name;
+  }
+
+  OutputStream wrapOutputStream(OutputStream os) throws IOException {
+    final String n = _name.toLowerCase();
+    switch (n) {
+      case "gzip":
+        return new GZIPOutputStream(os);
+      case "bzip2":
+        return wrapDynamic("org.python.apache.commons.compress.compressors.bzip2.BZip2CompressorOutputStream", os);
+      case "snappy":
+        return wrapDynamic("org.xerial.snappy.SnappyOutputStream", os);
+      default:
+        return wrapDynamic(_name, os);
+    }
+  }
+
+  private OutputStream wrapDynamic(String className, OutputStream os) {
+    try {
+      Class<?> cls = Class.forName(className);
+      Constructor<?> constructor = cls.getConstructor(OutputStream.class);
+      return (OutputStream) constructor.newInstance(os);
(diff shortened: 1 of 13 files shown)


michalkurka:
@Pscheidl still WIP - missing junits, please review the concept

h2o-core/src/main/java/water/util/CompressionFactory.java
```diff
@@ -0,0 +1,57 @@
...
+
+  private String _name;
+
+  private CompressionFactory(String name) {
+    _name = name;
+  }
+
+  OutputStream wrapOutputStream(OutputStream os) throws IOException {
+    final String n = _name.toLowerCase();
+    switch (n) {
```
Pscheidl: I'd suggest creating enumeration instead of string passing. The validation of an enum is also easier. In the handler, let's just use enum's `valueOf()` method to convert user-provided string to ENUM on the backend.

h2o-core/src/main/java/water/util/CompressionFactory.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.OutputStream;
+import java.lang.reflect.Constructor;
+import java.util.zip.GZIPOutputStream;
+
+public class CompressionFactory extends Iced<CompressionFactory> {
+
+  private String _name;
+
+  private CompressionFactory(String name) {
+    _name = name;
```
Pscheidl: Enum would provide us more safety here and would make sure the object is properly constructed, before failing later during a method call.

h2o-core/src/main/java/water/util/CompressionFactory.java
michalkurka: It supports more than just an enumerated set of values, you can specify any class name and it will try to use that. This way users are not limited to what we've baked in.

wdyt?

h2o-core/src/main/java/water/util/CompressionFactory.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  OutputStream wrapOutputStream(OutputStream os) throws IOException {
```
Pscheidl: This method could perfectly be static, with an enum of the compression algorithm. No need to instantiate.

h2o-core/src/main/java/water/util/CompressionFactory.java
```diff
@@ -0,0 +1,57 @@
...
+  
+  private void checkAvailability() {
+    try {
+      wrapOutputStream(new ByteArrayOutputStream());
+    } catch (IOException e) {
+      throw new IllegalStateException("Initialization failed for compression method " + _name, e);
+    }
+  }
+
+  public static CompressionFactory make(String name) {
```
Pscheidl: Please declare this method throws an exception.

h2o-core/src/main/java/water/util/CompressionFactory.java
michalkurka: It takes more than just an enum ;)

h2o-core/src/main/java/water/util/CompressionFactory.java
michalkurka: just curious: why? it is a private constructor and the `make` method prevents from returning an invalid instance; IMHO it is functionally equivalent

PR #3599, opened 2019-06-14
PUBDEV-6591: Allow to use a custom separator for CSV export

michalkurka:


h2o-core/src/main/java/water/api/FramesHandler.java
```diff
@@ -244,7 +244,8 @@ private FramesV3 doFetch(int version, FramesV3 s, boolean expensive) {
   public FramesV3 export(int version, FramesV3 s) {
     Frame fr = getFromDKV("key", s.frame_id.key());
     Log.info("ExportFiles processing (" + s.path + ")");
-    s.job = new JobV3(Frame.export(fr, s.path, s.frame_id.key().toString(), s.force, s.num_parts, s.compression));
+    Frame.CSVStreamParams csvParms = new Frame.CSVStreamParams().setSeparator(s.separator);
```
Pscheidl: can s.separator really be any char ? I guess yes, the user carries the consequences.

h2o-py/tests/testdir_misc/pyunit_export_sep.py
```diff
@@ -0,0 +1,37 @@
...
+from tests import pyunit_utils
+from os import path
+import pandas as pd
+'''
+Export file with h2o.export_file with a custom separator
+'''
+
+
+def export_custom_separator():
+    data = {'col1': [1, 2], 'col2': [3, 4]}
```
Pscheidl: I'd suggest trying values with escaped quotes.

h2o-core/src/main/java/water/api/FramesHandler.java
michalkurka: yep, we just like our input API takes any character

h2o-py/tests/testdir_misc/pyunit_export_sep.py
michalkurka: Can you explain bit more how is it related to this PR?

michalkurka:
@Pscheidl I will merge and add the new test in another PR after I get clarity on what would be its purpose.

PR #3600, opened 2019-06-17
PUBDEV-6592 fix code comparation

maurever:
Fix type error:

[2019-06-12T05:27:30.128Z] File ".../lib/python3.6/site-packages/h2o/utils/debugging.py", line 290, in find_code[2019-06-12T05:27:30.128Z] if hasattr(item, "_code") and item.code_ == code:[2019-06-12T05:27:30.128Z] File ".../python/lib/python3.6/site-packages/datatable/expr/expr.py", line 294, in _bool[2019-06-12T05:27:30.128Z] % self[2019-06-12T05:27:30.128Z] TypeError: Expression Expr:eq(Expr:col(0, 'code_'), <code object connect at 0x7fe508cf9b70, file ".../python/lib/python3.6/site-packages/h2o/h2o.py", line 58>) cannot be cast to bool.

JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6592?oldIssueView=true


maurever:
I invoked the error using @st-pasha code:

```
from datatable import f
import h2o
h2o.init(0)
```

and I found out, there is FrameProxy, which translate  `item.__code__ == code` to `ColSelectorExpr f['__code__ == code']`. The suggestion change `(item.__code__ == code) is True` throws the same error as before. 

So I tried to use `__eq__` method and switched the values to compare them. It passes after this change. But I am not sure if it is the right solution. WDYT @st-pasha ?

h2o-py/tests/testdir_jira/pyunit_NOPASS_pubdev_6592_type_error.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+from tests import pyunit_utils
+
+# NOPASS, after fixing a bug it throws a correct error.
+# Note: the try-except interrupts another error behind this error, so it can´t be used in this test.
+def test_type_error():
+    from datatable import f
+    import h2o
+    h2o.init(0)
```
st-pasha: Hmm, I don't think the test will work: 
after all, `h2o.init(0)` is supposed to throw an error, and what we are checking here is the behavior of the exception hook. Even catching the exception will not work, as we are interested in the behavior of the uncaught exception.

You could, however, launch this test in a separate python process:
```
try:
   subprocess.check_output(['python', '-c', 'from datatable import f; import h2o; h2o.init(0)'], stderr=subprocess.STDOUT)
except subprocess.CalledProcessError as e:
    text = e.output.decode()
    assert "Original exception was" not in text
    assert "cannot be cast to bool" not in text
    assert "Argument `url` should be a ?string, got integer 0" in text
    # etc...
```

st-pasha: Also, I don't know whether `datatable` is a dependency of `h2o` already, but if it isn't then it probably doesn't make sense to introduce such a dependency for the sake of a single test...
Luckily, object `f` can easily be mimicked: just create a class with overloaded `__code__` attribute which would return an object with `__eq__` method that returns another object not castable to bool.

PR #3609, opened 2019-06-20
[PUBDEV-6573] make XGBoost DART predictions deterministic

honzasterba:


h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostMojoModel.java
```diff
@@ -48,6 +48,8 @@ public static ObjectiveType fromXGBoost(String type) {
     }
   }
 
+  public String _boosterType;
```
michalkurka: IMHO these changes do warrant the MOJO version to be bumped up.

h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostNativeMojoModel.java
```diff
@@ -59,10 +61,16 @@ private static Booster makeBooster(byte[] boosterBytes) {
     try {
       dmat = new DMatrix(floats,1, floats.length, _sparse ? 0 : Float.NaN);
       final DMatrix row = dmat;
+      final int treeLimit;
+      if ("dart".equals(_boosterType)) {
```
michalkurka: Instead of doing the decision at predict time, I would just do it before this function is called.

It would make it a little bit more obvious in MOJO reader what is going on. WDYT?

honzasterba: yes it would be, but this method is static and all the context is passed via params, as such I would require a major refactor to get this done

h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostMojoModel.java
honzasterba: actually the added fields are handled in a backwards compatible way, so it is really not necessary

michalkurka: changing the version prevents accidentally using an older genmodel with a newer mojo

I think you are right here and there is no need to actually do that

michalkurka:
LGTM

PR #3610, opened 2019-06-20
Rename SVM to PSVM (naming collision in SW)

michalkurka:
FIY: @jakubhava @zainhaq-h2o


michalkurka:
Discussed with @jakubhava and @tomkraljevic

PR #3615, opened 2019-06-21
SVM: Bunch of smaller fixes - caught by Erin & Angela

michalkurka:


ledell:
I'm still a fan of changing `h2o.psvm` to `h2o.svm` but we could do that in a follow-up PR.

PR #3616, opened 2019-06-21
hotfix: disable last two examples from R package

ledell:
These two functions should also have a `\dontrun` around them because they still start up the H2O cluster upon CRAN submission (which we don't want) because of reasons cited [here](https://0xdata.atlassian.net/browse/PUBDEV-6501).  Follow-up to: https://github.com/h2oai/h2o-3/pull/3544

### DIFF
--- a/h2o-r/h2o-package/R/logging.R
+++ b/h2o-r/h2o-package/R/logging.R
@@ -5,7 +5,7 @@
 #' \dontrun{
 #' if (Sys.info()['sysname'] == "Darwin" && Sys.info()['release'] == '13.4.0') {
 #'   quit(save="no")
-#' } else{
+#' } else {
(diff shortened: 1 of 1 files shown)


h2o-r/h2o-package/R/logging.R
```diff
@@ -2,21 +2,25 @@
 #'
 #' @name aaa
 #' @examples
-#'if(Sys.info()['sysname'] == "Darwin" && Sys.info()['release'] == '13.4.0'){
-#'  quit(save="no")
-#'}else{
+#' \dontrun{
+#' if (Sys.info()['sysname'] == "Darwin" && Sys.info()['release'] == '13.4.0') {
+#'   quit(save="no")
+#' } else{
```
ledell:
```suggestion
#' } else {
```

PR #3621, opened 2019-06-22
[HOTFIX] Fix h2o.scale, H2OScaler provides "means" even for non-numerical columns

michalkurka:
It makes more sense to get the array for the whole Frame, this way
it is easier to map desired mean valus to the columns that should be
scaled.


michalkurka:
@Pscheidl more tests to come

PR #3619, opened 2019-06-22
PUBDEV-6410: Isolation Forest - early stopping on mean anomaly score

michalkurka:
This PR adds support for early stopping on mean `anomaly_score` in Isolation Forest. In order to do that, I had to refactor ScoreKeeper quite a bit: adding both @wendycwong and @maurever to the review.

PR #3622, opened 2019-06-24
PUBDEV-6613: Add stopping parameters to Isolation Forest

zainhaq-h2o:
"Isolation Forest" added to list of available algorithms for stopping_rounds, stopping_metric, and stopping_tolerance.

https://0xdata.atlassian.net/browse/PUBDEV-6613

*Do not merge until https://github.com/h2oai/h2o-3/pull/3619 has been merged

### DIFF
--- a/h2o-docs/src/product/automl.rst
+++ b/h2o-docs/src/product/automl.rst
@@ -72,7 +72,7 @@ Optional Miscellaneous Parameters
    The available options are:
 
     - ``auto``: This defaults to ``logloss`` for classification, ``deviance`` for regression, and ``anomaly_score`` for Isolation Forest. Note that custom and custom_increasing can only be used in GBM and DRF with the Python client. Must be one of: ``AUTO``, ``anomaly_score``. Defaults to ``AUTO``.
-    - ``anomaly_score``
+    - ``anomaly_score`` (Isolation Forest only)
     - ``deviance``
     - ``logloss``
     - ``mse``
--- a/h2o-docs/src/product/automl.rst
+++ b/h2o-docs/src/product/automl.rst
@@ -68,19 +68,23 @@ Optional Miscellaneous Parameters
 
 - `max_runtime_secs_per_model <data-science/algo-params/max_runtime_secs_per_model.html>`__: Specify the max amount of time dedicated to the training of each individual model in the AutoML run. Defaults to 0 (disabled). Note that setting this parameter can affect AutoML reproducibility.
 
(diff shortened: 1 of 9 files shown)


angela0xdata:
@zainhaq-h2o I just noticed this in Michal's pull request:
https://github.com/h2oai/h2o-3/pull/3619/files#diff-9cb131d06a7878d5de99508077e58223

The default value for stopping_tolerance in Isolation Forest is not the same as the other supported algos. Update the note in the stopping_tolerance Parameters Appendix entry to reflect this. You can also add this default value to the description of stopping_tolerance in the Isolation Forest chapter.

angela0xdata:
@zainhaq-h2o I forgot about this, but to be thorough, you should update the description of the `stopping_metric` option in the "Defining a Model" section in GBM, DRF, Deep Learning, AutoML, and XGBoost. What you have here already is great. Just make that same change in those other supported algorithms. The reason is that someone building a GBM model (for example) will see `anomaly_score` as an option when they run the Python or R help. This way, in the docs the restriction for IF is included.

h2o-docs/src/product/automl.rst
```diff
@@ -68,19 +68,23 @@ Optional Miscellaneous Parameters
...
 - `max_runtime_secs_per_model <data-science/algo-params/max_runtime_secs_per_model.html>`__: Specify the max amount of time dedicated to the training of each individual model in the AutoML run. Defaults to 0 (disabled). Note that setting this parameter can affect AutoML reproducibility.
 
-- `stopping_metric <data-science/algo-params/stopping_metric.html>`__: Specifies the metric to use for early stopping of the grid searches and individual models. Defaults to ``"AUTO"``.  The available options are:
+-  `stopping_metric <algo-params/stopping_metric.html>`__: Specify the metric to use for early stopping.
+   The available options are:
 
-    - ``AUTO``: This defaults to ``logloss`` for classification, ``deviance`` for regression
-    - ``deviance`` (mean residual deviance)
+    - ``auto``: This defaults to ``logloss`` for classification, ``deviance`` for regression, and ``anomaly_score`` for Isolation Forest. Note that custom and custom_increasing can only be used in GBM and DRF with the Python client. Must be one of: ``AUTO``, ``anomaly_score``. Defaults to ``AUTO``.
+    - ``anomaly_score``
```
angela0xdata: Can you change this to:
- ``anomaly_score`` (Isolation Forest only)

Same comment for other algos that use stopping_metric.

PR #3623, opened 2019-06-25
Fixing issues that caused local booklet build to fail

angela0xdata:
Fixed issues that were discovered when running .gradlew h2o-docs:booklets locally. Booklets now build successfully.


angela0xdata:
Closing this pull request because of errors. Will submit a new one.

PR #3624, opened 2019-06-25
Fixing issues that caused local booklet build to fail

angela0xdata:
Fixed issues that were discovered when running ./gradlew h2o-docs:booklets locally. Booklets now build successfully.


angela0xdata:
Closing. These changes are causing h2o-3 build to fail.

PR #3625, opened 2019-06-25
GBM MOJO model info printing/serialization POC

Pscheidl:
This is just a POC that does a lot of things at once. Should be discussed what to keep and what not to keep.


h2o-genmodel/build.gradle
```diff
@@ -28,9 +31,26 @@ dependencies {
   compile "net.sf.opencsv:opencsv:2.3"
   compile "com.google.code.gson:gson:2.6.2"
   compile "ai.h2o:h2o-tree-api:0.3.7"
+  compile group: 'com.google.protobuf', name: 'protobuf-java', version: '3.8.0'
```
michalkurka: Should this really be a compile dependency?

PR #3627, opened 2019-06-25
PUBDEV-5858: allow user to choose number of iteration for GLRM mojo scoring.

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-5858?filter=-1

1. Added glrmIterNumber into EasyPredictModelWrapper.java and PredictCSV.java
2. Added R/Python client support to add new parameter glrmIterNumber
3. Added R/Python unit tests to make sure the parameter is working.

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
@@ -68,6 +68,7 @@ public class EasyPredictModelWrapper implements Serializable {
   private final boolean enableGLRMReconstruct;  // if set true, will return the GLRM resconstructed value, A_hat=X*Y instead of just X
   private final boolean enableStagedProbabilities; // if set true, staged probabilities from tree agos are returned
   private final boolean enableContributions; // if set to true, will return prediction contributions (SHAP values) - for GBM & XGBoost
+  private final int glrmIterNumber; // allow user to set GLRM mojo iteration number in constructing x.
 
   private final PredictContributions predictContributions;
   
@@ -107,6 +108,7 @@ public class EasyPredictModelWrapper implements Serializable {
     private boolean enableGLRMReconstrut = false;
     private boolean enableStagedProbabilities = false;
     private boolean enableContributions = false;
+    private int glrmIterNumber = 100; // default set to 100
 
     /**
      * Specify model object to wrap.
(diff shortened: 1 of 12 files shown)


h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -114,6 +116,7 @@
     private boolean enableGLRMReconstrut = false;
     private boolean enableStagedProbabilities = false;
     private boolean enableContributions = false;
+    private int glrmIterNumber = 100; // default set to 1009
```
honzasterba: typo? comment in contrast whith reality commenting something that is apparent

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -75,6 +76,7 @@
   private final boolean enableGLRMReconstruct;  // if set true, will return the GLRM resconstructed value, A_hat=X*Y instead of just X
   private final boolean enableStagedProbabilities; // if set true, staged probabilities from tree agos are returned
   private final boolean enableContributions; // if set to true, will return prediction contributions (SHAP values) - for GBM & XGBoost
+  private final int glrmIterNumber; // allow user to set GLRM mojo iteration number in constructing x.
```
honzasterba: do we really want to leak algo-specific configs here? wouldnt it be better to put into a subclass specific for glmr?

just thinking out loud here, but GlrmMojoModel could be the place for it

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
wendycwong: Good catch.  Fat fingers.  Fixed.

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
wendycwong: GLRM is a little tricky.  It decomposes a dataset A = X*Y.  Sometimes the user just want X, sometimes the user want the reconstructed dataset from X*Y.  So, there are a lot of specific parameter for it.  Adding one more glrmIterNumber is yet another one.

Since we try to move all the scoring to mojo model directed by master Jedis (Michals), we will need to put these algo specific parameters somewhere.  Hence, in the mojo in this case.

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -21,6 +21,7 @@
 import java.awt.image.BufferedImage;
 import java.io.*;
 import java.net.URL;
+import java.security.InvalidParameterException;
```
michalkurka: This kind of exception shouldn't be used, use IllegalArgumentException instead

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
michalkurka: @wendycwong I think @honzasterba has a good point and I think you also have a good point that GLRM is very specific

it would be better to do it in a more generic way - please let me think about it

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
wendycwong: Done!

PR #3628, opened 2019-06-26
PUBDEV-6621 : Add TreeShap for DRF

navdeep-G:
* Some notes:
    - I created a new class, `SharedTreeModelWithContributions`, which extends `SharedTreeModel` and extends `Model.Contributions`. 
    - `SharedTreeModelWithContributions` is extended by `GBMModel` and `DRFModel`. The idea here is to localize the use of `Model.Contributions` to only DRF and GBM and avoid code duplication as both GBM and DRF rely on similar calculations.

cc @michalkurka @honzasterba


navdeep-G:
@michalkurka Just noticed a failing JUnit for MOJO scoring. Will try to fix.

h2o-algos/src/main/java/hex/tree/drf/DRFModel.java
```diff
@@ -55,7 +68,91 @@ public DRFParameters() {
     }
     return preds;
   }
+  
+  @Override
+  public Frame scoreContributions(Frame frame, Key<Frame> destination_key) {
```
michalkurka: ScoreContributionsTask and this function are very similar to the one in GBM, can we avoid the code duplication?

h2o-genmodel/src/main/java/hex/genmodel/algos/drf/DrfMojoModel.java
```diff
@@ -58,4 +62,47 @@ public DrfMojoModel(String[] columns, String[][] domains, String responseColumn)
         return score0(row, 0.0, preds);
     }
 
+    @Override
+    public PredictContributions makeContributionsPredictor() {
```
michalkurka: again - seems duplicated

h2o-algos/src/main/java/hex/tree/drf/DRFModel.java
navdeep-G: Yes, the main difference is how the contributions are calculated in the `map()` task for `ScoreContributionsTask`.

h2o-genmodel/src/main/java/hex/genmodel/algos/drf/DrfMojoModel.java
navdeep-G: Yes, main diff is just how we handle binomial and regression in DRF vs. GBM.

h2o-algos/src/main/java/hex/tree/drf/DRFModel.java
navdeep-G: I have moved ScoreContributionsTask into its own file and scoreContributions into SharedTreeModel.

h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
```diff
@@ -736,5 +738,107 @@ public void generate(JCodeSB out) {
   protected <T extends JCodeSB> T toJavaForestName(T sb, String mname, int t ) {
     return (T) sb.p(mname).p("_Forest_").p(t);
   }
+  
+  @Override
+  public Frame scoreContributions(Frame frame, Key<Frame> destination_key) {
```
honzasterba: IsolationForestModel also extends from this class. Since this adds functionality for that class as well should it be tested?

navdeep-G: It is only supported for DRF and GBM. I figured placing this functionality in SharedTreeModel would avoid duplications as @michalkurka mentioned.

navdeep-G: I have added a check here: https://github.com/h2oai/h2o-3/pull/3628/commits/4aa05c4d5d55d705f4557e4078c473b68f17f2da

h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
```diff
@@ -736,5 +738,110 @@ public void generate(JCodeSB out) {
...
+    if (_output.nclasses() > 2) {
+      throw new UnsupportedOperationException(
+              "Calculating contributions is currently not supported for multinomial models.");
+    }
+    if ((!_parms.algoName().equals("GBM")) && (!_parms.algoName().equals("DRF"))) {
```
honzasterba: there are certainly more elegant and less expensive ways of achieving the same result. 
why I dont like it:
1) string comparison to determine functionality
2) IsoFor implements a method that throws UnsupportedOp - this will surprise someone
few ideas how to improve:
1) create a SharedTreeModel subclass that also implements Contributions and contains this method, have GBM and DRF inherit from that
2) move this method to a separate class that only implements Contributions, use this new class in GBM and DRF model classes scoreContributions method.

h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
```diff
@@ -736,5 +738,110 @@ public void generate(JCodeSB out) {
...
+          input[i] = chks[i].atd(row);
+        }
+        for (int i = 0; i < contribs.length; i++) {
+          contribs[i] = 0;
+        }
+
+        // calculate Shapley values
+        _treeSHAP.calculateContributions(input, contribs, 0, -1, workspace);
+
+        if (_model._parms.algoName().equals("GBM")) { // GBM
```
honzasterba: I think there should be separate subclasses for this part of the method so that you can avoid doing a string comparison for every row of the frame and instatiate the correct subclass for given algo

navdeep-G: I have addressed this here (hopefully): https://github.com/h2oai/h2o-3/pull/3628/commits/2e5b7fa839247afd28740658bbede30382826275

h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
navdeep-G: I have addressed this here (hopefully): 2e5b7fa

h2o-algos/src/main/java/hex/tree/SharedTreeModelGBMDRF.java
```diff
@@ -0,0 +1,159 @@
...
+import water.fvec.NewChunk;
+import water.fvec.Vec;
+import water.util.*;
+
+import java.util.ArrayList;
+import java.util.List;
+import hex.Model;
+import water.Key;
+
+public abstract class SharedTreeModelGBMDRF<
```
honzasterba: why not SharedTreeModelWithContributions? too long?

h2o-algos/src/main/java/hex/tree/SharedTreeModelGBMDRF.java
```diff
@@ -0,0 +1,159 @@
...
+    adaptTestForTrain(adaptFrm, true, false);
+    // remove non-feature columns
+    adaptFrm.remove(_parms._response_column);
+    adaptFrm.remove(_parms._fold_column);
+    adaptFrm.remove(_parms._weights_column);
+    adaptFrm.remove(_parms._offset_column);
+
+    final String[] outputNames = ArrayUtils.append(adaptFrm.names(), "BiasTerm");
+    
+    if (_parms.algoName().equals("DRF")) {
```
honzasterba: the if is unnecessary and can be replaced with polymorphism, then also

    .doAll(outputNames.length, Vec.T_NUM, adaptFrm)
     .outputFrame(destination_key, outputNames, null);


will not need to duplicated. you can do

    return getScoreContributionsTask(args).doAll(...).outputFrame(...)

and then

    abstract getScoreContributionsTask(..)

and keep the common part of ScoreContributionsTask in this class and in DRF and GBM have DRFModel.ScoreContributionsTask extends ScoreContributionsTask

h2o-algos/src/main/java/hex/tree/SharedTreeModelGBMDRF.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+        for (int i = 0; i < contribs.length; i++) {
+          contribs[i] = 0;
+        }
+
+        // calculate Shapley values
+        _treeSHAP.calculateContributions(input, contribs, 0, -1, workspace);
+        
+        for (int i = 0; i < nc.length; i++) {
+          // Prediction of DRF tree ensemble is an average prediction of all trees. So, divide contribs by ntrees
```
honzasterba: only the inside of the for loop is different, it should be an abstract method in ScoreContributionsTask with implementation in DRF and GBM model classes

h2o-algos/src/main/java/hex/tree/SharedTreeModelGBMDRF.java
navdeep-G: I agree. Done.

h2o-algos/src/main/java/hex/tree/SharedTreeModelGBMDRF.java
navdeep-G: Agree. Just pushed this: https://github.com/h2oai/h2o-3/pull/3628/commits/ddb6a071d58f62d0f64eee8b6174d34874eacc1f

h2o-algos/src/main/java/hex/tree/SharedTreeModelGBMDRF.java
navdeep-G: Yes, just pushed a change for this: https://github.com/h2oai/h2o-3/pull/3628/commits/ddb6a071d58f62d0f64eee8b6174d34874eacc1f

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -0,0 +1,144 @@
...
+        // calculate Shapley values
+        _treeSHAP.calculateContributions(input, contribs, 0, -1, workspace);
+        
+        // Add contribs to new chunk
+        addContribToNewChunk(contribs, nc);
+
+      }
+    }
+
+    public void addContribToNewChunk(float[] contribs, NewChunk[] nc) {
```
honzasterba: originally my suggestion was to make this abstract, but I think its reasonable to consider it default impl
I think it should still be protected

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -0,0 +1,144 @@
...
+    final String[] outputNames = ArrayUtils.append(adaptFrm.names(), "BiasTerm");
+    
+    return getScoreContributionsTask(this, _output._ntrees, _output._treeKeys, _output._init_f)
+            .doAll(outputNames.length, Vec.T_NUM, adaptFrm)
+            .outputFrame(destination_key, outputNames, null);
+  }
+
+  protected abstract ScoreContributionsTask getScoreContributionsTask(SharedTreeModel model, int ntrees, Key<CompressedTree>[][] treeKeys, double init_f);
+  
+  public class ScoreContributionsTaskDRF extends ScoreContributionsTask {
```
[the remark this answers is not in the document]
navdeep-G: Good idea. Done.

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -0,0 +1,144 @@
...
+        for (int i = 0; i < contribs.length; i++) {
+          contribs[i] = 0;
+        }
+
+        // calculate Shapley values
+        _treeSHAP.calculateContributions(input, contribs, 0, -1, workspace);
+        
+        // Add contribs to new chunk
+        addContribToNewChunk(contribs, nc);
+
```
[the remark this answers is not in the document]
navdeep-G: thx. resolved.

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
navdeep-G: will changed to protected. thx.

PR #3634, opened 2019-06-27
Added more R toolbox suggests.

wendycwong:
My 2d pdp requires more R toolboxes.  I forgot to add them into the correct places.


h2o-r/h2o-DESCRIPTION.template
```diff
@@ -55,6 +55,10 @@ Suggests: ggplot2,
           slam,
           bit64 (>= 0.9.7),
           data.table (>= 1.9.8),
+          open3d,
+          persp3D,
+          persp3Drgl,
+          snapshot3d,
```
ledell:
```suggestion
```

ledell: This is a function name, not a library name.  The library is `persp3Drgl`.

h2o-r/h2o-DESCRIPTION.template
```diff
@@ -55,6 +55,10 @@ Suggests: ggplot2,
           slam,
           bit64 (>= 0.9.7),
           data.table (>= 1.9.8),
+          open3d,
```
ledell:
```suggestion
```

ledell: This is a function name, not a library name. The library is `persp3D`.

wendycwong: Erin:  

I removed all the packages and only keep rgl, plot3Drgl.

ledell:
@wendycwong Ok, this is good now!

h2o-r/h2o-DESCRIPTION.template
```diff
@@ -55,6 +55,8 @@ Suggests: ggplot2,
           slam,
           bit64 (>= 0.9.7),
           data.table (>= 1.9.8),
+          rgl,
```
michalkurka: I think it is better to be on a safe side here and specify the version that is being tested as the minimal version

h2o-r/h2o-package/R/models.R
```diff
@@ -3473,8 +3473,6 @@ row_index=-1) {
   if (!noPairs &&
       ("plot3Drgl" %in% rownames(installed.packages())) &&
       ("rgl" %in% rownames(installed.packages()))) {
```
michalkurka: please generate a warning when either of these packages is not available. the warning should instruct users what specifically they need to do in order to enable this functionality

wendycwong: Done.

docker/Makefile
```diff
@@ -16,7 +16,7 @@ else
 endif
 
 dev-base:
-	docker build -t harbor.h2o.ai/opsh2oai/h2o-3/dev-base$(GPU_SUFFIX):$(VERSION) \
+	docker build --no-cache -t harbor.h2o.ai/opsh2oai/h2o-3/dev-base$(GPU_SUFFIX):$(VERSION) \
```
michal-raska: I'm not sure we want `--no-cache`. For example it takes very long to build the R images. Why do you need it please?

honzasterba: I had a suspicion that a change I made did not propagate into the R images. Reverted.

PR #3638, opened 2019-06-28
PUBDEV-6590: support for AutoML deletion (including models and other dependencies)

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6590

User should be able to delete AutoML instance and all its dependencies from any client.
Especially, all dependencies should be removed from DKV.
This includes:
- trained models (including cross-validation models, SE…).
- metrics.
- predictions.
- frames generated and referenced by AutoML (those created outside AutoML but used during training should remain though).
- …

WIP: still need to implement R support

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
@@ -434,7 +434,7 @@ public class Leaderboard extends Keyed<Leaderboard> {
    */
   @Override
   protected Futures remove_impl(Futures fs) {
-    for (Model m : getModels())
+    for (Key<Model> m : models)
       m.remove(fs);
     for (Key k : leaderboard_set_metrics.keySet())
       k.remove(fs);
--- a/h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
@@ -434,8 +434,15 @@ public class Leaderboard extends Keyed<Leaderboard> {
    */
   @Override
   protected Futures remove_impl(Futures fs) {
-    for (Key<Model> m : models)
+    for (Key<Model> m : models) {
+      Model model = m.get();
+      if (model != null) {
+        model.deleteCrossValidationPreds();
+        model.deleteCrossValidationModels();
+        model.deleteCrossValidationFoldAssignment();
+      }
       m.remove(fs);
+    }
     for (Key k : leaderboard_set_metrics.keySet())
       k.remove(fs);
     return super.remove_impl(fs);
--- a/h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
@@ -430,18 +430,15 @@ public class Leaderboard extends Keyed<Leaderboard> {
   }
 
   /**
-   * Delete everything in the DKV that this points to.  We currently need to be able to call this after deleteWithChildren().
+   * Delete object and its dependencies from DKV, including models.
    */
-  void delete() {
-    for (Key k : leaderboard_set_metrics.keySet())
-      k.remove();
-    remove();
-  }
-
-  void deleteWithChildren() {
+  @Override
+  protected Futures remove_impl(Futures fs) {
     for (Model m : getModels())
-      m.delete();
-    delete();
+      m.remove(fs);
+    for (Key k : leaderboard_set_metrics.keySet())
+      k.remove(fs);
+    return super.remove_impl(fs);
   }
 
   private static double[] defaultMetricForModel(Model m) {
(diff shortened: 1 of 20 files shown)


h2o-py/tests/testdir_algos/automl/pyunit_automl_memory_cleanup.py
```diff
@@ -272,13 +292,192 @@ def test_param_disabled():
...
+        assert contains_event_log(project_name, keys)
+        expectations = dict(
+            models_base=max_models + 2,  # 2 SEs
+            cv_models=(max_models+2) * nfolds,  # 1 cv model per fold for all models, incl. SEs
+            predictions=(len(keys['cv_models'])  # cv predictions
+                         + len(keys['models_base'])  # cv holdout predictions
+                         ),
+            metrics=(len(keys['cv_models']) * 3  # for each cv model, 1 on training frame, 1 on validation frame (=training for cv), 1 on frame with null key (what is this?)
+                     + len(keys['models_base'])  # for each model, 1 on training_frame
+                     + (2 * 2)  # for each SE, 1 on levelone training, 1 for final scoring (preds_levelone)
```
seb-h2o: @michalkurka I'm quite happy with this test as it now gives me a – I expect complete – reference about all keys produced during AutoML and therefore for each model training.
For CV models although, I noticed there was 3 metrics stored in DKV for each of them: I can understand the one on training and validation frames, but a 3rd one is done against a frame with 'null' key. I spent some time searching in the code how it could happen but couldn't figure it out: any idea where this could come from? This looks like a temporary metric that should actually be deleted once the CV model is completed.
WDYT?

h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
```diff
@@ -430,18 +430,22 @@ Model getLeader() {
...
-  void delete() {
+  @Override
+  protected Futures remove_impl(Futures fs) {
+    for (Key<Model> m : models) {
+      Model model = m.get();
+      if (model != null) {
+        model.deleteCrossValidationPreds();
+        model.deleteCrossValidationModels();
+        model.deleteCrossValidationFoldAssignment();
+      }
```
seb-h2o: @michalkurka As you can see, when deleting a model, the CV details are not automatically deleted today.
I think you didn't want to include this to `model.delete()`: can you please remind me the reason for this?

h2o-py/tests/testdir_algos/automl/pyunit_automl_memory_cleanup.py
michalkurka: that could be the metric calculated on the adapted training/validation frame - this frame doesn't have a key - it is strange though

are the metrics same as train or validation or completely different?

h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
michalkurka: the argument was backwards compatibility, I don't think it is the best behavior but it could be a behavior people could be relying on

- imho we can transition to a better behavior - first give an option to delete with the CV models and later make that default

WDYT?

seb-h2o: @michalkurka this would be ideal transition, although we have to come up with something generic as RESTFul/client APIs are functional (as opposed to OO) so we have a limited way to extend it in a way that makes sense for every object.
For example in Python, we could use sth like:
```h2o.remove(model, cascade=False)```
as soon as it makes sense for all objects that can be deleted: frames, models, automl.
For frames, I guess the param could be ignored.
For models, `cascade=True` would additionally delete those CVs objects.
For AutoML, we could decide not to delete models and their dependencies if `cascade=False`.

I could try that: will submit a proposal.

File: h2o-py/tests/testdir_algos/automl/pyunit_automl_memory_cleanup.py (hunk 1 above)
Comment (seb-h2o) [reply h2o-py/tests/testdir_algos/automl/pyunit_automl_memory_cleanup.py]: @michalkurka Yep, just debugged this, and this are metrics computed on the adapted validation frame for CV holdout predictions.
It looks like this metrics key is stored as a byproduct of `Model#predictScoreImpl`, more exactly because we stored those metrics every time we set `computeMetrics=true`:
```
 if (computeMetrics && bs._mb != null) //metric builder can be null if training was interrupted/cancelled
      bs._mb.makeModelMetrics(this, fr, adaptFrm, bs.outputFrame());
```

It doesn't look necessary though in this case: tell me if I'm wrong but I don't see those metrics used anywhere.
Could create a small ticket to do some cleanup here.

seb-h2o:
@deil87 @ledell please review asap: I would like to have this in 3.26, it started as a request from DAI

h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
michalkurka: I very much like the `cascade=True` - IMHO that is a great way to go.

h2o-py/tests/testdir_algos/automl/pyunit_automl_memory_cleanup.py
michalkurka: Thanks, I also don't see the reason to keep them there - please make the jira.

seb-h2o:
build looks "good", merging

PR #3641, opened 2019-06-29
K refers to principal components and not clusters

kdpsingh:


michalkurka:
@kdpsingh thanks for the fix

@zainhaq-h2o please review

PR #3643, opened 2019-07-01
PUBDEV-6598: Update documentation for partial plots for 2D

zainhaq-h2o:
Updated the notes in H2O userguide to indicate support for 2-D pdps.
Driveby fixes to the Python documentation.

See: https://0xdata.atlassian.net/browse/PUBDEV-6598
https://0xdata.atlassian.net/browse/PUBDEV-6438

### DIFF
--- a/h2o-docs/src/product/performance-and-prediction.rst
+++ b/h2o-docs/src/product/performance-and-prediction.rst
@@ -464,7 +464,7 @@ Thus, the one-dimensional partial dependence of function :math:`g` on :math:`X_j
 
 - The partial dependence of a given feature is :math:`Xj` (where :math:`j` is the column index)
 - You can also change the equation to sum from 1 to N instead of 0 to N-1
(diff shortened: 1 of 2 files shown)


h2o-docs/src/product/performance-and-prediction.rst
```diff
@@ -464,6 +464,7 @@ Thus, the one-dimensional partial dependence of function :math:`g` on :math:`X_j
 
 - The partial dependence of a given feature is :math:`Xj` (where :math:`j` is the column index)
 - You can also change the equation to sum from 1 to N instead of 0 to N-1
+- Use the `col_pairs_2dpdp` option along with a list containing pairs of column names to generate 2D partial dependence plots
```
angela0xdata: change `col_pairs_2dpdp`  to ``col_pairs_2dpdp`` so that it shows up as a keyword rather than italics.

PR #3282, opened 2019-02-06
PUBDEV-6255: Mojo support for Target Encoding

deil87:

PR #3645, opened 2019-07-02
PUBDEV-6637: Refactored out RowToRawDataConverter 

deil87:
These changes were extracted from my  [TE Mojo Support PR](https://github.com/h2oai/h2o-3/pull/3282/files) to make it easier to merge both PRs.

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/easy/DomainMapConstructor.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/easy/DomainMapConstructor.java
@@ -1,0 +1,35 @@
+package hex.genmodel.easy;
+
+import hex.genmodel.GenModel;
+
+import java.util.HashMap;
+
+/**
+ *  Create map from input variable domain information.
+ */
+public class DomainMapConstructor {
+
+  public final GenModel _m;
+  
+  public DomainMapConstructor(GenModel m) {
+    _m = m;
+  }
+
+  public HashMap<Integer, HashMap<String, Integer>> create() {
+   
+    // This contains the categorical string to numeric mapping.
+    HashMap<Integer, HashMap<String, Integer>> domainMap = new HashMap<>();
+    for (int i = 0; i < _m.getNumCols(); i++) {
+      String[] domainValues = _m.getDomainValues(i);
+      if (domainValues != null) {
+        HashMap<String, Integer> m = new HashMap<>();
+        for (int j = 0; j < domainValues.length; j++) {
+          m.put(domainValues[j], j);
+        }
+
+        domainMap.put(i, m);
+      }
+    }
+    return domainMap;
+  }
+}
--- a/h2o-genmodel/src/main/java/hex/genmodel/easy/RowToRawDataConverter.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/easy/RowToRawDataConverter.java
@@ -1,0 +1,175 @@
+package hex.genmodel.easy;
+
+import hex.genmodel.GenModel;
+import hex.genmodel.algos.deepwater.DeepwaterMojoModel;
+import hex.genmodel.easy.exception.PredictException;
+import hex.genmodel.easy.exception.PredictNumberFormatException;
+import hex.genmodel.easy.exception.PredictUnknownCategoricalLevelException;
+import hex.genmodel.easy.exception.PredictUnknownTypeException;
+
+import javax.imageio.ImageIO;
+import java.awt.image.BufferedImage;
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.util.HashMap;
+
+/**
+ * This class is intended to transform a RowData instance - for which we want to get prediction to - into a raw array
+ */
+public class RowToRawDataConverter {
+
+  public final GenModel _m;
+  private final HashMap<String, Integer> _modelColumnNameToIndexMap;
+  private final HashMap<Integer, HashMap<String, Integer>> _domainMap;
+  private final EasyPredictModelWrapper.ErrorConsumer _errorConsumer;
+
+  private final boolean _convertUnknownCategoricalLevelsToNa;
+  private final boolean _convertInvalidNumbersToNa;
+  
+  public RowToRawDataConverter(GenModel m,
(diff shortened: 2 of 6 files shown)


h2o-genmodel/src/main/java/hex/genmodel/easy/RowToRawDataConverter.java
```diff
@@ -0,0 +1,175 @@
...
+
+      if (domainValues == null) {
+        // Column is either numeric or a string (for images or text)
+        double value = Double.NaN;
+        Object o = data.get(dataColumnName);
+        if (o instanceof String) {
+          String s = ((String) o).trim();
+          // Url to an image given
+          if (isImage) {
+            boolean isURL = s.matches("^(https?|ftp|file)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]");
```
seb-h2o: looks like a terrible idea (and I'm sure many wrong formatted urls match this regex).
Why not ```try { new URL(s); isUrl = true; } catch (Exception ignored) { isUrl = false; }``` instead?

I understand it's legacy, but my eyes still bleeding.

h2o-genmodel/src/main/java/hex/genmodel/easy/DomainMapConstructor.java
```diff
@@ -0,0 +1,35 @@
...
+
+import hex.genmodel.GenModel;
+
+import java.util.HashMap;
+
+/**
+ *  Create map from input variable domain information.
+ */
+public class DomainMapConstructor {
+
```
seb-h2o: static methods are not ALWAYS bad: `function(input) -> output` with no side effect = perfect candidate for a static method: I don't really see the benefit of the "kingdom of nouns" here.

h2o-genmodel/src/main/java/hex/genmodel/easy/RowToRawDataConverter.java
tomkraljevic: For this PR I think it's appropriate to change nothing from the old code except for moving it around.  But for the future, it would be safe to remove the Deep Water related stuff like isImage.

deil87: @seb-h2o How would you call converter then? I remember the article that you have sent to me about kingdom of Nouns. But I think that the truth is somewhere in the middle. There is such an abstraction as Conversion so I think in this case it is ok to add corresponding suffix. 
Regarding urls regexp... I was trying not to look inside the method :) Let's create separate class URLMatcher :)

tomkraljevic: This is really dead code at this point.  The best way to refactor it is to remove it (but not in this PR).

h2o-genmodel/src/main/java/hex/genmodel/easy/DomainMapConstructor.java
deil87: @seb-h2o Agree, was also thinking that probably it is too much.... but decided that it is not bounded to the scope of what is called EasyPredictModelWrapper. I got used to prefer classes over static methods because it is possible to mock/extend them during testing. Also I think that we have enough static methods on the project :)

h2o-genmodel/src/main/java/hex/genmodel/easy/RowToRawDataConverter.java
deil87: @tomkraljevic do you want me to create a jira for this?

tomkraljevic: sure.

h2o-genmodel/src/main/java/hex/genmodel/easy/RowToRawDataConverter.java
```diff
@@ -0,0 +1,175 @@
...
+   * 
+   * @param data instance of RowData we want to get prediction for
+   * @param rawData array that will be filled up from RowData instance and returned
+   * @return `rawData` array with data from RowData. 
+   * @throws PredictException Note: name of the exception feels like out of scope of the class with name `RowToRawDataConverter` 
+   * but this conversion is only needed to make it possible to produce predictions so it makes sense
+   */
+  public double[] convert(RowData data, double[] rawData) throws PredictException {  
+
+    // TODO: refactor
```
honzasterba: how about fixing this todo when aready refactoring this code? at least splitting the method into smaller chunks would be most welcome

h2o-genmodel/src/main/java/hex/genmodel/easy/RowToRawDataConverter.java
honzasterba: I tend to disagree, since the purpose of this PR is to refactor (move around) this code, I would vote to make it better. But I understand you have different fish to fry ATM

h2o-genmodel/src/main/java/hex/genmodel/easy/DomainMapConstructor.java
honzasterba: @seb-h2o true, they are not bad until you want to unit test the code that uses them.

h2o-genmodel/src/main/java/hex/genmodel/easy/RowToRawDataConverter.java
deil87: @honzasterba I'm really into refactoring even legacy code which is not really related to a scope of the task. Just in current situation I don't have time for this. I will create a jira for this piece of code for somebody to have fun with it later!

h2o-genmodel/src/main/java/hex/genmodel/easy/DomainMapConstructor.java
seb-h2o: @honzasterba that's why I mentioned "with no side-effect": there's no reason to mock pure functions when you want to test caller.

deil87: @seb-h2o side effects are not the only reasons for using mocks. For the sake of testing I might want a specific output to be returned from this method. But it is definitely  true that we are abusing ability to mutate everything around our methods in Java.

PR #3647, opened 2019-07-02
PUBDEV-6639: adding cascade param to remove API

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6639

An attempt at making `remove`/`delete` methods "cascadable".
Note that by contract, those methods were originally supposed to remove the object and all its subparts by default. Unfortunately, the contract was not fully respected by `Model` and `AutoML`.

Implemented the changes on backend and Py+R client.

FOR REVIEWER, main changes in:
- `Keyed`: https://github.com/h2oai/h2o-3/pull/3647/files#diff-eeadbba19e068aa9e00ffa0fa349f2b2
- `Lockable`: https://github.com/h2oai/h2o-3/pull/3647/files#diff-f748e31caa232a27f978e76704546969
- `RemoveHandler`: https://github.com/h2oai/h2o-3/pull/3647/files#diff-4491ddbd9d3bde94766d5e6058ca45fb
and for usage examples:
- `Model`: https://github.com/h2oai/h2o-3/pull/3647/files#diff-f7c84f8bfd6b9c51e43bccf2a55c6687
- `Leaderboard`:  https://github.com/h2oai/h2o-3/pull/3647/files#diff-c0a39c9d409ec866d644b215ae7580ad

### DIFF
--- a/h2o-py/h2o/h2o.py
+++ b/h2o-py/h2o/h2o.py
@@ -910,6 +910,7 @@ def remove(x, cascade=True):
 
     :param x: H2OFrame, H2OEstimator, or string, or a list of those things: the object(s) or unique id(s)
         pointing to the object(s) to be removed.
+    :param cascade: boolean, if set to TRUE (default), the object dependencies (e.g. submodels) are also removed.
     """
     item_type = U(str, Keyed)
     assert_is_type(x, item_type, [item_type])
--- a/h2o-py/h2o/h2o.py
+++ b/h2o-py/h2o/h2o.py
@@ -904,7 +904,7 @@ def log_and_echo(message=""):
     api("POST /3/LogAndEcho", data={"message": str(message)})
 
 
-def remove(x):
+def remove(x, cascade=True):
     """
     Remove object(s) from H2O.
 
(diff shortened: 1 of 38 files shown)


seb-h2o:
> it would be nice if there was an API method that users could rely on
@michalkurka apparently you mean an API method that would be also exposed to clients then.
In terms of use case, the only scenario for using this new `removeSubObjects` method would be if the user wants to delete the object tree but keep the parent object.
This could make sense for models when using xval, but definitely not for AutoML: with AutoML, it's rather the other way round, user could decide to keep some of the base models, but get rid of the AutoML object itself.
Also, it creates a logical order in which `remove` methods should be called (not really fan of that):
- user can call `removeSubObjects` and later `remove`.
- but `remove` followed with `removeSubObjects` can easily raise Exceptions when called from Java/Scala, and wouldn't make sense at all from Py/R.

Another concern is that for end-user, it's difficult to know what is a dependency-object and what is not: for example today, model metrics are directly attached to the output the model and in spite of having an aggregation relation with the main model (so very similar to the relation CV models have with the model), they're always deleted when the model is. So this shows that the semantic of what to delete or not in `removeSubObjects` is not that simple, and would be specific for each type of object.

A suggestion that would prove more control to the user would be, not to expose an additional `removeXXX` method, but a method listing all keys associated to the object, something like `listDependentKeys`.
This way, end-user can easily pass this to the `h2o.remove` method if he needs to delete them. And it also provides us with a method allowing us to control which keys are kept in store at every level (useful for testing and memory management also on client level).

wdyt?

h2o-py/h2o/h2o.py
```diff
@@ -904,7 +904,7 @@ def log_and_echo(message=""):
     api("POST /3/LogAndEcho", data={"message": str(message)})
 
 
-def remove(x):
+def remove(x, cascade=True):
```
seb-h2o: @michalkurka one question: for backwards compatibility. Should I default `cascade=False` (although I don't like it as original contract was for `remove` to delete object+dependencies).
Or should we document that users should use `cascade=False` if they have any issue to obtain the previous behaviour.

PR #3651, opened 2019-07-03
PUBDEV-6593: Get rid of dependency on response column from TargetEncoder apply method.

deil87:
This will allow to make a predictions with TargetEncoder on data without response column. Other option is to use Mojo model.


deil87:
@michalkurka Could you please take a look at this one?

PR #3657, opened 2019-07-08
PUBDEV-6638: remove unnecessary metrics computation on CV models

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-6638

### DIFF
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -1543,8 +1543,7 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
                                       CFuncRef customMetricFunc) {
     return new BigScore(domains[0],
                         names != null ? names.length : 0,
-                        adaptFrm.means(),
-                        _output.hasWeights() && adaptFrm.find(_output.weightsName()) >= 0,
+                        adaptFrm,
                         computeMetrics,
                         makePrediction,
                         j,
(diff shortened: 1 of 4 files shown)


h2o-core/src/main/java/hex/Model.java
```diff
@@ -1641,7 +1646,7 @@ protected void setupLocal() {
 
     @Override public void map(Chunk chks[], NewChunk cpreds[] ) {
       if (isCancelled() || _j != null && _j.stop_requested()) return;
-      Chunk weightsChunk = _hasWeights && _computeMetrics ? chks[_output.weightsIdx()] : null;
```
michalkurka: What is the purpose of this change? It changes the current behavior: "if computeMetrics is false then make a prediction for all rows regardless of the weights column"

seb-h2o: @michalkurka the rationale comes from the change in `ModelBuilder` below.
without it, if computeMetrics is false, then weights are ignored and we get predictions for all rows, including those with weight 0, as it is set in CV for the holdout fold.

maybe I'm missing something, but why does the current behaviour (as you described it) makes sense? why should computeMetrics flag be tied to weights? especially when we can test for weights completely independently?

Agree that I should add a test for this though, if you also agree with the change.

seb-h2o: When I noticed this, it scared me a bit as well as we use `computeMetrics=false` when getting predictions to preparing levelone frames in StackedEnsemble. Now I doubt this works correctly if user also specified column weights.

michalkurka: I think the issue of CV case is that we are overloading the weights and using them for another (internal purpose).

In the user-facing API, we have (in one place - for async predict) `computeMetrics=false` but the user should still see predictions for all rows even when the weights columns is provided and some of the weights are 0. From the user's point of view - weights should only matter for metrics, but shouldn't affect making predictions.

seb-h2o: looks like I misinterpreted `weights`: I thought those were column-wise weights but these are actually row-wise weights.
Need to verify how it impacts predictions though. I understand it disables prediction on a row if its weight is null, but what if its weight is 0.5? If it's applying this factor to the prediction, then my change would be bad.

Still don't understand why `computeMetrics` would be the right flag to switch this on/off.

seb-h2o: @michalkurka sorry, I wrote my last reply before reading yours.
Makes sense then: as you say, it's probably CV that is overloading weights here then.
Will find another approach then.

PR #3659, opened 2019-07-09
[PUBDEV-6627] setup special formatting for Zeppelin

honzasterba:
also fix some bugs in frame.show/summary

### DIFF
--- a/h2o-py/h2o/display.py
+++ b/h2o-py/h2o/display.py
@@ -22,6 +22,8 @@ class H2ODisplay(object):
 
     # True/False flag whether we are connected to a Jupyter notebook. Computed on the first use.
     _jupyter = None
+    # True/False flag whether we are connected to a Zeppelin note. Computed on the first use.
+    _zeppelin = None
 
     def __init__(self, table=None, header=None, table_header=None, **kwargs):
         self.table_header = table_header
@@ -41,7 +43,10 @@ class H2ODisplay(object):
             print()
             print(self.table_header + ":")
             print()
-        if H2ODisplay._in_ipy():
+        if H2ODisplay._in_zep():
+            print("%html " + H2ODisplay._html_table(self.table, self.header))
+            self.do_print = False
+        elif H2ODisplay._in_ipy():
             from IPython.display import display
             display(self)
             self.do_print = False
@@ -79,6 +84,13 @@ class H2ODisplay(object):
             except ImportError:
                 H2ODisplay._jupyter = False
         return H2ODisplay._jupyter
+        
+    @staticmethod
+    def _in_zep():  # are we in zeppelin? then use zeppelin pretty print support
+        if H2ODisplay._zeppelin is None:
+            import os
+            H2ODisplay._zeppelin = H2ODisplay._in_ipy() and "ZEPPELIN_RUNNER" in os.environ
+        return H2ODisplay._zeppelin
 
     # some html table builder helper things
     @staticmethod
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -448,15 +448,17 @@ class H2OFrame(object):
             print("This H2OFrame is empty.")
             return
         if not self._ex._cache.is_valid(): self._frame()._ex._cache.fill()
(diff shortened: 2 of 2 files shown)


h2o-py/h2o/frame.py
```diff
@@ -448,15 +448,17 @@ def show(self, use_pandas=False, rows=10, cols=200):
...
             return
         if not self._ex._cache.is_valid(): self._frame()._ex._cache.fill()
-        if H2ODisplay._in_ipy():
+        if H2ODisplay._in_zep():
+            print("%html " + self._ex._cache._tabulate("html", False, rows=rows))
+        elif H2ODisplay._in_ipy():
             import IPython.display
             if use_pandas and can_use_pandas():
-                IPython.display.display(self.head(rows=rows, cols=cols).as_data_frame(fill_cache=True))
+                IPython.display.display(self.head(rows=rows, cols=cols).as_data_frame(use_pandas=True))
```
jakubhava: What is please reason to remove`fill_cache` and introduce`use_pandas`?

h2o-py/h2o/frame.py
```diff
@@ -448,15 +448,17 @@ def show(self, use_pandas=False, rows=10, cols=200):
...
             else:
-                IPython.display.display_html(self._ex._cache._tabulate("html", False), raw=True)
+                IPython.display.display_html(self._ex._cache._tabulate("html", False, rows=rows), raw=True)
         else:
             if use_pandas and can_use_pandas():
-                print(self.head(rows=rows, cols=cols).as_data_frame(True))  # no keyword fill_cache
+                print(self.head(rows=rows, cols=cols).as_data_frame(use_pandas=True))
```
jakubhava: I can see the `if-else` checks for zeppelin & Jupyter more times in the code. What do you think of creating a method like `print/show/display` in `display.py` file which would handle this internally and the rest of the code would just call this method?

h2o-py/h2o/display.py
```diff
@@ -79,6 +84,13 @@ def _in_ipy():  # are we in ipy? then pretty print tables with _repr_html
             except ImportError:
                 H2ODisplay._jupyter = False
         return H2ODisplay._jupyter
+        
+    @staticmethod
+    def _in_zep():  # are we in zeppelin? then use zeppelin pretty print support
+        if H2ODisplay._zeppelin is None:
+            import os
+            H2ODisplay._zeppelin = H2ODisplay._in_ipy() and "ZEPPELIN_RUNNER" in os.environ
```
jakubhava: Probably extreme edge case, but what if someone has `ZEPPELIN_RUNNER` defined in their bash env for some obscure reasons?

h2o-py/h2o/frame.py
honzasterba: because as_data_frame has no argument fill_cache and this caused an error

h2o-py/h2o/display.py
honzasterba: yes, that is unfortunately a possibility, however very unlikely, because this env var is set directly by ZEPPELIN and not by user/config
I am aware this can cause issues, but would like your oppinion on this

h2o-py/h2o/frame.py
honzasterba: everywhere where it is possible we use H2ODisplay directly, the two frame method have special behavior where it is best to keep it like this.
It would be possible to rewrite the code to move most of the logic to H2ODisplay but it would be a major project with little benefit

h2o-py/h2o/display.py
jakubhava: I think it is fine to go this way, the case mentioned above is really rare I believe.

When I was initially looking into this issue, I was thinking of a solution which would print independently on Zeppelin/Jupyter, but I guess you have already try that.

h2o-py/h2o/frame.py
jakubhava: Thanks for the explanation!

Tagar:
Thank change broke two other methods 
model.summary and .model_performance()

I left a comment in 

https://0xdata.atlassian.net/browse/PUBDEV-6627

PR #3666, opened 2019-07-10
[PUBDEV-6657] H2O.CLOUD._memary(idx).getTimestamp returns 0 even though it should return valid timestamp of remote node

jakubhava:
Fix the issue with zero timestamp & move the timestamp related code to separate files to avoid making H2O class making even larger.


h2o-core/src/main/java/water/H2ONode.java
```diff
@@ -226,12 +226,15 @@ boolean removeClient() {
...
   // *interned*: there is only one per InetAddress.
   static private H2ONode intern(H2Okey key, short timestamp) {
-    final boolean foundPossibleClient = H2O.decodeIsClient(timestamp);
+    final boolean foundPossibleClient = H2ONodeTimestamp.decodeIsClient(timestamp);
     H2ONode h2o = INTERN.get(key);
     if (h2o != null) {
       if (foundPossibleClient || h2o.isPossibleClient()) {
         h2o.refreshClient(timestamp);
       }
+      if (h2o.getTimestamp() == H2ONodeTimestamp.TIMESTAMP_UNDEFINED) {
```
jakubhava: This was the main source of the issue. In case we are using flatfile, we intern the nodes in the flatfile, but we do not know the timestamp of the remote node at that time. Therefore we set the timestamp to undefined and set it to correct value once we hear from the remote node. The value 0 is not used for any valid timestamp so we can safely use it as undefined.

PR #3667, opened 2019-07-10
[PUBDEV-6658] Cloud hash should consist of timestamp as well

jakubhava:
The main purpose of this change is to be able to discover from client that the client is talking to old or new cluster simply by checking one field of the first node which is talking to client.

In the client code we can just check the of the cloud_hash and if it is not equal to previous name of the cloud hash the client was connected to, client should drop the previous state and connect to the new cluster.

Also this should not break any of the current behavior. `cloud_hash` is formed only from regular worker nodes (not clients) and regular worker nodes can't connect & disconnect from cluster, they are static. Therefore the worker nodes will have the same timestamp the whole time they are part of the cluster & therefore the `cloud_hash` will remain unchanged.

### DIFF
--- a/h2o-core/src/main/java/water/H2ONode.java
+++ b/h2o-core/src/main/java/water/H2ONode.java
@@ -352,9 +352,25 @@ public final class H2ONode extends Iced<H2ONode> implements Comparable {
     return sb.toString();
   }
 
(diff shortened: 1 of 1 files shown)


jakubhava:
@h2o-ops please test!

PR #3668, opened 2019-07-10
[XGBOOST] fix gblinear with GPU backend

honzasterba:
- set updater value to one supported by gblinear


honzasterba:
tested with previously failing xgb-smoke

PR #3669, opened 2019-07-10
PUBDEV-6492: Add Isolation Forest to Flow

michalkurka:


h2o-algos/src/main/java/hex/tree/isofor/IsolationForest.java
```diff
@@ -35,7 +35,7 @@
     };
   }
   @Override public BuilderVisibility builderVisibility() {
-    return BuilderVisibility.Beta;
+    return BuilderVisibility.Stable;
```
Pscheidl: This change also makes sense anyway :)

michalkurka:
@seb-h2o yes, they are executed in Jenkins pipeline

this particular test runs as a part of

    test-flow-headless-small:
	./scripts/run.py --wipeall --geterrs --test h2o-web/lib/h2o-flow/build/js/headless-test.js --jvm.xmx 4g --pto 14400 --ptt test-small


    [
      stageName: 'Flow Headless Small', target: 'test-flow-headless-small',
      timeoutValue: 75, component: pipelineContext.getBuildConfig().COMPONENT_JS
    ],

They are smoke tests by nature, no output is checked. It only verifies Flow is able to build the model. There are few of them for AutoML as well.

h2o-algos/src/main/java/hex/tree/isofor/IsolationForest.java
michalkurka: Thank you :)

PR #3670, opened 2019-07-10
PUBDEV-6015: Deprecate support for Java 7

michalkurka:
Java 7 support will be deprecated but the code and builds will still be Java 7 compatible and tested for Java 7 compatibility.

Users will be able to override the compatibility check if there is a good reason for it.

We will remove official java 7 support in H2O 3.26, in fix releases of 3.26 we will continue unofficially support Java 7. In H2O 3.28 neither code or builds will be Java 7 compatible.

### DIFF
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -1983,8 +1983,8 @@ final public class H2O {
     // Notes: 
     // - make sure that the following whitelist is logically consistent with whitelist in R code - see function .h2o.check_java_version in connection.R
     // - upgrade of the javassist library should be considered when adding support for a new java version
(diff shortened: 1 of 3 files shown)


tomkraljevic:
I would ask Nidhi whether the customer base (especially Hadoop) has mostly transitioned to Java 8.
My instinct is it's probably fine to move forward.

Just being conservative, I would say the MOJO runtime library should retain Java 7 support (especially if it's easy).

michalkurka:
@tomkraljevic MOJO can stay Java 7 compatible (forever)

Hadoop 2.7+ requires Java 7 or newer
Hadoop 3.0+ requires Java 8 or newer

It is definitely possible there is a Hadoop out there running on Java 7 but I've not seen one in any of my customer interactions. I will double-check with Nidhi.

angela0xdata:
Added a doc change to this PR.

PR #3675, opened 2019-07-11
PUBDEV-6452: Update GLM booklet to include information about minimizing negative log likelihood

zainhaq-h2o:
The "GLM in H2O" section has been updated to include information about minimizing negative log likelihood.

https://0xdata.atlassian.net/browse/PUBDEV-6452


wendycwong:
Zain:

Angela proposed something else here in this JIRA:  https://0xdata.atlassian.net/browse/PUBDEV-6452.


Wendy

PR #3680, opened 2019-07-12
PUBDEV-6518: Expose an option to collect logs in plain-text, not as a ZIP

michalkurka:


h2o-py/h2o/h2o.py
```diff
@@ -1031,12 +1031,13 @@ def download_csv(data, filename):
...
 
-def download_all_logs(dirname=".", filename=None):
+def download_all_logs(dirname=".", filename=None, container=None):
     """
     Download H2O log files to disk.
 
     :param dirname: a character string indicating the directory that the log file should be saved in.
-    :param filename: a string indicating the name that the CSV file should be. Note that the saved format is .zip, so the file name must include the .zip extension.
+    :param filename: a string indicating the name that the CSV file should be. Note that the default container format is .zip, so the file name must include the .zip extension.
+    :param container: a string indicating how to archive the logs, choice of "ZIP" and "TXT" (all logs concatenated together)
```
jakubhava: I just wonder whether `.log` extension is not more common for logs than `.txt` ?

h2o-core/src/main/java/water/util/GetLogsFromNode.java
```diff
@@ -76,26 +83,26 @@ public void doIt() {
...
-        int bytesIn = 0;
-        //loop through dirList, and zip the files
+        //loop through dirList, and archive the files
         for(int i=0; i<dirList.length; i++)
         {
           File sourceFile = new File(sourceDir, dirList[i]);
           File destinationFile = new File(destinationDir, dirList[i]);
           if(sourceFile.isDirectory())
           {
             //if the File object is a directory, call this
```
jakubhava: Can we remove `//if the File object is a directory, call this` and `//loop again` comments? They don't seem to bring any additional value in this case

jakubhava: There are actually more places where we could remove unnecessary comments, probably not worth it mixing this with this PR

h2o-core/src/main/java/water/util/LogArchiveWriter.java
```diff
@@ -0,0 +1,104 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Date;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+/**
+ * Abstract layer over different types of log containers (zip, txt,...)
+ * @param <OS>
```
jakubhava: NIT: doc for `<OS>` could be a but more explanatory (or removed)

h2o-core/src/main/java/water/api/RequestServer.java
```diff
@@ -877,39 +870,40 @@ private static String getOutputLogStem() {
...
+  private static byte[] archiveLogs(LogArchiveContainer container, Date now,
+                                    byte[][] results, byte[] clientResult, String topDir) throws IOException {
     int l = 0;
     assert H2O.CLOUD._memary.length == results.length : "Unexpected change in the cloud!";
     for (byte[] result : results) l += result.length;
     ByteArrayOutputStream baos = new ByteArrayOutputStream(l);
 
     // Add top-level directory.
-    ZipOutputStream zos = new ZipOutputStream(baos);
+    LogArchiveWriter zos = container.createLogArchiveWriter(baos);
```
jakubhava: `zos` implies ZipOutputStream but we use `LogArchiveWriter`

h2o-core/src/main/java/water/api/RequestServer.java
```diff
@@ -732,7 +724,10 @@ private static NanoResponse maybeServeSpecial(RequestUri uri) {
       // url "/3/Foo/bar" => path ["", "GET", "Foo", "bar", "3"]
       String[] path = uri.getPath();
       if (path[2].equals("")) return redirectToFlow();
-      if (path[2].equals("Logs") && path[3].equals("download")) return downloadLogs();
+      if (path[2].equals("Logs") && path[3].equals("download")) {
+        LogArchiveContainer container = path.length >= 6 ? LogArchiveContainer.valueOf(path[4].toUpperCase()) : LogArchiveContainer.ZIP;
```
jakubhava: This part could use some explanations. After looking into it, I think this is the part where we decide which type of `container` to use, but don't understand the logic.

What do you think of making this part ( maybe move it out to separate method) more self-explanatory?

michalkurka: yeah, this is bad but that is because of how the whole method is designed :)

I will try to make it better, thx

h2o-core/src/main/java/water/api/RequestServer.java
michalkurka: right, thx

h2o-py/h2o/h2o.py
michalkurka: hmm, good point - I think it does make sense to change it to LOG

h2o-core/src/main/java/water/util/LogArchiveWriter.java
michalkurka: yeah, will fix - this is ugly anyway

bilcus:
@michalkurka Can we add it to Flow as well?

h2o-core/src/main/java/water/api/RequestServer.java
jakubhava: Yup, thanks a lot!

PR #3681, opened 2019-07-12
[Hotfix] GBM Mojo compatibility - default link function not correctly set

michalkurka:


maurever:
Thank you @michalkurka for this fix! My previous quick fix was really bad. :(

PR #3683, opened 2019-07-12
PUBDEV-6672: fixed bug in python client code when no figure is plotted.

wendycwong:
Honza has found this bug for me: https://0xdata.atlassian.net/browse/PUBDEV-6672

I have fixed the bug if Python when no figure is plotted.  I added his test as the python unit test.

However, I am still working on refractoring the partial plot into separate functions.  Please start the review though.

### DIFF
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -1090,17 +1090,81 @@ class ModelBase(backwards_compatible(Keyed)):
                     kwargs["user_cols"] = user_cols
                     kwargs["user_splits"] = user_values
                     kwargs["num_user_splits"] = user_num_splits
-                else:
-                    kwargs["user_cols"] = None
-                    kwargs["user_splits"] = None
-                    kwargs["num_user_splits"] = None
-
+        else:
+            kwargs["user_cols"] = None
+            kwargs["user_splits"] = None
+            kwargs["num_user_splits"] = None
+            
         json = H2OJob(h2o.api("POST /3/PartialDependence/", data=kwargs),  job_type="PartialDependencePlot").poll()
         json = h2o.api("GET /3/PartialDependence/%s" % json.dest_key)
 
         # Extract partial dependence data from json response
         pps = json["partial_dependence_data"]
 
+        # Plot partial dependence plots using matplotlib
+        self.__generate_partial_plots(num_1dpdp, num_2dpdp, plot, server, pps, figsize, col_pairs_2dpdp, data, nbins,
+                                      kwargs["user_cols"], kwargs["num_user_splits"], plot_stddev, cols, save_to_file)
+        # totFig = num_1dpdp+num_2dpdp
+        # if plot and totFig>0:     # plot 1d pdp for now
+        #     plt = _get_matplotlib_pyplot(server)
+        #     if not plt: return pps
+        #     import matplotlib.gridspec as gridspec
+        #     fig = plt.figure(figsize=figsize)
+        #     gxs = gridspec.GridSpec(totFig, 1)
+        #     if num_2dpdp>0: # 2d pdp requested
+        #         axes3D = _get_mplot3d_pyplot("2D partial plots")
+        #         cm = _get_matplotlib_cm("2D partial plots")
+        #     figPlotted = 0  # indicated number of figures plotted
+        #     for i, pp in enumerate(pps):
+        #         if (i >= num_1dpdp): # plot 2D pdp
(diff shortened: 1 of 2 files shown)


h2o-py/h2o/model/model_base.py
```diff
@@ -1112,11 +1119,12 @@ def partial_plot(self, data, cols=None, destination_key=None, nbins=20, weight_c
             if num_2dpdp>0: # 2d pdp requested
                 axes3D = _get_mplot3d_pyplot("2D partial plots")
                 cm = _get_matplotlib_cm("2D partial plots")
-
+            figPlotted = 0  # indicated number of figures plotted
```
honzasterba: this looks like a boolean

h2o-py/tests/testdir_misc/pyunit_pubdev_6672_2D_pdp_fix.py
```diff
@@ -0,0 +1,23 @@
...
+
+'''
+Copied example from Honza to fix bug in 2D pdp.  Basically, when no figure is plotted, this bug will thrown an error.
+Hence, this code just needs to run to completion.  No assertion check is necessary for this test.
+'''
+def partial_plot_test_with_user_splits():
+    train = h2o.import_file(pyunit_utils.locate('smalldata/flow_examples/abalone.csv.gz'))
+    model = H2OGeneralizedLinearEstimator(training_frame=train)
+    model.train(y="C9")
+    model.partial_plot(train, col_pairs_2dpdp=[["C1", "C2"]], save_to_file="pdp.png")
```
honzasterba: should the test cleanup the created file? do we even need to save_to_file to reproduce this?

h2o-py/h2o/model/model_base.py
```diff
@@ -1140,21 +1148,26 @@ def partial_plot(self, data, cols=None, destination_key=None, nbins=20, weight_c
...
+            if (save_to_file is not None) and (figPlotted > 0): # only save when a figure is actually plotted
                 plt.savefig(save_to_file)
+            else:
+                print("No partial plot is generated and saved.  You may be missing toolboxes like "
+                  "mpl_toolkits.mplot3d, matplotlib")
 
-        return pps
-    
-    # change x, y, z to be 2-D numpy arrays in order to plot it.
+            # change x, y, z to be 2-D numpy arrays in order to plot it.
```
honzasterba: the indent seems wrong

h2o-py/h2o/model/model_base.py
```diff
@@ -1112,11 +1119,12 @@ def partial_plot(self, data, cols=None, destination_key=None, nbins=20, weight_c
             if num_2dpdp>0: # 2d pdp requested
                 axes3D = _get_mplot3d_pyplot("2D partial plots")
                 cm = _get_matplotlib_cm("2D partial plots")
-
+            figPlotted = 0  # indicated number of figures plotted
             for i, pp in enumerate(pps):
                 if (i >= num_1dpdp): # plot 2D pdp
```
honzasterba: I would also split the 2D and 1D PDP branches into separate functions and loops

h2o-py/h2o/model/model_base.py
wendycwong: Yeah, you are right.  Let me make it so.

h2o-py/tests/testdir_misc/pyunit_pubdev_6672_2D_pdp_fix.py
wendycwong: In this case, there is no file saved since no plot is generated.  No need to do clean up in this case.

h2o-py/h2o/model/model_base.py
wendycwong: Fixed

h2o-py/tests/testdir_misc/pyunit_pubdev_6672_2D_pdp_fix.py
wendycwong: We do not need to specify the save option to reproduce the error.

h2o-py/h2o/model/model_base.py
wendycwong: Fixed.

h2o-py/tests/testdir_misc/pyunit_pubdev_6672_2D_pdp_fix.py
honzasterba: thats right, im which case we should remove it from the test to avoid confusion

PR #3691, opened 2019-07-14
PUBDEV-6666: Add examples to GBM chapter

hannah-tillman:
Got it back down to 13 warnings locally

Again: missing examples for “custom_distribution_func”, “export_checkpoints_dir”,  “max_confusion_matrix”, “monotone_constraints”, “r2_stopping”, and “response_column”.


angela0xdata:
@hannah 
- You can find an example for custom_distribution_function here: https://github.com/h2oai/h2o-3/blob/master/h2o-docs/src/product/data-science/algo-params/custom_distribution_func.rst#example
- All examples already use response_column, so feel free to just borrow another one.
- An example for monotone_constraints can be found here: https://github.com/h2oai/h2o-3/blob/master/h2o-py/tests/testdir_misc/pyunit_xgboost_gbm_monotone.py
- An example for export_checkpoints_dir is here: https://github.com/h2oai/h2o-3/blob/master/h2o-py/tests/testdir_apis/H2O_Module/pyunit_h2ogrid_checkpoints.py
- r2_stopping is no longer supported. You can remove "Examples" from that.
- max_confusion_matrix is not supported.

PR #3692, opened 2019-07-15
[PUBDEV-6436] User friendly way to import MOJO models via Flow

honzasterba:


h2o-algos/src/main/java/hex/schemas/GenericV3.java
```diff
@@ -3,19 +3,18 @@
...
     public static final class GenericParametersV3 extends ModelParametersSchemaV3<GenericModelParameters, GenericParametersV3> {
         public static final String[] fields = new String[]{
                 "model_id",
-                "model_key"
+                "path"
         };
 
-        @API(direction = API.Direction.INOUT, level = API.Level.critical, help = "Key to the self-contained model archive already uploaded to H2O.")
-        public KeyV3.FrameKeyV3 model_key;
+        @API(required = true, level = API.Level.critical, help = "Path to file with self-contained model archive.")
```
Pscheidl: Removing model_key will definitely break APIs. What R and Python do is they import the file first and then use the file key to create the model.

honzasterba: I have updated the R and python APIs, do we need to be backwards compatible there?

Pscheidl: Ok, only the tests need fixing. Will do a patch for you.

h2o-r/h2o-package/R/models.R
```diff
@@ -4079,8 +4078,7 @@ h2o.genericModel <- function(model_file_path){
 #' }
 #' @export
 h2o.import_mojo <- function(mojo_file_path){
-  model_file_key <- h2o.importFile(mojo_file_path, parse = FALSE)
-  model <- h2o.generic(model_key = model_file_key)
+  model <- h2o.generic(path = mojo_file_path)
   print(model)
```
honzasterba: should this print be here?

Pscheidl: No there is no need, same case below, let's remove it.

Pscheidl:
There seems to be a problem with `fetchMojo` in `ModelsHandler`. After the changes, the call

`GET /3/Models/Generic_model_python_1563192004938_29/mojo, parms: {}` fails (the `/mojo` is not the root cause, it only directs the handler towards the correct method and has nothing to do with the new naming. On master, it currently behaves the same way and it works.)

h2o-r/h2o-package/R/models.R
```diff
@@ -4079,9 +4078,7 @@ h2o.genericModel <- function(model_file_path){
 #' }
 #' @export
 h2o.import_mojo <- function(mojo_file_path){
-  model_file_key <- h2o.importFile(mojo_file_path, parse = FALSE)
-  model <- h2o.generic(model_key = model_file_key)
-  print(model)
+  model <- h2o.generic(path = mojo_file_path)
   return(model)
```
Pscheidl: we could also return directly. Minor stuff.

Pscheidl:
Maybe looks like a problem with scopes with me, as the file is imported inside a scope started in Driver's compute2 method.

Pscheidl:
@michalkurka @honzasterba Please review. Danke !

honzasterba:
LGTM (cannot approve since this is originally my PR)

Pscheidl:
Shall we wait or shall we merge ? I approve too. Let's wait for the tests :rabbit:

honzasterba:
I think we should wait for tests and wait with merging for @michalkurka I am not sure more commits are allowed into the release.

michalkurka:
@honzasterba Let's wait with this one for after the release

I realized this would also need a docu change

PR #3696, opened 2019-07-15
Fix runit pubdev 4628 deeplearning mojo large r

wendycwong:
Veronica has caught test failure for runit_PUBDEV_4628_deeplearning_mojo.R and I caught it here: https://0xdata.atlassian.net/browse/PUBDEV-6680?filter=-2

I have done the following:
1. Broken deeplearning tests into autoencoders and non-autoencoders.
2. For each type, run through all combinations in terms of activation function, missing value handline, categorical encoding and others.

Next:
fix the bug.


wendycwong:
Opened another PR to replace this one.

PR #3697, opened 2019-07-16
Change seed value for GBMTest.testQuasiBinomial

wendycwong:
GBMTest failed here:  http://mr-0xc1:8080/view/H2O-3/job/h2o-3-nightly-pipeline/job/master/2912/testReport/junit/hex.tree.gbm/GBMTest/Java_11_JUnit___testQuasibinomial_1__gbm_EmulateConstraints__/

I relaxed the tolerance from 2e-3 to 2e-2 and 1e-3 to 1e-2.

### DIFF
--- a/h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
+++ b/h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
@@ -16,6 +16,7 @@ import hex.tree.SharedTreeModel;
 import org.junit.*;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 import water.*;
 import water.api.StreamingSchema;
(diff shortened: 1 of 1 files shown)


michalkurka:
@wendycwong should this also go into rel-yau?

wendycwong:
Talked to Arno.  He is not a big fan of changing the tolerance.  Instead, he proposed seed change.  I am going to try a few and we will see.

wendycwong:
Michal: 

Good point, it should go into rel-yau as it will fail all those tests as well.  I will rebase and come back again.  Thanks for pointing that out.   


Wendy

michalkurka:
@wendycwong do you know how is Java 11 causing the difference?

wendycwong:
Michalk:

It does seem like the Java version matters in this case.  However, I run with many seed values and found a bunch of them.

Basically random number generation is a finite state machine with initial seed (any value other than 0).  If they change the structure of the machine, the randomness will change.

PR #3698, opened 2019-07-16
[PUBDEV-6681] Refactor out method for obtaining logs in Java

jakubhava:
The main reason for this change is that in Sparkling Water, we are duplicating logic for downloading logs. With this change, we can remove the duplicated code.

### DIFF
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -2291,6 +2291,10 @@ final public class H2O {
     return downloadLogs(destinationDir.toString(), logContainer);
   }
 
+  public static URI downloadLogs(URI destinationDir, String logContainer) {
+    return downloadLogs(destinationDir, LogArchiveContainer.valueOf(logContainer));
+  }
+
   public static URI downloadLogs(String destinationDir, LogArchiveContainer logContainer) {
     String outputFileStem = LogsHandler.getOutputLogStem();
     String outputFileName = outputFileStem + "." + logContainer.getFileExtension();
@@ -2304,4 +2308,9 @@ final public class H2O {
     }
     return destination.toURI();
   }
+  
+  public static URI downloadLogs(String destinationDir, String logContainer) {
+    return downloadLogs(destinationDir, LogArchiveContainer.valueOf(logContainer));
+  }
+  
 }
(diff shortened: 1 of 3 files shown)


h2o-core/src/main/java/water/H2O.java
```diff
@@ -2298,4 +2286,31 @@ public static H2ONode getClientByIPPort(String ipPort){
...
+  public static URI downloadLogs(String destinationDir, LogArchiveContainer logContainer) {
+    String outputFileStem = LogsHandler.getOutputLogStem();
+    String outputFileName = outputFileStem + "." + logContainer.getFileExtension();
+    byte[] logBytes = LogsHandler.downloadLogs(logContainer, outputFileStem);
+    File destination = new File(destinationDir, outputFileName);
+
+    try (FileOutputStream fileOutputStream = new FileOutputStream(destination)) {
+      fileOutputStream.write(logBytes);
+    } catch (IOException e) {
+      e.printStackTrace();
```
Pscheidl: Shouldn't the reaction be more complex, e.g. use our `Log` class ?

h2o-core/src/main/java/water/api/LogsHandler.java
```diff
@@ -193,4 +192,111 @@ public LogsV3 fetch(int version, LogsV3 s) {
...
+        // Skip nodes that aren't healthy, since they are likely to cause the entire process to hang.
+        if (members[i].isHealthy()) {
+          GetLogsFromNode g = new GetLogsFromNode(i, logContainer);
+          g.doIt();
+          bytes = g.bytes;
+        } else {
+          bytes = StringUtils.bytesOf("Node not healthy");
+        }
+      }
+      catch (Exception e) {
```
Pscheidl: Minor: bad indentation. I just saw it, so I'm telling you. No big deal.

h2o-core/src/main/java/water/api/LogsHandler.java
```diff
@@ -193,4 +192,111 @@ public LogsV3 fetch(int version, LogsV3 s) {
...
+        archive.putNextEntry(ze);
+        archive.write(clientResult);
+        archive.closeEntry();
+      }
+
+      // Close the top-level directory.
+      archive.closeEntry();
+    } finally {
+      // Close the archive file.
+      archive.close();
```
Pscheidl: Minor: You can use try-with-resources as well.

h2o-core/src/main/java/water/H2O.java
jakubhava: Agree. The main purpose of this change was refactoring, I would like to avoid mixing fixes into this change. But I can improve this in the following PR, what do you think?

Pscheidl: Sure.

PR #3703, opened 2019-07-17
[PUBDEV-6685] Improve error handling in downloadLogs method

jakubhava:


jakubhava:
Thanks @Pscheidl , this was a good catch!

PR #3705, opened 2019-07-17
PUBDEV-6682 - Model metrics for IRF MOJO

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6682

These metrics are only for the IRF MOJO.

### DIFF
--- a/h2o-bindings/bin/gen_R.py
+++ b/h2o-bindings/bin/gen_R.py
@@ -198,8 +198,8 @@ def gen_module(schema, algo, module):
         yield "    beta_constraints <- as.h2o(beta_constraints)"
         yield "  }"
     elif algo == "generic":
-        yield "  # Required args: model_key"
-        yield "  if (is.null(model_key)) stop(\"argument 'model_key' must be provided\")"
+        yield "  # Required args: either model_key or path"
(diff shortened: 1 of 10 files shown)


h2o-bindings/bin/gen_R.py
```diff
@@ -198,8 +198,8 @@ def gen_module(schema, algo, module):
         yield "    beta_constraints <- as.h2o(beta_constraints)"
         yield "  }"
     elif algo == "generic":
-        yield "  # Required args: model_key"
-        yield "  if (is.null(model_key)) stop(\"argument 'model_key' must be provided\")"
+        yield "  # Required args: either model_key or path"
+        yield "  if (is.null(model_key) && is.null(path)) stop(\"argument 'model_key' or 'path' must be provided\")"
```
honzasterba: should the same be done for py?

Pscheidl: I'm doing that for all the algos in the very next PR. Will definitely be there.

Pscheidl: Python is more permissive and only checks against list of allowed algos, thus no hotfix is needed, but the check will be unified. There is a backend check as well btw.

Pscheidl: I'll backport the change here.

Pscheidl:
Do not resolve the issue just yet, there is one more PR coming.

PR #3706, opened 2019-07-17
PUBDEV-6561: fix deeplearning mojo test failure

wendycwong:
Veronica has caught test failure for runit_PUBDEV_4628_deeplearning_mojo.R and I caught it here: https://0xdata.atlassian.net/browse/PUBDEV-6680?filter=-2

I have done the following:

Broken deeplearning tests into autoencoders and non-autoencoders.
For each type, run through all combinations in terms of activation function, missing value handline, categorical encoding and others.
Next:
fix the bug.


wendycwong:
Moved PR to another place.  Sorry.

PR #3709, opened 2019-07-18
[XGBOOST] fix smoke tests on GPU

honzasterba:
I have verified that even xgboost native predict gives different results for gpu and cpu predictor. The worst diff encountered was on 3rd significant digit. Because of this I was forced to add frame comparison that would assert with relative precision - check n-th significant digit.

### DIFF
--- a/h2o-core/src/test/java/water/TestUtil.java
+++ b/h2o-core/src/test/java/water/TestUtil.java
@@ -18,7 +18,8 @@ import water.util.Log;
 import water.util.Timer;
 import water.util.TwoDimTable;
 
-import java.io.*;
+import java.io.File;
+import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.*;
@@ -670,11 +671,15 @@ public class TestUtil extends Iced {
   public static <T> T[] aro(T ...a) { return a ;}
 
   // ==== Comparing Results ====
-  
+
   public static void assertFrameEquals(Frame expected, Frame actual, double delta) {
+    assertFrameEquals(expected, actual, delta, false);
+  }
+
+  public static void assertFrameEquals(Frame expected, Frame actual, double delta, boolean relativeDelta) {
     assertEquals("Frames have different number of vecs. ", expected.vecs().length, actual.vecs().length);
     for (int i = 0; i < expected.vecs().length; i++) {
-      assertVecEquals(i + "/" + expected._names[i] + " ", expected.vec(i), actual.vec(i), delta);
+      assertVecEquals(i + "/" + expected._names[i] + " ", expected.vec(i), actual.vec(i), delta, relativeDelta);
     }
   }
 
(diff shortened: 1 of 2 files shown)


h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -13,6 +13,7 @@
 import water.parser.DefaultParserProviders;
 import water.parser.ParseDataset;
 import water.parser.ParseSetup;
+import water.rapids.ast.prims.reducers.AstCumu;
```
michalkurka: Unused import?

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -683,10 +688,20 @@ public static void assertVecEquals(Vec expecteds, Vec actuals, double delta) {
   }
 
   public static void assertVecEquals(String messagePrefix, Vec expecteds, Vec actuals, double delta) {
+    assertVecEquals(messagePrefix, expecteds, actuals, delta, false);
+  }
+
+  public static void assertVecEquals(String messagePrefix, Vec expecteds, Vec actuals, double delta, boolean relativeDelta) {
```
michalkurka: Another way to do this would be to allow specifying both absolute and relative deltas and use both if both specified or just one of them if one of them is NaN.

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -683,10 +688,20 @@ public static void assertVecEquals(Vec expecteds, Vec actuals, double delta) {
...
     assertEquals(expecteds.length(), actuals.length());
     for(int i = 0; i < expecteds.length(); i++) {
       final String message = messagePrefix + i + ": " + expecteds.at(i) + " != " + actuals.at(i) + ", chunkIds = " + expecteds.elem2ChunkIdx(i) + ", " + actuals.elem2ChunkIdx(i) + ", row in chunks = " + (i - expecteds.chunkForRow(i).start()) + ", " + (i - actuals.chunkForRow(i).start());
-      assertEquals(message, expecteds.at(i), actuals.at(i), delta);
+      double expectedVal = expecteds.at(i);
+      double actualVal = actuals.at(i);
+      double assertionDelta = delta;
+      if (relativeDelta) {
+        assertionDelta = Math.min(expectedVal, actualVal) * delta;
```
michalkurka: How does this work for negative numbers or if the minimum is 0? Then assertionDelta = 0. Which means exact match.

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostPredictImplComparisonTest.java
```diff
@@ -97,14 +93,17 @@ public void testPredictionsAreSame() throws IOException {
...
-            assertFrameEquals(predsNative, predsJava, getDelta());
+            assertFrameEquals(predsNative, predsJava, getDelta(), true);
         } finally {
             Scope.exit();
         }
     }
     
     private double getDelta() {
-        if ("gbtree".equals(booster)) {
+        if (XGBoost.hasGPU(H2O.CLOUD.members()[0], 0)) {
```
michalkurka: we should probably make the decision based on if GPU was actually used to train the algo not just if it is available? it might be the same thing in the current state of the test - but if things changed it wouldn't be obvious how is it actually being tested

honzasterba: fixed

h2o-core/src/test/java/water/TestUtil.java
honzasterba: good point, updated, the whole thing with min() was wrong, we should check the delta relative to the expected value only and use absolute value of the expected

h2o-core/src/test/java/water/TestUtil.java
honzasterba: I find this approach more user-friendly and I cannot imagine a case when you would need both

michalkurka: this would make it consistent with Model#testJavaScoring: double rel_epsilon, double abs_epsilon

for regression problems you might want to bind the absolute error as well as the relative error

PR #3712, opened 2019-07-18
Docs: Instructions to add cloud-boothook to force execution of user-data

pratapram:
User-data will not be run on custom AMIs automatically but can be forced with "#cloud-boothook".


angela0xdata:
This change should also go into rel-yau.

PR #3717, opened 2019-07-19
Fixed argument check for Generic model in Python

Pscheidl:
Discovered and fixed by @seb-h2o 

The condition could never be true. We only wanted to verify two of the arguments. Backend checks are present as well.

### DIFF
--- a/h2o-bindings/bin/gen_python.py
+++ b/h2o-bindings/bin/gen_python.py
@@ -184,8 +184,8 @@ def gen_module(schema, algo):
     yield "        names_list = {%s}" % bi.wrap(", ".join('"%s"' % p for p in param_names),
                                                 indent=(" " * 22), indent_first=False)
     if(algo == "generic"):
-        yield '        if(all(name is None for name in names_list)):'
-        yield '                raise H2OValueError("At least one of model-generic")'
+        yield '        if(all(kwargs.get(name, None) is None for name in [ "model_key", "path"])):'
+        yield '                raise H2OValueError("At least one of [\\"model_key\\", \\"path\\"] is required.")'
     yield '        if "Lambda" in kwargs: kwargs["lambda_"] = kwargs.pop("Lambda")'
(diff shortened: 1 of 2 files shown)


Pscheidl:
Waiting for tests, then merging.

h2o-bindings/bin/gen_python.py
```diff
@@ -184,8 +184,8 @@ def gen_module(schema, algo):
     yield "        names_list = {%s}" % bi.wrap(", ".join('"%s"' % p for p in param_names),
                                                 indent=(" " * 22), indent_first=False)
     if(algo == "generic"):
-        yield '        if(all(name is None for name in names_list)):'
-        yield '                raise H2OValueError("At least one of model-generic")'
+        yield '        if(all(kwargs.get(name, None) is None for name in [ "model_key", "path"])):'
+        yield '                raise H2OValueError("At least one of [\\"model_key\\", \\"path\\"] is required.")'
```
seb-h2o: you could avoid \\" by using single quotes if you want.

h2o-py/h2o/estimators/generic.py
```diff
@@ -24,8 +24,8 @@ def __init__(self, **kwargs):
         super(H2OGenericEstimator, self).__init__()
         self._parms = {}
         names_list = {"model_id", "model_key", "path"}
-        if(all(name is None for name in names_list)):
-                raise H2OValueError("At least one of model-generic")
+        if(all(kwargs.get(name, None) is None for name in [ "model_key", "path"])):
+                raise H2OValueError("At least one of [\"model_key\", \"path\"] is required.")
```
seb-h2o: btw ideally, parameters validation should go in the `train` method because user can usually create an estimator with empty constructor and set the params using the setters (this is actually the standard way when used in a sklearn pipeline), but we don't have any hook today to ensure that validation is done in the right place: I'll probably enforce this in my ongoing PR.

PR #3721, opened 2019-07-19
PUBDEV-6694: add AUCPR as early stopping metric

wendycwong:
This PR completes the work for JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6684?filter=myopenissues&orderby=priority%20DESC

Work done:
1. Added java backend support to allow AUCPR as an early-stopping parameter.
2. Added Runit and Pyunit test to ensure correct operation.

@ledell @NidhiMehta 
Please let me know if AUCPR is an okay name for the parameter.  I can change it to anything you guys want.  Just let me know please.

### DIFF
--- a/h2o-core/src/main/java/hex/ScoreKeeper.java
+++ b/h2o-core/src/main/java/hex/ScoreKeeper.java
@@ -126,9 +126,10 @@ public class ScoreKeeper extends Iced {
     RMSE(ConvergenceStrategy.LESS_IS_BETTER, true),
     MAE(ConvergenceStrategy.LESS_IS_BETTER, true),
     RMSLE(ConvergenceStrategy.LESS_IS_BETTER, true),
-    AUC(ConvergenceStrategy.MORE_IS_BETTER, false),
+    AUC(ConvergenceStrategy.MORE_IS_BETTER, true),
     lift_top_group(ConvergenceStrategy.MORE_IS_BETTER, false),
     misclassification(ConvergenceStrategy.LESS_IS_BETTER, true),
+    AUCPR(ConvergenceStrategy.MORE_IS_BETTER, true),
     mean_per_class_error(ConvergenceStrategy.LESS_IS_BETTER, true),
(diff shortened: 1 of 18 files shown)


h2o-core/src/main/java/hex/ScoreKeeper.java
```diff
@@ -126,9 +126,10 @@ else if (m instanceof ModelMetricsMultinomial) {
     RMSE(ConvergenceStrategy.LESS_IS_BETTER, true),
     MAE(ConvergenceStrategy.LESS_IS_BETTER, true),
     RMSLE(ConvergenceStrategy.LESS_IS_BETTER, true),
-    AUC(ConvergenceStrategy.MORE_IS_BETTER, false),
```
michalkurka: the lower bound only matters to LESS_IS_BETTER, imho it doesn't matter whether we have true or false there - but obviously AUC is actually lower bound by 0

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -6,6 +6,7 @@
 import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.rules.TestRule;
+import org.junit.runner.Computer;
```
michalkurka: unused import?

michalkurka:
@wendycwong the commit history includes other people's commits

can you please rebase this on top of rel-yau and set the target branch to rel-yau?

h2o-core/src/main/java/hex/ScoreKeeper.java
wendycwong: Got it.

wendycwong: Can I keep it as true though?  It looks better.

michalkurka: yes, no harm there - I agree it does look better

PR #3722, opened 2019-07-19
PUBDEV-6696 fix hex.deeplearning.DeepLearningCheckpointReporting.run

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6696

Work done:  I removed the following assertation:

Assert.assertTrue("Duration must be smooth; actual " + diff + ", expected at most " + sleepTime + " (before=" + durationBefore + ", after=" + durationAfter + ")", diff < sleepTime+1);

This one seems to assert the time difference between model was first trained and then train again with checkpoint.  However, it does nothing to verify checkpointing itself.  Our jenkins system seems to be heavily loaded these days and execution time is not guaranteed for multi-node, multi-thread system.

### DIFF
--- a/h2o-algos/src/test/java/hex/deeplearning/DeepLearningCheckpointReporting.java
+++ b/h2o-algos/src/test/java/hex/deeplearning/DeepLearningCheckpointReporting.java
@@ -14,6 +14,7 @@ import water.fvec.Frame;
 import water.fvec.NFSFileVec;
 import water.parser.ParseDataset;
 import water.util.TwoDimTable;
+import water.H2O;
 
 public class DeepLearningCheckpointReporting extends TestUtil {
   @BeforeClass() public static void setup() { stall_till_cloudsize(1); }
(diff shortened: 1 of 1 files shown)


michalkurka:
@wendycwong imho only the last commit is relevant to this change

h2o-algos/src/test/java/hex/deeplearning/DeepLearningCheckpointReporting.java
```diff
@@ -101,7 +98,7 @@
           durationAfter = durationAfter.substring(0, durationAfter.length()-4);
           double diff = Double.parseDouble(durationAfter) - Double.parseDouble(durationBefore);
           
-          Assert.assertTrue("Duration must be smooth; actual " + diff + ", expected at most " + sleepTime + " (before=" + durationBefore + ", after=" + durationAfter + ")", diff < sleepTime+1);
+          // Assert.assertTrue("Duration must be smooth; actual " + diff + ", expected at most " + sleepTime + " (before=" + durationBefore + ", after=" + durationAfter + ")", diff < sleepTime+1);
```
michalkurka: maybe we can keep this assert at least for single node? might be not worth it but still worth considering

h2o-algos are test both 1-node and multi-node

wendycwong: Good point.  Let me change it to reflect that.

wendycwong: Let me change it and see if it fails any test and we can discuss then.

wendycwong:
Michalk:  I messed up the base branch with this one too.  Good thing  you caught it.

PR #3725, opened 2019-07-20
PUBDEV-6319: Fix a race condition in group-by implementation

michalkurka:


wendycwong:
Do you want to merge this into rel-yau?  Need to fix it there too, no?  Wendy

michalkurka:
> So you just use a Hashset instead but with it delegating all works to the original nonblockingHashMap iced. Great! I would like to do a timing test after this is merge.

Exactly... that prevents users to use the HashMap the wrong way. But that is just a syntactic sugar; the core of the fix is in the first comment where I swapped working with keys (that can change) to values that are immutable once inserted.

michalkurka:
> Do you want to merge this into rel-yau? Need to fix it there too, no? Wendy

Yes, absolutely correct - I will merge into master first and when appropriate backport to a desired Yau fix release.

I didn't see any changes in performance in my tests but I will reach out to you and we can coordinate tests with your test suite.

michalkurka:
@wendycwong I couldn't do full tests yet but multinode tests show the performance is definitely not worse (times in seconds):

New:
Time taken to perform groupby is 129.045972824
Time taken to perform groupby is 139.069935083
Time taken to perform groupby is 136.047094822

Old:
Time taken to perform groupby is 152.91086483
Time taken to perform groupby is 162.909729958
Time taken to perform groupby is 171.010210991

This could be attributed to a faster serialization - we are only serializing the values before we were serializing both keys and values.

wendycwong:
@michalk:  Nice side effect.  You fixed the main problem and then gain more speed along the way.  Wish I could do that!

PR #3731, opened 2019-07-22
PUBDEV-6697 fix GBMTest.testQuasibinomial failures

wendycwong:
This PR completes the task in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6697?filter=-1.

Instead of trying out one seed value, I am allowing the test to try out more values.


wendycwong:
Veronica will take care of this one.

PR #3732, opened 2019-07-22
PUBDEV-6675: Python Docs: Add examples to DRF chapter

hannah-tillman:


h2o-py/h2o/estimators/random_forest.py
```diff
@@ -1568,6 +1568,31 @@ def export_checkpoints_dir(self):
         Automatically export generated models to this directory.
 
         Type: ``str``.
+
+        :examples:
+
+        >>> airlines = h2o.import_file("http://s3.amazonaws.com/h2o-public-test-data/smalldata/airlines/allyears2k_headers.zip", destination_frame="air.hex")
```
angela0xdata: For this example, we should probably include the required import statements at the beginning:
- import tempfile
- from os import listdir
- from h2o.grid.grid_search import H2OGridSearch

PR #3734, opened 2019-07-22
PUBDEV-6698: Add "aucpr" to list of stopping_metric options

zainhaq-h2o:
Updated documentation to reflect the addition of "aucpr" to the list of stopping_metric options.

https://0xdata.atlassian.net/browse/PUBDEV-6698


angela0xdata:
@zainhaq-h2o Minor detail. We try to arrange the list of options so that it matches the help. 
See this example: https://github.com/h2oai/h2o-3/blob/rel-yau/h2o-py/h2o/estimators/random_forest.py#L486

PR #3735, opened 2019-07-23
PUBDEV-6701: Use sort metric enum instead of string occurrences

jakubhava:
https://0xdata.atlassian.net/browse/PUBDEV-6701

Experimental state, cf. ticket description.

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -878,7 +878,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     if (parms._stopping_metric == StoppingMetric.AUTO) {
       SortMetric sort_metric = getSortMetric();
       parms._stopping_metric = sort_metric == null ? StoppingMetric.AUTO
-                              : sort_metric.equals("auc") ? StoppingMetric.logloss
+                              : sort_metric == SortMetric.auc ? StoppingMetric.logloss
                               : metricValueOf(sort_metric);
     }
 
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -290,7 +290,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
       }
     }
 
-      String sort_metric = buildSpec.input_spec.sort_metric == null ? null : buildSpec.input_spec.sort_metric.toLowerCase();
+      SortMetric sort_metric = buildSpec.input_spec.sort_metric == null ? null : SortMetric.valueOf(buildSpec.input_spec.sort_metric.toLowerCase());
       leaderboard = Leaderboard.getOrMake(projectName(), eventLog, this.leaderboardFrame, sort_metric);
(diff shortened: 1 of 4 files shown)


h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -876,7 +876,7 @@ private void setStoppingCriteria(Model.Parameters parms, Model.Parameters defaul
       parms._stopping_metric = buildSpec.build_control.stopping_criteria.stopping_metric();
 
     if (parms._stopping_metric == StoppingMetric.AUTO) {
-      String sort_metric = getSortMetric();
+      SortMetric sort_metric = getSortMetric();
```
seb-h2o: also needs to be changed below: `sort_metric = SortMetric.auc`

h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
```diff
@@ -162,22 +162,22 @@ private void setDefaultMetricAndDirection(Model m) {
     if (m._output.isBinomialClassifier()) { //Binomial
       metrics = new String[]{"logloss", "mean_per_class_error", "rmse", "mse"};
```
seb-h2o: Ideally use the enum in those string arrays as well

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
jakubhava: 👍

h2o-automl/src/main/java/ai/h2o/automl/SortMetric.java
```diff
@@ -0,0 +1,12 @@
+package ai.h2o.automl;
+
+public enum SortMetric {
```
seb-h2o: the main reason I'm hesitating is because it partly duplicates `StoppingMetric`, without the useful logic in the latter that could be used to remove duplicate logic in `Leaderboard`, and without the potential support for custom metric. Also, it would need AUTO.
I would prefer having this new enum used only in `AutoMLV99` and have it immediately converted to `StoppingMetric` and then have all the logic using only `StoppingMetric`

jakubhava: Sounds good!, Big fan of introducing `AUTO`. 

@seb-h2o Should I leave this PR open for you to work on it? I would like to spend on more urgent features in Sparkling Water

jakubhava: I just wanted to try and see if 5-minute solution would do, but the issue is more complex. Would rather let AutoML dev to deal with it

h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
jakubhava: good catch

h2o-automl/src/main/java/ai/h2o/automl/SortMetric.java
seb-h2o: sure, please leave it, I'm filing a Jira.
Thx @jakubhava

jakubhava:
Closing -> will be better if AutoML engineer starts from scratch. Want to avoid stall PR

PR #3737, opened 2019-07-23
[PUBDEV-6700] Set defaultValue to missingValuesHandling on XGBoostParameters

jakubhava:
Discovered that `missingValuesHandling` has default value as `null`. I believe, if we use `Enum` We should not use `null` as valid value, instead create new enum field.

Based on the current code, it seems that `null` has the same meaning as `MeanImputation`.

This change will also help on Sparkling Water side as I wanted to eliminate possibility of setting nulls on our pipeline wrappers when possible.

Please let me know if this makes sense & or if I'm wrong somewhere.

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
@@ -73,7 +73,7 @@ public class XGBoostModel extends Model<XGBoostModel, XGBoostModel.XGBoostParame
 
     // H2O GBM options
     public boolean _quiet_mode = true;
-    public MissingValuesHandling _missing_values_handling;
+    public MissingValuesHandling _missing_values_handling = MissingValuesHandling.MeanImputation;
 
     public int _ntrees=50; // Number of trees in the final model. Grid Search, comma sep values:50,100,150,200
     public int _n_estimators;  // This doesn't seem to be used anywhere... (not in clients)


angela0xdata:
@jakubhava we currently have an FAQ in XGBoost regarding missing values handling. Will this FAQ need to change as a result?

"How does the algorithm handle missing values?

Missing values are interpreted as containing information (i.e., missing for a reason), rather than missing at random. During tree building, split decisions for every node are found by minimizing the loss function and treating missing values as a separate category that can go either left or right. XGBoost will automatically learn which is the best direction to go when a value is missing."

jakubhava:
@angela0xdata this seems to confirm that `MeanImputation` is the default. I think it won't need to be changed

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -73,7 +73,7 @@
 
     // H2O GBM options
     public boolean _quiet_mode = true;
-    public MissingValuesHandling _missing_values_handling;
+    public MissingValuesHandling _missing_values_handling = MissingValuesHandling.MeanImputation;
```
michalkurka: XGBoost doesn't actually expose this option to users

https://github.com/h2oai/h2o-3/blob/master/h2o-extensions/xgboost/src/main/java/hex/schemas/XGBoostV3.java#L169

IMHO this should be removed from the code base completely because it doesn't make sense in case of XGBoost - probably copied from GLM because of DataInfo re-use.

@honzasterba or @Pscheidl - can you guys fix this in H2O?

michalkurka:
XGBoost doesn't do any mean imputation even when the option is flipped to MeanImputation, this is just very confusing piece of code :(

jakubhava:
Thank you @michalkurka! I was just trying to fix simple issue, but didn't know it goes this far.

michalkurka:
@jakubhava understood, thanks - if this parameter is exposed in SW - we should remove it there as well, it really doesn't have any actual purpose

jakubhava:
Yup, I already removed it from the API on SW side

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
honzasterba: will do https://0xdata.atlassian.net/browse/PUBDEV-6724

PR #3744, opened 2019-07-25
PUBDEV-6715: sklearn wrappers for AutoML and other Estimators

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6715

- expose a new Python module `h2o.sklearn`: the module makes use of `sklearn` but doesn't create any `sklearn` dependency for other modules.
- the module automatically generate sklearn wrappers for H2OAutoML and other estimators:
  - `H2OAutoML` will generate a generic `h2o.sklearn.H2OAutoMLEstimator`, an `h2o.sklearn.H2OAutoMLClassifier` and an `h2o.sklearn.H2OAutoMLRegressor`,
  - `H2OGradientBoostingEstimator` will generate a generic `h2o.sklearn.H2OGradientBoostingEstimator`, an `h2o.sklearn.H2OGradientBoostingClassifier` and an `h2o.sklearn.H2OGradientBoostingRegressor`.
  - unsupervised estimators generate only a generic Estimator.
  - transformers keep their original name.
- no duplicate logic, duplicate parameters and so on. It's only based on meta-programming: any parameter change in the original estimators (or in the estimators list in `estimators` module) will be reflected in the dynamically generated `h2o.sklearn` classes: generated on first import of the module.
- the generated wrappers automatically handle:
    - the connection with the backend.
    - the conversion from arrays to `H2OFrame` for `fit`, `transform` and `predict` methods.
    - the conversion back to numpy arrays for `predict` methods.
    - default `score` logic.
    - the wrappers can also deal in a H2O context: no conversion, nor back-conversion when using H2OFrames.
- support for `get_params`, `set_params` (includes **all** params, as recommended by `sklearn`, not only the ones set at instantiation) and usage in a `sklearn.Pipeline` 
- support for params discovery (auto-completion) in Jupyter/iPython environment.

status:
- AutoML and H2O estimators work on both Py3 and Py2.7, can be integrated into an sklearn pipeline containing sklearn transformers and/or a sklearn Grid, and using datasets in numpy format.
- compatibility with both `sklearn` and existing `H2O` transformers .

to REVIEWERS:
- all generated algos Py files have been regenerated (it's basically half the modified files here), due to the list of param names moved from constructor local scope to class-level scope: this is because those params are needed to dynamically generate the constructor of the wrapper (to allow auto-completion and feature discovery): we could probably think about having proper constructors for those original algos as well, feature discovery is important imo.
- see `h2o/sklearn/__init__.py` for the newly exposed API, dynamically generated.
- see `wrapper.py` for the wrapping logic common to all estimators and transformers.
- see tests for usage examples in `sklearn.Pipeline` or search contexts.


h2o-bindings/bin/gen_python.py
```diff
@@ -177,14 +177,14 @@ def gen_module(schema, algo):
     yield "    \"\"\""
     yield ""
     yield '    algo = "%s"' % algo
+    yield "    param_names = {%s}" % bi.wrap(", ".join('"%s"' % p for p in param_names),
+                                             indent=(" " * 19), indent_first=False)
```
seb-h2o: exposing param names as Py class property: needed to generate sklearn wrappers.

h2o-bindings/bin/gen_python.py
```diff
@@ -662,9 +662,8 @@ def gen_init(modules):
     yield "#"
     module_strs = []
     for module, clz, category in sorted(modules):
-        if clz == "H2OGridSearch": continue
-        module_strs.append('"%s"' % clz)
```
seb-h2o: was generating duplicates: cf. `estimators/__init__.py` below

h2o-py/h2o/model/regression.py
```diff
@@ -106,8 +106,8 @@ def h2o_r2_score(y_actual, y_predicted, weights=1.):
     :returns: R-squared (best is 1.0, lower is worse).
     """
     ModelBase._check_targets(y_actual, y_predicted)
-    numerator = (weights * (y_actual - y_predicted) ** 2).sum()
-    denominator = (weights * (y_actual - _colmean(y_actual)) ** 2).sum()
```
seb-h2o: this score method was ALWAYS returning `1.0` (that's how we cheat on benchmarks..) as `numerator` and `denominator` were still frames when the logic below expected numbers..

h2o-py/h2o/transforms/decomposition.py
```diff
@@ -70,8 +69,8 @@ def __init__(self, model_id=None, k=None, max_iterations=None, seed=None,
 
         """
         super(H2OPCA, self).__init__()
-        self._parms = locals()
-        self._parms = {k: v for k, v in self._parms.items() if k != "self"}
+        self._parms = locals()  # must be executed first, even parent constructor below add things to local scope
+        self._parms = {k: v for k, v in self._parms.items() if k not in ('self', '__class__')}
```
seb-h2o: Py2 also adds a nasty `__class__` to local context during construction

h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -391,7 +392,7 @@ def get_params(self, deep=True):
         :returns: A dict of parameters
         """
         out = dict()
-        for key, value in self.parms.items():
+        for key, value in self._parms.items():
```
seb-h2o: apparently `get_params` has **always** been returning the wrong set of params, making is unusable in several sklearn contexts.

h2o-py/h2o/sklearn/wrapper.py
```diff
@@ -0,0 +1,487 @@
...
+            return result
+
+        return wrapper
+
+    return decorator
+
+
+class BaseEstimatorMixin(object):
+
+    _classifier_distributions = ('bernoulli', 'binomial', 'quasibinomial', 'multinomial')
```
maurever: In h2o core, you can set distribution to `bernoulli`, `quasibinomial` and `multinomial` to get classifier, no `binomial` option.

seb-h2o: actually, GLM  is the exception: it is using `binomial` instead of `bernoulli` for its `family` parameter :(

seb-h2o:
@maurever I'm done with implementation and tests: was definitely more painful than imagined first, mainly to a lack of standardisation in some of our estimators.
I'll add more code documentation, thanks for noticing.

h2o-py/tests/testdir_sklearn/pyunit_sklearn_regression_all_estimators.py
```diff
@@ -0,0 +1,170 @@
...
+        test_scores_are_equivalent
+    ])
+
+
+failing = [
+    'H2OCoxProportionalHazardsRegressor',  # doesn't support regression?
+    'H2ODeepWaterRegressor',  # requires DW backend
+    'H2OStackedEnsembleRegressor',  # needs a separate test (requires models as parameters),
+    'H2OSupportVectorMachineRegressor',  # doc says that our SVM estimator also supports regression, but it seems to require categorical target
+]
```
seb-h2o: reviewers, please confirm if this is normal...

h2o-py/tests/testdir_sklearn/pyunit_sklearn_generic_all_estimators.py
```diff
@@ -0,0 +1,208 @@
...
+
+failing = [
+    'H2OAggregatorEstimator',  # can't use predict with it: looks more like a transformer, maybe should be removed from sklearn API
+    'H2OCoxProportionalHazardsEstimator',  # NPE at water.fvec.Frame.<init>(Frame.java:168) .... at hex.coxph.CoxPH$CollectTimes.collect(CoxPH.java:805)
+    'H2ODeepWaterEstimator',  # requires DW backend
+    'H2OGenericEstimator',  # maybe should be removed from sklearn API
+    'H2OStackedEnsembleEstimator',  # needs a separate test (requires models as parameters)
+    'H2OSingularValueDecompositionEstimator', # H2OResponseError: Server error water.exceptions.H2ONotFoundArgumentException: Error: POST /3/ModelBuilders/svd not found
+    'H2OWord2vecEstimator',  # needs a separate test (requires pre_trained model as parameter)
+]
```
seb-h2o: reviewers, please confirm if this is normal...

h2o-py/tests/testdir_sklearn/pyunit_sklearn_classification_all_estimators.py
```diff
@@ -0,0 +1,180 @@
...
+        test_estimator_with_numpy_arrays,
+        test_scores_are_equivalent
+    ])
+
+
+failing = [
+    'H2OCoxProportionalHazardsClassifier',  # NPE at water.fvec.Frame.<init>(Frame.java:168) .... at hex.coxph.CoxPH$CollectTimes.collect(CoxPH.java:805)
+    'H2ODeepWaterClassifier',  # requires DW backend
+    'H2OStackedEnsembleClassifier',  # needs a separate test (requires models as parameters)
+]
```
seb-h2o: reviewers, please confirm if this is normal...

h2o-py/tests/testdir_sklearn/pyunit_sklearn_regression_all_estimators.py
```diff
@@ -0,0 +1,170 @@
...
+    defaults = dict(
+        H2OCoxProportionalHazardsRegressor=dict(),
+        H2OGeneralizedLinearRegressor=dict(family='gaussian', seed=seed),
+    )
+    return defaults.get(estimator_cls.__name__, dict(seed=seed))
+
+
+def _get_custom_behaviour(estimator_cls):
+    custom = dict(
+        H2ODeepLearningRegressor=dict(scores_may_differ=True),
```
seb-h2o: deep learning with default params (but with a seed) doesn't look reproducible: doesn't look normal

h2o-py/tests/testdir_sklearn/pyunit_sklearn_classification_all_estimators.py
```diff
@@ -0,0 +1,180 @@
...
+    defaults = dict(
+        H2OCoxProportionalHazardsClassifier=dict(),
+        H2OGeneralizedLinearClassifier=dict(family='binomial', seed=seed),
+    )
+    return defaults.get(estimator_cls.__name__, dict(seed=seed))
+
+
+def _get_custom_behaviour(estimator_cls):
+    custom = dict(
+        H2ODeepLearningClassifier=dict(scores_may_differ=True),
```
seb-h2o: deep learning with default params (but with a seed) doesn't look reproducible on classification: doesn't look normal

h2o-py/tests/testdir_sklearn/pyunit_sklearn_automl_pipeline.py
```diff
@@ -0,0 +1,210 @@
+from __future__ import print_function
```
seb-h2o: @ledell This test file shows how it would work with AutoML.
The pipelines are simpler than than with `h2o-py/tests/testdir_sklearn/pyunit_sklearn_classification_pipeline.py` for example (didn't use transformers here for simplicity), but more complex pipelines or even sklearn searches would work as well here as for other H2O estimators.
In those tests, I just verified the various use cases:
- connection handling (autostart).
- data input format: H2OFrame vs. numpy arrays.
- setting parameters the sklearn way: constructor, set_params, as properties.
- sklearn API: `fit`, `predict`, `predict_proba`, `score`

ledell: @seb-h2o Do this support `pandas.DataFrame` in addition to numpy arrays?

h2o-py/h2o/estimators/__init__.py
```diff
@@ -26,16 +26,11 @@
 from .xgboost import H2OXGBoostEstimator
 
 __all__ = (
-    "H2OAggregatorEstimator", "H2OAggregatorEstimator", "H2OAutoML", "H2OCoxProportionalHazardsEstimator",
```
michalkurka: :)

h2o-py/h2o/model/regression.py
michalkurka: most likely not used by anyone :)

h2o-py/h2o/sklearn/__init__.py
```diff
@@ -0,0 +1,146 @@
+import inspect
+import sys
+
+from sklearn.base import ClassifierMixin, RegressorMixin
```
michalkurka: does this make us dependent on sklearn?

I didn't notice any changes in setup.py

h2o-py/tests/testdir_sklearn/pyunit_sklearn_regression_all_estimators.py
michalkurka: SVM only supports binary classification

cc: @angela0xdata

h2o-py/tests/testdir_sklearn/pyunit_sklearn_generic_all_estimators.py
maurever: SVD error is fixed [here](https://github.com/h2oai/h2o-3/pull/3757), wrong api version was called.

h2o-py/h2o/sklearn/__init__.py
seb-h2o: @michalkurka only when importing the `h2o.sklearn` module, which is fair enough I guess

h2o-py/tests/testdir_sklearn/pyunit_sklearn_automl_pipeline.py
seb-h2o: @ledell for data input yes, we can use `pandas.DataFrame`, but the auto-converted output for `predict` and `transform` will still be `numpy` arrays for now.
I need to double check what `sklearn` does currently: my first idea was do support `pandas` input -> `pandas` output, but I think that even `sklearn` does `pandas` input -> `numpy` output, that's why I removed conversion to `pandas`: should not be difficult to reintroduce though.

h2o-py/tests/testdir_sklearn/pyunit_sklearn_regression_all_estimators.py
angela0xdata: @michalkurka http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/svm.html#svm-algorithm
>>  This implementation can be used to solve binary classification problems.

I'll make it more prominent, though.

h2o-py/tests/testdir_sklearn/pyunit_sklearn_automl_pipeline.py
seb-h2o: I confirm that `sklearn` itself doesn't return `pandas.DataFrame` on `predict` and `transform` methods, so I don't need to support back-conversion to pandas.
However, it looks like my conversion from `pandas.DataFrame` as input data is broken (was relying on H2OFrame support for pandas but header is handled correctly only on some conditions): will fix this and add a test for `pandas`.

seb-h2o: pandas support fixed: it was a corner case in our H2OFrame pandas support when pandas DataFrame column names are not strings (e.g. when dataframe is directly created from numpy array)

h2o-py/h2o/sklearn/__init__.py
michalkurka: yes, that makes good sense - if you want to use sklearn functionality - you should have sklearn

makes me think that we should also have a minimal docker image with just the minimal requirements and run smoke tests with just minimum requirements as well

h2o-py/tests/testdir_sklearn/pyunit_sklearn_regression_all_estimators.py
michalkurka: thanks @angela0xdata 

@seb-h2o where did you see in doc SVM supports only regression?

h2o-py/h2o/sklearn/wrapper.py
michalkurka: yeah, GLM has it's own enum (which is sometimes a good thing but not in this case ;))

h2o-py/tests/testdir_sklearn/pyunit_sklearn_regression_all_estimators.py
seb-h2o: http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/svm.html#introduction
```
support vector machines are supervised learning models with associated learning algorithms that analyze data used for classification and regression analysis
```
that's why I thought we were also supporting SVR

seb-h2o: I removed the SVM regressor.

h2o-py/tests/testdir_sklearn/pyunit_sklearn_generic_all_estimators.py
seb-h2o: Reenabled it, seems to work now, thanks.

h2o-py/tests/testdir_sklearn/pyunit_sklearn_regression_all_estimators.py
michalkurka: deep learning is not reproducible by default even with a given seed

seb-h2o: always thought that is was on single node.
@ledell FYI

michalkurka: yeah, not even on a single node - to make it reproducible you need to use flag `reproducible=true`, http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/deep-learning.html

maurever:
Great work, @seb-h2o! Looks good to me, thanks for adding more doc.

One thing - there is a problem with transformer estimators (H2OPrincipalComponentAnalysisEstimator, H2OSingularValueDecompositionEstimator, H2OAggregatorEstimator, H2OGeneralizedLinearEstimator) which are estimators but with 'transform' property - the problem is described in this JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6144.

Your solution does not solve this problem, and if you use H2OSingularValueDecompositionEstimator instead of H2OSVD, it will fail. 

I temporarily solved this problem for PCA and SVD (https://0xdata.atlassian.net/browse/PUBDEV-6725), however, it does not work here. And it is not a nice solution, and also it is not applicable for Aggregator and GLM. 

Do you have any other idea connected to your wrapper to solve this problem? The worst solution is to refactor the 'transform' parameter. :( However, I don't have any other idea.

seb-h2o:
@maurever I see, you want to also expose `H2OAggregatorEstimator`, `H2OGeneralizedLowRankEstimator`, `H2OPrincipalComponentAnalysisEstimator` and `H2OSingularValueDecompositionEstimator` as estimators and transformers.
In which case, we probably wouldn't even need `H2OPCA` and `H2OSVD`.

I didn't add new methods to existing estimators yet, but it would be easy to do for those 4: at least `transform` + `fit_transform`

maurever:
> @maurever I see, you want to also expose `H2OAggregatorEstimator`, `H2OGeneralizedLowRankEstimator`, `H2OPrincipalComponentAnalysisEstimator` and `H2OSingularValueDecompositionEstimator` as both final estimators and transformers.
> In which case, we probably wouldn't even need `H2OPCA` and `H2OSVD`.
> 
> I didn't add new methods to existing estimators yet, but it would be easy to do for those 4: at least `transform` + `fit_transform`

Is not really a problem to add these methods, especially the `transform` method? Because there is `transform` property in these estimators and that causes the problem. 

```python
@property
    def transform(self):
        """
        Transformation of training data

        One of: ``"none"``, ``"standardize"``, ``"normalize"``, ``"demean"``, ``"descale"``  (default: ``"none"``).
        """
        return self._parms.get("transform")
```

seb-h2o:
@maurever I already implemented logic to solve those conflicts in `BaseSklearnEstimator`
see https://github.com/h2oai/h2o-3/pull/3744/files#diff-d00b46d85270db4a17a02baeb578d1e7R111 for a test using `H2OPCA`.
The idea is to keep the original property name (e.g. `transform`) for the constructor and the sklearn property access logic (`get_params`, `set_params`) but expose it as an object property with a trailing underscore (e.g. `estimator.transform_`): this way it's transparent for 90% of sklearn usages, and still intuitive for the last 10%.

I'll try it for the 4 estimators above.
For `H2OAggregatorEstimator` though, it's a bit different, this one should probably never have been an estimator: it's really just a transformer (without even any fitting step, it's a basic transformer), so I may wrap it into a pure transformer instead of estimator to avoid confusing users...

seb-h2o:
OK, I think I found out why my tests were failing on Jenkins: this is due to how we setup Py client in multinode. Any test closing the connection to the cloud created by `h2o_py_test_setup.py` is likely to screw up the following tests: those will create a new connection using `utilsPY.standaloneTest` starting a new instance on default 54321, and as many tests can run in parallel, this can quickly become nasty.
Therefore, removing my connection checks and autostart conditions in my tests here.

maurever:
> I'll try it for the 4 estimators above.
> For `H2OAggregatorEstimator` though, it's a bit different, this one should probably never have been an estimator: it's really just a transformer (without even any fitting step, it's a basic transformer), so I may wrap it into a pure transformer instead of estimator to avoid confusing users...

Cool, looks good! Thank you.

seb-h2o:
@maurever see h2o-py/tests/testdir_sklearn/pyunit_sklearn_estimators_as_transformers.py for a test checking support for `transform` method on the 4 estimators previously mentioned.

seb-h2o:
@michalkurka @maurever @ledell the new `h2o.sklearn` API looks complete now.
Please finish the review, have a look at it, play with it: in Jupyter it's easy, start to import the module and follow the autocompletion.
h2o_sklearn

h2o-py/tests/testdir_sklearn/pyunit_sklearn_classification_pipeline.py and h2o-py/tests/testdir_sklearn/pyunit_sklearn_classification_search_pipeline.py are also decent examples.

PR #3747, opened 2019-07-25
PUBDEV-6686: Debugging code to find out what is causing NPE in DRF training

michalkurka:
This doesn't fix anything - it should just show with column is `null`, having a vec=null would cause an exception like observed in PUBDEV-6686.

    06-24 18:18:17.745 127.0.0.1:54321       11855  FJ-1-93   ERRR: java.lang.NullPointerException
    06-24 18:18:17.745 127.0.0.1:54321       11855  FJ-1-93   ERRR:         at water.fvec.Frame.<init>   (Frame.java:168)
    06-24 18:18:17.745 127.0.0.1:54321       11855  FJ-1-93   ERRR:         at water.fvec.Frame.<init>(Frame.java:138)


michalkurka:
@mdymczyk that makes sense - however it is not easily possible to do that; the frame that is passed to this method is the training frame... it shouldn't have null vecs in the first place

because this is happening in the middle of the training - something must be going poorly during the training - we will catch this issue in the next iteration (hopefully) :)

PR #3748, opened 2019-07-26
PUBDEV-6716: Make sure TargetEncoderMojoModel and TargetEncoderModel return same values for unseen categorical levels

deil87:
This PR makes sure that both TargetEncoderMojoModel and TargetEncoderModel return posterior probability from NA level for unexpected values. If no missing values were encountered during training then prior probability from training data is being used.

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
@@ -318,23 +318,17 @@ public class TargetEncoder {
       int numberOfRowsInEncodingMap = (int) encodingMapForCurrentTEColumn.numRows();
       String lastDomain = encodingMapForCurrentTEColumn.domains()[0][numberOfRowsInEncodingMap - 1];
       boolean missingValuesWerePresent = lastDomain.equals(teColumnName + "_NA");
-      Frame encodingsForNALevel = null;
-      try {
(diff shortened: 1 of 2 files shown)


h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
```diff
@@ -313,6 +313,31 @@ Frame imputeWithMean(Frame fr, int columnIndex, double mean) {
...
       return fr;
     }
+    
+    Frame imputeWithPosteriorForNALevelOrWithPrior(String teColumnName, Frame fr, int columnIndex, Frame encodingMapForCurrentTEColumn, double priorMean) {
+      int numberOfRowsInEncodingMap = (int) encodingMapForCurrentTEColumn.numRows();
+      String lastDomain = encodingMapForCurrentTEColumn.domains()[0][numberOfRowsInEncodingMap - 1];
+      boolean missingValuesWerePresent = lastDomain.equals(teColumnName + "_NA");
+      Frame encodingsForNALevel = null;
+      try {
+        encodingsForNALevel = encodingMapForCurrentTEColumn.deepSlice(new long[]{numberOfRowsInEncodingMap - 1}, null);
```
michalkurka: What purpose does the `deepSlice` have - you are only working with a 2 value of that Frame - can you just use the original frame with the right row index on lines 325, 326?

deil87: @michalkurka good point! Fixed

PR #3749, opened 2019-07-26
PUBDEV-6709: add pr_auc() call to Python client

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6709?filter=-1.

Work done:
1. I have added the pr_auc() call to python model_base.py so that users can call it.
2. Added pyunit test to make sure it can be called.

### DIFF
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -764,7 +764,10 @@ class ModelBase(backwards_compatible(Keyed)):
         """
         tm = ModelBase._get_metrics(self, train, valid, xval)
         m = {}
-        for k, v in viewitems(tm): m[k] = None if v is None else v.auc()
+        for k, v in viewitems(tm):
+            if not(v == None) and not(is_type(v, h2o.model.metrics_base.H2OBinomialModelMetrics)):
+                raise H2OValueError("auc() is only available for Binomial classifiers.")
+            m[k] = None if v is None else v.auc()
         return list(m.values())[0] if len(m) == 1 else m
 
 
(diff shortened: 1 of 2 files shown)


h2o-py/h2o/model/model_base.py
```diff
@@ -764,7 +764,10 @@ def auc(self, train=False, valid=False, xval=False):
         """
         tm = ModelBase._get_metrics(self, train, valid, xval)
         m = {}
-        for k, v in viewitems(tm): m[k] = None if v is None else v.auc()
+        for k, v in viewitems(tm):
+            if not(v == None) and not(is_type(v, h2o.model.metrics_base.H2OBinomialModelMetrics)):
```
michalkurka: GLM is special in R, how about in Python - will it work as well?

h2o-py/tests/testdir_jira/pyunit_pubdev_6709_pr_auc_call.py
```diff
@@ -0,0 +1,71 @@
...
+    print(gbm_h2o)
+    print("pr_auc for GBM model is {0}".format(gbm_h2o.pr_auc()))
+
+    # Build H2O GLM classification model:
+    glm_h2o = H2OGeneralizedLinearEstimator(family='binomial', seed=seed)
+    glm_h2o.train(x=list(range(1,prostate_train.ncol)),y="CAPSULE", training_frame=prostate_train)
+    print("***************************   Printing GLM model")
+    print(glm_h2o)  # glm scoring history does not contain AUC, and hence no pr_auc
+    print("pr_auc for GLM model is {0}".format(glm_h2o.pr_auc()))
+
```
wendycwong: Michalk:  Here is the test for GLM.  You can use the same call pr_auc() or auc() to deal with GLM models.  GLM may be special, the metrics are the same as the others.

PR #3755, opened 2019-07-29
PUBDEV-6603 - H2OFrame.split_frame() leaks a _splitter object

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6603

## The problem

Method `split_frame` in `H2OFrame` python class attemps to delete the variable `tmp_runif` by issuing `        del tmp_runif` command on line `1885`.

Nothing is deleted, as `H2OFrame` does not override `__del__` properly.

## The solution

Implement `__del__` method for `H2OFrame` correctly.

## Discussion

1. Why doesn't python use `FrameSplitter` backend services ? In order not to break any contracts, this PR will focus solely on fixing the bug and won' attempt to fix what is possibly not broken (calling backend instead).

1. Other option would be to call `h2o.remove()` directly from `split_frame` method.

### DIFF
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -108,6 +108,9 @@ class H2OFrame(Keyed):
             self._upload_python_object(python_obj, destination_frame, header, separator,
                                        column_names, column_types, na_strings, skipped_columns)
 
+    def __del__(self):
+        h2o.remove(self)
+
     @staticmethod
     def _expr(expr, cache=None):
         # TODO: merge this method with `__init__`
@@ -1831,6 +1834,7 @@ class H2OFrame(Keyed):
                 # lower_boundary is 0.0
                 upper_boundary = boundaries[i]
                 tmp_slice = self[(tmp_runif <= upper_boundary), :]
+                print(tmp_slice.key)
             elif i == num_slices - 1:
                 lower_boundary = boundaries[i - 1]
                 # upper_boundary is 1.0
@@ -1848,7 +1852,6 @@ class H2OFrame(Keyed):
                 splits.append(tmp_slice)
 
             i += 1
-
         del tmp_runif
         return splits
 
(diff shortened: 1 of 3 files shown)


h2o-py/h2o/frame.py
```diff
@@ -108,6 +108,9 @@ def __init__(self, python_obj=None, destination_frame=None, header=0, separator=
             self._upload_python_object(python_obj, destination_frame, header, separator,
                                        column_names, column_types, na_strings, skipped_columns)
 
+    def __del__(self):
```
michalkurka: IMHO this is not a good idea. You might want to import a file in Flow, then go to Python, use a reference to the object and then continue in Flow. I don't think (all) Frames in H2O should copy the same lifecycle as the client objects.

In R:
- when you create a temporary frame and the frame is garbage collected, corresponding H2O object is delete on the backend
- non-temporary objects are not deleted in the H2O backend

I think you change affects all H2OFrames.

IMHO this PR should (ideally) only address the reported issue. If you want to make a global change - I think it would be better to make a new jira and describe what was changed and why.

maurever: In my opinion, we should use the FrameSplitter if it is possible (if there is API to use it from Python). However, I guess the reason why the Python code is used directly instead of Java is the speed, or not?

Pscheidl: Resolved, waiting for tests.

michalkurka: FrameSplitter is currently not used at all except for tests, it was created for the old Python client.

Split Frame is implemented the same way in R, Python & Flow (= 3x the same thing).

Having a backend implementation would make it IMHO make more efficient, I agree.

Pscheidl:
Massive test failures due to this change. Looks like some pieces of code actually rely on this "feature". Investigating.

Pscheidl:
Row slicing seems to be done lazily.

PR #3756, opened 2019-07-29
PUBDEV-6513 XGBoost MOJO import

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6513

Standard tests we already have for other algos.

### DIFF
--- a/h2o-algos/src/main/java/hex/generic/Generic.java
+++ b/h2o-algos/src/main/java/hex/generic/Generic.java
@@ -3,10 +3,20 @@ package hex.generic;
 import hex.ModelBuilder;
 import hex.ModelCategory;
 import hex.genmodel.*;
+import hex.genmodel.algos.drf.DrfMojoModel;
+import hex.genmodel.algos.gbm.GbmMojoModel;
+import hex.genmodel.algos.glm.GlmMojoModel;
+import hex.genmodel.algos.glm.GlmMultinomialMojoModel;
+import hex.genmodel.algos.glm.GlmOrdinalMojoModel;
+import hex.genmodel.algos.isofor.IsolationForestMojoModel;
+import hex.genmodel.algos.kmeans.KMeansMojoModel;
+import water.DKV;
 import water.H2O;
 import water.Key;
+import water.Scope;
 import water.fvec.ByteVec;
 import water.fvec.Frame;
+import water.util.ArrayUtils;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -72,10 +82,6 @@ public class Generic extends ModelBuilder<GenericModel, GenericModelParameters, 
             try {
                 final MojoReaderBackend readerBackend = MojoReaderBackendFactory.createReaderBackend(mojoBytes.openStream(_job._key), MojoReaderBackendFactory.CachingStrategy.MEMORY);
                 mojoModel = ModelMojoReader.readFrom(readerBackend, true);
-
-                if (!mojoModel.isH2OImportable()) {
-                    throw new IllegalArgumentException(String.format("Unsupported MOJO model %s. ", mojoModel.getClass().getName()));
-                }
 
                 final GenericModelOutput genericModelOutput = new GenericModelOutput(mojoModel._modelDescriptor, mojoModel._modelAttributes);
                 final GenericModel genericModel = new GenericModel(_result, _parms, genericModelOutput, mojoModel, mojoBytes);
--- a/h2o-algos/src/main/java/hex/generic/Generic.java
+++ b/h2o-algos/src/main/java/hex/generic/Generic.java
@@ -3,31 +3,29 @@ package hex.generic;
 import hex.ModelBuilder;
 import hex.ModelCategory;
 import hex.genmodel.*;
-import hex.genmodel.algos.drf.DrfMojoModel;
-import hex.genmodel.algos.gbm.GbmMojoModel;
-import hex.genmodel.algos.glm.GlmMojoModel;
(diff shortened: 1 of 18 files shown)


h2o-algos/src/main/java/hex/generic/Generic.java
```diff
@@ -29,8 +29,6 @@
  */
 public class Generic extends ModelBuilder<GenericModel, GenericModelParameters, GenericModelOutput> {
 
-    private static final Class[] SUPPORTED_MOJOS = new Class[]{GlmMojoModel.class, GlmMultinomialMojoModel.class,
```
michalkurka: Is there something that is replacing this check?

IMHO not all mojos are supported, or are they?

michalkurka:
Looks good - my only question is whether to keep the check on supported mojos - imho it is better to control what users can use instead of giving them all and not be sure what the user experience will be.

Pscheidl:
#response_container_BBPPID{font-family: initial; font-size:initial; color: initial;}  Yes, the check must be re-done and I forgot about that. Thanks.                                                                                                                                                                                                                                                                                                                        Od: notifications@github.comOdesláno: 29. července 2019 18:30Komu: h2o-3@noreply.github.comOdpověď komu: reply@reply.github.comKopie: pavel.junior@pscheidl.cz; assign@noreply.github.comPředmět: Re: [h2oai/h2o-3] PUBDEV-6513 XGBoost MOJO import (#3756)  Looks good - my only question is whether to keep the check on supported mojos - imho it is better to control what users can use instead of giving them all and not be sure what the user experience will be.

—You are receiving this because you were assigned.Reply to this email directly, view it on GitHub, or mute the thread.

h2o-py/tests/testdir_generic_model/pyunit_generic_model_mojo_xgboost.py
```diff
@@ -0,0 +1,76 @@
...
+def test(x, y, output_test, strip_part, algo_name, generic_algo_name):
+    airlines = h2o.import_file(path=pyunit_utils.locate("smalldata/testng/airlines_train.csv"))
+    xgb = H2OXGBoostEstimator(ntrees=1, nfolds=3)
+    xgb.train(x=x, y=y, training_frame=airlines, validation_frame=airlines)
+    print(xgb)
+    with Capturing() as original_output:
+        xgb.show()
+    
+    original_model_filename = tempfile.mkdtemp()
+    original_model_filename = xgb.download_mojo(original_model_filename)
```
maurever: Just cosmetics, but the old formula is confusing to me. :) But nothing important!

```suggestion
    original_model_filename = xgb.download_mojo(tempfile.mkdtemp())
```

h2o-algos/src/main/java/hex/generic/Generic.java
Pscheidl: Right now, all MOJOs are supported, but not tested. Basic metrics are available for all the MOJOs.

Pscheidl: Resolved, waiting for tests.

Pscheidl:
The check has been re-implemented in a different way, because XGBoost and possibly other extensions. Every GenModel extension has to declare whether is supports Generic model importability or not.

Pscheidl:
More changes to come.

h2o-algos/src/main/java/hex/generic/Generic.java
```diff
@@ -3,34 +3,30 @@
...
+import java.util.*;
 
 /**
  * Generic model able to do scoring with any underlying model deserializable into a format known by the {@link GenericModel}.
  * Only H2O Mojos are currently supported.
  */
 public class Generic extends ModelBuilder<GenericModel, GenericModelParameters, GenericModelOutput> {
+    
+    private static final Set<String> ALLOWED_MOJO_ALGOS;
+    static{
```
Pscheidl: We have tests that prove all these algorithms are supported and work. Both JUnit (`GenericModelTest`) and Python and R (pyunit_generic* or runit_generic*).

There weas a **false support** claimed for KMeans, but in fact it was not inteded, nor tested in any way. Probably a forgotten leftover. As this section had to be redesigned in this PR, took the opportunity to remove the support. WDYT ? cc @michalkurka

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
```diff
@@ -183,7 +183,10 @@ private void readAll(final boolean readModelDescriptor) throws IOException {
     checkMaxSupportedMojoVersion();
     readModelData();
     if (readModelDescriptor) {
-      _model._modelDescriptor = new ModelDescriptorBuilder(_model).build();
+      final String algoName = readkv("algo");
```
Pscheidl: This fixes a bug where algoName and full algo name are filled with one value only - the fully qualified class name of the model. We **do have these values in the MOJO**, and thus we should use those. Also, the reader builder relies on these values to construct the proper MOJO reader. As this PR is merged, those will be used to compare the algorithm against a list of algorithms allowed for MOJO import in `Generic` class.

```
    "algo": "gbm",
    "algo_full_name": "Gradient Boosting Machine",
```

Pscheidl:
During redesigning the section with `ALLOWED_ALGORITHMS`, I've found out that we claim false support for `KMeans`. It was not inteded and we're not testing it in any way. I apologize for the mistake. 

The actual import works and metrics are shown, but scoring is not working. Created a JIRA to support it `https://0xdata.atlassian.net/browse/PUBDEV-6735`

PR #3761, opened 2019-07-30
PUBDEV-6697 fix gbm quasibinomial

maurever:
The test sometimes failed. The reason is the quasibinomial create the first tree as a null tree.

I found it is caused by incorrect target distribution calculation at the beginning of algo.

JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6697

EDIT:
- I found we expected the input has to be {0, something}, but that is what GBM quasibinomial is about
- In all prints, there are mapping to {0, 1} without explanation how mapping was creating

TODO:
- allow correct input, for example {-1, 42}
- show correct domain in all prints?


h2o-core/src/main/java/water/util/MRUtils.java
```diff
@@ -116,14 +116,20 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
     @Override public void map(Chunk ys) {
       _ys = new double[_nclass];
       for( int i=0; i<ys._len; i++ )
-        if (!ys.isNA(i))
-          _ys[(int) ys.at8(i)]++;
+        if (!ys.isNA(i)) {
+          // PUBDEV-6697 quasibinomial GBM can have different domains than {0, 1}, for example {0, 2} or {0, 42}
+          int index = ys.at8(i) >= _ys.length ? 1 : (int) ys.at8(i);
```
maurever: If the domains were different than {0,1} for example {0,2}, it caused OutOfArrayIndexException here. It tried to reach `_ys[2]`, but the length of `_ys` is always 2 for quasibinomial.

h2o-algos/src/main/java/hex/tree/SharedTree.java
```diff
@@ -341,7 +341,7 @@ protected void validateRowSampleRate() {
         final int [] cons = new int[_nclass];
         for( int i=0; i<_nclass; i++ ) {
           names[i] = "NIDs_" + domain[i];
-          cons[i] = isSupervised() && ((_parms._distribution == DistributionFamily.quasibinomial || _model._output._distribution[i]==0)) ? -1 : 0;
```
maurever: This code causes the first tree in Quasibinomial GBM is null even though the target distribution is calculated correctly for {0, 1} domains. For other domains like {0, 42}, the target distribution calculation throws an exception. See other changes bellow.

h2o-core/src/main/java/water/util/MRUtils.java
```diff
@@ -128,6 +130,57 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
   }
 
+  /**
+   * Compute the class distribution for qusibinomial distribution from a class label vector
+   * (not counting missing values)
+   */
+  public static class ClassDistQuasibinomial extends MRTask<ClassDistQuasibinomial> {
+    final int _nclass;
+    protected double[] _ys;
+    protected Double[] _domain;
```
michalkurka: why Double?

h2o-core/src/main/java/water/util/MRUtils.java
```diff
@@ -128,6 +130,57 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+    
+    public ClassDistQuasibinomial(String[] domain) { 
+      _nclass = 2;
+      _domain = new Double[]{Double.valueOf(domain[0]), Double.valueOf(domain[1])};
```
michalkurka: seems like the code only needs the first element of the array

h2o-core/src/main/java/water/util/MRUtils.java
```diff
@@ -128,6 +130,57 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+      return _ys; 
+    }
+    
+    public final double[] relDist() {
+      final double sum = ArrayUtils.sum(_ys);
+      return ArrayUtils.div(Arrays.copyOf(_ys, _ys.length), sum);
+    }
+    
+    public final String[] domains(){
+      return new String[]{Double.toString(_domain[0]), Double.toString(_domain[1])};
```
michalkurka: it is passed in just to be retrieved?

h2o-core/src/main/java/water/util/MRUtils.java
```diff
@@ -128,6 +130,57 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+      _domain = new Double[]{Double.valueOf(domain[0]), Double.valueOf(domain[1])};
+      
+    }
+    
+    public final double[] dist() {
+      return _ys; 
+    }
+    
+    public final double[] relDist() {
+      final double sum = ArrayUtils.sum(_ys);
```
michalkurka: with 0 weights and NA responses we can have sum = 0, can't we?

maurever: I think the sum cannot be zero. At least one domain has to be set; there is a check for this before model start building.

h2o-core/src/main/java/water/util/MRUtils.java
maurever: Fixed.

h2o-core/src/main/java/water/util/MRUtils.java
maurever: Fixed.

h2o-core/src/main/java/water/util/MRUtils.java
maurever: True. Fixed.

h2o-core/src/main/java/water/util/MRUtils.java
michalkurka: even with 0 weights?

PR #3764, opened 2019-07-30
PUBDEV-6263 - Bypass proxy in Python when H2O is started on localhost

Pscheidl:
We don't want to use environment variables ,which is the official way to do it in `requests` package: https://github.com/psf/requests/pull/945

## Using freshly configured session

Session is created internally anyway when `requests.request(...)` is called, thus this introduces no overhead.

```
            session = requests.Session()
            session.trust_env = False
            resp = session.request(method=method, url=url, data=data, json=json, files=files, params=params,
                                    headers=headers, timeout=self._timeout, stream=stream,
                                    auth=self._auth, verify=self._verify_ssl_cert, proxies=self._proxies)
            session.close()
```

Setting the session not to trust the environment will make it ignore all proxies (not just localhost). This approach works, but as a side-effect, all certificates are ignored and possibly more.

## Using non-empty proxy map

Once the map of proxies is empty or `None`, the environment settings are respected instead by the `requests` library. Passing a made-up key-value inside the map can also do the trick. This has no side effects besides disabling all the proxies.

```
NO_PROXY = {
    'no': 'pass',
}

r = requests.get(uri, proxies=NO_PROXY)
```

https://github.com/psf/requests/issues/879


Pscheidl:
Also trying to find out a way how to test it.

michalkurka:
@Pscheidl regarding testing - it probably falls into the same category as the "py init" tests

Pscheidl:
For R, there is another JIRA created:
https://0xdata.atlassian.net/browse/PUBDEV-6931

h2o-py/h2o/backend/connection.py
```diff
@@ -399,7 +411,7 @@ def request(self, endpoint, data=None, json=None, filename=None, save_to=None):
 
             headers = {"User-Agent": "H2O Python client/" + sys.version.replace("\n", ""),
                        "X-Cluster": self._cluster_id,
-                       "Cookie": self._cookies}
+                       "Cookie": self._cookies}
```
honzasterba: extra whitespace

h2o-py/tests/testdir_algos/kmeans/pyunit_init_proxy.py
```diff
@@ -0,0 +1,27 @@
...
+    else:
+        assert len(conn._proxies) == 0
+        assert conn._proxies is None
+
+
+
+if __name__ == "__main__":
+    pyunit_utils.standalone_test(init_proxy)
+else:
+    init_proxy()
```
honzasterba: should this be in algos folder? it has nothing to do with kmeans, I would put it elsewhere

h2o-py/h2o/backend/connection.py
Pscheidl: done

h2o-py/tests/testdir_algos/kmeans/pyunit_init_proxy.py
Pscheidl: definitely not. moved.

h2o-py/h2o/backend/connection.py
```diff
@@ -315,6 +316,18 @@ def open(server=None, url=None, ip=None, port=None, name=None, https=None, auth=
...
                          "This may interfere with your H2O Connection." % name)
+                    
+            local_if_host = socket.gethostbyname(socket.gethostname()) # Local interface
+            if ("localhost" in conn._base_url
+                or "127.0.0.1" in conn._base_url
+                or local_if_host in conn._base_url):
+                # Empty list will cause requests library to respect the default behavior.
+                # Thus a non-existing proxy is inserted.
+
+                conn._proxies = {
```
michalkurka: What if I have a proxy defined on the OS level?

h2o-py/h2o/backend/connection.py
```diff
@@ -315,6 +316,18 @@ def open(server=None, url=None, ip=None, port=None, name=None, https=None, auth=
                 if name.lower() == scheme + "_proxy":
                     warn("Proxy is defined in the environment: %s. "
                          "This may interfere with your H2O Connection." % name)
+                    
+            local_if_host = socket.gethostbyname(socket.gethostname()) # Local interface
+            if ("localhost" in conn._base_url
```
michalkurka: why not to use the `ip` property directly? this will cause issues for urls like eg. http://remote-server/localhostinsky

PR #3767, opened 2019-08-01
PUBDEV-6741: Create examples for the Autoencoder section in the Python documentation

hannah-tillman:


h2o-py/h2o/model/autoencoder.py
```diff
@@ -17,6 +17,26 @@ def anomaly(self, test_data, per_feature=False):
...
+        >>> test = test[predictors]
+        >>> ae_model = H2OAutoEncoderEstimator(activation="Tanh",
+        ...                                    hidden=[2],
+        ...                                    l1=1e-5,
+        ...                                    ignore_const_cols=False,
+        ...                                    epochs=1)
+        >>> ae_model.train(x=predictors,training_frame=train)
+        >>> test_rec_error = ae_model.anomaly(test)
+        >>> test_rec_error
+        >>> test_recon = ae_model.predict(test)
```
angela0xdata: Hmm... instead of a predict example, can you add a per_features=True example?

test_rec_error_features = ae_model.anomaly(test, per_feature=True)
test_rec_error_features

PR #3769, opened 2019-08-01
PUBDEV-6742: Create examples for the Tree Class section in the Python documentation

hannah-tillman:
The example for `show()` shows up in the source code but not in the docs when I build locally. When I tried to add the framing for the function (the "@property" at the beginning and the "return self._show" at the end), none of the html page loaded when I locally built it. The function works; it just isn't building properly for some reason. Any ideas?


h2o-py/h2o/tree/tree.py
```diff
@@ -29,37 +29,113 @@ def __init__(self, model, tree_number, tree_class=None):
```
angela0xdata: Missing this change at the top:

Also, the description needs to change from:

"Represents a model of a Tree built by one of H2O’s algorithms (GBM, Random Forest)."
to
"Represents a model of a Tree built by one of H2O’s algorithms (GBM, Random Forest, XGBoost, Isolation Forest)."

h2o-py/h2o/tree/tree.py
```diff
@@ -29,37 +29,113 @@ def __init__(self, model, tree_number, tree_class=None):
 
     @property
```
angela0xdata: Also at the top, can you add parameters under the H2OTree section (before "left_children")? 
Currently, the function is specified as "class h2o.tree.H2OTree(model, tree_number, tree_class=None)" but model, tree_number, and tree_class are not described.

From the R documentation:
model - The model this tree is related to.
tree_number - An integer representing the order in which the tree has been built in the model.
tree_class - A string representing the name of the tree's class. The number of tree classes equals the number of levels in categorical response column. As there is exactly one class per categorical level, the name of the tree's class is equal to the corresponding categorical level of the response column. In case of regression and binomial models, the name of the categorical level is ignored and can be omitted, as there is exactly one tree built in both cases.

angela0xdata:
Let me do some digging for show()

PR #3777, opened 2019-08-05
PUBDEV-6756: Fix NPE when creating XGBoost sparse matrices

honzasterba:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/SparseMatrixDimensions.java
```diff
@@ -33,7 +33,7 @@ public SparseMatrixDimensions(int[] nonZeroElementsCounts, int[] rowIndicesCount
             precedingNonZeroCount += nonZeroElementsCounts[i];
         }
 
-        _nonZeroElementsCount = ArrayUtils.sum(nonZeroElementsCounts);
+        _nonZeroElementsCount = ArrayUtils.suml(nonZeroElementsCounts);
```
Pscheidl: I wonder if we need overflow detection Because `(2^31)^2  < 2^61` (our 2D array has potentially the size of `(2^31)^2`).

Pscheidl: Even then, I wonder what the reaction would be.

michalkurka: IMHO we don't have to worry about overflow here - here we would be approaching millions of terabytes of memory.

Is there a practical scenario where we could hit the limit?

PR #3782, opened 2019-08-05
PUBDEV-6758: added glrm to hadoop smoke test

wendycwong:
This PR completes task in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6758?filter=-1

After talking to Honza, we have decided to add a GLRM test to hadoop smoke test.


honzasterba:
does the INTERNAL prefix do anything? will the test run?

wendycwong:
Honza:  It is format that we use for old hadoop test.  I will remove it.

wendycwong:
I noticed that hadoop smoke test is not part of the jenkins test for PR.  How do I make sure it actually works?

honzasterba:
you can add your branch to the jenkins job

wendycwong:
Honza:

I added my branch to hadoop smoke test.  Once the test passed, I will remove it from it.

PR #3783, opened 2019-08-05
Switch to Java 8 as a target for all modules except genmodel

michalkurka:
Functionality for MOJO-scoring stays Java 7 compatibile

### DIFF
--- a/scripts/jenkins/groovy/defineTestStages.groovy
+++ b/scripts/jenkins/groovy/defineTestStages.groovy
@@ -38,11 +38,6 @@ def call(final pipelineContext) {
       component: pipelineContext.getBuildConfig().COMPONENT_JS
     ],
     [
-      stageName: 'Java 7 Smoke', target: 'test-junit-7-smoke-jenkins', javaVersion: 7, timeoutValue: 20,
-      component: pipelineContext.getBuildConfig().COMPONENT_JAVA,
-      image: "${pipelineContext.getBuildConfig().DOCKER_REGISTRY}/opsh2oai/h2o-3/dev-openjdk-7:${pipelineContext.getBuildConfig().DEFAULT_IMAGE_VERSION_TAG}"
-    ],
-    [
       stageName: 'Java 8 Smoke', target: 'test-junit-smoke-jenkins', javaVersion: 8, timeoutValue: 20,
       component: pipelineContext.getBuildConfig().COMPONENT_JAVA
     ],
@@ -96,11 +91,6 @@ def call(final pipelineContext) {
     [
       stageName: 'Py3.5 Small AutoML', target: 'test-pyunit-small-automl', pythonVersion: '3.5',
       timeoutValue: 90, component: pipelineContext.getBuildConfig().COMPONENT_PY
-    ],
-    [
-      stageName: 'R3.4 Init Java 7', target: 'test-r-init', rVersion: '3.4.1', javaVersion: 7,
-      timeoutValue: 10, hasJUnit: false, component: pipelineContext.getBuildConfig().COMPONENT_R,
-      image: "${pipelineContext.getBuildConfig().DOCKER_REGISTRY}/opsh2oai/h2o-3/dev-r-3.4.1-openjdk-7:${pipelineContext.getBuildConfig().DEFAULT_IMAGE_VERSION_TAG}"
     ],
     [
       stageName: 'R3.4 Init Java 8', target: 'test-r-init', rVersion: '3.4.1', javaVersion: 8,
--- a/settings.gradle
+++ b/settings.gradle
@@ -47,7 +47,7 @@ if ("true".equals(System.getenv("H2O_BUILD_GRPC"))) {
 apply plugin: 'scala-cross-build'
 
 scalaCrossCompile {
-  targetVersions "2.10.4", "2.11.7"
+  targetVersions "2.10.7", "2.11.12"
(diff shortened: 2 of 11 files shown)


michalkurka:
@jakubhava please check this is okay to do: https://github.com/h2oai/h2o-3/pull/3783/commits/20b28849e19b4375adc75fc6698838ee7e9961ce

I had to upgrade in order to make the build work on Java 8, 10 and 11

scripts/jenkins/groovy/defineTestStages.groovy
```diff
@@ -37,11 +37,6 @@ def call(final pipelineContext) {
       stageName: 'Flow Headless Smoke', target: 'test-flow-headless-smoke',timeoutValue: 20,
       component: pipelineContext.getBuildConfig().COMPONENT_JS
     ],
-    [
-      stageName: 'Java 7 Smoke', target: 'test-junit-7-smoke-jenkins', javaVersion: 7, timeoutValue: 20,
-      component: pipelineContext.getBuildConfig().COMPONENT_JAVA,
-      image: "${pipelineContext.getBuildConfig().DOCKER_REGISTRY}/opsh2oai/h2o-3/dev-openjdk-7:${pipelineContext.getBuildConfig().DEFAULT_IMAGE_VERSION_TAG}"
-    ],
```
seb-h2o: we can trust the compiler restrictions only for the language features but not for the changes in JDK. So, do you plan to have a Java 7 stage to test the MOJOs?

settings.gradle
```diff
@@ -47,7 +47,7 @@ if ("true".equals(System.getenv("H2O_BUILD_GRPC"))) {
 apply plugin: 'scala-cross-build'
 
 scalaCrossCompile {
-  targetVersions "2.10.4", "2.11.7"
+  targetVersions "2.10.7", "2.11.12"
```
seb-h2o: how is it related to Java 8 support? https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html
Just surprised that a simple upgrade in scala patch version is necessary to be compatible with Java 8.

Is it to be actually compatible with Java 11?

jakubhava:
@michalkurka @mmalohlava @mn-mikke This change also means that we can't support Spark 2.1 anymore with new H2O versions as Spark 2.1 is the only one requiring Java 7.

Is it ok to switch to maintenance mode of Spark 2.1 & don't put there any new features (and H2O updates) ?

settings.gradle
michalkurka: It is not related to Java 8 but it is related to the other changes in this PR - our Jenkins builds uses Java 8 and everything works fine but trying to compile on Java 11/12 without this change produces an error without the patch:

Java HotSpot(TM) 64-Bit Server VM warning: Ignoring option MaxPermSize; support was removed in 8.0
error: error while loading package, Missing dependency 'object java.lang.Object in compiler mirror', required by /Users/mkurka/.gradle/caches/modules-2/files-2.1/org.scala-lang/scala-library/2.10.4/9aae4cb1802537d604e03688cab744ff47b31a7d/scala-library-2.10.4.jar(scala/package.class)
error: error while loading package, Missing dependency 'object java.lang.Object in compiler mirror', required by /Users/mkurka/.gradle/caches/modules-2/files-2.1/org.scala-lang/scala-library/2.10.4/9aae4cb1802537d604e03688cab744ff47b31a7d/scala-library-2.10.4.jar(scala/runtime/package.class)
error: scala.reflect.internal.MissingRequirementError: object java.lang.Object in compiler mirror not found.
        at scala.reflect.internal.MissingRequirementError$.signal(MissingRequirementError.scala:16)
        at scala.reflect.internal.MissingRequirementError$.notFound(MissingRequirementError.scala:17)
        at scala.reflect.internal.Mirrors$RootsBase.getModuleOrClass(Mirrors.scala:48)
        at scala.reflect.internal.Mirrors$RootsBase.getModuleOrClass(Mirrors.scala:40)

scripts/jenkins/groovy/defineTestStages.groovy
michalkurka: We should do that, that is a good idea. I think MOJO Compatibility suite should definitely run with J7 as well.

settings.gradle
michalkurka: (also added this info to the commit message)

mmalohlava:
@jakubhava good question, we need to figure out active users of Spark 2.1 (i do not expect any, but better to verify).

michalkurka:
Spark 2.1 requires J7+ 

@jakubhava @mmalohlava I think we can still support Spark 2.1, just not on Java 7, or not?

I can imagine there is someone running Spark 2.1 but I've not seen Java 7 in production for a long time.

jakubhava:
Good point @michalkurka, I just verified and you can run Spark on Java 8 so we're good.

Anyway - I would like to stop putting new features into Spark 2.1 at some point anyway, but that is for different discussion which does not relate to this PR

jakubhava:
I will switch to Java 8 on Sparkling Water for Spark 2.1 in next SW=H2O major release

scripts/jenkins/groovy/defineTestStages.groovy
michalkurka: @seb-h2o Added a MOJO Compatibility test on Java 7.

Let's see if this PR passes :)

PR #3784, opened 2019-08-06
PUBDEV-6765 - MOJO model JUnit tests compare incorrect bytes

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6765

JUnit tests in GenericModelTest compare MOJOs byte wise. There are two problems:

     Generic MOJO bytes are compared to the same bytes, original MOJO bytes are not taken into account.

    Original MOJO bytes are written via ZipOutputStream, Generic bytes are stored internally via ByteVec and might be written into ByteArrayOutputStream in different order. Writing them as a file and comparing the files byte-wise should help.

**The goal** is to compare apples with apples. Both models are written into a file, then the file is read via a neutral function from Java SDK and then the bytes are compared.

### DIFF
--- a/h2o-algos/src/test/java/hex/generic/GenericModelTest.java
+++ b/h2o-algos/src/test/java/hex/generic/GenericModelTest.java
@@ -14,6 +14,7 @@ import hex.tree.gbm.GBM;
 import hex.tree.gbm.GBMModel;
 import hex.tree.isofor.IsolationForest;
 import hex.tree.isofor.IsolationForestModel;
+import org.apache.commons.io.FileUtils;
 import org.junit.Before;
 import org.junit.Test;
 import water.*;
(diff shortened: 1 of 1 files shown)


h2o-algos/src/test/java/hex/generic/GenericModelTest.java
```diff
@@ -528,26 +529,25 @@ public void downloadable_mojo_gbm() throws IOException {
             parms._distribution = AUTO;
             parms._response_column = trainingFrame._names[1];
             parms._ntrees = 1;
+            parms._seed = 0XFEED;
```
deil87: @Pscheidl do we need this in all downloadable tests? or we can remove it here?

Pscheidl: We can remove it. It is actually better to produce a variety of models. I'll do it.

PR #3785, opened 2019-08-06
PUBDEV-6567: add support for xgboost predict_leaf_assignment

honzasterba:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -496,6 +496,16 @@ public Frame scoreContributions(Frame frame, Key<Frame> destination_key, boolean
     return approx ? new PredictContribApproxTask(_parms, model_info, _output, di) : new PredictTreeSHAPTask(di, model_info(), _output);
   }
 
+  @Override
+  public Frame scoreLeafNodeAssignment(
+      Frame frame, LeafNodeAssignmentType type, Key<Frame> destination_key
+  ) {
+    AssignLeafNodeTask task = AssignLeafNodeTask.make(model_info.dataInfo(), _output, model_info._boosterBytes, type);
```
michalkurka: should we use scoring info instead?

```
  protected BigScorePredict setupBigScorePredict(BigScore bs) {
    DataInfo di = model_info().scoringInfo(false); // always for validation scoring info for scoring (we are not in the training phase)
```

honzasterba: fixed

PR #3786, opened 2019-08-06
PUBDEV-6739: Expose original parameters of imported MOJO model

deil87:
This PR adds support for model parameters that were stored in a JSON representation but were not parsed and exposed as ModelAttributes in mojo model.

Example of the printout of the mojo's parameters:
image

By default R approach was taken - model parameters will be printed in a truncated fashion. For Py there is a dedicated method to retrieve model parameters as TwoDimTable independently... in R there no such a method.

Questions for discussion:

1. whether we want to print all the parameters by default or truncated list is fine
2. in R I found only methods that just print model's details/summary/varimp/whatever and there is no dedicated helper methods for extracting them - is it fine?


deil87:
@Pscheidl feel free to review this one. Some tests are failing but they don't seem to be relevant. I will restart tests manually to see how consistent failures are.

Pscheidl:
Reply to the discussion in the description of the PR: In R, there is a `model@parameters` slot. Such slot should provide list of all the parameters.

h2o-py/h2o/model/model_base.py
```diff
@@ -447,6 +459,8 @@ def show(self):
             model["scoring_history"].show()
         if "variable_importances" in model and model["variable_importances"]:
             model["variable_importances"].show()
+        if "model_parameters" in model and model["model_parameters"]:
```
Pscheidl: We only want to make them available, not print them explicitely, as the amout of parameters could be quite large. WDYT ?

h2o-algos/src/main/java/hex/generic/GenericModelOutput.java
```diff
@@ -234,6 +241,58 @@ private static Object convertObjects(final Object source, final Object target) {
         return target;
     }
 
+    private static TwoDimTable convertModelParameters(final ModelParameter[] modelParameters) {
```
Pscheidl: I believe the MOJO part is done in a good way. However, couldn't we just reuse current API by filling _parms ? I know, the _parms is strongly typed and `GenericParameters` can't cover all the cases.

I'm thinking whether it would be possible to create static alternatives to all the param classes ? This way, a param once coded in could't be removed. Or provide two versions of `_parms` - a dynamic one like `_dynamic_parms` backed by a `TwoDimTable` (basically what you are doing here), but acting as a map. On frontend, just convert it. Or can we adjust the conversion on backend in order for the frontend client not to recognize anything ?

h2o-py/h2o/model/model_base.py
```diff
@@ -410,6 +410,18 @@ def summary(self):
         print("No model summary for this model")
 
 
+    def model_parameters(self):
```
Pscheidl: I guess we want to leverage curent `_parms` coming from backend and use usual ways of displaying them. Is this necessary ?

h2o-py/h2o/model/model_base.py
deil87: @Pscheidl By default TwoDimTable will be printed in a truncated fashion - 5 top rows and 5 last rows. Don't think it is too much. It could serve as a hint that parameters are exposed and user can go a get whole list.

h2o-py/h2o/model/model_base.py
deil87: @Pscheidl thanks, I will check if parameters could be exposed through _parms

Pscheidl: It would make your work futureproof.

I had the idea with providing multiple subclasses of GenericModelParameters and their V3 counterparts. This is costly and ineffective.

Just thinking out loud, but having a Map of these values and just creating the V3 objects would be easier (Schema.fillObject would have to be modified for that case).

Or, have one generic V3 serialized. In this case, the code generator would have to be modified as well, but it would be a truly general way and could even be used for my generic metrics system, which is now purely static.

deil87: @Pscheidl I think that knowing how to convert types of fields within those classes of parameters is one of the most general ways to deserialise. We don't need to care about classes. Right now recursion and reflection are being used and it is not static - ie if the field of supported type is added to a parameters class it will be deserialised just fine. 
I don't get exactly what you described above but I think we can merge current version ( which is done in the same fashion as existing functionality). I'm aware of approaches where json is being deserialised as Map and of course we can create code generators for each class of parameters but don't think it is really necessary. 
I think that there is no `futureproof` solutions as nobody knows future.
Anyway if you want to improve it please go ahead and create proposal PR after this one gets merged. But this proposal might not be something that contributes to extending functionality of the project.

deil87:
@Pscheidl could you please consider to approve PR ?

Pscheidl:
@deil87 Sure, I'll do the review. I didn't know it was finished yet :)

h2o-r/tests/testdir_algos/generic/runit_generic_model_mojo_model_parameters.R
```diff
@@ -0,0 +1,27 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../../scripts/h2o-r-test-setup.R")
+
+
+
+test.model.generic.drf <- function() {
```
Pscheidl: Copy & paste error in the naming of the test.

h2o-r/tests/testdir_algos/generic/runit_generic_model_mojo_model_parameters.R
```diff
@@ -0,0 +1,27 @@
...
+    print(generic_model)
+
+    print("Model parameters:")
+    if(!is.null(generic_model@parameters)){
+        print(generic_model@parameters)
+        expect_equal(length(generic_model@parameters), 50)
+    }
+}
+
+doTest("Generic model from DRF MOJO has its parameters exposed", test.model.generic.drf )
```
Pscheidl: Copy & paste error in the description and test name.

h2o-r/tests/testdir_algos/generic/runit_generic_model_mojo_model_parameters.R
```diff
@@ -0,0 +1,27 @@
...
+source("../../../scripts/h2o-r-test-setup.R")
+
+
+
+test.model.generic.drf <- function() {
+    data <- h2o.importFile(path = locate('smalldata/testng/airlines_train.csv'))
+    
+    # Binomial
+    cols <- c("Origin", "Distance")
+    original_model <- h2o.randomForest(x=cols, y = "IsDepDelayed", training_frame = data, validation_frame = data, nfolds = 3)
```
Pscheidl: Only random forest is tested. Ideally, all supported Generic MOJO models should be tested. Including this test into already existing pyunit and R tests would:

1. Reduce the load on test pipeline (another model trained, while we could check the presence of `_parms` in models already trained.
2. Test all supported models at once.

h2o-py/h2o/model/model_base.py
Pscheidl: Good enough to me.

h2o-py/tests/testdir_generic_model/pyunit_generic_model_mojo_gbm.py
```diff
@@ -46,6 +51,8 @@ def test(x, y, output_test, strip_part, algo_name, generic_algo_name):
     assert len(generic_mojo_model_from_file._model_json["output"]["variable_importances"]._cell_values) > 0
     assert generic_mojo_model_from_file._model_json["output"]["model_summary"] is not None
     assert len(generic_mojo_model_from_file._model_json["output"]["model_summary"]._cell_values) > 0
+    assert generic_mojo_model_from_file._parms is not None
+    assert len(generic_mojo_model_from_file._parms._cell_values) > 0
```
Pscheidl: When I try to inspect the params inside `generic_mojo_model_from_file.parms`

I get

```
<class 'dict'>: {'model_id': {'__meta': {'schema_version': 3, 'schema_name': 'ModelParameterSchemaV3', 'schema_type': 'Iced'}, 'name': 'model_id', 'label': 'model_id', 'help': 'Destination id for this model; auto-generated if not specified.', 'required': False, 'type': 'Key<Model>', 'default_value': None, 'actual_value': {'__meta': {'schema_version': 3, 'schema_name': 'ModelKeyV3', 'schema_type': 'Key<Model>'}, 'name': 'Generic_model_python_1566302911212_23', 'type': 'Key<Model>', 'URL': '/3/Models/Generic_model_python_1566302911212_23'}, 'level': 'critical', 'values': [], 'is_member_of_frames': [], 'is_mutually_exclusive_with': [], 'gridable': False}, 'model_key': {'__meta': {'schema_version': 3, 'schema_name': 'ModelParameterSchemaV3', 'schema_type': 'Iced'}, 'name': 'model_key', 'label': 'model_key', 'help': 'Key to the self-contained model archive already uploaded to H2O.', 'required': False, 'type': 'Key<Frame>', 'default_value': None, 'actual_value': {'__meta': {'schema_version': 3, 'schema_name': 'FrameKeyV3', 'schema_type': 'Key<Frame>'}, 'name': 'nfs://tmp/tmpgy08jywf/GBM_model_python_1566302911212_1.zip', 'type': 'Key<Frame>', 'URL': '/3/Frames/nfs://tmp/tmpgy08jywf/GBM_model_python_1566302911212_1.zip'}, 'level': 'critical', 'values': [], 'is_member_of_frames': [], 'is_mutually_exclusive_with': [], 'gridable': False}, 'path': {'__meta': {'schema_version': 3, 'schema_name': 'ModelParameterSchemaV3', 'schema_type': 'Iced'}, 'name': 'path', 'label': 'path', 'help': 'Path to file with self-contained model archive.', 'required': False, 'type': 'string', 'default_value': None, 'actual_value': '/tmp/tmpgy08jywf/GBM_model_python_1566302911212_1.zip', 'level': 'critical', 'values': [], 'is_member_of_frames': [], 'is_mutually_exclusive_with': [], 'gridable': False}}
```

**Question**: Is that all that is there ?

I've downloaded the MOJO created in this test 
[tmpmojo.zip](https://github.com/h2oai/h2o-3/files/3520539/tmpmojo.zip) - 
 and I can see much more in the parameters section of modelDetails than is present in the model._parms.

h2o-r/tests/testdir_algos/generic/runit_generic_model_mojo_drf.R
```diff
@@ -19,7 +19,10 @@ test.model.generic.drf <- function() {
     print(generic_model)
     
     original_output <- capture.output(print(original_model))
+    
     generic_output <- capture.output(print(generic_model))
+    generic_output <- drop_model_parameters_from_printout(generic_output)
```
Pscheidl: Why are we dropping the parameters from printout ? I don't say the output of parameters must be 1:1 exactly the same as I did it with output, but I don't see where the presence of model parameters is tested in these R tests.

Currently, there is only one test for random forest. I'd prefer having tests for each supported algorithm - these tests are already created, we only need to adjust them. Or at worst copy&paste them, if the output capturing does not suit your needs. Is there something preventing us from doing it ?

h2o-r/h2o-package/R/generic.R
```diff
@@ -40,5 +40,17 @@ h2o.generic <- function(model_id = NULL,
 
   # Error check and build model
   model <- .h2o.modelJob('generic', parms, h2oRestApiVersion=3, verbose=FALSE)
+    
+  # Exposing model's parameters represented as TwoDimTable as list
```
Pscheidl: This part should be placed into `gen_R.py`, otherwise, it'll be rewritten every time someone builds the project and triggers `gen_R.py`.

h2o-r/h2o-package/R/classes.R
```diff
@@ -179,6 +179,12 @@ setMethod("show", "H2OModel", function(object) {
   if( !is.null(model.parts$xm) ) print(model.parts$xm)
   cat("\n")
   if( !is.null(model.parts$xms) ) print(model.parts$xms)
+    
+  # model's parameters
+  has_model_params <- !is.null(m$model_parameters)
```
Pscheidl: Purpose of this code is to print parameters by default ? I don't think this we want to change this behavior for our users, we only want to expose the parameters.

h2o-py/h2o/estimators/generic.py
```diff
@@ -30,13 +30,20 @@ def __init__(self, **kwargs):
...
             if pname == 'model_id':
                 self._id = pvalue
-                self._parms["model_id"] = pvalue
             elif pname in names_list:
                 # Using setattr(...) will invoke type-checking of the arguments
                 setattr(self, pname, pvalue)
             else:
                 raise H2OValueError("Unknown parameter %s = %r" % (pname, pvalue))
 
+    def train(self, x=None, y=None, training_frame=None, offset_column=None, fold_column=None, weights_column=None,
```
Pscheidl: These changes should go into `gen_python.py`, otherwise they'll be overwritten once the project is built and/or `gen_python.py` is invoked.

h2o-algos/src/main/java/hex/generic/GenericModelOutput.java
```diff
@@ -17,6 +19,7 @@
     final ModelCategory _modelCategory;
     final int _nfeatures;
     TwoDimTable _variable_importances;
+    TwoDimTable _model_parameters;
```
Pscheidl: Having a universal structure is fine to me, but let's discuss about the reasoning. Having one `GenericModelParameters` per algorithm introduces a lot of new objects, but the clients should be able to work with such parameters object without any adjustments. Have you tried that ? Downside of that is a lot of new objects introduced (which I did with Metrics, which are slightly more complex objects and it might change in future).

This obviously leads to params being deserialized as list instead of a dict in Python. Which changes the experience for our users. Not sure about R.

Pscheidl:
Closing. Not fit for merging.

- Does not solve the original issue for reasons stated in comments above. Most importantly, this PR introduces new API for users to use, while the goal is for the user not to differentiate between generic model and any other "native" model.
- MOJO "backend" part is programatically done right, unfortunately is part of the "bad solution".

PR #3788, opened 2019-08-06
PUBDEV-6740: Add support for custom documentation (e.g. usage examples) for algos properties: Py + R APIs

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6740

To be able to provide a relatively simple way to add documentation and examples to generated Py and R code, this PR started to externalize parts of algo's code generation to a dedicated file.

Putting this logic further, and to make this code generation simpler when modifying existing algos, almost all custom code and doc has been externalized.
This has the following benefits:
- separate code from documentation/examples: safer and easier for our writers.
- improving standardization of generated code, by defining template: (1) params validation, (2) set params, (3) train model, (4) post process model.
- for reviews: if making changes for algo ALGO, then only files `gen_ALGO.py` should be edited without any risk of changes that could impact other algos: there is still `gen_python.py`, `gen_R.py` and/or `gen_defaults.py` for changes that apply to multiple algos.
- thanks to template mechanism, we can remove `if/else` hell with its mix of conditions on algos, property names, types and so on (esp. in R: it was a real pain to understand what's going on).
- fixed indentations + code layout issues (esp in R): strings in template should simply be left aligned (no indent calculation needed).
- flexibility if we need to add more custom hooks in the future.

and drawbacks:
- although I tried to minimize them, some small blocks of code are sometimes duplicated between algos, esp. in R for the validation of `x`, `y`, `training_frame` args. But it also has the benefit tof making them easier to modify...
- dev need to understand the template structure: will explain below.

Structure of code generation.
For Python and R, the code generation doesn't reside only in `gen_python.py` and `gen_R.py` respectively anymore.
Instead, we have the following structure:
```
h2o-bindings/bin
├── custom
│   ├── R
│   │   ├── gen_aggregator.py
│   |   ├── ...
│   │   ├── gen_defaults.py
│   |   ├── ...
│   │   └── gen_xgboost.py
│   └── python
│       ├── gen_aggregator.py
│       ├── ...
│       ├── gen_defaults.py
│       ├── ...
│       └── gen_xgboost.py
├── gen_R.py
├── gen_python.py
```
the `gen_defaults.py` has the same structure as other `gen_ALGO.py` files but is almost empty for Python and provides code blocks that are common to most algos for R (to avoid too much duplicates). `defaults` is a fallback, it is used iff its properties haven't been overridden in the dedicated algo file.

for Python the `gen_ALGO.py` files have the following possible entries (none of them is required):
```
rest_api_version = 99   # if algo doesn't use the default 3

def update_params(name, param):
    # a hook to modify schema params on the fly: mainly used to remove some enum values...
    pass

extensions = dict(
    __imports__="a string with additional imports",
    __class__="a string or a function defining additional methods for the default algo class",
    __module__="a string or a function defining additional methods for the algo module",
    __init__validation="a string defining additional code for validation phase of __init__: used only for generic",
    __init__setparams="a string defining additional code for set_params phase of __init__: used only for word2vec",
)

overrides = dict(
    foo = dict(
         getter="string defining code overriding the getter for property foo",
         setter="string defining code overriding the setter for property foo",
    ),
)

doc = dict(
    __class__ = "additional documentation added at class level",
    foo = "additional documentation added to the getter of property foo"
)

examples = dict(
    __class__ = "doc examples added at class level",
    foo = "doc examples added to the getter of property foo"
)
```

for R the `gen_ALGO.py` files have the following possible entries (none of them is required):
```
rest_api_version = 99   # if algo doesn't use the default 3

def update_params(name, param):
    # a hook to modify schema params on the fly: mainly used to remove some enum values...
    pass

extensions = dict(
    required_params=['x', 'y', 'training_frame'], #required params on top of signature: list of names or tuple(name, default_value)
    frame_params=['training_frame', 'validation_frame'], #by default (have special validation handling)
    extra_params=[('verbose', 'FALSE')],  #additional params not in schema: list of tuple(name, default_value)
    validate_required_params="specific code to validated required params",
    validate_params="specific code to validate params",
    set_required_params="specific code to set required params to `parms` object",
    skip_default_set_params_for=['foo'], # list of params for which we need to remove default set param mechanism because it will be customized below
    set_params="specific code to set other params to `parms` object",
    with_model="code to postprocess the model",
    module="additional R functions added to the module",
)

doc = dict(
    preamble="doc preamble",
    params=dict(
        foo="doc for param foo (without tag): used for params not in schema, otherwise overrides help doc from schema",
        bar="doc for param bar",
    ),
    returns="@return doc (without tag)",
    seealso="@seealso doc (without tag)",
    references="@references doc (without tag)",
    examples="@examples doc (code only)",
)
```

Future improvements (not in this PR):
- modify signature of `__init__` for Python algos to provide autocompletion and improve feature discovery.
- add documentation for params to Python `__init__`.


h2o-bindings/bin/custom/python/gen_gbm.py
```diff
@@ -0,0 +1,63 @@
+class_doc = """
+Builds gradient boosted trees on a parsed data set, for regression or classification.
+The default distribution function will guess the model type based on the response column type.
+Otherwise, the response column must be an enum for "bernoulli" or "multinomial", and numeric
+for all other distributions.
+"""
+
```
seb-h2o: @angela0xdata @hannah-tillman this is how the template for examples will look:
this simple `python/gen_gbm.py` file can define some basic string variables into which you can just add your doc and examples.

For object properties examples, what would you prefer?
1. a flat structure like 
``` 
property_training_frame_examples = ... 
property_validation_frame_examples = ...
```
2. a dictionary:
```
properties_examples = dict(
    training_frame="""...""",
    validation_frame="""...""",
)
```
see below to see how it looks like for real.
The second one may be slightly more readable, but don't forget your commas :)

h2o-py/h2o/estimators/gbm.py
```diff
@@ -58,6 +58,20 @@ def training_frame(self):
         Id of the training data frame.
 
         Type: ``H2OFrame``.
+
```
seb-h2o: @angela0xdata @hannah-tillman this is how it looks like once the code is generated

seb-h2o:
@michalkurka this is WIP, but feel free to provide any feedback/idea for this

h2o-bindings/bin/custom/python/gen_gbm.py
angela0xdata: So to provide an example for `balance_classes` in GBM, we can just add to that flat file or dict?
(flat file example)

```
property_balance_classes_examples = """
>>> cars = h2o.import_file("https://s3.amazonaws.com/h2o-public-test-data/smalldata/junit/cars_20mpg.csv")
>>> cars["economy_20mpg"] = cars["economy_20mpg"].asfactor()
>>> predictors = ["displacement","power","weight","acceleration","year"]
>>> response = "economy_20mpg"
>>> train, valid = cars.split_frame(ratios = [.8], seed = 1234)
>>> cars_gbm = H2OGradientBoostingEstimator(balance_classes=True, seed = 1234)
>>> cars_gbm.train(x = predictors,
...                y = response,
...                training_frame = train,
...                validation_frame = valid)
>>> cars_gbm.auc(valid=True)
```

h2o-bindings/bin/custom/python/gen_aggregator.py
```diff
@@ -0,0 +1,13 @@
+class_init_extra = """
+self._parms["_rest_version"] = 99
+"""
+
+class_extras = """
+@property
```
michalkurka: Is there any chance we could define this as actual python code instead of strings?

We could have a mixin class that would be added to the generated one.

h2o-bindings/bin/custom/python/gen_gbm.py
seb-h2o: @angela0xdata 👍 for (1) or for (2)? :)
I won't support both.
As I'm externalizing more of those customizations, I tend to go towards this dictionary structure:
```
properties = dict(
     training_frame = dict(
          examples = """
>>> statement
>>> another statement
         """,
    ),
    validation_frame = dict(
          examples = """
>>> statement
>>> another statement
         """,
    ),
    some_property_with_many_customizations = dict(
        doc = """ This property also does this and that""",  # doc will be added to getter only
        examples = """cf. above""",  # examples will be added to getter only
        getter = """
        if isinstance(prop, str):
            return converted_to_list
        else:
            return prop
        """,
        setter = """
        blah blah blah
        """
    ),
)
```

it hurts a bit more the eyes here, but looks fine with color syntax. 
Would it be fine for you?

angela0xdata: Either way would work for me, and I have no preference :) ... I just wanted to make sure I understood correctly.

h2o-bindings/bin/custom/python/gen_aggregator.py
seb-h2o: @michalkurka yeah, I've been thinking about something like that too.
For now, my approach works for any language as these are just strings.

But for python we could do much better, especially for those extras:
I could also declare those extras as a proper mixin as you suggest (would appear as code in this file), copy the source code using `inspect` module and add the mixin in the base classes dynamically.
Will definitely try it later.

michalkurka: cool, I think this is definitely a step in a good direction

michalkurka:
@seb-h2o I don't see any issue with the proposed direction - thanks for pushing us forward!

h2o-bindings/bin/custom/python/gen_gbm.py
seb-h2o: @angela0xdata yep, I didn't see your previous comment, you got it right

seb-h2o:
@michalkurka I think I'm done with this one: the R part was a real pain due to the barely intelligible logic there: I hope it's a bit better now.
Please see PR descriptions for how the template works.
There's almost no changes in generated code outside indentation fixes and better code layout.

The custom documentation and examples are separated from code to make it easier for writers.

Finally, for Python, I could move most custom code into proper functions, although it mainly helps for color syntax (those functions are not runnable directly due to import issues).

I'd like to be able to merge this soon (before my vacations): also maybe apply the changes to `rel-yau` to avoid painful conflicts each time `rel-yau` makes a change in the code generators.

michalkurka:
@seb-h2o looks really good - for the sake of the review, do you think it would it would be possible to separate the generated files in a separate commit?

seb-h2o:
@michalkurka separate commit, I can try. Separate PR definitely not possible.

seb-h2o:
@michalkurka https://github.com/h2oai/h2o-3/pull/3788/commits/c9ecbb877d1de7dbdcc8cd09e4a731dfe38c2796
provides this, except that the diff is reverted (new version on the left, master on the right): commit view is always with the previous one, I can't figure out how to get a diff with master per commit...

angela0xdata:
Thank you, @seb-h2o!!!

hannah-tillman:
Looks awesome--thank you so much! :D

PR #3793, opened 2019-08-09
HEXDEV-740: AstGroup OOM

wendycwong:
This PR completes step 2 of JIRA https://0xdata.atlassian.net/browse/HEXDEV-740?filter=-1

I have replaced the Group G key _gs with byte[][] from double[].

### DIFF
--- a/h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
@@ -1,7 +1,10 @@
 package water.rapids.ast.prims.mungers;
 
 import water.*;
-import water.fvec.*;
+import water.fvec.Chunk;
+import water.fvec.Frame;
+import water.fvec.NewChunk;
+import water.fvec.Vec;
 import water.rapids.Env;
 import water.rapids.Merge;
 import water.rapids.Val;
@@ -16,6 +19,7 @@ import water.util.IcedHashMap;
 import water.util.Log;
 
 import java.util.Arrays;
+import java.io.FileOutputStream;
 
 /**
  * GroupBy
@@ -80,7 +84,7 @@ public class AstGroup extends AstPrimitive {
         return ds[0] / n;
       }
     },
-    sum() {
+    sum() { // no need for ns
       @Override
       public void op(double[] d0s, double d1) {
         d0s[0] += d1;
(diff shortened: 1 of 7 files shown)


h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -709,32 +721,63 @@ public G(int ncols, AGG[] aggs, boolean hasMedian) {
       }
     }
 
+    public byte[] convertRowVal2ByteArr(Chunk chk, int row) {
```
wendycwong: change a double to byte array using AutoBuffer.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -407,6 +413,16 @@ public boolean equals(Object o) {
       });
   }
 
+  public static double convertByte2Double(byte[] val) {
```
wendycwong: Change keys in byte[] back to double

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -673,24 +688,20 @@ public MedianResult(int len) {
...
-    public NAHandling[] _na;*/
     public MedianResult medianR = null;
 
     public G(int ncols, AGG[] aggs) {
       this(ncols, aggs, false);
     }
 
     public G(int ncols, AGG[] aggs, boolean hasMedian) {
-      _gs = new double[ncols];
+      _gsB = new byte[ncols][];
```
michalkurka: I thought we would use just a single byte array to store everything -this way you would get the most efficient compressed representation

the layout is known - we do know how many bytes we need to represent each value

wendycwong: I thought about it but the problem is that we actually use the actual value of each column to perform calculations in later parts.  If you combine all the keys into one byte array, I will need another array to remember which column bytes starts and which one ends....

michalkurka: yes, however you need this array just once - all rows can share the same layout

wendycwong:
Will focus of Hierarchical GLM.  Michalk will take over.

PR #3794, opened 2019-08-09
PUBDEV-6776 - Introduce Transform operation together with Target Encoding Model

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6776

## Summary
The goal of this PR is to expose Target Encoder functionality to our users in Python and R. We're deprecating the custom-built API in R and re-using mechanisms around `Model` class, which are already built into H2O, tested and kind-of time-proven.

This PR introduces a new function named `transform` to generalize behavior of our Models - with TE, models are not 100% predictive. R API that already existed remains the same (already existing tests written by Andrej are kept to prove that). Internally, `transform` only delegates to `predict`. The whole change is only about correct naming. In R, there was a small problem, as transform function had already existed for `Word2Vec`. I think I've solved it using ellipsis. 

Other than that, the rest of this PR is just some heavy lifting regarding integration of TE as a model.

Details are described below.

## TE as a Model
Previously, Target Encoder implementation has not been exposed via H2O Model.
:heavy_check_mark:  From technical point of view, `TargetEncoder` can be exposed via `TargetEncoderModel` just fine. This has several effects:

1. R & Python code generators are re-used.
2. Both clients may use built-in validations for free, thanks to code generation mechanics mentioned in previous point.
3. H2O Model has many tested features already built-in. ColSpecifiers, enum validators and some others, to name a few.
4. It is automatically seen in the list of algorithms (currently marked as BETA in this PR).

Unnecessary built-in procedures that are not needed not only for TargetEncoder, but to any data munging model.

## Problems & motivation
Generally speaking, data mungers can also be though of as Models (please discuss). However, there is **one flaw**. H2O's models were built as `predictive` models. Therefore, the methods/functions exposed on the client are called `predict`. That is **inappropriate naming** for a model that only does data transformation distinct from any prediction.

In Java, we only use `score` or `predict` as well.

The solution to his could be the introduction of a new method, named `transform`.  The motivation is very simple:

- With the introduction of Target Encoder, not all models are predictive.
- Target Encoder is probably not the last data non-predictive, data-transforming model.

## What was done

Therefore, **in this PR**, a new `transform` function has been introduced. Currently, it is only available in Python and R. Transform delegates to "predict" and only acts as a facade.

In R, there was one specific problem to solve - such function already existed for `Word2Vec` algorithm. I've found a way to make the old API work. Now, there is a new, word2Vec specific function named `h2o.transform_word2vec` that has the old signature and behavior. The new, general function named `transform` only delegates the job to `h2o.transform_word2vec` when `Word2Vec` model is passed as an argument inside `h2o.transform`. The general `h2o.transform` function still supports original parameters via `ellipsis` - a nice thing to have in R !

Python usage demonstration: https://github.com/h2oai/h2o-3/pull/3794/files#diff-5dc1dbde5cd9b87fc791979b111e1572

R usage demonstration: https://github.com/h2oai/h2o-3/pull/3794/files#diff-4154e4115d963046d7c8a1d7d16f65b7

In Java, I propose to keep the "score" methods for now, as we can change it internally a little bit later, once we're 100% sure and gather some feedback. Those are hidden to users anyway.

## Tests

For both R and Python, the following behavior of TE model is tested. This is not an exhaustive list, there are more tests included.

1. Model creation
1. Applying the TE map, aka calling `transform`.
1. MOJO download
1. Various parameters, including fold column.
1. Transform function works on various models and gives the same result as calling predict.
1. Calling `predict` on TargetEncoder won't fail and will give the correct answer.
1. New transformation function for Word2VEc
1. Old Word2Vec API is still functional (old tests are working)

## Other stuff

Target encoder model parameters have their naming changed, because H2O was unable to map them. This is solveable, but decision was made by me and @michalkurka to rename them to H2O standards.

I did not know what to put into model summary table. So there is a map of the old columns to the new ones.

```
Target Encoder model summary.: Summary for target encoder model

original_name    encoded_column_name
---------------  ---------------------
cabin            cabin_te
embarked         embarked_te
home.dest        home.dest_te
```


h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
```diff
@@ -60,11 +61,43 @@ public TargetEncoder (String[] columnNamesToEncode) {
       this(columnNamesToEncode, new BlendingParams(20, 10));
     }
 
-    public static class DataLeakageHandlingStrategy {
-        public static final byte LeaveOneOut  =  0;
-        public static final byte KFold  =  1;
-        public static final byte None  =  2;
+  public enum DataLeakageHandlingStrategy {
```
Pscheidl: Enums are more descriptive and easier to check against. So I've introduced it. The byte inside is only there for legacy reasons to keep the old API honor and intact. I believe the only external users are now Sparkling water people (cc @jakubhava ) and if they say we don't need the by-level API, I'll just remove it. There is a method to migrate from the old API to the new one, used on some places of the code. Could serve the same purpose.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
```diff
@@ -508,6 +541,18 @@ public void map(Chunk cs[]) {
       }
     }
 
+  public Frame applyTargetEncoding(Frame data,
```
Pscheidl: We needed target encoding to respect key of the frame created when necessary. This method is here to represent the old API and not break anything. Used mainly by tests and maybe by Sparkling water.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
```diff
@@ -520,27 +565,30 @@ public void map(Chunk cs[]) {
...
                                      Map<String, Frame> columnToEncodingMap,
-                                     byte dataLeakageHandlingStrategy,
+                                     DataLeakageHandlingStrategy dataLeakageHandlingStrategy,
                                      String foldColumnName,
                                      boolean withBlendedAvg,
                                      double noiseLevel,
-                                     boolean imputeNAsWithNewCategory,
-                                     long seed) {
+                                     long seed,
+                                     Key<Frame> encodedFrameKey) {
```
Pscheidl: There might be a desire to have the output frame named as desired. Attempt was made to leave the original code as intact as possible and just make the final frame be known in DKV under this encoded frame key.

Pscheidl: Plase note `imputeNAsWithNewCategory` was unused parameter. I've removed it and with it made the appropriate adjustments to other parts of code (including tests).

My model API call would have to provide the value, so it was interfering with any new code. That's why it is removed in this API.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
```diff
@@ -520,27 +565,30 @@ public void map(Chunk cs[]) {
...
 
         if(noiseLevel < 0 )
             throw new IllegalStateException("`_noiseLevel` must be non-negative");
 
         Frame dataWithAllEncodings = null;
         try {
+          
+          if(encodedFrameKey == null){
+            encodedFrameKey = Key.make();
```
Pscheidl: But we do have support for no encodedFrameKey provided - in that case, we'll just make up our own !

h2o-core/src/main/java/hex/Model.java
```diff
@@ -53,6 +53,10 @@
   protected ScoringInfo[] scoringInfo;
   public IcedHashMap<Key, String> _toDelete = new IcedHashMap<>();
 
+  public enum Type {
+    PREDICTIVE, DATA_TRANSFORMING
```
Pscheidl: From now on, models won't be just predictors, but also data transformers. There is no transform() method yet, but this is just a matter of API changes right now. The method `predict` is going to behave as `transform` for now.

At several places, there is a need to know what kind of model we're working with. E.g. to create a name for the destination frame, where for data transformers, it is not desired to have the resulting data frame named with prefix "predictions_".

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
deil87: @Pscheidl `imputeNAsWithNewCategory` was hanging there due to a limitation in our merge implementation. Take a look at jira https://0xdata.atlassian.net/browse/PUBDEV-6729. The point is that we in the end of the day want this option/parameter to be available for users.

Pscheidl: When I think about it, Sparkling Water might integrate on it and this would break the API for them.

Otherwise, the preferred way to access TE from now now would be Model and its parameters. If there was no SW, we could dump the parameter anyway for now and move it back there when it actually does something.

h2o-core/src/main/java/hex/Model.java
deil87: @Pscheidl I would suggest to separate these two types by leveraging class hierarchy. This would allow us to keep code that is specific to a particular model type in a separate class/hierarchy branch. With just `enum` the separation is not strong enough and everything will be in one place/class. Just thinking out loud.  @michalkurka  wdyt?

Pscheidl: How'd you want to do that, without throwing away everything from current Model class ?
Currently, there is a clear hierarchy. The new one would require creation of new services around a new type of model, including frontend code generation mechanisms and other stuff. 

 The plan is to bring a new general method named `transform` to current models, so that it can be used from now on and delegate it's behavior to former `predict` or `score` where necessary (in all previous predictive models). This way, it won't take forever to implement it.

Btw. you can see not transform method yet, because it's not yet implemented, as this PR is a Work In Progress currently.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
deil87: @Pscheidl I agree with you that enums are more convenient but for ultimate performance passing bytes(from clients or wherever) is more efficient. I was inspired by h2o-3 code base to use it this way.  But truth is somewhere in between of the usability and performance. @michalkurka Wanted to know if we have any considerations or conventions about it on the project?

h2o-core/src/main/java/hex/Model.java
deil87: @Pscheidl Why throwing away? Moving is not throwing. What if we had code generation for one class that is doing everything in our app.... sooner or later we would want to leverage benefits of OOP and introduce new classes.  I'm just suggesting **to consider** this option....because what you are doing is just a half step. The fact that there are a lot of things that needs to be changed is an obstacle of course but it doesn't mean that what I'm suggesting is a wrong thing. Though it might be a wrong thing but it should not be argued against with just time expenses.
Regarding general method that will delegate to either score or transform based on enum's value.... is a sign that you are combining two things that you already have acknowledged to be a distinguishable things. You are increasing complexity of the class Model. OOP approach is to expand hierarchy and make it easier to reason as soon as we have opportunity to separate things. Having `transform` and `score` methods in two separate classes that are calling same parent's method - lets say 'process` or again `transform` -would be better IMHO.
I know that this PR is in progress but you are writing comments for others and I think that discussing them along the way is better than at the end of the work.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
deil87: I agree, just giving you more context.

h2o-core/src/main/java/hex/Model.java
Pscheidl: The transform method will exist algonside predict and score, not replacing it. Nor there will be a router (at the moment). Again, we're talking about WIP here.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoder.java
Pscheidl: Sure. I am working with a Model and from user's perspective, I wanted strings with as few mapping layers in between as possible.

h2o-core/src/main/java/hex/Model.java
Pscheidl: The complexity of such task is non-trivial. This has various reasons and there are not few.

Most of the current algos somehow depend on current `Model` class. What is there may not be used by all, but every part is used by some of the underlying implementations.  H2O is designed this way. And this affects the whole core and algos module related to models. For example parameters. Some of the parameters are unused by some of the models (obviously GenericModel for example). This way, we might squeeze any transformation machine in.

Creating a new predecessor only for transformations with a common method for transformations/predictions with existing models, one would have to do something like:

```
   Transformation
      <                 >
   <                      >
Model                  DataTransformation
```

Which effectively means creating new frontend generators, new V3 generators and checks (or new piece of Rest API to get rid of V3 objects if required). New frontend validation and display logic (or reuse it and wire it, not trivial in any case). Make it compatible with Flow. And test it all.

This is what I wanted to avoid, as I still believe a Model can still describe a universal transformation and the `train` and `computeImpl` methods actually allow us to do any customizations we want internally. All that needs to be done is to expose new `transform` function, that will be usable both for predicting and for data transformation. On the client. I find this much easier, much less error prone and much more likely to deliver in time. 

Your solution sounds nicer, of course. But it basically means writing and re-writing lots of code, which requires lots of time, testing and eventually bugfixing.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoderBuilder.java
```diff
@@ -22,6 +23,22 @@ public TargetEncoderBuilder(TargetEncoderModel.TargetEncoderParameters parms) {
...
 
+  public TargetEncoderBuilder(final boolean startupOnce) {
+    super(new TargetEncoderModel.TargetEncoderParameters(), startupOnce);
+  }
+
+  @Override
+  public void init(boolean expensive) {
+    super.init(expensive);
+
+    if (_parms._leakageHandlingStrategy == null || _parms._leakageHandlingStrategy.trim().isEmpty()) {
```
michalkurka: Why is this not an enum?

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoderModel.java
```diff
@@ -109,23 +111,34 @@ public int nfeatures() {
...
   }
   
   @Override
   protected double[] score0(double data[], double preds[]){
     throw new UnsupportedOperationException("TargetEncoderModel doesn't support scoring. Use `transform()` instead.");
   }
 
+  @Override
+  public Frame score(Frame fr, String destination_key, Job j, boolean computeMetrics, CFuncRef customMetricFunc) throws IllegalArgumentException {
+    final TargetEncoder.DataLeakageHandlingStrategy leakageHandlingStrategy = TargetEncoder.DataLeakageHandlingStrategy.valueOf(_parms._leakageHandlingStrategy);
```
michalkurka: where is the frame adapted?

h2o-automl/src/main/java/hex/schemas/BlendingParamsV3.java
```diff
@@ -0,0 +1,13 @@
+package hex.schemas;
+
+import ai.h2o.automl.targetencoding.BlendingParams;
+import water.api.API;
+import water.api.schemas3.SchemaV3;
+
+public class BlendingParamsV3 extends SchemaV3<BlendingParams, BlendingParamsV3> {
+  
+  @API(help="K blending parameter", direction=API.Direction.OUTPUT)
```
michalkurka: direction "output"?

h2o-automl/src/main/java/hex/schemas/TargetEncoderModelOutputV3.java
```diff
@@ -0,0 +1,7 @@
+package hex.schemas;
+
+import ai.h2o.automl.targetencoding.TargetEncoderModel;
+import water.api.schemas3.ModelOutputSchemaV3;
+
+public class TargetEncoderModelOutputV3 extends ModelOutputSchemaV3<TargetEncoderModel.TargetEncoderOutput, TargetEncoderModelOutputV3> {
```
michalkurka: empty output? is this final?

h2o-core/src/main/java/hex/Model.java
deil87: @Pscheidl As usual, just by always following the easiest and fastest way eventually you will find yourself in a big unbearable trouble. I'm sure you know about this. All what I wanted is to bring up/discuss another option. I believe that just by bearing in mind multiple options we can end up with a better solution.

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
```diff
@@ -0,0 +1,36 @@
...
+import ai.h2o.automl.targetencoding.TargetEncoderModel;
+import water.api.API;
+import water.api.schemas3.ModelParametersSchemaV3;
+
+import java.util.List;
+
+public class TargetEncoderV3 extends ModelBuilderSchema<TargetEncoderBuilder, TargetEncoderV3, TargetEncoderV3.TargetEncoderParametersV3> {
+  public static class TargetEncoderParametersV3 extends ModelParametersSchemaV3<TargetEncoderModel.TargetEncoderParameters, TargetEncoderParametersV3> {
+    
+    @API(help = "Is blending used ? True by default.", mapsTo = "_withBlending")
```
michalkurka: it would be better to rename the parameters to conform to H2O's naming conventions (eg. `_withBlending` violates the rules).

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
```diff
@@ -0,0 +1,36 @@
...
+    public boolean blending;
+    
+    @API(help = "Columnds to encode.", mapsTo = "_columnNamesToEncode")
+    public String[] encoded_columns;
+    
+    @API(help = "Target column for the encoding", mapsTo = "_response_column")
+    public String target_column;
```
michalkurka: This should probably be FrameV3.ColSpecifierV3 instead.

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.List;
+
+public class TargetEncoderV3 extends ModelBuilderSchema<TargetEncoderBuilder, TargetEncoderV3, TargetEncoderV3.TargetEncoderParametersV3> {
+  public static class TargetEncoderParametersV3 extends ModelParametersSchemaV3<TargetEncoderModel.TargetEncoderParameters, TargetEncoderParametersV3> {
+    
+    @API(help = "Is blending used ? True by default.", mapsTo = "_withBlending")
+    public boolean blending;
+    
+    @API(help = "Columnds to encode.", mapsTo = "_columnNamesToEncode")
+    public String[] encoded_columns;
```
michalkurka: ColSpecifierV3[]?

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
```diff
@@ -0,0 +1,36 @@
...
+    
+    @API(help = "Target column for the encoding", mapsTo = "_response_column")
+    public String target_column;
+    
+    @API(help = "Parameters used for blending (if enabled). Blending is to be enabled separaterly using the 'blending' parameter.", mapsTo = "_blendingParams")
+    public BlendingParamsV3 blending_parameters;
+
+    @API(help = "Data leakage handling strategy. Default to None.", mapsTo = "_leakageHandlingStrategy")
+    public String data_leakage_handling;
```
michalkurka: should be an enum

h2o-core/src/main/java/hex/Model.java
```diff
@@ -53,6 +53,10 @@
   protected ScoringInfo[] scoringInfo;
   public IcedHashMap<Key, String> _toDelete = new IcedHashMap<>();
 
+  public enum Type {
```
michalkurka: what is this used for?

h2o-core/src/main/java/water/api/API.java
```diff
@@ -77,4 +77,6 @@
    * Identify grid-able parameter.
    */
   boolean gridable() default false;
+  
+  String mapsTo() default "";
```
michalkurka: please no! just conform to H2O naming conventions

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
```diff
@@ -367,7 +367,15 @@ public JobV3 predictAsync(int version, final ModelMetricsListSchemaV3 s) {
       if (parms._exemplar_index >= 0) {
         parms._predictions_name = "members_" + parms._model._key.toString() + "_for_exemplar_" + parms._exemplar_index;
       } else {
-        parms._predictions_name = "predictions" + Key.make().toString().substring(0, 5) + "_" + parms._model._key.toString() + "_on_" + parms._frame._key.toString();
+        switch (parms._model.getModelType()) {
```
michalkurka: This seems to be the only usage of model type. Is this really worth complicating the API?

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
```diff
@@ -0,0 +1,36 @@
...
+  
+    @Override
+    public String[] fields() {
```
michalkurka: Please explain - why is this not needed in other models? Is TE special in any way? Can it be just like others? :)

h2o-py/h2o/estimators/targetencoder.py
```diff
@@ -0,0 +1,113 @@
...
+            elif pname in names_list:
+                # Using setattr(...) will invoke type-checking of the arguments
+                setattr(self, pname, pvalue)
+            else:
+                raise H2OValueError("Unknown parameter %s = %r" % (pname, pvalue))
+
+    @property
+    def blending(self):
+        """
+        Is blending used ? True by default.
```
michalkurka: Default seems to be false ;)

h2o-py/h2o/model/metrics_base.py
```diff
@@ -761,3 +761,8 @@ class H2OCoxPHModelMetrics(MetricsBase):
     def __init__(self, metric_json, on=None, algo=""):
         super(H2OCoxPHModelMetrics, self).__init__(metric_json, on, algo)
 
+class H2OTargetEncodermetrics(MetricsBase):
+
+    def __init__(self, metric_json, on=None, algo=""):
+        super(H2OCoxPHModelMetrics, self).__init__(metric_json, on, algo)
```
michalkurka: copy&paste error

h2o-r/h2o-package/R/targetencoder.R
```diff
@@ -1,166 +1,50 @@
...
+#' @param x A vector containing the \code{character} names of the predictors in the model.
+#' @param blending \code{Logical}. Is blending used ? True by default. Defaults to FALSE.
+#' @param encoded_columns Columnds to encode.
+#' @param target_column Target column for the encoding
+#' @param blending_parameters Parameters used for blending (if enabled). Blending is to be enabled separaterly using the 'blending'
+#'        parameter.
+#' @param data_leakage_handling Data leakage handling strategy. Default to None.
+#' @param model_id Destination id for this model; auto-generated if not specified.
 #' @export
-h2o.target_encode_fit <- function(frame, x, y, fold_column = NULL)
```
michalkurka: are we deprecating the old API???

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoderModel.java
Pscheidl: This is indeed a bug. I wrongly thought TargetEncoder is going to do the adaptation itself. Or more precisely, it will do the validation on it's own. 

But it is indeed advantageous to do to he adaptation.

h2o-py/h2o/estimators/targetencoder.py
Pscheidl: Corrected at the correct place.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoderModel.java
Pscheidl: Nowehere. A bug. Adaptation should indeed be done. Adding adaptation revealed some other bugs.

Pscheidl: Adaptation may remove some columns from the remaining dataset, which might be useful for making predictions, but not for transformations like target encoding, where necessary predecessor is the dataset having the required features in place.

I'm also suspicious some features of the adaptation process, like categorical levels matching might even hurt the process of Target Encoding. This is all up to the `TargetEncoder` class, which should do the validation.

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
Pscheidl: This can be done without any delay.

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
Pscheidl: `08-14 11:08:16.759 192.168.0.103:54321   20866  #56930-17 FATAL: Don't know how to cast an array of: class water.api.schemas3.FrameV3$ColSpecifierV3 to an array of: class java.lang.String`

Even classic Model.Params do not use array of ColSpecifierV3. I'll try to add the functionality and if it will take too long, use string as the rest of the Model.Parameters instead.

Pscheidl: I see, VecSpecifiers have to be used. Ok, the code can be adapted without problems.

h2o-r/h2o-package/R/targetencoder.R
Pscheidl: I see. It is overridden by the autogenerated changes. It should still be there, only delegate to the new API and mark it as deprecated.

h2o-py/h2o/estimators/targetencoder.py
Pscheidl: Default is true.

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
Pscheidl: Please note this is present in all `*ParametersV3` classes, as there is a need to specify only the exposed fields.

I've only introduced `extractDeclaredApiParameters(getClass());` helper method (this is called during startup & code generation only) to make it a little bit easier to specify declared fields + some selected inherited ones.

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
Pscheidl: Done.

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
Pscheidl: Removed.

h2o-automl/src/main/java/hex/schemas/TargetEncoderModelOutputV3.java
Pscheidl: Added only prior_mean and renamed ModelOutput to make proper implicit attribute mapping possible. 

Note (to self): We should prbably provide proper model summary for TargetEncoderModel.

Pscheidl: @deil87 Important question: Please have at what goes into model summary and model output in general and let me know if there is something more to expose and I'll do it.

h2o-core/src/main/java/hex/Model.java
Pscheidl: removed.

h2o-automl/src/main/java/hex/schemas/TargetEncoderModelOutputV3.java
deil87: @Pscheidl As we discussed in Slack... prior mean is probably the only reasonable thing for now. Encoding map might be too big.... maybe truncated version is ok but doesn't make a lot of sense. Info about Blending could be exposed through exposure of model's parameters. We can provide some helpful summary but we will need to compute these statistics e.g. number of instances which were imputed with prior mean; how many categories were represented with 1 instance, 2 instances etc.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoderModelAlgoRegister.java
```diff
@@ -0,0 +1,20 @@
...
+import water.api.AlgoAbstractRegister;
+import water.api.RestApiContext;
+import water.api.SchemaServer;
+
+public class TargetEncoderModelAlgoRegister extends AlgoAbstractRegister {
+  
+  @Override
+  public void registerEndPoints(RestApiContext context) {
+    ModelBuilder targetEncoderModelBuilder = new TargetEncoderBuilder(true);
+    registerModelBuilder(context, targetEncoderModelBuilder, SchemaServer.getStableVersion());
```
honzasterba: why not use RegisterAlgos?

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
```diff
@@ -0,0 +1,61 @@
...
+    @API(help = "Is blending used ? True if blending is used, false if not. True by default.")
+    public boolean blending;
+    
+    @API(help = "Columnds to encode.")
+    public FrameV3.ColSpecifierV3[] encoded_columns;
+    
+    @API(help = "Target column for the encoding")
+    public FrameV3.ColSpecifierV3 target_column;
+    
+    @API(help = "Parameter 'f' used for blending (if enabled). Blending is to be enabled separaterly using the 'blending' parameter.")
```
honzasterba: seems like docs for f and k are misplaced

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoderModelAlgoRegister.java
Pscheidl: Not available there, would require playing with dependencies. But your comment raises another question:

We have this functionality in the AutoML module and the rest of AutoML behaves the same way (acts as an extension, even though our core modules directly depend on it and could be considered part of core) - it is registered as every H2O extension is.

However, even though Target Encoding was primarily meant for AutoML, it makes sense to me to relocate it to core algos (can be done any day), as it's definitely not tied to AutoML and usable **independently** by our users.

WDYT everoyne ?

h2o-automl/src/main/java/hex/schemas/TargetEncoderV3.java
Pscheidl: Thank you. Resolved.

h2o-automl/src/main/java/ai/h2o/automl/targetencoding/TargetEncoderModelAlgoRegister.java
honzasterba: exactly my thinging, if we are exposing this as a regular algo it should live in core

Pscheidl: Made sense to me from the beginning. Let's make this bold move and make it part of H2O Core, as AutoML is of great importance. :)

Pscheidl: First, I need to make sure SW is ok with it @mn-mikke .

Pscheidl: Done. Sparkling Water is fine with it.

h2o-py/h2o/model/metrics_base.py
```diff
@@ -761,3 +761,8 @@ class H2OCoxPHModelMetrics(MetricsBase):
     def __init__(self, metric_json, on=None, algo=""):
         super(H2OCoxPHModelMetrics, self).__init__(metric_json, on, algo)
 
+class H2OTargetEncodermetrics(MetricsBase):
```
maurever: Just a typo:

```suggestion
class H2OTargetEncoderMetrics(MetricsBase):
```

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_target_encoding_model.py
```diff
@@ -0,0 +1,59 @@
...
+    assert transformed is not None
+    print(transformed.names)
+    assert transformed.ncols == trainingFrame.ncols + len(teColumns)
+    for te_col in teColumns:
+        assert te_col + "_te" in transformed.names
+    
+    assert transformed.nrows == 1309
+    
+    # Test fold_column proper handling + kfold data leakage strategy defined
+    te = H2OTargetencoderEstimator(encoded_columns = teColumns, target_column = targetColumnName, fold_column = "pclass",
```
ledell: In regular estimators in Python we have "data related" arguments in the `.train()` method and not the Estimator constuctor (including exact parameters like `fold_column`.  Maybe that's something we want to follow here as well for some or all of the data related parameters?

h2o-r/tests/testdir_algos/targetencoder/runit_targetencoder_model.R
```diff
@@ -0,0 +1,26 @@
...
+    data <- h2o.importFile(path = locate('smalldata/gbm_test/titanic.csv'), col.types=list(by.col.name=c("survived"),types=c("factor")))
+    encoded_columns <- c('home.dest', 'cabin', 'embarked')
+    target_encoder <- h2o.targetencoder(training_frame = data, encoded_columns= encoded_columns, target_column = "survived")
+    encoded_data <- h2o.transform(target_encoder, data) # For now, there is only the predict method
+    expect_false(is.null(encoded_data))
+    expect_equal(h2o.ncol(data) + length(encoded_columns), h2o.ncol(encoded_data))
+    expect_true(h2o.nrow(data) == 1309)
+    
+    # Test fold_column proper handling + kfold data leakage strategy defined
+    target_encoder <- h2o.targetencoder(training_frame = data, encoded_columns= encoded_columns, target_column = "survived",
```
ledell: We now have three different TE interfaces in R... @meganjkurka's original, @deil87's recent one and now the `h2o.targetencoder()`.  @Pscheidl I assume making a new API is the only way to achieve what you're trying to do?

ledell:
@Pscheidl Thank you very much for the detailed description of this PR.  I left some comments about the API (we might want to make some changes to Python), but overall I think it's good.

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_target_encoding_model.py
Pscheidl: Thank you. Done.

Now, our users will be able to do:

```
    te.train(training_frame=trainingFrame, fold_column="pclass", data_leakage_handling="kfold",
             target_column=targetColumnName, encoded_columns=teColumns)
```

h2o-py/h2o/model/metrics_base.py
Pscheidl: Done. Thank you.

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_target_encoding_model.py
Pscheidl: Tested as well.

h2o-r/tests/testdir_algos/targetencoder/runit_targetencoder_model.R
Pscheidl: True. In history, there were other attempts to bypass model auto-generated API due to various reasons. Word2Vec was one of them, TargetEncoder is the most recent API.

This PR, besides integrating TE as a Model, it also paves the way (not only by introducing a delegating `transform method, but also by many enhancements on our backend) for future transformers of all kinds to be usable as models. It will also serve as a showcase for others on how to re-use what we already have. This is **what I tried to achieve**, together with unification of the APIs in future.

Therefore, if I want to unify current and future models under one API and reuse current ecosystem at the same time, avoiding the long journey it takes to develop, test and polish new API, there is no other way to me. If there was, I'd take it.

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_target_encoding_model.py
```diff
@@ -36,10 +36,13 @@ def test_target_encoding_fit_method():
     assert transformed.nrows == 1309
     
     # Test fold_column proper handling + kfold data leakage strategy defined
-    te = H2OTargetencoderEstimator(encoded_columns = teColumns, target_column = targetColumnName, fold_column = "pclass",
-                                   k = 0.7, f = 0.3, data_leakage_handling = "kfold")
-    te.train(training_frame = trainingFrame)
+    te = H2OTargetencoderEstimator(k=0.7, f=0.3)
+    te.train(training_frame=trainingFrame, fold_column="pclass", data_leakage_handling="kfold",
```
ledell: I think it's debatable whether `data_leakage_handling` belongs in the constructor or the train method... but I think we just put arguments that relate to the given `training_frame` in `.train()`.  So my instinct is to keep `data_leakage_handling` in the constructor since it's a characteristic of the TE process, rather than the frame you are operating on.

h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -108,24 +108,30 @@ def train(self, x=None, y=None, training_frame=None, offset_column=None, fold_co
...
         has_default_training_frame = hasattr(self, 'training_frame') and self.training_frame is not None
         training_frame = H2OFrame._validate(training_frame, 'training_frame',
                                             required=self._requires_training_frame() and not has_default_training_frame)
         validation_frame = H2OFrame._validate(validation_frame, 'validation_frame')
         algo = self.algo
         parms = self._parms.copy()
         if(algo == "targetencoder"):
-            x = parms["encoded_columns"]
-            y = parms["target_column"]
+            if ("data_leakage_handling" in kwargs):
```
ledell: I think it's debatable whether `data_leakage_handling` belongs in the constructor or the train method... but I think we just put arguments that relate to the given `training_frame` in `.train()`.  So my instinct is to keep `data_leakage_handling` in the constructor since it's a characteristic of the TE process, rather than the frame you are operating on.

ledell:
@Pscheidl I put one last comment on your latest commit: https://github.com/h2oai/h2o-3/pull/3794/commits/46f4a85a5cfc075e7ce9aa1f537dff4ff204177d

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_target_encoding_model.py
Pscheidl: Removed.

h2o-py/h2o/estimators/estimator_base.py
Pscheidl: Discussed over Slack yesterday, removed from train method's arguments.

h2o-algos/src/main/java/water/rapids/prims/mungers/mungers/AstTargetEncoderTransform.java
```diff
@@ -40,7 +40,7 @@ public ValFrame apply(Env env, Env.StackHelp stk, AstRoot asts[]) {
     Frame[] encodingMapFrames = getEncodingMapFrames(env, stk, asts);
     Frame frame = getFrameToTransform(env, stk, asts);
     String[] teColumnsToEncode = getTEColumns(env, stk, asts);
-    byte dataLeakageHandlingStrategy = getDataLeakageHandlingStrategy(env, stk, asts);
+    TargetEncoder.DataLeakageHandlingStrategy dataLeakageHandlingStrategy = TargetEncoder.DataLeakageHandlingStrategy.fromVal(getDataLeakageHandlingStrategy(env, stk, asts));
```
honzasterba: inside getDataLeakageHandlingStrategy you are doing getVal() and here you construct the enum again using fromVal, seems unnecessary

h2o-algos/src/main/java/water/rapids/prims/mungers/mungers/AstTargetEncoderTransform.java
```diff
@@ -121,14 +121,14 @@ private Frame getFrameToTransform(Env env, Env.StackHelp stk, AstRoot asts[]) {
   private byte getDataLeakageHandlingStrategy(Env env, Env.StackHelp stk, AstRoot asts[]) {
```
honzasterba: I thik this should return DataLeakageHandlingStrategy object, working with the byt val directly is dangerous

h2o-algos/src/test/java/hex/targetencoding/TargetEncoderBuilderTest.java
```diff
@@ -218,7 +222,7 @@ public void transform_KFold_scenario(){
 
       encodingMapFromTargetEncoder = tec.prepareEncodingMap(fr2, responseColumnName, foldColumnName);
 
-      Frame transformedTrainWithTargetEncoder = tec.applyTargetEncoding(fr2, responseColumnName, encodingMapFromTargetEncoder, strategy, foldColumnName, targetEncoderParameters._withBlending, true, seed);
+      Frame transformedTrainWithTargetEncoder = tec.applyTargetEncoding(fr2, responseColumnName, encodingMapFromTargetEncoder, strategy, foldColumnName, targetEncoderParameters._blending,false, seed);
```
honzasterba: whitespace around ,

h2o-algos/src/main/java/water/rapids/prims/mungers/mungers/AstTargetEncoderTransform.java
Pscheidl: This code has been removed only - FYI. I'm not the author. I probably have to ask the author for permission to change it :) Will do :)

h2o-algos/src/main/java/water/rapids/prims/mungers/mungers/AstTargetEncoderTransform.java
Pscheidl: Done, thanks.

h2o-algos/src/main/java/water/rapids/prims/mungers/mungers/AstTargetEncoderTransform.java
Pscheidl: Resolved, thanks.

h2o-algos/src/test/java/hex/targetencoding/TargetEncoderBuilderTest.java
Pscheidl: Reformatted, thanks.

seb-h2o:
@Pscheidl @michalkurka : due to changes in code generation for Py and R, this definitely won't be able to get merged to `master` later: if we want to have this in `rel-yau`, then I suggest to merge https://github.com/h2oai/h2o-3/pull/3808 first, then I'll be happy to help migrating the changes here to the new code generation.

h2o-core/src/main/java/hex/Model.java
seb-h2o: @Pscheidl I know the discussion is outdated, but FYI  it's hard not to consider that `AggregatorModel`, `SVDModel`, `PCAModel`, `GLRMModel` are already actual transformers with an unfortunate `predict` API.

Pscheidl: @seb-h2o Let's create other JIRA once this new transform API is merged. I'll take care of it. Great point, thank you.

h2o-py/h2o/estimators/targetencoder.py
```diff
@@ -0,0 +1,171 @@
...
+    @fold_column.setter
+    def fold_column(self, fold_column):
+        assert_is_type(fold_column, None, str)
+        self._parms["fold_column"] = fold_column
+
+
+
+    def transform(self, frame, data_leakage_handling="None", noise=-1, seed=-1):
+        """
+        Deprecated API. Please use H2OTargetencoderEstimator instead.
```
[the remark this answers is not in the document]
Pscheidl: Yes.

h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -91,7 +91,7 @@ def join(self):
 
     def train(self, x=None, y=None, training_frame=None, offset_column=None, fold_column=None,
               weights_column=None, validation_frame=None, max_runtime_secs=None, ignored_columns=None,
-              model_id=None, verbose=False):
+              model_id=None, verbose=False, **kwargs):
```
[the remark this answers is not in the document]
Pscheidl: I'll use your hook functionality. This is bad design. Thanks for noticing it.

h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -108,17 +108,28 @@ def train(self, x=None, y=None, training_frame=None, offset_column=None, fold_co
...
+        parms = self._parms.copy()
+        if(algo == "targetencoder"):
+            if ("target_column" in kwargs):
+                parms["target_column"] = kwargs["target_column"]
+            if ("encoded_columns" in kwargs):
+                parms["encoded_columns"] = kwargs["encoded_columns"]
+            
+            y = parms["target_column"] if "target_column" in parms else y
+            parms["encoded_columns"] = parms["encoded_columns"] if "encoded_columns" in parms else x
+
```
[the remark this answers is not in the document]
Pscheidl: As mentioned above, I'll use your hook functionality to resolve this. Thank you.

h2o-py/h2o/estimators/estimator_base.py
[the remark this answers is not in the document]
Pscheidl: Removed.

Pscheidl:
@seb-h2o Let me know when it's merged and then let's merge this PR.

Pscheidl:
Rebased on latest rel-yau. There were some merging problems, so I resolved them. 

Me and @seb-h2o both agree on not merging this PR before his (#3808) gets merged.

Pscheidl:
There were quite some changes:

## API Changes
- R API `transform` methods now use S4 `setGeneric` system for dispatching, which is way better then the ellipsis I created before, + S4 is able to differentiate among parameter types.
- Python and R API now allow to change some parameters during `transform` call. IF those specific parameters are not explicitely stated, the ones provided in setup parameters will be taken. If those are not specified as well, defaults are used (this is especially true for blending parameters - those have defaults).
- After quite some investigation and consultation with various people, I've removed one general transform method in both APIs with `...` in R (ellipsis) and `kwargs` in Python. There is a custom handler for the `transform` operation. Such handler will be generated in future, but not now. Current code will only be replaced by the generated one. Goal is to enable users to see full method documentation per each model/transformer. It also separates layers in the code quite nicely, making it shorter and more understandable, with less branches in one method.
- Removed state from `TargetEncoder` class. Part of the state was passed via parameters and part of the state was there from the very moment the `TargetEncoder` object had been created. This made it hard to manage lifecycle of the object and more importantly, manage lifecycle of it's state, as the user might change the state in different phases (first pass as models params, then override with `transform method`, then go back to defaults etc ...),, Currently, the model is holding and managing the state, if any. Not the TargetEncoder, which contains pure logic.
- Some tests were fixed.
- Added R and Python documentation.

Pscheidl:
+ some major and minor bugfixes & polishment. Especially the one mentioned by @seb-h2o was useful. I used hooks for parameter management in Python API, instead of writing it into the general logic.

Pscheidl:
@michalkurka @ledell This is now ready for the final re-review. Sorry for having to review it twice, but the first attempt was not 100% perfect.

Everything @ledell wanted from the API is still there. Only technical stuff changed.

Thank you.

Pscheidl:
Note to self- after the extraction to a module-extension, see if all the tests run correctly.

h2o-r/h2o-package/R/targetencoder.R
```diff
@@ -12,7 +15,16 @@
...
 #' @param data_leakage_handling Data leakage handling strategy. Default to None. Must be one of: "None", "KFold", "LeaveOneOut".
 #' @param model_id Destination id for this model; auto-generated if not specified.
+#' @param training_frame Id of the training data frame.
 #' @param fold_column Column with cross-validation fold index assignment per observation.
+#' @examples
+#' \dontrun{
+#' # library(h2o)
+#' # h2o.init()
+#' 
+#' # encoder = H2OTargetencoderEstimator(encoded_columns = columns, target_column = target_column_name, k = 0.7, f = 0.3, data_leakage_handling = "none")
```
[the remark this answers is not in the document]
Pscheidl: Will be fixed in `gen_R.py`. Thank you.

h2o-r/h2o-package/R/targetencoder.R
```diff
@@ -1,166 +1,68 @@
...
+#' @param data_leakage_handling Data leakage handling strategy. Default to None. Must be one of: "None", "KFold", "LeaveOneOut".
+#' @param model_id Destination id for this model; auto-generated if not specified.
+#' @param training_frame Id of the training data frame.
+#' @param fold_column Column with cross-validation fold index assignment per observation.
+#' @examples
+#' \dontrun{
+#' # library(h2o)
+#' # h2o.init()
+#' 
+#' # encoder = H2OTargetencoderEstimator(encoded_columns = columns, target_column = target_column_name,
```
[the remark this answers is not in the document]
Pscheidl: Resolved in https://github.com/h2oai/h2o-3/pull/3794/commits/9fc4c8ee7268209aa114d52105377fe05523fbd0.

h2o-py/h2o/estimators/targetencoder.py
```diff
@@ -0,0 +1,189 @@
...
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from h2o.estimators.estimator_base import H2OEstimator
+from h2o.exceptions import H2OValueError
+from h2o.frame import H2OFrame
+from h2o.utils.typechecks import assert_is_type, Enum, numeric
+import h2o
+
+
+class H2OTargetencoderEstimator(H2OEstimator):
```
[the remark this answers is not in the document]
Pscheidl: Resolved in https://github.com/h2oai/h2o-3/pull/3794/commits/9fc4c8ee7268209aa114d52105377fe05523fbd0

h2o-bindings/bin/gen_R.py
```diff
@@ -661,6 +670,15 @@ def help_example_for(algo):
         # generic_model <- h2o.genericModel("/path/to/model.zip")
         # predictions <- h2o.predict(generic_model, dataset)
         }"""
+    if algo == "targetencoder":
+         return """\dontrun{
+        # library(h2o)
+        # h2o.init()
+        
+        # encoder = H2OTargetencoderEstimator(encoded_columns = columns, target_column = target_column_name,
```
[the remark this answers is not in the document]
Pscheidl: Resolved in https://github.com/h2oai/h2o-3/commit/2d89e893e91b015024fdded2bb353d9f16d1f3b3. Thank you.

h2o-bindings/bin/gen_R.py
```diff
@@ -666,6 +670,14 @@ def help_example_for(algo):
         # generic_model <- h2o.genericModel("/path/to/model.zip")
         # predictions <- h2o.predict(generic_model, dataset)
         }"""
+    if algo == "targetencoder":
+         return """\dontrun{
+        # library(h2o)
+        # h2o.init()
+        
+        # encoder = H2OTargetencoderEstimator(encoded_columns = columns, target_column = target_column_name, k = 0.7, f = 0.3, data_leakage_handling = "none")
```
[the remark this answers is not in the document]
Pscheidl: Resolved in https://github.com/h2oai/h2o-3/commit/2d89e893e91b015024fdded2bb353d9f16d1f3b3. Thank you for reviewing it.

h2o-r/h2o-package/R/targetencoder_deprecated.R
```diff
@@ -0,0 +1,168 @@
...
+#'
+#' @param frame An H2OFrame object with which to create the target encoding map.
+#' @param x List of categorical column names or indices that we want apply target encoding to. 
+#'          Case when item in the list is a list of multiple columns itself is not supported for now.
+#' @param y The name or column index of the response variable in the frame.
+#' @param fold_column (Optional) The name or column index of the fold column in the frame.
+#' @return Returns an object containing the target encoding mapping for each column in `x`.
+#' @seealso \code{\link{h2o.target_encode_transform}} for applying the target encoding mapping to a frame.
+#' @export
+h2o.target_encode_fit <- function(frame, x, y, fold_column = NULL)
```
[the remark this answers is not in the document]
Pscheidl: We can't change this API I believe. This is Andrej's old API, only moved to a different file and marked as deprecated. There might be people using it.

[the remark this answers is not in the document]
Pscheidl: But if changing the contract is possible, I'll gladly change it. Another option is to add a new method with new name, only delegating to the old method.

ledell:
Hey @Pscheidl I added some comments about a naming inconsistency and some small issues with the R documentation.

Pscheidl:
Ok, this is it, waiting for tests.

Pscheidl:
Two more fixed.
- Forgot to remove a deprecated test was the first failure.
- Accidentally deleted a line in JUnit test - blending was not enabled in a model, but it was for the results the model's output had been compared with. Fixed.

PR #3796, opened 2019-08-10
PUBDEV-6782: Fix swapped by mistake API help messages for ModelParametersSchemaV3 class

deil87:


deil87:
@seb-h2o we discussed this one over Slack. Could you please take a quick look?

PR #3797, opened 2019-08-10
PUBDEV-6783: Fix wrong pyunit_generic_model_mojo_irf.py test and removed code duplication for PUBDEV-6136

deil87:
This PR 

- fixes test which was comparing same models' outputs:

```
with Capturing() as original_output:
        ifr.show()
```
and
```
with Capturing() as generic_output:
        ifr.show()
```

- removes code duplication for Capturing in Py 

- removes code duplication for compare_output function in R

-  removes unused and unnecessary code in SharedTreeModelAttributes

- renamed irf to ifr.


deil87:
@Pscheidl As a side effect of working closely with your code I found and fix some issues. Please take a look. Also one question - do we need this test `/h2o-r/tests/testdir_algos/generic/runit_generic_mojo_convenience.R`?

Pscheidl:
I'm a little bit confused about what is the actual fix and what is renaming. But yes, that part was wrong.

Yes, we need to test the convenience, as the API is still usable by end users.

deil87:
@Pscheidl Because renaming of the file changes ownership and it is difficult to see changes I put explanation into the title and description of the PR. In pyunit_generic_model_mojo_irf.py  test you were comparing same output `ifr.show()` - take a look at your original code. You were using `model` variable for generic model but was comparing same `ifr` model. I have removed code duplication in Py and R and renamed `irf` suffixes into `ifr` as it is I believe a correct abbreviation for **I**solation**F**o**R**est.

deil87:
@Pscheidl Regarding test for convenience... It is just unclear from the body of the test what exactly is supposed to be convenient for the user. It looks like test just checks that both imported and uploaded models produce same (in shape) predictions.

deil87:
@Pscheidl could you please take a look and consider to approve this one?

PR #3798, opened 2019-08-11
PUBDEV-6768 POC graphviz-java for use with Tree visualization in MOJO

koniecsveta:
Task PUBDEV-6768 implementation for review.

https://0xdata.atlassian.net/browse/PUBDEV-6768


h2o-genmodel/build.gradle
```diff
@@ -27,6 +27,8 @@ dependencies {
   testCompile "junit:junit:${junitVersion}"
   compile "net.sf.opencsv:opencsv:2.3"
   compile "com.google.code.gson:gson:2.6.2"
+  compile "guru.nidi:graphviz-java:0.9.0"
```
Pscheidl: This dependency "weights" `172 Kb`.

h2o-genmodel/build.gradle
```diff
@@ -27,6 +27,8 @@ dependencies {
...
+  compile "org.slf4j:slf4j-log4j12:1.7.10"
```
Pscheidl: This dependency could be acceptable as test-scoped. We don't want SLF4J implementations to be introduced to the classpath by H2O - Genmodel uses system output directly. SLF4J dependencies are even specifically excluded few lines below.

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -1,32 +1,45 @@
...
+  private PrintTreeOptions pTreeOptions;
+  private boolean internal;
+  private String outputPngName = null;
+  private static final String tmpOutputFileName = "tmpOutputFileName.gv";
 
   public static void main(String[] args) {
+    // Provide configuration for graphviz logging
+    BasicConfigurator.configure();
+    // Select a way to process dot files
+    Graphviz.useEngine(new GraphvizJdkEngine());
```
Pscheidl: We could do this lazily, e.g. around line `250`. Only when the graph is actually printed to a file.

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -189,26 +208,51 @@ private void validateArgs() {
...
+    SharedTreeGraphConverter treeBackedModel = (SharedTreeGraphConverter) genModel;
+    final SharedTreeGraph g = treeBackedModel.convert(treeToPrint, null);
+    if (printRaw) {
+      g.print();
+    }
+    g.printDot(os, maxLevelsToPrintPerEdge, detail, optionalTitle, pTreeOptions);
+    if (outputPngName != null) {
+      try (FileInputStream is = new FileInputStream(dotSourceFilePath.toFile())) {
+        MutableGraph gr = Parser.read(is);
+        Graphviz.fromGraph(gr).render(Format.PNG).toFile(new File(outputPngName));
```
Pscheidl: This part takes roughly > 8 seconds (no hotspot kicked in) on my computer. I'd guess it's the default DPI settings. I believe this can be tweaked later.

h2o-genmodel/build.gradle
```diff
@@ -9,10 +9,10 @@ javadoc {
   options.overview = "src/main/java/overview.html" // relative to source root
 
   doLast {
-     copy {
-       from "src/main/java/doc-files"
-       into "$buildDir/docs/javadoc/doc-files"
-     }
+    copy {
+      from "src/main/java/doc-files"
```
Pscheidl: Unnecessary formatting changes. Should be removed before merging.

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -219,7 +263,7 @@ private void run() throws Exception {
 
     public PrintTreeOptions(boolean setdecimalplaces, int nplaces, int fontsize, boolean internal) {
       _setDecimalPlace = setdecimalplaces;
-      _nPlaces = _setDecimalPlace?nplaces:_nPlaces;
+      _nPlaces = _setDecimalPlace ? nplaces : _nPlaces;
```
Pscheidl: Minor: Formatting should normally be done separately. But this makes sense and is ok to me.

h2o-genmodel/src/test/java/hex/genmodel/PrintMojoTest.java
```diff
@@ -0,0 +1,147 @@
...
+        targetFile = Files.createTempFile("", "test_h2o.csv.gv");
+        outputPngFilename = "exampleh2o.png";
+        originalSecurityManager = System.getSecurityManager();
+        System.setSecurityManager(new PreventExitSecurityManager());
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        System.setSecurityManager(originalSecurityManager);
+        Files.deleteIfExists(gbmMojoFile);
```
Pscheidl: Minor: JUnit has TemporaryFolder rule. We could adopt it to our advantage.

```java
    @Rule
    TemporaryFolder temporaryFolder = new TemporaryFolder();
```

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
Pscheidl: Question is: Can we make it faster ? What is the default DPI ?

h2o-genmodel/src/test/java/hex/genmodel/PrintMojoTest.java
```diff
@@ -0,0 +1,147 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.security.Permission;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public class PrintMojoTest {
```
michalkurka: Nice to see tests :) Thank you @koniecsveta

Pscheidl:
Thank you @koniecsveta ! I really like what I see. Well tested, the API design is clear and in-line with H2O standards. Really good contribution.

There are couple of issues that are hard to know beforehand, let's talk about them in code review comments directly.

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
Pscheidl: I've investigated this a little bit and the performance problems seems to be observed with the `GraphvizJdkEngine` only. If the default pipeline is used, another one is chosen (probably a binding to the native library) and the time is significantly reduced ( < 1000 ms). Is there anything preventing us from using the default engine pipeline ?

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
koniecsveta: Will fix

h2o-genmodel/build.gradle
koniecsveta: Will fix

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
koniecsveta: Default DPI is 96:
https://github.com/nidi3/graphviz-java/blob/master/graphviz-java/src/main/java/guru/nidi/graphviz/engine/Graphviz.java#withDefaultDpi

I tried to lower it by `gr.graphAttrs().add("dpi",10);` before calling `Graphviz.fromGraph(gr).render(Format.PNG).toFile(new File(outputPngName));` and also tried to fix the width like `Graphviz.fromGraph(gr).width(200).render(Format.PNG).toFile(new File(outputPngName));` but it doesn't seem to have any impact on speed.

h2o-genmodel/build.gradle
koniecsveta: 1. slf4j can be removed from dependencies. However, there will be these logs written into console on each execution: 

`Zuzanas-MacBook-Pro:h2oexp zuzanaolajcova$  java -cp h2o.jar hex.genmodel.tools.PrintMojo --tree 0 -i test_h2o.csv.zip -o test_h2o.csv.gv -f 20 -d 3 --direct mynewpng.png 
log4j:WARN No appenders could be found for logger (guru.nidi.graphviz.service.CommandLineExecutor).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.` 

But maybe there is a way how to suppress it.

2. slf4j cannot be excluded from graphviz, since it is used by its classes and it would lead to java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory

h2o-genmodel/build.gradle
koniecsveta: Excluding `guru.nidi.com.kitfox:svgSalamander:1.1.2` leads to 

```
java.lang.IllegalStateException: - Rasterizer explicitly set no null or
- neither Batik nor Salamander found on classpath.
	at guru.nidi.graphviz.engine.Renderer.toImage(Renderer.java:114)
	at guru.nidi.graphviz.engine.Renderer.toFile(Renderer.java:70)
	at guru.nidi.graphviz.engine.Renderer.lambda$toFile$3(Renderer.java:56)
	at guru.nidi.graphviz.engine.EngineResult.mapIO(EngineResult.java:63)
	at guru.nidi.graphviz.engine.Renderer.toFile(Renderer.java:54)
	at hex.genmodel.tools.PrintMojo.generateDotFromSource(PrintMojo.java:252)
	at hex.genmodel.tools.PrintMojo.handlePrintingToOutputFile(PrintMojo.java:220)
	at hex.genmodel.tools.PrintMojo.run(PrintMojo.java:209)
	at hex.genmodel.tools.PrintMojo.main(PrintMojo.java:46)
```

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
koniecsveta: There is nothing preventing us. I developed it like this in order to demonstrate how non native graphviz version can be used. Default will check for native (GraphvizCmdLineEngine()) and if graphviz is not installed, it will go for the next engine (GraphvizV8Engine()) and next etc... This seems to be a good approach because if native is installed then it is used and if not, then it is used non native (maybe slower, but without need to install new tool). 

The only “problem” is that there will be logs about it from graphviz-java in console (now there wont be logs about which engine was(not) found to be used, because there will be warnings caused by not including log4j depencency and not setting it up together with graphviz-java).

h2o-genmodel/build.gradle
Pscheidl: There are multiple options we have now. 

## Providing minimal SLF4J implementation
The first option requires making ourselves comfortable with slf4j being present on the classpath. GenModel is aimed to be present on the classpath of other projects directly and we need to make sure any SLF4J implementations and most importantly configurations do not interfere, override or change the behavior for our customers.

It should also be done in the minimal way possible (we keep an eye for `genmodel` module's dependencies to be as slim as possible). This means preventing any transitive dependencies on SLF4J API and only supplying a minimal implementation. LOG4J previously present has around 500k, it's only true alternative Logback is about the same size. Using the `SimpleLogger` provided by SLF4J directly introduces a 15K dependency:

```groovy
  compile ("guru.nidi:graphviz-java:0.9.0"){
    exclude group: 'org.slf4j'
  }
  compile group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.27'
```

Please note any transitive slf4j dependencies are excluded, thus no API will be directly present (the scope of the API could actually be compileOnly or "provided" in Maven terms in the graphviz-java project). It is perfectly fine with the `'slf4j-simple` implementation.

There is only one disadvantage - `SimpleLogger` by default prints into System.err, not stdout. This can be further configured, if required. However, I am almost 100% sure this is OK, as this is mostly used as a command line tool and the output for debugging will still be there somewhere, even if the error output is not shown to the user directly in some cases.

## Defaulting to NO-OP

Looks nasty, I'd not do it. There is still some transitive dependency required (otherwise we'll get `NoClassDefFoundError`), so this really solves nothing for us.

## Fork graphviz-java

Forking the repository and substituting `Log.*` with direct usage if `System.out` would be possible and it wouldn't even be that much work, considering the size of the project. However, the downsides are obvious. No easy updates, which means no bugfixes for free. Plus the cost of custom maintenance.

## Outcome
Personally, I'd choose the way of **providing minimal SLF4J implementation**. WDYT @michalkurka , is it safe to include minimal SLF4J implementation with size of 15k, wihch only contains the bare minimum ? The source code that will get included can be seen here: https://github.com/qos-ch/slf4j/tree/master/slf4j-simple/src/main/java/org/slf4j/simple

This way, the classpath content regarding SLF4J would be controlled by us.

What is your opinion @koniecsveta ?

h2o-genmodel/build.gradle
Pscheidl: I see. Thank you. Let's keep it in then. Still, the size is acceptable.

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
Pscheidl: The log content, if it makes sense, is not a bad thing. It could even be there to our own advantage. It would be very useful during debugging. Let's pretend there is a situation where this functionality is not working for an imaginary customer. Having the knowledge of which backend has been used and which was skipped could be very useful. Therefore, as I think about it, I'm not really against this being logged.

Pscheidl: Let's use the defaults then. By default, it is much, much faster then the JDK backend and it uses JDK as fallback on it's own. Win-win situation.

Pscheidl:
There are two major things:

- Make the logging implementation minimal or come up with a very different solution.
- Use the default backend pipeline.

Once we have these two things resolved, we're good to squash and merge. Again, thank you for your time @koniecsveta . My intentions are to really merge this functionality, as it's of great quality and represents a good addition to H2O.

h2o-genmodel/build.gradle
koniecsveta: I think that if there is logging already implemented in graphviz-java, we should use it (logs can be helpful). I would go for simple logger which would store graphviz-java log-output with some acceptable backup strategy (size/date). The question is what is acceptable in this case? If logging is not acceptable and the tool we want to integrate has it as a prerequisite, then we should maybe try different tool (or create own by fork but that comes with mentioned cons).

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
koniecsveta: Resolved in 3fc46f1d1f10090356a1077c42ee7bcdbd9fcc90

Pscheidl: I'll try to merge tomorrow - there is a possible relocation issue that needs to be checked. Otherwise we're good.

Pscheidl: Thank you.

Pscheidl:
Could you please open the PR for changes from the maintainers ? Or, alternatively, apply these two patches:

## SLF4J package relocation
```patch
Index: h2o-assemblies/genmodel/build.gradle
IDEA additional info:
<+>UTF-8
--- h2o-assemblies/genmodel/build.gradle	(revision 1ad0481b35b31c8de628f75e54f2b498430bcff3)
+++ h2o-assemblies/genmodel/build.gradle	(revision 03e5e410cc0a860499e27bb9a7a1862a0f8a2908)
@@ -34,6 +34,7 @@
   mergeServiceFiles()
   classifier = ''
   relocate 'com.google.gson', 'ai.h2o.com.google.gson' // Should be eventually removed: PUBDEV-5508
+  relocate 'org.slf4j', 'ai.h2o.org.slf4j'
   exclude 'META-INF/*.DSA'
   exclude 'META-INF/*.SF'
   exclude 'synchronize.properties'
```

## Explicit engine release

```patch
Index: h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
IDEA additional info:
<+>UTF-8
--- h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java	(revision 03e5e410cc0a860499e27bb9a7a1862a0f8a2908)
+++ h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java	(revision a50dfb501421732ecdd4f4207980871b7852489a)
@@ -244,6 +244,8 @@
       try (FileInputStream is = new FileInputStream(dotSourceFilePath.toFile())) {
         MutableGraph gr = Parser.read(is);
         Graphviz.fromGraph(gr).render(Format.PNG).toFile(new File(outputPngName));
+      } finally {
+        Graphviz.releaseEngine();
       }
     }
   }
```

After that, we're finally good to merge. Thank you.

koniecsveta:
I see "edits from maintainers" as allowed. In case of problems with this I can apply patches in the evening.

michalkurka:
The size of the final assembled genmodel.jar increased considerably 

Before:
`13M genmodel.jar`

After:
`34M  genmodel.jar`

The bundle native libraries (linux, osx, win) are responsible for the increase. IMHO that makes it infeasible to be included in the genmodel that is meant to be relatively small. We can think about including just linux which will bring down the size to 20MB.

I don't think this can be merged in the current state but as a POC I think it absolutely fulfills its purpose. Thank you @koniecsveta.

Pscheidl:
One way to drop this library is to plot it in a custom way and save it. JavaFX has nice plotting API, so we'd need to do just the conversion. However, I just learned it's no longer part of standard JDK.

Pscheidl:
Tried this project, as it is lightweight and uses basic Java 2D libs to draw the graphs: http://graphstream-project.org/

But it was unable to parse our graph in DOT format. It can't read the blocks enclosed in `{}`.

PR #3799, opened 2019-08-12
PUBDEV-6775: added null pointer check to Python and R clients.

wendycwong:
This PR fixes the problem inhttps://0xdata.atlassian.net/browse/PUBDEV-6775.

Honza:
I referred to variables without checking if they are null first.  Added the null pointer check in R and Python with unit tests.  Thank you, Wendy

### DIFF
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -1205,7 +1205,7 @@ class ModelBase(backwards_compatible(Keyed)):
             print("Numpy not found.  Cannot plot 2D partial plots.")
         ycol = colPairs[1]
         nBins = nbins
-        if not(user_cols==None) and (ycol in user_cols):
+        if user_cols is not None and ycol in user_cols:
             ind = user_cols.index(ycol)
             nBins = user_num_splits[ind]
         nrow = int(len(x)/nBins)
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -1205,7 +1205,7 @@ class ModelBase(backwards_compatible(Keyed)):
             print("Numpy not found.  Cannot plot 2D partial plots.")
         ycol = colPairs[1]
         nBins = nbins
-        if ycol in user_cols:
+        if not(user_cols==None) and (ycol in user_cols):
             ind = user_cols.index(ycol)
             nBins = user_num_splits[ind]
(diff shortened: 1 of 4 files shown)


h2o-py/h2o/model/model_base.py
```diff
@@ -1205,7 +1205,7 @@ def __predFor3D(self, x, y, z, colPairs, nbins, user_cols, user_num_splits):
             print("Numpy not found.  Cannot plot 2D partial plots.")
         ycol = colPairs[1]
         nBins = nbins
-        if ycol in user_cols:
+        if not(user_cols==None) and (ycol in user_cols):
```
honzasterba: I think `if user_cols is not None` looks more Python-y

h2o-py/tests/testdir_misc/pyunit_pubdev_6775_2D_pdp_xgboost.py
```diff
@@ -0,0 +1,34 @@
...
+    gbm_model = H2OGradientBoostingEstimator(ntrees=50, learn_rate=0.05, seed=12345)
+    gbm_model.train(x=x, y=y, training_frame=data)
+
+    # pdp without weight or NA
+    fn = "pdp_simple.png"
+    gbm_model.partial_plot(data=data, server=True, plot=True, 
+        col_pairs_2dpdp=[['AGE', 'PSA'],['AGE', 'RACE']],   
+        save_to_file="pdp_simple.png")
+
+    if os.path.isfile(fn):
```
honzasterba: I generally dont like tests without assert, at least assert true at the end to demonstrate it ran fine

you could however assert that the plot file was written and exists, also, you should use tempfile when writing a file to avoid colisions between tests

h2o-py/h2o/model/model_base.py
honzasterba: and also you dont need the brackets (), python has pretty good operator precedence

wendycwong: Yeap, you caught me.  Not a python coder.  Fixed.  Thanks.

h2o-py/tests/testdir_misc/pyunit_pubdev_6775_2D_pdp_xgboost.py
wendycwong: I am going to delete this whole save file thing and added an assert.

PR #3802, opened 2019-08-12
PUBDEV-6413: fixed variance calculation over cross-validation metrics

wendycwong:
This PR fix the bug in https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6413?filter=myopenissues&orderby=priority%20DESC.

Fix the variance calculation (and hence standard deviation) in MathUtils.BasicStats.  I added a Pyunit test to make sure the calculation is correct.

### DIFF
--- a/h2o-core/src/main/java/hex/ModelBuilder.java
+++ b/h2o-core/src/main/java/hex/ModelBuilder.java
@@ -1689,14 +1689,13 @@ abstract public class ModelBuilder<M extends Model<M,P,O>, P extends Model.Param
       }
       i++;
     }
-
+    MathUtils.SimpleStats simpleStats = new MathUtils.SimpleStats(numMetrics);
     for (i=0;i<N;++i)
-      stats.add(vals[i],1);
+      simpleStats.add(vals[i],1);
     for (i=0;i<numMetrics;++i) {
-      table.set(i, 0, (float)stats.mean()[i]);
-      table.set(i, 1, (float)stats.sigma()[i]);
+      table.set(i, 0, (float)simpleStats.mean()[i]);
+      table.set(i, 1, (float)simpleStats.sigma()[i]);
     }
-
     Log.info(table);
     return table;
   }
--- a/h2o-core/src/main/java/water/util/MathUtils.java
+++ b/h2o-core/src/main/java/water/util/MathUtils.java
@@ -165,6 +165,70 @@ public class MathUtils {
 
     public boolean isSparse(int col) {return _nzCnt[col] < _nobs;}
   }
+
+  public static final class SimpleStats extends Iced {
+    double[] _wsums;
+    double[] _xsumsqs;
+    double[] _xsums;
+    double[] _var;
+    double[] _sd;
+    double[] _mean;
+
+    public SimpleStats(int n) {
+      _xsumsqs = MemoryManager.malloc8d(n);
+      _xsums = MemoryManager.malloc8d(n);
+      _wsums = MemoryManager.malloc8d(n);
+    }
+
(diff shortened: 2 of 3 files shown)


h2o-core/src/main/java/water/util/MathUtils.java
```diff
@@ -139,8 +142,11 @@ public void reduce(BasicStats bs) {
 
     private double[] variance(double[] res) {
       for (int i = 0; i < res.length; ++i) {
-        long nobs = _nobs - _naCnt[i];
-        res[i] = nobs == 0?0:(nobs / (nobs - 1.0)) * _m2[i] / _wsums[i];
+        //long nobs = _nobs - _naCnt[i];
+        //res[i] = nobs == 0?0:(nobs / (nobs - 1.0)) * _m2[i] / _wsums[i];
```
wendycwong: The original calculation is too advanced for me to understand.  I just added my own calculation here.

h2o-core/src/main/java/water/util/MathUtils.java
```diff
@@ -144,13 +147,34 @@ public void reduce(BasicStats bs) {
       }
       return res;
     }
+    
+    public double[] varianceNorm() {
+      return variance(MemoryManager.malloc8d(_mean.length));
+    }
+    // This function calculation variance the normal way
+    private double[] varianceNorm(double[] res) {
```
wendycwong: started my own variance calculation since the original variance calculation was actually used by deeplearning even though the values are incorrect.  I think they are using the for different purpose and should really not be called variance.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1694,7 +1694,7 @@ private TwoDimTable makeCrossValidationSummaryTable(Key[] cvmodels) {
       stats.add(vals[i],1);
     for (i=0;i<numMetrics;++i) {
       table.set(i, 0, (float)stats.mean()[i]);
-      table.set(i, 1, (float)stats.sigma()[i]);
+      table.set(i, 1, (float)stats.sigmaNorm()[i]);
```
michalkurka: @wendycwong could we just use `org.apache.commons.math3.stat.descriptive.moment.StandardDeviation` instead?

We are just using few values here (no big data), imho implemeting this in the framework of BasicStats is not necesasary.

h2o-core/src/main/java/water/util/MathUtils.java
```diff
@@ -66,10 +66,13 @@ static public double computeWeightedQuantile(Vec weight, Vec values, double alph
     private final double[] _mean;
     private final double[] _m2;
     double[] _wsums;
+    double[] _wsumsq; // store sum of squares of original element
```
michalkurka: I would avoid modifying this class if there is another way to do it. Adding anything here will make things just more confusing :)

h2o-core/src/main/java/hex/ModelBuilder.java
wendycwong: The problem here is I like the BasicStats framework.  They take care of NA counts, weighted means.  I just added two fields there, _varianceNorm, _sigmaNorm.  If the functions varianceNorm() or sigmaNorm() are not called, the two new fields will be null.

h2o-core/src/main/java/water/util/MathUtils.java
wendycwong: Done implementing the apache class for std and mean

michalkurka:
Thanks @wendycwong! Looks bullet proof now!

PR #3806, opened 2019-08-13
PUBDEV-6789: Fix GLM plotting

michalkurka:


michalkurka:
If there was just one iteration, don't plot lines - plot the point instead (will happen always for regression).

Screen Shot 2019-08-13 at 4 31 44 PM

michalkurka:
Example output for binomial

Screen Shot 2019-08-13 at 4 31 58 PM

PR #3809, opened 2019-08-14
PUBDEV-6761: Create examples for the Assembly section in the Python documentation

hannah-tillman:


angela0xdata:
The examples look great! Let's also add descriptions and returns for each, though.

angela0xdata:
@hannah-tillman can you add "Returns" sections too?

hannah-tillman:
> @hannah-tillman can you add "Returns" sections too?

Sorry about that--thought you meant the output for each T_T

hannah-tillman:
> The examples look great! Let's also add descriptions and returns for each, though.

Also; don't know what to do for returns for the very first example. Any ideas?

angela0xdata:
The first example won't need a returns because the entire example is already described. But the returns would be "a new H2OFrame"

PR #3810, opened 2019-08-15
GLRM Py test showing suspicious behaviour

seb-h2o:
This behaviour makes GLRM unusable in sklearn context:
let's say you have data as a numpy array `np_data`, then depending on the implementation of `fit_transform` or `fit_predict`, we will obtain different results when running:
```
# np_data will be converted twice to H2OFrame
glrm.fit(np_data)
glrm.predict(np_data)
#or np_data converted only once to H2OFrame
glrm.fit_predict(np_data)
```

### DIFF
--- a/h2o-py/tests/testdir_algos/glrm/pyunit_glrm_idempotence.py
+++ b/h2o-py/tests/testdir_algos/glrm/pyunit_glrm_idempotence.py
@@ -1,0 +1,79 @@
+from __future__ import print_function
+import sys
+sys.path.insert(1,"../../../")
+import h2o
+from tests import pyunit_utils
+from h2o.estimators.glrm import H2OGeneralizedLowRankEstimator
+
+
+def test_glrm_indempotence():
+    # df = h2o.upload_file(pyunit_utils.locate("smalldata/iris/iris_wheader.csv"), destination_frame='iris')
+    df = h2o.import_file(path=pyunit_utils.locate("smalldata/extdata/australia.csv"), destination_frame='autralia')
+
+    estimator = H2OGeneralizedLowRankEstimator(k=2)
+
+    print("input data before train:")
+    print(df)
+
+    estimator.train(training_frame=df)
+    # estimator.show()
+    print("input data after train:")
+    print(df)
+
+    preds1 = estimator.predict(df)
+    print("input data after predict:")
+    print(df)
+    print("predictions on input data:")
+    print(preds1)
+
+    estimator.train(training_frame=df)
+    preds2 = estimator.predict(df)
+    print("2nd predictions on input data:")
+    print(preds2)
+
+    # assertion below should fail
+    assert not (preds1 == preds2).all(), "GLRM is now indempotent, rather a good thing: this is fixed then!"
+
+
+
+def test_glrm_predict_on_clone():
+    # df = h2o.upload_file(pyunit_utils.locate("smalldata/iris/iris_wheader.csv"), destination_frame='iris')
+    df = h2o.import_file(path=pyunit_utils.locate("smalldata/extdata/australia.csv"), destination_frame='autralia')
+
+    estimator = H2OGeneralizedLowRankEstimator(k=2)
+
+    print("input data before train:")
+    print(df)
+
+    estimator.train(training_frame=df)
+    # estimator.show()
+    print("input data after train:")
+    print(df)
+
+    preds1 = estimator.predict(df)
+    print("input data after predict:")
+    print(df)
+    print("predictions on input data:")
+    print(preds1)
+
+    preds2 = estimator.predict(df)
+    print("2nd predictions on input data:")
+    print(preds2)
+
+    assert (preds1 == preds2).all(), "Consecutive predicts don't return same predictions"
+
+    df_clone = h2o.deep_copy(df, df.frame_id+'_clone')
+    print("cloned data:")
+    print(df)
+    preds1_clone = estimator.predict(df_clone)
+    print("predictions on cloned data:")
+    print(preds1_clone)
+
+    # assertion below should fail
+    assert not(preds1 == preds1_clone).all(), "GLRM now does provide similar predictions on cloned data, looks like this is fixed!"
+
+
+pyunit_utils.run_tests([
+    test_glrm_indempotence,
+    test_glrm_predict_on_clone
+])


seb-h2o:
@michalkurka @wendycwong I don't expect this to be a "normal" behaviour:
once the GLRM is trained, `predict` should be independent of the frame used for training.

Besides, when printing the frame, I can't see any changes after training (even when doing a `df.refresh()`, so I'm not even sure that the frame is modified in-place, it rather looks like some other reference to the frame or cached value is stored somewhere.
Any thoughts?

wendycwong:
Seb:

In H2O-3, we only guarantee repeatable results when you set and use the same seed value in your training.  I just changed the way you prepare the model as: 

estimator = H2OGeneralizedLowRankEstimator(k=2, seed=12345)

replace your assert statement with:
pyunit_utils.compare_frames_local(preds1, preds2, prob=1.0, tol=0.0)

and the code ran fine.

h2o-py/tests/testdir_algos/glrm/pyunit_glrm_idempotence.py
```diff
@@ -0,0 +1,79 @@
...
+from tests import pyunit_utils
+from h2o.estimators.glrm import H2OGeneralizedLowRankEstimator
+
+
+def test_glrm_indempotence():
+    # df = h2o.upload_file(pyunit_utils.locate("smalldata/iris/iris_wheader.csv"), destination_frame='iris')
+    df = h2o.import_file(path=pyunit_utils.locate("smalldata/extdata/australia.csv"), destination_frame='autralia')
+
+    estimator = H2OGeneralizedLowRankEstimator(k=2)
+
```
wendycwong: I changed this one to 

estimator = H2OGeneralizedLowRankEstimator(k=2, seed=12345)

h2o-py/tests/testdir_algos/glrm/pyunit_glrm_idempotence.py
```diff
@@ -0,0 +1,79 @@
...
+    print("input data before train:")
+    print(df)
+
+    estimator.train(training_frame=df)
+    # estimator.show()
```
wendycwong: Changed this to
estimator = H2OGeneralizedLowRankEstimator(k=2, seed=12345)

h2o-py/tests/testdir_algos/glrm/pyunit_glrm_idempotence.py
```diff
@@ -0,0 +1,79 @@
...
+    print("predictions on input data:")
+    print(preds1)
+
+    estimator.train(training_frame=df)
+    preds2 = estimator.predict(df)
+    print("2nd predictions on input data:")
+    print(preds2)
+
+    # assertion below should fail
+    assert not (preds1 == preds2).all(), "GLRM is now indempotent, rather a good thing: this is fixed then!"
```
wendycwong: change this to

pyunit_utils.compare_frames_local(preds1, preds2, prob=1.0, tol=0.0)

wendycwong:
For your test_glrm_predict_on_clone(), the first assert statement will be true regardless of you setting the seed since you are calling only predict multiple times.

Say you have a dataset A, GLRM will perform the following for you:

A = A1*Y.

If you call predict with the same dataset A, the algorithm will return A1*Y.

However, if you call another dataset B which contains the same dataset as in A but with different model_id, the algorithm attempts to do the following, 

B = B1*Y.  

In this case, it will fix the matrix Y that you obtained from your training and just try to find B1.  The operations of A = A1*Y and B = B1*Y are different and hence you will get slightly different answers.  The reason we implemented the algorithm this way is because of the following use cases:

1. You have a training dataset and you want to get A = XY during training;
2. During scoring, you will have new datasets B, C, ..... In this case, the algorithm is trying to approximate B, C as B1*Y, C1*Y.  During scoring, the program will generate B1, C1 from Y calculated during training.

Note that X is stored as H2OFrame and Y is just a double array.  If you have a big dataset and have many rows, X can be big and have the same number of rows as your dataset.

You can retrieve your X from training by doing:
>>X_frame = h2o.get_frame(estimator._model_json['output']['representation_name'])

You can then save your X_frame for later use if you want.

h2o-py/tests/testdir_algos/glrm/pyunit_glrm_idempotence.py
```diff
@@ -0,0 +1,79 @@
...
+    print("predictions on input data:")
+    print(preds1)
+
+    preds2 = estimator.predict(df)
+    print("2nd predictions on input data:")
+    print(preds2)
+
+    assert (preds1 == preds2).all(), "Consecutive predicts don't return same predictions"
+
+    df_clone = h2o.deep_copy(df, df.frame_id+'_clone')
```
wendycwong: df_clone is now treated as a new dataset by GLRM and the predict will try to do 
approximate B as B1*Y instead of returning X*Y to you

h2o-py/tests/testdir_algos/glrm/pyunit_glrm_idempotence.py
seb-h2o: with the seed, `assert(preds1 == preds2).all()` works fine too

h2o-py/tests/testdir_algos/glrm/pyunit_glrm_idempotence.py
```diff
@@ -0,0 +1,79 @@
...
+    print("input data after predict:")
+    print(df)
+    print("predictions on input data:")
+    print(preds1)
+
+    preds2 = estimator.predict(df)
+    print("2nd predictions on input data:")
+    print(preds2)
+
+    assert (preds1 == preds2).all(), "Consecutive predicts don't return same predictions"
```
seb-h2o: this one is expected, I just wanted to be sure

h2o-py/tests/testdir_algos/glrm/pyunit_glrm_idempotence.py
```diff
@@ -0,0 +1,79 @@
...
+
+    df_clone = h2o.deep_copy(df, df.frame_id+'_clone')
+    print("cloned data:")
+    print(df)
+    preds1_clone = estimator.predict(df_clone)
+    print("predictions on cloned data:")
+    print(preds1_clone)
+
+    # assertion below should fail
+    assert not(preds1 == preds1_clone).all(), "GLRM now does provide similar predictions on cloned data, looks like this is fixed!"
```
seb-h2o: @wendycwong so here, we're saying that by design, even if we have a seed, it's normal to get different predictions?

PR #3813, opened 2019-08-16
Use Pandas to display H2OTwoDimTables if possible

michalkurka:


wendycwong:
Michalk:  Do you want to add a suggestion to have user install panda for better printing results before you fix it at the root?

h2o-py/h2o/two_dim_table.py
```diff
@@ -92,29 +92,32 @@ def as_data_frame(self):
...
-        if nr > 20:  # create a truncated view of the table, first/last 5 rows
-            trunc_table = []
-            trunc_table += [v for v in table[:5]]
-            trunc_table.append(["---"] * len(table[0]))
-            trunc_table += [v for v in table[(nr - 5):]]
-            table = trunc_table
-
-        H2ODisplay(table, self._col_header, numalign="left", stralign="left")
+
+        (nr, is_pandas, table) = self._as_show_table()
```
seb-h2o: small detail: usually, when returning tuples, idiom is to start from most important data to optional info, so, would probably put the table first

michalkurka: thanks @seb-h2o, will fix

PR #3814, opened 2019-08-16
PUBDEV-6805: Implement a re-try mechanism for requesting the flatfile on Hadoop

michalkurka:


h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
```diff
@@ -528,36 +524,54 @@ else if (type == MapperToDriverMessage.TYPE_EXIT) {
...
-          // This is bad.
-          System.out.println("ERROR: Duplicate node registered (" + entry + "), exiting");
-          System.exit(1);
+        if (_dupChecker.containsKey(entry)) {
+          int prevAttempt = _dupChecker.get(entry);
+          if (prevAttempt == attempt) {
+            // This is bad.
+            fatalError("Duplicate node registered (" + entry + "), exiting");
+          } else if (prevAttempt > attempt) {
+            // Suspicious, we are receiving attempts out-of-order, stick with the .
```
jakubhava: small note: missing end of the sentence? Stick with the .. ?

h2o-mapreduce-generic/src/test/java/water/hadoop/CallbackManagerTest.java
```diff
@@ -0,0 +1,59 @@
...
+    Socket s = new Socket();
+    cm.registerNode("localhost", 54, 1, s);
+    cm.registerNode("localhost", 54, 0, new Socket());
+
+    assertEquals(cm._socks, Collections.singletonList(s));
+    assertEquals(cm._nodes, Collections.singletonList("localhost:54"));
+  }
+
+  @Test
+  public void testRegisterNode_duplicate() {
```
jakubhava: Should we add tests also for this case:

`Inconsistency found: old node entry for a repeated register node attempt doesn't exist` ?

michalkurka: thanks, added

h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
michalkurka: yep! good catch, thank you

PR #3815, opened 2019-08-19
[build] multinode host detection sometimes failed on mac

honzasterba:
- implement more robust approach


michal-raska:
Would it be possible to rewrite this using the `ip` instead of `ifconfig`? Otherwise we'll have to add `net-tools` to the base docker image, because `ifconfig` is not available by default.

honzasterba:
this is only when running directly on darwin where ifconfig is available by default, so it should have no impact on jenkins/docker images

michal-raska:
> this is only when running directly on darwin where ifconfig is available by default, so it should have no impact on jenkins/docker images

I missed that, sorry

PR #3817, opened 2019-08-19
[PUBDEV-6802] Move initFrame and finalizeFrame to ExternalFrame API

jakubhava:
Preparation for https://0xdata.atlassian.net/browse/SW-1530

### DIFF
--- a/h2o-core/src/main/java/water/ExternalFrameHandler.java
+++ b/h2o-core/src/main/java/water/ExternalFrameHandler.java
@@ -3,7 +3,6 @@ package water;
 
 import java.io.IOException;
 import java.nio.channels.ByteChannel;
-import java.nio.channels.SocketChannel;
 
 /**
  * <p>This class is used to coordinate the requests for accessing/obtaining H2O frames from non-H2O environments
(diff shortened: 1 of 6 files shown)


h2o-core/src/main/java/water/ExternalFrameHandler.java
```diff
@@ -55,30 +54,27 @@
...
-     * would lead to unwanted states.
+     * Byte used for signaling that reading from h2o frame has finished.
      */
     static final byte CONFIRM_READING_DONE = 1;
-    static final byte CONFIRM_WRITING_DONE = 2;
 
-    /**
-     * Main task codes
-     */
-    static final byte CREATE_FRAME = 0;
```
jakubhava: Move to `ExternalBackendRequestType`

h2o-core/src/main/java/water/ExternalFrameWriterBackend.java
```diff
@@ -73,16 +84,6 @@ static void handleWriteToChunk(ByteChannel sock, AutoBuffer ab) throws IOExcepti
...
         // close chunks at the end
         ChunkUtils.closeNewChunks(nchnk);
-
-        // Flag informing sender that all work is done and
-        // chunks are ready to be finalized.
-        //
-        // This also needs to be sent because in the sender we have to
-        // wait for all chunks to be written to DKV; otherwise we get race during finalizing and
-        // it happens that we try to finalize frame with chunks not ready yet
-        AutoBuffer outputAb = new AutoBuffer();
```
jakubhava: We no longer need this hacky confirmation as the initialization, writing to chunks and frame finalization happens here. The external code (=sparkling water) just calls the API methods:
```
writer.initFrame()
writer.writeToChunk()
writer.finalizeFrame()
```

h2o-core/src/main/java/water/ExternalFrameWriterClient.java
```diff
@@ -65,17 +68,29 @@ public ExternalFrameWriterClient(ByteChannel channel){
...
     /**
-     * Create chunks on the h2o backend. This method creates chunk in en empty frame.
+     * Create single chunk on the already existing empty frame.
      * @param frameKey name of the frame
      * @param expectedTypes expected types
      * @param chunkId chunk index
      * @param totalNumRows total number of rows which is about to be sent
      */
-    public void createChunks(String frameKey, byte[] expectedTypes, int chunkId, int totalNumRows, int[] maxVecSizes) throws IOException {
+    public void createChunk(String frameKey, byte[] expectedTypes, int chunkId, int totalNumRows, int[] maxVecSizes) throws IOException {
```
jakubhava: Rename to more appropriate name

h2o-core/src/main/java/water/ExternalFrameWriterClient.java
```diff
@@ -150,28 +179,7 @@ public void sendDenseVector(double[] values) throws IOException {
...
     }
-
-    /**
-     * This method ensures the application waits for all bytes to be written before continuing in the control flow.
-     *
-     * It has to be called at the end of writing.
-     * @param timeout timeout in seconds
-     * @throws ExternalFrameConfirmationException
-     */
-    public void waitUntilAllWritten(int timeout) throws ExternalFrameConfirmationException {
```
jakubhava: not required for the same reason as mentioned https://github.com/h2oai/h2o-3/pull/3817/files#r315192609

jakubhava:
just fyi: Need to do additional changes after testing, I would prefer not putting this into current release

h2o-core/src/main/java/water/ExternalFrameWriterBackend.java
jakubhava: At the end we need to confirmation to ensure the request is done on the backend side. Otherwise the client will just continue with work & the back-end might have not finished the current task ( init/ write to chunks/finalize)

jakubhava:
Looks good now, thanks all!

PR #3818, opened 2019-08-19
[PUBDEV-6803] Remove read confirmation timeout as it is not actually needed

jakubhava:


h2o-core/src/main/java/water/ExternalFrameReaderClient.java
```diff
@@ -157,28 +157,7 @@ public Timestamp readTimestamp() {
     public boolean isLastNA() {
         return isLastNA;
     }
-
-    /**
-     * This method ensures the application waits for all bytes to be received before continuing in the
```
jakubhava: See JIRA why this was removed https://0xdata.atlassian.net/browse/PUBDEV-6803

jakubhava:
just fyi: Need to do additional changes after testing, I would prefer not putting this into current release

PR #3819, opened 2019-08-19
PUBDEV-6753: Fix mismatch between sorted grid metric and corresponding metric from the model

deil87:
Before the fix threshold(actually index that maximise metric's value) for computing any metrics was the one that maximise F1 metric. This is the only reason why f1 scores were fine whereas f0point5 and f2 were not.

### DIFF
--- a/h2o-core/src/main/java/hex/ModelMetrics.java
+++ b/h2o-core/src/main/java/hex/ModelMetrics.java
@@ -112,7 +112,6 @@ public class ModelMetrics extends Keyed<ModelMetrics> {
   public double mse() { return _MSE; }
   public double rmse() { return Math.sqrt(_MSE);}
   public ConfusionMatrix cm() { return null; }
-  public ConfusionMatrix cm(AUC2.ThresholdCriterion criterion) { return null; }
   public float[] hr() { return null; }
   public AUC2 auc_obj() { return null; }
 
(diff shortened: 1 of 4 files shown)


h2o-core/src/main/java/hex/ModelMetrics.java
```diff
@@ -135,7 +136,20 @@ public static double getMetricFromModelMetric(ModelMetrics mm, String criterion)
     Method method = null;
     Object obj = null;
     criterion = criterion.toLowerCase();
-    ConfusionMatrix cm = mm.cm();
+    
+    // Constructing confusion matrix based on criterion
+    ConfusionMatrix cm;
+    AUC2.ThresholdCriterion criterionAsEnum = null;
+    try {
```
michalkurka: What is the purpose of the try-catch?

deil87: @michalkurka the reason is that criterion might not be in enum. For example by default `logloss` will be passed into this method and it is not present in AUC2.ThresholdCriterion enum. There is no `exist` method in enums. I can rewrite it with searching through all values  approach. Or what was exactly your concern about?

michalkurka: Catching an exception without handling in any way is an anti-pattern. From your description it doesn't seem like this is necessary, please fix.

deil87: @michalkurka I know that it is anti-pattern. It is just being used all over the method(and also whole code base) at lines 157 and 166. We should at least log something in the catch blocks, otherwise it might be a blackhole. I will fix

PR #3821, opened 2019-08-19
PUBDEV-5504  Another mode to treat missing values: plug values

michalkurka:
Just use means but add another field to represent it, should pass all tests

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLMModel.java
+++ b/h2o-algos/src/main/java/hex/glm/GLMModel.java
@@ -183,7 +183,7 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
   public static class GLMParameters extends Model.Parameters {
 
     public enum MissingValuesHandling {
-      MeanImputation, Skip
+      MeanImputation, PlugValues, Skip
     }
 
     public String algoName() { return "GLM"; }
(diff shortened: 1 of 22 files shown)


h2o-algos/src/test/java/hex/glm/GLMBasicTestRegression.java
```diff
@@ -134,7 +134,7 @@ public void testSparse() {
     parms._objective_epsilon = 0;
     parms._gradient_epsilon = 1e-10;
     parms._max_iterations = 1000;
-    parms._missing_values_handling = DeepLearningModel.DeepLearningParameters.MissingValuesHandling.Skip;
+    parms._missing_values_handling = GLMModel.GLMParameters.MissingValuesHandling.Skip;
     try {
```
wendycwong: OMG:  This is really bad!  Thanks for fixing this one.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -330,8 +337,14 @@ public void validate(GLM glm) {
             H2O.fail();
         }
       }
+      if (_missing_values_handling != null) {
+        if (!(_missing_values_handling instanceof DeepLearningModel.DeepLearningParameters.MissingValuesHandling) &&
+                !(_missing_values_handling instanceof MissingValuesHandling)) {
```
wendycwong: the !(_missing_values_handling instanceof DeepLearningModel.DeepLearningParameters.MissingValuesHandling) part is for backward compatibility, right?

michalkurka: exactly!

wendycwong:
Thank you for doing this and fixing errors along the way.

PR #3822, opened 2019-08-19
PUBDEV-6703: Add links to custom distribution and custom loss function demos

zainhaq-h2o:
Links for the updated custom distribution and custom loss function demos have been added to the docs.

See: https://0xdata.atlassian.net/browse/PUBDEV-6703

### DIFF
--- a/h2o-docs/src/product/data-science/algo-params/custom_distribution_func.rst
+++ b/h2o-docs/src/product/data-science/algo-params/custom_distribution_func.rst
@@ -14,6 +14,7 @@ Use this option to specify a custom distribution function, which can then be use
 **Notes**: 
 
 - This option is only supported in the Python client.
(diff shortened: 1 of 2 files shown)


h2o-docs/src/product/data-science/algo-params/custom_distribution_func.rst
```diff
@@ -14,6 +14,7 @@ Use this option to specify a custom distribution function, which can then be use
 **Notes**: 
 
 - This option is only supported in the Python client.
+- Demos for the custom distribution and custom loss functions are located at https://github.com/h2oai/h2o-3/blob/master/h2o-py/demos/custom_loss_function_demo.ipynb.
```
angela0xdata: Change here and in the GBM chapter to:

"A demo for the custom distribution function is available here: https://github.com/h2oai/h2o-3/blob/master/h2o-py/demos/custom_loss_function_demo.ipynb"

PR #3823, opened 2019-08-20
[xgboost] move docker images for xgb build into this repo

honzasterba:
lets have all docker infrastrucure at one place

### DIFF
--- a/docker/xgb_build/scripts/Setup.dist.patched
+++ b/docker/xgb_build/scripts/Setup.dist.patched
@@ -1,0 +1,377 @@
+# -*- makefile -*-
+# The file Setup is used by the makesetup script to construct the files
+# Makefile and config.c, from Makefile.pre and config.c.in,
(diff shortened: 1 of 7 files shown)


docker/xgb_build/scripts/Setup.dist.patched
```diff
@@ -0,0 +1,377 @@
+# -*- makefile -*-
```
michalkurka: I am not sure, at least this file doesn't seem like it belongs to this repo

michalkurka:
what is the benefit of keeping it all here?

is there going to be an issue with versioning? how do I build an older xgboost image?

honzasterba:
We already have images used during the build process here - the ones used to verify the libraries, the runtime images, I think it best we have all the images for the build in one place, that is here. 

Good point about the old version, the v1 images can be built from the original files in xgboost repo.

PR #3828, opened 2019-08-22
PUBDEV-6790: Fix response column annotation in PSVM

michalkurka:


michalkurka:
@angela0xdata thank you for the quick review

PR #3830, opened 2019-08-23
Experiment with running of github-action executing JUnit Smoke Tests

michalkurka:


h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -355,13 +356,36 @@ public static Frame parse_test_file( String fname, int[] skipped_columns) {
...
+    return Boolean.getBoolean("H2O_JUNIT_ALLOW_NO_SMALLDATA");
+  }
+  
+  private static void downloadTestFileFromS3(String fname) throws IOException {
+    if (fname.startsWith("./"))
+      fname = fname.substring(2);
+    File f = new File(fname);
+    if (! f.exists()) {
+      f.getParentFile().mkdirs();
+      File tmpFile = File.createTempFile(f.getName(), "tmp", f.getParentFile());
```
Pscheidl: Shouldn't we call `deleteOnExit()`, otherwise the file won't be deleted ? https://stackoverflow.com/questions/16691437/when-are-java-temporary-files-deleted

Pscheidl: The better solution would be to manually manage the file in it's scope and delete it afterwards in a finally block. This would ensure deletion even when the JVM exits abnormally :hammer:

michalkurka: Maybe, the goal though is to keep the file - download it to a temporary file and then move it atomically to the target file.

Pscheidl: I see, just wanted to point it out. Thanks. Resolving.

PR #3833, opened 2019-08-24
Initial version of Terraform template to start H2O on EMR

michalkurka:


templates/emr/terraform/modules/emr/main.tf
```diff
@@ -0,0 +1,121 @@
+##
+## Provider Definition
+##
+provider "aws" {
```
michal-raska: I would recommend specifying a provider version here, something like `version = "~> 2.12"`. I would also recommend specifying required terraform version - https://www.terraform.io/docs/configuration-0-11/terraform.html#specifying-a-required-terraform-version

templates/emr/terraform/modules/emr/main.tf
```diff
@@ -0,0 +1,121 @@
...
+data "aws_vpc" "main" {
+  id = "${var.aws_vpc_id}"
+}
+
+data "aws_subnet" "main" {
+  id = "${var.aws_subnet_id}"
+}
+
+resource "aws_key_pair" "key" {
+  public_key = "ssh-rsa ${var.aws_ssh_public_key == "" ? "AAAAB3NzaC1yc2EAAAADAQABAAABAQC0eX0fhy3WTIHF13DuSTHBFjLzKRssFRrW6e2B+/9Oh2Ua/zsEoIeLyX5YtPAqeR22DVJBA+sOvKMQnenAVUa0XG7y6rzEPgugqWNv6NVsFgbgHMfWpRYcuPuOo42T0AQD/9rLViyAzy6lRDid3gpN3PkSBhDLGPEZYs9Lzucawm2FZV92/9u5CxgvRZBAAIrWtgHwGpos3mVuisNxHjH3uEv0B43NzN5hJfBYiEyHhwi2eyjTuDFvVQ8rywcrDZ+aR2BTRX+roR7eVq7isjyOq41qy+pRsRLl8/9ULA6HvDYyozN+jCd5xhFJHTMG1IInapIUcRewtqzsgA9XggyT" : var.aws_ssh_public_key}"
```
michal-raska: Why do we provide a default SSH key here please?

templates/emr/terraform/modules/emr/main.tf
```diff
@@ -0,0 +1,121 @@
...
+#!/bin/bash
+set -x -e
+
+mkdir -p /home/hadoop/h2o
+cd /home/hadoop/h2o
+
+wget http://h2o-release.s3.amazonaws.com/h2o/rel-${var.h2o_codename}/${var.h2o_fix_version}/h2o-${var.h2o_main_version}.${var.h2o_fix_version}-hdp2.6.zip
+unzip -o h2o*.zip 1> /dev/null & wait
+aws s3 cp ${format("s3://%s/realm.properties", aws_s3_bucket.h2o_bucket.bucket)} realm.properties
+keytool -genkey -keyalg RSA -keystore h2o.jks -keypass h2oh2o -storepass h2oh2o -keysize 2048 -dname "CN=Jonh Smith, OU=H2O, O=H2O.ai, L=Mountain View, S=California, C=US"
```
michal-raska: nit: there's a typo in `CN=Jonh`

templates/emr/terraform/modules/emr/security_groups.tf
```diff
@@ -0,0 +1,46 @@
+resource "aws_security_group" "master" {
+  description = "Security group for master node"
+  vpc_id = "${data.aws_vpc.main.id}"
+  revoke_rules_on_delete = true
+  ingress {
```
michal-raska: Would it be possible to restrict the ingress rule, this seems too open.

templates/emr/terraform/modules/emr/variables.tf
```diff
@@ -0,0 +1,32 @@
...
+variable "aws_access_key" {}
+variable "aws_secret_key" {}
+variable "aws_vpc_id" {}
+variable "aws_subnet_id" {}
+
+variable "aws_ssh_public_key" {
+  default = ""
+}
+variable "aws_region" {
+  default = "us-east-1"
```
michal-raska: IMHO we should not set default for AWS_REGION. I think that if the user has the `AWS_REGION` env var set, it's expected to use the region from this env var, but it'll use the default value instead. If we don't provide a default it'll fail and the user will be informed that the region needs to be set as terraform variable.

PR #3835, opened 2019-08-26
[PUBDEV-6822] Expose number of rows on ColV3 rest api endpoint

jakubhava:


jakubhava:
Number of rows is possible to get via FrameV3

PR #3837, opened 2019-08-27
Fix for TargetEncodingTargetColumnTest

Pscheidl:
Test failure reproduced locally on a cluster of `5`.

## Core issues:
- Not waiting for the `export` MRTask to finish properly. Waiting was done via `Thread.sleep()`.
- One frame re-used, causing possible key leaks and exceptions if one test fails (the @After part might actually find an old `Frame` or `null`).
- Temporary file names re-used.

## Other tweaks
- New test frames now have random keys.
- Scopes are now used.
- Removed unused imports.
- Temporary files are now managed by JUnit and properly placed on the filesystem & deleted without introducing tons of our additional code and taking care of the lifecycle. The tmp file deletion did not count with many exceptional states, e.g. failure to create the temporary file. JUnit handles this for us via `TemporaryFolder` class.

### DIFF
--- a/h2o-automl/src/test/java/ai/h2o/automl/targetencoding/TargetEncodingTargetColumnTest.java
+++ b/h2o-automl/src/test/java/ai/h2o/automl/targetencoding/TargetEncodingTargetColumnTest.java
@@ -1,145 +1,111 @@
 package ai.h2o.automl.targetencoding;
 
-import org.junit.After;
 import org.junit.BeforeClass;
-import org.junit.Ignore;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
 import water.*;
 import water.fvec.*;
-import water.rapids.Rapids;
-import water.rapids.Val;
-import water.util.FrameUtils;
-import water.util.TwoDimTable;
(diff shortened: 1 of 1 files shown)


Pscheidl:
@michalkurka This is what you wanted me to investigate yesterday, but @honzasterba reminded me today about the problem.

h2o-automl/src/test/java/ai/h2o/automl/targetencoding/TargetEncodingTargetColumnTest.java
```diff
@@ -1,145 +1,111 @@
...
               .withDataForCol(1, ar(1, 2, 3))
               .build();
-
-      tmpName = UUID.randomUUID().toString();
-      Frame.export(fr, tmpName, fr._key.toString(), true, 1);
-      tmpName2 = UUID.randomUUID().toString();
-      Frame.export(fr2, tmpName2, fr2._key.toString(), true, 1);
-
-      try {
-        Thread.sleep(1000); // TODO why we need to wait? otherwise `parsedFrame` is empty.
```
honzasterba: 👍

PR #3838, opened 2019-08-27
PUBDEV-6824: Small tidying up in AutoML class

deil87:
This PR is an attempt to digest bigger PR https://github.com/h2oai/h2o-3/pull/3183 by smaller pieces.

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -185,12 +185,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
 
   public static AutoML makeAutoML(Key<AutoML> key, Date startTime, AutoMLBuildSpec buildSpec) {
 
-    AutoML autoML = new AutoML(key, startTime, buildSpec);
-
-    if (null == autoML.trainingFrame)
-      throw new H2OIllegalArgumentException("No training data has been specified, either as a path or a key.");
-
-    return autoML;
+    return new AutoML(key, startTime, buildSpec);
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


deil87:
@seb-h2o You have already seen this changes -> https://github.com/h2oai/h2o-3/pull/3183#discussion_r244973730

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -267,34 +262,54 @@ public AutoML(Key<AutoML> key, Date startTime, AutoMLBuildSpec buildSpec) {
...
     }
+  }
 
-      String sort_metric = buildSpec.input_spec.sort_metric == null ? null : buildSpec.input_spec.sort_metric.toLowerCase();
-      leaderboard = Leaderboard.getOrMake(projectName(), eventLog, this.leaderboardFrame, sort_metric);
-    } catch (Exception e) {
-      delete(); //cleanup potentially leaked keys
-      throw e;
+  private void handleCVParameters(AutoMLBuildSpec buildSpec) {
+    if (null != buildSpec.input_spec.fold_column && 5 != buildSpec.build_control.nfolds) {
```
deil87: @seb-h2o Please help me to understand whether we need this block. During merge with current state of the AutoML I have not found validation for this scenario. Should we remove it?

seb-h2o: @deil87 you just added it, right? 
I don't see why it would be necessary: either user provides a `fold_column` and we use it (it takes precedence over `nfolds`, and we print the warning message below saying that nfolds value is ignored) or we use `nfolds`, possibly the default value of 5.
Besides, checking against a non-trivial hardcoded value is always suspicious.

I'd remove it

deil87:
@seb-h2o This block was there before. Here is a commit where it was removed by you e07332307e7abb68ab6566bdace1b043f2040640 . So I just wanted to double check. I will remove it. Thanks!

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: @ledell if we really want to prevent the user from doing this, this is the kind of validation that should be done on client side (Py, R, schemas): setting both is harmless as soon as we know which one takes precedence, and imo, business logic should only validate against parameters that could fail the autoML process.

seb-h2o: besides, even if I would not raise an error for this in H2O models, at least the error is consistent: it will always raise an error if `nfolds` > 0 and `fold_column` is set.
But for AutoML, by doing this validation in business logic, it would raise an error if setting `nfolds=6, fold_column='folds'` but not if setting `nfolds=5, fold_column='folds'` which is totally weird: only clients can handle this correctly.

deil87: @ledell I think that AutoML plays role of a client which is calling base models...that is why we do some extra validation there. If we know intentions... then we can of course pass parameters down in a right way (with warning). Though warnings are difficult to notice. But in case `nfolds=6` and `fold_column='folds'` we should fail I would say as there is no way to infer intention. Anyway, I think discussing could be moved somewhere else as topic is not directly related to this PR

seb-h2o: @ledell well, not on my page I guess: if I had to change something, I would change the way `ModelBuilder` validates those params, not `AutoML`.
There's an implicit rule in software: you don't fail if it doesn't prevent it from working normally or if you can easily and non-ambiguously recover from a situation, we even invented "warnings" for the later.
The question is not about inferring intentions, the code de facto gives precedence to `fold_column` (even in H2O models), we should just inform user with warnings and makes it clear in documentation. Failing for ignored params is a step towards painful API (when you want to change a param, you have to change 2 instead of 1), whereas precedence+warnings is lenient and makes is more flexible.

PR #3839, opened 2019-08-28
Revert "[PUBDEV-6821] Expose leader node via REST API (#3834)"

Pscheidl:
This reverts commit 46695420 merged in https://github.com/h2oai/h2o-3/pull/3834.

It causes failure of local Python tests:

```
  File "../../../../h2o/backend/connection.py", line 848, in __new__
    if schema == "CloudV3": return H2OCluster.from_kvs(keyvals)
  File "../../../../h2o/backend/cluster.py", line 47, in from_kvs
    raise AttributeError("Attribute %s cannot be set on H2OCluster (= %r)" % (k, v))
AttributeError: Attribute leader_idx cannot be set on H2OCluster (= 0)
```


jakubhava:
@Pscheidl anything I could do to help with is? It would be great if we can make it for the next release

PR #3840, opened 2019-08-28
[PUBDEV-6815] Checkpointing for xgboost

honzasterba:


h2o-algos/src/main/java/hex/util/CheckpointUtils.java
```diff
@@ -0,0 +1,78 @@
...
+            if (!Arrays.equals(builder.train().names(), checkpointModel._output._names)) {
+                throw new IllegalArgumentException("The columns of the training data must be the same as for the checkpointed model");
+            }
+            if (!Arrays.deepEquals(builder.train().domains(), checkpointModel._output._domains)) {
+                throw new IllegalArgumentException("Categorical factor levels of the training data must be the same as for the checkpointed model");
+            }
+        } catch (H2OIllegalArgumentException e) {
+            builder.error(e.values.get("argument").toString(), e.values.get("value").toString());
+        }
+        if (builder._parms instanceof Model.GetNTrees && checkpointModel._output instanceof Model.GetNTrees) {
```
michalkurka: what if this condition is not satisfied? should we just continue?

h2o-py/tests/testdir_algos/xgboost/pyunit_checkpoint_xgboost.py
```diff
@@ -0,0 +1,81 @@
...
+import h2o
+from tests import pyunit_utils
+import os
+import random
+from h2o.estimators.xgboost import H2OXGBoostEstimator
+from h2o.exceptions import H2OResponseError
+ 
+def xgboost_checkpoint():
+
+  milsong_train = h2o.upload_file(pyunit_utils.locate("bigdata/laptop/milsongs/milsongs-train.csv.gz"))
```
michalkurka: is it possible to make a test that will make a model of say 50 trees and then train a model with 25, resume from checkpoint and train another 25 and make sure the predictions are the same?

h2o-algos/src/main/java/hex/util/CheckpointUtils.java
michalkurka: ok, thank you - makes sense

PR #3842, opened 2019-08-28
PUBDEV-6366: fixed broken metrics (e.g. no thresholds) on clients

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6366

### DIFF
--- a/h2o-py/tests/testdir_misc/pyunit_make_metrics.py
+++ b/h2o-py/tests/testdir_misc/pyunit_make_metrics.py
@@ -79,6 +79,10 @@ def pyunit_make_metrics():
     print("m2:")
     print(m2)
 
+    assert m0.accuracy()[0][1] + m0.error()[0][1] == 1
+    assert m1.accuracy()[0][1] + m1.error()[0][1] == 1
+    assert m2.accuracy()[0][1] + m2.error()[0][1] == 1
+
     # Testing base metric methods
     # FIXME: check the same failures for other ModelMetrics impl. and then fix'emall or move them out of base class...
     base_metrics_methods_failing_on_H2OBinomialModelMetrics = ['aic', 'mae', 'mean_per_class_error', 'mean_residual_deviance', 'rmsle']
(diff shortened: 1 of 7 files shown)


h2o-r/h2o-package/R/models.R
```diff
@@ -1857,15 +1860,43 @@ h2o.hit_ratio_table <- function(object, train=FALSE, valid=FALSE, xval=FALSE) {
 #' h2o.F1(perf)
 #' }
 #' @export
-h2o.metric <- function(object, thresholds, metric) {
-  if(!is(object, "H2OModelMetrics")) stop(paste0("No ", metric, " for ",class(object)," .Should be a H2OModelMetrics object!"))
-  if(is(object, "H2OBinomialMetrics")){
-    if(!missing(thresholds)) lapply(thresholds, function(t,object,metric) h2o.find_row_by_threshold(object, t)[, metric], object, metric)
-    else {
-     if(missing(metric)) object@metrics$thresholds_and_metric_scores else object@metrics$thresholds_and_metric_scores[, c("threshold", metric)]
+h2o.metric <- function(object, thresholds, metric, transform=NULL) {
```
seb-h2o: @ledell I had to fix and extend this function to support various cases:
- fix the `error` metrics: one was completely broken, the other was returning confusing (and actually incorrect results): this is fixed here by the generic `transform` param.
- support for "threshold maximizing the metric" like in Python: this is added here by allowing a new special value `thresholds='max'`

h2o-py/tests/testdir_misc/pyunit_make_metrics.py
```diff
@@ -79,6 +79,24 @@ def pyunit_make_metrics():
...
+    assert m0.accuracy(thresholds=0.4).value + m0.error(thresholds=0.4).value == 1
+
+    assert isinstance(m0.accuracy(thresholds=[0.4]).value, list)
+    assert len(m0.accuracy(thresholds=[0.4]).value) == 1
+    assert m0.accuracy(thresholds=[0.4]).value[0] == m0.accuracy(thresholds=0.4).value
+
+    assert isinstance(m0.accuracy(thresholds=[0.4, 0.5]).value, list)
+    assert len(m0.accuracy(thresholds=[0.4, 0.5]).value) == 2
+    assert m0.accuracy(thresholds=[0.4, 0.5]).value == [m0.accuracy(thresholds=0.4).value, m0.accuracy(thresholds=0.5).value]
+
```
seb-h2o: @ledell just added the `value` property: see this test for actual contract

PR #3844, opened 2019-08-29
EMR Terraform: Setup EMR on a private subnet expose Flow using load balancer

michalkurka:


michalkurka:
@michal-raska this is still work-in-progress, needs cleaning up.

The main issue is that the first invocation fails:

    The VPC/subnet configuration was invalid: No route to any external sources detected in Route Table for Subnet: subnet-09a8dd1de8c5a9eb6 for VPC: vpc-03bcaeb337bb00ca3

Consecutive invocation works.

This looks like the dependencies between the tasks are not defined correctly. I was not yet able to decipher where.

PR #3846, opened 2019-08-30
PUBDEV-6778 - h2o.mojo_predict_df corrupts/removes the tmp directory in R

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6778

After `h2o.mojo_predict_df` was called, the temp folder was deleted ! This was solved inside by re-creating it once it's called again, but destroying everything inside temp R temp folder.

Now, the files are unlinked as they're created after the `h2o.mojo_predict_df` is finishing, not touching the `tempdir`. Any attempts to create or delete the tempdir were removed from the code. The method now only relies on `tempdir()`, as it should.

### DIFF
--- a/h2o-r/h2o-package/R/predict.R
+++ b/h2o-r/h2o-package/R/predict.R
@@ -180,15 +180,15 @@ h2o.mojo_predict_csv <- function(input_csv_path, mojo_zip_path, output_csv_path=
 #' @return Returns a data.frame containing computed predictions
 #' @export
 h2o.mojo_predict_df <- function(frame, mojo_zip_path, genmodel_jar_path=NULL, classpath=NULL, java_options=NULL, verbose=F, setInvNumNA=F) {
-	input_csv_path <- file.path(tempdir(), 'input.csv')
-	prediction_csv_path <- file.path(tempdir(), 'prediction.csv')
+	input_csv_path <- file.path(tempdir(), paste0('h2o_input_', stringi::stri_rand_strings(1,20), '.csv'))
+	prediction_csv_path <- file.path(tempdir(), paste0('h2o_prediction_', stringi::stri_rand_strings(1,20), '.csv'))
 	tryCatch(
 		{
 			write.csv(frame, file=input_csv_path, row.names=F)
-			return(h2o.mojo_predict_csv(input_csv_path=input_csv_path, mojo_zip_path=mojo_zip_path, output_csv_path=prediction_csv_path, genmodel_jar_path=genmodel_jar_path, classpath=classpath, java_options=java_options, verbose=verbose, setInvNumNA=setInvNumNA))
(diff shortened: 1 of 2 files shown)


h2o-r/h2o-package/R/predict.R
```diff
@@ -180,17 +180,16 @@ h2o.mojo_predict_csv <- function(input_csv_path, mojo_zip_path, output_csv_path=
...
 	tryCatch(
 		{
-			input_csv_path <- file.path(tmp_dir, 'input.csv')
-			prediction_csv_path <- file.path(tmp_dir, 'prediction.csv')
 			write.csv(frame, file=input_csv_path, row.names=F)
 			return(h2o.mojo_predict_csv(input_csv_path=input_csv_path, mojo_zip_path=mojo_zip_path, output_csv_path=prediction_csv_path, genmodel_jar_path=genmodel_jar_path, classpath=classpath, java_options=java_options, verbose=verbose, setInvNumNA=setInvNumNA))
 		},
 		finally={
-			unlink(tmp_dir, recursive=T)
+			unlink(input_csv_path)
```
honzasterba: broken indent

h2o-r/h2o-package/R/predict.R
```diff
@@ -180,17 +180,16 @@ h2o.mojo_predict_csv <- function(input_csv_path, mojo_zip_path, output_csv_path=
 #' @return Returns a data.frame containing computed predictions
 #' @export
 h2o.mojo_predict_df <- function(frame, mojo_zip_path, genmodel_jar_path=NULL, classpath=NULL, java_options=NULL, verbose=F, setInvNumNA=F) {
-	tmp_dir <- tempdir()
-	dir.create(tmp_dir)
+	input_csv_path <- file.path(tempdir(), 'input.csv')
+	prediction_csv_path <- file.path(tempdir(), 'prediction.csv')
```
michalkurka: can we create a sub-directory in the temp directory and remove that one instead? this seems prone to errors

Pscheidl: It's the same thing with the subdirectory. It can be overridden by one, unless having a unique name.

The right solution seems to be generating those filenames with a portion of randomness. Luckily, H2O has transitive dependency on the `stringi` library, which makes it easy to do.

h2o-r/h2o-package/R/predict.R
Pscheidl: Reindented in R studio. Thanks.

h2o-r/h2o-package/R/predict.R
honzasterba: I dont think using a transitive dependency is correct, you need to add it as an direct dependcy when used in the code. however I guess to venture that you dont need an external lib to generate a random number in R

Pscheidl: Letters are used instead of base 10, as the string is then shorter for the same number of possible combinations (there are more letters per place & only 10 numbers). But I guess numbers are fine as well in this case.

runif() is able to quickly produce pseudo-random numbers with uniform distribution once the right arguments are put inside (source: https://renenyffenegger.ch/notes/development/languages/R/functions/runif ). Ran a test myself, the histogram on 10k numbers is nearly uniform, which is good enough for me.

PR #3850, opened 2019-09-02
PUBDEV-6839 - Unify TE arguments with standard H2O models

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6839

From the perspective of the JIRA and it's goals, as proposed by @seb-h2o , this part can be reviewed and addressed in comments right not. But there is a bug that got revealed by this change - the very way scoring is done. There is a map serialized into MOJO which basically points to a very specific index for the given feature in the dataset. Not only this implies the production dataset given to the MOJO must have the features ordered exactly as in the training dataset, it also implies all the features must be there, including fold column and others. 

This unfortunately righteously breaks MOJO tests, as currently, the number of features in TargetEncoderOutput started to work as in other models, showing only the number of features actually used, and not the whole dataset.

I'm currently working on a way to dig it out in a way that is compatible with what has already been released. In case you'd like to investigate the current behavior, see `TargetEncoderMojoModel` class, method `  public double[] score0(double[] row, double[] preds)`, row 51 `double categoricalLevelIndex = row[indexOfColumnInRow]; ` and its surroundings. It's obvious when the behavior of `rel-yau` and this branch is compared.

### DIFF
--- a/h2o-core/src/test/java/water/TestUtil.java
+++ b/h2o-core/src/test/java/water/TestUtil.java
@@ -1131,6 +1131,25 @@ public class TestUtil extends Iced {
     return true;
   }
 
+  public static final String[] ignoredColumns(final Frame frame, final String... usedColumns) {
+    final Set<String> usedColumnsSet = new HashSet<>(usedColumns.length);
+    for (String usedColumn : usedColumns) {
+      usedColumnsSet.add(usedColumn);
+    }
+    
+    final String[] names = frame.names();
+    final String[] ignoredcolumns = new String[names.length - usedColumns.length];
+
+    int ignoredColumnsPointer = 0;
+    for (String name : names) {
+      if (!usedColumnsSet.contains(name)) {
+        ignoredcolumns[ignoredColumnsPointer++] = name;
+      }
+    }
+
+    return ignoredcolumns;
+  }
+
   public static boolean compareFrames(final Frame f1, final Frame f2) throws IllegalStateException {
     return compareFrames(f1, f2, 0);
   }
--- a/h2o-core/src/test/java/water/TestUtil.java
+++ b/h2o-core/src/test/java/water/TestUtil.java
@@ -1132,22 +1132,9 @@ public class TestUtil extends Iced {
   }
 
   public static final String[] ignoredColumns(final Frame frame, final String... usedColumns) {
-    final Set<String> usedColumnsSet = new HashSet<>(usedColumns.length);
-    for (String usedColumn : usedColumns) {
-      usedColumnsSet.add(usedColumn);
-    }
-    
-    final String[] names = frame.names();
-    final String[] ignoredcolumns = new String[names.length - usedColumns.length];
-
-    int ignoredColumnsPointer = 0;
-    for (String name : names) {
-      if (!usedColumnsSet.contains(name)) {
-        ignoredcolumns[ignoredColumnsPointer++] = name;
-      }
-    }
-
-    return ignoredcolumns;
+    Set<String> ignored = new HashSet(Arrays.asList(frame.names()));
+    ignored.removeAll(Arrays.asList(usedColumns));
+    return ignored.toArray(new String[ignored.size()]);
   }
 
   public static boolean compareFrames(final Frame f1, final Frame f2) throws IllegalStateException {
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
@@ -4,7 +4,6 @@ import hex.ModelBuilder;
 import hex.ModelCategory;
 import water.Scope;
 import water.fvec.Frame;
-import water.util.ArrayUtils;
 import water.util.IcedHashMapGeneric;
 import water.util.Log;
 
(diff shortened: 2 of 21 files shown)


h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java
```diff
@@ -33,21 +26,13 @@
     public String[] fields() {
       final List<String> params = extractDeclaredApiParameters(getClass());
       params.add("model_id");
+      params.add("ignored_columns");
```
seb-h2o: I don't think this one needs to be added here: for other algos, it is "write-only".
Adding it as a field mainly causes its exposition as a public property in Python, why would you need this?

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -1131,6 +1131,25 @@ public static boolean compareFrames(final Frame f1, final Frame f2, double delta
...
+
+    int ignoredColumnsPointer = 0;
+    for (String name : names) {
+      if (!usedColumnsSet.contains(name)) {
+        ignoredcolumns[ignoredColumnsPointer++] = name;
+      }
+    }
+
+    return ignoredcolumns;
+  }
```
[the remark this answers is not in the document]
Pscheidl: Agreed. Did exactly that.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
```diff
@@ -64,8 +61,16 @@ public void computeImpl() {
...
 
 
-    private String[] encodedColumnsFromIgnored(final String[] minued, final String[] subtrahend, final String responseName,
-                                                      final String foldColumnName) {
+    private String[] encodedColumnsFromIgnored(final Frame trainingFrame, final String[] subtrahend, final String responseName,
+                                               final String foldColumnName) {
+      // Ignore non-categorical columns by default
+      final ArrayList<String> minued = new ArrayList();
+      for (int i = 0; i < trainingFrame.numCols(); i++) {
+        if (trainingFrame.vec(i).isCategorical() && !trainingFrame.name(i).equals(responseName)) {
```
seb-h2o: duplicate logic with response: it's filtered out in `minued` and then removed in `subtrahend`.
Would be easier to keep `minued` = trainingFrame._names, and then specify all the things you remove.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
```diff
@@ -36,8 +36,9 @@ public void init(boolean expensive) {
   private class TargetEncoderDriver extends Driver {
     @Override
     public void computeImpl() {
-      final String[] encoded_columns = Frame.VecSpecifier.vecNames(_parms._encoded_columns);
-      TargetEncoder tec = new TargetEncoder(encoded_columns);
+      final String[] encodedColumns = encodedColumnsFromIgnored(_parms.train(), _parms._ignored_columns,
```
michalkurka: This doesn't look correct to me. At this point the Frame should only contain columns that will actually be used, it shouldn't skip any additional columns. You should also never work with `_parms.train()` here because it is a different frame that the one that was adapted by the model builder initialization.

There is a specific mechanism for column fitlering already this one should be used.

Pscheidl: It's not correct and I've fixed that. There is no need to do this manually - just use the names from the already adapted training frame in ModelBuilder's init and remove fold column name and response column name.

seb-h2o: @michalkurka `There is a specific mechanism for column filtering already this one should be used.` Do you mean that it should be done in `TargetEncoderBuilder.init()`? Would be the right place imo.
@Pscheidl why did you completely remove the logic? I don't think that the non-categorical columns are excluded anymore.
Wouldn't it be fine if it were just moved to `init`, overriding `_parms. _encoded_columns`?
By the way:
```
  public TargetEncoderBuilder(TargetEncoderModel.TargetEncoderParameters parms) {
    super(parms);
    super.init(false);
  }
```
should be :
```
  public TargetEncoderBuilder(TargetEncoderModel.TargetEncoderParameters parms) {
    super(parms);
    init(false);
  }
```
this could lead to bugs if overriding `init`.

michalkurka: > Do you mean that it should be done in TargetEncoderBuilder.init()?

@seb-h2o yes, it should be done as a part of the init method, specifically, I meant to override `ignoreBadColumns` which is supposed to drop columns that are not relevant to the particular algo (in this case we also need to drop the numeric columns).

seb-h2o: `ignoredBadColumns`: nice, never noticed this method.
Yes, could definitely be overridden in `TargetEncoderBuilder`: @Pscheidl

seb-h2o: @michalkurka the thing is: we still want the numeric columns in the output, so we don't want to remove them from train, we just want to remove them from `encoded_columns` param.
So `init` should be a better place for this.

seb-h2o: actually nowhere in the current code do I see anything close to:
```
encoded_columns = categorical train columns not in ignored_columns
```
not the "categorical train columns" part, 
nor the "not in ignored_columns" part.
So I'm even confused how it even works when used from  clients currently (as they provide only `ignored_columns`)

michalkurka: > @michalkurka the thing is: we still want the numeric columns in the output, so we don't want to remove them from train, we just want to remove them from encoded_columns param.

Hmm, okay... if the goal is to leave the dataset intact then I think we also need to defeat the default login of `ignoreBadColumns`.

Why does TE need to know about columns that are not being used for training? IMHO they should just be dropped.

seb-h2o: @michalkurka sorry, forget my previous post, I'm afraid I got confused again with this part. We don't need those columns for training, we just want them in the predict/transform result, and this should work anyway if they are removed from `train` before training.
So back to your suggestion: `ignoredBadColumns` should be overridden in `TargetEncoderBuilder` to remove non-numeric columns.

PR #3852, opened 2019-09-02
Alternative to network-based clouding on Hadoop

michalkurka:


h2o-mapreduce-generic/src/main/java/water/hadoop/FSCloudingEventSource.java
```diff
@@ -0,0 +1,150 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+import java.io.*;
+import java.util.*;
+import java.util.stream.Collectors;
+
+class FSCloudingEventSource {
```
honzasterba: since we already have a class FileSystemBasedClouding I would prefer calling this FileSystemCloudingEventSource

h2o-mapreduce-generic/src/main/java/water/hadoop/FSCloudingEventSource.java
```diff
@@ -0,0 +1,150 @@
+package water.hadoop;
```
honzasterba: I think this file should be split into separate classes and maybe put into a package

h2o-mapreduce-generic/src/main/java/water/hadoop/FSCloudingEventSource.java
```diff
@@ -0,0 +1,150 @@
...
+      return newEvents;
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to read events", e);
+    }
+  }
+
+}
+
+class CloudingEvent implements Comparable<CloudingEvent> {
+  CloudingEventType _type;
```
honzasterba: it would be a bit cleaner if all those fields were final and we had an init constructor

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
```diff
@@ -0,0 +1,140 @@
...
+    System.out.println("FileSystemBasedClouding print()");
+    System.out.println("    Clouding directory: " + ((_path != null) ? _path : "(null)"));
+    System.out.println("    Target cloud size: " + _n);
+  }
+  
+  private void writeFile(Path path, String content) throws IOException {
+    Path temp = new Path(path.getParent(), path.getName() + ".temp");
+    try (OutputStream os = _fs.create(temp, false);
+         PrintWriter wr = new PrintWriter(os)) {
+      wr.write(content);
```
honzasterba: formating + printwriter can println

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
```diff
@@ -0,0 +1,140 @@
...
+    public boolean accept(Path path) {
+      String name = path.getName();
+      if (name.endsWith(".node")) {
+        return !_nodes.containsKey(name);
+      }
+      return name.endsWith(".exit");
+    }
+  }
+  
+  private static URI h2oUri(InetAddress ip, int port) {
```
honzasterba: always used with toString why not return string?

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
```diff
@@ -0,0 +1,140 @@
...
+    try {
+      writeFile(path, h2oUri(leaderIp, leaderPort).toString());
+    } catch (IOException e) {
+      e.printStackTrace();
+      exit(162);
+    }
+  }
+
+  @Override
+  public void notifyAboutEmbeddedWebServerIpPort(InetAddress ip, int port) {
```
honzasterba: maybe also remove the huge TODO in NetworkInit ;)

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
```diff
@@ -0,0 +1,140 @@
...
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
+
+public class FileSystemBasedClouding extends AbstractClouding {
+
+  private InetAddress _self_ip;
+  private int _self_port = -1;
+  
+  private int _n;
```
honzasterba: I would like a more descriptive name like _cloudSize

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
```diff
@@ -0,0 +1,140 @@
...
+          return null;
+        }
+        try (InputStream is = _fs.open(fs.getPath());
+             BufferedReader br = new BufferedReader(new InputStreamReader(is))) {
+          String node = br.readLine();
+          _nodes.put(name, URI.create(node));
+        }
+      }
+      try {
+        Thread.sleep(1000);
```
honzasterba: the correct call here is wait(), bussy waiting is not good
also InterruptedException should never be eaten, the best way to handle it to do         Thread.currentThread().interrupt();

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
```diff
@@ -0,0 +1,140 @@
...
+  
+  @Override
+  public void exit(int status) {
+    Path exitPath = toNodePath(_self_ip, _self_port, "exit");
+    try {
+      writeFile(exitPath, String.valueOf(status));
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+    invokeExit(status);
```
honzasterba: maybe put this into finally?

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
```diff
@@ -0,0 +1,140 @@
...
+    return sb.toString();
+  }
+  
+  @Override
+  public void exit(int status) {
+    Path exitPath = toNodePath(_self_ip, _self_port, "exit");
+    try {
+      writeFile(exitPath, String.valueOf(status));
+    } catch (Exception e) {
+      e.printStackTrace();
```
honzasterba: eating exceptions is always bad, I would wrap in RuntimeException and rethrow

h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
```diff
@@ -498,25 +491,188 @@ else if (type == MapperToDriverMessage.TYPE_EXIT) {
...
+      conf.set(h2omapper.H2O_CLOUDING_IMPL, FileSystemBasedClouding.class.getName());
+      conf.set(h2omapper.H2O_CLOUDING_DIR_KEY, _event_source.getPath().toString());
+      conf.setInt(h2omapper.H2O_CLOUD_SIZE_KEY, targetCloudSize());
+    }
+
+    private void sleep() {
+      try {
+        Thread.sleep(1000);
+      } catch (InterruptedException e) {
+        maybeLogException(e);
```
honzasterba: same as above, interrupt should be handled and then re-interrupted

h2o-mapreduce-generic/src/main/java/water/hadoop/AbstractClouding.java
```diff
@@ -0,0 +1,19 @@
+package water.hadoop;
+
+import org.apache.hadoop.conf.Configuration;
+
+abstract class AbstractClouding extends water.init.AbstractEmbeddedH2OConfig {
```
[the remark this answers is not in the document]
michalkurka: AbstractEmbeddedH2OConfig is also used by Sparkling Water

h2o-mapreduce-generic/src/main/java/water/hadoop/EmbeddedH2OConfig.java
```diff
@@ -1,19 +1,18 @@
...
+import org.apache.hadoop.conf.Configuration;
 import water.util.Log;
 
 import java.io.IOException;
-import java.io.OutputStream;
 import java.net.InetAddress;
 import java.net.Socket;
 
-class EmbeddedH2OConfig extends water.init.AbstractEmbeddedH2OConfig {
+class EmbeddedH2OConfig extends AbstractClouding {
```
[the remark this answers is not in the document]
michalkurka: sure, makes sense

h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
```diff
@@ -498,25 +491,188 @@ else if (type == MapperToDriverMessage.TYPE_EXIT) {
     }
   }
 
+  abstract class CloudingManager extends Thread {
```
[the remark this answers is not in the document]
michalkurka: it does (for now) because it manipulates fields of the outer class - we can further refactor

h2o-mapreduce-generic/src/main/java/water/hadoop/FSCloudingEventSource.java
michalkurka: good idea

h2o-mapreduce-generic/src/main/java/water/hadoop/FSCloudingEventSource.java
michalkurka: ok

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
michalkurka: that would defeat the purpose, the node would not actually shutdown

this is safe to do in Hadoop context

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
michalkurka: yep

h2o-mapreduce-generic/src/main/java/water/hadoop/FSCloudingEventSource.java
michalkurka: done

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
michalkurka: sure

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
michalkurka: agreed, it is deprecated at this point

h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
michalkurka: fixed

h2o-mapreduce-generic/src/main/java/water/hadoop/FileSystemBasedClouding.java
michalkurka: Agreed about the handling of the interrupt - I will just propagate it to the caller (who will shut down h2o immediately).

Regarding `wait` vs `sleep`, imho it doesn't matter here - wait() could make the whole thing more complex due to spurious wakeups: https://hazelcast.com/blog/spurious-wakeups-are-real-4/

michalkurka:
> Would be great if we could e2e test this with our hadoop smoke tests

Agreed, I will add this in a separate task: https://0xdata.atlassian.net/browse/PUBDEV-6859

PR #3855, opened 2019-09-03
PUBDEV-6841: Limitations update for XGBoost

zainhaq-h2o:
Update to the supported platform table in the documentation for XGBoost - the supported Ubuntu version has been changed.


h2o-docs/src/product/data-science/xgboost.rst
```diff
@@ -244,15 +244,15 @@ The list of limitations include:
...
-    |Linux     | yes             | yes | yes | Ubuntu 14.04, g++ 4.7 |
-    +----------+-----------------+-----+-----+-----------------------+
-    |OS X      | yes             | no  | no  | OS X 10.11            |
-    +----------+-----------------+-----+-----+-----------------------+
-    |Windows   | no              | no  | no  | NA                    |
-    +----------+-----------------+-----+-----+-----------------------+
+    +----------+-----------------+-----+-----+---------------------+
+    | Platform | Minimal XGBoost | OMP | GPU | Compilation OS      |
+    +==========+=================+=====+=====+=====================+
+    |Linux     | yes             | yes | yes | Ubuntu 18.04, g++ 5 |
```
honzasterba: primary compilation OS was always Centos, now it is Centos 7

honzasterba:
maybe we should mention that we test on ubuntu 16 and centos 7 with GCC 5 and those are considered supported?

PR #3856, opened 2019-09-04
Fix PCA Mojo in order to play nicely with MOJO pipeline builder

michalkurka:
Implement a test that shows PCA can be used in MOJO Pipelines

Improvements & fixes in MOJO Import
    
    - added convenience method for loading of MOJO models
    - added an option to skip the check of white-listed algos
    - fixed an issue when the Frame MOJO is leaked (when user loads from path)
    - unsupported MOJOs don't get proper metric, they only get dummy metric object
    - fixed an issue of loading MOJOs that don't have model metadata

Fix PCA Mojo in order to play nicely with MOJO pipeline builder

### DIFF
--- a/h2o-algos/src/main/java/hex/generic/Generic.java
+++ b/h2o-algos/src/main/java/hex/generic/Generic.java
@@ -7,6 +7,7 @@ import water.H2O;
 import water.Key;
 import water.fvec.ByteVec;
 import water.fvec.Frame;
+import water.util.Log;
 
 import java.io.IOException;
 import java.util.*;
@@ -81,17 +82,20 @@ public class Generic extends ModelBuilder<GenericModel, GenericModelParameters, 
                 final MojoReaderBackend readerBackend = MojoReaderBackendFactory.createReaderBackend(mojoBytes.openStream(_job._key), MojoReaderBackendFactory.CachingStrategy.MEMORY);
                 mojoModel = ModelMojoReader.readFrom(readerBackend, true);
                 
-                if(!ALLOWED_MOJO_ALGOS.contains(mojoModel._modelDescriptor.algoName().toLowerCase())){
-                    throw new IllegalArgumentException(String.format("Unsupported MOJO model '%s'. ", mojoModel._modelDescriptor.algoName()));
+                if(! ALLOWED_MOJO_ALGOS.contains(mojoModel._modelDescriptor.algoName().toLowerCase())) {
+                    if (_parms._disable_algo_check)
+                        Log.warn(String.format("MOJO model '%s' is not supported but user disabled white-list check. Trying to load anyway.", mojoModel._modelDescriptor.algoName()));
+                    else
+                        throw new IllegalArgumentException(String.format("Unsupported MOJO model '%s'. ", mojoModel._modelDescriptor.algoName()));
                 }
 
                 final GenericModelOutput genericModelOutput = new GenericModelOutput(mojoModel._modelDescriptor, mojoModel._modelAttributes);
-                final GenericModel genericModel = new GenericModel(_result, _parms, genericModelOutput, mojoModel, mojoBytes);
+                final GenericModel genericModel = new GenericModel(_result, _parms, genericModelOutput, mojoModel, dataKey);
 
                 genericModel.write_lock(_job);
                 genericModel.unlock(_job);
             } catch (IOException e) {
-                throw new IllegalStateException("Unreachable MOJO file: " + mojoBytes._key, e);
+                throw new IllegalStateException("Unreachable MOJO file: " + dataKey, e);
             }
         }
     }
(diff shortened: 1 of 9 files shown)


michalkurka:
@wendycwong the fix in PCA MOJO is minimal

@Pscheidl I can split into 2 PRs but I wanted to have the option to easily test PCA pipeline in H2O

h2o-algos/src/main/java/hex/generic/Generic.java
```diff
@@ -81,17 +82,20 @@ public void computeImpl() {
                 final MojoReaderBackend readerBackend = MojoReaderBackendFactory.createReaderBackend(mojoBytes.openStream(_job._key), MojoReaderBackendFactory.CachingStrategy.MEMORY);
                 mojoModel = ModelMojoReader.readFrom(readerBackend, true);
                 
-                if(!ALLOWED_MOJO_ALGOS.contains(mojoModel._modelDescriptor.algoName().toLowerCase())){
-                    throw new IllegalArgumentException(String.format("Unsupported MOJO model '%s'. ", mojoModel._modelDescriptor.algoName()));
+                if(! ALLOWED_MOJO_ALGOS.contains(mojoModel._modelDescriptor.algoName().toLowerCase())) {
```
Pscheidl: Scoring is going to work for all MOJOs. I have a different strategy on my mind - just allowing all the algos and only warning the user that model metrics are not available. Not necessarily to be solved in this PR, just a thought to work with :)

Pscheidl: As you're using "Dummy metrics builder", we could simply make scoring work for every MOJO and only display the warning, as you wrote it.

h2o-algos/src/main/java/hex/generic/GenericModel.java
```diff
@@ -52,34 +55,79 @@ private static MojoModel reconstructMojo(ByteVec mojoBytes) {
...
+            case AnomalyDetection:
+                return new ModelMetricsAnomaly.MetricBuilderAnomaly("");
+            default:
+                throw H2O.unimpl();
+        }
+    }
+    
+    private ModelMetrics.MetricBuilder unsupportedMetricsBuilder() {
+        if (_parms._disable_algo_check) {
+            Log.warn("Using dummny metric builder for unsupported model category: " + _output._modelCategory);
```
Pscheidl: Model category 'XY' is currently unsupported. Model metrics will not be available.

h2o-algos/src/main/java/hex/generic/GenericModel.java
```diff
@@ -52,34 +55,79 @@ private static MojoModel reconstructMojo(ByteVec mojoBytes) {
...
+        }
+
+        @Override
+        public double[] perRow(double[] ds, float[] yact, Model m) {
+            return ds;
+        }
+
+        @Override
+        public ModelMetrics makeModelMetrics(Model m, Frame f, Frame adaptedFrame, Frame preds) {
+            return null;
```
Pscheidl: Why don't we return an instance of Generic metrics ?

h2o-algos/src/test/java/hex/pca/PCATest.java
```diff
@@ -442,4 +453,59 @@ public void testPCAPredMojoPojoNumericsOnly() {
...
+      new MojoPipelineBuilder()
+              .addModel("pca", new File(pcaMojoUri))
+              .addMapping("PC1", "pca", 0)
+              .addMapping("PC2", "pca", 1)
+              .addMapping("PC3", "pca", 2)
+              .addMapping("PC4", "pca", 3)
+              .addMainModel("kmeans", new File(kmeansMojoUri))
+              .buildPipeline(pipelineFile);
+
+      GenericModel pipelineModel = Generic.importMojoModel(pipelineFile.getAbsolutePath(), true);
```
Pscheidl: Test the dummy metrics are in place ?

h2o-algos/src/main/java/hex/generic/Generic.java
michalkurka: Interesting, I think that is a good idea.

h2o-algos/src/main/java/hex/generic/GenericModel.java
michalkurka: Done, thanks

h2o-algos/src/test/java/hex/pca/PCATest.java
michalkurka: We do not actually calculate any metrics. The metric builder is only needed for BigScore to work.

michalkurka:
@Pscheidl thanks for the comments and the idea of allowing all mojos except for the metrics, I think that is a good idea and we should look into it

PR #3857, opened 2019-09-04
PUBDEV-6847 - Show MOJO Import possibility when importing model from a different version

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6847

Should we also include the documentation link ? http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/mojo_import.html

### DIFF
--- a/h2o-core/src/main/java/water/AutoBuffer.java
+++ b/h2o-core/src/main/java/water/AutoBuffer.java
@@ -277,8 +277,7 @@ public final class AutoBuffer {
   private void checkVersion(String version) {
     final boolean ignoreVersion = Boolean.getBoolean(H2O_SYSTEM_SERIALIZATION_IGNORE_VERSION);
     if (! version.equals(H2O.ABV.projectVersion())) {
-      String msg = "Found version "+version+", but running version "+H2O.ABV.projectVersion()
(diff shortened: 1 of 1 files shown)


h2o-core/src/main/java/water/AutoBuffer.java
```diff
@@ -277,7 +277,8 @@ public AutoBuffer( InputStream is ) {
   private void checkVersion(String version) {
     final boolean ignoreVersion = Boolean.getBoolean(H2O_SYSTEM_SERIALIZATION_IGNORE_VERSION);
     if (! version.equals(H2O.ABV.projectVersion())) {
-      String msg = "Found version "+version+", but running version "+H2O.ABV.projectVersion();
+      String msg = "Found version "+version+", but running version "+H2O.ABV.projectVersion()
+              + ". If you wish to import older model into a newer version of H2O, please refer to the MOJO import functionality.";
```
michalkurka: Can we turn this into an H2O technote, eg.https://0xdata.atlassian.net/browse/TN-1? (see H2O#technote)

Pscheidl: Added https://0xdata.atlassian.net/browse/TN-14

PR #3858, opened 2019-09-04
PUBDEV-6848 - Target Encoder should ignore non-categorical encoded columns

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6848

Target encoder won't ignore numeric columns. This PR ignores them if they're specified and supplies a standard warning message for the client to display it (tested manually, works).

The simplest use case for this is not to specify x in the client at all - nothing gets ignored then, as the default behavior is to use all the columns.

### DIFF
--- a/h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_target_encoding_model.py
+++ b/h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_target_encoding_model.py
@@ -52,7 +52,15 @@ def test_target_encoding_fit_method():
 
     # Argument check
     te.train(training_frame=trainingFrame, fold_column="pclass", y=targetColumnName, x=teColumns)
-    
+
+    # Drop all non-categorical columns
+    te.train(x=None, y=targetColumnName, training_frame=trainingFrame, fold_column="pclass")
+    transformed = te.transform(trainingFrame, data_leakage_handling="kfold", seed=1234)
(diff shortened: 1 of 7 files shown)


h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_target_encoding_model.py
```diff
@@ -52,7 +52,15 @@ def test_target_encoding_fit_method():
...
     # Argument check
     te.train(training_frame=trainingFrame, fold_column="pclass", y=targetColumnName, x=teColumns)
-    
+
+    # Drop all non-categorical columns
+    te.train(x=None, y=targetColumnName, training_frame=trainingFrame, fold_column="pclass")
+    transformed = te.transform(trainingFrame, data_leakage_handling="kfold", seed=1234)
+    assert transformed.col_names == ['home.dest', 'pclass', 'embarked', 'cabin', 'sex', 'survived', 'name', 'age',
+                                     'sibsp', 'parch', 'ticket', 'fare', 'boat', 'body', 'kfold_column',
+                                     'sex_te', 'cabin_te', 'embarked_te', 'home.dest_te'] # 4 encoded columns
```
seb-h2o: great! now it's clear

Pscheidl:
The issue was found and suggested by @seb-h2o , I only coded it.

PR #3863, opened 2019-09-06
[rel-yau] Improvements & fixes in MOJO Import

michalkurka:
- added convenience method for loading of MOJO models
- added an option to skip the check of white-listed algos
- fixed an issue when the Frame MOJO is leaked (when user loads from path)
- unsupported MOJOs don't get proper metric, they only get dummy metric object
- fixed an issue of loading MOJOs that don't have model metadata


Pscheidl:
All tests are green except for one SVM. Merging.

PR #3867, opened 2019-09-09
PUBDEV-6840: AutoML: offer the possibility to specify the order in which training steps are executed

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6840

# To Reviewers:
The ticket above was assigned for `3.26.0.5`, although given the large amount of refactoring necessary to implement this in a declarative way, I decided to work on `master`.
There shouldn't be any backwards incompatibility at API level, but it's not 100% guaranteed on Java side, although it should be fine for users using only public fields and methods, including SW, which should be tested against this branch.

you may want to start with `pyunit_automl_model_selection.py` for usage examples.

A minimal Python example:
```
    aml = H2OAutoML(project_name="training_plan_simple_syntax",
                    training_plan=[
                        ('DRF', ['XRT', 'def_1']),
                        ('GBM', 'grids'),
                        ('StackedEnsemble', ['best'])
                    ])
    aml.train(y=target, training_frame=train)
```
will build only the following models in this order: XRT, DRF (default model 1), GBM (all default grids), StackedEnsemble (only best of family)

# Changes:
## Refactoring
There's a lot of refactoring there to make the model/grid building and executions more dynamic and flexible, but very few changes in terms of logic.

To make this feature possible, I had to do major refactoring in AutoML in order to remove the pure sequential way of how models used to be built.
There is now a strict separation between:
- definition of the sequence of steps to be executed.
- job creation (configured model builders).
- job execution.
The logic deciding if a model is executed or not based on timeout, max models... is also separated from job creation/execution, and remains the responsibility of the AutoML instance itself.

## API
The new API changes concerns mainly the introduction of a `training_plan` input param and a `trained_steps` output param using similar format, and can then be re-injected as input for strong reproducibility requirements.
The API changes are exposed on REST, Python and R. Given the complexity of the object.

## Business logic
In terms of logical changes, the sequence of models is therefore now defined in a declarative way (separation between sequence definition and its execution).

This dynamic flavour opens the door to various future improvements:
- Different default order for different kind of problems
- AutoML lifecycle: start/pause/restart/stop
- Dynamic Generation of training steps: for example for hyperparameter optimization.

## Customization/Expert
As a bonus feature, the training steps are now registered as a service, so users can extend AutoML internals by plugging in their own steps (backend only).

The new model definitions should just be added to the classpath and registered to the service loader: for example, a customer could add it's own model/grid definitions, or he could reuse those from a previous H2O version.


seb-h2o:
@deil87 this is only refactoring for now, but I'd still be glad if you could provide some early feedback, thanks.

deil87:
@seb-h2o Sure, I will try to find time today for a review

seb-h2o:
@ledell, @deil87 Please review.
Given the large amount of refactoring, it would be much appreciated if the review doesn't last too long, as any new change to `AutoML` during that period will lead to git merge nightmares.
Thanks.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -46,97 +32,23 @@
  */
 public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
 
-  static class WorkAllocations extends Iced<WorkAllocations> {
-
```
seb-h2o: class extracted to dedicated file

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -193,12 +101,6 @@ public static AutoML makeAutoML(Key<AutoML> key, Date startTime, AutoMLBuildSpec
     return AutoMLV99.AutoMLKeyV3.class;
   }
 
-  enum JobType {
-    Unknown,
-    ModelBuild,
-    HyperparamSearch
-  }
-
```
seb-h2o: moved with `WorkAllocations`

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -215,34 +117,34 @@ public AutoMLBuildSpec getBuildSpec() {
...
 
-  private Date startTime;
-  private static Date lastStartTime; // protect against two runs with the same second in the timestamp; be careful about races
-  private Countdown runCountdown;
-  private Job job;                  // the Job object for the build of this AutoML.
+  Date startTime;
+  Countdown runCountdown;
+  Job<AutoML> job;                  // the Job object for the build of this AutoML.
+  WorkAllocations workAllocations;
+  StepDefinition[] trainedSteps;
```
seb-h2o: made those properties package private: `AutoML` instance is injected to `TrainingSteps` and `TrainingStep` so that each model definition can access as much information as possible: by being package private, they are available to the classes already mentioned, and subclasses can access them through delegate accessors.

I still don't want to make those properties public to avoid backwards compatibility issues with future changes.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -573,104 +458,28 @@ private void handleDatafileParameters(AutoMLBuildSpec buildSpec) {
...
     }
-    DKV.put(this);
   }
 
 
-  //*****************  Jobs Build / Configure / Run / Poll section (model agnostic, kind of...) *****************//
-
+  //*****************  Training Jobs  *****************//
 
-  private void pollAndUpdateProgress(Stage stage, String name, WorkAllocations.Work work, Job parentJob, Job subJob) {
```
seb-h2o: polling logic moved to `TrainingStepsExecutor`

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -684,671 +493,39 @@ private int nextInstanceCounter(String algoName, NonBlockingHashMap<String, Inte
...
-  private Key<Model> modelKey(String algoName, boolean with_counter) {
+  Key<Model> modelKey(String algoName, boolean with_counter) {
     String counterStr = with_counter ? "_" + nextInstanceCounter(algoName, algoInstanceCounters) : "";
     return Key.make(algoName + counterStr + "_AutoML_" + timestampFormatForKeys.format(this.startTime));
   }
 
-  Job<Model> trainModel(Key<Model> key, WorkAllocations.Work work, Model.Parameters parms) {
-    return trainModel(key, work, parms, false);
-  }
-
```
seb-h2o: common `trainModel` logic moved with barely no changes to `TrainingStep.ModelStep` inner class.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -684,671 +493,39 @@ private int nextInstanceCounter(String algoName, NonBlockingHashMap<String, Inte
...
 
-  /**
-   * Do a random hyperparameter search.  Caller must eventually do a <i>get()</i>
-   * on the returned Job to ensure that it's complete.
-   * @param gridKey optional grid key
-   * @param work  the allocated work item: used to check various executions limits and to distribute remaining work (time + models)
-   * @param baseParms ModelBuilder parameter values that are common across all models in the search
-   * @param searchParms hyperparameter search space
-   * @return the started hyperparameter search job
-   */
```
seb-h2o: common `hyperparameterSearch` logic moved with barely no changes to `TrainingStep.GridStep` inner class.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -684,671 +493,39 @@ private int nextInstanceCounter(String algoName, NonBlockingHashMap<String, Inte
...
-    Log.debug("Hyperparameter search: "+algo.name()+", time remaining (ms): "+timeRemainingMs());
-    return GridSearch.startGridSearch(
-        gridKey,
-        baseParms,
-        searchParms,
-        new GridSearch.SimpleParametersBuilderFactory<>(),
-        searchCriteria
-    );
-  }
-
```
seb-h2o: stack logic moved with barely no changes to `StackedEnsembleSteps`

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -684,671 +493,39 @@ private int nextInstanceCounter(String algoName, NonBlockingHashMap<String, Inte
...
-    stackedEnsembleParameters._metalearner_nfolds = buildSpec.build_control.nfolds;
-
-    stackedEnsembleParameters.initMetalearnerParams();
-    stackedEnsembleParameters._metalearner_parameters._keep_cross_validation_models = buildSpec.build_control.keep_cross_validation_models;
-    stackedEnsembleParameters._metalearner_parameters._keep_cross_validation_predictions = buildSpec.build_control.keep_cross_validation_predictions;
-
-    Key modelKey = modelKey(modelName, false);
-    return trainModel(modelKey, work, stackedEnsembleParameters, true);
-  }
-
```
seb-h2o: `setCommonModelBuilderParams` and `setStoppingCriteria` logic moved with barely no changes to abstract class `TrainingStep`

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -1381,44 +558,6 @@ protected Futures remove_impl(Futures fs, boolean cascade) {
...
-    modelCount.addAndGet(after - before);
-  }
-
-  private void addModel(final Model newModel) {
-    int before = leaderboard().getModelCount();
-    leaderboard().addModel(newModel);
-    int after = leaderboard().getModelCount();
-    modelCount.addAndGet(after - before);
-  }
-
```
seb-h2o: `addModel` methods moved to `TrainingStepsExecutor` will the polling logic

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -1381,44 +558,6 @@ protected Futures remove_impl(Futures fs, boolean cascade) {
...
-        String[] attempts = { name, name.toUpperCase(), name.toLowerCase() };
-        for (String attempt : attempts) {
-          try {
-            return StoppingMetric.valueOf(attempt);
-          } catch (IllegalArgumentException ignored) { }
-        }
-        return StoppingMetric.AUTO;
-    }
-  }
-
```
seb-h2o: metric utility methods moved to `TrainingStep`

h2o-automl/src/main/java/ai/h2o/automl/Leaderboard.java
```diff
@@ -27,7 +27,7 @@
  * <p>
  * TODO: make this robust against removal of models from the DKV.
  */
-public class Leaderboard extends Keyed<Leaderboard> {
+public class Leaderboard extends Lockable<Leaderboard> {
   /**
```
seb-h2o: changes in Leaderboard have been back-ported to https://github.com/h2oai/h2o-3/pull/3898
As the refactoring raised this issue much more consistently.

h2o-automl/src/main/java/ai/h2o/automl/TrainingStep.java
```diff
@@ -0,0 +1,339 @@
...
+
+    protected TrainingStep(Algo algo, String id, int weight, AutoML autoML) {
+        _algo = algo;
+        _id = id;
+        _weight = weight;
+        _aml = autoML;
+        _description = algo.name()+" "+id;
+    }
+
+    protected abstract Work getWork();
```
seb-h2o: I should rename this to `getAllocatedWork` to avoid confusion with `makeWork` as they're not used during the same phase.

h2o-automl/src/main/resources/META-INF/services/ai.h2o.automl.TrainingStepsProvider
```diff
@@ -0,0 +1,6 @@
+ai.h2o.automl.training.DeepLearningSteps$Provider
+ai.h2o.automl.training.DRFSteps$Provider
+ai.h2o.automl.training.GBMSteps$Provider
+ai.h2o.automl.training.GLMSteps$Provider
+ai.h2o.automl.training.StackedEnsembleSteps$Provider
+ai.h2o.automl.training.XGBoostSteps$Provider
```
seb-h2o: looks like a new line would be better to end this file

h2o-py/tests/testdir_algos/automl/pyunit_automl_model_selection.py
```diff
@@ -0,0 +1,259 @@
...
+    df = h2o.import_file(path=pu.locate("smalldata/prostate/{}".format("prostate_complete.csv.zip" if larger else "prostate.csv")))
+    target = "CAPSULE"
+    df[target] = df[target].asfactor()
+    #Split frames
+    fr = df.split_frame(ratios=[.8,.1], seed=seed)
+    #Set up train, validation, and test sets
+    return pu.ns(train=fr[0], valid=fr[1], test=fr[2], target=target, target_idx=1)
+
+
+def get_partitioned_model_names(leaderboard):
```
deil87: @seb-h2o It's not super relevant to current scope but just an observation that we have this method in many py scripts. Maybe we can move some most generic version to utils or somewhere?

h2o-py/tests/testdir_algos/automl/pyunit_automl_model_selection.py
```diff
@@ -0,0 +1,259 @@
...
+    except h2o.exceptions.H2OTypeError:
+        pass
+
+
+def test_training_plan_using_full_syntax():
+    ds = import_dataset()
+    aml = H2OAutoML(project_name="py_training_plan_full_syntax",
+                    max_models=2,
+                    training_plan=[
+                        dict(name='GLM', steps=[dict(id='def_1')]),
```
deil87: @seb-h2o How would users know how many default models we have inside? only 'training_step' is a source of the knowledge probably.

h2o-py/tests/testdir_algos/automl/pyunit_automl_model_selection.py
```diff
@@ -0,0 +1,259 @@
...
+        ])
+    except AssertionError as e:
+        assert "each definition must have only 1 or 2 keys" in str(e)
+
+    try:
+        H2OAutoML(training_plan=[
+            dict(name="GBM", alias='all_steps')
+        ])
+    except AssertionError as e:
+        assert "alias must be one of ['all', 'defaults', 'grids']" in str(e)
```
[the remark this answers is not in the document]
seb-h2o: yes, or rather the absence of keys in simple syntax is interpreted as 'all' on backend: cf. translation logic in `TrainingStepsRegistry` the component that turn definitions into a list of `TrainingStep`.

h2o-py/tests/testdir_algos/automl/pyunit_automl_model_selection.py
seb-h2o: he can know by running it fully once and looking at `trained_steps`: we can also document this, but I would not suggest to document internals.
That's the main reason for aliases, to care less about internals.

h2o-py/tests/testdir_algos/automl/pyunit_automl_model_selection.py
seb-h2o: yes I know, I just copy/paste it, sometimes there are small differences specific for the test, sometimes they're the exact same.
I'm less concerned about abstraction in tests to be honest.

h2o-core/src/main/java/water/util/PojoUtils.java
```diff
@@ -640,27 +652,39 @@ private static Object fillFromMap(Object o, Map<String, Object> setFields) {
...
+              }
+              f.set(o, valuesDest);
+            } catch (IllegalAccessException e) {
+              throw new IllegalArgumentException("Cannot get value of the field: '" + key + "' on object " + o);
+            } catch (InstantiationException e) {
+              throw new IllegalArgumentException("Field = " + key + " element cannot be set to value = " + value, e);
+            }
+        } else {
+          setField(o, key, value);
+        }
```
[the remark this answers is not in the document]
deil87: @seb-h2o I see, but maybe we should separate these core changes and ask MichalK to review it?

h2o-py/tests/testdir_algos/automl/pyunit_automl_model_selection.py
deil87: @seb-h2o ok, just thought it might be more convenient

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -149,20 +61,17 @@ public static AutoML startAutoML(AutoMLBuildSpec buildSpec) {
     synchronized (AutoML.class) {
       // protect against two runs whose startTime is the same second
       if (lastStartTime != null) {
-        while (lastStartTime.getYear() == startTime.getYear() &&
-            lastStartTime.getMonth() == startTime.getMonth() &&
-            lastStartTime.getDate() == startTime.getDate() &&
-            lastStartTime.getHours() == startTime.getHours() &&
-            lastStartTime.getMinutes() == startTime.getMinutes() &&
-            lastStartTime.getSeconds() == startTime.getSeconds())
+        while (Math.abs(lastStartTime.getTime() - startTime.getTime()) < 1e3 ) {
```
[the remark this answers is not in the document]
seb-h2o: deprecating some of those `public static` is a good idea indeed, as it's basically the only thing we can do with those `public` methods: I know that I deprecated `public static AutoML makeAutoML` in a separate PR.
Will check if it's reasonable to deprecate one of those `startAutoML` ones.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -149,20 +61,17 @@ public static AutoML startAutoML(AutoMLBuildSpec buildSpec) {
...
+          try {
+            Thread.sleep(500);
```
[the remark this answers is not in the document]
seb-h2o: I've also been thinking about tricking the Date instance indeed.
Honestly, I don't think anyone creates AutoML instances in a loop anyway, as it would hurt performance: H2O-3 doesn't scale in those situations as model building is trying to use all the nodes/threads (even if it doesn't always).
Also, I know that the loop could check more often, but given that the AutoML run will last several minutes (hours?), who can't wait a second more?
Still, maybe the trick with the Date would be nicer than this.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -215,34 +117,34 @@ public AutoMLBuildSpec getBuildSpec() {
...
-  private Frame trainingFrame;    // required training frame: can add and remove Vecs, but not mutate Vec data in place
-  private Frame validationFrame;  // optional validation frame; the training_frame is split automagically if it's not specified
-  private Frame blendingFrame;
-  private Frame leaderboardFrame; // optional test frame used for leaderboard scoring; if not specified, leaderboard will use xval metrics
+  public StepDefinition[] getTrainedSteps() { return trainedSteps; }
 
-  private Vec responseColumn;
-  private Vec foldColumn;
-  private Vec weightsColumn;
+  Frame trainingFrame;    // required training frame: can add and remove Vecs, but not mutate Vec data in place
```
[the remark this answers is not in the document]
seb-h2o: @deil87 yes, I use the _ convention in every new class.
Maybe I should use this refactoring as an opportunity to apply it to `AutoML` as well, as soon as there's no `public` attributes. Good point.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -309,8 +213,15 @@ private void handleCVParameters(AutoMLBuildSpec buildSpec) {
     }
   }
 
+  TrainingStep[] getExecutionPlan() {
+    if (executionPlan == null) {
+      executionPlan = trainingStepsRegistry.getOrderedSteps(this);
+    }
+    return executionPlan;
```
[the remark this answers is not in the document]
seb-h2o: @deil87 no, no, I like the ?: operator and use it as often as I can: maybe the if condition changed over time and didn't notice it was just an expression at the end. Will change.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -272,6 +171,11 @@ public AutoML(Key<AutoML> key, Date startTime, AutoMLBuildSpec buildSpec) {
       handleEarlyStoppingParameters(buildSpec);
 
       initLeaderboard(buildSpec);
+
+      StepDefinition[] trainingPlan = buildSpec.build_models.training_plan == null ? defaultTrainingPlan : buildSpec.build_models.training_plan;
```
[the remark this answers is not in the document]
seb-h2o: @deil87 good point!
I'm not happy this those 4 lines either.
I need `trainingStepsExecutor` for the `learn` method though.
What I'm rather thinking of, is to remove the parameter from the `TrainingStepsRegistry`.
Also, I don't want to burry those registry and executor too much: some scenario I have in mind is for very large datasets, where we could have a first plan executed against a sample to produce first models quickly and then another plan executed against the full dataset (this plan could also be designed based on the results of the first). 
Basically, want to use those 2 components to increase flexibility in AutoML execution, but yes, will try to group this logic into `planWork`, makes sense.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -331,28 +242,21 @@ WorkAllocations planWork() {
...
-            .allocate(Algo.XGBoost, 3, JobType.ModelBuild, 10)
-            .allocate(Algo.XGBoost, 1, JobType.HyperparamSearch, 100)
-            .allocate(Algo.StackedEnsemble, 2, JobType.ModelBuild, 15)
-            .end();
-
+    for (TrainingStep step: trainingSteps) {
+      workAllocations.allocate(step.makeWork());
+    }
     for (Algo skippedAlgo : skippedAlgos) {
       eventLog().info(Stage.ModelTraining, "Disabling Algo: "+skippedAlgo+" as requested by the user.");
```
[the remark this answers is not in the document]
seb-h2o: True, but there's another warning message for those.
Not sure it's worth making this logic more complex.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -361,15 +265,11 @@ public void run() {
...
-    for (Job j : jobs) j.get(); // Hold until they all completely stop.
-    jobs = null;
-
-    runCountdown.stop();
+    if (null == trainingStepsExecutor) return; // already stopped
+    trainingStepsExecutor.stop();
     eventLog().info(Stage.Workflow, "AutoML build stopped: " + EventLogEntry.dateTimeFormat.format(runCountdown.stop_time()))
             .setNamedValue("stop_epoch", runCountdown.stop_time(), EventLogEntry.epochFormat);
-    eventLog().info(Stage.Workflow, "AutoML build done: built " + modelCount + " models");
+    eventLog().info(Stage.Workflow, "AutoML build done: built " + trainingStepsExecutor.modelCount() + " models");
```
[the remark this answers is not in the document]
seb-h2o: `it is just to summarise what was logged along the execution`
I think it was the original purpose.
Each event is first logged during execution at its proper level (I remember this because I added automatic logging when calling `eventLog().info` for example).

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -573,104 +458,28 @@ private void handleDatafileParameters(AutoMLBuildSpec buildSpec) {
...
-        Grid<?> grid = (Grid) subJob.get();
-        int totalGridModelsBuilt = grid.getModelCount();
-        if (totalGridModelsBuilt > lastTotalGridModelsBuilt) {
-          eventLog().debug(stage, "Built: " + totalGridModelsBuilt + " models for search: " + name);
-          this.addModels(grid.getModelKeys());
+  private void learn() {
+    List<TrainingStep> executed = new ArrayList<>();
+    for (TrainingStep step : getExecutionPlan()) {
+        if (!exceededSearchLimits(step._description, step._ignoreConstraints)) {
+          if (trainingStepsExecutor.submit(step, job())) {
```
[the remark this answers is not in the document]
seb-h2o: getting an object from DKV is basically free when the object has been loaded once: DKV has inner cache.
And I prefer to use the DKV cache than having to manipulate 2 references for each variable.

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsExecutor.java
```diff
@@ -0,0 +1,180 @@
...
+                int totalGridModelsBuilt = grid.getModelCount();
+                if (totalGridModelsBuilt > lastTotalGridModelsBuilt) {
+                    eventLog().debug(Stage.ModelTraining, "Built: " + totalGridModelsBuilt + " models for search: " + jobDescription);
+                    this.addModels(grid.getModelKeys());
+                    lastTotalGridModelsBuilt = totalGridModelsBuilt;
+                }
+            }
+
+            try {
+                Thread.sleep(1000);
```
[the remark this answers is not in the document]
seb-h2o: agree, will make it a constant for now

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsExecutor.java
```diff
@@ -0,0 +1,180 @@
...
+        if (null != parentJob) {
+            parentJob.update(work._weight - lastWorkedSoFar);
+        }
+        work.consume();
+        _jobs.remove(job);
+    }
+
+
+    // If we have multiple AutoML engines running on the same project they will be updating the Leaderboard concurrently,
+    // so always use leaderboard() instead of the raw field, to get it from the DKV.
```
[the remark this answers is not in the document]
seb-h2o: sure, will do

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsExecutor.java
```diff
@@ -0,0 +1,180 @@
...
+        int after = leaderboard().getModelCount();
+        _modelCount.addAndGet(after - before);
+    }
+
+    private EventLog eventLog() {
+        return _eventLog == null ? null : (_eventLog = _eventLog._key.get());
+    }
+
+    private Leaderboard leaderboard() {
+        return _leaderboard == null ? null : (_leaderboard = _leaderboard._key.get());
```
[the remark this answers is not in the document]
seb-h2o: mmm, this null-check doesn't make sense anyway: `EventLog` and `Leaderboard` are injected through constructor and are `Iced` components (and therefore Serializable), so they can't be null.
However, we still need to ensure we always have a fresh instance, so I guess it can be simplified to:
```
private Leaderboard leaderboard() {
        return (_leaderboard = _leaderboard._key.get());
```

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsRegistry.java
```diff
@@ -0,0 +1,106 @@
...
+    public TrainingStep[] getOrderedSteps(AutoML aml) {
+        return getOrderedSteps(aml, _defaultTrainingPlan);
+    }
+
+
+    /**
+     * @param aml the AutoML instance responsible to execute the {@link TrainingStep}s.
+     * @return the list of {@link TrainingStep}s to execute according to the given training plan.
+     */
+    public TrainingStep[] getOrderedSteps(AutoML aml, StepDefinition[] trainingPlan) {
```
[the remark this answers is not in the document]
seb-h2o: For now yes.
Actually planned to use it later in case of multiple training plans, but now I think I should rather get rid of the `_defaultTrainingPlan`, which I mainly introduced for encapsulation but it is not really useful.

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsRegistry.java
```diff
@@ -0,0 +1,106 @@
...
+            try {
+                TrainingSteps steps = clazz.getConstructor(AutoML.class).newInstance(aml);
+                TrainingStep[] toAdd;
+                if (def._alias != null) {
+                    toAdd = steps.getSteps(def._alias);
+                } else if (def._steps != null) {
+                    toAdd = steps.getSteps(def._steps);
+                    if (toAdd.length < def._steps.length) {
+                        List<String> toAddIds = Stream.of(toAdd).map(s -> s._id).collect(Collectors.toList());
+                        Stream.of(def._steps)
```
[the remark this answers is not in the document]
seb-h2o: :)

h2o-core/src/main/java/water/util/PojoUtils.java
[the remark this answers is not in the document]
seb-h2o: Asking him, although this logic is used only by `AutoML` currently.
@michalkurka: if you want to have a look at changes in `PojoUtils`

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -46,97 +32,23 @@
...
-
-  }
-
   private final static boolean verifyImmutability = true; // check that trainingFrame hasn't been messed with
   private final static SimpleDateFormat timestampFormatForKeys = new SimpleDateFormat("yyyyMMdd_HHmmss");
 
+  private static StepDefinition[] defaultTrainingPlan = {
+          new StepDefinition(Algo.XGBoost.name(), Alias.defaults),
+          new StepDefinition(Algo.GLM.name(), Alias.defaults),
+          new StepDefinition(Algo.DRF.name(), new String[]{ "def_1" }),
```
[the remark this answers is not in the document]
seb-h2o: it's just an identifier, not sure I want to introduce a constant for this: it's only used for `DRF` besides, as aliases are usually good enough to express the current execution plan.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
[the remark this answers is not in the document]
deil87: @seb-h2o sorry, obviously I was not clear enough in explanation :) I was talking about reducing 500ms to a smaller value ( and making loops shorter) let's say 10ms or 50 ms - it will not hurt and just reduce maximum of possible waste of time. Agree, it is not a big difference after all but if we can easily mitigate this loss then maybe it is also not a big deal.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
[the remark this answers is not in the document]
deil87: @seb-h2o maybe those short and full work plans could be still created within planWork. Not sure about possible obstacles for that right now. But I think you got my point.

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsExecutor.java
```diff
@@ -0,0 +1,180 @@
...
+
+
+/**
+ * Class responsible for starting all the {@link TrainingStep}s and monitoring their associated {@link Job},
+ * i.e. polling jobs and adding their result model(s) to the {@link Leaderboard}.
+ */
+class TrainingStepsExecutor extends Iced<TrainingStepsExecutor> {
+
+    EventLog _eventLog;
+    Leaderboard _leaderboard;
```
michalkurka: this will serialize the leaderboard, but doesn't leaderboard have its own update mechanism? should we just remember the leaderboard key here?

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsExecutor.java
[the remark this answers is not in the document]
michalkurka: in that case - my comment from above is most likely valid - leaderboard shoudln't be serialized a part of this class

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsRegistry.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * The registry responsible for loading all {@link TrainingStepsProvider} using service discovery,
+ * and providing the list of {@link TrainingStep} to execute.
+ */
+public class TrainingStepsRegistry extends Iced<TrainingStepsRegistry> {
```
michalkurka: this is great, it allows for writing user extensions without modifying automl code

- first application could be xgboost, xgboost is built as an h2o-extension and it would be great if automl didn't have to depend on the xgboost module directly

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsRegistry.java
```diff
@@ -0,0 +1,106 @@
...
+
+    static final NonBlockingHashMap<String, Class<TrainingSteps>> stepsByName = new NonBlockingHashMap<>();
+
+    static {
+        ServiceLoader<TrainingStepsProvider> trainingStepsProviders = ServiceLoader.load(TrainingStepsProvider.class);
+        for (TrainingStepsProvider provider : trainingStepsProviders) {
+            stepsByName.put(provider.getName(), provider.getStepsClass());
```
michalkurka: is there a reason why the providers are not factories for the training steps - what is the intention of registering a class instead of the provider?

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsRegistry.java
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * @param aml the AutoML instance responsible to execute the {@link TrainingStep}s.
+     * @return the list of {@link TrainingStep}s to execute according to the given training plan.
+     */
+    public TrainingStep[] getOrderedSteps(AutoML aml, StepDefinition[] trainingPlan) {
+        aml.eventLog().info(Stage.Workflow, "Loading execution steps: "+Arrays.toString(trainingPlan));
+        List<TrainingStep> orderedSteps = new ArrayList<>();
+        for (StepDefinition def : trainingPlan) {
+            Class<TrainingSteps> clazz = stepsByName.get(def._name);
```
michalkurka: imho this could be simplified if the providers were itself the factory classes

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsRegistry.java
seb-h2o: @michalkurka I didn't have `xgboost` in mind when doing it, but yes, agree that the XGB steps could be moved to the xgb extension.
Won't do it in this PR, but will try afterwards.

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsRegistry.java
```diff
@@ -0,0 +1,106 @@
...
+                TrainingStep[] toAdd;
+                if (def._alias != null) {
+                    toAdd = steps.getSteps(def._alias);
+                } else if (def._steps != null) {
+                    toAdd = steps.getSteps(def._steps);
+                    if (toAdd.length < def._steps.length) {
+                        List<String> toAddIds = Stream.of(toAdd).map(s -> s._id).collect(Collectors.toList());
+                        Stream.of(def._steps)
+                                .filter(s -> !toAddIds.contains(s._id))
+                                .forEach(s -> aml.eventLog().warn(Stage.Workflow,
```
michalkurka: a warning is generated but where is the step actually skipped?

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsRegistry.java
seb-h2o: Very good point, don't know why they are not factories taking `AutoML` param, probably because I introduced them a posteriori... will try to turn them into proper factories and get rid of the Java reflection.

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsRegistry.java
seb-h2o: agree

h2o-core/src/main/java/water/util/OSUtils.java
```diff
@@ -53,6 +53,10 @@ public static boolean isLinux() {
     return getOsName().toLowerCase().startsWith("linux");
   }
 
+  public static boolean isMac() {
```
michalkurka: appears to be unused

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsExecutor.java
seb-h2o: yes, we should, will change that.

h2o-core/src/main/java/water/util/OSUtils.java
seb-h2o: after all, yes, unused: ended up using the `isLinux` one as multinode is supported only on linux.
Want me to remove it or can it become useful somehow?

michalkurka:
This is a very nice change, opens up doors to 3rd party extensions in AutoML.

Another use would be for "recipes" - prioritizing different algos for different problems (multiclass, wide datasets, ...). Perhaps the training step could take a peek at the training data and calculate a "priority" values - which would then play a role in the ordering of steps.

What would be really cool to eventually have is to be able to generate an execution plan for a particular kind of problem - present it to the user in a readable form and let them edit it to their likings and submit to the backend.

h2o-core/src/main/java/water/util/OSUtils.java
michalkurka: up to you, doesn't bother me much - impossible to test in current infrastructure though

h2o-automl/src/main/java/ai/h2o/automl/TrainingStepsRegistry.java
seb-h2o: it's done when listing the steps: didn't want to introduce the warning there in `getSteps` as I can't generate a meaningful message there: in general, I prefer generating user-facing messages at a rather high-level.

h2o-automl/src/main/java/ai/h2o/automl/TrainingSteps.java
```diff
@@ -0,0 +1,75 @@
...
+                if (step._weight != Step.DEFAULT_WEIGHT) {
+                    tStep._weight = step._weight;  // override default weight
+                }
+                tSteps.add(tStep);
+            });
+        }
+        return tSteps.toArray(new TrainingStep[0]);
+    }
+
+    TrainingStep[] getSteps(String[] ids) {
```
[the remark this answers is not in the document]
seb-h2o: correct, not needed anymore -> removing

seb-h2o:
@michalkurka thanks, yes, I'm definitely glad that this change is putting behind the static training of AutoML. Now we need to define all the strategies we want to apply (like https://0xdata.atlassian.net/browse/PUBDEV-6864). For large datasets, maybe we could also prioritize them based on models trained on data sample and so on... given that this should be mainly (but not only) useful for large datasets where AutoML doesn't always have time to try all the algos or to explore the hyperparameter space with the grids.

For the user interaction: this would be a good expert feature, and quite easy to implement (most of the work is done in this PR).

PR #3872, opened 2019-09-12
PUBDEV-6873: add aucpr

wendycwong:
This PR completes the task in https://0xdata.atlassian.net/browse/PUBDEV-6873?filter=-1

I tried to use Pasha backward compatibility class but it is not working.  It did not even work for giniCoef.

Instead I just let aucpr to call pr_auc.
Added a pyunit test to make sure the fix works.


wendycwong:
pr_auc() is the very first API.  No backward compatibility issues here.

PR #3882, opened 2019-09-16
PUBDEV-6323: Implement GLM native multinomial classification

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6323?filter=-1

1. add solver IRLSM_NATIVE and if enabled will solve the GLM multinomial in one shot instead of one versus many.
2. Generate gradient for multinomial and a Junit test to check with manual calculation
3. Generate gram and xy and a Junit test to check with manual calculation
4. change ADMM and LS so that it works with my multinomial. Added a junit test for confirm that.
5. Add Python and R unit tests.
6. Add tests for lambda_search.

The trick to understand this PR is to realize that we are dealing with all the coefficients together.  Hence, some of the thresholds need to change because of that.


h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -215,6 +227,23 @@ public int compare(Integer o1, Integer o2) {
     return res;
   }
 
+  private static double [] extractSubRange(int N, int [] ids, double [] src, int offset) {
+    if(ids == null) return Arrays.copyOfRange(src,offset,offset+ids.length);
```
honzasterba: this looks like it will cause NPE. ids == null -> ids.lenght => NPE

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -215,6 +227,23 @@ public int compare(Integer o1, Integer o2) {
     return res;
   }
 
+  private static double [] extractSubRange(int N, int [] ids, double [] src, int offset) {
```
honzasterba: param N not used

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -265,25 +328,78 @@ public GLMSubsetGinfo(GLMGradientInfo fullInfo, int N, int c, int [] ids) {
...
   public GradientSolver gslvrMultinomial(final int c) {
     final double [] fullbeta = _beta.clone();
     return new GradientSolver() {
       @Override
-      public GradientInfo getGradient(double[] beta) {
-        fillSubRange(_activeData.fullN()+1,c,_activeDataMultinomial[c].activeCols(),beta,fullbeta);
-        GLMGradientInfo fullGinfo =  _gslvr.getGradient(fullbeta);
-        return new GLMSubsetGinfo(fullGinfo,_activeData.fullN()+1,c,_activeDataMultinomial[c].activeCols());
+      public GradientInfo getGradient(double[] beta) { // beta stores coeff of one class for other
+        if (_multinomialSpeedup) {  // beta only contains active coefficients of all classes, need to expand it back to fullbeta
```
honzasterba: I would very much extract the IF outside of the class and used inharitance to separate the two if branches

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -530,10 +647,27 @@ public boolean converged(){
   }
 
   protected double updateState(double [] beta,GLMGradientInfo ginfo){
-    _betaDiff = ArrayUtils.linfnorm(_beta == null?beta:ArrayUtils.subtract(_beta,beta),false);
+    _betaDiff = _activeColsAll==null
```
honzasterba: please replace multi line ternary operator with if/else, also looks like a candidate for extract to method

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -628,33 +762,46 @@ public static GramXY addCols(double[] beta, final int[] newActiveCols, final int
     }
   }
 
+  // activeData contains only active columns of one class
   protected GramXY computeNewGram(DataInfo activeData, double [] beta, GLMParameters.Solver s){
     double obj_reg = _parms._obj_reg;
     if(_glmw == null) _glmw = new GLMModel.GLMWeightsFun(_parms);
-    GLMTask.GLMIterationTask gt = new GLMTask.GLMIterationTask(_job._key, activeData, _glmw, beta,_activeClass).doAll(activeData._adaptedFrame);
+    GLMTask.GLMIterationTask gt = _parms._solver.equals(GLMParameters.Solver.IRLSM_NATIVE)?
```
honzasterba: this is not readable, extract method and use if/else

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1582,21 +1672,21 @@ else if(_parms._lambda_search) { // lambda search prefers coordinate descent
...
-
+    
     private static double boundedX(double x, double lb, double ub) {
       if (x < lb) x = lb;
       if (x > ub) x = ub;
       return x;
     }
 
-    public GramSolver(Gram gram, double[] xy, double lmax, double betaEps, boolean intercept) {
+/*    public GramSolver(Gram gram, double[] xy, double lmax, double betaEps, boolean intercept) {
```
honzasterba: if it can be commented out it can be deleted

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1634,73 +1725,158 @@ public GramSolver(Gram gram, double[] xy, boolean intercept, double l2pen, doubl
...
-        rhos[icpt] = 1;//(xy[icpt] >= 0 ? xy[icpt] : -xy[icpt]);
+
+      int lastIndex = coeffPClass - ii;
+      if (activeColAll==null) { // find rhos for full length beta
+        if (l1pen != 0) {   // no need to play with rho if l1pen is zero
+          double min = Double.POSITIVE_INFINITY;
+          for (int classInd = 0; classInd < classIndBound; classInd++) {
+            for (int i = 0; i < lastIndex; ++i) { // this loop excludes the intercept
+              double d = xy[i + classInd * coeffPClass];
+              d = d >= 0 ? d : -d;
```
honzasterba: use Math.abs is more readable

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1634,73 +1725,158 @@ public GramSolver(Gram gram, double[] xy, boolean intercept, double l2pen, doubl
...
-        double x = ((y - ybar * xbar) / ((gram.get(i, i) - xbar * xbar) + l2pen));///gram.get(i,i);
-        rhos[i] = ADMM.L1Solver.estimateRho(x, l1pen, lb == null ? Double.NEGATIVE_INFINITY : lb[i], ub == null ? Double.POSITIVE_INFINITY : ub[i]);
-      }
-      // do the intercept separate as l1pen does not apply to it
-      if (intercept && (lb != null && !Double.isInfinite(lb[icptCol]) || ub != null && !Double.isInfinite(ub[icptCol]))) {
-        int icpt = xy.length - 1;
-        rhos[icpt] = 1;//(xy[icpt] >= 0 ? xy[icpt] : -xy[icpt]);
+
+      int lastIndex = coeffPClass - ii;
+      if (activeColAll==null) { // find rhos for full length beta
```
honzasterba: this method is far too long, the comment is indicative of this being a good place to split it

h2o-algos/src/main/java/hex/glm/ComputationState.java
wendycwong: Thank you Honza for the review!  Let me fix it.

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -78,7 +90,7 @@ public void setLambda(double lambda) {
     adjustToNewLambda(lambda, 0);
   }
   public double [] beta(){
-    if(_activeClass != -1)
+    if(_activeClass != -1 && (!_multinomialSpeedup))
```
honzasterba: extra () braces make this less readable

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -215,6 +227,23 @@ public int compare(Integer o1, Integer o2) {
...
+    double [] res = MemoryManager.malloc8d(ids.length);
+    int j = 0;
+    for(int i:ids)
+      res[j++] = src[offset+i];
+    return res;
+  }
+
+  /**
+   * 
+   * @param N: number of coefficients per class
```
honzasterba: javadocs does not like : here

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -59,12 +62,21 @@ public ComputationState(Job job, GLMParameters parms, DataInfo dinfo, BetaConstr
     _nclasses = (parms._family == Family.multinomial||parms._family == Family.ordinal)?nclasses:1;
     _alpha = _parms._alpha[0];
   }
-
+
```
Pscheidl:
```suggestion

```

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -59,12 +62,21 @@ public ComputationState(Job job, GLMParameters parms, DataInfo dinfo, BetaConstr
...
   }
+  public void setActiveColsAll() {
+    _activeColsAll = new int[_nclasses][];
+    for (int classInd=0; classInd < _nclasses; classInd++) {
+      int[] original = activeDataMultinomial(classInd)._activeCols;
+      int activeColsLen = original.length;
+      _activeColsAll[classInd] = new int[activeColsLen];
+      System.arraycopy(original, 0, _activeColsAll[classInd], 0, activeColsLen);
+    }
+  }
```
Pscheidl: I don't see the advantage of copying the data. Is there any ? Can't we just link to it ? Correct me if I'm wrong please. Like that:
```suggestion
  public void setActiveColsAll() {
    _activeColsAll = new int[_nclasses][];
    for (int classInd=0; classInd < _nclasses; classInd++) {
      int[] original = activeDataMultinomial(classInd)._activeCols;
      _activeColsAll[classInd] = activeDataMultinomial(classInd)._activeCols;
    }
  }
```

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -105,7 +117,7 @@ public String toString() {
   private void adjustToNewLambda(double lambdaNew, double lambdaOld) {
     double ldiff = lambdaNew - lambdaOld;
     if(ldiff == 0 || l2pen() == 0) return;
-    double l2pen = .5*ArrayUtils.l2norm2(_beta,true);
+    double l2pen = .5*ArrayUtils.l2norm2(_beta,true, _multinomialSpeedup?_nclasses:1);
```
Pscheidl: Formatting only.

```suggestion
    double l2pen = .5 * ArrayUtils.l2norm2(_beta, true, _multinomialSpeedup ? _nclasses : 1);
```

h2o-algos/src/main/java/hex/glm/ComputationState.java
Pscheidl: Can this really happen ? I couldn't reproduce it with current test battary. If yes, there should be a test.
Is the solution to copy the full length ? `if (ids == null) return Arrays.copyOfRange(src, offset, src.length);` ?

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -215,6 +227,23 @@ public int compare(Integer o1, Integer o2) {
...
+  }
+
+  /**
+   * 
+   * @param N: number of coefficients per class
+   * @param c: class index
+   * @param ids: column indices of active columns
+   * @param src: values of active columns only
+   * @param dst: coefficients of all classes stacked up
+   */
```
Pscheidl:
```suggestion
  /**
   * 
   * @param N number of coefficients per class
   * @param c class index
   * @param ids column indices of active columns
   * @param src values of active columns only
   * @param dst coefficients of all classes stacked up
   */
```

wendycwong:
From my test results off Hadoop cluster, I do not find significant gain in speed.

PR #3883, opened 2019-09-16
Also release new cdh 15 and 16 hdp jars

honzasterba:


michalkurka:
I already added these

PR #3890, opened 2019-09-17
[PUBDEV-6821] Expose leader node idx via rest api

jakubhava:
Reincarnation of https://github.com/h2oai/h2o-3/pull/3834 as it was reverted due to failure mentioned here https://github.com/h2oai/h2o-3/pull/3839.

Let's wait for infra to verify this will do the trick :)

CC: @mn-mikke in the rest api based external backend, we should always use the leader node for the Flow UI and rest API

### DIFF
--- a/h2o-core/src/test/java/water/api/CloudHandlerTest.java
+++ b/h2o-core/src/test/java/water/api/CloudHandlerTest.java
@@ -1,0 +1,25 @@
+package water.api;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+import water.TestUtil;
+import water.api.schemas3.CloudV3;
+
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class CloudHandlerTest extends TestUtil {
+
+  @BeforeClass
+  public static void beforeClass() {
+    TestUtil.stall_till_cloudsize(1);
+  }
+
+  @Test
+  public void status() {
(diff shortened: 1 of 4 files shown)


jakubhava:
Seems like unrelated xgboost test failures, cc: @honzasterba

h2o-core/src/test/java/water/api/CloudHandlerTest.java
```diff
@@ -0,0 +1,25 @@
...
+  public static void beforeClass() {
+    TestUtil.stall_till_cloudsize(1);
+  }
+
+  @Test
+  public void status() {
+    final CloudHandler cloudHandler = new CloudHandler();
+    final CloudV3 status = cloudHandler.status(3, new CloudV3());
+    assertNotNull(status);
+    assertNotEquals(status.leader_idx, -1);
```
honzasterba: it would be nice to send a request to establish a leader and then do another where the return value would be >= 0

honzasterba:
yes, those are unrelated

h2o-core/src/test/java/water/api/CloudHandlerTest.java
jakubhava: Good idea! In this tests we already have leader node assigned and test that the value is >=0

honzasterba: oh true, I missed the Not in assertNotEquals

PR #3904, opened 2019-09-19
Better error reporting when a chunk is missing (avoid NPE)

michalkurka:


michalkurka:
@jakubhava should help with your debugging

PR #3906, opened 2019-09-20
PUBDEV-6902: Fix model summary & model performance in Zeppelin

michalkurka:


michalkurka:
I decided to avoid html-print formatting for most of the output in this PR and I prefer to print the tables using Pandas, the reason is that plain-text output doesn't mix well with html-output with Zeppelin (notice how the metrics are on a single line instead of correctly on separate lines):

issue

h2o init:

init

model.show()

model-show

model.model_performance()

model-performance

model.summary()

model-summary

frame.summary()

frame-summary

PR #3908, opened 2019-09-21
[PUBDEV-6904] Fix issue with missing chunk on external backend

jakubhava:
https://0xdata.atlassian.net/browse/PUBDEV-6904

### DIFF
--- a/h2o-core/src/main/java/water/ExternalFrameWriterClient.java
+++ b/h2o-core/src/main/java/water/ExternalFrameWriterClient.java
@@ -124,73 +124,74 @@ final public class ExternalFrameWriterClient {
     waitForRequestToFinish(timeout, requestType);
   }
 
-    public void close() throws IOException {
+    public void close() throws IOException, ExternalFrameConfirmationException {
+        waitForRequestToFinish(timeout, ExternalBackendRequestType.WRITE_TO_CHUNK.getByte());
         channel.close();
     }
 
-    public void sendBoolean(boolean data) throws IOException, ExternalFrameConfirmationException {
+    public void sendBoolean(boolean data) throws IOException {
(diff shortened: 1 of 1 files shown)


h2o-core/src/main/java/water/ExternalFrameWriterClient.java
```diff
@@ -124,73 +124,74 @@ public void finalizeFrame(String keyName, long[] rowsPerChunk, byte[] colTypes,
     waitForRequestToFinish(timeout, requestType);
   }
 
-    public void close() throws IOException {
+    public void close() throws IOException, ExternalFrameConfirmationException {
+        waitForRequestToFinish(timeout, ExternalBackendRequestType.WRITE_TO_CHUNK.getByte());
```
jakubhava: always wait for this request on backend to finish in close method as it was designed for it. In case the user of this API sends the empty partition no sent method will be called and therefore we never wait for request to finish

h2o-core/src/main/java/water/ExternalFrameWriterClient.java
```diff
@@ -214,15 +215,10 @@ private void waitForRequestToFinish(int timeout, byte confirmation) throws Exter
     }
   }
 
-  private void increaseCurrentColIdx() throws IOException, ExternalFrameConfirmationException {
+  private void increaseCurrentColIdx() {
     currentColIdx++;
     if (currentColIdx == numCols) {
       currentRowIdx++;
-      if (currentRowIdx == numRows) {
-        writeToChannel(ab, channel);
```
jakubhava: there is nothing to write to channel as we write data directly as part of `send` methods. This invocation can be safely removed

michalkurka:
thanks @jakubhava

PR #3909, opened 2019-09-23
PUBDEV-6870 - Error in PredictCSV unused column detection

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6870

This fixed the original issue as described in the JIRA. However, this solution as implemented here still has some limitations:

1. If the dataset predictions are made on has no header, it is going to report the first line as unused - which is wrong.
2. If the dataset predictions are made on has different header, it is going to report those haders as unused columns. 

Both points of course have no impact on prediction. This is hard to fix, the code has no way to know whether there is header or no, despite some heuristics can be done (detection of numbers in headers etc. - but still, there are valid datasets with features numbered as plain integers)

Also, weak spot in our CSV writer is found again in the output:

```
Downloading MOJO @... /tmp/tmp2666f_ch
    => /tmp/tmp2666f_ch/model.zip  (81897 bytes)
    Time taken = 0.051s
    => /tmp/tmp2666f_ch/model.zip  (81897 bytes)
    => /tmp/tmp2666f_ch/h2o-genmodel.jar  (12872947 bytes)
Detected 1 unused columns in the input data set: {}
```

Reason ? Here:
```
Model column names: [AGE, RACE, DPROS, DCAPS, PSA, VOL, GLEASON, CAPSULE]
Parsed column names: [, AGE, RACE, DPROS, DCAPS, PSA, VOL, GLEASON]
Parsed response names: CAPSULE
```

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/tools/PredictCsv.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/tools/PredictCsv.java
@@ -418,7 +418,7 @@ public class PredictCsv {
       if (!parsedColumnNames.contains(columnName) && !columnName.equals(model.m._responseColumn)) {
         missingColumns.add(columnName);
       } else {
-        parsedColumnNames.remove(missingColumns);
+        parsedColumnNames.remove(columnName);
       }
     }
     
--- a/h2o-py/tests/testdir_misc/pyunit_mojo_predict.py
+++ b/h2o-py/tests/testdir_misc/pyunit_mojo_predict.py
@@ -181,7 +181,7 @@ def mojo_predict_csv_test(target_dir):
 
     download_mojo(multi_gbm, mojo_zip_path)
 
-    print("\nPerforming Binomial Prediction using MOJO @... " + target_dir)
+    print("\nPerforming Multinomial Prediction using MOJO @... " + target_dir)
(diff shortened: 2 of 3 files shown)


h2o-genmodel/src/main/java/hex/genmodel/tools/PredictCsv.java
```diff
@@ -418,7 +418,7 @@ private void checkMissingColumns(final String[] parsedColumnNamesArr) {
       if (!parsedColumnNames.contains(columnName) && !columnName.equals(model.m._responseColumn)) {
         missingColumns.add(columnName);
       } else {
-        parsedColumnNames.remove(missingColumns);
+        parsedColumnNames.remove(columnName);
```
Pscheidl: Actual bugfix.

h2o-py/tests/testdir_misc/pyunit_mojo_predict.py
```diff
@@ -181,7 +181,7 @@ def mojo_predict_csv_test(target_dir):
 
     download_mojo(multi_gbm, mojo_zip_path)
 
-    print("\nPerforming Binomial Prediction using MOJO @... " + target_dir)
+    print("\nPerforming Multinomial Prediction using MOJO @... " + target_dir)
```
Pscheidl: Copy & paste error

michal-raska: Thanks!

PR #3910, opened 2019-09-23
PUBDEV-6494: training_frame requirements for StackedEnsemble in blending mode

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6494
https://0xdata.atlassian.net/browse/PUBDEV-6612 (minor change, related to blending mode as well)

In blending mode:
- `training_frame` is not required for the training of the SE model.
- base models are allowed to have been trained using different `training_frame`s.

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
@@ -279,7 +279,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
       throw new H2OIllegalArgumentException("Cannot specify fold_column and nfolds at the same time.");
 
     Model aModel = null;
-    boolean beenHere = false;
+    boolean retrievedFirstModelParams = false;
     boolean blending_mode = _parms._blending != null;
     boolean cv_required_on_base_model = !blending_mode;
     boolean require_consistent_training_frames = !blending_mode && !_parms._is_cv_model;
@@ -302,7 +302,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
         throw new H2OIllegalArgumentException("Base model is not supervised: "+aModel._key.toString());
       }
 
-      if (beenHere) {
+      if (retrievedFirstModelParams) {
         // check that the base models are all consistent with first based model
 
         if (modelCategory != aModel._output.getModelCategory())
@@ -311,7 +311,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
 
         if (! responseColumn.equals(aModel._parms._response_column))
           throw new H2OIllegalArgumentException("Base models are inconsistent: they use different response columns."
-                  +" Found: "+responseColumn+"(StackedEnsemble) and "+aModel._parms._response_column+" (model "+k+").");
+                  +" Found: "+responseColumn+" (StackedEnsemble) and "+aModel._parms._response_column+" (model "+k+").");
 
         if (require_consistent_training_frames) {
           if (trainingFrameRows < 0) trainingFrameRows = _parms.train().numRows();
@@ -362,7 +362,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
         // TODO: If we're set to DistributionFamily.AUTO then GLM might auto-conform the response column
         // giving us inconsistencies.
       } else {
-        // !beenHere: this is the first base_model
+        // !retrievedFirstModelParams: this is the first base_model
         this.modelCategory = aModel._output.getModelCategory();
         this._dist = DistributionFactory.getDistribution(distributionFamily(aModel));
         responseColumn = aModel._parms._response_column;
@@ -377,7 +377,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
         basemodel_fold_column = aModel._parms._fold_column;
         seed = aModel._parms._seed;
         _parms._distribution = aModel._parms._distribution;
-        beenHere = true;
+        retrievedFirstModelParams = true;
       }
 
     } // for all base_models
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
@@ -283,6 +283,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
     boolean beenHere = false;
     boolean blending_mode = _parms._blending != null;
     boolean cv_required_on_base_model = !blending_mode;
+    boolean require_consistent_training_frames = !blending_mode && !_parms._is_cv_model;
     
     //following variables are collected from the 1st base model (should be identical across base models), i.e. when beenHere=false
     int basemodel_nfolds = -1;
@@ -291,7 +292,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
     long seed = -1;
     //end 1st model collected fields
     
-    trainingFrameRows = _parms.train().numRows();
+    trainingFrameRows = require_consistent_training_frames ? _parms.train().numRows() : 0;
 
     for (Key<Model> k : _parms._base_models) {
       aModel = DKV.getGet(k);
@@ -312,7 +313,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
 
         // NOTE: if we loosen this restriction and fold_column is set add a check below.
         Frame aTrainingFrame = aModel._parms.train();
-        if (trainingFrameRows != aTrainingFrame.numRows() && !this._parms._is_cv_model)
+        if (require_consistent_training_frames && trainingFrameRows != aTrainingFrame.numRows())
           throw new H2OIllegalArgumentException("Base models are inconsistent: they use different size (number of rows) training frames." +
               " Found: "+trainingFrameRows+" (StackedEnsemble) and "+aTrainingFrame.numRows()+" (model "+k+").");
 
@@ -320,11 +321,6 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
           throw new H2OIllegalArgumentException("Base models are inconsistent: they use different response columns." +
               " Found: "+responseColumn+"(StackedEnsemble) and "+aModel._parms._response_column+" (model "+k+").");
         
-//        if (blending_mode && _parms._blending.equals(aModel._parms._train)) {
-//          throw new H2OIllegalArgumentException("Base model `"+k+"` was trained with the StackedEnsemble blending frame.");
-//        }
-
-        // TODO: we currently require xval; loosen this iff we add a separate holdout dataset for the ensemble
         if (cv_required_on_base_model) {
           if (aModel._parms._fold_assignment != basemodel_fold_assignment) {
             if ((aModel._parms._fold_assignment == AUTO && basemodel_fold_assignment == Random) ||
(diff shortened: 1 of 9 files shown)


seb-h2o:
@ledell @michalkurka @deil87 anyone to review this change? PR is currently in sleeping mode. Thx

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -283,79 +282,72 @@ void checkAndInheritModelProperties() {
...
-            }
-          }
-
-          // If we have a fold_column make sure nfolds from base models are consistent
-          if (aModel._parms._fold_column == null && basemodel_nfolds != aModel._parms._nfolds)
-            throw new H2OIllegalArgumentException("Base models are inconsistent: they use different values for nfolds.");
+          throw new H2OIllegalArgumentException("Base models are inconsistent: they use different response columns."
+                  +" Found: "+responseColumn+"(StackedEnsemble) and "+aModel._parms._response_column+" (model "+k+").");
+
+        if (require_consistent_training_frames) {
```
deil87: @seb-h2o I believe there is still a chance that training frames are different even though number of rows is not. We decided not to consider this case as it is a highly unlikely use case?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -283,79 +282,72 @@ void checkAndInheritModelProperties() {
...
     long seed = -1;
     //end 1st model collected fields
-    
-    trainingFrameRows = _parms.train().numRows();
 
     for (Key<Model> k : _parms._base_models) {
       aModel = DKV.getGet(k);
       if (null == aModel) {
-        Log.warn("Failed to find base model; skipping: " + k);
+        Log.warn("Failed to find base model; skipping: "+k);
```
deil87: @seb-h2o tiny question, is it a new convention that we should remove spaces in Java?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -283,79 +282,72 @@ void checkAndInheritModelProperties() {
...
         continue;
       }
-      Log.debug("Checking properties for model " + k);
+      Log.debug("Checking properties for model "+k);
       if (!aModel.isSupervised()) {
-        throw new H2OIllegalArgumentException("Base model is not supervised: " + aModel._key.toString());
+        throw new H2OIllegalArgumentException("Base model is not supervised: "+aModel._key.toString());
       }
 
       if (beenHere) {
```
deil87: @seb-h2o I know this is not you code, but I would rename `beenHere` as it is too generic.
```
 if(!isFirstBaseModel) {}
```
or 
```
 if(!isReferenceBaseModel) {}
```
would be clearer and we wont need some comments to explain `beenHere`. Ignore this if you don't feel like touching legacy code :)

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -283,79 +282,72 @@ void checkAndInheritModelProperties() {
...
-          if (aModel._parms._fold_column == null && aModel._parms._nfolds < 2)
-            throw new H2OIllegalArgumentException("Base model does not use cross-validation: " + aModel._parms._nfolds);
+        if (cv_required_on_base_model) {
 
-          // NOTE: we already check that the training_frame checksums are the same, so
-          // we don't need to check the Vec checksums here:
-          if (aModel._parms._fold_column != null &&
-              ! aModel._parms._fold_column.equals(basemodel_fold_column))
-            throw new H2OIllegalArgumentException("Base models are inconsistent: they use different fold_columns.");
+          if (aModel._parms._fold_assignment != basemodel_fold_assignment
```
deil87: @seb-h2o what if we have following situation:
```
aModel._parms._fold_assignment == Random && basemodel_fold_assignment == AUTO
```
shouldn't we exclude it as well? or it is impossible to get into this state?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
seb-h2o: No, but it's not an uncommon practice to remove spaces around + operator just for string concatenations (in any language that supports it) as it improves readability:
- "+ and +" appears as placeholder delimiters without interrupting the read (we ignore the + quickly when reading):
          -  "hi, my name is "+name+" and I live in "+city.
          -  "hi, my name is " + name + " and I live in " + city.
- you really see the spaces in the sentence and are less likely to forget them:
         - "hi, my name is "+name+"and I live in"+city.
         - "hi, my name is " + name + "and I live in" + city.

for all other use-cases, I use spaces, but I think that for string concat, this was a bad convention in the first place.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
seb-h2o: L376:
` if (basemodel_fold_assignment == AUTO) basemodel_fold_assignment = Random;`

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
seb-h2o: I agree, will rename

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
seb-h2o: @deil87 I don't think it's a question of likely/unlikely.
It is just a question of what is necessary for StackedEnsemble to be able to stack the models.

When using xval, we allow the various models to be trained with different training frames, they may have a different set of predictors or have different feature engineering for example, but the response column should be the same, and xval predictions MUST have the same number of rows, that's all. If user is using a completely different training frame with the same number of rows, and a response column with a similar name (e.g. "class"), then we won't detect it. Maybe we could store the hash of the response column for additional safety? Wasn't done until now.

When using blending, the rows restriction is even removed, remains only the name/type of the response... you could almost blend orange with apples, which is not that uncommon in juice.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
deil87: @seb-h2o i think it is subjective which way is more readable :) I don't mind any of the formatting but it is better to stick to one across the project/team. I just remember someone was suggesting me during reviews to add those spaces in Java and remove ones in python  :)  Never mind.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
deil87: @seb-h2o I see now. It was not on the surface let's say :)

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
seb-h2o: It's hard to compare Python and Java in terms of style.
Python is very opinionated as it aims at optimal readability (because we read 100x times more code than we write), there's even a PEP for it: https://www.python.org/dev/peps/pep-0008/
Java commonly-used code style is just a convention following old SUN internal code style, but it varies so much that people implemented tools to apply different standards by company for example...
And I noticed that you often tend to use the word "subjective" when you don't have a strong opinion on sth. Are you so sure it is? Do you think that Python is only subjectively more readable than Perl for example? Can't there be any objective truth in styling? due to our physiological/psychological/neurological limitations? :)

deil87: @seb-h2o I do have strong opinion on many things. And my strong opinion was that this is subjective :)  I believe that every developer/human being is unique. You have provided examples of code as a proof( to support objectiveness) but I strongly believe that this example does not prove objectiveness of your point of view. 
Styling is about order and order(any reasonable order actually) makes it easier to interpret(read) anything. Because we get used to something and it becomes easier for us. Btw i have not found convention in this PEP regarding string concatenation. Maybe because it is arguable? 

Code styles exist but they are different even though they all seek for optimal readability. It is another proof that not every code style rule could be considered as universally correct one.

>  Can't there be any objective truth in styling? due to our physiological/psychological/neurological limitations? 

exactly because of our limitations, which varies from person to person like how good your eyesight is, it is hard to find common truth in such questions.

seb-h2o: that's a bit of a flawed reasoning:
> You have provided examples of code as a proof
No, examples can never serve as a proof, they just serve as examples. Only counter-examples can prove the falseness of a statement. That's the eternal problem of induction.

> I believe that every developer/human being is unique
> our limitations, which varies from person to person like how good your eyesight is, it is hard to find common truth in such questions
Unique doesn't mean that we're all different. Otherwise, you can throw away all researches in biology, psychology, social sciences, and what's even the point in building models if we don't expect them to apply because everyone is unique.
Yes, people are unique **at some point**, otherwise how boring, but they're also very similar, otherwise how impossible to communicate, live in cities with millions of them...
And when I say that adding extra spaces make it more difficult to distinguish the spaces inside the strings from the ones outside, I think you're just arguing for the sake of arguing.
Objective doesn't mean absolute eternal truth (that would be a straw man): for example red and blue objects can be said to be objectively different, although there's nothing like green and red colours, they're just an effect of perception due to our "physiological/psychological/neurological" system. However, yes, you can find people that won't make the difference, but even this fact, we can explain it objectively, because the people who don't see the difference are likely to have defective cones in the retina.
But if you reject the premise `red and blue objects can be said to be objectively different`, you just reject the work objectivity altogether, saying it's meaningless, in which case, subjectivity becomes also meaningless because well... everything is subjective.

and...
> Btw i have not found convention in this PEP regarding string concatenation. Maybe because it is arguable?
C'mon, no one uses string concatenation in Python! the Python way is string formatting or `join`.

michalkurka: > However, yes, you can find people that won't make the difference, but even this fact, we can explain it objectively, because the people who don't see the difference are likely to have defective cones in the retina.

It can also be just too dark.

deil87: @seb-h2o 
>No, examples can never serve as a proof, they just serve as examples. Only counter-examples can prove the falseness of a statement. That's the eternal problem of induction.

It means it was counter-example, right? Because why on earth we need just examples if they just examples. Anyway my point was that whatever your example was it is subjective. People read / interpret / parse differently, they have different environments (dark mode, white mode, small screen vs big screen, good eyesight vs poor eyesight) etc. 

> And when I say that adding extra spaces make it more difficult to distinguish the spaces inside the strings from the ones outside, I think you're just arguing for the sake of arguing.

I've said "Never mind" - it means that I did not want to argue from the beginning. 

> Yes, people are unique at some point, otherwise how boring, but they're also very similar,

And who decides where this "some point" is? In conversation like this it is just about difference in our personal thresholds. 

> However, yes, you can find people that won't make the difference, but even this fact, we can explain it objectively, because the people who don't see the difference are likely to have defective cones in the retina.

I does not mean that people who have good cones in their retinas( can distinguish the difference) can not ignore the difference as it is something that really make no difference for them. Why don't we argue then about what font we should all use and what size of font? We can try to find objectiveness there as well. Again, I will repeat myself, I think that the most important thing is conventions within particular communities. Every language has its own coding style even though every language seek for optimal readability. We get used to some, stick to it and it makes it convenient for the team to read the code. 

> C'mon, no one uses string concatenation in Python! the Python way is string formatting or join.

Java also have string formatting. Why do we use string concatenation then?

PS. I started to follow your idea about naming of the long test methods in Java with underscores even though it is against Java camel case style convention. I subjectively agreed to this. But if the rest of the team would say me that let's stick to a convention I would follow the majority as it is more important. It is just in this particular case I don't feel like that without spaces it is more convenient to read. And I asked the initial question whether we have new convention or not, that is it.

PR #3911, opened 2019-09-23
PUBDEV-6825: Added ability to make copy of the ModelBuilder

deil87:
This functionality is needed for integration of TE into AutoML. 

We need to search for TE hyperparameters for a given `ModelBuilder`. Current implementation of `trainModelOnH2ONode()` method of the `ModelBuilder` prohibits us from reusing same instance of ModelBuilder hence the need of copying/cloning.

part of the PUBDEV-5742 task ( https://github.com/h2oai/h2o-3/pull/3183)


h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -170,6 +170,29 @@ private static int ensureBuilderIndex(String urlName) {
     return mb;
   }
 
+  /**
+   * Method to create deep copy of ModelBuilder so that it could be trained independently without affecting original builder.
+   */
+  public <B extends ModelBuilder> B makeCopy() {
```
deil87: Not sure what would be the best name for the method. This is not deepCloning as I did not care about all possible copies of objects that might be referencing by ModelBuilder's instance. It is not a shallow cloning as well as I do deep copy of the frames.

h2o-automl/src/test/java/hex/ModelBuilderAutoMLTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+import water.*;
+import water.fvec.Frame;
+
+import java.util.Random;
+
+import static org.junit.Assert.assertEquals;
+
+public class ModelBuilderAutoMLTest extends TestUtil {
```
deil87: I'm testing copy of MB based on the ability to be trained and score same results on test split with arbitrary algo. This is the reason why I could not place test in the core module's tests as I needed algos. I assumed that testing with just arbitrary algo is enough as the main goal is to test ability to create new training job and use copy of the original data. Maybe adding test to `h2o-core` module with just  checking of the data being copied also makes sense. Will rely on reviewers comments.

h2o-core/src/main/java/hex/ModelBuilder.java
seb-h2o: @deil87 can you please provide a concrete example why you would need a copy of the `ModelBuilder` itself?
Why not simply cloning the parameters?
Why deep copying the frames? looks like this could be a quite expensive operation.

deil87: @seb-h2o please read javadoc for `trainModelOnH2ONode` method of MB. 
```
* Warning: the nature of this method prohibits further use of this instance of the model builder after the method
   *          is called.
   *
   * This is intended to reduce training time in client-mode setups, it pushes all computation to a regular H2O node
   * and avoid exchanging data between client and H2O cluster. This also lowers requirements on the H2O client node.
```
Training process leaves some intermediate data in the original frames which makes it dangerous to reuse them. I was relying on the comment's warning when decided to make a copy of the MB itself.  `ModelBuilder` is quite mutable structure.... there are `Countdowns` as you might know.  Also we need a job with new key to be able to submit training with different TE hyperparameters. 

I think that the most expensive operation is a copying of the data. We can probably come up with non-disruptive training method but this on the other hand will impact speed of training(again according to a javadoc). If you have any suggestions I would be glad to discuss them.

seb-h2o: @deil87 My point is that you can create a new `ModelBuilder` just with the `Model.Parameters` object, and you can create as many as you want from the same parameters.
Now given that we know that for now, training may mutate those parameters, you can just clone them, and have a new `ModelBuilder` out if it.
I just don't understand the necessity of another method for that (`new ModelBuilder(params.clone())`.
But here, you also deep clone training and validation frame, which should not be modified by MB afaik, so I'm just wondering about the reason behind this.

deil87: @seb-h2o  There would not be such a Warning in a comment  to the method `trainModelOnH2ONode` if everything was that simple. Please take a look at screenshot that shows `size=-1` for the training frame right after training is finished and also pay attention to the names of the columns.  Screenshot 2019-09-23 at 23 06 17 . 

Frames are being modified by the training process. Again, please see the comment to the method - it is done to `...avoid exchanging data between client and H2O cluster.` And as frames are being modified at some point you will need to deep copy them from parameters. PR could be rewritten from perspective where we don't clone ModelBuilder but clone parameters with deepCopy of the frames - that is actually what I'm doing but I put the code inside ModelBuilder.

michalkurka: `trainModelOnH2ONode` is unrelated IMHO, client mode refers to sparkling water client mode in this context.

deil87: Thank you guys. Was confused by the javadoc. Maybe it does not make sense to introduce this method after all. 
@seb-h2o params stay intact and looks like we don't even need to clone them. Did not find constructor like in your comment as ModelBuilder is an abstract class, but I guess it was just a pseudocode to express an idea.

h2o-automl/src/test/java/hex/ModelBuilderAutoMLTest.java
```diff
@@ -0,0 +1,141 @@
+package hex;
```
seb-h2o: does it make sense to have this test suite in AutoML module?

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -170,6 +170,16 @@ private static int ensureBuilderIndex(String urlName) {
...
+  /**
+   * Factory method to create a ModelBuilder instance from a clone of a given {@code parms} instance of Model.Parameters.
+   */
+  public static <B extends ModelBuilder, MP extends Model.Parameters> B make(MP parms) {
+    Key<Model> mKey = ModelBuilder.defaultKey(parms.algoName());
+    Job<Model> mJob = new Job<>(mKey, parms.javaName(), parms.algoName());
+    B newMB = ModelBuilder.make(parms.algoName(), mJob, mKey);
+    newMB._parms = parms.clone();
+    return newMB;
+  }
```
seb-h2o: looks good to me

h2o-core/src/main/java/hex/Model.java
```diff
@@ -213,6 +213,10 @@ public void setTrain(Key<Frame> train) {
       this._train = train;
     }
 
+    public void setValid(Key<Frame> valid) {
```
michalkurka: Unused

h2o-core/src/main/java/hex/ModelBuilder.java
michalkurka: the method looks fine, it deserves a simple unit test in h2o-core that will:

1. check that created instance is of expected type
2. check that _job, and _result are not null
3. check that parameters a cloned, not the same instance

h2o-core/src/main/java/hex/Model.java
deil87: @michalkurka yeah, I wanted to keep it as it should have been there before.... like `setTrain()`.  Don't want to do a refactoring to make it used on the other hand.

h2o-core/src/main/java/hex/ModelBuilder.java
deil87: @michalkurka Yes, I was thinking about this kind of test that you described. 
One of the reasons why I didn't write tests in h2o-core at once was that we need algos or concrete classes that implements Model.Parameters abstract class( which also live in h2o-algos module).
In `ModelBuilderTest` we have `DummyModelBuilder` and `DummyModelParameters` and everything is dummy. This is maybe fine for a non-static functionality but in my case I need to test `ModelBuilder.make()` static method, which in addition uses static field of `ModelBuilder` -> ALGOBASES which is empty if we don't do registration of algos. 

All these ^^ makes it quite difficult to test.

Also I don't feel like those 3  checks guarantee that the ModelBuilder could be successfully trained. For example _job, and _result are not null even for ModelBuilder that was already trained once. Maybe checking for job ids would be needed. 

Is it convenient to register algos from within the test suite?

h2o-automl/src/test/java/hex/ModelBuilderAutoMLTest.java
deil87: @seb-h2o I also don't like this. That is why I left this comment from the beginning https://github.com/h2oai/h2o-3/pull/3911/files#r327223191 . It's complicated to write test in h2o-core module ( see answer https://github.com/h2oai/h2o-3/pull/3911#discussion_r328227400). And for me it is better to have test in a not the best place than don't have it at all. I will try to do something about it, maybe @michalkurka will suggest me something.

h2o-core/src/main/java/hex/ModelBuilder.java
deil87: @michalkurka I have found a solution - registered DummyModelBuilder through the side effect of the existing method/constructor. This is similar to how we register algos in `RegisterAlgos.java`. Let me know if you have another ideas.

deil87:
@michalkurka Could you please consider to approve this PR as I believe all comments were addressed?

h2o-core/src/test/java/hex/ModelBuilderTest.java
```diff
@@ -121,9 +121,25 @@ public void testMakeUnknownModel() {
...
+      assertTrue( e.getMessage().startsWith("Algorithm 'invalid' is not registered. Available algos: [")); 
     }
   }
+  
+  @Test
+  public void testMakeByModelParameters() {
+    
+    String registrationAlgoName = "dummymodelbuilder"; // by default it is a lower case of the ModelBuilder's simple name
+    DummyModelParameters params = new DummyModelParameters("dummy_MSG", Key.make( "dummyGBM_key"), registrationAlgoName);
+    new DummyModelBuilder(params, true); // as a side effect DummyModelBuilder will be registered in a static field ModelBuilder.ALGOBASES
```
michalkurka: I would register this in `@BeforeClass` just to make sure this particular test method doesn't have side effects.

Otherwise, it looks good to me.

michalkurka:
@deil87 thanks for addressing the comments, LGTM

(minor inline suggestion posted)

h2o-core/src/test/java/hex/ModelBuilderTest.java
michalkurka: Fine with me, this is a minor thing.

PR #3913, opened 2019-09-24
PUBDEV-6903: Add support for SPNEGO to h2odriver

honzasterba:


scripts/jenkins/groovy/kerberosCommands.groovy
```diff
@@ -7,15 +7,24 @@
...
         case H2O_HADOOP_STARTUP_MODE_STANDALONE:
             return getCommandStandalone(stageConfig)
         default:
             error("Startup mode ${stageConfig.customData.mode} for H2O with Hadoop is not supported")
     }
 }
 
-private GString getCommandHadoop(final stageConfig) {
+private GString getCommandHadoop(final stageConfig, final spnegoAuth) {
+    loginArgs = ""
```
michal-raska: I think there's a missing `def`

scripts/jenkins/groovy/hadoopStage.groovy
```diff
@@ -53,6 +54,7 @@ def call(final pipelineContext, final stageConfig) {
 private String getMakeTargetSuffix(final stageConfig) {
     switch (stageConfig.customData.mode) {
         case H2O_HADOOP_STARTUP_MODE_HADOOP:
+        case H2O_HADOOP_STARTUP_MODE_HADOOP_SPNEGO:
```
michal-raska: Could we remove the fallthrough please?

scripts/jenkins/groovy/kerberosCommands.groovy
honzasterba: thanks, also the if was broken :)

scripts/jenkins/groovy/hadoopStage.groovy
honzasterba: fixed

Tagar:
Nice.

If we enable SPNEGO authentication, would it be possible to whitelist only certain users, or perhaps use an LDAP query to get list of their Active Directory groups, and then only whitelist only certain groups for authorization. 

Thank you

honzasterba:
Yes, I believe all that should be possible, you assign a service principal to the H2O cluster and then within your kerberos kdc you can define which users are allowed to access given service.

Tagar:
Thank you @honzasterba

PR #3916, opened 2019-09-24
PUBDEV-6906: Update gen_gbm Examples

hannah-tillman:


h2o-bindings/bin/custom/python/gen_gbm.py
```diff
@@ -54,15 +54,15 @@ def update_param(name, param):
 ...                training_frame=cars)
 >>> cars_gbm.auc()
 """,
-    keep_cross_validation_models="""
+   keep_cross_validation_models="""
```
angela0xdata: This is now indented 3 spaces instead of 4. All others are indented 4 spaces.

PR #3918, opened 2019-09-24
Moved R deeplearing mojo test to NOPASS.

wendycwong:
A JIRA was opened to resolve this failure: https://0xdata.atlassian.net/browse/PUBDEV-6680?filter=-1


michalkurka:
@wendycwong why is it failing?

wendycwong:
@michalk:

It is failing because there is a bug in the deeplearning mojo dealing with categorical columns.  They do it differently in deeplearning compared to all the other algos.  I have moved the python test away but forgot about the R test.

PR #3919, opened 2019-09-25
PUBDEV-6749 - Allow Grid Search to Continue After Cluster Restart

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6749

Goal of this PR is to introduce a way for users to resume once-saved Grids. Even after a cluster wipe or complete cluster restart. This means saving and loading the related models as well.

## New API

Python method definitions: https://github.com/h2oai/h2o-3/pull/3919/files#diff-cbbd4c777a0e751240db11aaea1722de

R method definitions: https://github.com/h2oai/h2o-3/pull/3919/files#diff-eda7d39913d5a8ee1048abd909b10ee3

### Loading the Grid
Core of the new functionality is a new function to load a saved Grid

**Python**
```python
grid = h2o.load_grid(grid_directory="/home/pavel/example, grid_id=grid.grid_id)
```
**R**
```R
  grid <- h2o.load_grid(grid_directory = tempdir(), grid_id = grid_id)
```

Two arguments are supplied: The folder with the saved grid and the ID of the Grid.

**Important note:** With the grid, models are loaded as well. Models are looked for in the very same folder.

### Saving the Grid

There are two modes
- Use auto-checkpointing and supply `export_checkpoints_dir` parameter
- Call new function `h2o.save_grid` for manual export

Both available in Py & R of course.

**R checkponting example** (full example in the R test: https://github.com/h2oai/h2o-3/pull/3919/files#diff-2beac8db59bc28f335f3b14691dcdfb1)
```R
  baseline_grid <- h2o.grid("gbm", grid_id="gbm_grid_test", x=1:4, y=5, training_frame=iris.hex, hyper_params = hyper_parameters, export_checkpoints_dir = tempdir())
grid_id <- baseline@grid_id
  grid <- h2o.load_grid(grid_directory = tempdir(), grid_id = grid_id)
```

**Python checkpointing example** ( full example in the test https://github.com/h2oai/h2o-3/pull/3919/files#diff-db0bffe7d30a708c65be5c3e6c17acbe)
```Python
    gs = H2OGridSearch(H2OGradientBoostingEstimator, hyper_params=hyper_parameters,
                       export_checkpoints_dir=export_dir)
    grid_id = gs.grid_id
    grid = h2o.load_grid(grid_directory=export_dir, grid_id=grid_id)
```
**R manual export example** (full example in the R test: https://github.com/h2oai/h2o-3/pull/3919/files#diff-a2aa48625f3059c164984a08eb254db7)
```R
  baseline_grid <- h2o.grid("gbm", grid_id="gbm_grid_test", x=1:4, y=5, training_frame=iris.hex, hyper_params = hyper_parameters)
  grid_id <- baseline_grid@grid_id
  h2o.save_grid(grid_directory = tempdir(), grid_id = grid_id)
  grid <- h2o.load_grid(grid_directory = tempdir(), grid_id = grid_id)
```

**Python manual export example** (full example in the corresponding test https://github.com/h2oai/h2o-3/pull/3919/files#diff-4e7f8642d6b625956071fe607e584f6e)
```Python
    gs = H2OGridSearch(H2OGradientBoostingEstimator,
                       hyper_params=hyper_parameters)
    gs.train(x=list(range(4)), y=4, training_frame=train)
    grid_id = gs.grid_id
    h2o.save_grid(export_dir, grid_id)
    grid = h2o.load_grid(grid_directory=export_dir, grid_id=grid_id)
```


Pscheidl:
test

Pscheidl:
In the `Grid` class, there is:
```java
  /**
   * Write out K/V pairs
   */
  @Override
  protected AutoBuffer writeAll_impl(AutoBuffer ab) {
    for (Key<Model> k : _models.values())
      ab.putKey(k);
    return super.writeAll_impl(ab);
  }
```
And the `putKey` method looks like this:
```java
  // Write Key's target IFF the Key is not null; target can be null.
  public AutoBuffer putKey(Key k) {
    if( k==null ) return this;    // Key is null ==> write nothing
    Keyed kd = DKV.getGet(k);
    put(kd);
    return kd == null ? this : kd.writeAll_impl(this);
  }
```

This piece of code is important:
```java
    Keyed kd = DKV.getGet(k);
    put(kd);
```

Which means models were serialized together with Grid, as well as serialized besides it. I kept the original behavior and added a new method named `writeWithoutModels`:)

Pscheidl:
@ledell @michalkurka @honzasterba and possibly others, please do the API review. I'll be glad for general code review as well.

At this point, there is something fishy going on and it seems like the autoechkpointing variant of the tests creates more models than expected. There are 2*3 hyperparameter options, resulting in 6 models. With the manual export, there should also be 6 models, as there hyperspace is fully depleted and "explored". However, 6 more are added.

Because of this, the final part of the python and R contains code that reproduces this issue. I'm looking into it, but it should not affect the rest of the code and I wanted to give you an overview of the API as soon as possible to re-do what needs to be re-done early.

Thank you for your reviews !

karthikkannappan:
Does this still assume that all models in the (completed) grid fit in memory after a new cluster comes up? There's a customer use case where they wanted to train more models than would fit in typical cluster memory, so they were looking at the possibility of something like this to essentially push some models to disk and only retain their scores, for example - then if you found a model you really liked, you have that on disk anyway. Their use case involves training say, 100 models where only ~ 15 fit in memory, and essentially they want to push those 15 to disk and continue on with training the rest in cluster memory.

michalkurka:
@karthikkannappan they need to fit in memory

for you use case, you can create a new feature request for that customer

h2o-py/h2o/h2o.py
```diff
@@ -434,6 +434,32 @@ def import_file(path=None, destination_frame=None, parse=True, header=0, sep=Non
...
+    """
+    Export a Grid and it's all its models into the given folder
+    :param grid_directory: A string containing the path to the folder for the grid to be saved to.
+    :param grid_id: A chracter string with identification of the Grid in H2O. 
+    """
+    assert_is_type(grid_directory, str)
+    assert_is_type(grid_id, str)
+    response = api("POST /3/Grid/export", {"grid_directory": grid_directory, "grid_id": grid_id})
+    return get_grid(response["name"])
+
```
seb-h2o: are those methods the equivalent for grids as those existing ones for model?
- `save_model`
- `load_model`
If so, it would be nice to have a symmetric naming, and same signature:
`save_grid(grid, path)` returning the full path 
and `load_grid(path)`

h2o-r/h2o-package/R/grid.R
```diff
@@ -228,3 +234,90 @@ h2o.getGrid <- function(grid_id, sort_by, decreasing) {
       summary_table     = json$summary_table
       )
 }
+
```
seb-h2o: same comment as for Python.
R has 2 modules `import` and `export`, those functions are probably more suitable there for consistency.
Also, it would prefer having naming and signature symmetric with models: `h2o.loadGrid`, `h2o.saveGrid`.

h2o-core/src/main/java/water/api/RegisterV3Api.java
```diff
@@ -464,6 +464,14 @@ public void registerEndPoints(RestApiContext context) {
...
             "GET /3/Capabilities/API", CapabilitiesHandler.class, "listRest",
             "List of all registered Rest API capabilities");
+
+    context.registerEndpoint("import_grid_search",
+            "POST /3/Grid/import", GridImportExportHandler.class, "importGrid",
+            "Import previously saved grid model");
+
+    context.registerEndpoint("export_grid_search",
+            "POST /3/Grid/export", GridImportExportHandler.class, "exportGrid",
+            "Export a Grid and its models.");
```
seb-h2o: also, for those building on top on the REST API, would be nice to align with models here as well:
```
    // Model serialization - import/export calls
    context.registerEndpoint("importModel",
            "POST /99/Models.bin/{model_id}", ModelsHandler.class, "importModel",
            "Import given binary model into H2O.");

    context.registerEndpoint("exportModel",
            "GET /99/Models.bin/{model_id}", ModelsHandler.class, "exportModel",
            "Export given model.");
```
I have difficulties to understand the GET for export, but for the rest we could use similar URLs.

seb-h2o: @michalkurka any plan to officialise those "experimental" `/99/` APIs to `/3/`?
Also asking because AutoML is also using `/99/`: we could duplicate the handlers for the lifetime of a major release, and switch all clients to use `/3/`.

h2o-py/h2o/h2o.py
Pscheidl: Renamed the methods, will deal with the signature later.

h2o-r/h2o-package/R/grid.R
Pscheidl: Renamed the methods for now. Will deal with the signature later.

@ledell Do we want the tidyverse version, or the "traditional H2O" with camelCase ?

h2o-core/src/main/java/water/api/RegisterV3Api.java
michalkurka: Introducing /99/ shouldn't have been done. However, if we move it - we will certainly break something for people who use portions of the API directly. There was a jira to move everything out of /99/ it but it is not something is a priority right now.

I think from AutoML point of view - I would just move it.

All new stuff should go to /3/.

Pscheidl: Using {grid_id} is definitely a good idea.

Pscheidl:
Note to self: Provide export information to stdout.

h2o-core/src/main/java/water/api/GridImportExportHandler.java
```diff
@@ -0,0 +1,91 @@
...
+    final Persist persist = H2O.getPM().getPersistForURI(gridUri);
+    try (final InputStream inputStream = persist.open(gridUri.toString())) {
+      final AutoBuffer gridAutoBuffer = new AutoBuffer(inputStream);
+      final Freezable freezable = gridAutoBuffer.get();
+      if (!(freezable instanceof Grid)) {
+        throw new IllegalArgumentException(String.format("Given file '%s' is not a Grid", gridImportV3.grid_directory));
+      }
+      final Grid grid = (Grid) freezable;
+
+      loadGridModels(grid,gridImportV3);
```
honzasterba: missing space

h2o-core/src/main/java/water/api/GridImportExportHandler.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @SuppressWarnings("unused")
+  public KeyV3.GridKeyV3 exportGrid(final int version, final GridImportExportV3 gridExportV3) throws IOException {
+    validateGridImportExportV3(gridExportV3);
+    if(DKV.get(gridExportV3.grid_id) == null){
+      throw new IllegalArgumentException(String.format("Grid with id '%s' has not been found.", gridExportV3.grid_id));
+    }
+    final Iced possibleGrid = DKV.get(gridExportV3.grid_id).get();
+    assert possibleGrid instanceof Grid;
```
honzasterba: there is no guarantee that the ID given by the user points to a grid so this should not be an assert but a regular if check

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -139,6 +138,29 @@ private GridSearch(Key<Grid> gkey, HyperSpaceWalker<MP, ?> hyperSpaceWalker) {
...
+       * If export_checkpoints_dir is defined in parameters, saves the whole Grid into that
+       * folder with checkpoints.
+       *
+       * @param caller the task invoking this method (which may
+       */
+      @Override
+      public void onCompletion(CountedCompleter caller) {
+        final MP params = _hyperSpaceWalker.getParams();
+        if (params._export_checkpoints_dir != null) {
+          final String gridPath = params._export_checkpoints_dir + "/" + grid._key;
```
honzasterba: since export binary is a method of Grid I thing it would be cleaner just to pass the dir and leave the attaching + grid._key for the method, it would also reduce code duplication

h2o-core/src/main/java/water/api/GridImportExportHandler.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Basic sanity check for given Grid parameters (both import and export).
+   *
+   * @param input An instance of {@link GridImportExportV3}
+   */
+  private void validateGridImportExportV3(final GridImportExportV3 input) {
+    Objects.requireNonNull(input);
+    if (input.grid_directory == null || input.grid_directory.isEmpty()) {
+      throw new IllegalArgumentException(String.format("Given grid directory '%s' is not a valid directory.",
```
honzasterba: this can be accomblished by required=true in annotation can it?

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -139,6 +138,29 @@ private GridSearch(Key<Grid> gkey, HyperSpaceWalker<MP, ?> hyperSpaceWalker) {
...
       }
+
+      /**
+       * If export_checkpoints_dir is defined in parameters, saves the whole Grid into that
+       * folder with checkpoints.
+       *
+       * @param caller the task invoking this method (which may
+       */
+      @Override
+      public void onCompletion(CountedCompleter caller) {
```
honzasterba: should we maybe save the grid after every succeful model train? I think that would be best

h2o-core/src/main/java/water/api/GridImportExportHandler.java
Pscheidl: The annotation is even there. :roll_eyes:

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Done. Thanks.

h2o-core/src/main/java/water/api/GridImportExportHandler.java
Pscheidl: Yes, that should be done. And it was. Thank you.

h2o-core/src/main/java/water/api/GridImportExportHandler.java
Pscheidl: Done. Thanks.

h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -475,4 +492,35 @@ public TwoDimTable createScoringHistoryTable() {
...
     return ScoringInfo.createScoringHistoryTable(_scoring_infos, (scoring_info != null ? scoring_info.validation : false), (scoring_info != null ? scoring_info.cross_validation: false), m._output.getModelCategory(), (scoring_info != null ? scoring_info.is_autoencoder : false));
   }
+
+  /**
+   * Exports this Grid in a binary format using {@link AutoBuffer}. Related models are not saved.
+   *
+   * @param gridFilePath Full path to the name of the file this {@link Grid} should be saved to
+   * @throws IOException Error serializing the grid.
+   */
+  public void export_binary(final String gridFilePath) throws IOException {
```
honzasterba: I think this should take just the parent dir as argument, since the path is always DIR + "/" + this.id

h2o-core/src/main/java/hex/grid/GridSearch.java
honzasterba: where? I cant find it

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -245,6 +245,17 @@ private void gridSearch(Grid<MP> grid) {
           _job.update(1);
           // Always update grid in DKV after model building attempt
           grid.update(_job);
+          final MP generalParams = _hyperSpaceWalker.getParams();
+          if (generalParams._export_checkpoints_dir != null) {
+            final String gridPath = generalParams._export_checkpoints_dir + "/" + grid._key;
+            try {
+              grid.exportBinary(gridPath);
+              // Models are exported automatically when export_checkpoints_dir is defined
+            } catch (IOException e) {
```
honzasterba: is this try-catch necessary here? since we are basically ignoring the exception what is the point of throwing it?

Pscheidl: ## My POV
From the point of `exportBinary` method, it gives the caller more freedom to decide what to do when the export fails. But with power comes great responsibility to catch the exceptions and treat them "properly".

In this case, we only log something went bad, because taking down the whole grid only because it could not be exported is not appropriate. Maybe where will be other handlings in future. 

## Conclusion
As there is no other use case for it, I adopted your logic and removed the try catch. The `exportBinary` method now logs this itself.

Pscheidl: From the point of exportBinary method, it gives the caller more freedom to decide what to do when the export fails. But with power comes great responsibility to catch the exceptions and treat them "properly".

In this case, we only log something went bad, because taking down the whole grid only because it could not be exported is not appropriate. Maybe where will be other handlings in future.

It's also used in `exportGrid` in `GridImportExportHandler` - which doesn't catch the exception, but rather propagates it all the way to the client.

h2o-core/src/main/java/hex/grid/Grid.java
Pscheidl: Provides less control for naming, but not required now. Did as you suggested.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Was not pushed yet. It's pushed now. Sorry.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Done.

h2o-py/h2o/h2o.py
Pscheidl: Done. save_grid now returns the full path usable later in `load_grid`.

h2o-r/h2o-package/R/grid.R
Pscheidl: Done. Re-aligned with older methods using camelCase, even though we're only doing snake_case now. Makes sense to me - would be too much entropy for the users.

Moved to import/export. Changed the examples accordingly.

h2o-core/src/main/java/water/api/RegisterV3Api.java
Pscheidl: In-url {grid_id} param for export done.

Pscheidl:
Thank you for your comments. Please re-review @honzasterba @seb-h2o @ledell @michalkurka .

Addressed all your comments and did the necessary changes.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -258,7 +272,8 @@ private void gridSearch(Grid<MP> grid) {
       grid.unlock(_job);
     }
   }
-
+  
+  private static final Set<String> IGNORED_FIELDS_PARAM_HASH = new HashSet(Arrays.asList("_export_checkpoints_dir"));
```
honzasterba: you seem to want this constant, in which cas I would use Collections.immutableSet

Pscheidl: Good advice. Made it so.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -258,7 +270,8 @@ private void gridSearch(Grid<MP> grid) {
       grid.unlock(_job);
     }
   }
-
+  
+  private static final Set<String> IGNORED_FIELDS_PARAM_HASH = Collections.unmodifiableSet(new HashSet(Arrays.asList("_export_checkpoints_dir")));
```
michalkurka: Collections.singleton("_export_checkpoints_dir")

h2o-core/src/main/java/water/api/RegisterV3Api.java
```diff
@@ -464,6 +464,14 @@ public void registerEndPoints(RestApiContext context) {
     context.registerEndpoint("list_rest_capabilities",
             "GET /3/Capabilities/API", CapabilitiesHandler.class, "listRest",
             "List of all registered Rest API capabilities");
+
+    context.registerEndpoint("import_grid_search",
```
michalkurka: import_grid_search or import_grid?

h2o-py/tests/testdir_algos/grid/pyunit_grid_manual_export_resume_new_hyperspace.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Run GBM Grid Search
+    ntrees_opts = [1, 3]
+    learn_rate_opts = [0.1, .05]
+    hyper_parameters = OrderedDict()
+    hyper_parameters["learn_rate"] = learn_rate_opts
+    hyper_parameters["ntrees"] = ntrees_opts
+    print("GBM grid with the following hyper_parameters:", hyper_parameters)
+    
+    export_dir = tempfile.mkdtemp()
```
michalkurka: files shouldn't be created in temp directory - use test sandbox directory instead

h2o-py/tests/testdir_algos/grid/pyunit_grid_resume.py
```diff
@@ -0,0 +1,48 @@
...
+
+def grid_resume():
+    train = h2o.import_file(path=pyunit_utils.locate("smalldata/iris/iris_wheader.csv"))
+    # Run GBM Grid Search
+    ntrees_opts = [1,5]
+    hyper_parameters = OrderedDict()
+    hyper_parameters["ntrees"] = ntrees_opts
+    print("GBM grid with the following hyper_parameters:", hyper_parameters)
+
+    export_dir = tempfile.mkdtemp()
```
michalkurka: same as above

h2o-r/tests/testdir_algos/grid/runit_grid_manual_export_resume.R
```diff
@@ -0,0 +1,43 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../../scripts/h2o-r-test-setup.R")
+
+
+
+test.grid.resume <- function() {
+  iris.hex <- h2o.uploadFile(locate("smalldata/iris/iris.csv"), destination_frame="iris.hex")
```
michalkurka: please use importFile instead

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Done.

h2o-core/src/main/java/water/api/RegisterV3Api.java
Pscheidl: import_grid definíitely

Pscheidl: Also did the change for export_grid. Done.

h2o-py/tests/testdir_algos/grid/pyunit_grid_manual_export_resume_new_hyperspace.py
Pscheidl: Swapped for `pyunit_utils.locate("results")`. Thanks.

h2o-py/tests/testdir_algos/grid/pyunit_grid_resume.py
Pscheidl: Swapped for pyunit_utils.locate("results"). Thanks.

h2o-r/tests/testdir_algos/grid/runit_grid_manual_export_resume.R
Pscheidl: Two tests affected. Replaced. Thank you.

h2o-core/src/main/java/water/api/RegisterV3Api.java
seb-h2o: @Pscheidl this is binary import, right?
I we specify this in the URL (like with `/Models.bin/` vs. `/Models.mojo/`), I think you'll have more flexibility in the future if for example we want to have a feature allowing to export all grid models as mojo at once: you'll then be able to use `/Grid.bin/{grid_id}/export` vs. `/Grid.mojos/{grid_id}/export` and possibly support future formats in a consistent way.
Just my suggestion.

Pscheidl: Yes, this looks reasonably good.

Pscheidl:
There was some amount of test failure due to unadjusted asserts in old grid tests with autosaving features.

Pscheidl:
Did the adjustments as @seb-h2o pointed out to the REST API. Now waiting for tests.

Pscheidl:
Had to fix another two failing grid tests. True failures caused by this PR - the asserts measured number of files exported to a folder when autoexporting is active. And now, there is one more file :)

PR #3922, opened 2019-09-25
PUBDEV-6780: Add equivalent of Pandas Melt

michalkurka:
https://0xdata.atlassian.net/browse/PUBDEV-6780

### DIFF
--- a/h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstMelt.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstMelt.java
@@ -1,0 +1,99 @@
+package water.rapids.ast.prims.mungers;
+
+import water.MRTask;
+import water.fvec.*;
+import water.rapids.Val;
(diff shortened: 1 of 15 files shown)


h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstMelt.java
```diff
@@ -0,0 +1,99 @@
...
+import water.rapids.Val;
+import water.rapids.ast.AstBuiltin;
+import water.rapids.vals.ValFrame;
+import water.util.ArrayUtils;
+
+public class AstMelt extends AstBuiltin<AstPivot> {
+
+  @Override
+  public String[] args() {
+    return new String[]{"frame", "id_vars", "value_vars", "var_name", "value_name", "skip_na"};
```
Pscheidl: Could be a constant, but hey, let's not keep it in memory. if it was a constant, it could be used in nargs. Ok, this is actually better.

PR #3923, opened 2019-09-26
PUBDEV-6829: sort bug for multiple nodes

wendycwong:
This PR fixed the bug reported in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6829?filter=-1.

A dataset was found to sort to completion with one node but with 2 nodes, it will hang.

The reasons for this hanging are:
1.  As the number of nodes increases, the number of chunks also increase.
2. Original implementation will walk through all rows per frame more than 4 times in order to determine which chunk index that row belongs to and which H2O node is that chunk located on.

I fixed this hanging by writing my own new class SortCombine to only walk through the whole frames twice.  This seems to solve the issue.  I have generated a small dataset which will hang with the original implementation and but will not hang with my fix.

The last thing I will do is to checkout the sort run time with the original customer dataset with 1 to 8 nodes.


wendycwong:
@michalk:  I will need to clean up the BinaryMerge.java at some point.  However, it is very messy to touch right now.

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -70,6 +69,8 @@ public static Frame merge(final Frame leftFrame, final Frame riteFrame, final in
   // single-threaded driver logic.  Merge left and right frames based on common columns.
   public static Frame merge(final Frame leftFrame, final Frame riteFrame, final int leftCols[], final int riteCols[],
                             boolean allLeft, int[][] id_maps, int[] ascendingL, int[] ascendingR) {
+    if (allLeft)
+      return sortOnly(leftFrame,  leftCols, id_maps, ascendingL);
     final boolean hasRite = riteCols.length > 0;
```
wendycwong: diverge to my own method if sorting is performed.

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -340,7 +341,121 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
     return fr;
   }
 
-  private static RadixOrder createIndex(boolean isLeft, Frame fr, int[] cols, int[][] id_maps, int[] ascending) {
+  public static Frame sortOnly(final Frame leftFrame, final int leftCols[], int[][] id_maps, int[] ascendingL) {
+
+    createIndex(true ,leftFrame,leftCols,id_maps, ascendingL);  // sort the columns.
+    System.out.print("Making BinaryMerge RPC calls ... ");
```
wendycwong: createIndex already has fixed sorting based on the sorted columns.  All we need to do next is to collect the rows into the sorted rows discovered by createIndex.

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -340,7 +341,121 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+    long t0 = System.nanoTime();
+    ArrayList<SortCombine> bmList = new ArrayList<>();
+    Futures fs = new Futures();
+
+      for (int leftMSB=0; leftMSB<=255; leftMSB++) {  // only add to RPC call if MSB is not empty
+        SingleThreadRadixOrder.OXHeader leftSortedOXHeader = DKV.getGet(getSortedOXHeaderKey(/*left=*/true, leftMSB));
```
wendycwong: To perform merging, one RPC is used per MSB.  There will be 256 RPCs generated here.

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -340,7 +341,121 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+    }
+    
+    Key<Vec> key = Vec.newKey();
+    Vec[] vecs = new Vec(key, Vec.ESPC.rowLayout(key, espc)).makeCons(numColsInResult, 0, doms, types);
+    System.out.println("took: " + (System.nanoTime() - t0) / 1e9);
+
+    System.out.print("Finally stitch together by overwriting dummies ...");
+    t0 = System.nanoTime();
+    Frame fr = new Frame(names, vecs);
+    ChunkStitcher ff = new ChunkStitcher(chunkSizes, chunkLeftMSB, chunkRightMSB, chunkBatch);
```
wendycwong: ChunkStitcher will stitch together Chunks formed for all MSBs into one big frame.

h2o-core/src/main/java/water/rapids/SortCombine.java
```diff
@@ -0,0 +1,407 @@
...
+  private void setPerNodeNumsToFetch() {
+    Vec anyVec = _leftSB._frame.anyVec();
+    int nbatch = _leftKO._order.length;
+    _numRowsPerCidx = new int[nbatch][anyVec.nChunks()];
+    for (int batchNum =0; batchNum < nbatch; batchNum++) {
+      int batchSize = _leftKO._order[batchNum].length;
+      for (int index=0; index < batchSize; index++) {
+        long globalRowNumber = _leftKO._order[batchNum][index];
+        int chkIdx = _leftSB._vec.elem2ChunkIdx(globalRowNumber);
+        _leftKO._perNodeNumRowsToFetch[_leftSB._chunkNode[chkIdx]]++;
```
wendycwong: we go over the whole dataframe once here in order to discuss for each batch, how many rows per chunks need to be fetched to form the final sorted chunks for that MSB

h2o-core/src/main/java/water/rapids/SortCombine.java
```diff
@@ -0,0 +1,407 @@
...
+    for (int nbatch=0; nbatch < numBatch; nbatch++) {
+      int sortedRowIndex = -1;
+      chkIndices[nbatch] = new int[_chunkNum];
+      int batchSize = _leftKO._order[nbatch].length;
+      for (int batchNum=0; batchNum < batchSize; batchNum++) {
+        sortedRowIndex++;
+        long row = _leftKO._order[nbatch][batchNum];
+        int chkIdx =  _leftSB._vec.elem2ChunkIdx(row); //binary search in espc
+        perNodeLeftRowsCidx[nbatch][chkIdx][chkIndices[nbatch][chkIdx]] = row;
+        perNodeLeftIndices[nbatch][chkIdx][chkIndices[nbatch][chkIdx]] = sortedRowIndex;
```
wendycwong: This is the second time we go over the whole dataframe to figure out for each batch, and for each chunk, what the global row number of the row (perNodeLeftRowsCidx) should be and in what order (perNodeLeftIndices) should we put it back together.

h2o-core/src/main/java/water/rapids/SortCombine.java
```diff
@@ -0,0 +1,407 @@
...
+        frameLikeChunksLong[col][b] = null; //free mem as early as possible (it's now in the store)
+      } else {
+        Chunk ck = new NewChunk(frameLikeChunks[col][b]).compress();
+        DKV.put(BinaryMerge.getKeyForMSBComboPerCol(_leftSB._msb, -1, col, b), ck, fs, true);
+        frameLikeChunks[col][b] = null; //free mem as early as possible (it's now in the store)
+      }
+    }
+    fs.blockForPending();
+  }
+
```
wendycwong: This class will collect all sorted rows for each H2O node.

wendycwong:
image

wendycwong:
image

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -340,7 +341,121 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+      types[j] = leftFrame.vec(j).get_type();
+      doms[j] = leftFrame.domains()[j];
+      names[j] = leftFrame.names()[j];
+    }
+    
+    Key<Vec> key = Vec.newKey();
+    Vec[] vecs = new Vec(key, Vec.ESPC.rowLayout(key, espc)).makeCons(numColsInResult, 0, doms, types);
+    System.out.println("took: " + (System.nanoTime() - t0) / 1e9);
+
+    System.out.print("Finally stitch together by overwriting dummies ...");
```
Pscheidl: `water.util.Log`

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -340,7 +341,121 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+        }
+      }
+      
+    System.out.println("took: " + String.format("%.3f", (System.nanoTime() - t0) / 1e9));
+
+    t0 = System.nanoTime();
+    System.out.println("Sending BinaryMerge async RPC calls in a queue ... ");
+    fs.blockForPending();
+
+    System.out.println("took: " + (System.nanoTime() - t0) / 1e9);
```
Pscheidl: Seconds, millis ? The user's won't know It's nanos divides by 1e9.

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Good idea.  This will speed things up even more.  Done.

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: I added the seconds unit here to clarify.  Thanks.

h2o-core/src/main/java/water/rapids/SortCombine.java
```diff
@@ -0,0 +1,407 @@
...
+        int chkIdx = _leftSB._vec.elem2ChunkIdx(globalRowNumber);
+        _leftKO._perNodeNumRowsToFetch[_leftSB._chunkNode[chkIdx]]++;
+        _numRowsPerCidx[batchNum][chkIdx]++;  // number of rows to fetch per Cidx for a certain MSB
+      }
+    }
+  }
+
+  // Holder for Key & Order info
+  private static class KeyOrder {
+    public final transient long _batchSize;
```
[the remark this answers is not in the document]
wendycwong: You are right, they are all the same size and they are not adaptive anyway.

h2o-core/src/main/java/water/rapids/Merge.java
honzasterba: use Log.

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -327,26 +329,140 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
-    System.out.println("took: " + (System.nanoTime() - t0) / 1e9);
+    Log.info("took: " + (System.nanoTime() - t0) / 1e9+" seconds.");
 
     System.out.print("Finally stitch together by overwriting dummies ...");
     t0 = System.nanoTime();
     Frame fr = new Frame(names, vecs);
     ChunkStitcher ff = new ChunkStitcher(chunkSizes, chunkLeftMSB, chunkRightMSB, chunkBatch);
     ff.doAll(fr);
-    System.out.println("took: " + (System.nanoTime() - t0) / 1e9);
+    Log.info("took: " + (System.nanoTime() - t0) / 1e9+" seconds");
```
honzasterba: are all these time infos important to the user? maybe some of them should be just DEBUG

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -327,26 +329,140 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
 
     //Merge.cleanUp();
     return fr;
   }
 
-  private static RadixOrder createIndex(boolean isLeft, Frame fr, int[] cols, int[][] id_maps, int[] ascending) {
-    System.out.println("\nCreating "+(isLeft ? "left" : "right")+" index ...");
+  public static Frame sortOnly(final Frame leftFrame, final int leftCols[], int[][] id_maps, int[] ascendingL) {
```
honzasterba: this method is too long and impossible to understand
still some Sys.outs left

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -3306,6 +3306,54 @@ def check_sorted_2_columns(frame1, sorted_column_indices, prob=0.5, ascending=[T
                                                                                              "row {2}: {3}".format(rowInd, frame1[rowInd,colInd],
                                                                                                                    rowInd+1, frame1[rowInd+1,colInd])
 
+def check_sorted_2_columns(frame1, sorted_column_indices, prob=0.5, ascending=[True, True]):
+    for colInd in sorted_column_indices:
+        for rowInd in range(0, frame1.nrow-1):
+            if (random.uniform(0.0,1.0) < prob):
```
honzasterba: I really really dont like this patter, if something is wrong this test will fail only half of the time

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -3306,6 +3306,54 @@ def check_sorted_2_columns(frame1, sorted_column_indices, prob=0.5, ascending=[T
                                                                                              "row {2}: {3}".format(rowInd, frame1[rowInd,colInd],
                                                                                                                    rowInd+1, frame1[rowInd+1,colInd])
 
+def check_sorted_2_columns(frame1, sorted_column_indices, prob=0.5, ascending=[True, True]):
```
honzasterba: this does not seem to be used, should it be removed?

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: I think you have a point

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -327,26 +329,140 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+    ChunkStitcher ff = new ChunkStitcher(chunkSizes, chunkLeftMSB, chunkRightMSB, chunkBatch);
+    ff.doAll(fr);
+    Log.info("took: " + (System.nanoTime() - t0) / 1e9+" seconds.");
+    
+    return fr;
+  }
+
+
+    private static RadixOrder createIndex(boolean isLeft, Frame fr, int[] cols, int[][] id_maps, int[] ascending) {
+    Log.info("\nCreating "+(isLeft ? "left" : "right")+" index ...");
```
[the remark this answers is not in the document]
wendycwong: fixed

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Let me add more comments and break it up if possible.

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: I left the ones telling user what the system is doing as info but the speed info as debug.

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: Honza:  You can set prob=1.0 and this test will check row.  I am doing this to save time.  Copied this from Arno originally.

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: It is used here: h2o-py/tests/testdir_munging/pyunit_pubdev_4870_sort_bug_pubdev_4404_desc.py

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Sorry that I have small eyes and miss things like that!

h2o-core/src/main/java/water/rapids/SortCombine.java
```diff
@@ -0,0 +1,407 @@
...
+import water.fvec.Vec;
+import water.parser.BufferedString;
+
+import java.util.Arrays;
+
+/**
+ * Before this class is called, sorting has been completed on the sorted columns.  The job of this class is to
+ * gather the sorted rows of one MSB into chunks which will later be stitched together to form the whole frame.
+ */
+class SortCombine extends DTask<SortCombine> {
```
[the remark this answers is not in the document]
wendycwong: Very funny Honza!  GRRRRRRRRRR

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: It is used in this test: h2o-py/tests/testdir_munging/pyunit_pubdev_4870_sort_bug_pubdev_4404_desc.py

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -327,26 +329,145 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+    Log.info("Making BinaryMerge RPC calls ... ");
+    ArrayList<SortCombine> bmList = gatherSameMSBRows(leftFrame); // For each MSB, gather sorted rows with same MSB into one spot 
+    
+    Log.info("Allocating and populating chunk info (e.g. size and batch number) ...");
+    return allocatePopulateChunk(bmList, leftFrame);
+
+  }
+
+
+    private static RadixOrder createIndex(boolean isLeft, Frame fr, int[] cols, int[][] id_maps, int[] ascending) {
```
honzasterba: formating

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -327,26 +329,145 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
 
     //Merge.cleanUp();
     return fr;
   }
 
-  private static RadixOrder createIndex(boolean isLeft, Frame fr, int[] cols, int[][] id_maps, int[] ascending) {
-    System.out.println("\nCreating "+(isLeft ? "left" : "right")+" index ...");
+  public static ArrayList<SortCombine> gatherSameMSBRows(Frame leftFrame) {
+    long t0 = System.nanoTime();
+    ArrayList<SortCombine> bmList = new ArrayList<>();
```
honzasterba: knitpicking: but you should use interfaces for collections when possible (e.g. List<SortCombine> bmList here), there are few more cases of this in this PR

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -327,26 +329,145 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+        chunkSizes[k] = thisChunkSizes[j];
+        chunkLeftMSB [k] = thisbm._leftSB._msb;
+        chunkRightMSB[k] = -1;
+        chunkBatch[k] = j;
+        k++;
+      }
+    }
+    Log.debug("took: " + (System.nanoTime() - t0) / 1e9 +" seconds.");
+
+    // Now we can stitch together the final frame from the raw chunks that were
```
honzasterba: this comment is a sign that the method should be split here

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -327,26 +329,145 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+
+    Log.info("Finally stitch together by overwriting dummies ...");
+    t0 = System.nanoTime();
+    Frame fr = new Frame(names, vecs);
+    ChunkStitcher ff = new ChunkStitcher(chunkSizes, chunkLeftMSB, chunkRightMSB, chunkBatch);
+    ff.doAll(fr);
+    Log.debug("took: " + (System.nanoTime() - t0) / 1e9+" seconds.");
+    return fr;
+  }
+  public static Frame sortOnly(final Frame leftFrame, final int leftCols[], int[][] id_maps, int[] ascendingL) {
```
honzasterba: too much whitespace

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -327,26 +329,145 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+    Log.debug("took: " + (System.nanoTime() - t0) / 1e9+" seconds");
+
+    Log.info("Finally stitch together by overwriting dummies ...");
+    t0 = System.nanoTime();
+    Frame fr = new Frame(names, vecs);
+    ChunkStitcher ff = new ChunkStitcher(chunkSizes, chunkLeftMSB, chunkRightMSB, chunkBatch);
+    ff.doAll(fr);
+    Log.debug("took: " + (System.nanoTime() - t0) / 1e9+" seconds.");
+    return fr;
+  }
```
honzasterba: missing empty line

h2o-core/src/main/java/water/rapids/SortCombine.java
```diff
@@ -0,0 +1,407 @@
...
+  final FFSB _leftSB;
+  private transient KeyOrder _leftKO;
+
+  private transient long _leftFrom;
+  private transient int _retBatchSize;
+  private int[][] _numRowsPerCidx;
+  private int _chunkNum;
+  private boolean[] _stringCols;
+  private boolean[] _intCols;
+  SingleThreadRadixOrder.OXHeader _leftSortedOXHeader;
```
honzasterba: seems like this could be final too

h2o-py/tests/pyunit_utils/utilsPY.py
honzasterba: I know its copied everywhere, but that makes me like it even less

h2o-py/tests/pyunit_utils/utilsPY.py
honzasterba: strange, then why does it show up here as new code? was it moved from somewhere?

wendycwong: I might have moved it.  Not sure.

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Fixed

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Fixed

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Fixed

h2o-core/src/main/java/water/rapids/SortCombine.java
wendycwong: fixed

h2o-core/src/main/java/water/rapids/Merge.java
honzasterba: it is not, the indent of the method header is wrong

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -327,26 +329,253 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+  
+  public static Frame sortOnly(final Frame leftFrame, final int leftCols[], int[][] id_maps, int[] ascendingL) {
+    createIndex(true ,leftFrame,leftCols,id_maps, ascendingL);  // sort the columns.
+    Log.info("Making BinaryMerge RPC calls ... ");
+    ArrayList<SortCombine> bmList = gatherSameMSBRows(leftFrame); // For each MSB, gather sorted rows with same MSB into one spot 
+    
+    Log.info("Allocating and populating chunk info (e.g. size and batch number) ...");
+    return allocatePopulateChunk(bmList, leftFrame);
+
+/*    createIndex(true ,leftFrame,leftCols,id_maps, ascendingL);  // sort the columns.
```
honzasterba: why is this commented out?

h2o-core/src/main/java/water/rapids/Merge.java
Pscheidl: I'd personally stress this more than the formatting. Let's use the `List` interface there please.

PR #3925, opened 2019-09-26
Fix pyunit_pubdev_5454 tests

wendycwong:
I added one extra argument as:

compare_frames_local_onecolumn_NA(f1, f2, prob=0.5, tol=1e-6, returnResult=False, oneLessRow=False):

This fixed tests:
pyunit_pubdev_5456_one_group_only_large.py and pyunit_pubdev_5456_rank_within_group_large.py

### DIFF
--- a/h2o-py/tests/pyunit_utils/utilsPY.py
+++ b/h2o-py/tests/pyunit_utils/utilsPY.py
@@ -3562,15 +3562,19 @@ def compare_frames_local_svm(f1, f2, prob=0.5, tol=1e-6, returnResult=False):
 
 
 # frame compare with NAs in column
-def compare_frames_local_onecolumn_NA(f1, f2, prob=0.5, tol=1e-6, returnResult=False):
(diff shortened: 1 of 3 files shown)


h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -3570,9 +3570,9 @@ def compare_frames_local_onecolumn_NA(f1, f2, prob=0.5, tol=1e-6, returnResult=F
     temp2 = f2.as_data_frame(use_pandas=False)
     assert (f1.nrow==f2.nrow) and (f1.ncol==f2.ncol), "The two frames are of different sizes."
     if oneLessRow:
-        lastRow = f2.nrow
```
honzasterba: this is weird, all I can see is a whitespace change

wendycwong: I accidentally pushed to rel-yau.  I thought I reverted it back but guessed not.

wendycwong: I re-did the whole thing and tried again here.  Please check again.  It should contain three files to be fixed.  Good catch.  Thank you Honza.

PR #3926, opened 2019-09-27
Add support for specifying CA_bundle file in h2o.init/h2o.connect

michalkurka:


michalkurka:
@Pscheidl I've found a good way to make an automated test in our infra, any suggestions please?

PR #3927, opened 2019-09-27
PUBDEV-6884: Disable XGBoost native comparison tests on py2

honzasterba:
...xgboost python lib is not supported on py2 anymore


h2o-py/tests/testdir_algos/xgboost/pyunit_H2OXGBoost_native_XGBoost_enum_binomial_compare_sparse_singleNode.py
```diff
@@ -1,6 +1,6 @@
-import xgboost as xgb
 import time
-
+import sys
+sys.path.insert(1,"../../../")
```
Pscheidl: Thank you !

PR #3931, opened 2019-09-30
[PUBDEV-6922] Send data to and from external backend not by each item, but after specified block size

jakubhava:
The goal of this change is to send the data to and from external backend after some pre-defined ( currently hard-coded) batch size to improve the overall performance

### DIFF
--- a/h2o-core/src/main/java/water/ExternalFrameWriterClient.java
+++ b/h2o-core/src/main/java/water/ExternalFrameWriterClient.java
@@ -51,15 +51,15 @@ import static water.ExternalFrameUtils.writeToChannel;
  * </p>
  */
 final public class ExternalFrameWriterClient {
-  private AutoBuffer ab;
-  private ByteChannel channel;
+  private final AutoBuffer ab;
+  private final ByteChannel channel;
   private byte[] expectedTypes;
   private int currentColIdx = 0;
   private int currentRowIdx = 0;
-  private int timeout;
+  private final int timeout;
   private int numRows;
   private int numCols;
-  private long blockSize;
+  private final long blockSize;
 
   /**
    * Initialize the External frame writer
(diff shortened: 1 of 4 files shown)


mn-mikke:
This feature improves performance of conversion Spark DataFrames to H2O Frames with external backend significantly:
```
Tested on EMR 4+4 nodes (1 core each)

before:
300M rows, 1 column:    312225 milliseconds
300M rows, 10 columns: 1527652 milliseconds
300M rows, 100 columns: timeout after 4 hours

after:
300M rows, 1 column:     46142 milliseconds
300M rows, 10 columns:  111857 milliseconds
300M rows, 100 columns: 630779 milliseconds
```

h2o-core/src/main/java/water/ExternalFrameWriterClient.java
```diff
@@ -60,6 +59,7 @@
   private int timeout;
   private int numRows;
   private int numCols;
+  private long blockSize;
```
Pscheidl: Minor: these could be final ?

h2o-core/src/main/java/water/ExternalFrameWriterClient.java
```diff
@@ -124,77 +125,79 @@ public void finalizeFrame(String keyName, long[] rowsPerChunk, byte[] colTypes,
...
 
-    public void close() throws IOException, ExternalFrameConfirmationException {
-        waitForRequestToFinish(timeout, ExternalBackendRequestType.WRITE_TO_CHUNK.getByte());
-        channel.close();
-    }
+  public void close() throws IOException, ExternalFrameConfirmationException {
+    // write remaining data
+    writeToChannel(ab, channel);
+    waitForRequestToFinish(timeout, ExternalBackendRequestType.WRITE_TO_CHUNK.getByte());
+    channel.close();
```
Pscheidl: What if the wait operation times out ? Is the channel.close() is going to be closed and resources freed ?

h2o-core/src/main/java/water/ExternalFrameWriterClient.java
mn-mikke: Good point, thanks!

mn-mikke:
@Pscheidl I've tried to address your changes. Please take a look, thanks!

Pscheidl:
I think it's great. Thanks.

jakubhava:
Great speedup, thanks for measuring Marek. Did you did manual benchmarks or do we have these already in our benchmarks suite on aws?

PR #3936, opened 2019-09-30
PUBDEV-6897: Add GBM POJO support for OneHotExplicit encoding

michalkurka:
This enables users to export a POJO for a model that was trained
with 1-hot-explicit encoding.

There is currently no support in EasyPredictModelWrapper - users will
need to apply the encoding externally and let the Wrapper know by
flipping a `useExternalEncoding` to true.

### DIFF
--- a/h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
+++ b/h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
@@ -115,6 +115,39 @@ public class GBMTest extends TestUtil {
     }
   }
 
+  @Test public void testOneHotExplicitWithPOJO() {
+    try {
+      Scope.enter();
+      final String response = "CAPSULE";
+      Frame fr = parse_test_file("./smalldata/logreg/prostate.csv")
+              .toCategoricalCol("RACE")
+              .toCategoricalCol(response);
+      fr.remove("ID").remove();
+      Scope.track(fr);
+      DKV.put(fr);
+
+      GBMModel.GBMParameters parms = makeGBMParameters();
+      parms._train = fr._key;
+      parms._response_column = response;
+      parms._ntrees = 5;
+      parms._categorical_encoding = Model.Parameters.CategoricalEncodingScheme.OneHotExplicit;
+
+      GBM job = new GBM(parms);
+      GBMModel gbm = job.trainModel().get();
+      Scope.track_generic(gbm);
+
+      // Done building model; produce a score column with predictions
+      Frame scored = Scope.track(gbm.score(fr));
+
+      // Build a POJO & MOJO, validate same results
+      Assert.assertTrue(gbm.testJavaScoring(fr, scored,1e-15));
+
+    } finally {
+      Scope.exit();
+    }
+
+  }
+  
   @Test public void testBasicGBM() {
     // Regression tests
     basicGBM("./smalldata/junit/cars.csv",
--- a/h2o-genmodel/src/main/java/hex/genmodel/CategoricalEncoding.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/CategoricalEncoding.java
@@ -1,0 +1,6 @@
+package hex.genmodel;
+
+public enum CategoricalEncoding {
+  AUTO,
+  OneHotExplicit
+}
--- a/h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
@@ -1,10 +1,7 @@
 package hex.genmodel.easy;
 
 import hex.ModelCategory;
-import hex.genmodel.GenModel;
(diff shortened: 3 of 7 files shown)


h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
```diff
@@ -115,6 +115,39 @@ Constraints emptyConstraints(Frame f) {
...
+
+      GBM job = new GBM(parms);
+      GBMModel gbm = job.trainModel().get();
+      Scope.track_generic(gbm);
+
+      // Done building model; produce a score column with predictions
+      Frame scored = Scope.track(gbm.score(fr));
+
+      // Build a POJO & MOJO, validate same results
+      Assert.assertTrue(gbm.testJavaScoring(fr, scored,1e-15));
```
Pscheidl: Minor: The absolute delta can easily be zero in this case. But 15 decimal places is also considered enough.

h2o-genmodel/src/main/java/hex/genmodel/CategoricalEncoding.java
```diff
@@ -0,0 +1,6 @@
+package hex.genmodel;
```
Pscheidl: Should this really be in a top-level packages ? Maybe somewhere in  the `algos` subpackage ? Would be hard to change in future, as this is user-facing.

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -197,6 +195,13 @@ public Config setEnableContributions(boolean val) throws IOException {
 
     public boolean getEnableContributions() { return enableContributions; }
 
+    public Config setUseExternalEncoding(boolean val)  {
```
Pscheidl: We do have JavaDoc on top of most of these methods. Not all, but these are directly user-facing and I'd personally provide it.

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -299,7 +304,13 @@ public EasyPredictModelWrapper(Config config) {
     } else {
       predictContributions = null;
     }
-    
+
+    if ((m.getCategoricalEncoding() != CategoricalEncoding.AUTO) && !config.useExternalEncoding) {
+      throw new UnsupportedOperationException("Categorical Encoding `" + m.getCategoricalEncoding() +
```
Pscheidl: Would this be a good candidate for a technical note ?

h2o-genmodel/src/main/java/water/genmodel/IGeneratedModel.java
```diff
@@ -54,6 +61,9 @@
     /** Returns domain values for all columns including response column. */
     public String[][] getDomainValues();
 
+    /** Returns original domain values for all columns including response column. */
+    public String[][] getOrigDomainValues();
```
Pscheidl: I don't understand why this new method is introduced and then it always returns null in `GenModel`.

At least in some cases, could `getOrigDomainValues` return the same values as `getDomainValues` ?

h2o-genmodel/src/main/java/hex/genmodel/CategoricalEncoding.java
michalkurka: IMHO this is not an entirely bad location, it is where all the genmodel interfaces live (IGenModel, GenModel itself). It is part of the genmodel interface. It could be in algos if you think that is better.

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
michalkurka: good point, will add

h2o-genmodel/src/main/java/hex/genmodel/CategoricalEncoding.java
Pscheidl: No need. Just wanted to point out whether there could be any issue related to the location.

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
michalkurka: could be, there is no technote infrastructure in genmodel though - I will make a reference to a jira on github where the functionality will be added

h2o-genmodel/src/main/java/water/genmodel/IGeneratedModel.java
michalkurka: this method is only useful when the categorical encoding is not handled by the model (!= AUTO), this is the case when we need the original domain and the original column names - it is overridden in the generated POJO

imho this mimics the behavior of model._output._origDomains - they are only set in certain scenarios an are null in others

PR #3939, opened 2019-10-01
PUBDEV-6883: CountingErrorConsumer can now optionally collect unseen values

michalkurka:


Tagar:
Thank you @michalkurka !

h2o-genmodel/src/main/java/hex/genmodel/easy/error/CountingErrorConsumer.java
```diff
@@ -69,6 +86,19 @@ public void dataTransformError(String columnName, Object value, String message)
...
   public void unseenCategorical(String columnName, Object value, String message) {
     unknownCategoricalsPerColumn.get(columnName).incrementAndGet();
+    if (collectUnseenCategoricals) {
+      ConcurrentMap<Object, AtomicLong> columnCollector = unseenCategoricalsCollector.get(columnName);
+      assert columnCollector != null;
+      AtomicLong counter = columnCollector.get(value);
+      if (counter != null) {
+        counter.incrementAndGet(); // best effort to avoid creating new AtomicLongs on each invocation 
+      } else {
+        counter = columnCollector.putIfAbsent(value, new AtomicLong(1));
```
Tagar: .. just trying to follow the logic if `collectUnseenCategoricals` is not enabled (false),
then we should count only at column level - how many unknowns were seen for 
that column as a whole, without having a counter for each `value` separately.. 
Why is `value` present then in the this `else` branch? 
Thanks !

michalkurka: This else branch doesn't belong to the top level if-statement that decides whether to count on column level or also value level.

Tagar: That explains it :) That was a long day. Thanks

PR #3943, opened 2019-10-03
PUBDEV-6899: Enable SPNENO authentication in R client

honzasterba:


h2o-r/h2o-package/R/classes.R
```diff
@@ -69,7 +70,7 @@ setRefClass("H2OConnectionMutableState",
 setClass("H2OConnection",
          representation(ip="character", port="numeric", name="character", proxy="character",
                         https="logical", cacert="character", insecure="logical",
-                        username="character", password="character",
+                        username="character", password="character", use_spnego="logical",
```
Pscheidl: Good naming !

h2o-r/h2o-package/R/connection.R
```diff
@@ -148,6 +149,10 @@ h2o.init <- function(ip = "localhost", port = 54321, name = NA_character_, start
     stop("`cacert` must be a character string or NA_character_")
   if(!is.logical(insecure) || length(insecure) != 1L || is.na(insecure))
     stop("`insecure` must be TRUE or FALSE")
+  if(!is.logical(use_spnego) || length(use_spnego) != 1L || is.na(use_spnego))
```
Pscheidl: We could generalize these check in one place somewhere in future. Just saying.

michalkurka: @honzasterba this option is a good candidate to have in ~./h2oconfig - please add reading the value from the config to the h2o.init function (see eg. `init.cacert`)

michalkurka:
thanks @honzasterba 

Please add getting the default value from config. This should make it a bit more user-friendly.

Would there be a way to enable the option automatically based on the response of the server? It would be nice if it just worked without the extra parameter.

honzasterba:
reading from config added

regarding of activating spnego automatically - I think this could be done, but to me it does not sound like the correct thing to do - I dont think we should share users security credentials (even encrypted) unless instructed to do so

anyway sound like out of scope of this PR

PR #3944, opened 2019-10-03
PUBDEV-6941: Added blending of posterior and prior during imputation of unseen values

deil87:
See details in https://0xdata.atlassian.net/browse/PUBDEV-6941

### DIFF
--- a/h2o-extensions/target-encoder/build.gradle
+++ b/h2o-extensions/target-encoder/build.gradle
@@ -9,6 +9,8 @@ dependencies {
         
     testCompile "junit:junit:${junitVersion}"
     testCompile "org.mockito:mockito-core:2.23.0"
+    testCompile "com.pholser:junit-quickcheck-core:0.9"
+    testCompile 'com.pholser:junit-quickcheck-generators:0.9'
     testCompile project(path: ":h2o-core", configuration: "testArchives")
     testCompile project(path: ":h2o-algos", configuration: "testArchives")
     testRuntimeOnly project(":${defaultWebserverModule}")
--- a/h2o-extensions/target-encoder/src/test/java/ai/h2o/targetencoding/TEMojoIntegrationTest.java
+++ b/h2o-extensions/target-encoder/src/test/java/ai/h2o/targetencoding/TEMojoIntegrationTest.java
@@ -1,5 +1,8 @@
 package ai.h2o.targetencoding;
 
+import com.pholser.junit.quickcheck.Property;
+import com.pholser.junit.quickcheck.generator.InRange;
+import com.pholser.junit.quickcheck.runner.JUnitQuickcheck;
 import hex.genmodel.MojoModel;
 import hex.genmodel.algos.targetencoder.EncodingMap;
 import hex.genmodel.algos.targetencoder.EncodingMaps;
@@ -11,6 +14,7 @@ import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
 import water.DKV;
 import water.Scope;
 import water.TestUtil;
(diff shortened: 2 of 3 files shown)


h2o-extensions/target-encoder/build.gradle
```diff
@@ -9,6 +9,8 @@ dependencies {
         
     testCompile "junit:junit:${junitVersion}"
     testCompile "org.mockito:mockito-core:2.23.0"
+    testCompile "com.pholser:junit-quickcheck-core:0.9"
```
deil87: Decided to add it within this PR but might consider to add this dependency separately if there will be concerns about it

michalkurka: Not an issue - the usage is an issue, see bellow

h2o-extensions/target-encoder/src/test/java/ai/h2o/targetencoding/TEMojoIntegrationTest.java
```diff
@@ -486,6 +582,80 @@ public void check_that_encodings_for_unexpected_values_are_the_same_in_TargetEnc
     }
   }
 
+  @Property(trials = 5)
```
michalkurka: I don't see any reason to attempt to do 5 trials, it doesn't increase coverage and running the test once doesn't give use any confidence of what it is is actually doing. Good tests have seeds and don't rely on probabilistic approach that your edge scenario will occur.

deil87: @michalkurka I can't agree with you. Please take a look at property-based testing (PBT) concept. Here for example https://dev.to/jdsteinhauser/intro-to-property-based-testing-2cj8 . This particular PR is a proof actually that PBT is a useful thing. Test was working for only specific hardcoded values and would fail for rest of the cases. And it does increase coverage. Over time it will test all values that developer has specified through the range of the possible values. 
Seed is only a way to provide reproducibility. PBT framework will provide you values that led to a failure. It can also tell you what exact seed was used as we can generate seed in a same way we generate `randomInflectionPoint `. Good test should check all the possible values that can come as an input to your functionality.

michalkurka: Yes, PBT is certainly useful for test development. However, discovered edge cases that you didn't think of during the test development should still be turned into unit tests. Is that the case of this PR?

deil87: @michalkurka If I found one use case that was not working then there is a probability that there are others. I see your point but the right answer is that from the beginning we should have used PBT. I current situation maybe it would be more explicit to add separate unit test with the value that was causing the failure. But I would still keep the PBT version as in the long run( in case when number of runs is relatively small) test will cover all possible cases. In a given situation we have pivot value that is equal to a number of missing values in the training set and only two tests are enough (less than pivot and greater than pivot).... but purpose of our tests is actually to be a guard against regression caused by the future changes. So how do you like the idea of keeping two tests?

deil87: @michalkurka Actually I do have separate tests for edge cases(including one that was failing) in the test suite: with large `inflection_point` and small `inflection_point`. PBT-like test was added for another scenario just to cover more use cases. Are you ok with me merging this PR?

michalkurka: Ok, thank you for confirming. In that case, I don't think having it there is an issue (as long is doesn't take too much time).

Regarding merging - I didn't review it in detail yet. Please give me a little bit more time.

deil87:
Tests are fine( failures are unrelated) and also PR was tested against SW => merging

PR #3945, opened 2019-10-03
Fix testChunkIdxWithDeletedVec

michalkurka:
The original test cannot run reliably in multi-node

### DIFF
--- a/h2o-core/src/test/java/water/fvec/VecTest.java
+++ b/h2o-core/src/test/java/water/fvec/VecTest.java
@@ -197,26 +197,17 @@ public class VecTest extends TestUtil {
   }
 
   @Test public void testChunkIdxWithDeletedVec() {
+    Key<Vec> k = null;
     try {
       Scope.enter();
       Vec v = Scope.track(makeConN((long) 1e6, 16));
-      new MRTask() {
-        @Override
-        protected void setupLocal() {
-          Vec v = _fr.anyVec();
-          if (v != null) {
-            v.remove(); // evil happens here
-          }
-        }
-        @Override
-        public void map(Chunk c) {
-          if (c.vec().get_type() != Vec.T_NUM)
-            throw new IllegalStateException("Expected a numeric Vec");
-        }
-      }.doAll(v);
+      k = v._key;
+      v.remove();
+      v.chunkForChunkIdx(0);
+      fail("chunkForChunkIdx didn't fail");
     } catch (Exception e) {
       String msg = e.getMessage();
-      if (msg == null || !msg.contains("Missing chunk") || !msg.contains("is not in DKV"))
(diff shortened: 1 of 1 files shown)


h2o-core/src/test/java/water/fvec/VecTest.java
```diff
@@ -197,26 +197,16 @@ public void map(Chunk c) {
...
-        }
-        @Override
-        public void map(Chunk c) {
-          if (c.vec().get_type() != Vec.T_NUM)
-            throw new IllegalStateException("Expected a numeric Vec");
-        }
-      }.doAll(v);
+      k = v._key;
+      v.remove();
+      v.chunkForChunkIdx(0);
```
seb-h2o: if you expect an exception, you should probably add a `fail()` statement before `catch`.

h2o-core/src/test/java/water/fvec/VecTest.java
```diff
@@ -197,26 +197,16 @@ public void map(Chunk c) {
...
     } catch (Exception e) {
       String msg = e.getMessage();
-      if (msg == null || !msg.contains("Missing chunk") || !msg.contains("is not in DKV"))
```
seb-h2o: maybe checking `e.toString()` instead of `e.getMessage()` would work here?

michalkurka: Can you please explain what issue it addresses?

seb-h2o: I thought that as DistributedException may have been wrapping the original exception without message using basic constructor `public DistributedException(Throwable cause)`, then your test would always fail.

seb-h2o: for example in `RPC` and `MRTask`:
```
        throw (re instanceof DistributedException)?new DistributedException(re.getMessage(),re.getCause()):new DistributedException(re);
```
for some (unknown) reason, the message is propagated only if the cause exception is already a `DistributedException`

seb-h2o: my mistake, didn't see the full logic: by default it's prefixing the cause message...

michalkurka: Okay, I see - that would be for the original code.

Current code won't involve distributed execution - I will always get exactly the exception I am throwing.

The issue with the previous one was not wrapping into DistributedException, it was an NPE in the initialization of the MRTask on the remote nodes.

seb-h2o: yes, I'm using split view (there, it is clear it was a comment on original test).
with unified view, my comment may have been confusing

seb-h2o:
@michalkurka is it ready for merge?

PR #3946, opened 2019-10-04
PUBDEV-6893 - Saving files in Python not properly creating directories

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6893

This is a problem with paths - with splits being done in the code, there are scenarios where the path given might be considered as ending with a filename, rather than a directory. This happens when users forgets to put slash at the end of the string (like the issue reporter did : `h2o.download_pojo(model=m, path='test_ae_pojo', get_jar=True)`). Therefore the "NotADirectory" error reported.

Solution is to simply use `os.path.join` to consider the user-given path to end with a diretory all the time. `os.path.join` manages to do it in an os-independent way (so the documentation says).

H2O documentation of the `download_pojo` method says: `    :param path: an absolute path to the directory where POJO should be saved.`.

### DIFF
--- a/h2o-py/h2o/backend/connection.py
+++ b/h2o-py/h2o/backend/connection.py
@@ -736,11 +736,13 @@ class H2OConnection(backwards_compatible()):
         status_code = response.status_code
         if status_code == 200 and save_to:
             if save_to.startswith("~"): save_to = os.path.expanduser(save_to)
+            save_to = os.path.join(str(save_to), '')
             if os.path.isdir(save_to) or save_to.endswith(os.path.sep):
-                dirname = os.path.abspath(save_to)
+                dirname = os.path.join(os.path.abspath(save_to), '')
                 filename = H2OConnection._find_file_name(response)
             else:
                 dirname, filename = os.path.split(os.path.abspath(save_to))
+                dirname = os.path.join(dirname, '')
             fullname = os.path.join(dirname, filename)
             try:
                 if not os.path.exists(dirname):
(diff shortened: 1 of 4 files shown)


h2o-py/h2o/backend/connection.py
```diff
@@ -736,11 +736,13 @@ def _process_response(response, save_to):
         status_code = response.status_code
         if status_code == 200 and save_to:
             if save_to.startswith("~"): save_to = os.path.expanduser(save_to)
+            save_to = os.path.join(str(save_to), '')
```
honzasterba: I think it would be more readable to do `if save_to.endswith("/"): save_to = save_to + "/"` and then mkdirs() (pardon my pseudo-python)

Pscheidl: `os.path.join(str(save_to), '')` has no effect if it already ends with "/" - the check is inside, this is why I omitted it here. The premature optimization inside me made me do this !

Also, they already solved the os-independency. E.g. there are some Windows users out there who use "\" instead of "/".

Pscheidl:
Second attempt, this time a sane and better one. Two problems:

1. Python's function `os.path.abspath(save_to)` returns the absolute path, but without path separator at the end, which makes `os.makedirs(dirname)` a few lines down treat it badly !
Because the block I "adjusted" only treats directory, it is made sure the correct line separator per OS is present. 
2. :warning: Controversial: The `download_pojo` function makes sure the input `path` variable is always treated as folder by appending os-specific path separator at the end all the time.

PR #3948, opened 2019-10-04
PUBDEV-6948: Disable HTTP TRACE Method

michalkurka:
https://0xdata.atlassian.net/browse/PUBDEV-6948

### DIFF
--- a/h2o-core/src/main/java/water/server/ServletUtils.java
+++ b/h2o-core/src/main/java/water/server/ServletUtils.java
@@ -30,6 +30,8 @@ public class ServletUtils {
    */
   private static final boolean DISABLE_CORS = Boolean.getBoolean(H2O.OptArgs.SYSTEM_PROP_PREFIX + "disable.cors");
 
+  private static final String TRACE_METHOD = "TRACE";
+
   private static final ThreadLocal<Long> _startMillis = new ThreadLocal<>();
   private static final ThreadLocal<Integer> _status = new ThreadLocal<>();
   private static final ThreadLocal<String> _userAgent = new ThreadLocal<>();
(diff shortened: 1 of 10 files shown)


michalkurka:
cc: @mn-mikke - we will need to make a nightly build of SW that bundles this fix to test with an user

h2o-core/src/test/java/water/webserver/H2OHttpViewImplTest.java
```diff
@@ -0,0 +1,33 @@
...
+    };
+
+    HttpServletRequest request = mock(HttpServletRequest.class);
+    HttpServletResponse response = mock(HttpServletResponse.class);
+    
+    when(request.getMethod()).thenReturn("TRACE");
+    
+    httpView.gateHandler(request, response);
+
+    verify(response).setStatus(405);
```
Pscheidl: Great. 405 is the correct answer indeed.

h2o-core/src/main/java/water/server/ServletUtils.java
```diff
@@ -172,6 +174,10 @@ public static boolean isXhrRequest(final HttpServletRequest request) {
     return "XMLHttpRequest".equals(requestedWithHeader);
   }
 
+  public static boolean isTraceRequest(final HttpServletRequest request) {
+    return TRACE_METHOD.equalsIgnoreCase(request.getMethod());
```
Pscheidl: Ignoring case is necessary. Some implementations actually might screw up the case. I've seen this many times.

Pscheidl:
Re-started the tests manually. No PR-related failures observed, one failure due to infrastructure.

mn-mikke:
> we will need to make a nightly build of SW that bundles this fix to test with an user
@michalkurka No problem, just let me know once you have the nightly H2O-3 build.

Pscheidl:
Went through the test failures - unrelated.

PR #3951, opened 2019-10-05
PUBDEV-6949: Got rid of H2O.STORE.clear() in FrameTest.java and enabled deepSlice test

deil87:
This is related to an old Jira https://0xdata.atlassian.net/browse/PUBDEV-6019

After removal of the H2O.STORE.clear() from the `FrameTest.testRemoveColumn()` test was able to enable deepSlice test.

### DIFF
--- a/h2o-core/src/test/java/water/fvec/FrameTest.java
+++ b/h2o-core/src/test/java/water/fvec/FrameTest.java
@@ -55,12 +55,11 @@ public class FrameTest extends TestUtil {
   @Test
   public void testRemoveColumn() {
     Scope.enter();
-    Set<Vec> removedVecs = null;
+    Set<Vec> removedVecs = new HashSet<>();
 
     try {
       Frame testData = parse_test_file(Key.make("test_deep_select_1"), "smalldata/sparse/created_frame_binomial.svm.zip");
       Scope.track(testData);
-      removedVecs = new HashSet<>();
 
       // dataset to split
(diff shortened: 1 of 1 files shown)


h2o-core/src/test/java/water/fvec/FrameTest.java
```diff
@@ -55,10 +55,13 @@ public void testNonEmptyChunks() {
...
   public void testRemoveColumn() {
     Scope.enter();
-    Frame testData = parse_test_file(Key.make("test_deep_select_1"), "smalldata/sparse/created_frame_binomial.svm.zip");
-    Set<Vec> removedVecs = new HashSet<>();
+    Set<Vec> removedVecs = null;
 
     try {
+      Frame testData = parse_test_file(Key.make("test_deep_select_1"), "smalldata/sparse/created_frame_binomial.svm.zip");
+      Scope.track(testData);
+      removedVecs = new HashSet<>();
```
seb-h2o: merge this with L58.
otherwise you can get a NPE in the finally block if you get an error during parsing.

deil87: @seb-h2o thanks! I was thinking about it as well. I wanted even to introduce helper Scope.track method vararg of the vecs. But maybe for now there is a simpler solution -> see commit

deil87:
Failing tests are unrelated to the scope of the change. Going to merge.

PR #3953, opened 2019-10-06
PUBDEV-6936 fix custom distribution multinomial gradient

maurever:
JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6936

### DIFF
--- a/h2o-core/src/main/java/water/udf/CDistributionFunc.java
+++ b/h2o-core/src/main/java/water/udf/CDistributionFunc.java
@@ -7,6 +7,7 @@ package water.udf;
  * - link: link function transforms the probability of response variable to a continuous scale that is unbounded
  * - init: computes numerator and denominator of the initial value.
  * - gradient: computes (Negative half) Gradient of deviance function at predicted value for actual response
+ * - lgradient: computes (Negative half) Gradient of deviance function at predicted value for actual response and a class label (only for multinomial case)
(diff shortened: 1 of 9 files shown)


h2o-core/src/main/java/water/udf/CDistributionFunc.java
```diff
@@ -35,6 +36,16 @@
...
 
+    /**
+     * (Negative half) Gradient of deviance function at predicted value f, for actual response y.
+     * Important for customization of a loss function.
+     * @param y (actual) response
+     * @param f (predicted) response in link space (including offset)
+     * @param l (class label) label of a class (converted lexicographically from original labels to 0-number of class - 1)  
+     * @return gradient
+     */
+    double lgradient(double y, double f, int l);
```
maurever: I decided to add a new method rather than add a new parameter to the gradient method. It is compatible with the old code, and it allows using a class label to customize gradient calculation now. 

WDYT @michalkurka?

maurever:
Tests look good (only unrelated xgboost fails), I rerun once again to sure it is working.

h2o-core/src/main/java/water/udf/CDistributionFunc.java
maurever: Finally, I removed `lgradient` method and override gradient method.

PR #3954, opened 2019-10-07
Fix message when flatfile has ignored entries

michalkurka:


h2o-core/src/main/java/water/Paxos.java
```diff
@@ -160,7 +160,7 @@ static private void lockCloud_impl(Object reason) {
       if(H2O.isFlatfileEnabled()){
         for(H2ONode n: H2O.getFlatfile()){
           if(!n._heartbeat._client && !PROPOSED.containsKey(n._key)){
-            Log.info("Flatile::" + n._key + " not active in this cloud. Removing it from the list.");
```
michalkurka: Typo :)

PR #3956, opened 2019-10-08
Update the list of contributors

Pscheidl:
Found out you three are missing and should definitely be there. WDYT ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -799,6 +799,9 @@ Leland Wilkinson
 Wendy Wong
 Nikhil Shekhar
 Pavel Pscheidl
+Michal Kurka
+Veronika Maurerova
+Jan Sterba
 ```
 
 <a name="Advisors"></a>


maurever:
@michalkurka  is not on the list of contributors? :-O

michalkurka:
@maurever 

contributing?

let me quote The Office (http://officequotes.net/no8-04.php)

> Robert: You people say I led you, but it wasn't me. You want to toast the man who led you to success, but the boss is irrelevant. Andy and I, **we produce nothing. We do nothing. We sit in our offices and demand, I want this and that right now, like petulant children. You know, the difference between a crying baby and a manager, one day the baby will grow up**. But, without you, Andy and I would be sitting in our dirty diapers, waiting for someone to change us, wipe us. I should be toasting you, thanking you, for allowing me to have the easiest job in the universe. Cheers.

BTW where is constrained k-means? 👯‍♂

michalkurka:
@Pscheidl seems like this change is breaking several junit tests ;)

maurever:
> BTW where is constrained k-means? :-)

In my head and my local git.... :woman_technologist:

But against you, @michalkurka, my contribution is tiny, so please, let add me to the list, when it will be comparable! :)

PR #3958, opened 2019-10-09
XGBoost: make gblinear deterministic

honzasterba:
- set updater for gblinear when running on CPU to coord_descent to get rid of indeterminism

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
@@ -212,6 +212,27 @@ public class XGBoostModel extends Model<XGBoostModel, XGBoostModel.XGBoostParame
       }
     }
   }
+  
+  private static XGBoostParameters.Backend findSuitableBackend(XGBoostParameters p) {
+    if ( p._backend == XGBoostParameters.Backend.auto || p._backend == XGBoostParameters.Backend.gpu ) {
+      if (H2O.getCloudSize() > 1) {
+        Log.info("GPU backend not supported in distributed mode. Using CPU backend.");
+        return XGBoostParameters.Backend.cpu;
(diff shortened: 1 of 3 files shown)


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -212,6 +212,27 @@ public void dump(String format) {
...
+        return XGBoostParameters.Backend.gpu;
+      } else {
+        Log.info("No GPU (gpu_id: " + p._gpu_id + ") found. Using CPU backend.");
+        return XGBoostParameters.Backend.gpu;
+      }
+    } else {
+      Log.info("Using CPU backend.");
+      return XGBoostParameters.Backend.cpu;
+    }
+  }
```
honzasterba: I separated the part of the code that does the backend decision from the part that sets the params based on selected backend and booster

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostPredictImplComparisonTest.java
```diff
@@ -30,23 +30,20 @@ public static void setup() {
             {"gbtree", "gamma", "AGE"},
             {"gbtree", "poisson", "AGE"},
             {"gbtree", "tweedie", "AGE"},
-            {"gbtree", "gamma", "AGE"},
```
honzasterba: duplicit param settings removed

michalkurka:
Can we assume that users who don't set the seed explicitly don't care about model reproducibility and resort to the safe version coordinate descent only when the seed is set?

Any deviation from xgboost default parameters should also be documented.

honzasterba:
I dont like the idea of using different updaters with different seed settings. I would not want to document that. Since determinism with same seed is important to us I would stick with coord_descent and document it.

PR #3959, opened 2019-10-09
PUBDEV-6739 Expose original parameters of imported MOJO model

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6739

### DIFF
--- a/h2o-algos/src/main/java/hex/generic/GenericModelParameters.java
+++ b/h2o-algos/src/main/java/hex/generic/GenericModelParameters.java
@@ -1,8 +1,16 @@
 package hex.generic;
 
 import hex.Model;
+import hex.genmodel.attributes.parameters.ColumnSpecifier;
+import hex.genmodel.attributes.parameters.ModelParameter;
+import hex.genmodel.attributes.parameters.ParameterKey;
+import water.Iced;
+import water.IcedWrapper;
 import water.Key;
+import water.api.schemas3.FrameV3;
+import water.api.schemas3.ModelParameterSchemaV3;
 import water.fvec.Frame;
+import water.util.Log;
 
 public class GenericModelParameters extends Model.Parameters {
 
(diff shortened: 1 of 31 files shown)


h2o-algos/src/main/java/hex/generic/GenericModelParameters.java
```diff
@@ -22,6 +30,55 @@
...
+    public ModelParameterSchemaV3[] _modelParameters;
+
+    protected static ModelParameterSchemaV3[] convertParameters(final ModelParameter[] originalParams) {
+        final ModelParameterSchemaV3[] convertedParams = new ModelParameterSchemaV3[originalParams.length];
+
+        for (int i = 0; i < originalParams.length; i++) {
+            final ModelParameter originalParam = originalParams[i];
+            final ModelParameterSchemaV3 convertedParam = new ModelParameterSchemaV3();
+            // Hand-built mapping for better performance compared to reflection
+            convertedParam.name = originalParam.name;
```
michalkurka: some of these parameters are only good for model building (eg. gridable) - should we store these as well?

h2o-core/src/main/java/water/api/schemas3/ModelParametersSchemaV3.java
```diff
@@ -288,11 +290,17 @@ public static AutoBuffer writeParametersJSON(AutoBuffer ab, ModelParametersSchem
     }
 
     compute_transitive_closure_of_is_mutually_exclusive(metadata);
-
+    if (additionalParameters != null) {
+      metadata = (ModelParameterSchemaV3[]) ArrayUtils.addAll(metadata, additionalParameters);
```
michalkurka: water.util.ArrayUtils#append(T[] a, T... b) handles the null check for you

h2o-algos/src/main/java/hex/generic/GenericModelParameters.java
Pscheidl: It gives our users the full view. This is the only argument I can find. Besides unknown use-cases from distant future, like upgrading Mojo to a full-fledged generic model container, capable of checkpointing. Then, these could be useful. Not sure if that would be backwards compatible all the way.

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/ColumnSpecifier.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.io.Serializable;
+
+public class ColumnSpecifier implements Serializable {
+
+  private final String columnName;
+  private final String[] is_member_of_frames;
+
+  public ColumnSpecifier(String columnName, String[] is_member_of_frames) {
+    this.columnName = columnName;
```
Pscheidl: RequireNotNull missing.

h2o-core/src/main/java/water/api/schemas3/ModelParametersSchemaV3.java
Pscheidl: Thank you. Used our own ArrayUtils instead.

h2o-algos/src/main/java/hex/generic/GenericModelParameters.java
Pscheidl: Not exposing old model_id. As that is irrelevant and hard to expose in a non-confusing way. The imported generic model has it's own model_id. It can be added any time later.

PR #3960, opened 2019-10-09
PUBDEV-6874 StackedEnsemble bad predictions due to one bad model in the stack

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6874

Issue explained in details in ticket above, as well as investigation and results.

The fix consists in disabling standardization for GLM metalearner in AUTO mode: user who wants to use explicit GLM metalearner will still have the default GLM params.


seb-h2o:
@ledell can you please tell if you approve this small change?
Details + benchmark results are in the Jira ticket.

seb-h2o:
@ledell Great, thanks. Didn't know that we were documenting those internal parameters.

PR #3969, opened 2019-10-11
PUBDEV-6961: remove custom nvidia-docker scripts

honzasterba:
- use custom docker runtime arg to replace


michal-raska:
Thanks 👍

PR #3972, opened 2019-10-11
PUBDEV-6928: expose get_automl from `h2o.automl` module

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6928

adding alias and use the occasion to raise a better Error if instance doesn't exist (was previously raising uncaught error)


seb-h2o:
@ledell this is a tiny one

PR #3974, opened 2019-10-11
PUBDEV-6964: Add to docs that MOJOs and POJOs are threadsafe

angela0xdata:
Updated FAQs for MOJOs and POJOs to indicate that all MOJOs/POJOs are thread safe except for Deep Learning

### DIFF
--- a/h2o-docs/src/product/mojo-quickstart.rst
+++ b/h2o-docs/src/product/mojo-quickstart.rst
@@ -611,7 +611,7 @@ FAQ
 
 -  **Are MOJOs thread safe?**
 
-  All of H2O-3 MOJOs are thread safe except for Deep Learning MOJOs. 
+  Yes, all of H2O-3 MOJOs are thread safe.
 
 -  **How can I use an XGBoost MOJO with Maven?**
 
--- a/h2o-docs/src/product/mojo-quickstart.rst
+++ b/h2o-docs/src/product/mojo-quickstart.rst
@@ -611,7 +611,7 @@ FAQ
 
 -  **Are MOJOs thread safe?**
 
-  All of H2O-3 MOJOs are threadsafe except for Deep Learning MOJOs. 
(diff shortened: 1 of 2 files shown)


h2o-docs/src/product/mojo-quickstart.rst
```diff
@@ -609,6 +609,10 @@ The following code snippet shows how to download a MOJO from R and run the Print
 FAQ
 ~~~
 
+-  **Are MOJOs thread safe?**
+
+  All of H2O-3 MOJOs are threadsafe except for Deep Learning MOJOs.
```
laurendiperna: can we change threadsafe to thread-safe

h2o-docs/src/product/pojo-quickstart.rst
```diff
@@ -186,6 +186,10 @@ The following use cases are demonstrated with code examples:
 FAQ
 ~~~
 
+-  **Are POJOs thread safe?**
+
+  All of H2O-3 POJOs are threadsafe except for Deep Learning POJOs.
```
laurendiperna: can we change threadsafe to thread-safe?

h2o-docs/src/product/mojo-quickstart.rst
angela0xdata: Actually, it should be without the hyhen. Will update.

laurendiperna:
thanks!

angela0xdata:
@laurendiperna I don't generally like adding "this will be fixed in the future" links to PUBDEV issues because they end up becoming stale, but I am a little confused. 

Notice that the POJO issue is marked as Resolved > Fixed. Also confused because of the number of issues tied to the MOJO thread safe fix. Or is that just supposed to be https://0xdata.atlassian.net/browse/PUBDEV-6615. And hmm... If PUBDEV-6615 is supposed to be fixed in 3.26.0.7, then this FAQ will have to be updated to remove DL from caveats. I'll keep an eye on that when I'm working on the release notes.

PR #3975, opened 2019-10-13
PUBDEV-6970: introduced AllCategorical and Threshold TE application strategies

deil87:
Added two TE application strategies:
1) All categorical
2) Threshold strategy, meaning categorical columns with number of levels higher than threshold will be subjects for TE encoding

### DIFF
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/AllCategoricalTEApplicationStrategy.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/AllCategoricalTEApplicationStrategy.java
@@ -2,25 +2,30 @@ package ai.h2o.targetencoding.strategy;
 
 import water.fvec.Frame;
 
-import java.util.ArrayList;
+import java.util.Arrays;
 
 public class AllCategoricalTEApplicationStrategy extends TEApplicationStrategy {
   
   private Frame _frame;
-  private String _responseColumnName;
-  
-  public AllCategoricalTEApplicationStrategy(Frame frame, String responseColumnName) {
-    _frame = frame; 
-    _responseColumnName = responseColumnName;
+  private String[] _excludedColumnNames;
+
+  /**
+   *  Constructor for selection of categorical columns strategy 
+   * @param frame the frame selection is being done from
+   * @param excludedColumnNames the column names we want to exclude from the result 
+   *                            ( i.e. response column for classification tasks , fold column in case it is categorical etc.)
+   */
+  public AllCategoricalTEApplicationStrategy(Frame frame, String[] excludedColumnNames) {
+    _frame = frame;
+    _excludedColumnNames = excludedColumnNames;
   }
   
   public String[] getColumnsToEncode() {
-    ArrayList<String> categoricalColumnNames = new ArrayList<>();
-    for( String vecName : _frame.names()) {
-      if(_frame.vec(vecName).isCategorical() && !vecName.equals(_responseColumnName))
-        categoricalColumnNames.add(vecName);
-    }
-    return categoricalColumnNames.toArray(new String[0]);
+    return Arrays.stream(_frame.names())
+        .filter(columnName ->
+            _frame.vec(columnName).isCategorical() && ! Arrays.asList(_excludedColumnNames).contains(columnName)
+        )
+        .toArray(String[]::new);
   }
 
 }
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/AllCategoricalTEApplicationStrategy.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/AllCategoricalTEApplicationStrategy.java
@@ -1,0 +1,26 @@
+package ai.h2o.targetencoding.strategy;
+
+import water.fvec.Frame;
+
+import java.util.ArrayList;
+
+public class AllCategoricalTEApplicationStrategy extends TEApplicationStrategy {
+  
+  private Frame _frame;
+  private String _responseColumnName;
+  
+  public AllCategoricalTEApplicationStrategy(Frame frame, String responseColumnName) {
+    _frame = frame; 
+    _responseColumnName = responseColumnName;
+  }
+  
+  public String[] getColumnsToEncode() {
+    ArrayList<String> categoricalColumnNames = new ArrayList<>();
+    for( String vecName : _frame.names()) {
+      if(_frame.vec(vecName).isCategorical() && !vecName.equals(_responseColumnName))
+        categoricalColumnNames.add(vecName);
+    }
+    return categoricalColumnNames.toArray(new String[0]);
+  }
+
+}
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/ThresholdTEApplicationStrategy.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/ThresholdTEApplicationStrategy.java
@@ -1,40 +1,38 @@
 package ai.h2o.targetencoding.strategy;
 
 import water.fvec.Frame;
-import water.fvec.Vec;
-
-import java.util.ArrayList;
+import java.util.Arrays;
 
 /**
- * Strategy that will select only categorical columns which cardinality is greater of equal than specified threshold
+ * Strategy that will select only categorical columns with cardinality of greater of equal than specified threshold
  */
 public class ThresholdTEApplicationStrategy extends TEApplicationStrategy {
   
   private Frame _frame;
-  private String _responseColumnName;
+  private String[] _excludedColumnNames;
   private long _threshold;
 
   /**
-   * 
-   * @param frame
-   * @param responseColumn
-   * @param threshold
+   *  Constructor for selection of categorical columns strategy based on threshold value
+   * @param frame the frame selection is being done from
+   * @param excludedColumnNames the column names we want to exclude from the result 
+   *                            ( i.e. response column for classification tasks , fold column in case it is categorical etc.)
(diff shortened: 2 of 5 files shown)


michalkurka:
@deil87 this code doesn't seem to be referenced from anywhere, is there a PR that requires this change?

deil87:
> @deil87 this code doesn't seem to be referenced from anywhere, is there a PR that requires this change?

@michalkurka Sorry, I was about to add this dependency but you were faster :) It is part of the integration of TE into AutoML https://github.com/h2oai/h2o-3/pull/3183

deil87:
@seb-h2o @michalkurka please feel free to review this small PR. We can merge it after upcoming major release.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/AllCategoricalTEApplicationStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+  
+  public AllCategoricalTEApplicationStrategy(Frame frame, String responseColumnName) {
+    _frame = frame; 
+    _responseColumnName = responseColumnName;
+  }
+  
+  public String[] getColumnsToEncode() {
+    ArrayList<String> categoricalColumnNames = new ArrayList<>();
+    for( String vecName : _frame.names()) {
+      if(_frame.vec(vecName).isCategorical() && !vecName.equals(_responseColumnName))
```
seb-h2o: should we not filter out non-predictors columns like `fold_column`, `weights_column`, `offset_column` as well?

seb-h2o: well, those can never be categorical... even `fold_column`?

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/ThresholdTEApplicationStrategy.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * 
+   * @param frame
+   * @param responseColumn
+   * @param threshold
+   */
+  public ThresholdTEApplicationStrategy(Frame frame, Vec responseColumn, long threshold) {
+    _frame = frame; 
+    _responseColumnName = _frame.name(_frame.find(responseColumn)); // Question: can we add .name(Vec vec) to Frame's API?
```
seb-h2o: I'd be in favour of answering "yes" to your question.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/AllCategoricalTEApplicationStrategy.java
seb-h2o: from doc:
```
The fold column can include integers (0, 1, 2, … , N-1 values or 1, 2, 3… , N values) or categorical values.
```

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/ThresholdTEApplicationStrategy.java
michalkurka: `.name(Vec vec)`? I don't know what that would be good for

this is imho recipe for disaster:

     _frame.name(_frame.find(responseColumn))

you shouldn't be using the Vec to reverse look-up its name - for many reasons, including the very strange one that the Frame actually _can_ have 2 different columns backed by the same Vec

I would say this constructor should instead look like this:

    public ThresholdTEApplicationStrategy(Frame frame, String responseColumn, long threshold)

then there is no issue with reverse look-up and the API is impossible to use the wrong way

michalkurka: (I also have the same objections toward `.find(Vec vec)` - which is an existing method - only used in MRUtils (btw) - I don't think that one is a good idea in general either

seb-h2o: but you have the same problem with `name` or `key` then, don't we?
`find(Vec)` is doing a lookup by Vec key, and so is `find(Key)`.
Btw, do we actually prevent duplicate Vec names in a frame? I'm not 100% sure it's the case, at least according to `Frame` constructor.

seb-h2o: Ok, convinced now. 
I would suggest to deprecate those public methods that are not recommended then, if they can lead to subtle bugs.
Also, regarding the signature, and given the usage mane of response column, it looks like it could be profitably replaced by `String[] excludedColumnNames` as not only response, but also `foldColumn` will need to get the same logic applied.
Same for `AllCategoricalTEApplicationStrategy` constructor.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/AllCategoricalTEApplicationStrategy.java
deil87: @seb-h2o I did filtering ( _responseColumnName) but I did not know that fold_column can be categorical. Thanks! I will consider to add `excludeColumnNames` parameter as you suggested because yes... it is already a second column we need to filter out.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/ThresholdTEApplicationStrategy.java
deil87: @michalkurka 
> you shouldn't be using the Vec to reverse look-up its name - for many reasons, including the very strange one that the Frame actually can have 2 different columns backed by the same Vec

I was relying on existing API. If we don't want to add `.name(Vec)` for mentioned above reasons then we should get rid of(deprecate) `.find(Vec)` and`.find(Key)`. So I agree with @seb-h2o regarding deprecation.

Regarding reverse look-up.... My first thought was that Vec should maybe know its name. But ok, design is different: Frame is a holder of vecs and maps names to vecs. This gives us a freedom to have multiple columns being backed by the same Vec - ok.  But if we are ok with  this design... I think it is eligible in general to ask Frame about "what is the name for a given Vec you have". So I would say reverse lookup on its own is a fine idea. But we allow columns to share Vecs and that is what makes this lookup dangerous. 

> I would say this constructor should instead look like this:

I would do it this way of course but the problem is that in AutoML we store responseColumn as a Vec (L224) and I would have to do this reverse lookup anyway either outside of the `TEApplicationStrategy` or inside. And as there were no other usages for `responseColumnName` as a String outside of the scope of `TEApplicationStrategy` I decided to move this reverse lookup logic inside the component increase cohesion.

deil87: @seb-h2o Yeap. That would be enough! Thanks!

seb-h2o: I would say, in `AutoML`, feel free to change the package protected attributes:
```
Vec _responseColumn -> String _responseColumnName
```
and modify the public corresponding accessors accordingly:
```
  public Vec getResponseColumn() { return _trainingFrame.vec(_responseColumnName); }

```

deil87:
@seb-h2o excludeColumnNames parameter was added as you suggested. Added test for case when fold column is categorical.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/strategy/ThresholdTEApplicationStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+   * @param threshold categorical columns with higher cardinality than {@code threshold} value will be selected
+   */
+  public ThresholdTEApplicationStrategy(Frame frame,  long threshold, String[] excludedColumnNames) {
+    _frame = frame;
+    _excludedColumnNames = excludedColumnNames; 
+    _threshold = threshold;
+  }
+  
+  public String[] getColumnsToEncode() {
+    return Arrays.stream(_frame.names())
```
michalkurka: @seb-h2o @deil87 This is not an ideal design. TE can only be applied to categorical columns and "excludedColumnNames" seem to be repeated in both strategies. In that case, it is better to have one class responsible for filtering the columns and the strategy will just be applied on a column level - that is strategy will have on the input only single column and will only output yes/no for the given column.

deil87: @michalkurka 
> "excludedColumnNames" seem to be repeated in both strategies

yes, I was thinking about this fact.  I wanted to move it to abstract `TEApplicationStrategy` class to highlight its importance in all possible strategies. But sometimes it feels better when you have field to be defined in the class where it is actually used ( in our case in the constructors of inherited classes). Probably it was not you point here but just wanted to share my thoughts with respect to this parameter. I also had a desire not to duplicate filtering part but having separate class for just filtering is an overhead especially when we don't have other use cases for such a class.

> In that case, it is better to have one class responsible for filtering the columns and the strategy will just be applied on a column level - that is strategy will have on the input only single column and will only output yes/no for the given column

How would you represent for example AllCategoricalsStrategy with suggested design? Abstraction that I had in mind was broader than just "column -> yes/no". We can have ratio strategy for example. We also can select categorical columns to encode based on relative cardinality of them...let's say based on average cardinality. I don't think that we should restrict ourselves with just single column as an input.

PR #3976, opened 2019-10-14
Don't restrict Hadoop smoke tests to `micro` machines

michalkurka:
This is blocking SW

### DIFF
--- a/scripts/jenkins/groovy/defineTestStages.groovy
+++ b/scripts/jenkins/groovy/defineTestStages.groovy
@@ -382,7 +382,7 @@ def call(final pipelineContext) {
         kerberosPrincipal: 'HTTP/localhost@H2O.AI',
         kerberosConfigPath: 'scripts/jenkins/config/kerberos.conf',
         kerberosPropertiesPath: 'scripts/jenkins/config/kerberos.properties',
-      ], pythonVersion: '2.7', nodeLabel: 'docker && micro',
+      ], pythonVersion: '2.7',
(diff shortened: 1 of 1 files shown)


michal-raska:
The reason why Hadoop tests were pinned to micro nodes is, that when running on Jenkins node with multiple executors the Hadoops (from different jobs) tried to communicate. Docker has the inter-container communication (https://docs.docker.com/v17.09/engine/userguide/networking/default_network/container-communication/#communication-between-containers) by default. That's why the hadoop instances in different containers can communicate. There was a plan to disable this (it's a daemon param when starting the docker), but I'm not sure if this is implemented on all machines. Anyway I think this is worth a try, we can fix it properly (on level of hadoop configuration) if we see issues.

michal-raska:
I'm not sure that the tests do not download some datasets... but that can be avoided for sure. Otherwise the `--network=none` seems like a good option.

michalkurka:
@michal-raska they shouldn't communicate - multicast is disabled in Hadoop mode

maybe it was done for the "H2O Standalone on Hadoop"

michal-raska:
@michalkurka I meant Hadoop itself, not H2O-3

michalkurka:
> @michalkurka I meant Hadoop itself, not H2O-3

ok, thanks for the info - it is a bit puzzling though - is this because of port conflicts?

michalkurka:
@honzasterba IMHO we cannot disable networking completely because we are also testing  that S3 subsystem is not broken

PR #3980, opened 2019-10-15
[WIP/POC] PUBDEV-6614: predict OOME when parsing and stop if in danger

honzasterba:
JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6614

For large datasets it seems the starting estimate can sometimes be very innacurate, this accuracy gets better as time goes on, but can be very much off until about 40% of the frame is loaded.

For this reason I think loading should not be terminated until we get to some threshold of dangerously low available memory.

Suggestions welcome.

### DIFF
--- a/h2o-core/src/main/java/water/fvec/NewChunk.java
+++ b/h2o-core/src/main/java/water/fvec/NewChunk.java
@@ -112,6 +112,17 @@ public class NewChunk extends Chunk {
       else if (_vals4 != null) _vals4 = Arrays.copyOf(_vals4, len);
       _len = len;
     }
+
+    public long byteSize() {
+      long mem = 4;
+      if (_vals1 != null) {
+        mem += _vals1.length;
+      }
+      if (_vals4 != null) {
+        mem += _vals4.length * 4;
+      }
+      return mem;
+    }
   }
 
   /**
@@ -218,6 +229,20 @@ public class NewChunk extends Chunk {
       else if(_vals4 != null) _vals4 = Arrays.copyOf(_vals4,len);
       else if(_vals8 != null) _vals8 = Arrays.copyOf(_vals8,len);
     }
+
+    public long byteSize() {
+      long mem = 4;
+      if (_vals1 != null) {
+        mem += _vals1.length;
+      }
+      if (_vals4 != null) {
+        mem += _vals4.length * 4;
+      }
+      if (_vals8 != null) {
+        mem += _vals8.length * 8;
+      }
+      return mem;
+    }
   }
 
   public final int _cidx;
@@ -239,6 +264,33 @@ public class NewChunk extends Chunk {
   private transient double _ds[];   // Doubles, for inflating via doubles
   public transient byte[]   _ss;   // Bytes of appended strings, including trailing 0
   private transient int    _is[];   // _is[] index of strings - holds offsets into _ss[]. _is[i] == -1 means NA/sparse
+    
+  public long byteSize() {
+    long mem = 0;
+    if (_ms != null) {
+      mem += _ms.byteSize();
+    }
+    if (_missing != null) {
+      mem += _missing.size() / 8; // size() returns number of bits, convert to bytes, approx
+    }
+    if (_xs != null) {
+      mem += _xs.byteSize();
+    }
+    if (_id != null) {
+      mem += _id.length * 4;
+    }
+    if (_ds != null) {
+      mem += _ds.length * 8;
+    }
+    if (_ss != null) {
+      mem += _ss.length;
+    }
+    if (_is != null) {
+      mem += _is.length * 8;
+    }
+    return mem;
+  }
+
 
   int   [] alloc_indices(int l)  { return _id = MemoryManager.malloc4(l); }
   public double[] alloc_doubles(int l)  {
@@ -295,7 +347,6 @@ public class NewChunk extends Chunk {
     _ms = new Mantissas(4);
     _xs = new Exponents(4);
     if(sparse) _id = new int[4];
-
   }
 
   public NewChunk(double [] ds) {
--- a/h2o-core/src/main/java/water/parser/FVecParseWriter.java
+++ b/h2o-core/src/main/java/water/parser/FVecParseWriter.java
@@ -6,6 +6,7 @@ import water.fvec.AppendableVec;
 import water.fvec.NewChunk;
 import water.fvec.Vec;
 import water.util.ArrayUtils;
+import water.util.FrameSizeMonitor;
 
 import java.util.Arrays;
 import java.util.UUID;
@@ -27,17 +28,18 @@ public class FVecParseWriter extends Iced implements StreamParseWriter {
   private final Vec.VectorGroup _vg;
   private long _errCnt;
   int[] _parse_columns_indices;
+  private final FrameSizeMonitor _monitor;
 
   public FVecParseWriter(Vec.VectorGroup vg, int cidx, Categorical[] categoricals, byte[] ctypes, int chunkSize,
                          AppendableVec[] avs) {
-    this(vg, cidx, categoricals, ctypes, chunkSize, avs, null);
+    this(vg, cidx, categoricals, ctypes, chunkSize, avs, null, null);
   }
   // note that if parse_columns_indices==null, it implies all columns are parsed.
   public FVecParseWriter(Vec.VectorGroup vg, int cidx, Categorical[] categoricals, byte[] ctypes, int chunkSize,
-                         AppendableVec[] avs, int[] parse_columns_indices) {
+                         AppendableVec[] avs, int[] parse_columns_indices, FrameSizeMonitor monitor) {
 
     boolean ctypesShrunk = false;
-         // Required not-null
+    // Required not-null
     if ((parse_columns_indices!=null) && (categoricals!=null) &&
             (parse_columns_indices.length == categoricals.length)) { // for nextChunk() calls in gzip/zip parser
       _ctypes = ctypes;
@@ -83,6 +85,10 @@ public class FVecParseWriter extends Iced implements StreamParseWriter {
     _cidx = cidx;
     _vg = vg;
     _chunkSize = chunkSize;
+    _monitor = monitor;
+    if (_monitor != null) {
+      _monitor.register(this);
+    }
   }
 
   @Override public FVecParseWriter reduce(StreamParseWriter sdout){
@@ -111,15 +117,20 @@ public class FVecParseWriter extends Iced implements StreamParseWriter {
   }
   @Override public FVecParseWriter close(Futures fs){
     if( _nvs == null ) return this; // Might call close twice
+    long mem = 0;
     for(int i=0; i < _nvs.length; i++) {
       _nvs[i].close(_cidx, fs);
+      mem += _nvs[i].chk2().byteSize();
       _nvs[i] = null; // free immediately, don't wait for all columns to close
+    }
+    if (_monitor != null) {
+      _monitor.closed(this, mem);
     }
     _nvs = null;  // Free for GC
     return this;
   }
   @Override public FVecParseWriter nextChunk(){
-    return  new FVecParseWriter(_vg, _cidx+1, _categoricals, _ctypes, _chunkSize, _vecs, _parse_columns_indices);
+    return new FVecParseWriter(_vg, _cidx+1, _categoricals, _ctypes, _chunkSize, _vecs, _parse_columns_indices, _monitor);
   }
 
   @Override public void newLine() {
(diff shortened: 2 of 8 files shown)


Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6614

h2o-core/src/main/java/water/fvec/NewChunk.java
```diff
@@ -239,6 +264,33 @@ public void resize(int len) {
...
   public transient byte[]   _ss;   // Bytes of appended strings, including trailing 0
   private transient int    _is[];   // _is[] index of strings - holds offsets into _ss[]. _is[i] == -1 means NA/sparse
+    
+  public long byteSize() {
+    long mem = 0;
+    if (_ms != null) {
+      mem += _ms.byteSize();
+    }
+    if (_missing != null) {
+      mem += _missing.size() / 64;
```
Pscheidl: The internal array is long[] (is that guaranteed ?), but the `size()` method is supposed to return total amount of **bits** allocated. By diving it by 64, don't we get the length of the `long[]` internal array ? In other words, don't we get the `.length` of that array effectively ? If we want bytes, shouldn't we just divide by 8 ?

Sorry if there is something I'm missing.

honzasterba: good catch!

honzasterba: the fact that internaly it uses long is not guarateed, all this is on a best-effort basis

h2o-core/src/main/java/water/util/FrameSizeMonitor.java
```diff
@@ -0,0 +1,151 @@
...
+                    job.stop();
+                    break;
+                }
+                if (nextProgress < 0.1f) {
+                    nextProgress *= 2;
+                }
+                System.out.println("NEXT " + nextProgress);
+            }
+            try {
+                Thread.sleep(SLEEP_MS);
```
Pscheidl: Rarely, this might not be enough. The parsers tends to be really quick. 

We could do the checks before any memory is allocated directly in the thread touched. But that would require additional overhead, even if the monitoring feature would be disabled. Which is probably undesirable in high-perf core code. I guess hotspot could eliminate the performance impact once it kicks in.

Pscheidl: VERY rarely. Very. Just thinking out loud and pointing out the weaknesses I believe @honzasterba is well aware of.

Pscheidl: By the "affected thread", I mean the NewChunk itself - doing the test there before allocating new memory or doing memory changes (sparsity <-> density transitions etc.) would make sense ?

h2o-core/src/main/java/water/util/FrameSizeMonitor.java
```diff
@@ -0,0 +1,151 @@
...
+                if (writer.getNvs() != null) {
+                    writers.put(writer, getUsedMemory(writer));
+                }
+                usedMemory += e.getValue();
+            }
+        }
+        return usedMemory;
+    }
+
+    private long getUsedMemory(FVecParseWriter writer) {
```
Pscheidl: Minor: I can imagine making these methods publicly available in a different class for others to use.

h2o-core/src/main/java/water/util/FrameSizeMonitor.java
Pscheidl: One adjustment I can think of is to only use `SLEEP_MS` as a default value and adjust it based on the relative speed of the parser against free memory. This would be relatively cheap and easy check that could make this almost bulletproof. Not going over 100ms (current constant value, or any other), but in well-reasoned cases shorten the interval to be 100% sure.

h2o-core/src/main/java/water/util/FrameSizeMonitor.java
honzasterba: yes, but thats out of scope

h2o-core/src/main/java/water/util/FrameSizeMonitor.java
honzasterba: actually checking more frequently does not have much benefit because the progress does not change that much and also memory is usually allocated in big batches, to avoid false positives I have added a check for overall memory usage and only terminate when a certain overall memory usage is reached and it seems the predicted size will not fit

h2o-core/src/main/java/water/parser/ParseDataset.java
```diff
@@ -1010,11 +1009,9 @@ private void freeMem(Chunk in) {
       }
       @Override public void reduce(DistributedParse dp) {
         _dout.reduce(dp._dout);
-
       }
 
       @Override public void postGlobal() {
-        super.postGlobal();
```
honzasterba: noop

h2o-core/src/main/java/water/util/FrameSizeMonitor.java
```diff
@@ -0,0 +1,178 @@
...
+    public FrameSizeMonitor(Key<Job> jobKey) {
+        this.jobKey = jobKey;
+    }
+
+    public static FrameSizeMonitor start(Key<Job> jobKey) {
+        FrameSizeMonitor task = new FrameSizeMonitor(jobKey);
+        if (ENABLED) {
+            H2O.submitTask(new LocalMR(task, 1));
+        }
+        return task;
```
honzasterba: maybe it would be better if this returned null in case the feature is not enabled, the rest of the code is ready for that

h2o-core/src/main/java/water/util/FrameSizeMonitor.java
```diff
@@ -0,0 +1,178 @@
...
+
+    @Override
+    protected void map(int id) {
+        float nextProgress = 0.02f;
+        Job<Frame> job = DKV.getGet(jobKey);
+        while (job.isRunning() && nextProgress < 1f) {
+            //job = DKV.getGet(jobKey); // refresh progress
+            float currentProgress = job.progress();
+            if (currentProgress >= nextProgress) {
+                if (isMemoryUsageOverLimit() && isFrameSizeOverLimit(currentProgress, job)) {
```
honzasterba: to avoid false positives I have added a check for overall memory usage and only terminate when a certain overall memory usage is reached and it seems the predicted size will not fit

h2o-core/src/main/java/water/fvec/NewChunk.java
```diff
@@ -239,6 +264,33 @@ public void resize(int len) {
   private transient double _ds[];   // Doubles, for inflating via doubles
   public transient byte[]   _ss;   // Bytes of appended strings, including trailing 0
   private transient int    _is[];   // _is[] index of strings - holds offsets into _ss[]. _is[i] == -1 means NA/sparse
+    
+  public long byteSize() {
```
Tagar: if we have some categoricals stored as strings.. will the logic to detect that string size be somewhere here too? 
empty strings in jvm for example use up 40 bytes. 
https://www.javaworld.com/article/2077496/java-tip-130--do-you-know-your-data-size-.html

honzasterba: string data is stored in _ss and _is and added to the sum

h2o-core/src/main/java/water/parser/FVecParseWriter.java
```diff
@@ -111,15 +117,20 @@ public FVecParseWriter(Vec.VectorGroup vg, int cidx, Categorical[] categoricals,
   }
   @Override public FVecParseWriter close(Futures fs){
     if( _nvs == null ) return this; // Might call close twice
+    long mem = 0;
     for(int i=0; i < _nvs.length; i++) {
       _nvs[i].close(_cidx, fs);
+      mem += _nvs[i].chk2().byteSize();
```
michalkurka: Why do we need the "byteSize" method on NewChunk? At this point, NewChunk is already compressed (close method was called) and we know the exact size in memory.

I don't think I understand the concept.

h2o-core/src/main/java/water/parser/FVecParseWriter.java
```diff
@@ -83,6 +85,10 @@ public FVecParseWriter(Vec.VectorGroup vg, int cidx, Categorical[] categoricals,
     _cidx = cidx;
     _vg = vg;
     _chunkSize = chunkSize;
+    _monitor = monitor;
+    if (_monitor != null) {
+      _monitor.register(this);
```
michalkurka: This will increase memory usage... IMHO by a lot: essentially this causes the `_nvs` data to stay in memory even when we are not working with that chunk anymore.

h2o-core/src/main/java/water/parser/FVecParseWriter.java
honzasterba: at this point byteSize is 0, we need byteSize to compute size of newChunk in memory before it gets compressed, after newchunk is compressed we use the size of compressed chunk

h2o-core/src/main/java/water/parser/FVecParseWriter.java
honzasterba: _nvs is set to null after chunks are closed so the data should be freed from memory once not needed, this might however keep the FVecParseWriter in memory for longer
any suggestion of a better way of how to keep track of all the created chunks and their size is welcome

h2o-core/src/main/java/water/fvec/NewChunk.java
Tagar: great. thanks for confirming @honzasterba

h2o-core/src/main/java/water/parser/FVecParseWriter.java
michalkurka: okay, I see - you are monitoring even the chunks that are just being written - shooting for the stars :)

now I am closer to the understanding

h2o-core/src/main/java/water/parser/FVecParseWriter.java
michalkurka: yep - I just checked that, close will kill the _nvs array, mea culpa

in that case - do we need to to expose `getNvs`, can there eg. be a race condition in https://github.com/h2oai/h2o-3/pull/3980/files#diff-ad31e0d8480e348d69e75a53b763a9ffR140? 

you check that getNvs is not null, but that can change before the next method is executed - synchronizing on writers won't help

wendycwong:
@honzasterba:

I am wondering if you may be able to estimate the file size during parse setup where you actually will parse the file to determine the column type?  It the column types are known, you can estimate the memory size per row.  Then, if you can monitor the size of the row change to determine if you can quit early and won't parse the frame?

Just a wild guess here.

Wendy

h2o-core/src/main/java/water/parser/FVecParseWriter.java
honzasterba: you are correct, there is a tricky race condition, will look into it

h2o-core/src/main/java/water/parser/FVecParseWriter.java
honzasterba: interesting point, counting just finished chunks would certainly simplify things, but during parsing in-progress chunks take significant amount of memory so I would like to keep this

h2o-core/src/main/java/water/parser/FVecParseWriter.java
honzasterba: fixed

h2o-core/src/main/java/water/fvec/NewChunk.java
```diff
@@ -239,6 +264,33 @@ public void resize(int len) {
...
+    long mem = 0;
+    if (_ms != null) {
+      mem += _ms.byteSize();
+    }
+    if (_missing != null) {
+      mem += _missing.size() / 8; // size() returns number of bits, convert to bytes, approx
```
Tagar: Perhaps this is a bit optimistic?
One `BitSet` is at least 4 bytes for `wordsInUse`
https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/BitSet.java#L95
plus 4 bytes for `sizeIsSticky`
https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/BitSet.java#L101
plus 24 bytes for headers overhead for the `long[] words` ?
https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/BitSet.java#L90
see https://stackoverflow.com/questions/26874631/how-much-memory-does-a-java-array-use
I hope I get this right?
Would be great to see if all these estimates correlate well with actual jvm memory usage. 
Thanks !

PR #3982, opened 2019-10-16
replace nonexistent 'continue' keyword in R with 'next'

seb-h2o:
was causing test errors in some conditions, for example in `PUBDEV_5705_drop_columns_parser_arff_large.R`:
```
>
[2019-10-16 15:09:31] [ERROR] : Error: Test failed: 'Test ARFF Parse with skipped columns'
* object 'continue' not found
1: withWarnings(test())
2: withCallingHandlers(expr, warning = wHandler)
3: test()
4: assertCorrectSkipColumns(fileName, fullFrameR, skip99Per, TRUE, h2o.getTypes(f1))
```


seb-h2o:
@honzasterba it's used in test utils only, and the condition where is was used is apparently rarely met...

PR #3987, opened 2019-10-17
Fix a test class to be compatible with Python 2.7 syntax

michalkurka:


Pscheidl:
You know, that's just me. I always get carried away by dreaming of thing I cannot have. Always lost in those fantasies where platforms are type-safe and compilers provide as much safety as they can by making me type one more lexeme...

michalkurka:
yep, seems like Mr. Marty McFly-Pscheild came from some kind of a future universe or something

Pscheidl:
https://docs.python.org/3/library/typing.html

michalkurka:
@Pscheidl I cannot access that link with my Netscape Navigator

Pscheidl:
I'll extract the text and ship it to you on a persistent storage ? What do you Apple folks have these days ? Is 5" floppy ok ?

michalkurka:
@Pscheidl My PowerBook 1400 prefers Zip drives. Please ship it over using Federal Express.

PR #3989, opened 2019-10-17
PUBDEV-6945: add instruction RunXGBoostNativeHadoopREADME.txt

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6945?filter=-1

H2O XGBoost results are not reproducible over multiple nodes.  However, if the native XGBoost results are not reproducible across multiple nodes are not reproducible, then, there is nothing we can do here at H2O about it.  It needs to be fixed at the root.

I have made a readme file which contains instructions on how to run native XGBoost off hadoop.

### DIFF
--- a/scripts/RunXGBoostNativeHadoopREADME.txt
+++ b/scripts/RunXGBoostNativeHadoopREADME.txt
@@ -1,0 +1,70 @@
+This file shows all steps need to run xgboost off Hadoop on mr-0xd2.  All AWS keys and stuff have been set up on this machine.  In addition, I put the data in the following place: s3://h2o-public-test-data/xgboost-dist/.  Do a s3cmd ls s3://h2o-public-test-data/xgboost-dist/ to see what directories and files are there.  You will notice that there are airlines and higgs directories containing airlines and higgs dataset in the correct format for xgboost in ingest.  
+
+For XGBoost 0.82, they use Python3.5 or higher.  However, there are several errors concerning bytes-object and string.  Need to replace the following codes:
+1. yarn.py, line 48, replace with out = out.decode('utf-8').split('\n')[0].split()
+2. yarn.py, line 53, replace classpath = classpath.strip() with classpath = classpath.decode('utf-8').strip()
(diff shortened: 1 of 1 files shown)


scripts/RunXGBoostNativeHadoopREADME.txt
```diff
@@ -0,0 +1,70 @@
+This file shows all steps need to run xgboost off Hadoop on mr-0xd2.  All AWS keys and stuff have been set up on this machine.  In addition, I put the data in the following place: s3://h2o-public-test-data/xgboost-dist/.  Do a s3cmd ls s3://h2o-public-test-data/xgboost-dist/ to see what directories and files are there.  You will notice that there are airlines and higgs directories containing airlines and higgs dataset in the correct format for xgboost in ingest.
```
honzasterba: I think it would be very useful to document how these datasets were produced.

scripts/RunXGBoostNativeHadoopREADME.txt
```diff
@@ -0,0 +1,70 @@
...
+# Task Parameters
+# the number of round to do boosting
+num_round = 50
+#num_boost_round = 500
+# 0 means do not save any model except the final round model
+save_period = 0
+# The path of training data
+data = "s3://h2o-public-test-data/xgboost-dist/airlines/train"
+eval[test]="s3://h2o-public-test-data/xgboost-dist/airlines/train"
+model="s3://h2o-public-test-data/xgboost-dist/airlines/model1"
```
honzasterba: shoud be model_dir=

scripts/RunXGBoostNativeHadoopREADME.txt
wendycwong: Honza:

I only have the higgs and airline dataset conversion and feel that once you wrote your general dataset conversion, that will be more useful than what I have here.  Let me know what you think.  I can easily added those two scripts I sent you here.

scripts/RunXGBoostNativeHadoopREADME.txt
wendycwong: Fixed!  Thank you for the catch.

PR #3990, opened 2019-10-17
PUBDEV-6774 - Parallelize Grid Search

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6774

Besides new code, there is one refactor done in this PR: Each of the hyperspace walkers is now a separate class. Other people are currently adding more and having them all in one interface resulted in insanely long file, which was hard to orient in.

I'm using a very basic strategy: the "leader node" determines some level of parallelism (or the user sets it). Then it walks the hyperspace and tries to keep that given level of parallelism (number of models built in parallel) until the hyperparameter space is fully explored. Currently, the strategy is simple: once a model is finished building (or fails to build), a new model takes its place, keeping the same level of parallelism at all time.

There is a setting for H2O to determine the level of parallelism automatically, or it can be overridden by user.

**By default**, everything is sequential. Goal is for current users to see no difference in the API and behavior, unless explicitely enabled.


h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -0,0 +1,56 @@
+package hex;
+
+import jsr166y.ForkJoinTask;
+import java.util.*;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+public class ParallelModelBuilder extends ForkJoinTask<ParallelModelBuilder> {
```
Pscheidl: This should be moved to autoML module probably. Only a showcase, but fully functional. Will be instantiated inside a method and won't be serialized anywhere.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -0,0 +1,56 @@
...
+  private BiConsumer<Model, ParallelModelBuilder> _modelFeeder;
+  private Lock _waitingModelsLock;
+
+  public ParallelModelBuilder(BiConsumer<Model, ParallelModelBuilder> modelFeeder) {
+    Objects.requireNonNull(modelFeeder);
+    _modelFeeder = modelFeeder;
+    _waitingModelsLock = new ReentrantLock();
+  }
+
+  public void run(ModelBuilder[] modelBuilders) {
```
Pscheidl: To make this an isolated showcase, I had to remove few lines of code. This makes the impression the lock is there in vain. In this example, it is less useful then it seems, but it serves one purpose -prevents start of training of multiple models at once, in case two model feeders invoke this method at similar point in time.

h2o-algos/src/test/java/hex/grid/GridTest.java
```diff
@@ -89,6 +135,42 @@ public void testFaileH2OdParamsCleanup() throws FileNotFoundException {
     }
   }
 
+  @Test
+  public void testParallelGridSearch() {
```
Pscheidl: Please us this test to see the basic parallel approach that is now simply hardcoded for showcase.

The |hyperspace| = 12. There are quick-to-finish variants, as well as longer lasting ones (number of trees is the main factors affecting time in this test).

Pscheidl:
Also did first manual benchmark, really orientational. For really small and quickly trained models based on Iris with the following params:
```
      final Frame trainingFrame = parse_test_file("smalldata/iris/iris_train.csv");
      Scope.track(trainingFrame);

      final Integer[] ntreesArr = new Integer[]{50, 51, 52,53,54,55};
      final Integer[] maxDepthArr = new Integer[]{2, 3, 4, 5};
      HashMap<String, Object[]> hyperParms = new HashMap<String, Object[]>() {{
        put("_distribution", new DistributionFamily[]{DistributionFamily.multinomial});
        put("_ntrees", ntreesArr);
        put("_max_depth", maxDepthArr);
      }};
```

Both parallel and sequential ran separately with 20 iterations. Parallel approach with parallelism level of 10. The typical duration for parallel is `210 ms`, while for sequential it is `350` ms. Rough average from 20 runs. Looks like the improvement is there. The smaller the models are, the less is the difference, but for small models, level of parallelism can be increased - then the difference becomes higher again.

Pscheidl:
Did another, much longer benchmark of'

```
      final Frame trainingFrame = parse_test_file("./smalldata/testng/airlines_train.csv");
      Scope.track(trainingFrame);

      final Integer[] ntreesArr = new Integer[]{50, 51, 52, 53, 54, 55};
      final Integer[] maxDepthArr = new Integer[]{2, 3, 4, 5, 6, 7};
      HashMap<String, Object[]> hyperParms = new HashMap<String, Object[]>() {{
        put("_distribution", new DistributionFamily[]{DistributionFamily.multinomial});
        put("_ntrees", ntreesArr);
        put("_max_depth", maxDepthArr);
      }};
```

Results:
# Sequential
Took 9938 millis on average
# Parallel
Took 3757 millis on average

Same parallelism level as above (max 10 models at a time). More parallelism speeds it up even more.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -160,6 +160,61 @@ public long getModelCount() {
...
+    private final HyperSpaceWalker.HyperSpaceIterator<MP> hyperSpaceWalker;
+    private final Grid grid;
+    private final Lock lock = new ReentrantLock();
+
+    public ModelFeeder(HyperSpaceWalker.HyperSpaceIterator<MP> hyperSpaceWalker, Grid grid) {
+      this.hyperSpaceWalker = hyperSpaceWalker;
+      this.grid = grid;
+    }
+
+    private void feedModel(final Model model, final ParallelModelBuilder parallelModelBuilder) {
```
Pscheidl: Not cool. Really not cool, as it is responsible for manipulating internal state of the parallelModelBuilder, done this way only for simplicity.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -238,6 +242,11 @@ public void compute2() {
         Scope.exit();
       }
       tryComplete();
+      if(callback != null) {
+        final ModelBuildingResult modelBuildingResult = new ModelBuildingResult(Optional.of(_job.get()),
```
michalkurka: When can job for a model be null?

Pscheidl: This branch is not exceptional completion, which means the model will be there. The Optional is there for the consumer to honor the contract - The ModelBuildingResult might as well contain an exception.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -308,21 +317,28 @@ public void run() {
...
+      return _job.start(trainModelImpl(callback), _parms.progressUnits(), _parms._max_runtime_secs);
 
     // cross-validation needs to be forked off to allow continuous (non-blocking) progress bar
     return _job.start(new H2O.H2OCountedCompleter() {
                         @Override
                         public void compute2() {
                           computeCrossValidation();
                           tryComplete();
+                          if(callback != null) {
+                            final ModelBuildingResult modelBuildingResult = new ModelBuildingResult(Optional.of(_job.get()),
```
michalkurka: this API seems strange because it seems like there are 2 use cases handled by a single method:

1. successful completion
2. exceptional completion

ModelBuildingResult - first 2 arguments to this class seem mutually exclusive

why does the callback not have 2 methods for these 2 use cases?

Pscheidl: To only have one universal callback. We might have two as well - no problem.

Pscheidl: Introduced dual callback - one on failure and one on success.

Pscheidl:
Rebased on latest master - there were some changes related to `AutoMLBuildSpec` which uses hyperspace walkers.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1357,6 +1357,7 @@ public void computeImpl() {
     }
 
     @Override public boolean onExceptionalCompletion(Throwable t, CountedCompleter caller){
+      super.onExceptionalCompletion(t, caller);
       doCleanup();
       return true;
```
seb-h2o: Should it not return result from call to `super`?

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -15,12 +15,25 @@
...
 import java.util.*;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 
 /**
  *  Model builder parent class.  Contains the common interfaces and fields across all model builders.
  */
 abstract public class ModelBuilder<M extends Model<M,P,O>, P extends Model.Parameters, O extends Model.Output> extends Iced {
 
+  Consumer<Model> _onSuccessCallback = null;
```
michalkurka: What do you think about a single interface with 2 methods? IMHO that gives a little cleaner semantics. Is there a case where you would set just one of them?

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+   * one of the callbacks (on model failure / on model completion) is called.
+   *
+   * @param modelBuilders An {@link Collection} of {@link ModelBuilder} to execute in parallel.
+   */
+  public void run(final Collection<ModelBuilder> modelBuilders) {
+      for (final ModelBuilder modelBuilder : modelBuilders) {
+        _modelInProgressCounter.incrementAndGet();
+
+        // Set the callbacks
+        final Consumer<Model> onModelSuccessfulBuild = this::onModelSuccessfullBuild;
```
michalkurka: Are you sure the callbacks will never be serialized? It would be less than ideal to serialize them because it would require to serialize this whole class.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+   * {@link ParallelModelBuilder}. Any threads waiting for the instance of {@link ParallelModelBuilder} to finish (via join)
+   * will be notified.
+   */
+  public void noMoreModels() {
+    // Prevent completing te parallel model builder twice
+    final boolean previouslyCompleted = _completed.getAndSet(true);
+    if(previouslyCompleted) return;
+
+    // Do not block the caller
+    new Thread(() -> {
```
michalkurka: This is what AutoML is doing - it is not ideal.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -65,23 +65,28 @@
...
+ * @see HyperSpaceWalker
  * @see #startGridSearch(Key, HyperSpaceWalker)
  */
 public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSearch> {
   public final Key<Grid> _result;
   public final Job<Grid> _job;
+  public final int _parallelismLevel;
+
+  public static final int SEQUENTIAL_MODE = 1;
+  public static final int ADAPTIVE_PARALLEL_MODE = 0;
```
seb-h2o: Can't you use an enum instead here? Also, how is it different from `GridSearchStrategy` enum?
It's a bit annoying when having methods with many parameters to figure out what this `int` is... without mentioning the better error detection when using an enum.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -119,9 +124,9 @@ private GridSearch(Key<Grid> gkey, HyperSpaceWalker<MP, ?> hyperSpaceWalker) {
     long gridWork=0;
     if (gridSize > 0) {//if total grid space is known, walk it all and count up models to be built (not subject to time-based or converge-based early stopping)
       int count=0;
-      while (it.hasNext(model) && (it.max_models() > 0 && count++ < it.max_models())) { //only walk the first max_models models, if specified
+      while (it.hasNext(Optional.ofNullable(model)) && (it.max_models() > 0 && count++ < it.max_models())) { //only walk the first max_models models, if specified
         try {
-          Model.Parameters parms = it.nextModelParameters(model);
+          Model.Parameters parms = it.nextModelParameters(Optional.of(model));
```
seb-h2o: so model was nullable 2 lines above, and it isn't anymore?

h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -287,10 +287,10 @@ private void appendFailedModelParameters(final Key<Model> modelKey, final MP par
    * @param params model parameters which caused model builder failure
    * @params e  exception causing a failure
    */
-  void appendFailedModelParameters(final Key<Model> modelKey, final MP params, final Exception e) {
+  void appendFailedModelParameters(final Key<Model> modelKey, final MP params, final Throwable t) {
```
seb-h2o: is it necessary? are you catching Java Errors somewhere? It's not recommended in most cases...

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -136,7 +141,15 @@ private GridSearch(Key<Grid> gkey, HyperSpaceWalker<MP, ?> hyperSpaceWalker) {
     // Install this as job functions
     return _job.start(new H2O.H2OCountedCompleter() {
       @Override public void compute2() {
-        gridSearch(grid);
+
+        if (_parallelismLevel == 1) {
+            gridSearch(grid);
+        } else if (_parallelismLevel > 1 || _parallelismLevel == 0) {
```
seb-h2o: You should document somewhere the semantic of `parallelism = 0`.
I see below, that is means `parallelism = 2 * H2O.NUMCPUS`: it's worth being written explicitly.
Why the factor 2 by the way? Aren't model training already trying to use all CPUs?

h2o-core/src/main/java/hex/grid/hyperspace/HyperSpaceSearchCriteria.java
```diff
@@ -1,4 +1,4 @@
-package hex.grid;
```
seb-h2o: any backwards compatibility concern?
I think this is part of Java `API`

h2o-core/src/main/java/hex/grid/hyperspace/CartesianWalker.java
```diff
@@ -0,0 +1,125 @@
+package hex.grid.hyperspace;
```
seb-h2o: can you please point to the changes from the original impl?

h2o-core/src/main/java/hex/grid/hyperspace/RandomDiscreteValueWalker.java
```diff
@@ -0,0 +1,185 @@
+package hex.grid.hyperspace;
```
seb-h2o: same as for CartesianWalker, if any change from original, please show us.

h2o-py/h2o/grid/grid_search.py
```diff
@@ -175,6 +180,7 @@ def train(self, x=None, y=None, training_frame=None, offset_column=None, fold_co
         # dictionaries have special handling in grid search, avoid the implicit conversion
         parms["search_criteria"] = None if self.search_criteria is None else str(self.search_criteria)
         parms["export_checkpoints_dir"] = self.export_checkpoints_dir
+        parms["parallelism_level"] = self._parallelism
```
seb-h2o: I prefer `parallelism` everywhere: wondering why you have `parallelism_level` or `parallelismLevel` on backend/RestAPI

h2o-py/tests/testdir_algos/grid/pyunit_grid_parallel.py
```diff
@@ -0,0 +1,31 @@
...
+    # Run GBM Grid Search
+    ntrees_opts = [1,5]
+    hyper_parameters = OrderedDict()
+    hyper_parameters["ntrees"] = ntrees_opts
+    print("GBM grid with the following hyper_parameters:", hyper_parameters)
+
+    gs = H2OGridSearch(H2OGradientBoostingEstimator, hyper_params=hyper_parameters, parallelism = 1)
+    gs.train(x=list(range(4)), y=4, training_frame=train)
+    assert gs is not None
+    assert len(gs.model_ids) == len(ntrees_opts)
```
seb-h2o: we're checking that it works, but can we measure if parallelization is faster on small and not-that-small data?
I understand it was the original goal.

h2o-core/src/main/java/hex/grid/GridSearch.java
seb-h2o: gotcha this, is the `ADAPTIVE_PARALLELISM_MODE`: got confused by the `_MODE` suffix

h2o-core/src/main/java/hex/schemas/GridSearchSchema.java
```diff
@@ -122,6 +126,24 @@
...
+      } catch (NumberFormatException e) {
+        final String errorMessage = String.format("Could not parse given parallelism_level value: '%s' - not a number.",
+                parallelismLevelProperty);
+        throw new IllegalArgumentException(errorMessage, e);
+      }
+      parms.remove("parallelism_level");
+    } else {
+      this.parallelism_level = 0; // Operate in sequential mode by default
+    }
+
```
seb-h2o: Waw! Why is this impl of `fillFromParms` so complicated in this class? Why is it not done automatically like for other schemas? Someone needs to fix this.
Look at how complicated it is just to set a `int`.
With auto-binding, you would just have to check that the bound value is `>= 0`...

h2o-core/src/main/java/hex/grid/GridSearchStrategy.java
```diff
@@ -0,0 +1,5 @@
+package hex.grid;
+
+public enum GridSearchStrategy {
+  SEQUENTIAL, PARALLEL
```
seb-h2o: no need to shout :)

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -248,14 +379,7 @@ private void gridSearch(Grid<MP> grid) {
           grid.update(_job);
           final MP generalParams = _hyperSpaceWalker.getParams();
           if (generalParams._export_checkpoints_dir != null) {
```
seb-h2o: I'd remove this condition for clarity as it is repeated in the `attemptGridSave` method.

seb-h2o: I'd remove this condition for clarity as it is repeated in the attemptGridSave method.

h2o-py/tests/testdir_algos/grid/pyunit_grid_parallel.py
michalkurka: Yes - good point, also note that there will be a follow-up - the backend is going to change in order to run multiple model builders on _replicated_ dataset across the cluster (use case small data & many worker nodes).

We do want to increase performance / cluster utilization in this PR but there is more wrok coming :)

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -160,6 +173,124 @@ public long getModelCount() {
...
+    }
+
+    if(!startModels.isEmpty()) {
+      parallelModelBuilder.run(startModels);
+      parallelModelBuilder.join();
+    }
+    grid.update(_job);
+
+    final String export_checkpoints_dir = _hyperSpaceWalker.getParams()._export_checkpoints_dir;
+    if (export_checkpoints_dir != null) {
```
seb-h2o: also redundant with logic in `attemptGridSave`

h2o-core/src/main/java/hex/ParallelModelBuilder.java
seb-h2o: Which make me think that I should find some time to rewrite that part in AutoML some day..

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -160,6 +173,124 @@ public long getModelCount() {
...
+
+    for (int i = 0; i < mps.size(); i++) {
+      final MP nextModelParameters = mps.get(i);
+      final long checksum = nextModelParameters.checksum(IGNORED_FIELDS_PARAM_HASH);
+      if (grid.getModelKey(checksum) == null) {
+        startModels.add(ModelBuilder.make(nextModelParameters));
+      } else {
+        //TODO: Compensate for this scenario - we need to keep the level of paralelism at pre-set level
+      }
+    }
```
seb-h2o: not a big fan of this approach.
I think you could use an `ArrayBlockingQueue` limited to `parallelism`, and fill it continuously as soon as it has capacity (waiting when it's full) and as soon as you have next model to train.

seb-h2o: not a big fan of this approach.
I think you could use an ArrayBlockingQueue limited to parallelism, and fill it continuously as soon as it has capacity (waiting when it's full) and as soon as you have next model to train.

seb-h2o: not a big fan of this approach.
I think you could use an ArrayBlockingQueue limited to parallelism, and fill it continuously as soon as it has capacity (waiting when it's full) and as soon as you have next model to train.

seb-h2o: This would also solve your TODO

seb-h2o: another advantage of queues here: you don't need your callbacks to "know" what to do next:
like in `onModelBuildFailed`:
```      attemptBuildNextModel(parallelModelBuilder, null);```

this simplifies the organization of the code: communication between components is unidirectional, producer feeds the queue, consumer consumes, and possible a last queue consumes all the callbacks. No locks, no dialogue between components...

Comment (seb-h2o) [reply h2o-core/src/main/java/hex/grid/GridSearch.java]: I love queues, but it's really because they simplify things a lot most of the time.

File: h2o-py/tests/testdir_algos/grid/pyunit_grid_parallel.py (hunk 19 above)
Comment (Pscheidl) [reply h2o-py/tests/testdir_algos/grid/pyunit_grid_parallel.py]: We did some benchmarking and the performance gain varies. Especially in cases where one subspace (ro more) of the hyperspace result in slow-to-train models, creating bottlenecks in sequential mode, this speedup is significant (up to 50x speedup in extreme artificial cases). If there is a small hyperspace and models are really quick to train, the speedup can be as low as 2-3x.

Comment (Pscheidl) [reply h2o-py/tests/testdir_algos/grid/pyunit_grid_parallel.py]: But I'll come up with a benchmark, that's definitely planned.

File: h2o-algos/src/test/java/hex/tree/drf/DRFGridTest.java [hunk 25]
```diff
@@ -235,7 +235,7 @@ public void testRandomCarsGrid() {
       params._train = fr._key;
       params._response_column = "economy (mpg)";
       // Get the Grid for this modeling class and frame
-      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms);
+      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms, GridSearch.SEQUENTIAL_MODE);
```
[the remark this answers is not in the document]
honzasterba: why not keep the old  GridSearch.startGridSearch() that would call the new with SEQUENTIAL_MODE?

h2o-core/src/main/java/hex/ModelBuilder.java
honzasterba: but then you would not be able to use lambdas!

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Dispatcher for parallel model building. Starts building models every time the run method is invoked.
+ * After each model is finished building, the `modelFinished` method is invoked, which in turn invokes modelFeeder callback.
+ * ModelFeeder receives the model built and can deal with it in any way - e.g. put it into a Grid, or discard it if the resulting model
+ * is a failure. It also has the power to invoke the training of any number of new models. Or stop the parallel model builder,
+ * released the barrier inside.
+ */
+public class ParallelModelBuilder extends ForkJoinTask<ParallelModelBuilder> {
+
+  private BiConsumer<Model, ParallelModelBuilder> _onBuildSuccessCallback;
```
honzasterba: I think most of these could be final

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+   *
+   * @param modelBuilders An {@link Collection} of {@link ModelBuilder} to execute in parallel.
+   */
+  public void run(final Collection<ModelBuilder> modelBuilders) {
+      for (final ModelBuilder modelBuilder : modelBuilders) {
+        _modelInProgressCounter.incrementAndGet();
+
+        // Set the callbacks
+        final Consumer<Model> onModelSuccessfulBuild = this::onModelSuccessfullBuild;
+        modelBuilder.setOnSuccessCallback(onModelSuccessfulBuild);
```
honzasterba: why the variable, modelBuilder.setOnSuccessCallback(this::onModelSuccessfullBuild) is so much nicer

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+  public void noMoreModels() {
+    // Prevent completing te parallel model builder twice
+    final boolean previouslyCompleted = _completed.getAndSet(true);
+    if(previouslyCompleted) return;
+
+    // Do not block the caller
+    new Thread(() -> {
+      do {
+        try {
+          Thread.sleep(100);
```
honzasterba: bussy waiting is such an antipattern, if you have to spawn a thread here at least use wait()

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -65,23 +65,28 @@
...
- * @see hex.grid.HyperSpaceWalker
+ * @see HyperSpaceWalker
  * @see #startGridSearch(Key, HyperSpaceWalker)
  */
 public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSearch> {
   public final Key<Grid> _result;
   public final Job<Grid> _job;
+  public final int _parallelismLevel;
+
+  public static final int SEQUENTIAL_MODE = 1;
```
honzasterba: why not enum? I think we should have 2 arguments here instead of ints with hidden meaning - paralelism type (signle, parallel_adaptive, parallel_fixed) and an int argument for parallel_fixed case, better still dont use enum but use an object that contains all the info

h2o-core/src/main/java/hex/grid/GridSearch.java
honzasterba: agree with seb, this is confusing, see my comment above

h2o-core/src/main/java/hex/schemas/GridSearchSchema.java
```diff
@@ -122,6 +126,24 @@
...
+                  parallelism_level));
+        }
+      } catch (NumberFormatException e) {
+        final String errorMessage = String.format("Could not parse given parallelism_level value: '%s' - not a number.",
+                parallelismLevelProperty);
+        throw new IllegalArgumentException(errorMessage, e);
+      }
+      parms.remove("parallelism_level");
+    } else {
+      this.parallelism_level = 0; // Operate in sequential mode by default
```
honzasterba: sequential mode is 1 no? 0 is addaptive? see? ints are dangerous

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -160,6 +173,124 @@ public long getModelCount() {
     return _hyperSpaceWalker.getMaxHyperSpaceSize();
   }
 
+  private class ModelFeeder<MP extends Model.Parameters> {
```
honzasterba: I think this class is long enought and this does not need to be inner class

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Ints are dangerous and if you have a look at the commits, it used to have two arguments:
- Parallelism enabled/disabled
- Degree of parallelism

Then, I discussed this with more customer-facing people in the office and they all suggested libraries like scikit have integers only and this is more understandable. Hence my motivation to use integer.

Not sure which way is better. Personally, I like two params more.

h2o-algos/src/test/java/hex/grid/GridTest.java
```diff
@@ -23,12 +30,164 @@
...
+      Scope.track(trainingFrame);
+      GBMModel.GBMParameters parms = new GBMModel.GBMParameters();
+      parms._train = trainingFrame._key;
+      parms._distribution = AUTO;
+      parms._response_column = "IsDepDelayed";
+      parms._ntrees = 10;
+      final Lock lock = new ReentrantLock();
+
+      // Must be thread safe, otherwise won't work and will lock itself at the end.
+      final BiConsumer<Model, ParallelModelBuilder> onModelComplete = (model, parallelModelBuilder) -> {
```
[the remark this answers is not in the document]
Pscheidl: Is that a good thing or a bad thing to you ?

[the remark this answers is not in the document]
honzasterba: lambdas are great, when used correctly

h2o-core/src/main/java/hex/grid/GridSearchStrategy.java
Pscheidl: This is actually forgotten and obsolete from times when there were two parameters (strategy and degree of parallelism usable in parallel mode). - unless we want to implement two arguments for the users.

honzasterba: I am fine with keeping the py/R api int based but internally I would use something more high-level

h2o-algos/src/test/java/hex/grid/GridTest.java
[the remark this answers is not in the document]
Pscheidl: Are these used incorrectly ? Where is correctness of lada usage defined ?

h2o-core/src/main/java/hex/schemas/GridSearchSchema.java
Pscheidl: No idea, ask the original author. Refactoring this is not part of this PR. But you're always welcome to create a PR with a fix.

h2o-algos/src/test/java/hex/grid/GridTest.java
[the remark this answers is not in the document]
honzasterba: of course the correctnes is not defined anywhere, its based on personal prefference, my original comment was out of pure joy of finally seing lambdas used in our codebase and I thing this particular use qualifies as correct, dont worry.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Agreed. Will do.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
Pscheidl: This is indeed garbage, a little bit forgotten. The whole point of having everything even-driven was the lack busy waiting. Thanks for pointing it out, I was so blinded to leave it there.

I believe the whole parallel building can be completed easily on an even-driven base, without any waiting threads.

h2o-algos/src/test/java/hex/tree/drf/DRFGridTest.java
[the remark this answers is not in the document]
Pscheidl: Because then, someone would surely 100% show up complaining about having too much of these `startGridSearch` methods - we already have 3 of 'em. So, to reduce amount of these methods, I did this.

I personally prefer doing it as you say. So it will be done. Together with other changes (let's revert back the integer).

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
```diff
@@ -7,13 +7,11 @@
...
 import hex.Model.Parameters.FoldAssignmentScheme;
 import hex.ModelBuilder;
-import hex.ScoreKeeper;
 import hex.ScoreKeeper.StoppingMetric;
 import hex.ensemble.StackedEnsembleModel;
 import hex.grid.Grid;
 import hex.grid.GridSearch;
-import hex.grid.HyperSpaceSearchCriteria;
-import hex.grid.HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria;
+import hex.grid.hyperspace.HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria;
```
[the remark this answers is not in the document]
Pscheidl: If this is a problem, let's revert. It was just an attempt to refactor.

h2o-core/src/main/java/hex/ModelBuilder.java
Pscheidl: There is a way to satisfy both:

- Have an interface with two methods inside, with one setter.
- Introduce another "setter" method, accepting two lambdas, wrapping those in an anonymous class implementing the interface.

Sounds crazy even to me, but the price is low and people would be happy. I don't have a preference. Chose the way that was quickest for me to write.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
Pscheidl: Not now. ParallelModelBuilder is only spawned in the `parallelGridSearch` method and it's scope is lifetime of that method only.

But you have a great point - what about the callbacks in the Driver ? In future, the plan is to abandon chunking for smaller datasets and just do one model per node (or multiple per node). This might mean moving a lot of things locally - would such local callbacks still make sense ? Will we be forced to deprecate the Driver callbacks and use H2O's built in RPC to notify the leader node a model is ready ? 

Is the only purpose of the callback this PR and event-driven nature of the ParallelModelBuilder, which I find much better than "looping threads" ?

h2o-core/src/main/java/hex/ParallelModelBuilder.java
Pscheidl: The whole point of having ParallelModelBuilder was to have everything even-driven. I completely removed this piece of forgotten code. No more waiting threads.

h2o-core/src/main/java/hex/grid/Grid.java
Pscheidl: A little bit short-sighted of me. I used `Throwable`, because ModelBuilder's `onExceptionalCompletion` method uses `Throwable` as well - even though it might be a little bit more justified there. 

Yes, errors are nowhere caught as far as I can see in the model builders. But what if ? What if the model builder fails and calls `onExceptionalCompletion` with an `Error` (not an `Exception`) as an argument ? Shall we discard the error ? Basically, the job may fail due to the error, or it might be caught somewhere..but this method is about telling the grid which parameters failed and why. And the `why` is typed as `Throwable` in ModelBuilder.

1. Filter out errors and only pass exceptions ?
2. Modify ModelBuiler ?
3. Something else ?

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: I do agree with you. I am also for enums - those were originally there. There are various comments about that and let's address them all here.

What I'D like to do is to have i homogeneous in the client API and at the backend. Are you ok with 2 arguments, one enabling parallelism as a whole, defaulting to H2O-determined level of parallelism....and a second one, which makes it possible for our users to optionally set their own level of parallelism ?

h2o-core/src/main/java/hex/grid/hyperspace/HyperSpaceSearchCriteria.java
Pscheidl: Oh, is it ? Then we should leave it as-is I guess and not refactor it ever, or more precisely ion a backwards compatible way. If yes, I'll revert it.

h2o-core/src/main/java/hex/grid/hyperspace/CartesianWalker.java
Pscheidl: None besides this:

- The `Model` instance given to `hasNext` and `next` methods was allowed to be null. Now is Optional, which I believe shouts more clearly the model is optional. Both for users and future implementors.
- New methods for getting `n` intial models - those might be vended using a different strategy than calling `next` - especially they're guaranteed to know their surroundings and previous models.

Nothing else.

h2o-core/src/main/java/hex/grid/hyperspace/RandomDiscreteValueWalker.java
Pscheidl: Replied https://github.com/h2oai/h2o-3/pull/3990#discussion_r340614774

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Completely true. Fixed. Thanks.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
Pscheidl: Removed it from the code.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Thank you. That was really unnecessary. Removed completely.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -160,6 +173,125 @@ public long getModelCount() {
...
+      return params;
+    }
+  }
+
+  /**
+   * Searches the hyperspace and builds models in a parallel way - building the models in parallel.
+   *
+   * @param grid Grid to add models to
+   */
+  private void parallelGridSearch(final Grid<MP> grid) {
```
seb-h2o: I don't see any support for early stopping anywhere in this implementation, which would make it unusable for `RandomGridSearch`. Did I miss it?

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -160,6 +173,125 @@ public long getModelCount() {
...
+
+      return enoughTime;
+    }
+
+    private MP getNextModelParams(final HyperSpaceWalker.HyperSpaceIterator<MP> hyperSpaceWalker, final Model model, final Grid grid){
+      MP params = null;
+
+      while (params == null) {
+        if (hyperSpaceWalker.hasNext(Optional.ofNullable(model))) {
+          params = hyperSpaceWalker.nextModelParameters(Optional.ofNullable(model));
```
seb-h2o: In legacy
`public MP nextModelParameters(Optional<Model> previousModel)`
`public boolean hasNext(Optional<Model> previousModel)`
you're allowed to pass the previous model as if it was enough to decide about the next model params, or if there should be a next model. 
But this makes sense only in sequential mode. 
In parallel mode, the `previousModel` can be any of the models that was started until now. Only the Grid itself should be the source of truth to decide what's next.
Btw, none of the methods listed above is currently using this parameter, which is already suspicious... I would simply remove it.

h2o-algos/src/test/java/hex/tree/drf/DRFGridTest.java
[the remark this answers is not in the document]
michalkurka: @Pscheidl where does the user specify the level of parallelism? in `params`?

IMHO the mode of execution should be specified in the same place where the level of parallelism is defined.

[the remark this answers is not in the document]
Pscheidl: Your comments is a proof I made the API worse. `SEQUENTIAL_MODE` is a number constant with a value of 1. Level of parallelism of `1` means sequential. This will change. It's an easy change, so I'm saving it as a last one.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: The contract is set, the `model` argument is optional and with Java 8, we're only making it more obvious. The API does not state the model, if provided, is the previous model time-wise. 

There might be strategies based on previous model finished I can imagine, based on model's metrics for example. With model, parameters used to build that model are available. This implies the hyperspace walker knows which part of the hyperspace to avoid/focus on, based on model performance (as one example). Such strategies are usable even in parallel regime. Therefore, I'd not close our way to obtaining the information about the model built, even if it is currently unused.

I'd discuss my changes to the API as a whole. If this is part of Java API, then we might as well want to keep those walkers as they are right now, not adding the Optional, nor moving them to a separate package.

h2o-algos/src/test/java/hex/tree/drf/DRFGridTest.java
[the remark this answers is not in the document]
michalkurka: not to think too much ahead - but the parallel execution mode will most likely get more parameters - eg. to replicated the dataset to worker nodes

h2o-core/src/main/java/hex/grid/GridSearch.java
seb-h2o: This belongs to the Walkers, I'm not sure we can consider those as the Java API (at least not from a user perspective), and I doubt that there are custom implementations of those walkers (although possible why not..., we could always find a way to make the change backwards-compatible).

I agree that there might be strategies that would rely on built models to decide about next step.
My point is just that (and especially in parallel mode), the last built model is non significant, the strategy should rely on all previously built models... there's no reason to limit this to only one model.
To me it just shows that the legacy API was designed with `sequential` mode in mind.

seb-h2o: To be clear, what I had in mind was early stopping in parallel mode.
If it were implemented in the random walker `hasNext` instead of the way it is currently done, then you would have it for free in parallel mode too... somehow. But for this you need to pass the full grid, not only the last model.

Sure we can always imagine strategies that would rely only on some characteristics of one single model, but I can't see any interesting one tbh.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
seb-h2o: > Will we [...] use H2O's built in RPC to notify the leader node a model is ready ?

This looks like a good approach even now I think. You broadcast the jobId+status, and the listener has enough info to continue.

h2o-core/src/main/java/hex/grid/GridSearch.java
seb-h2o: @pavel
> I wanted to get rid of any threads waiting in loop to do something...

But there's no interruption every `t` millis with queues. It's event-driven, the event is just encapsulated by the queue itself... When the queue is full, the producer just sleeps until it get notified that it can add more. so that producer thread wouldn't be "busy-waiting". The consumer on the other side, could be the main/current thread without any problem.

> Of course, a callback doesn't need to know what to do next, but these callbacks are intended to know what to do next. They're free to invoke other builders

It's just a question of separation of concerns. Your current callbacks manage 2 concerns: (1) adding the new model to the grid, (2) trigger the build for the next model (this concern being also partly managed by your initialization code).
Queue clearly separate those concerns as it would take care fully and only of providing model builders.
Of course, there's always some logic needed to decide when to stop: and in this case, this should be the consumer=current thread=thread responsible to start the training jobs. It takes a ModelBuilder from the queue, checks it (for example, if queue was fed with `Optional.empty()`, then it stops everything) as well as the time/max models conditions. If conditions met, it starts training, otherwise, it can interrupt the producer thread...

> Current version, despite considered as "ugly" by some, gives the callback the will to spawn n new jobs at a time.

This would clearly break the SoC principle imo, I'd not recommend this. I think there should always be one single place deciding what to do next. 

> Thinking of it as a queue won't allow me to get rid of any business logic and locks. 

Correct, the feeding queue won't allow you to get rid of the locks that you need to fill the grid. Maybe my sentence was not clear... was just saying that callbacks themselves could be queued, in which case you don't need the locks.
But I don't see this last queue as interesting as the feeding one: outside getting rid of the locks, it doesn't bring much. The feeding queue, on the other side has more benefits.

> Or, the queue can be serviced completely thread-free by callbacks only

You can't do that with blocking queues.

michalkurka:
@Pscheidl please let us know when you are done incorporating the review changes - so that we can do a final review, thank you!

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+      return _throwable;
+    }
+
+    public Model.Parameters getParameters() {
+      return _parameters;
+    }
+  }
+
+  /**
+   * Indicate this builder there will be no more models. After this methods i called (by whoever calls it first),
```
honzasterba: typo in 'methods i called'

honzasterba: the javadoc does not seem to reflect what the method does

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+   * it waits for all other models to finish execution and then returns, completing the job represented by the very instance of
+   * {@link ParallelModelBuilder}. Any threads waiting for the instance of {@link ParallelModelBuilder} to finish (via join)
+   * will be notified.
+   */
+  public void noMoreModels() {
+    _completed.set(true);
```
honzasterba: the _complted var seems is never read

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+  
+  private void attemptComplete(){
+    if(_modelInProgressCounter.get() != 0) return;
```
honzasterba: can it happen that this 0 because a model has finished and new one has not started yet?

h2o-core/src/main/java/hex/grid/Grid.java
honzasterba: Error should never be caught, we should use Exception here

h2o-core/src/main/java/hex/grid/GridSearch.java
honzasterba: this comment needs to be resolved

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -136,7 +138,15 @@ private GridSearch(Key<Grid> gkey, HyperSpaceWalker<MP, ?> hyperSpaceWalker) {
...
     return _job.start(new H2O.H2OCountedCompleter() {
       @Override public void compute2() {
-        gridSearch(grid);
+
+        if (_parallelism == 1) {
+          gridSearch(grid);
+        } else if (_parallelism > 1) {
+          parallelGridSearch(grid);
+        } else {
+          throw new IllegalArgumentException(String.format("Grid search parallelism level must be >= 0. Give value is '%d'.",
```
honzasterba: error in message, the parallelsim here is expected to be >= 1
also I would move this check into constructor and have a simple if-else here

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
```diff
@@ -346,7 +344,8 @@ protected Work getAllocatedWork() {
                     baseParms,
                     searchParms,
                     new GridSearch.SimpleParametersBuilderFactory<>(),
-                    searchCriteria
+                    searchCriteria,
+                    GridSearch.getParallelismLevel(GridSearch.SEQUENTIAL_MODEL_BUILDING)
```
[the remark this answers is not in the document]
Pscheidl: Parallelism level is required, putting magical number is not the best option. I can either overload the method again or use just `GridSearch.SEQUENTIAL_MODEL_BUILDING`. Which one is preferred ? I can see no objective rule to decide.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -513,4 +686,38 @@ public MP build() {
...
+      return determineParallelismLevel();
+    } else {
+      return parallelism;
+    }
+  }
+
+  /**
+   * @return An integer with dynamically calculated level of parallelism based on Cluster's properties.
+   */
+  private static int determineParallelismLevel() {
```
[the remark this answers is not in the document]
Pscheidl: It's not confusing to me.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: This is for the case anyone will call this from tests. The check is on multiple places. I don't think it hurts to have it here.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
Pscheidl: I accidentally deleted another check.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
Pscheidl: Related to the situation below, I deleted it by accident. This breaks tests. Resolved.

h2o-algos/src/main/java/hex/glm/GLM.java
Pscheidl: No.

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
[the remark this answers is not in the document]
Pscheidl: All changes to these classes are reverted. I am not doing any refactoring whatsoever.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: I completely reverted all the changes to Hyperspace walkers, leaving the original behavior.

h2o-core/src/main/java/hex/grid/Grid.java
michalkurka: We are not actually catching Throwable, this comes from CountedCompleter - https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CountedCompleter.html

I think for now keeping the Throwable is fine - @Pscheidl is this a user-facing or just an internal API?

Pscheidl: Internal. Appending the failed results happens inside GridSearch.

Pscheidl: I am unable to explain this originates from core and in order to change it, we must take one of the actions mentioned above.

h2o-core/src/main/java/hex/ModelBuilder.java
michalkurka: @Pscheidl this is my thinking:

- ModelBuilder is Iced, the callbacks are not declared `transient` which can lead to some very non-obvious issues because Consumer is not even Serializable by default
- IMHO it is easier to document a custom interface - the only place where to document in this code seems to be the setter methods which is not ideal

For both practical and semantics reasons I am leaning toward a custom interface / abstract class

h2o-py/tests/testdir_algos/grid/pyunit_grid_parallel.py
michalkurka: @Pscheidl can you please make a jira for the benchmark? We shouldn't release this without a benchmark, however, since this is a new functionality in active development - it is okay to merge this PR without completing the benchmark (similar to what we did with TE).

h2o-core/src/main/java/hex/grid/Grid.java
michalkurka: As long as this is something we can change in the future without affecting users - I think we do not have to address it in this PR. Existing GridSearch catches Throwable - this should be revisited in a separate activity, not in the scope of this PR.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
michalkurka: Regarding the callbacks - I am arguing above in favor of making them a custom interface / abstract class to gain more control of what users can pass here.

h2o-core/src/main/java/hex/grid/GridSearch.java
michalkurka: Good question - is early stopping thought of @Pscheidl?

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: After reading through it, I still see no point a (fixed-sized) queue. All I need is `n` models training in parallel. And the `n` can change. Queue has no benefits over this model to me, other than subjective benefits of "better readability" and other stuff like that.

Concerns are separable here as well - just introduce another method to do it.

My scenario: Model finishes building -> model is saved -> `n` new models are started, potentially based on what has just been finished building.
Queue scenario: Model is finished building -> waiting thread is notified -> model or `n` new models are started, but taken from the queue, no direct (time-wise) feedback ( (model length, performance of the latest model ...).

Not saying queues are a bad thing, I just don't see they're somehow superior. If you're ok with it, I'd leave this part as-is.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Early stopping by walker not "vending" other parameters has been added. The same way it works in sequential. It is guaranteed for the walker to receive every model so far + the latest finished. It is then free to stop the vending, because the `stopEarly` method returning true will stop new models from being "queued".

Early stopping by time constraint is tested in `testParallelModelTimeConstraint`. Btw. cancellation by user is also tested.

h2o-core/src/main/java/hex/grid/Grid.java
Pscheidl: Great, thank you for understanding. Let's no address refactorings in this PR whatsoever - these are hot topics. @seb-h2o please create a JIRA to address it if you want `Throwable` to be gone, or I can create one, no problemo.

I only agree the Throwable is weird. But it's honoring existing contract.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Oh, the message has been fixed.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
Pscheidl: Fixed.

h2o-algos/src/test/java/hex/grid/GridTest.java
```diff
@@ -23,12 +30,170 @@
...
+        put("_max_depth", maxDepthArr);
+      }};
+
+      GBMModel.GBMParameters params = new GBMModel.GBMParameters();
+      params._train = trainingFrame._key;
+      params._response_column = "IsDepDelayed";
+      params._seed = 42;
+
+      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms,
+              GridSearch.getParallelismLevel(GridSearch.ADAPTIVE_PARALLELISM_LEVEL));
```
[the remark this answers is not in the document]
Pscheidl: Adapted to your liking.

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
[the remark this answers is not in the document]
Pscheidl: Chose to remove the call. Using `SEQUENTIAL_MODEL_BUILDING` only.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: The stopEarly method acccepts "last model" and has access to all the previous scoring infos of models built in time. This contract I believe can be honored even in parallel mode.

Pscheidl: But I can also see the potential problems with that. The APIs are vaguely defined, but I don't have the balls to refactor them right now. Most of those are truly user-facing and there'd be a battle far beyond scope of this PR. For current walkers, it will work just fine. For future walkers ... not so sure.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: It uses a lot from the grid itself. I don't want to externalize it because of that. I think it's a justified case for inner class just because of the amount of  reuse and maybe more importantly it's limited context to one place.

h2o-core/src/main/java/hex/ModelBuilder.java
Pscheidl: Used Iced abstract class as this is easier/quicker for any others to implement.

h2o-core/src/main/java/hex/grid/hyperspace/CartesianWalker.java
Pscheidl: Completely discarded the refactor.

h2o-py/tests/testdir_algos/grid/pyunit_grid_parallel.py
Pscheidl: Here @michalkurka https://0xdata.atlassian.net/browse/PUBDEV-7031

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -238,7 +245,19 @@ public void compute2() {
         Scope.exit();
       }
       tryComplete();
+      if (_modelBuilderListener != null) {
+        _modelBuilderListener.onModelSuccess(_job.get());
```
michalkurka: What will happen in the case when the MB is running nested in another job (`trainModelNested`)? Is there another way to access the model instance?

Pscheidl: Got inspired by `get` method from this very `ModelBuilder`. - `  public final M get() { assert _job._result == _result; return _job.get(); }`

I guess I can use `_result.get()` (pushed).

Pscheidl:
There is a problem with lambda serialization: http://mr-0xc1:8080/job/h2o-3-pipeline/job/PR-3990/45/artifact/r3.4-small-client-mode/h2o-3/h2o-r/tests/results/failed/testdir_algos_grid_runit_grid_parallel.R.out.txt

Pscheidl:
Commit https://github.com/h2oai/h2o-3/pull/3990/commits/b114e66442de118052428b92959908e099e7c518 should fix it - got rid of lambdas.

h2o-py/tests/testdir_algos/grid/pyunit_grid_parallel.py
michalkurka: Thank you

Pscheidl:
@seb-h2o Would you be ok with this ? Despite the lack of queues and RPCs missing.

The RPCs do not need to be there, as this logic is now only executed on leader node. @michalkurka Is working to add a way to avoid chunking, distribute smaller `Frame`s across the whole cluster and calculate models independently on many nodes. This scenario will like require some kind of RPC. However, right now there is nothing to code against.

I've created the benchmark JIRA.

Pscheidl:
Thank you.

PR #3992, opened 2019-10-18
PUBDEV-6986: Python Docs: Add examples to TargetEncoder

hannah-tillman:


h2o-bindings/bin/custom/python/gen_targetencoder.py
```diff
@@ -15,8 +15,22 @@ def transform(self, frame, data_leakage_handling="None", noise=-1, seed=-1):
...
+        >>> fold_col = "kfold_column"
+        >>> titanic[fold_col] = titanic.kfold_column(n_folds=5, seed=1234)
+        >>> titanic_te = H2OTargetEncoderEstimator(k = 0.7,
+        ...                                        f = 0.3,
+        ...                                        data_leakage_handling = "None",
+        ...                                        blending = True)
+        >>> titanic_te.train(x = predictors,
+        ...                  y = response,
+        ...                  training_frame = titanic)
+        >>> transformed = titanic_te.transform(frame = titanic,
```
seb-h2o: minor remark: `frame` is a positional/required param, as such, it is not recommended to use it as a keyword param (there's even a new feature in Py 3.8. to prevent this). This works fine, just doesn't improve readability.

minor minoris remark: Py recommendation is to avoid spaces around `=` in function parameters (to distinguish it from `=` as an assignment.
https://www.python.org/dev/peps/pep-0008/#other-recommendations

minimus remark: there's an old typo line 5: `trains()` method doesn't exist.

h2o-bindings/bin/custom/python/gen_targetencoder.py
```diff
@@ -15,8 +15,22 @@ def transform(self, frame, data_leakage_handling="None", noise=-1, seed=-1):
...
-                            >>> encodedTrain = targetEncoder.transform(frame=trainFrame, data_leakage_handling="None", seed=1234, is_train_or_valid=True)
+        >>> titanic = h2o.import_file("https://s3.amazonaws.com/h2o-public-test-data/smalldata/gbm_test/titanic.csv")
+        >>> predictors = ["home.dest", "cabin", "embarked"]
+        >>> response = "survived"
+        >>> titanic["survived"] = titanic["survived"].asfactor()
+        >>> fold_col = "kfold_column"
+        >>> titanic[fold_col] = titanic.kfold_column(n_folds=5, seed=1234)
+        >>> titanic_te = H2OTargetEncoderEstimator(k = 0.7,
+        ...                                        f = 0.3,
+        ...                                        data_leakage_handling = "None",
```
seb-h2o: in the examples, please avoid specifying parameters if you're using their default value: here `data_leakage_handling`. This applies to many of the examples below.

This param has a flaw: it provides 2 representations for the same value, `data_leakage_handling=None` (default) and `data_leakage_handling="None" which can only create confusion, so let's stick with the usual None value.

h2o-bindings/bin/custom/python/gen_targetencoder.py
```diff
@@ -15,8 +15,22 @@ def transform(self, frame, data_leakage_handling="None", noise=-1, seed=-1):
...
+        ...                                        blending = True)
+        >>> titanic_te.train(x = predictors,
+        ...                  y = response,
+        ...                  training_frame = titanic)
+        >>> transformed = titanic_te.transform(frame = titanic,
+        ...                                    data_leakage_handling="None",
```
seb-h2o: this one can be removed as well for clarity, and many others below

h2o-bindings/bin/custom/python/gen_targetencoder.py
```diff
@@ -15,8 +15,22 @@ def transform(self, frame, data_leakage_handling="None", noise=-1, seed=-1):
         :param int seed: a random seed used to generate draws from the uniform distribution for random noise. Defaults to -1.
 
         :example:
-        >>> targetEncoder = TargetEncoder(encoded_columns=te_columns, target_column=responseColumnName, blended_avg=True, inflection_point=10, smoothing=20)
-                            >>> encodedTrain = targetEncoder.transform(frame=trainFrame, data_leakage_handling="None", seed=1234, is_train_or_valid=True)
+        >>> titanic = h2o.import_file("https://s3.amazonaws.com/h2o-public-test-data/smalldata/gbm_test/titanic.csv")
+        >>> predictors = ["home.dest", "cabin", "embarked"]
+        >>> response = "survived"
+        >>> titanic["survived"] = titanic["survived"].asfactor()
```
seb-h2o: I guess users using TE should know this by now – that response needs to be converted to a categorical to enforce classification –, but let's make it even more explicit in the examples and reuse the variable just defined:
`titanic[response] = titanic[response].asfactor()`

h2o-bindings/bin/custom/python/gen_targetencoder.py
```diff
@@ -29,3 +43,134 @@ def transform(self, frame, data_leakage_handling="None", noise=-1, seed=-1):
...
+    data_leakage_handling="""
+>>> titanic = h2o.import_file("https://s3.amazonaws.com/h2o-public-test-data/smalldata/gbm_test/titanic.csv")
+>>> predictors = ["home.dest", "cabin", "embarked"]
+>>> response = "survived"
+>>> titanic["survived"] = titanic["survived"].asfactor()
+>>> fold_col = "kfold_column"
+>>> titanic[fold_col] = titanic.kfold_column(n_folds=5, seed=1234)
+>>> titanic_te = H2OTargetEncoderEstimator(k = 0.7,
+...                                        f = 0.3,
+...                                        data_leakage_handling= "k_fold",
```
seb-h2o: `data_leakage_handling="k_fold"`: this is the correct value.
Looks like there's an error in the documentation lines 10-12 then: 'KFold" -> 'k_fold' ...

h2o-bindings/bin/custom/python/gen_targetencoder.py
```diff
@@ -29,3 +43,134 @@ def transform(self, frame, data_leakage_handling="None", noise=-1, seed=-1):
...
+>>> titanic_te = H2OTargetEncoderEstimator(k = 0.7,
+...                                        f = 0.3,
+...                                        data_leakage_handling = "None",
+...                                        blending = True)
+>>> titanic_te.train(x = predictors,
+...                  y = response,
+...                  training_frame = titanic)
+>>> titanic_te
+""",
+    ignored_columns="""
```
seb-h2o: please don't document this param (anywhere, I should say): code generator unfortunately exposes it but it's aimed at internal use only.

h2o-bindings/bin/custom/python/gen_targetencoder.py
```diff
@@ -29,3 +43,134 @@ def transform(self, frame, data_leakage_handling="None", noise=-1, seed=-1):
...
+>>> titanic_te = H2OTargetEncoderEstimator(k = 0.7,
+...                                        f = 0.3,
+...                                        data_leakage_handling = "None",
+...                                        blending = True)
+>>> titanic_te.train(x = predictors,
+...                  y = response,
+...                  training_frame = titanic)
+>>> titanic_te
+""",
+    response_column="""
```
seb-h2o: same as for `ignored_columns`

h2o-bindings/bin/custom/python/gen_targetencoder.py
seb-h2o: Just double checked that both spelling work, `k_fold` and `KFold`, but for documentation, maybe we should just stick to one. I suggest the `k_fold` case, as it is what the code generator is using for auto-generated doc.

seb-h2o: It's one of those cases when trying to be clever somewhere (in parsing values on backend) can create confusion somewhere else...

hannah-tillman: Both versions work for each, but I'll switch it over to the `k_fold` case

h2o-bindings/bin/custom/python/gen_targetencoder.py
```diff
@@ -15,8 +15,22 @@ def transform(self, frame, data_leakage_handling="None", noise=-1, seed=-1):
...
         :example:
-        >>> targetEncoder = TargetEncoder(encoded_columns=te_columns, target_column=responseColumnName, blended_avg=True, inflection_point=10, smoothing=20)
-                            >>> encodedTrain = targetEncoder.transform(frame=trainFrame, data_leakage_handling="None", seed=1234, is_train_or_valid=True)
+        >>> titanic = h2o.import_file("https://s3.amazonaws.com/h2o-public-test-data/smalldata/gbm_test/titanic.csv")
+        >>> predictors = ["home.dest", "cabin", "embarked"]
+        >>> response = "survived"
+        >>> titanic["survived"] = titanic["survived"].asfactor()
+        >>> fold_col = "kfold_column"
+        >>> titanic[fold_col] = titanic.kfold_column(n_folds=5, seed=1234)
+        >>> titanic_te = H2OTargetEncoderEstimator(k = 0.7,
```
deil87: Btw why we switched from `inflection_point` and `smoothing` parameters to `k` and `f`?  Even though original paper from Daniele Micci-Barreca uses `k` and `f` I think it is more meaningful to use naming from previous API from `TargetEncoder` class. DAI uses something similar to `inflection_point` and `smoothing` as well. But yeah, it is doc-related PR and these concerns are outside of the current scope. 
Also I would suggest to use integers for K and F. At least for K we definitely should prefer integer values as it is about number of rows.  
Minor note:  imho when you see values like 0.7 and 0.3 it feels like values should be from the [0, 1] range. 
cc @ledell

seb-h2o: @deil87 your point about naming is a very fair point. I thought you were involved in the writing of this api, but apparently you weren't...
@ledell Are we comfortable with changing this now? This API is already available in `rel-yau`...
Or we could add aliases in client APIs but it adds complexity.

PR #3994, opened 2019-10-20
PUBDEV-6991: Added malloc4 for 2D int case. Removed code duplication for malloc 2D double case

deil87:


h2o-algos/src/main/java/hex/coxph/CoxPHUtils.java
```diff
@@ -4,13 +4,6 @@
...
 class CoxPHUtils {
 
-  static double[][] malloc2DArray(final int d1, final int d2) {
-    final double[][] array = new double[d1][];
-    for (int j = 0; j < d1; ++j)
-      array[j] = MemoryManager.malloc8d(d2);
-    return array;
-  }
-
   static double[][][] malloc3DArray(final int d1, final int d2, final int d3) {
```
deil87: @michalkurka maybe we can move this 3D method to MemoryManager as well? or you think it is not that common  to put it there?

deil87:
Java 8 unit tests are fine, so merging this one

PR #3995, opened 2019-10-20
PUBDEV-6992: Deprecated Frame.find(Vec) and Frame.find(Key)

deil87:
This PR is an implementation of the suggestion from discussion https://github.com/h2oai/h2o-3/pull/3975#discussion_r335547805

### DIFF
--- a/h2o-core/src/main/java/water/fvec/Frame.java
+++ b/h2o-core/src/main/java/water/fvec/Frame.java
@@ -384,7 +384,8 @@ public class Frame extends Lockable<Frame> {
 
   /**   Finds the matching column index, or -1 if missing
    *  @return the matching column index, or -1 if missing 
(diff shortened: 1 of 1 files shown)


h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -383,7 +383,10 @@ public int find( String name ) {
   }
 
   /**   Finds the matching column index, or -1 if missing
-   *  @return the matching column index, or -1 if missing */
+   *  @return the matching column index, or -1 if missing 
+   *  @deprecated as many columns in a Frame could be backed by the same Vec, we can't return single column index that corresponds to a given {@code vec}
```
seb-h2o: Thank you.
Maybe just add a suggestion that the developer should modify its code to be able to use `find(String name)` instead.

deil87: @seb-h2o yep. added suggestion.

PR #3998, opened 2019-10-21
PUBDEV-6995: Swapped default TE blending params to be consistent with clients' defaults

deil87:
https://0xdata.atlassian.net/browse/PUBDEV-6995

### DIFF
--- a/h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderTransformParametersV3.java
+++ b/h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderTransformParametersV3.java
@@ -4,6 +4,7 @@ import ai.h2o.targetencoding.TargetEncoder;
 import ai.h2o.targetencoding.TargetEncoderModel;
 import ai.h2o.targetencoding.TargetEncoderTransformParameters;
 import water.api.API;
+import water.api.EnumValuesProvider;
 import water.api.schemas3.KeyV3;
 import water.api.schemas3.SchemaV3;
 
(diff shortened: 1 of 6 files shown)


deil87:
@Pscheidl could you please check whether you have relied somewhere on previous values of parameters? Sure tests should reveal that, but just to be double cautious.

h2o-r/h2o-package/R/targetencoder.R
```diff
@@ -13,11 +13,10 @@
 #' @param training_frame Id of the training data frame.
 #' @param blending \code{Logical}. Blending enabled/disabled Defaults to FALSE.
 #' @param k Inflection point. Used for blending (if enabled). Blending is to be enabled separately using the 'blending'
-#'        parameter. Defaults to 20.
-#' @param f Smoothing. Used for blending (if enabled). Blending is to be enabled separately using the 'blending'
 #'        parameter. Defaults to 10.
-#' @param data_leakage_handling Data leakage handling strategy. Default to None. Must be one of: "None", "KFold", "LeaveOneOut". Defaults to
```
deil87: here we had duplication of "Defaults to None" so I had to remove this piece from the original API message

deil87:
@michalkurka thanks! Addressed your comment in last commit.

michalkurka:
@deil87 thank you!

regarding the change of defaults - we usually try to be backward compatible but I am not sure to what extent it applies here

I will approve the PR from a technical point of view and please ask @ledell about being backward compatible and if we need to document the change somewhere

deil87:
@ledell when you have time, please take a look at this PR. Please let me know to what extent we want to be backward compatible here. IMHO if we want to be consistent across all our TE functionality  we need to merge this, sooner of later.

h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderTransformParametersV3.java
```diff
@@ -13,7 +13,7 @@
   public KeyV3.ModelKeyV3<TargetEncoderModel> model;
   @API(help = "Seed value")
   public long seed;
-  @API(help = "Data leakage handling strategy. Default to None.", values = {"None", "KFold", "LeaveOneOut"})
+  @API(help = "Data leakage handling strategy.", values = {"None", "KFold", "LeaveOneOut"})
```
seb-h2o: @deil87 FYI: there's possibility to use `valuesProvider=LeakageStrategyHandler.class` if you want this to be updated automatically when adding a new strategy:
you just need to have `LeakageStrategyHandler` to extend `EnumValuesProvider` (cf. other examples).

deil87: @seb-h2o thanks! will definitely try this approach

PR #4003, opened 2019-10-22
PUBDEV-6982: improve xgboost updater/tree-method configuration

honzasterba:
- fix the name of max_bin native param
- properly handle hist tree method


michalkurka:
LGTM - please wait with merging for after the release

PR #4005, opened 2019-10-22
Add ability to specify algo parameters (e.g. monotonicity constraints) in H2O AutoML

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6737

Current implementation restricts the parameters that can be assigned by the user to:
- `monotone_constraints` as required by initial customer request.

Added a System property to allow overriding of `any` parameter: the plan is to use this for development/benchmarking.
Tuning algorithms, and running benchmarks against versions with different parameters is currently a hassle (recompile, rename built jar, upload it, keep reference...).
Having those parameters configurable from client will allow us to try more combinations faster (may also support hyperparameters override in the future for benchmarking as well).

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
@@ -1,11 +1,23 @@
 package ai.h2o.automl;
 
+import hex.Model;
 import hex.ScoreKeeper;
+import hex.deeplearning.DeepLearningModel;
+import hex.ensemble.StackedEnsembleModel;
+import hex.glm.GLMModel;
 import hex.grid.HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria;
+import hex.tree.drf.DRFModel;
+import hex.tree.gbm.GBMModel;
+import hex.tree.xgboost.XGBoostModel;
 import water.Iced;
 import water.Key;
 import water.api.schemas3.JobV3;
+import water.exceptions.H2OIllegalArgumentException;
 import water.fvec.Frame;
+import water.util.ArrayUtils;
+import water.util.IcedHashMap;
+import water.util.Log;
+import water.util.PojoUtils;
 
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
@@ -174,6 +186,77 @@ public class AutoMLBuildSpec extends Iced {
     public Algo[] exclude_algos;
     public Algo[] include_algos;
     public StepDefinition[] modeling_plan;
+    public AutoMLCustomParameters algo_parameters;
+  }
+
+  public static final class AutoMLCustomParameters extends Iced {
+
+    public boolean has(Algo algo) {
+      return algo_parameter_names.get(algo.name()) != null;
+    }
+
+    public boolean has(Algo algo, String param) {
+      return ArrayUtils.contains(algo_parameter_names.get(algo.name()), param);
+    }
+
+    public <V> void set(String param, V value) {
+      for (Algo algo : Algo.values()) {
+        set(algo, param, value);
+      }
+    }
+
+    public <V> void set(Algo algo, String param, V value) {
+      Model.Parameters customParams = getCustomParameters(algo);
+      try {
+        PojoUtils.setField(customParams, param, value, PojoUtils.FieldNaming.CONSISTENT);
+        PojoUtils.setField(customParams, param, value, PojoUtils.FieldNaming.DEST_HAS_UNDERSCORES);
+        addParameterName(algo, param);
+      } catch (IllegalArgumentException iae) {
+        Log.debug("Could not set custom param "+param+" for algo "+algo+": "+iae.getMessage());
+      }
+    }
+
+    public String[] getCustomParameterNames(Algo algo) {
+      return algo_parameter_names.get(algo.name());
+    }
+
+    public Model.Parameters getCustomParameters(Algo algo) {
+      if (!algo_parameters.containsKey(algo.name())) algo_parameters.put(algo.name(), defaultParameters(algo));
+      return algo_parameters.get(algo.name());
+    }
+
+    public void setCustomParameters(Algo algo, Model.Parameters destParams) {
+      if (has(algo)) {
+          PojoUtils.copyProperties(destParams, getCustomParameters(algo), PojoUtils.FieldNaming.CONSISTENT, null, getCustomParameterNames(algo));
+      }
+    }
+
+    private Model.Parameters defaultParameters(Algo algo) {
+      switch (algo) {
+        case DeepLearning: return new DeepLearningModel.DeepLearningParameters();
+        case DRF: return new DRFModel.DRFParameters();
+        case GBM: return new GBMModel.GBMParameters();
+        case GLM: return new GLMModel.GLMParameters();
+        case StackedEnsemble: return new StackedEnsembleModel.StackedEnsembleParameters();
+        case XGBoost: return new XGBoostModel.XGBoostParameters();
+        default: throw new H2OIllegalArgumentException("Custom parameters are not supported for "+algo.name()+".");
+      }
+    }
+
+    private void addParameterName(Algo algo, String param) {
+      if (!algo_parameter_names.containsKey(algo.name())) {
+        algo_parameter_names.put(algo.name(), new String[] {param});
+      } else {
+        String[] names = algo_parameter_names.get(algo.name());
+        if (!ArrayUtils.contains(names, param)) {
+          algo_parameter_names.put(algo.name(), ArrayUtils.append(names, param));
+        }
+      }
+    }
+
+    public IcedHashMap<String, String[]> algo_parameter_names = new IcedHashMap<>(); // stores the parameters names overridden, by algo name
+    public IcedHashMap<String, Model.Parameters> algo_parameters = new IcedHashMap<>(); //stores the parameters values, by algo name
+
   }
 
   public AutoMLBuildControl build_control;
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
@@ -2,13 +2,13 @@ package ai.h2o.automl;
 
 import hex.Model;
 import hex.ScoreKeeper;
-import hex.deeplearning.DeepLearningModel;
-import hex.ensemble.StackedEnsembleModel;
-import hex.glm.GLMModel;
+import hex.deeplearning.DeepLearningModel.DeepLearningParameters;
+import hex.ensemble.StackedEnsembleModel.StackedEnsembleParameters;
+import hex.glm.GLMModel.GLMParameters;
 import hex.grid.HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria;
-import hex.tree.drf.DRFModel;
-import hex.tree.gbm.GBMModel;
-import hex.tree.xgboost.XGBoostModel;
+import hex.tree.drf.DRFModel.DRFParameters;
+import hex.tree.gbm.GBMModel.GBMParameters;
+import hex.tree.xgboost.XGBoostModel.XGBoostParameters;
 import water.H2O;
 import water.Iced;
 import water.Key;
@@ -277,12 +277,12 @@ public class AutoMLBuildSpec extends Iced {
 
     private Model.Parameters defaultParameters(Algo algo) {
       switch (algo) {
-        case DeepLearning: return new DeepLearningModel.DeepLearningParameters();
-        case DRF: return new DRFModel.DRFParameters();
-        case GBM: return new GBMModel.GBMParameters();
-        case GLM: return new GLMModel.GLMParameters();
-        case StackedEnsemble: return new StackedEnsembleModel.StackedEnsembleParameters();
-        case XGBoost: return new XGBoostModel.XGBoostParameters();
+        case DeepLearning: return new DeepLearningParameters();
+        case DRF: return new DRFParameters();
+        case GBM: return new GBMParameters();
+        case GLM: return new GLMParameters();
+        case StackedEnsemble: return new StackedEnsembleParameters();
+        case XGBoost: return new XGBoostParameters();
         default: throw new H2OIllegalArgumentException("Custom parameters are not supported for "+algo.name()+".");
       }
     }
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
@@ -12,7 +12,6 @@ import hex.tree.xgboost.XGBoostModel;
 import water.H2O;
 import water.Iced;
 import water.Key;
-import water.api.schemas3.JobV3;
 import water.exceptions.H2OIllegalArgumentException;
 import water.exceptions.H2OIllegalValueException;
 import water.fvec.Frame;
@@ -348,9 +347,6 @@ public class AutoMLBuildSpec extends Iced {
   public final AutoMLBuildControl build_control = new AutoMLBuildControl();
   public final AutoMLInput input_spec = new AutoMLInput();
   public final AutoMLBuildModels build_models = new AutoMLBuildModels();
-
-  // output
-  public JobV3 job;
 
   public String project() {
     if (build_control.project_name == null) {
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
@@ -13,11 +13,15 @@ import water.Iced;
 import water.Key;
 import water.api.schemas3.JobV3;
 import water.exceptions.H2OIllegalArgumentException;
+import water.exceptions.H2OIllegalValueException;
 import water.fvec.Frame;
 import water.util.ArrayUtils;
 import water.util.IcedHashMap;
 import water.util.Log;
 import water.util.PojoUtils;
+import water.util.PojoUtils.FieldNaming;
+
+import java.util.stream.Stream;
 
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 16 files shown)


h2o-automl/src/main/java/ai/h2o/automl/modeling/GBMStepsProvider.java
```diff
@@ -111,7 +113,6 @@ GBMParameters prepareModelParameters() {
                         GBMParameters gbmParameters = prepareModelParameters();
 
                         Map<String, Object[]> searchParams = new HashMap<>();
-                        searchParams.put("_ntrees", new Integer[]{10000});
```
seb-h2o: there was no point in keeping this as an hyperparam if there is only one value

h2o-core/src/main/java/water/util/JSONValue.java
```diff
@@ -0,0 +1,66 @@
+package water.util;
+
+import water.Iced;
+import water.api.Schema;
+
+import java.lang.reflect.Array;
+import java.util.Objects;
+
+public class JSONValue<V> extends Iced {
```
seb-h2o: @michalkurka any concern if I add this new class to core?
AutoML is posting whole payload in POST body as JSON.
This class is intended to provide a safe encapsulation of some parts of the body that can be parsed dynamically later.

See the test and the use in `POJOUtils` below.

michalkurka: I think this is a very clean solution, I don't see any issue having it in core.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -63,20 +57,28 @@ public AutoMLBuildControl() {
     public boolean keep_cross_validation_models = false;
     public boolean keep_cross_validation_fold_assignment = false;
     public String export_checkpoints_dir = null;
+
+    public AutoMLBuildControl() {
+      stopping_criteria = new AutoMLStoppingCriteria();
+
+      // reasonable defaults:
+      stopping_criteria.set_max_models(0);
```
ledell: Is -1 or 0 more common as a replacement for None/NULL for an integer-valued parameter on the backend?  I have seen both... (which I guess means we can choose whichever one is "better"?).

h2o-r/h2o-package/R/automl.R
```diff
@@ -46,7 +46,11 @@
 #         Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
 #' @param include_algos Vector of character strings naming the algorithms to restrict to during the model-building phase. This can't be used in combination with exclude_algos param.
 #         Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
-#' @param modeling_plan List. The list of modeling steps to be used by the AutoML engine (they may not all get executed, depending on other constraints). Optional.
+#' @param modeling_plan List. The list of modeling steps to be used by the AutoML engine (they may not all get executed, depending on other constraints). Optional (Expert usage only).
+#' @param algo_parameters List.
+#' @param algo_parameters List. A list of param_name=param_value to be passed to internal models. Defaults to none (Expert usage only).
```
ledell: `algo_parameters` is here twice.  I think this is a typo.

h2o-r/h2o-package/R/automl.R
```diff
@@ -46,7 +46,11 @@
 #         Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
 #' @param include_algos Vector of character strings naming the algorithms to restrict to during the model-building phase. This can't be used in combination with exclude_algos param.
 #         Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
-#' @param modeling_plan List. The list of modeling steps to be used by the AutoML engine (they may not all get executed, depending on other constraints). Optional.
+#' @param modeling_plan List. The list of modeling steps to be used by the AutoML engine (they may not all get executed, depending on other constraints). Optional (Expert usage only).
+#' @param algo_parameters List.
```
```suggestion
```

ledell:
@seb-h2o Can you update the description to remove `ntrees` and to add `algo_parameters` instead? 

My concern with `algo_parameters` is that if someone only wants to limit `ntrees` in GBMs, for example, there is no way to limit that to just GBMs.  Since DRF and XGBoost also has an `ntrees` parameter, it will also limit it there too (from my understanding of your current implementation).  I guess the only way to give further precision is to have it be a nested dict/list:
```
algo_parameters = {"GBM": {"ntrees": 1000, "learn_rate": 0.01}, "DRF": {"ntrees": 200}}
```

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -174,14 +176,177 @@ public RandomDiscreteValueSearchCriteria getSearchCriteria() {
...
+
+    public Model.Parameters getCustomParameters(Algo algo) {
+      if (!algo_parameters.containsKey(algo.name())) algo_parameters.put(algo.name(), defaultParameters(algo));
+      return algo_parameters.get(algo.name());
+    }
+
+    public void applyCustomParameters(Algo algo, Model.Parameters destParams) {
+      if (hasCustomParams(algo)) {
+        String[] paramNames = getCustomParameterNames(algo);
+        String[] onlyParamNames = ArrayUtils.append(Stream.of(paramNames).map(p -> "_"+p).toArray(String[]::new), paramNames);
```
deil87: @seb-h2o do we really need to append to the original names? I think that just with-underscore-prefix ones should be fine

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
seb-h2o: @ledell I didn't change that code, just moved it down: usually in Java, attributes are declared before constructor, I was just applying this "norm" here.

Regarding the -1/0 debate... there's no rule written in stone. -1 is commonly used for "unset", and when there's no ambiguity like here, I tend to use 0 for "unlimited". Maybe it deserves a tiny comment next to it.

h2o-r/h2o-package/R/automl.R
seb-h2o: thx

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
```diff
@@ -163,6 +164,37 @@ public AlgoProvider() {
...
+    private static final String[] SCOPES = ArrayUtils.append(new String[]{ANY_ALGO}, new AlgoProvider().values());
+
+    @Override
+    public String[] values() {
+      return SCOPES;
+    }
+  }
+
+  public static final class AutoMLCustomParameterV99<V> extends Schema<Iced, AutoMLCustomParameterV99<V>> {
+    @API(help="", valuesProvider=ScopeProvider.class, direction=API.Direction.INPUT)
```
deil87: minor: no help for at all for all these fields? maybe they are too internal but this is conflicting  with the fact that APIs are for someone out there

h2o-core/src/main/java/water/api/SchemaMetadata.java
```diff
@@ -423,6 +423,9 @@ private static String getEnumSchemaName(Class<?> type) {
     StringBuffer sb = new StringBuffer(type.getCanonicalName());
     sb.delete(0, sb.indexOf(".")+1);
     sb.setCharAt(0, Character.toUpperCase(sb.charAt(0)));
-    return sb.toString().replace(".", "").replace("$", "");
+    return sb.toString()
```
deil87: @seb-h2o just curious if we had tests for this functionality

h2o-r/tests/testdir_algos/automl/runit_automl_model_selection.R
```diff
@@ -211,7 +244,8 @@ automl.model_selection.suite <- function() {
     test_modeling_plan_full_syntax,
     test_modeling_plan_minimal_syntax,
     test_modeling_steps,
-    test_exclude_algos_is_applied_on_top_of_modeling_plan
+    test_exclude_algos_is_applied_on_top_of_modeling_plan,
+    test_monotone_constraints_can_be_passed_as_algo_parameter,
```
deil87: @seb-h2o for Py you have two more tests 
``` 
    test_algo_parameter_can_be_applied_only_to_a_specific_algo,
    test_cannot_set_unauthorized_algo_parameter
```
just in case it was forgotten.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -174,14 +176,177 @@ public RandomDiscreteValueSearchCriteria getSearchCriteria() {
...
+    // let's limit the list of allowed custom parameters by default for now: we can always decide to open this later.
+    private static final String[] ALLOWED_PARAMETERS = {
+            "monotone_constraints",
+            "ntrees",
+    };
+    private static final String ROOT_PARAM = "algo_parameters";
+
+    public static final class AutoMLCustomParameter<V> extends Iced {
+      public AutoMLCustomParameter(String name, V value) {
+        this.name = name;
```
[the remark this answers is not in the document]
seb-h2o: @deil87 I do, but I'm getting all confused in this particular file as it is API class and most fields are public, so they don't have underscore.
I should use `_` prefix for private fields though, correct.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -174,14 +176,177 @@ public RandomDiscreteValueSearchCriteria getSearchCriteria() {
...
+
+    /**
+     * Must always be called right after adding all custom parameters with {@link #add(String, Object)} methods.
+     * This is necessary as the custom parameters must be applied in a certain order,
+     * and we can't assume that the consumer of this API will add them in the right order.
+     * @return the current instance with custom parameters properly assigned.
+     */
+    public AutoMLCustomParameters end() {
+      if (!algo_parameters.isEmpty()) {
+        throw new IllegalStateException("Cannot call `build` if AutoML algo_parameters are already assigned.");
```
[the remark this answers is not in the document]
seb-h2o: It should, guess how the method used to be named?

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
seb-h2o: Will review this, but I think it's necessary for the PojoUtils magic.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
seb-h2o: my bad, forgot to add it

h2o-core/src/main/java/water/api/SchemaMetadata.java
seb-h2o: it's used at build time: build calls metadata API to generate code

h2o-r/tests/testdir_algos/automl/runit_automl_model_selection.R
seb-h2o: it was not really forgotten... always wondering if Py and R should have the same tests.
Usually I cover REST API + Py client in Py tests, whereas I just cover R client in R tests...

seb-h2o: `test_algo_parameter_can_be_applied_only_to_a_specific_algo` could be added to R though

seb-h2o:
> Can you update the description to remove `ntrees` and to add `algo_parameters` instead?

> Can you update the description to remove `ntrees` and to add `algo_parameters` instead?

@ledell you mean in the JIRA ticket?
regarding the implementation, it is actually possible to set params for a specific algo using the following syntax (works both in R and Py):
```
algo_parameters = dict(
    ntrees=200,    # applies to all algos supporting ntrees
    GBM__ntrees=1000 .   # applies only to GBMs
)
```
Anyway, as soon as we're not sure about exposing `ntrees`, waiting for discussion with customer, I'll keep this feature hidden.
Remains the question of `monotone_constraints`: do we want to have the param at top level as Michal suggests?

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -174,14 +176,177 @@ public RandomDiscreteValueSearchCriteria getSearchCriteria() {
...
+      private Algo algo;
+      private String name;
+      private V value;
+    }
+
+    private final IcedHashMap<String, String[]> algo_parameter_names = new IcedHashMap<>(); // stores the parameters names overridden, by algo name
+    private final IcedHashMap<String, Model.Parameters> algo_parameters = new IcedHashMap<>(); //stores the parameters values, by algo name
+
+    // used for building phase only
+    private final transient List<AutoMLCustomParameter> any_algo_params = new ArrayList<>();
```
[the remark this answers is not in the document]
seb-h2o: I know, if there were more scopes, it would make sense.
I considered it was not worth having a Map when the partition is binary: any vs. specific.
Mainly because there's always a bit more logic when you need to manipulate map values.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -174,14 +176,177 @@ public RandomDiscreteValueSearchCriteria getSearchCriteria() {
...
+      // apply "all" scope first, then algo-specific ones.
+      for (AutoMLCustomParameter param : any_algo_params) {
+        if (!addParameter(param.name, param.value))
+          throw new H2OIllegalValueException(param.name, ROOT_PARAM, param.value);
+      }
+      for (AutoMLCustomParameter param : specific_algo_params) {
+        if (!addParameter(param.algo, param.name, param.value))
+          throw new H2OIllegalValueException(param.name, ROOT_PARAM, param.value);
+      }
+      return this;
```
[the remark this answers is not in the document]
seb-h2o: Yeah, I hesitated into turning this into a proper "builder" (that's also why I renamed the `build` method).
Mainly trying to keep it simple: not sure builders are justified most of the time.
When I see
```
 AutoMLCustomParameters algo_parameters;
```
my reflex is:
```
algo_parameters = new AutoMLCustomParameters();
```
not: "damn, constructor is private, any factory method? `AutoMLCustomParameters.newInstance(...)`? damn, what then? Oh found it `new AutoMLCustomParametersBuilder()....build()`"

So trying to combine the best of both:
```
algo_parameters = new AutoMLCustomParameters().add(...)
                                                                                         .add(...)
                                                                                         .end();
```
It's not without flaws though, conscious about this. Will reconsider and try to find sth more satisfying.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -174,14 +176,177 @@ public RandomDiscreteValueSearchCriteria getSearchCriteria() {
...
+    }
+
+    public String[] getCustomParameterNames(Algo algo) {
+      return algo_parameter_names.get(algo.name());
+    }
+
+    public Model.Parameters getCustomParameters(Algo algo) {
```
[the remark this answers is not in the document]
seb-h2o: Not sure this should be public anymore actually, which would make the naming less important.
`getCustomDefaults` has something oxymoronic in it that makes me like it somehow :)

ledell:
> @ledell you mean in the JIRA ticket?
@seb-h2o No I mean in this PR description (it still says `ntrees`).

Thanks for clarifying about how to apply the parameter to a particular algorithm.  I remember seeing this `algoname__paramname` format before (was that here or do we use this format in another part of H2O?).

Can you list the strings that you accept for each algorithm?  I want to make sure we are using consistent naming of the algorithms... I filed this [JIRA](https://0xdata.atlassian.net/browse/PUBDEV-6990) recently because I noticed that H2O has several sets of algorithm names.  It would be great to unify this across H2O moving forward.

Regarding `algo_parameters` in general -- I know that we had planned to allow users to fully customize the AutoML "plan" at some point (we had this on a roadmap, but it's not a JIRA yet).  Would the `algo_parameters` parameter become redundant or conflicting at that point, or do you think that it will still provide value?  I wonder how often people will want to fix a particular hyperparameter value across all the models, since the goal in AutoML is usually to try many values for a single hyperparameter...instead of just one.  Just thinking out-loud.

seb-h2o:
@ledell  
> Thanks for clarifying about how to apply the parameter to a particular algorithm. I remember seeing this `algoname__paramname` format before (was that here or do we use this format in another part of H2O?).
It is the "standard" flat format used by scikit learn to specify nested parameters: I must say that I like it. When we have the possibility to avoid an additional level of depth without possibility of confusion, it brings only clarity I think.

> Can you list the strings that you accept for each algorithm? I want to make sure we are using consistent naming of the algorithms...
AutoML is using the algo names listed in https://github.com/h2oai/h2o-3/blob/master/h2o-automl/src/main/java/ai/h2o/automl/Algo.java everywhere.
I also noticed that the parsing logic we use to extract params from the JSON body is case-sensitive with enums, meaning we usually can't use `gbm` but we have to use `GBM`.
I think we should be more lenient here and accept both: it's a very easy fix.
 
> Regarding `algo_parameters` in general - I know that we had planned to allow users to fully customize the AutoML "plan" at some point

For me `algo_parameters` is useful when user wants to set 1 or 2 specific constraints on algos, not to define complete models, at least not at end API level (tbh, when exposing this feature, I've been wondering if we could not create a model designer on top of it, only client side: it would be easy to have a "custom steps" in Java that got only fed by some of those params, internally all "custom___xxx" params for example...), also thinking out loud.

But overall, I think the idea of full customization should be thought like DAI recipes, with a mini-language describing additional models, grids, feature engineering...
This needs to be designed top down.

This being said, after disabling `ntrees`, there will be only `monotone_constraints` allowed by default (as said in the description, I keep the possibility to fully enable the feature for dev/benchmark through a `java / h2o.init` parameter).
So the question remains if we want to keep `monotone_constraints` here only, or if expose it as a 1st level param in the API.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
seb-h2o: You're right, I misread your question: I added the original names only for "safety", just in case there were params in `Model.Parameters` subclasses not starting with `_`.
Unnecessary for now though, will remove.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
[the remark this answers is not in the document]
seb-h2o: done for the private fields

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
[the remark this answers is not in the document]
seb-h2o: removed, unnecessary now that there's a proper builder

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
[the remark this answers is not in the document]
seb-h2o: ended up with `AutoMLCustomParameters.create().add(...).build()`
`create` returns an internal Builder instance`

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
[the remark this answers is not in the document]
seb-h2o: renamed, and lowered visibility (used in tests)

h2o-r/tests/testdir_algos/automl/runit_automl_model_selection.R
seb-h2o: added R test above (and discovered a bug from it)

seb-h2o:
@ledell As all concerns on this PR have been resolved, I'm merging it.
See additional PR https://github.com/h2oai/h2o-3/pull/4020 for the decision to expose `monotone_constraints` as a top level param in `AutoML`.

PR #4008, opened 2019-10-23
PUBDEV-6975: Attempt to reuse existing tools for grid search over TE hyper parameters

deil87:
This PR is related to https://github.com/h2oai/h2o-3/pull/3183 task and essentially an attempt to reuse existing grid walkers for TargetEncoder.

In original PR https://github.com/h2oai/h2o-3/pull/3183 I had independently implemented RGS as before we did not treat TE  as a model( and HyperSpaceWalker is highly coupled with the model concept). But now we do threat TE as a regular model hence the effort of this PR.

As stated in jira https://0xdata.atlassian.net/browse/PUBDEV-7006 we are missing `noise` parameter.  If we will follow this jira proposal https://0xdata.atlassian.net/browse/PUBDEV-7007 then we will remove `data_leakage_handling` and `noise` parameters from the `TargetEncoderParameters` and therefore we won't be able to grid-search over them. 

The problem is that parameters we want to grid search over are not necessarily a Model parameters. Some parameters might be needed during application of the Model. 

Pipeline is as follows:
1) Train TE model
2) apply TE model with different params to different frames
3) train predictive Model (GBM, DRF etc) on training frame and predict on test frame
4) evaluate quality of predictions

So we would ideally want to operate with Pipeline parametes ( or collection of the Model parameters) when we do grid search with `HyperSpaceWalker` in order to be able to correctly reuse existing tools for RGS.

Update:
There is also a difficulty with BlendingParams field in `TargetEncoderParameters` as it is a class and not a primitive. Need to investigate how difficult it will be to adapt the code of random walker.
Update Update:
As a result this PR do the following things:
1) introduce `_noice_level` parameter to TargetEncoderParameters class. (Flow will reflect this)
2) exposed `seed` to a Flow UI so that `noise_level` could be reproducible
3) got rid of field of `BlendingParameters`  type and replaced it with just `k` and `f` blending parameters.
4) added test that shows that we can use `HyperSpaceWalker` to iterate over `TargetEncoderParameters`

### DIFF
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
@@ -35,7 +35,8 @@ public class TargetEncoderModel extends Model<TargetEncoderModel, TargetEncoderM
     public double _k = TargetEncoder.DEFAULT_BLENDING_PARAMS.getK();
     public double _f = TargetEncoder.DEFAULT_BLENDING_PARAMS.getF();
     public TargetEncoder.DataLeakageHandlingStrategy _data_leakage_handling = TargetEncoder.DataLeakageHandlingStrategy.None;
-    public double _noise_level = 0;
+    public double _noise_level = 0.01;
+    
     @Override
     public String algoName() {
       return ALGO_NAME;
@@ -160,7 +161,7 @@ public class TargetEncoderModel extends Model<TargetEncoderModel, TargetEncoderM
             _parms._data_leakage_handling != null ? _parms._data_leakage_handling : TargetEncoder.DataLeakageHandlingStrategy.None;
     
     return _targetEncoder.applyTargetEncoding(fr, _parms._response_column, this._output._target_encoding_map, leakageHandlingStrategy,
-            _parms._fold_column, _parms._blending, _parms._seed,false, Key.<Frame>make(destination_key), blendingParams);
+            _parms._fold_column, _parms._blending, _parms._noise_level, _parms._seed, Key.<Frame>make(destination_key), blendingParams );
   }
   
 
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
@@ -150,7 +150,7 @@ public class TargetEncoderModel extends Model<TargetEncoderModel, TargetEncoderM
   
   @Override
   protected double[] score0(double data[], double preds[]){
-    throw new UnsupportedOperationException("TargetEncoderModel doesn't support scoring. Use `transform()` instead.");
+    throw new UnsupportedOperationException("TargetEncoderModel doesn't support scoring on raw data. Use transform() or score() instead.");
   }
 
   @Override
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
@@ -34,7 +34,7 @@ public class TargetEncoderModel extends Model<TargetEncoderModel, TargetEncoderM
     public boolean _blending = false;
     public BlendingParams _blending_parameters = TargetEncoder.DEFAULT_BLENDING_PARAMS;
     public TargetEncoder.DataLeakageHandlingStrategy _data_leakage_handling = TargetEncoder.DataLeakageHandlingStrategy.None;
-    
+    public double _noise_level = 0;
     @Override
     public String algoName() {
       return ALGO_NAME;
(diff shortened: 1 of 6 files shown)


michalkurka:
I see no issues with this approach - please proceed with your changes.

And it is great to see that making a TE formally a model pays off.

h2o-extensions/target-encoder/src/test/java/ai/h2o/targetencoding/strategy/TargetEncoderRGSTest.java
```diff
@@ -0,0 +1,57 @@
...
+      HashMap<String, Object[]> hpGrid = new HashMap<>();
+      hpGrid.put("blending", new Boolean[]{true, false});
+      hpGrid.put("noise_level", new Double[]{0.0, 0.01,  0.1});
+      // TODO figure out how to work with hierarchical parameters BlendingParams(inflection_point, smoothing) or BlendingParams(k, f)
+      hpGrid.put("k", new Double[]{1.0, 2.0, 3.0});
+      hpGrid.put("f", new Double[]{5.0, 10.0, 20.0});
+      
+      TargetEncoderModel.TargetEncoderParameters parameters = new TargetEncoderModel.TargetEncoderParameters();
+
+      GridSearchHandler.DefaultModelParametersBuilderFactory<TargetEncoderModel.TargetEncoderParameters, TargetEncoderV3.TargetEncoderParametersV3> modelParametersBuilderFactory = new GridSearchHandler.DefaultModelParametersBuilderFactory<>();
```
deil87:
```
// Note: Naming for DefaultModelParametersBuilderFactory and SimpleParametersBuilderFactory is confusing. 
// There are no hints that one actually uses Schema and another uses ModelParameters class itself.
GridSearch.SimpleParametersBuilderFactory simpleParametersBuilderFactory = new GridSearch.SimpleParametersBuilderFactory();

with `SimpleParametersBuilderFactory` it is probably less painful to support ability to specify hierarchical hyper parameters in the grid ( e.g. `_blending_parameters._k`) then with `DefaultModelParametersBuilderFactory` as we won't need to have a match  between fields names in `TargetEncoderParametersV3 ` and in `TargetEncoderParameters` classes. Need to investigate a little bit more to decide which approach is more preferable .

File: h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java [hunk 2]
```diff
@@ -21,6 +21,9 @@
 
     @API(help = "Data leakage handling strategy. Default to None.", values = {"None", "KFold", "LeaveOneOut"})
     public TargetEncoder.DataLeakageHandlingStrategy data_leakage_handling;
+
+    @API(help = "Noise level", required = false, direction = API.Direction.INPUT, gridable = true)
+    public double noise_level;
```
seb-h2o: Is it a percentage or sth? The doc string may need more explanations.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
```diff
@@ -119,7 +123,7 @@ private TwoDimTable constructSummary(){
    */
   public Frame transform(Frame data, byte strategy, double noiseLevel, final boolean useBlending, BlendingParams blendingParams,
```
seb-h2o: off-topic: no plan have an alternative signature using the corresponding enum instead of `byte` for the `strategy` param?
Looks like this signature is for backwards compatibility, but it's not very user-friendly.

h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java
seb-h2o: OK, checked impl. details: no dimension.

deil87:
@michalkurka 
> And it is great to see that making a TE formally a model pays off.

One can always find benefits of unifying concepts. But when things are not exactly the same one always sacrifice something. So if you look at `TargetEncoderModel.java` you will find `score` method. But there is no easy way to score with different parameters( leakageHandlingStrategy, blending parameters) - and the real user's scenario requires those abilities. The only hack is to train two models with different parameters ( which means compute encoding map twice). This is why I believe we have an issue in Flow right now.

deil87:
@Pscheidl do you think that `score0` method of the `TargetEncoderModel.java` throws misleading message? We kind of do support scoring but with different API/signature. I suggest something like this `TargetEncoderModel doesn't support scoring on raw data. Use `transform()` or score() instead.`. wdyt?

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
deil87: @seb-h2o I agree that this might be improved but this is out of scope of this PR. Maybe you can create jira which could become a simple `good first task`.

Pscheidl:
Makes absolute sense to me @deil87. I agree.

deil87:
@michalkurka @seb-h2o this I think is ready to be reviewed. This should be a quick one I guess.

deil87:
@Pscheidl could you please review this one as you should be familiar with RGS logic due to your recent activities and it will be easy for you to review. See last statements in the description for an overview of this small change.

h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java
seb-h2o: Still, this `help` is used as field documentation in Flow, would be nice to provide a recommended range for example.

PR #4009, opened 2019-10-23
Refactoring: get rid of obsolete references to UDP

michalkurka:
People are sometimes confused by error messages that mention "UDP" packets.

UDP was originally used to exchange small messages between H2O nodes, I removed the references from places that only use TCP.

I left all UDP classes - the meaning of UDP over the time changed to "small messages", we can kill the rest of the UDP references eventually as well but IMHO this change should be good enough to clear the confusion.

### DIFF
--- a/h2o-core/src/main/java/water/H2ONode.java
+++ b/h2o-core/src/main/java/water/H2ONode.java
@@ -74,8 +74,8 @@ public final class H2ONode extends Iced<H2ONode> implements Comparable {
     _sendThread = null;
   }
 
-  private UDP_TCP_SendThread startSendThread() {
-    UDP_TCP_SendThread newSendThread = new UDP_TCP_SendThread(); // Launch the UDP send thread  
+  private SmallMessagesSendThread startSendThread() {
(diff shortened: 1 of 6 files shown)


h2o-core/src/main/java/water/H2ONode.java
```diff
@@ -105,16 +105,24 @@ private UDP_TCP_SendThread startSendThread() {
...
       }
     }
-    public int htm_port() { return getPort()-H2O.ARGS.port_offset; }
-    public int udp_port() { return getPort()  ; }
-    @Override public String toString() { return getAddress()+":"+htm_port(); }
+    /**
+     * Retrieves the public communication port on which the REST API is exposed
+     * @return port number
+     */
+    int api_port() { return getPort()-H2O.ARGS.port_offset; }
```
honzasterba: since we are renaming this I would switch it to camelCase as is javaStandard

michalkurka: thanks, that makes sense

PR #4010, opened 2019-10-23
Do not schedule H2O jobs to micro nodes - leave them to SW

michalkurka:
FIY: @mn-mikke


michalkurka:
> I wonder if we should get rid of the !mr-0xc8, it makes no sense to me, why is it still there?

@honzasterba not sure - the jenkins worker is offline since August - imho it makes sense to remove it

michalkurka:
this should also help with timeouts - the `micro` nodes are not really capable of completing Java unit tests in 3h (only 4CPUs)

mn-mikke:
Thanks a lot guys!

PR #4017, opened 2019-10-24
Fix internal function .str.list

michalkurka:
.str.list(c("a", "B")) produced a warning

thanks to @seb-h2o for noticing


h2o-r/h2o-package/R/frame.R
```diff
@@ -1191,7 +1191,7 @@ h2o.pivot <- function(x, index, column, value){
 #' @param skipna if enabled, do not include NAs in the result (default: FALSE)
 #' @return an unpivoted H2OFrame
 #' @export
-h2o.melt <- function(x, id_vars, value_vars=NA_character_, var_name="variable", value_name="value", skipna=FALSE) {
+h2o.melt <- function(x, id_vars, value_vars=NULL, var_name="variable", value_name="value", skipna=FALSE) {
```
seb-h2o: is it related?

PR #4018, opened 2019-10-24
PUBDEV-6987: Fix bug for merge involving empty frames

wendycwong:
This PR resolve the issue in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6987?filter=-1

A user reported that merge will get stuck if one of the frames is empty.

I did the following:
1. Check if we have an empty frame before merging.  If one frame is empty, return the other frame right away.  If both frames are empty, throw an error.
2. Added pyunit test to make sure my fix works.

### DIFF
--- a/h2o-core/src/main/java/water/rapids/Merge.java
+++ b/h2o-core/src/main/java/water/rapids/Merge.java
@@ -8,7 +8,8 @@ import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-
+import static java.math.BigInteger.ZERO;
+import static java.math.BigInteger.ONE;
 import static water.rapids.SingleThreadRadixOrder.getSortedOXHeaderKey;
 
 public class Merge {
@@ -52,7 +53,6 @@ public class Merge {
   public static Frame merge(final Frame leftFrame, final Frame riteFrame, final int leftCols[], final int riteCols[],
                             boolean allLeft, int[][] id_maps) {
     int[] ascendingL, ascendingR;
-
     if (leftCols != null && leftCols.length>0) {
       ascendingL = new int[leftCols.length];
       Arrays.fill(ascendingL, 1);
@@ -109,6 +109,7 @@ public class Merge {
 
     // Running 3 consecutive times on an idle cluster showed that running left
     // and right in parallel was a little slower (97s) than one by one (89s).
+    // empty frame will come back with base = Long.MIN_VALUE (-9223372036854775808).  
     // TODO: retest in future
     RadixOrder leftIndex = createIndex(true ,leftFrame,leftCols,id_maps, ascendingL);
     RadixOrder riteIndex = createIndex(false,rightFrame,riteCols,id_maps, ascendingR);
@@ -125,7 +126,16 @@ public class Merge {
     final BigInteger riteBase = hasRite ? riteIndex._base [0] : leftBase;
 
     // initialize for double columns, may not be used....
-    long leftMSBfrom = riteBase.subtract(leftBase).shiftRight(leftShift).longValue();
+    boolean leftFrameEmpty = (leftFrame.numRows()==0);
+    boolean riteFrameEmpty = (riteFrame.numRows()==0);
+    long leftMSBfrom = riteBase.subtract(leftBase).shiftRight(leftShift).longValue();  // value when all frames nonempty
+    if (riteFrameEmpty) { // if riteFrame is empty, need to modify it because riteBase is Long.MIN_VALUE
+      if (leftFrameEmpty) { // if both left and rite frames are empty.
+        leftMSBfrom = 0;
+      } else {  // only riteFrame is empty
+        leftMSBfrom = ZERO.subtract(leftBase).shiftRight(leftShift).longValue();
+      }
+    }
     boolean riteBaseExceedsleftBase=riteBase.compareTo(leftBase)>0;
     // deal with the left range below the right minimum, if any
     if (riteBaseExceedsleftBase) {  // right branch has higher minimum column value
@@ -153,9 +163,14 @@ public class Merge {
     }
 
     BigInteger rightS = BigInteger.valueOf(256L<<riteShift);
-    long leftMSBto = riteBase.add(rightS).subtract(BigInteger.ONE).subtract(leftBase).shiftRight(leftShift).longValue();
-    // -1 because the 256L<<riteShift is one after the max extent.  
-    // No need -for +1 for NA here because, as for leftMSBfrom above, the NA spot is on -both sides
+    long leftMSBto = riteBase.add(rightS).subtract(ONE).subtract(leftBase).shiftRight(leftShift).longValue();
+    if (riteFrameEmpty) { // rite frame is empty
+      if (leftFrameEmpty) { // both left and rite frames are empty
+        leftMSBto = ZERO.add(rightS).subtract(ONE).shiftRight(leftShift).longValue();
+      } else {  // only riteFrame is empty
+        leftMSBto = ZERO.add(rightS).subtract(ONE).subtract(leftBase).shiftRight(leftShift).longValue();
+      }
+    }
 
     // deal with the left range above the right maximum, if any.  For doubles, -1 from shift to avoid negative outcome
     boolean leftRangeAboveRightMax = leftIndex._isCategorical[0]?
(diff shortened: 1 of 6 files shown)


h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -52,6 +52,15 @@ public static Frame sort(final Frame fr, int[] cols, int[] ascending) {
   public static Frame merge(final Frame leftFrame, final Frame riteFrame, final int leftCols[], final int riteCols[],
                             boolean allLeft, int[][] id_maps) {
     int[] ascendingL, ascendingR;
+    
+    if (leftFrame.numRows()==0 && riteFrame.numRows() > 0) { // if one frame contains zero rows, returns the other frame which may contain zero rows
+      return riteFrame;
```
michalkurka: imho the output should be the empty frame, or shouldn't it?

h2o-py/tests/testdir_munging/pyunit_6987_merge_one_empty.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import print_function
+import sys
+sys.path.insert(1,"../../")
+import h2o
+from tests import pyunit_utils
+import random
+
+def mergeOneEmptyFrame():
+    # PUBDEV-6987: merge with one empty frame and one normal frame.
```
honzasterba: if one of these frames is empty, do we need an empty csv in small data?

h2o-core/src/main/java/water/rapids/Merge.java
honzasterba: agree

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -52,6 +52,15 @@ public static Frame sort(final Frame fr, int[] cols, int[] ascending) {
...
+    }
+    if (riteFrame.numRows()==0 && leftFrame.numRows() > 0) {
+      return leftFrame;
+    }
+    if (riteFrame.numRows()==0 && leftFrame.numRows()==0)
+      throw new IllegalArgumentException("Cannot perform merge on empty frames.");
```
honzasterba: why not?

h2o-py/tests/testdir_munging/pyunit_6987_merge_one_empty.py
wendycwong: The csv contains the column names in this case.  It is not empty.

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: That should not be hard to do to just return an empty frame.

h2o-py/tests/testdir_munging/pyunit_6987_merge_one_empty.py
```diff
@@ -0,0 +1,24 @@
...
+    file1 = h2o.import_file(pyunit_utils.locate("smalldata/jira/pubdev-6987-a.csv"))
```
Pscheidl: I believe we don't need it. Let's create a pandas data frame `df = pd.DataFrame({"A1": [1], "A2": [0]})` and use `h2o.H2OFrame(df)`.

h2o-py/tests/testdir_munging/pyunit_6987_merge_one_empty.py
```diff
@@ -0,0 +1,24 @@
...
+    file2 = h2o.import_file(pyunit_utils.locate("smalldata/jira/pubdev-6987-b.csv"))
```
Pscheidl: Same story as above.

h2o-core/src/main/java/water/rapids/Merge.java
honzasterba: Please explain why you marked this as resolved, this seems incorrect

h2o-py/tests/testdir_munging/pyunit_6987_merge_one_empty.py
```diff
@@ -0,0 +1,26 @@
...
+    # PUBDEV-6987: merge with one empty frame and one normal frame.
+    file1 = h2o.H2OFrame(pd.DataFrame({"A1":[1], "A2":[0]}))
+    file3 = h2o.H2OFrame(pd.DataFrame({"A1":[1], "A2":[0]}))
+    file2 = file3.drop([0], axis=0) # create frame with empty content, just the headers
+
+    f1Mergef2 = file1.merge(file2)
+    f2Mergef1 = file2.merge(file1)
+    f2Mergef2 = file2.merge(file2)  # merging of empty frame with just headers
+
+    assert f1Mergef2[0,0]==file1[0,0] and f1Mergef2[0,1]==file1[0,1], "Merge file error!"
```
honzasterba: this should be an empty frame I think

honzasterba:
also apparently we dont need pandas to create frames out of thin air, h2o has APIs of its own:
h2o.H2OFrame({'A':['Hello', 'World', 'Welcome', 'To', 'H2O', 'World'],
                                'n': [0,1,2,3,4,5]})

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Honza:

If both frames are empty, the original code will return an frame with zero rows.  Hence, I remove my change and it works as you requested.

h2o-py/tests/testdir_munging/pyunit_6987_merge_one_empty.py
wendycwong: It is an empty frame but do contain headers  That is what that assert statement is comparing.

wendycwong:
You are right Honza.  I use the thin air method to create the frames now.  Thanks!

h2o-py/tests/testdir_munging/pyunit_6987_merge_one_empty.py
honzasterba: I still think that an explicit assert for nrows would add clarity to the test

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Checked with Matt Dowle: if we are joining two frames and if one of them is empty, the return result should be empty as well.  Michalk/Honza, you guys are correct.  I made a mistake to mark this as resolved.  Sorry

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -125,7 +125,13 @@ public static Frame merge(final Frame leftFrame, final Frame riteFrame, final in
     final BigInteger riteBase = hasRite ? riteIndex._base [0] : leftBase;
 
     // initialize for double columns, may not be used....
-    long leftMSBfrom = riteBase.subtract(leftBase).shiftRight(leftShift).longValue();
+    boolean leftFrameEmpty = leftFrame.numRows()==0;
+    boolean riteFrameEmpty = riteFrame.numRows() == 0;
+    BigInteger bigZero = BigInteger.ZERO;
+    long leftMSBfrom = riteFrameEmpty? (leftFrameEmpty? bigZero.shiftRight(leftShift).longValue()
```
wendycwong: Add boolean to check if a frame is empty.

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -125,7 +125,13 @@ public static Frame merge(final Frame leftFrame, final Frame riteFrame, final in
...
+                    : bigZero.shiftRight(leftShift).longValue()):
+            (leftFrameEmpty?riteBase.shiftRight(leftShift).longValue():
+                    riteBase.subtract(leftBase).shiftRight(leftShift).longValue());
     boolean riteBaseExceedsleftBase=riteBase.compareTo(leftBase)>0;
```
wendycwong: leftMSBfrom/leftMSBto/riteMSBfrom/riteMSBto are all changed if there is empty frame involved.  If nothing is done, one of the values will be Long.MAX_VALUE and this will create the stall the user has reported.

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -125,7 +125,13 @@ public static Frame merge(final Frame leftFrame, final Frame riteFrame, final in
     final BigInteger riteBase = hasRite ? riteIndex._base [0] : leftBase;
 
     // initialize for double columns, may not be used....
-    long leftMSBfrom = riteBase.subtract(leftBase).shiftRight(leftShift).longValue();
+    boolean leftFrameEmpty = leftFrame.numRows()==0;
```
honzasterba: incosistent formatting

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -125,7 +125,13 @@ public static Frame merge(final Frame leftFrame, final Frame riteFrame, final in
...
+    boolean riteFrameEmpty = riteFrame.numRows() == 0;
+    BigInteger bigZero = BigInteger.ZERO;
```
honzasterba: you could just static import ZERO

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -153,7 +159,10 @@ public static Frame merge(final Frame leftFrame, final Frame riteFrame, final in
     }
 
     BigInteger rightS = BigInteger.valueOf(256L<<riteShift);
-    long leftMSBto = riteBase.add(rightS).subtract(BigInteger.ONE).subtract(leftBase).shiftRight(leftShift).longValue();
+    long leftMSBto = riteFrameEmpty?(leftFrameEmpty? rightS.subtract(BigInteger.ONE).shiftRight(leftShift).longValue():
```
honzasterba: this is unreadable, please extract into method or just a line of if/else statements

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Got it.  Fixed.  Please check.

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Got it!

h2o-core/src/main/java/water/rapids/Merge.java
wendycwong: Expanded it to if/else statements.  Hope it is clear.  Thanks, Wendy

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -125,7 +126,16 @@ public static Frame merge(final Frame leftFrame, final Frame riteFrame, final in
...
     // initialize for double columns, may not be used....
-    long leftMSBfrom = riteBase.subtract(leftBase).shiftRight(leftShift).longValue();
+    boolean leftFrameEmpty = (leftFrame.numRows()==0);
+    boolean riteFrameEmpty = (riteFrame.numRows()==0);
+    long leftMSBfrom = riteBase.subtract(leftBase).shiftRight(leftShift).longValue();  // value when all frames nonempty
+    if (riteFrameEmpty) { // if riteFrame is empty, need to modify it because riteBase is Long.MIN_VALUE
+      if (leftFrameEmpty) { // if both left an
+        leftMSBfrom = 0;
+      } else {  // only riteFrame is empty
+      leftMSBfrom = ZERO.subtract(leftBase).shiftRight(leftShift).longValue();
```
honzasterba: indent

wendycwong: I will fix it.

PR #4019, opened 2019-10-25
PUBDEV-6339 split test to two

maurever:
Split the test to two - the first one uses smalldata datasets, the second one uses bigdata datasets.


michalkurka:
doesn't look like it is needed anymore - closing

PR #4020, opened 2019-10-25
PUBDEV-6737: expose monotone_constraints as a top level parameter in AutoML

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6737

This PR applies on top of https://github.com/h2oai/h2o-3/pull/4005
if we decide to expose `monotone_constraints` as a top level param in clients.

### DIFF
--- a/h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
+++ b/h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
@@ -209,9 +209,24 @@ public class AutoMLBuildSpecV99 extends SchemaV3<AutoMLBuildSpec, AutoMLBuildSpe
     @API(help="Custom algorithm parameters.", direction=API.Direction.INPUT)
     public AutoMLCustomParameterV99[] algo_parameters;
 
+    public KeyValueV3[] monotone_constraints;
+
     @Override
     public AutoMLBuildSpec.AutoMLBuildModels fillImpl(AutoMLBuildSpec.AutoMLBuildModels impl) {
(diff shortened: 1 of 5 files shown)


h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
```diff
@@ -209,9 +209,24 @@ V getFormattedValue() {
     @API(help="Custom algorithm parameters.", direction=API.Direction.INPUT)
     public AutoMLCustomParameterV99[] algo_parameters;
 
+    public KeyValueV3[] monotone_constraints;
+
```
seb-h2o: this may be necessary for support in `Flow` anyway.

h2o-py/h2o/automl/autoh2o.py
```diff
@@ -273,6 +276,11 @@ def assert_is_step(s):
                             plan.append(dict(name=name, steps=[dict(id=i) for i in ids]))
             self.build_models['modeling_plan'] = plan
 
+        if monotone_constraints is not None:
+            if algo_parameters is None:
+                algo_parameters = {}
+            algo_parameters['monotone_constraints'] = monotone_constraints
+
```
seb-h2o: Py+R clients don't need the REST API changes, they can just use the `algo_parameters` logic.

seb-h2o:
@ledell we didn't decide if we want to have `monotone_constraints` as a top level param.
This PR does it on top of the https://github.com/h2oai/h2o-3/pull/4005.
Please confirm if you want this, or if the generic param is enough.

PR #4021, opened 2019-10-27
PUBDEV-7015: Added MaterialisedWalker

deil87:
For more details see jira https://0xdata.atlassian.net/browse/PUBDEV-7015

"Added filterFunctions for interdependent hyper parameters" part of the original PR was moved to a separate PR https://github.com/h2oai/h2o-3/pull/4040

Materialised walker will be finished later once we get back to SMBO.
WIP.

### DIFF
--- a/h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
+++ b/h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
@@ -8,6 +8,9 @@ import water.exceptions.H2OIllegalArgumentException;
 import water.util.PojoUtils;
 
 import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static java.lang.StrictMath.min;
 
@@ -293,7 +296,7 @@ public interface HyperSpaceWalker<MP extends Model.Parameters, C extends HyperSp
       return _paramsBuilderFactory;
     }
 
-    protected MP getModelParams(MP params, Object[] hyperParams) {
+    public MP getModelParams(MP params, Object[] hyperParams) {
       ModelParametersBuilderFactory.ModelParametersBuilder<MP>
           paramsBuilder = _paramsBuilderFactory.get(params);
       for (int i = 0; i < _hyperParamNames.length; i++) {
@@ -408,6 +411,123 @@ public interface HyperSpaceWalker<MP extends Model.Parameters, C extends HyperSp
         public Object[] getCurrentRawParameters() {
           Object[] hyperValues = new Object[_hyperParamNames.length];
           return hypers(_currentHyperparamIndices, hyperValues);
+        }
+      }; // anonymous HyperSpaceIterator class
+    } // iterator()
+
+    /**
+     * Cartesian iteration over the hyper-parameter space, varying one hyperparameter at a
+     * time. Mutates the indices that are passed in and returns them.  Returns NULL when
+     * the entire space has been traversed.
+     */
+    private int[] nextModelIndices(int[] hyperparamIndices) {
+      // Find the next parm to flip
+      int i;
+      for (i = 0; i < hyperparamIndices.length; i++) {
+        if (hyperparamIndices[i] + 1 < _hyperParams.get(_hyperParamNames[i]).length) {
+          break;
+        }
+      }
+      if (i == hyperparamIndices.length) {
+        return null; // All done, report null
+      }
+      // Flip indices
+      for (int j = 0; j < i; j++) {
+        hyperparamIndices[j] = 0;
+      }
+      hyperparamIndices[i]++;
+      return hyperparamIndices;
+    }
+  } // class CartesianWalker
+
+
+  /**
+   * Hyperparameter space walker which visits each combination of hyperparameters in order.
+   */
+  public static class MaterializedRandomWalker<MP extends Model.Parameters>
+          extends BaseWalker<MP, HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria> {
+
+    public MaterializedRandomWalker(MP params,
+                                    Map<String, Object[]> hyperParams,
+                                    ModelParametersBuilderFactory<MP> paramsBuilderFactory,
+                                    HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria search_criteria,
+                                    ArrayList<Function<HashMap<String, Object>, HashMap<String, Object>>> filterFunctions) {
+      super(params, hyperParams, paramsBuilderFactory, search_criteria);
+    }
+
+    @Override
+    public HyperSpaceIterator<MP> iterator() {
+      ArrayList<HashMap<String, Object>> allElements = new ArrayList<>();
+
+      _hyperParams.forEach((hpKey, hpValues) -> {
+        if(allElements.isEmpty()) {
+          Arrays.stream(hpValues).forEach(hpValue -> {
+            HashMap<String, Object> newGridItem = new HashMap<>();
+            newGridItem.put(hpKey, hpValue);
+            allElements.add(newGridItem);
+          });
+        } else {
+          Stream<HashMap<String, Object>> expandedResult = allElements.stream().flatMap(existingItem -> {
+            return Arrays.stream(hpValues).map(hpValue -> {
+              HashMap<String, Object> clone = (HashMap<String, Object>) existingItem.clone();
+              clone.put(hpKey, hpValue);
+              return clone;
+            });
+          });
+          ArrayList<HashMap<String, Object>> collect = expandedResult.collect(Collectors.toCollection(ArrayList::new));
+          allElements.clear();
+          allElements.addAll(collect);
+        }
+      });
+      Collections.shuffle(allElements);
+
+      // TODO Apply `filterFunctions` here to `allElements`
+              
+      Iterator<Object[]> iterator = allElements.stream().map(item -> Arrays.stream(_hyperParamNames).map(item::get).toArray()).collect(Collectors.toCollection(ArrayList::new)).iterator();
+
+      return new HyperSpaceIterator<MP>() {
+        /** Hyper params permutation.
+         */
+
+        @Override
+        public MP nextModelParameters(Model previousModel) {
+          if (hasNext(previousModel)) { // or just iterator.hasNext()
+            // Get clone of parameters
+            MP commonModelParams = (MP) _params.clone();
+            // Fill model parameters
+            MP params = getModelParams(commonModelParams, getCurrentRawParameters()); // or just iterator.next()
+
+            return params;
+          } else {
+            throw new NoSuchElementException("No more elements to explore in hyper-space!");
+          }
+        }
+
+        @Override
+        public boolean hasNext(Model previousModel) {
+          return iterator.hasNext();
+        }
+
+        @Override public void reset() {
+          //TODO
+        }
+
+        @Override
+        public double time_remaining_secs() { return Double.MAX_VALUE; }
+
+        @Override
+        public double max_runtime_secs() { return Double.MAX_VALUE; }
+
+        public int max_models() { return _maxHyperSpaceSize > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int)_maxHyperSpaceSize; }
+
+        @Override
+        public void modelFailed(Model failedModel) {
+          // nada
+        }
+
+        @Override
+        public Object[] getCurrentRawParameters() {
+          return iterator.next();
         }
       }; // anonymous HyperSpaceIterator class
     } // iterator()
--- a/h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
+++ b/h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
@@ -442,17 +442,36 @@ public interface HyperSpaceWalker<MP extends Model.Parameters, C extends HyperSp
 
 
   /**
-   * Hyperparameter space walker which visits each combination of hyperparameters in order.
+   * Random iteration over the hyper-parameter space.  Does not repeat previously-visited combinations. 
+   * Implementation differs from {@link RandomDiscreteValueWalker} as we first eagerly evaluate the whole space, shuffle it and then iterate sequentially.
+   * Materialised grid might be taking considerable amount of memory so it needs to be taken into account.
    */
   public static class MaterializedRandomWalker<MP extends Model.Parameters>
           extends BaseWalker<MP, HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria> {
 
+    Random random;
+    ArrayList<Function<HashMap<String, Object>, HashMap<String, Object>>> _filterFunctions;
+
     public MaterializedRandomWalker(MP params,
                                     Map<String, Object[]> hyperParams,
                                     ModelParametersBuilderFactory<MP> paramsBuilderFactory,
-                                    HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria search_criteria,
+                                    HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria searchCriteria) {
+      this(params, hyperParams, paramsBuilderFactory, searchCriteria, null);
+    }
+    
+    public MaterializedRandomWalker(MP params,
+                                    Map<String, Object[]> hyperParams,
+                                    ModelParametersBuilderFactory<MP> paramsBuilderFactory,
+                                    HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria searchCriteria,
                                     ArrayList<Function<HashMap<String, Object>, HashMap<String, Object>>> filterFunctions) {
-      super(params, hyperParams, paramsBuilderFactory, search_criteria);
+      super(params, hyperParams, paramsBuilderFactory, searchCriteria);
+      
+      if (-1 == searchCriteria.seed())
+        random = new Random();                       
+      else
+        random = new Random(searchCriteria.seed());
+      
+      _filterFunctions = filterFunctions;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -437,6 +440,123 @@ public void modelFailed(Model failedModel) {
     }
   } // class CartesianWalker
 
+
+  /**
+   * Hyperparameter space walker which visits each combination of hyperparameters in order.
```
michalkurka: In what order? It seems like the order is random.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -437,6 +440,123 @@ public void modelFailed(Model failedModel) {
...
+              clone.put(hpKey, hpValue);
+              return clone;
+            });
+          });
+          ArrayList<HashMap<String, Object>> collect = expandedResult.collect(Collectors.toCollection(ArrayList::new));
+          allElements.clear();
+          allElements.addAll(collect);
+        }
+      });
+      Collections.shuffle(allElements);
```
michalkurka: All operations involving a random number generator need to be seeded. Grid search should be reproducible.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -293,7 +296,7 @@ public MP getParams() {
       return _paramsBuilderFactory;
     }
 
-    protected MP getModelParams(MP params, Object[] hyperParams) {
+    public MP getModelParams(MP params, Object[] hyperParams) {
```
michalkurka: fiy @Pscheidl is making changes in the grid search API right now

Pscheidl: I'll do the merge on my own or help with the merge if required, whichever comes first. There is a small refactor which actually extracts all the walkers out of the interface class (otherwise it would get really big and messy). Nothing else I can think of that touches this PR.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -437,6 +440,123 @@ public void modelFailed(Model failedModel) {
...
+
+        public int max_models() { return _maxHyperSpaceSize > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int)_maxHyperSpaceSize; }
+
+        @Override
+        public void modelFailed(Model failedModel) {
+          // nada
+        }
+
+        @Override
+        public Object[] getCurrentRawParameters() {
```
michalkurka: just to make sure - is this correct? this method gets called when the model building fails

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -437,6 +440,123 @@ public void modelFailed(Model failedModel) {
...
   } // class CartesianWalker
 
+
+  /**
+   * Hyperparameter space walker which visits each combination of hyperparameters in order.
+   */
+  public static class MaterializedRandomWalker<MP extends Model.Parameters>
+          extends BaseWalker<MP, HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria> {
+
+    public MaterializedRandomWalker(MP params,
```
michalkurka: Is this meant to be accessible to users? The search space for the default AutoML GBM grid search is about 80k. That is a lot of hashmap objects.

h2o-extensions/target-encoder/src/jmh/java/hex/NewRGSBench.java
```diff
@@ -0,0 +1,100 @@
...
+@Warmup(iterations = 2)
+@Measurement(iterations = 3)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class NewRGSBench {
+
+
+  @Benchmark
+  public void traverseWholeGridNew() {
+    HashMap<String, Object[]> hpGrid = new HashMap<>();
```
michalkurka: this seems to be a copy of code that is in the MaterializedRandomWalker

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: @michalkurka It is a copy paste issue( see CartesianWalker's comment).  PR had WIP label and I had not assigned any reviewers.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: @michalkurka yes, definitely it will be added.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
michalkurka: > PR had WIP label and I had not assigned any reviewers.

No problem! Happy to help and give early feedback!

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: @michalkurka thanks, I know that @Pscheidl is working on the parallelisation of RGS. I had taken a look at his changes before starting to this PR. I think that merge should not be too difficult. But yes, we are working "close to each other" and should take it into account.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: @michalkurka thank you! I appreciate early feedback. Just unfortunately it is difficult at this stage to avoid issues like with this comment.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: @michalkurka most likely it is not correct, and taking into account name of the method we should not return next element but last one, i.e. current one. Thanks!

h2o-extensions/target-encoder/src/jmh/java/hex/NewRGSBench.java
deil87: @michalkurka yes, hence the comment here https://github.com/h2oai/h2o-3/pull/4021/files#diff-73353bb7545487cc487e075203b698feR31 . Firstly I was benchmarking the core logic for traversing the grid then I've integrated it into MaterialisedRandomWalker and did separate benchmarking of the logic+walker_wrapper. In the end we can remove redundant  benchmarks

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: @michalkurka yes. I'm conscious about the tradeoff between lazy and eager evaluations. That is why it put word `materialised` in the name of the class `MaterializedRandomWalker`. Maybe this walker will not be perfect for all use cases(grids) as any solution in general might not be, but we will definitely need something like this for Bayesian approaches. Later we can change representation to make it more memory efficient. I would say that we can store grid items as just ordered(based on array of grid keys) arrays Object[]. 

Also as it is a Stream we can do lazy materialisation as well. But shuffling will be a difficulty here. We can materialise in batches and shuffle them but it might not be as good shuffling as given whole array of grid items.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -437,6 +440,132 @@ public void modelFailed(Model failedModel) {
...
+  /**
+   * Random iteration over the hyper-parameter space.  Does not repeat previously-visited combinations. 
+   * Implementation differs from {@link RandomDiscreteValueWalker} as we first eagerly evaluate the whole space, shuffle it and then iterate sequentially.
+   * Materialised grid might be taking considerable amount of memory so it needs to be taken into account.
+   */
+  public static class MaterializedRandomWalker<MP extends Model.Parameters>
+          extends BaseWalker<MP, HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria> {
+
+    Random random;
+    ArrayList<Function<HashMap<String, Object>, HashMap<String, Object>>> _filterFunctions;
```
seb-h2o: does it have to be `HashMap`? wouldn't work if the filter only manipulates `Map`?

deil87: I will think on the best representation of the grid item as we will need to consider memory impact.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -437,6 +440,132 @@ public void modelFailed(Model failedModel) {
...
+          ArrayList<HashMap<String, Object>> collect = expandedResult.collect(Collectors.toCollection(ArrayList::new));
+          allElements.clear();
+          allElements.addAll(collect);
+        }
+      });
+      
+      Collections.shuffle(allElements, random);
+
+      if(_filterFunctions != null) {
+        // unfortunately we don't have foldLeft/foldRight methods
```
seb-h2o: you mean `reduce` because it looks like that's what you're trying to do here with your clean/addAll.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
seb-h2o: > Later we can change representation to make it more memory efficient. I would say that we can store grid items as just ordered(based on array of grid keys) arrays Object[].

Why not thinking only in terms of indices?
Param names are already sorted according to `_hyperParamNames`.
You know how many potential values for each `hyperParam`.
So , with simple arithmetic, you can build a unique match between each param combination and an index `i`.
You build an array [0....max], and this is easy to randomize.
Then for each number, you reverse the arithmetic to recover the params name/value, apply the filter and so on...

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: @seb-h2o no, I meant fold as I'm iterating over functions but want to return grid items. Pseudocode from Scala:
```
val finterFunctions = Seq(function1, function2)
val allGridItems: Seq[HashMap[String, Object]] = Seq(gridItem1,.... gridItemN)
val filteredGridItems:Seq[HashMap[String, Object]] = filterFuntions.foldLeft(allGridItems)(case (nextFilterFunction, result) => nextFilterFunction.apply(result) ) 
```

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: @seb-h2o as I mentioned above I will think about optimisation part once I prepare everything else . It is always a tradeoff..... once you store as indices you will spend more time for applying functions and the more functions you have the more times you will need to convert. We can change representation after we filtered everything - that is what I was thinking about. Let me finish and then we can discuss.... there is a WIP label.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
seb-h2o: but `fold` is just `reduce` with a starting element, right?
I'm pretty sure this is equivalent to your fold: https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#reduce-T-java.util.function.BinaryOperator-

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
seb-h2o: It's arithmetic + random access! It's always fast!

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: No, it is not. You can do something with reduce that would be similar. But you can't implement fold reduce. https://stackoverflow.com/questions/41240414/equivalent-of-scalas-foldleft-in-java-8 . I will try to do it with reduce...this was the reason why I added comment for myself.

seb-h2o: Yeah, they are semantically different in their constraints, but for what you are trying to achieve, `reduce` would work, and maybe even better `collect`.

deil87: The global issue here is that I wanted to start with not an identity element but with collection of all grid elements. These would prevent us from being able to use parallelisation 
```
_filterFunctions.parallelStream().reduce(allElements, ...)
```
as we won't be able to combine properly two filtered array lists.

But if we want to use no-parallel stream then we might be fine but only with `reduce` not `collect`. I think you can't do it with `collect` in a functional way:

First thing to point out is that  collect() is using BiConsumer and it is working only through side-effects. 
From BiConsumer's javadoc 
```
Represents an operation that accepts two input arguments and returns no
 * result.  This is the two-arity specialization of {@link Consumer}.
 * Unlike most other functional interfaces, {@code BiConsumer} is expected
 * to operate via side-effects.
```
So given the fact that BiConsumer operates only through side-effect, solution would not be any better then a current one.
```
BiConsumer<ArrayList<HashMap<String, Object>>, Function<HashMap<String, Object>, HashMap<String, Object>> >
                biConsumerApplyFunToArray = (a, b) -> {
          ArrayList<HashMap<String, Object>> filtered = a.stream().map(b::apply)
                  .filter(Objects::nonNull) // TODO we actually need only one of these actions `filter`/`distinct` depending on the type of the function 
                  .distinct()
                  .collect(Collectors.toCollection(ArrayList::new));
          allElements.clear();
          allElements.addAll(filtered);
        };

        BiConsumer<ArrayList<HashMap<String, Object>>, ArrayList<HashMap<String, Object>>> combiner = (a,b) -> {
          a.addAll(b); 
          allElements.clear();
          allElements.addAll(a);
        };

        ArrayList<HashMap<String, Object>> result = _filterFunctions.stream().collect(() -> allElements, biConsumerApplyFunToArray, combiner);
        allElements.clear();
        allElements.addAll(result);
``` 
or something similar as I did not get it working actually.


With `reduce` we can do better, but only one of the overloaded versions will work for us as here we can introduce type U ( which is similar to the fold):
```
<U> U reduce(U identity,
                 BiFunction<U, ? super T, U> accumulator,
                 BinaryOperator<U> combiner);
```
But what I don't like with this approach is that you will have to provide dummy `combiner` 
that will never be actually used ( as we can't use `.parallelStream`).
Combiner:
```  
(a, b) -> {
          return a;
        }
```
or something like this(can't put `null`).

So possible solution with `reduce` will be as follows:

```
BiFunction<ArrayList<HashMap<String, Object>>, Function<HashMap<String, Object>, HashMap<String, Object>>, ArrayList<HashMap<String, Object>> >
                biFunctionApplyFunToArray = (a, b) -> {
          return a.stream().map(b::apply)
                  .filter(Objects::nonNull)
                  .distinct()
                  .collect(Collectors.toCollection(ArrayList::new));
        };

        ArrayList<HashMap<String, Object>> result = _filterFunctions.stream().reduce(allElements, biFunctionApplyFunToArray, (a, b) -> {
          return a;
        });
        allElements.clear();
        allElements.addAll(result);
```

PS I would not say that `reduce` with identity element is a reduce. It is closer to fold but not exactly fold. We are fine here as we don't need combiner for parallelisation. If we needed parallelisation we are on one hand lucky that we use monoid like type ArrayList( integers for example would not combine) but on the other hand unlucky that we don't want to start with empty identity element.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
honzasterba: you should definitely use interfaces here, you can weight in memory/performace when actualy instantiating the values

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
michalkurka: @deil87 just noticed this:

> _filterFunctions.parallelStream().reduce(allElements, ...)

I don't really think we can use "parallel" in Java because it uses a different Fork/Join pool than H2O. Just thought I would warn you early :)

BTW - why do we need to materialize all the parameters? Is that somehow required by the filtering function?

deil87: @michalkurka 
> I don't really think we can use "parallel" in Java because it uses a different Fork/Join pool than H2O

Thanks, I was just talking in general about `fold `and `reduce`. Was not going to use `.parallelStream` in H2O.

> BTW - why do we need to materialize all the parameters? Is that somehow required by the filtering function?

I understand your interest here. There are couple of points from my side:

1) I don't think it is a requirement. But I was thinking about performance benefits. I would say that under certain conditions sequential evaluation on the whole collection of grid items is more efficient(than one by one) as inputs to filtering functions will be reducing in sizes towards the end of the collection of functions. 
2) We need grid to be materialised to be able to apply Bayesian approaches( e.g. Sequential model based optimisation) in the near future (probably mentioned t his point before). And overall I think that random selection of indices is inefficient way of traversing a space. In theory it might take forever to visit all possible permutations.  With materialised version we don't need to 1)compute hashes of indices, 2) check whether current combination has already been visited 3) store hashes in memory( not a big deal of memory required though. btw we can use HashSet for visited permutations instead of LinkedHashSet as we don't use what `Linked` is providing). Micro-benchmarking proves that materialised version is faster.  But yes, it will take more space( actually not more but just we will be keeping more at the same time). 
Ideally, as it always the case with trade-off situations, we should decide dynamically - if the grid is big then lazy evaluation if it is relatively small then we better work with materialised version.

Summary:
1) There is no hard dependency between materialisation and filtering
2) We will need Materialised version of the grid for SMBO
3) Filtering could and should be added to all Walkers. This PR was just an example on how it could be implemented for a particular walker. 

I think that I will split this PR into two:
1) Filtering support for existing BaseWalkers 
2) MaterialisedWalker(with filtering support as well)

deil87: Forgot to mention one point that made me favour materialised version - it was easier to implement "shadowing" of hyperparameters. When you evaluate one combination from the subgroup and you need to mark all the other from this subgroup as redundant. But it could be implemented without materialisation as well.

PR #4024, opened 2019-10-28
Bring back deleted helper method & make Wendy happy again!

michalkurka:


wendycwong:
Please in the future, do not use the phrase,

make xxxx xxxx again! 

Bad memories there.


Thanks, Wendy

PR #4027, opened 2019-10-30
Better reporting when Watchdog Client disconnets

michalkurka:
This PR combines 2 changes

- if an H2O node doesn't see the watchdog client but the other nodes do - report the ones that do see it
- only the leader node can shutdown the cluster if the watchdog disappears (consensus is reached) - this has 2 purposes - the logs are clearer + there should be no connection error messages when all nodes are trying to shutdown the cluster at the same time

### DIFF
--- a/h2o-core/src/main/java/water/FailedNodeWatchdogExtension.java
+++ b/h2o-core/src/main/java/water/FailedNodeWatchdogExtension.java
@@ -246,9 +246,15 @@ public class FailedNodeWatchdogExtension extends AbstractH2OExtension {
                             Log.warn("Asking the rest of the nodes in the cloud whether watchdog client is really gone.");
                             List<H2ONode> nodesWithClient = tsk.doAllNodes().getNodesWithConnectedClient();
                             if(nodesWithClient.isEmpty()) {
-                                Log.fatal("Stopping H2O cloud since the watchdog client is disconnected from all nodes in the cluster!");
-                                // we should fail with negative status as this is not planned shutdown
-                                H2O.shutdown(-1);
+                                // we let the leader do the work to have a single point of shutdown initialization
+                                // Note: we know leader is working properly because it responded to MRTask
+                                if (H2O.SELF == H2O.CLOUD.leader()) {
+                                    Log.fatal("Stopping H2O cloud since the watchdog client is disconnected from all nodes in the cluster!");
+                                    // we should fail with negative status as this is not planned shutdown
+                                    H2O.shutdown(-1);
+                                } else {
+                                    Log.warn("Watchdog client is disconnected - leader node " + H2O.CLOUD.leader() + " will shutdown the cluster.");
+                                }
                             } else {
                                 Log.warn("This H2O node doesn't see client " + client + " but others do: " + nodesWithClient);
                             }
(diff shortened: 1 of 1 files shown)


h2o-core/src/main/java/water/FailedNodeWatchdogExtension.java
```diff
@@ -169,28 +172,42 @@ public void run() {
...
         public void reduce(WatchdogClientDisconnectedTask mrt) {
-            this.clientDisconnectedConsensus = this.clientDisconnectedConsensus && mrt.clientDisconnectedConsensus;
+            for (int i = 0; i < clientConnected.length; i++) {
+                clientConnected[i] &= mrt.clientConnected[i];
+            }
         }
 
         @Override
         protected void setupLocal() {
+            clientConnected = new boolean[H2O.getCloudSize()];
```
mn-mikke: Just a question: Is the number of nodes final at this point?

h2o-core/src/main/java/water/FailedNodeWatchdogExtension.java
```diff
@@ -227,10 +244,19 @@ public void run() {
...
                             Log.warn("Asking the rest of the nodes in the cloud whether watchdog client is really gone.");
-                            if((tsk.doAllNodes()).clientDisconnectedConsensus) {
-                                Log.fatal("Stopping H2O cloud since the watchdog client is disconnected from all nodes in the cluster!");
-                                // we should fail with negative status as this is not planned shutdown
-                                H2O.shutdown(-1);
+                            List<H2ONode> nodesWithClient = tsk.doAllNodes().getNodesWithConnectedClient();
+                            if(nodesWithClient.isEmpty()) {
+                                // we let the leader do the work to have a single point of shutdown initialization
+                                // Note: we know leader is working properly because it responded to MRTask
+                                if (H2O.SELF == H2O.CLOUD.leader()) {
```
mn-mikke: Maybe in this case it's irrelevant, but is a good practice to compare `H2ONode`s based on references?

h2o-core/src/main/java/water/FailedNodeWatchdogExtension.java
michalkurka: Yes, it should be. The cloud will be locked. The reason why it is final is that we actually do see a client. Because the client is run only after the cloud is fully formed we can relly on the fact the list of nodes won't change.

h2o-core/src/main/java/water/FailedNodeWatchdogExtension.java
michalkurka: This is safe to do because all the nodes are interned. And you can be sure that SELF is in `H2O.CLOUD._memary`.

h2o-core/src/main/java/water/FailedNodeWatchdogExtension.java
mn-mikke: Ok, perfect. Thanks for the explanation!

h2o-core/src/main/java/water/FailedNodeWatchdogExtension.java
mn-mikke: Ok, make sense. Thanks!

jakubhava:
@michalkurka I believe we can close this PR as watchdog client is already removed. Feel free to re-open if you need it

PR #4028, opened 2019-10-30
Report when a client actually does connect to the H2ONode

michalkurka:


h2o-core/src/main/java/water/H2ONode.java
```diff
@@ -39,14 +39,28 @@
   transient public volatile HeartBeat _heartbeat;  // My health info.  Changes 1/sec.
   transient public int _tcp_readers;               // Count of started TCP reader threads
     
-  transient private short _timestamp = H2ONodeTimestamp.UNDEFINED;
+  transient private short _timestamp;
```
jakubhava: Minor: I like the explicit initialization to stress out the default value

michalkurka: The issue is that intellij reports it as unused and it can get confusing because the default might even be different (if someone changes it and forgets to do it here)

jakubhava: I will just explain my initial motivation for this: The goal was to use `H2ONodeTimestamp.UNDEFINED` everywhere so we have single place to change it if the default value changes. I would say it is just coincidence that UNDEFINED is equal to 0.

michalkurka: yes, but that is not why it is redundant - the value is always overwritten in the constructor of the class - which is why this one is unused, regardless of what value it has (and that is why there is a risk of going stale)

PR #4032, opened 2019-10-31
Build: use an absolute value for free space check

michalkurka:


michal-raska:
I think we should bump the version of hadoop images

PR #4034, opened 2019-10-31
PUBDEV-7026: Added an option to disable H2O Flow

michalkurka:


michalkurka:
@jakubhava thanks for the SW issue

ottok92:
How can we disable h2o flow in Sparkling-Water via Spark configs?

ottok92:
FYI I found it here:
https://github.com/h2oai/sparkling-water/pull/877

PR #4036, opened 2019-11-01
[PUBDEV-7027] Cleanup of code used for client propagation

jakubhava:
This is the small cleanup we discussed on Slack -> explicitly call intern of the client with the correct timestamp instead of serializing H2ONode instance

### DIFF
--- a/h2o-core/src/main/java/water/UDPClientEvent.java
+++ b/h2o-core/src/main/java/water/UDPClientEvent.java
@@ -2,6 +2,8 @@ package water;
 
 import water.util.Log;
 
+import java.net.InetAddress;
+
 /**
  * A simple message which informs cluster about a new client
  * which was connected.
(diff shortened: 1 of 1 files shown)


michalkurka:
This PR breaks one invariant that currently exists in the codebase -> the timestamp is only defined if the client actually did connect to the h2o instance.

I think it is useful to keep that invariant, wdyt?

jakubhava:
Yup, that make sense, I can just remove the timestamp specification

h2o-core/src/main/java/water/UDPClientEvent.java
```diff
@@ -26,22 +28,21 @@ AutoBuffer call(AutoBuffer ab) {
         // Connect event is not sent in multicast mode
         case CONNECT:
           if (H2O.isFlatfileEnabled()) {
-            Log.info("Client reported via broadcast message " + ce.clientNode + " from " + ab._h2o);
-
-            // It is important to propagate Client's HeartBeat information to the rest of the nodes
```
michalkurka: Is this supposed to be removed?

jakubhava: Not removed, just moved -> we do first intern and then report

jakubhava:
Removed the timestamp specification so it is undefined now.

btw: It would be good to document these invariants at some place

h2o-core/src/main/java/water/UDPClientEvent.java
michalkurka: I am sorry, I do not see it there. Where is `setHeartBeat` being called?

jakubhava: ah, this :) This is removed. The thought behind this is that the heartbeat will be filled when the node starts communication with the client

michalkurka: I see. This means we are losing the information that a given node was a client until the heartbeat actually does come.

It might not be an issue, however, there are places that assume the information is available: like this one in Paxos:

```
      if(H2O.isFlatfileEnabled()){
        for(H2ONode n: H2O.getFlatfile()){
          if(!n._heartbeat._client && !PROPOSED.containsKey(n._key)){
```

jakubhava: I see, I will revert this change, don't want to change any behaviour with this change. Thank you for great catch!

jakubhava: I have put back the heart beat, please have a look @michalkurka

jakubhava:
The code will be probably removed in the future + not critical, want to avoid stall PR

PR #4038, opened 2019-11-04
PUBDEV-6945: small refactorings on Rabit tracker

honzasterba:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitTrackerH2O.java
```diff
@@ -96,11 +95,12 @@ public void stop() {
         }
     }
 
-    private class RabitTrackerH2OThread extends Thread {
-        private RabitTrackerH2O tracker;
+    private static class RabitTrackerH2OThread extends Thread {
+
+        private final RabitTrackerH2O tracker;
 
         private LinkMap linkMap;
```
Pscheidl: This varable's scope can be the `run` method only. No need to define it here ?

honzasterba: nice catch!

PR #4042, opened 2019-11-05
PUBDEV-6580 Python client fail on unsupported java version

koniecsveta:


h2o-py/h2o/backend/server.py
```diff
@@ -367,7 +368,14 @@ def _check_java(java, verbose):
                                   "Please download the latest 64-bit Java SE JDK from Oracle.")
         if "Client VM" in jver:
             warn("  You have a 32-bit version of Java. H2O works best with 64-bit Java.\n"
-                 "  Please download the latest 64-bit Java SE JDK from Oracle.\n")
+                 "  Please download the latest 64-bit Java SE JDK from Oracle.\n") 
+        H2OLocalServer._has_compatible_version(jver)
+
+    @staticmethod
+    def _has_compatible_version(java_version):
+        pattern = re.compile("1\\.[1-7]\\.")
```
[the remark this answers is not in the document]
koniecsveta: fixed in 38084c3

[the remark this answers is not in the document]
michalkurka: @honzasterba

The way it was meant to work: H2O in Java maintains a whitelist, while the client maintains a blacklist. The blacklist items are the Java versions that H2O definitely cannot work with. The blacklist only contains old versions, we are not saying anything about future versions. This is how it was done in R.

H2O-3 can actually work on Java 13 - it is just not officially supported. H2O-3 cannot work on Java 6 for example, it is not compiled for that.

IMHO we should not exclude Java 13 here. The check on Java side can be defeated - this check cannot be disabled.

michalkurka:
@koniecsveta 

Please see my inline comments - on the client-side we should only maintain a blacklist (existing Java versions that we don't support - eg. Java 6, on master Java 7). Just like in R.

My take on the jira would be different: I would let H2O (in Java) do the work of determining if the java version is supported or not but make sure we clearly communicate the back to the users.

So maybe this has 2 tasks:
1) Quickly reject obviously unsupported Javas - 6, 7 (on master). Basically want to reject any version that H2O was not compiled for. If the target was 8, we should reject anything bellow that.
2) Extend the check in Java to propagate the Java version error better to the client. So that the client gets a nice and clear message about what was wrong.

cc: @honzasterba

h2o-app/src/main/java/water/H2OApp.java
```diff
@@ -1,10 +1,13 @@
 package water;
 
 public class H2OApp extends H2OStarter {
+  
+  public static int BAD_JAVA_VERSION_RETURN_CODE = 3;
```
koniecsveta: Maybe it would make sense to do enum for all return codes used in code? What do you think? @michalkurka

h2o-py/h2o/backend/server.py
```diff
@@ -344,7 +345,10 @@ def _launch_server(self, port, baseport, mmax, mmin, ea, nthreads, jvm_custom_ar
         giveup_time = time.time() + self._TIME_TO_START
         while True:
             if proc.poll() is not None:
-                raise H2OServerError("Server process terminated with error code %d" % proc.returncode)
+                if proc.returncode == self._BAD_JAVA_VERSION_RETURN_CODE_:
```
honzasterba: we should do the same for R (can be a separate PR)

h2o-py/h2o/backend/server.py
```diff
@@ -344,7 +345,10 @@ def _launch_server(self, port, baseport, mmax, mmin, ea, nthreads, jvm_custom_ar
...
+                    raise H2OServerError("Java version check failed: Only Java 8, 9, 10, 11 and 12 are supported.")
```
honzasterba: I think this error message needs to say that the server failed to start or something along those lines as well

h2o-py/h2o/backend/server.py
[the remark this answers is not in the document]
koniecsveta: fixed in 8900af9

h2o-py/h2o/backend/server.py
koniecsveta: fixed in 80d407c

h2o-py/h2o/backend/server.py
koniecsveta: found this one where it is done: c86af86

honzasterba: I think we should keep Python and R consisent and implement the same solution here

koniecsveta: stderr added to the the server error in a3da706

PR #4044, opened 2019-11-05
[PUBDEV-7040] Expose version details in cloud info

jakubhava:


jakubhava:
Thank you @pavel for review, I waited fro tests and fixed one issue, tests are green now

PR #4049, opened 2019-11-06
PUBDEV-7006: Added missing parameters noise and seed to R new TE API

deil87:
This PR is only dedicated to a fix in R.

Agenda:
`h2o.transform` method for H2OTargetEncoderModel in R was missing `noise` parameter and also `seed` parameter.


h2o-r/tests/testdir_algos/targetencoder/runit_targetencoder_model.R
```diff
@@ -1,6 +1,10 @@
 setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
 source("../../../scripts/h2o-r-test-setup.R")
 
+doTestAndContinue <- function(testDesc, test) {
```
michalkurka: Please use the already existing method `doTest`. This is copy-paste-modify. Not maintainable.

h2o-r/h2o-package/R/models.R
```diff
@@ -514,7 +514,9 @@ setMethod("h2o.transform", signature("H2OTargetEncoderModel"), function(model, d
                                                                         data_leakage_handling = NULL,
                                                                         use_blending = NULL,
                                                                         inflection_point = -1,
-                                                                        smoothing = -1) {
+                                                                        smoothing = -1,
```
michalkurka: Please document the added parameters.

h2o-r/tests/testdir_algos/targetencoder/runit_targetencoder_model.R
deil87: @michalkurka sure, I'm aware of `doTest` method. Btw I wonder why we are ok with for example `get_partitioned_model_names` method in python then? But I agree that it is better to have utils that are being shared between all tests. The reason why I added this modification is that I don't wan't one-file-per-test approach. And I don't want to put all use cases into one test like we usually do in R. Maybe we should agree upon some util method and add it to test utils?

h2o-r/h2o-package/R/models.R
deil87: @michalkurka Thanks! I've missed that.

h2o-r/tests/testdir_algos/targetencoder/runit_targetencoder_model.R
deil87: I've found `doSuite` in utils so I think it will work for me.

deil87:
@michalkurka tests look fine. Going to merge this one today.

PR #4051, opened 2019-11-08
PUBDEV-6959 add more info into asserts

maurever:
JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-6959

### DIFF
--- a/h2o-core/src/main/java/water/util/ArrayUtils.java
+++ b/h2o-core/src/main/java/water/util/ArrayUtils.java
@@ -1729,9 +1729,13 @@ public class ArrayUtils {
 
     int end=Arrays.binarySearch(sortedSplitPoints, maxEx);
     if (end<0) end=-end-1;
-    assert(end>0 && end<= sortedSplitPoints.length);
-    assert(end>=start);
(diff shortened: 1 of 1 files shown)


h2o-core/src/main/java/water/util/ArrayUtils.java
```diff
@@ -1729,9 +1729,13 @@ else if (pos > min && pos < maxEx && (i==0 || pos != last)) {
 
     int end=Arrays.binarySearch(sortedSplitPoints, maxEx);
     if (end<0) end=-end-1;
-    assert(end>0 && end<= sortedSplitPoints.length);
-    assert(end>=start);
-    assert(sortedSplitPoints[end-1] < maxEx);
+    String splitPointInfo = "Split points info - split points length: "+sortedSplitPoints.length + " values: ";
```
michalkurka: This is not needed by a regular execution. Can we execute it only if the assert actually fails?

maurever: Yes, we can. :)

PR #4052, opened 2019-11-08
[PUBDEV-7053] Make Name of LdapLoginModule Independent on Jetty Version

mn-mikke:
Sparkling Water utilize Jetty 9 and expects `ai.h2o.org.eclipse.jetty.jaas.spi.LdapLoginModule` in the login conf. When SW runs in external backend, it starts H2O workers via H2O extended jar which contains Jetty 8 for hadoop 2 distributions. Thus H2O workers expect `ai.h2o.org.eclipse.jetty.plus.jaas.spi.LdapLoginModule` in the login conf.


mn-mikke:
@jakubhava I've added both classes so that the user doesn't have to think about what LDAP module belong to Jetty 8 and Jetty 9. If one the class doesn't make sense, I can remove it.

PR #4053, opened 2019-11-08
PUBDEV-6945: Make multi-node xgboost reproducible

honzasterba:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitWorker.java
```diff
@@ -154,6 +151,6 @@ public void assignRank(int rank, Map<Integer, RabitWorker> waitConn, LinkMap lin
 
     @Override
     public int compareTo(RabitWorker o) {
-        return host.compareTo(o.host);
+        return jobId.compareTo(o.jobId);
```
michalkurka: What about the `jobToRankMap`?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitTrackerH2O.java
```diff
@@ -39,8 +39,8 @@ public RabitTrackerH2O(int workers) {
         envs.put("DMLC_NUM_SERVER", "0");
         envs.put("DMLC_TRACKER_URI", H2O.SELF_ADDRESS.getHostAddress());
         envs.put("DMLC_TRACKER_PORT", Integer.toString(port));
+        envs.put("DMLC_TASK_ID", Integer.toString(H2O.SELF.index()));
```
michalkurka: What about the `jobToRankMap`?

Do we need to change/clean-up the codebase?

honzasterba: Now jobToRankMap was a noop structure, now it can actually work if we want to support restarting of crashed xgb worker processes

PR #4056, opened 2019-11-09
Disable client mode unless explicitly enabled

michalkurka:
H2O instances will need to be started with option `-allow_clients` to enable client connections.

For H2O on Hadoop, the `h2odriver` will need to be started with a flag `-sw_ext_backend`. This will automatically add `-allow_clients`. In the future, this option can enforce validation of security parameters (eg. make sure when security is enabled, internal_secure_connections is enabled as well).


michalkurka:
@jakubhava @mn-mikke please take a look and let me know if this is okay from SW point of view.

PR #4057, opened 2019-11-10
PUBDEV-7056: SPNEGO athentication would pass with any HTTP-Basic header

honzasterba:


h2o-core/src/main/java/water/webserver/H2OHttpViewImpl.java
```diff
@@ -68,16 +68,18 @@ public boolean authenticationHandler(HttpServletRequest request, HttpServletResp
...
+    if (request.getUserPrincipal() != null) {
+      String loginName = request.getUserPrincipal().getName();
+      if (loginName.equals(config.user_name)) {
+        return false;
+      } else {
+        Log.warn("Login name (" + loginName + ") does not match cluster owner name (" + config.user_name + ")");
+        ServletUtils.sendResponseError(response, HttpServletResponse.SC_UNAUTHORIZED, "Login name does not match cluster owner name");
+      }
+    } else {
+      ServletUtils.sendResponseError(response, HttpServletResponse.SC_UNAUTHORIZED, "Missing authentication details");
```
michalkurka: @honzasterba why is this needed? This handler should only be concerned with authorization = admit only users who are allowed to use the cluster. In the case of hash-authentication - anybody who is the config file, in case of other methods - only the user who started the cluster.

h2o-jetty-8/src/main/java/ai/h2o/org/eclipse/jetty/security/authentication/SpnegoAuthenticator.java
```diff
@@ -0,0 +1,141 @@
+package ai.h2o.org.eclipse.jetty.security.authentication;
+
+import java.io.IOException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
```
michalkurka: Please include all the original headers and any license information

PR #4054, opened 2019-11-08
[PUBDEV-7036] Update ModelJsonReader.java to use stderr for errors/warnings

Tagar:
PUBDEV-7036: Update ModelJsonReader's error output

Issue described in https://0xdata.atlassian.net/browse/PUBDEV-7036

Attempt 2 of https://github.com/h2oai/h2o-3/pull/4039

PR #4058, opened 2019-11-11
PUBDEV-7036 - Use stderr instead of stdout to show warnings in MojoReader

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7036

Continuation of https://github.com/h2oai/h2o-3/pull/4054.


Pscheidl:
The tests were already successful in the previous branch, but ran against the `master` branch. Waiting for tests in this branched, aimed towards `rel-yau`, then merging.

PR #4059, opened 2019-11-11
PUBDEV-7055: Change examples to insecure=False on Security Page

hannah-tillman:


h2o-docs/src/product/security.rst
```diff
@@ -237,7 +237,7 @@ with HTTPS:
 
 ::
 
-    h2o.init(ip = "a.b.c.d", port = 54321, https = TRUE, insecure = TRUE)
+    h2o.init(ip = "a.b.c.d", port = 54321, https = TRUE, insecure = FALSE)
```
angela0xdata: whoops, looks like this has been incorrect for a while. Python uses True/False, not TRUE/FALSE. 
h2o.init(ip="a.b.c.d", port=54321, https=True, insecure=False)

PR #4060, opened 2019-11-11
PUBDEV-6895: Fix internal secure connections

michalkurka:


h2o-core/src/main/java/water/network/SecurityUtils.java
```diff
@@ -96,15 +96,24 @@ public static String generateSSLConfig(SSLCredentials credentials) throws IOExce
...
+        sslConfig.put("h2o_ssl_jts", credentials.jts.name);
         sslConfig.put("h2o_ssl_jts_password", credentials.jts.pass);
         FileOutputStream output = new FileOutputStream(file);
-        sslConfig.store(output, "");
+        try {
+            sslConfig.store(output, "");
+            output.close();
+        } finally {
+            try {
+                output.close();
```
honzasterba: doing output.close(); twice, once in finally is enough?

h2o-core/src/main/java/water/network/SecurityUtils.java
```diff
@@ -96,15 +96,24 @@ public static String generateSSLConfig(SSLCredentials credentials) throws IOExce
...
+            } catch (IOException e) {
+                // ignore
```
honzasterba: do we really want to eat this exception?

h2o-core/src/main/java/water/network/SecurityUtils.java
michalkurka: It is not, this is one of the recommended approaches closing a resource if you don't use try-with-resources and don't want to lose the Exception accidentally. 

The point of the close inside of the try block is to make sure that if a close() produces an exception it is actually thrown and not suppressed. If close() inside of the finally-block generates an Exception we know it is okay to ignore because that can only happen if there was another root cause.

h2o-core/src/main/java/water/network/SecurityUtils.java
michalkurka: Yes, we can - thanks to the close() in try-statement.

PR #4062, opened 2019-11-11
PUBDEV-6895 bug case added to hadoopCommands.groovy

koniecsveta:


michalkurka:
thank you @koniecsveta

PR #4064, opened 2019-11-12
Scoring info not allowed to be ran simultaneously for paralle grid builder.

Pscheidl:
Constructing scoring info in parallel is potentially dangerous - the logic sorts scoring infos held inside an array internally. Uses built `_models` for that as well.


Pscheidl:
@honzasterba 

This should fix http://mr-0xc1:8080/job/h2o-3-nightly-pipeline/job/master/3464/artifact/r3.4-small-client-mode-disconnect-attack/h2o-3/h2o-r/tests/results/failed/testdir_algos_grid_runit_grid_parallel.R.out.txt

I'll wait for the tests to confirm. The issue is not hit 100% of a time, at least locally.

PR #4066, opened 2019-11-13
PUBDEV-6890: add platt scaling for XGBoost

honzasterba:


honzasterba:
testing TBD

h2o-algos/src/main/java/hex/tree/PlattScalingHelper.java
```diff
@@ -0,0 +1,124 @@
...
+import water.*;
+import water.fvec.Frame;
+import water.fvec.Vec;
+
+import static hex.ModelCategory.Binomial;
+
+public class PlattScalingHelper {
+    
+    public interface ModelBuilderWithCalibration<M extends Model<M , P, O>, P extends Model.Parameters, O extends Model.Output> {
+        void warn(String field, String msg);
```
maurever: Maybe I am blind, but I can't find where are the warn and error methods implemented...? Does the ModelBuilder override and implement these methods?

h2o-algos/src/main/java/hex/tree/PlattScalingHelper.java
```diff
@@ -0,0 +1,124 @@
...
+        int nclasses();
+        Frame init_adaptFrameToTrain(Frame f, String label, String param, boolean expensive);
+        Frame getCalibrationFrame();
+        void setCalibrationFrame(Frame f);
+        void setCalibrationOutput(GLM calibrationOutput, M model);
+    }
+
+    public interface ParamsWithCalibration {
+        boolean isCVModel();
+        String weightsColumn();
```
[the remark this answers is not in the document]
honzasterba: unfortunately yes, since ParamsWithCalibration does not extend Parameters so if I want to use an interface the only other option would be to cast ParamsWithCalibration to Parameters and access the public _fields, I went with the interface options, looks cleaner to me

h2o-algos/src/main/java/hex/tree/PlattScalingHelper.java
honzasterba: yes these are implemented in ModelBuilder, again I thought about adding something like IModelBuilder but I really hate I-prefixed I-nterfaces so instead I went with this.

h2o-algos/src/main/java/hex/tree/PlattScalingHelper.java
```diff
@@ -0,0 +1,124 @@
...
+                calibBuilder._parms._intercept = true;
+                calibBuilder._parms._response_column = "response";
+                calibBuilder._parms._train = calibInput._key;
+                calibBuilder._parms._family = GLMModel.GLMParameters.Family.binomial;
+                calibBuilder._parms._lambda = new double[] {0.0};
+                if (calibWeights != null) {
+                    calibBuilder._parms._weights_column = "weights";
+                }
+
+                builder.setCalibrationOutput(calibBuilder, model);
```
michalkurka: Can this method instead return the calibration model? I think it is a little unexpected that responsibility of this method is also to set it to the original builder.

h2o-algos/src/main/java/hex/tree/PlattScalingHelper.java
```diff
@@ -0,0 +1,124 @@
...
+                builder.error("_calibrate_model", "Calibration frame was not specified.");
+        }
+    }
+    
+    public static <M extends Model<M , P, O>, P extends Model.Parameters, O extends Model.Output> void buildCalibrationModel(
+        boolean finalScoring, ModelBuilderWithCalibration<M, P, O> builder, 
+        ParamsWithCalibration parms, Job job, M model
+    ) {
+        // Model Calibration (only for the final model, not CV models)
+        if (finalScoring && parms.calibrateModel() && (!parms.isCVModel())) {
```
michalkurka: maybe calibrateModel could also consider whether this is a cvModel - then on we could move the logic on the caller side

   if (finalScoring && params.calibrateModel())

IMHO better than to have the logic here

h2o-algos/src/main/java/hex/tree/PlattScalingHelper.java
michalkurka: they are already public on model builder, I know there is no pretty solution but maybe a little nicer would be to just add a method getModelBuilder() to gain access to all model builder methods? otherwise, you need to keep declaring them here whenever you need one

a similar approach with the parameter object

honzasterba: done

h2o-algos/src/main/java/hex/tree/PlattScalingHelper.java
honzasterba: done

h2o-algos/src/main/java/hex/tree/PlattScalingHelper.java
honzasterba: done

PR #4068, opened 2019-11-13
[PUBDEV-7060] Propagate offset_column to MOJO models

mn-mikke:
This PR suggests to propagate `offset_column` to MOJO models of GBM, GLM and DeepLearning.


Pscheidl:
Fold column can be found in  `experimental/modelDetails.json`. If required, it could be taken from there for older MOJOs. FYI. Als, it's marked as experimental, but in fact we rely on it heavily.

We do have offset column and other stuff in `model.ini`, so I guess such a basic information can be there as well.

michalkurka:
@mn-mikke thanks - please don't merge yet

Pscheidl:
@michalkurka put the offset column for all algos in https://github.com/h2oai/h2o-3/pull/4086

mn-mikke:
Thanks a lot @michalkurka

PR #4069, opened 2019-11-14
[PUBDEV-7061] Expose additional methods on SecurityUtils

jakubhava:
See https://0xdata.atlassian.net/browse/PUBDEV-7061 and

https://github.com/h2oai/sparkling-water/pull/1616


jakubhava:
@michalkurka The API is not changing, I just exposed 2 additional methods. Please also see the SW PR

jakubhava:
I can avoid changing this file, but then need to duplicate generateSSLConfig on Sparkling Water side

PR #4072, opened 2019-11-14
Jenkins: fix hdp-krb tests timeouts

honzasterba:
- too mum xmx causes hive tasks to wait forever
- cleanup memory in hadoop tests


michalkurka:
@honzasterba why is it needed to do the clean-up? a clean-up is automatically done before each test in, see utilsPY#standalone_test - it starts with `h2o.remove_all()`

honzasterba:
@michalkurka thanks for pointing that out, you are correct, these influence nothing, I have seen some OOME and thought this was the case, looks like it was not

PR #4078, opened 2019-11-18
PUBDEV-6754: provide actual ntree value

koniecsveta:


h2o-py/h2o/estimators/gbm.py
```diff
@@ -953,3 +953,10 @@ def check_constant_response(self, check_constant_response):
         self._parms["check_constant_response"] = check_constant_response
 
 
+    """
+      Returns actual number of trees in GBM model (whether CV is enabled or not).
```
michalkurka: How is this related to CV?

I think we can mention early stopping instead.

h2o-r/h2o-package/R/gbm.R
```diff
@@ -317,5 +317,8 @@ h2o.gbm <- function(x,
 
   # Error check and build model
   model <- .h2o.modelJob('gbm', parms, h2oRestApiVersion=3, verbose=verbose)
+
+  # Provide actual ntree value 
+  model@model$ntree_actual <- model@model$model_summary['number_of_trees'][,1]
```
michalkurka: Should we not provide a similar convenience method as we have in Python? Especially for the parity of APIs.

h2o-py/tests/testdir_algos/gbm/pyunit_PUBDEV-6754_gbm_ntree_actual.py
```diff
@@ -0,0 +1,30 @@
...
+  prostate_gbm = H2OGradientBoostingEstimator(nfolds=5,ntrees=ntrees_original, distribution="bernoulli", stopping_metric="MSE", stopping_tolerance=0.01, stopping_rounds=5)
+  prostate_gbm.train(x=list(range(2,9)), y=1, training_frame=prostate)
+  
+  print("\n")
+  print("number of trees set by user before building the model is:")
+  print(ntrees_original)
+  print("number of trees built with early-stopping is:")
+  print(prostate_gbm.ntree_actual())
+  
+  assert prostate_gbm.ntree_actual() == prostate_gbm._model_json['output']['model_summary']['number_of_trees'][0]
```
michalkurka: I would also suggest checking that the model actually stopped early. That is actual number of trees < 1000.

h2o-py/h2o/estimators/gbm.py
koniecsveta: It is taken from the ticket description. Will fix.

h2o-bindings/bin/custom/python/gen_gbm.py
```diff
@@ -3,6 +3,19 @@ def update_param(name, param):
         param['values'].remove('ordinal')
         return param
     return None  # param untouched
+ 
+def class_extensions():
```
honzasterba: do we want the same for other tree algos as well? drf, xgboost, isofor?

h2o-bindings/bin/custom/python/gen_gbm.py
```diff
@@ -3,6 +3,19 @@ def update_param(name, param):
...
+    """
+      Returns actual number of trees in GBM model (whether CV is enabled or not).
+
+      Type: ``float``
+      """
+    def ntree_actual(self):
```
honzasterba: I dont like the name. Everywhere we use ntrees, my suggestions:
- ntrees_actual
- ntrees_built

koniecsveta: Taken from the ticket description. Will change :)

h2o-bindings/bin/custom/python/gen_gbm.py
koniecsveta: not specified on the ticket but clarifying in #h2o-3-r-client

h2o-r/h2o-package/R/gbm.R
koniecsveta: After clarification with @ledell and @Pscheidl this version is considered to be sufficient since we don’t have accessor method for all the other model parameters (just a few) + it will be implemented for all tree-based algorithms.

michalkurka: Thanks @koniecsveta

h2o-py/tests/testdir_algos/sharedtree/pyunit_PUBDEV-6754_ntrees_actual_for_tree_algos.py
```diff
@@ -0,0 +1,68 @@
...
+from h2o.estimators.gbm import H2OGradientBoostingEstimator
+from h2o.estimators.isolation_forest import H2OIsolationForestEstimator
+from h2o.estimators.random_forest import H2ORandomForestEstimator
+from h2o.estimators.xgboost import H2OXGBoostEstimator
+
+def tree_algos_ntree_actual():
+  prostate = h2o.import_file(path=pyunit_utils.locate("smalldata/logreg/prostate.csv"))
+  prostate[1] = prostate[1].asfactor()
+  prostate.summary()
+  ntrees_original = 1000
```
honzasterba: ideally I would chose a much lower number here, so that the test runs as fast as possible (same for R test)

h2o-py/h2o/estimators/gbm.py
koniecsveta: fixed

h2o-py/tests/testdir_algos/gbm/pyunit_PUBDEV-6754_gbm_ntree_actual.py
koniecsveta: done

h2o-bindings/bin/custom/python/gen_gbm.py
koniecsveta: added for all tree algos

h2o-bindings/bin/custom/python/gen_gbm.py
koniecsveta: done

h2o-py/tests/testdir_algos/sharedtree/pyunit_PUBDEV-6754_ntrees_actual_for_tree_algos.py
koniecsveta: The test runs faster because of early stopping. trees_original is 1000 to give enough space for build to be early stopped and to be illustrative / that there is a difference between trees_original and trees_actual (so it builds ~20 trees for each model).

h2o-py/h2o/model/model_base.py
```diff
@@ -390,6 +390,20 @@ def scoring_history(self):
         print("No score history for this model")
 
 
+    def ntrees_actual(self):
+        """
+        Returns actual number of trees in GBM model. If early stopping enabled, GBM can reset the ntrees value.
```
honzasterba: mentions GBM when in facts applies to other algos

h2o-py/h2o/model/model_base.py
```diff
@@ -390,6 +390,20 @@ def scoring_history(self):
         print("No score history for this model")
 
 
+    def ntrees_actual(self):
```
honzasterba: I am not sure I like this approach, this adds this method to models where it does not make sense. @michalkurka what do you think?

michalkurka: Good point @honza, this is however a common approach - this class already does have algo-specific method.

I would keep this for now.

michalkurka: If you want to make a jira - I think we can fix this in the future (we can eg. dynamically remove methods from the instances).

honza: :wave:

h2o-py/h2o/model/model_base.py
koniecsveta: fixed

PR #4079, opened 2019-11-18
PUBDEV-6940: fix list of metrics exposed in Flow

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6940

And added support for `AUCPR` in AutoML.

### DIFF
--- a/h2o-core/src/main/java/hex/ModelMetricsBinomial.java
+++ b/h2o-core/src/main/java/hex/ModelMetricsBinomial.java
@@ -74,6 +74,8 @@ public class ModelMetricsBinomial extends ModelMetricsSupervised {
 
   // expose simple metrics criteria for sorting
   public double auc() { return auc_obj()._auc; }
+  public double pr_auc() { return auc_obj()._pr_auc; }
+  public double aucpr() { return auc_obj()._pr_auc; } // for compatibility with naming in ScoreKeeper.StoppingMetric annotation
   public double lift_top_group() { return gainsLift().response_rates[0] / gainsLift().avg_response_rate; }
 
   /**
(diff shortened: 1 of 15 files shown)


h2o-core/src/main/java/hex/ScoreKeeper.java
```diff
@@ -131,9 +131,10 @@ else if (m instanceof ModelMetricsMultinomial) {
     misclassification(ConvergenceStrategy.LESS_IS_BETTER, true),
     AUCPR(ConvergenceStrategy.MORE_IS_BETTER, true),
     mean_per_class_error(ConvergenceStrategy.LESS_IS_BETTER, true),
+    anomaly_score(ConvergenceStrategy.NON_DIRECTIONAL, false),
     custom(ConvergenceStrategy.LESS_IS_BETTER, false),
     custom_increasing(ConvergenceStrategy.MORE_IS_BETTER, false),
-    anomaly_score(ConvergenceStrategy.NON_DIRECTIONAL, false);
+    ;
```
seb-h2o: reordered for display in Flow

h2o-core/src/main/java/hex/ModelMetricsBinomial.java
```diff
@@ -74,6 +74,8 @@ public ConfusionMatrix cm(AUC2.ThresholdCriterion criterion) {
 
   // expose simple metrics criteria for sorting
   public double auc() { return auc_obj()._auc; }
+  public double pr_auc() { return auc_obj()._pr_auc; }
+  public double aucpr() { return auc_obj()._pr_auc; } // for compatibility with naming in ScoreKeeper.StoppingMetric annotation
```
seb-h2o: should we also add the `pr_auc` one?
`aucpr` is needed for the reflection logic in `ScoringInfo`

ledell: @seb-h2o Are there advantages to adding `pr_auc()` other than consistency (all the other clients have both?).

seb-h2o: other than consistency? no, as it was exposed only through REST API, which doesn't need those methods.
R currently has only `pr_auc` and Python has  `model.pr_auc()` though...

PR #4080, opened 2019-11-18
Adds custom assembly for h2odriver for CDP 7.0

michalkurka:
CDH 7.0 is unreleased, this built is based on dependencies for CDH 6.3
used in compile-time. All Hadoop depenedencies are then excluded -
this makes this driver build compatible with generally any Hadoop 3.x
distribution. However, further dependency changes will be made to made
this CDP 7.0 specific.

### DIFF
--- a/h2o-hadoop-3/h2o-cdp7.0-assembly/build.gradle
+++ b/h2o-hadoop-3/h2o-cdp7.0-assembly/build.gradle
@@ -1,0 +1,130 @@
+ext {
+    hadoopVersion = 'cdh6.3.0'
+    hadoopMavenArtifactVersion = '3.0.0-cdh6.3.0'
+    orcSupported = true
+    orcHiveExecVersion = "2.1.1-cdh6.3.0"
+}
+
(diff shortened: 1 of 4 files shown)


h2o-hadoop-3/h2o-cdp7.0-assembly/build.gradle
```diff
@@ -0,0 +1,130 @@
...
+
+apply plugin: 'java'
+apply plugin: 'com.github.johnrengelman.shadow'
+
+description = 'H2O HDFS client shadowjar for Hadoop ' + hadoopVersion
+
+sourceCompatibility = 1.7
+targetCompatibility = 1.7
+
+configurations {
```
honzasterba: do I understand it right that we can remove all this once cdh7 is released and replace it all with the usual apply from: '../assemblyjar.gradle'?

michalkurka: yes, that is correct

PR #4085, opened 2019-11-20
[PUBDEV-7083] JUnit tests are not being run for multiple modules

deil87:
jira with details https://0xdata.atlassian.net/browse/PUBDEV-7083

Modules that were not being testes:

- h2o-genmodel 

-  h2o-genmodule-extensions

 - h2o-jetty-8

- h2o-jetty-9

- h2o-mapreduce-generic

PS to improve further we can consider to move `H2OTestRunner` into a separate module that other modules can depend on. This way we will be able to avoid dependencies on h2o-core module especially for modules that don't want to do anything with h2o-core (e.g. `h2o-genmodel`). Still current solution imo should be fine as it is only `testCompile` type of dependency.


h2o-mapreduce-generic/build.gradle
```diff
@@ -17,6 +17,18 @@ dependencies {
     exclude module: "${defaultWebserverModule}"
   }
   compile project(":h2o-security")
+
+  //To make H2OTestRunner available
+  testCompile project(":h2o-core").sourceSets.test.output
```
honzasterba: Agree this is very ugly, would rather move the required stuff into separate module

deil87:
Tests were being run actually. The issue is that xml reporting reporting is not appropriately configured for `test` task of mentioned modules.

PR #4086, opened 2019-11-20
PUBDEV-7060: Propagate offset_column to MOJO models

michalkurka:


h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
```diff
@@ -180,6 +180,9 @@ private void addCommonModelInfo() throws IOException {
     writekv("n_classes", model.nclasses());
     writekv("n_columns", model.columnNames().length);
     writekv("n_domains", n_categoricals);
+    if (model.offsetColumn() != null) {
```
Pscheidl: Having it for all makes more sense. Great.

h2o-genmodel/src/main/java/hex/genmodel/GenModel.java
```diff
@@ -112,6 +112,10 @@ public GenModel(String[] names, String[][] domains) {
     return _domains.length - 1;
   }
 
+  @Override public String getOffsetName() {
```
Pscheidl: getOffsetColName ? Doesn't matter really, the var is named `_offsetColumn`, not offset.

michalkurka: this is modeled to be consistent with `getResponseName`

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
michalkurka: thank you

PR #4089, opened 2019-11-21
PUBDEV-7072: add AUCPR column to AutoML leaderboard for binomial classification

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-7072

Leaderboard now looks like this:
```
> aml@leaderboard
                                        model_id       auc    logloss     aucpr
1                   GBM_1_AutoML_20191126_062347 0.9950717 0.11876769 0.9850958
2 XGBoost_grid__1_AutoML_20191126_062347_model_5 0.9949223 0.10305487 0.9480312
3                   GLM_1_AutoML_20191126_062347 0.9947730 0.10012978 0.9848061
4 XGBoost_grid__1_AutoML_20191126_062347_model_2 0.9947357 0.09782578 0.9849690
5 XGBoost_grid__1_AutoML_20191126_062347_model_1 0.9936529 0.10097153 0.9838275
6                   GBM_4_AutoML_20191126_062347 0.9935783 0.10115086 0.9840811
  mean_per_class_error      rmse        mse
1           0.03345281 0.1800188 0.03240677
2           0.02882318 0.1604014 0.02572860
3           0.03345281 0.1710249 0.02924950
4           0.03345281 0.1655949 0.02742167
5           0.02882318 0.1654495 0.02737353
6           0.02942055 0.1645574 0.02707914
```

Also, renamed renamed leaderboard column `mean_residual_deviance` to `deviance` for consistency.

### DIFF
--- a/h2o-core/src/main/java/hex/ModelMetricsRegression.java
+++ b/h2o-core/src/main/java/hex/ModelMetricsRegression.java
@@ -12,10 +12,21 @@ import water.util.ArrayUtils;
 import water.util.MathUtils;
 
 public class ModelMetricsRegression extends ModelMetricsSupervised {
-  /** For all algos except GLM this is mean residual deviance.  For GLM it's total residual deviance. */
-  public double residual_deviance() { return _mean_residual_deviance; }
-  public double mean_residual_deviance() { return _mean_residual_deviance; }
   public final double _mean_residual_deviance;
+  /**
+   * Alias for {@link #mean_residual_deviance()}.
+   *
+   * Required for consistency when retrieving metrics using reflection, as it is exposed for input as {@link ScoreKeeper.StoppingMetric#deviance}.
+   * @return {@link #mean_residual_deviance()}
+   */
+  @SuppressWarnings("unused")
+  public double deviance() { return _mean_residual_deviance; }
+  /**
+   * @return {@link #mean_residual_deviance()} for all algos except GLM, for which it means "total residual deviance".
+   **/
+  public double residual_deviance() { return _mean_residual_deviance; }
+  @SuppressWarnings("unused")
+  public double mean_residual_deviance() { return _mean_residual_deviance; }
   public final double _mean_absolute_error;
   public double mae() { return _mean_absolute_error; }
   public final double _root_mean_squared_log_error;
--- a/h2o-core/src/main/java/hex/ModelMetricsRegression.java
+++ b/h2o-core/src/main/java/hex/ModelMetricsRegression.java
@@ -14,14 +14,6 @@ import water.util.MathUtils;
 public class ModelMetricsRegression extends ModelMetricsSupervised {
   public final double _mean_residual_deviance;
   /**
-   * Alias for {@link #mean_residual_deviance()}.
-   *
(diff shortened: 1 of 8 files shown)


h2o-core/src/main/java/hex/ModelMetricsRegression.java
```diff
@@ -12,10 +12,21 @@
...
-  /** For all algos except GLM this is mean residual deviance.  For GLM it's total residual deviance. */
+  public final double _mean_residual_deviance;
+  /**
+   * Alias for {@link #mean_residual_deviance()}.
+   *
+   * Required for consistency when retrieving metrics using reflection, as it is exposed for input as {@link ScoreKeeper.StoppingMetric#deviance}.
+   * @return {@link #mean_residual_deviance()}
+   */
+  @SuppressWarnings("unused")
+  public double deviance() { return _mean_residual_deviance; }
```
seb-h2o: @wendycwong, @michalkurka any concern about this new `deviance` alias?

ledell:
@seb-h2o I think the reason that the term `mean_residual_deviance` (instead of `deviance`) was used in the leaderboard is because that's what the accessor functions are called in [R](http://docs.h2o.ai/h2o/latest-stable/h2o-r/docs/reference/h2o.mean_residual_deviance.html) and [Python](http://docs.h2o.ai/h2o/latest-stable/h2o-py/docs/metrics.html?highlight=deviance#h2o.model.metrics_base.MetricsBase.mean_residual_deviance) and it's more explicit.  In R, the metric is stored at `m@model$training_metrics@metrics$mean_residual_deviance`

Almost everywhere except for the `early_stopping` value, we use the full term, "Mean Residual Deviance" since we also have another term, null deviance, which can be confusing.

```
> h2o.mean_residual_deviance(m)
[1] 0.1302183

H2ORegressionMetrics: deeplearning
** Reported on training data. **
** Metrics reported on full training frame **

MSE:  0.1302183
RMSE:  0.3608577
MAE:  0.2850873
RMSLE:  0.05111677
Mean Residual Deviance :  0.1302183
```

Since the term `"deviance"` is used for the `early_stopping` parameter everywhere in H2O, it's definitely an internal inconsistency that won't be entirely resolved even if  fix it here.  I think to be more in line with the metric name as it's used elsewhere in H2O, we may just want to tolerate the internal inconsistency and leave it as is.  

As an alternative, we could propose that we add an alias of `"mean_residual_deviance"` for the `early_stopping` parameter in H2O in general.  What do you think about this approach?

ledell:
@seb-h2o I think the placement of the "aucpr" column is good.  I updated the description with a copy of the leaderboard for reference.

seb-h2o:
@ledell I'll revert changes for `deviance` then and keep `mean_residual_deviance` in leaderboard (`deviance` had one advantage for display, it kept the column narrower).
Regarding the alias for `early_stopping`, as much as I don't mind method aliases (it's relatively common), I don't like it much for enums: those are more annoying to maintain: each time the enum is used somewhere, we have to ensure that both values are handled in the same way.

> H2O, it's definitely an internal inconsistency that won't be entirely resolved even if fix it here. I think to be more in line with the metric name as it's used elsewhere in H2O, we may just want to tolerate the internal inconsistency and leave it as is.

I can live with that.

PR #4090, opened 2019-11-21
[PUBDEV-7082] Expose Location of Chunks in REST API

mn-mikke:


h2o-core/src/main/java/water/api/schemas3/FrameChunksV3.java
```diff
@@ -0,0 +1,45 @@
...
+    public static class FrameChunkV3 extends SchemaV3<Iced, FrameChunkV3> {
+
+        @API(help="An identifier unique in scope of a given frame", direction=API.Direction.OUTPUT)
+        public int chunk_id;
+
+        @API(help="Number of rows represented byt the chunk", direction=API.Direction.OUTPUT)
+        public int row_count;
+        
+        @API(help="Description of H2O node where the chunk is located", direction=API.Direction.OUTPUT)
+        public CloudV3.NodeV3 location;
```
michalkurka: NodeV3 has quite a bit of information in it, this info will be replicated many times as many chunks will reside on the same node. I would use node index instead. WDYT?

h2o-core/src/test/java/water/api/APIThrPriorTest.java
```diff
@@ -97,7 +97,7 @@
       vec.sigma();
       Assert.assertNotNull(DKV.get(rskey)); // Rollups on my zeros not computed yet
       serve("/3/Frames/iris.hex", null,200); // Rollups already done at parse, but gets ChunkSummary
-
+      serve("/3/FrameChunks/iris.hex", null,200);
```
michalkurka: doesn't really belong here imho

h2o-core/src/main/java/water/api/schemas3/FrameChunksV3.java
mn-mikke: I like your idea, but I can't find anything like this on `NodeV3` is the node index equal to the position in `CloudeV3.nodes`?

h2o-core/src/test/java/water/api/APIThrPriorTest.java
mn-mikke: ok, will remove it.

h2o-core/src/main/java/water/api/schemas3/FrameChunksV3.java
michalkurka: you can call `key.home().index()`

  // index of this node in the current cloud... can change at the next cloud.
  public int index() { return H2O.CLOUD.nidx(this); }

mn-mikke: I mentioned the position in `CloudV3.nodes` in order to be able to correctly map the node index to the correct `NodeV3` instance within the array. [It seems that Sparkling Water already relies on the equality of the node order within `CloudV3.nodes` and the node index.](https://github.com/h2oai/sparkling-water/blob/features/dynamic-allocation/core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala#L102)

michalkurka: Yes, the order is given once the cloud is formed. You can rely on that.

mn-mikke: Cool, thanks!

mn-mikke:
@jakubhava @michalkurka The call of this new REST API was causing failure of the whole H2O node. I made same changes to avoid that. Please can you have a second look?

h2o-core/src/main/java/water/api/schemas3/FrameChunksV3.java
```diff
@@ -5,9 +5,9 @@
 import water.api.API;
 import water.fvec.Vec;
 
-public class FrameChunksV3 extends SchemaV3<Frame, FrameChunksV3> {
+public class FrameChunksV3 extends SchemaV3<Iced, FrameChunksV3> {
```
michalkurka: I think this is a bit unusual but I don't see any issues with it

michalkurka:
@mn-mikke I don't see any issue - I think it is good to merge

h2o-core/src/main/java/water/api/schemas3/FrameChunksV3.java
mn-mikke: There is a problem that `Frame` doesn't have a parameterless constructor and [this line](https://github.com/h2oai/h2o-3/blob/b9b0954b45c956d6d3d08537ada3543c85efa557/h2o-core/src/main/java/water/api/Schema.java#L167) will cause H2O node failure. I guess that's the reason why the `Frames` class was introduced.

Regarding the usage of `Iced`, I've gotten inspired in [`FindV3`](https://github.com/h2oai/h2o-3/blob/43f8ab952a69a8bc9484bd0ffac909b6e3e820ca/h2o-core/src/main/java/water/api/schemas3/FindV3.java#L7).

Pscheidl:
The tests are OK. Failures completely unrelated. :heavy_check_mark:

mn-mikke:
Thank guys for your reviews!

PR #4091, opened 2019-11-21
PUBDEV-6572: AutoML set default max_runtime_secs=1h iff max_models is not set by user

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-6572

- Grouped AutoML validation/defaults in validateBuildSpec() method, including time/model constraints defaults.
- Exposed result buildSpec to clients for testing+debugging purpose.
- Changed and tested max_runtime_secs default on client side so that it's limited to 1h iff user didn't set max_models.


h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
```diff
@@ -142,7 +142,7 @@
             level = API.Level.secondary)
     public int max_models;
 
-    @API(help = "Maximum time to spend building all models (optional).",
+    @API(help = "Maximum time to spend building all models. Defaults to 1h if max_models is not provided. (optional).",
```
deil87: @seb-h2o should we say `optional` if we have already said `Defaults to...`?

seb-h2o: It doesn't hurt I guess.
I'm not even sure we're always consistent with those "optional" info.

h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
```diff
@@ -298,10 +298,55 @@ def test_frames_cannot_be_passed_as_key():
...
+    assert max_models == 0
+
+
+def test_max_runtime_secs_can_be_set_in_combination_with_max_models():
+    ds = import_dataset()
+    aml = H2OAutoML(project_name="py_all_stopping_constraints", seed=1, max_models=1, max_runtime_secs=12)
+    aml.train(y=ds['target'], training_frame=ds['train'])
+    max_runtime = aml._build_resp['build_control']['stopping_criteria']['max_runtime_secs']
+    max_models = aml._build_resp['build_control']['stopping_criteria']['max_models']
+    assert max_runtime == 12
```
deil87: @seb-h2o can we test that actually max_runtime_secs is being ignored?

h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
```diff
@@ -298,10 +298,55 @@ def test_frames_cannot_be_passed_as_key():
...
+    assert max_models == 1
+
+
+def test_default_max_runtime_if_no_max_models_provided():
+    ds = import_dataset()
+    aml = H2OAutoML(project_name="py_no_stopping_constraints", seed=1, verbosity='Info')
+    with pu.Timeout(5, on_timeout=lambda: aml._job.cancel()):
+        try:
+            aml.train(y=ds['target'], training_frame=ds['train'])
```
deil87: @seb-h2o could not we just do this? 
```
aml.train(y=ds['target'], training_frame=ds['train'])
aml._job.cancel()
```
Or it is not enough time to get response?

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -47,9 +47,30 @@
...
 
 
+class Timeout:
+
+    def __init__(self, timeout_secs, on_timeout=None):
+        enabled = timeout_secs is not None and timeout_secs >= 0
+        self.timer = threading.Timer(timeout_secs, on_timeout) if enabled else None
+
+    @property
+    def active(self):
```
deil87: @seb-h2o is this method being used?

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -164,76 +166,142 @@ public AutoML() {
...
     }
   }
 
-  private void initLeaderboard(AutoMLBuildSpec buildSpec) {
-    String sort_metric = buildSpec.input_spec.sort_metric;
-    sort_metric = sort_metric == null || StoppingMetric.AUTO.name().equalsIgnoreCase(sort_metric) ? null : sort_metric.toLowerCase();
-    _leaderboard = Leaderboard.getOrMake(_key.toString(), _eventLog, _leaderboardFrame, sort_metric);
-  }
+  private void validateBuildSpec(AutoMLBuildSpec buildSpec) {
+    // Validate all buildSpec parameters here and provide reasonable defaults or parameter cleaning if necessary.
```
deil87: @seb-h2o should n't we add comments about whole method in a Javadoc style outside of the method's scope?

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -164,76 +166,142 @@ public AutoML() {
...
+    // When assigning default or modifying the original value, use `eventLog` to inform the user about the changes.
+    // Please no additional logic!
```
deil87: @seb-h2o "Please no additional logic!" - this IMO is redundant. It might be confusing if we will need to add extra validation for buildSpec

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -164,76 +166,142 @@ public AutoML() {
...
 
-  private void handleReproducibilityParameters(AutoMLBuildSpec buildSpec) {
-    eventLog().info(Stage.Workflow, "Build control seed: " + buildSpec.build_control.stopping_criteria.seed() +
-            (buildSpec.build_control.stopping_criteria.seed() == -1 ? " (random)" : ""));
+    // Can possibly use this validation entry point to provide parameters validation for Flow later.
+    // Can also reuse ModelBuilder.ValidationMessage for this.
+
+    validateInput(buildSpec.input_spec);
```
deil87: @seb-h2o when reading over the file I found it slightly confusing that we have `validateInput` and then `prepareData`. So we first validate and then prepare/change data which looks unusual to me.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -164,76 +166,142 @@ public AutoML() {
...
-      buildSpec.build_control.stopping_criteria.set_default_stopping_tolerance_for_frame(_trainingFrame);
-      eventLog().info(Stage.Workflow, "Setting stopping tolerance adaptively based on the training frame: " +
-              buildSpec.build_control.stopping_criteria.stopping_tolerance());
-    } else {
-      eventLog().info(Stage.Workflow, "Stopping tolerance set by the user: " + buildSpec.build_control.stopping_criteria.stopping_tolerance());
-      double default_tolerance = AutoMLBuildSpec.AutoMLStoppingCriteria.default_stopping_tolerance_for_frame(_trainingFrame);
-      if (buildSpec.build_control.stopping_criteria.stopping_tolerance() < 0.7 * default_tolerance){
-        eventLog().warn(Stage.Workflow, "Stopping tolerance set by the user is < 70% of the recommended default of " + default_tolerance + ", so models may take a long time to converge or may not converge at all.");
+  private void validateInput(AutoMLInput input) {
+    if (null == DKV.getGet(input.training_frame))
```
deil87: @seb-h2o minor. this is not a Java style to put null upfront I believe

h2o-py/tests/pyunit_utils/utilsPY.py
seb-h2o: no, you're right, I should remove this: I took this from my utilities in the benchmark app :)

h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
seb-h2o: why ignored? it's not ignored, it's the first that wins.
I thought we already had a test though checking this scenario in `pyunit_automl_args.py` when both are set, but apparently not: will add.

h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
seb-h2o: you can be certain that I tried... and it obviously doesn't work: `aml.train` is blocking. And I decided that Jenkins had a better use of its time than waiting for 1h.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: yeah, bad wording, will fix.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: I usually don't like Javadoc on private methods, but you're probably right.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -164,76 +166,142 @@ public AutoML() {
...
+      throw new H2OIllegalArgumentException("No training data has been specified, either as a path or a key, or it is not available anymore.");
+
+    final Frame trainingFrame = DKV.getGet(input.training_frame);
+    final Frame validationFrame = DKV.getGet(input.validation_frame);
+    final Frame blendingFrame = DKV.getGet(input.blending_frame);
+    final Frame leaderboardFrame = DKV.getGet(input.leaderboard_frame);
+
+    Map<String, Frame> compatibleFrames = new LinkedHashMap<String, Frame>(){{
```
deil87: @seb-h2o do we need Linked implementation? do we rely on order from iterator?

h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
deil87: @seb-h2o 
> why ignored? it's not ignored, it's the first that wins.

you are right. forgot for a second

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: yeah, validation is only about checking that all buildSpec parameters are correct, consistent with each other... its only allowed side effect is to modify some of those params to set some defaults dynamically.
Before validation is completed, you don't know yet if you'll start the training.

As a comparison, you start preparing data when the training part is engaged.

The distinction is clear from a user perspective, like ModelBuilder does. We could make a request to AutoML just to check the parameters, and display messages, highlight wrong parameter fields... this should be fast, and you don't want to start preparing data in this request.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: I've always seen both. 
I must have copied it because I usually don't use it this way, but I can change if it hurts your eyes.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
deil87: @seb-h2o I understand. It is only the fact that within validation of buildSpec we actually do validate Input ( which is data as well) that confused me

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -164,76 +166,142 @@ public AutoML() {
...
+      put("training", trainingFrame);
+      put("validation", validationFrame);
+      put("blending", blendingFrame);
+      put("leaderboard", leaderboardFrame);
+    }};
```
deil87: @seb-h2o this double bracket syntax looks convenient but FYI I have heard that there are some concern about it:
```
Note that, we must try to avoid this initialization technique as it creates an anonymous extra class at every usage, and also holds hidden references to the enclosing object and might cause memory leak issues.
``` 
https://www.baeldung.com/java-initialize-hashmap

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: yeah, somehow.
As it is throwing on first error, I wanted to be sure it throws on the most "important" frame. I think it's a bit friendlier to get a message about the trainingFrame if response column is missing everywhere, than on the blendingFrame...

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: the instance scope is only local to the validation method, so it's safe: no leak concerns.

seb-h2o: It's just the same concerns as for any anonymous class btw.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -164,76 +166,142 @@ public AutoML() {
...
+    }
+
+    if (input.ignored_columns != null) {
+      List<String> ignoredColumns = new ArrayList<>(Arrays.asList(input.ignored_columns));
+      Map<String, String> doNotIgnore = new LinkedHashMap<String, String>(){{
+        put("response_column", input.response_column);
+        put("fold_column", input.fold_column);
+        put("weights_column", input.weights_column);
+      }};
+      for (Map.Entry<String, String> entry: doNotIgnore.entrySet()) {
```
deil87: @seb-h2o when I see `for` with just `if` inside I want to use streams with filters:
```
doNotIgnore.entrySet().stream()
              .filter(entry -> entry.getValue() != null && ignoredColumns.contains(entry.getValue()))
              .forEach(entry -> {
                eventLog().info(Stage.DataImport, "Removing "+entry.getKey()+" '"+entry.getValue()+"' from list of ignored columns.");
                ignoredColumns.remove(entry.getValue())
              });
```

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
ledell:
```suggestion
    @API(help = "Maximum time to spend building all models. Defaults to 1 hour if max_models is not provided. (optional).",
```

h2o-docs/src/product/automl.rst
```diff
@@ -30,7 +30,7 @@ Required Stopping Parameters
 
 One of the following stopping strategies (time or number-of-model based) must be specified.  When both options are set, then the AutoML run will stop as soon as it hits one of either of these limits. 
 
-- `max_runtime_secs <data-science/algo-params/max_runtime_secs.html>`__: This argument controls how long the AutoML will run at the most, before training the final Stacked Ensemble models. Defaults to 3600 seconds (1 hour).
+- `max_runtime_secs <data-science/algo-params/max_runtime_secs.html>`__: This argument controls how long the AutoML will run at the most, before training the final Stacked Ensemble models. Defaults to 3600 seconds (1 hour) if ``max_models`` is not provided.
```
ledell:
```suggestion
- `max_runtime_secs <data-science/algo-params/max_runtime_secs.html>`__: This argument specifies the maximum time that the AutoML process will run for, prior to training the final Stacked Ensemble models. Defaults to 3600 seconds (1 hour) if ``max_models`` is not provided.
```

h2o-py/h2o/automl/autoh2o.py
```diff
@@ -83,9 +83,9 @@ def __init__(self,
           factors will be automatically computed to obtain class balance during training. Requires ``balance_classes``.
         :param float max_after_balance_size: Maximum relative size of the training data after balancing class counts (can be less than 1.0).
           Requires ``balance_classes``. Defaults to ``5.0``.
-        :param int max_runtime_secs: This argument controls how long the AutoML run will execute. Defaults to ``3600`` seconds (1 hour).
+        :param int max_runtime_secs: This argument controls how long the AutoML run will execute. If ``max_models`` is not set, then defaults to ``3600`` seconds (1 hour).
```
```suggestion
        :param int max_runtime_secs: This argument specifies the maximum time that the AutoML process will run for, prior to training the final Stacked Ensemble models. If ``max_models`` is not set, then defaults to ``3600`` seconds (1 hour).
```

h2o-r/h2o-package/R/automl.R
```diff
@@ -28,9 +28,9 @@
 #'        be automatically computed to obtain class balance during training. Requires balance_classes.
 #' @param max_after_balance_size Maximum relative size of the training data after balancing class counts (can be less than 1.0). Requires
 #'        balance_classes. Defaults to 5.0.
-#' @param max_runtime_secs Maximum allowed runtime in seconds for the entire model training process. Use 0 to disable. Defaults to 3600 secs (1 hour).
+#' @param max_runtime_secs Maximum allowed runtime in seconds for the entire model training process. Use 0 to disable. If `max_models` is not set, then defaults to 3600 secs (1 hour).
```
```suggestion
#' @param max_runtime_secs This argument specifies the maximum time that the AutoML process will run for, prior to training the final Stacked Ensemble models. If `max_models` is not set, then it defaults to 3600 seconds (1 hour).
```

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: yeah, difference is not that big tbh.
Also, I'm not a complete fan for `forEach` in streams. I like to know when there are side effects and where there are none: so I use streams with pure functions and `for` for side effects.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
ledell: The "(optional)" thing is an old R convention.  We are not consistent about it and I think having it there only part of the time is more confusing than not having it there at all.  At some point, we should just go through and remove it everywhere.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
deil87: @seb-h2o in functional programming `forEach` signifies side effects. `map` is for pure things.

seb-h2o: I know that. And that's why it doesn't bring much value in a language like Java.

PR #4092, opened 2019-11-21
PUBDEV-7086: fixed two tests with leaked keys due to missing extension from TestUtil class

deil87:


Pscheidl:
Thanks for fixing it for me.

PR #4097, opened 2019-11-22
[PUBDEV-7095] Allocate more time for clouding to fix Cloud 1 under 4 issue

deil87:
For details see https://0xdata.atlassian.net/browse/PUBDEV-7095

### DIFF
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -1850,11 +1850,14 @@ final public class H2O {
 
   public static void waitForCloudSize(int x, long ms) {
     long start = System.currentTimeMillis();
+    Log.info("Started clouding. Target number of nodes: " + x);
     while (System.currentTimeMillis() - start < ms) {
       if (CLOUD.size() >= x && Paxos._commonKnowledge)
         break;
       try { Thread.sleep(100); } catch (InterruptedException ignore) {}
     }
+    Log.info("Clouding process has timedout ( last retry achieved " + CLOUD.size() + " out of " + x + " nodes cloud ). " 
+            + "Consider to increase `DEFAULT_TIME_FOR_CLOUDING`. Continuing....");
     if (CLOUD.size() < x)
       throw new RuntimeException("Cloud size " + CLOUD.size() + " under " + x);
   }
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -1850,16 +1850,14 @@ final public class H2O {
 
   public static void waitForCloudSize(int x, long ms) {
     long start = System.currentTimeMillis();
-    Log.info("Started clouding. Target number of nodes: " + x);
+    Log.info("Waiting for clouding to finish. Target number of nodes: " + x);
     while (System.currentTimeMillis() - start < ms) {
       if (CLOUD.size() >= x && Paxos._commonKnowledge)
         break;
       try { Thread.sleep(100); } catch (InterruptedException ignore) {}
     }
-    Log.info("Clouding process has timedout ( last retry achieved " + CLOUD.size() + " out of " + x + " nodes cloud ). " 
-            + "Consider to increase `DEFAULT_TIME_FOR_CLOUDING`. Continuing....");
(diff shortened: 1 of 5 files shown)


deil87:
Tested with at least 4 CI  runs - no `Cloud 1 under 4` errors.

michalkurka:
@deil87 thanks for the change

> Java 8 JUnit / ai.h2o.targetencoding.TargetEncoderModelTest.testTargetEncoderModel

> On the screenshot it could be seen that test starts after about 50 seconds after clouding was started and cloud reaches 4 nodes only after tests have already been started. 


The reason why the test appears started are different for both provided examples:

- TargetEncoderModelTest - stall_until_cloud_size should be declared on a @BeforeClass setup method - not on a @Before method

- BoosterParmsTest - doesn't require cloud at all, it is a regular junit test - it can run at any time

deil87:
@michalkurka 
> BoosterParmsTest - doesn't require cloud at all, it is a regular junit test - it can run at any time

I noticed this as well but I was trying to understand why we get this error for that kind of tests when we execute `testMultiNode` task. See error for `BoosterParmsTest`
http://mr-0xc1:8080/view/H2O-3/job/h2o-3-pipeline/job/rel-yau/208/testReport/junit/hex.tree.xgboost/BoosterParmsTest/Java_8_XGBoost_Multinode_JUnit___testGetLocalized/

> TargetEncoderModelTest - stall_until_cloud_size should be declared on a @BeforeClass setup method - not on a @before method

Still i think that even if placed in BeforeClass we will have a chance to get a time out. Increasing timeout helped to got rid of the errors anyway. Will try to improve logic.

deil87:
@honzasterba I decided to return timeout to the previous value and fixed Before annotation as @michalkurka suggested. The strange thing still is that another test that don't have any clouding logic had `Cloud 1 under 4` error output as well. I believe it is an issue with our reporting logic ( test took output from another failed test). 
Anyway test keeps failing  http://mr-0xc1:8080/view/h2o-3-monitor/job/h2o-3-pipeline/job/rel-yau/209/ - I would merge this one but waiting for @michalkurka as he asked not to merge anything to `rel-yau` without asking him.

deil87:
@honzasterba actually failure message has reference to another test XGBoostTest
```
java.lang.RuntimeException: Cloud size 1 under 4
	at hex.tree.xgboost.XGBoostTest.null()
```

Looks like we are experiencing side effect from  XGBoostTest. Does anything look unusual in the setup of the test(you might be more familiar with XGBoost related tests)?

honzasterba:
this is failing in the @BeforeClass with stall_till_cloudsize so nothing special

deil87:
@honzasterba what I can see from logs highlights two things:
1) tests that do not have any `stall_till_cloud_size` logic will fail with errors from other tests classes. `BoosterParmsTest.testGetLocalized` failed with error from `XGBoostPredictImplComparisonTest`. See http://mr-0xc1:8080/view/h2o-3-monitor/job/h2o-3-pipeline/job/rel-yau/208/testReport/junit/hex.tree.xgboost/BoosterParmsTest/Java_8_XGBoost_Multinode_JUnit___testGetLocalized/
2) Even with BeforeClass annotation we are failing to prepare the cloud. See log from 1) . XGBoostPredictImplComparisonTest has timed out. So it I think justifies increase of the default timeout.

michalkurka:
@deil87 I would also suggest using less number of clouds in the TE tests - I think it just one cloud is enough to run all the tests

currently, there are 5 clouds:

    CLUSTER_BASEPORT_1=44000
    CLUSTER_BASEPORT_2=45000
    CLUSTER_BASEPORT_3=46000
    CLUSTER_BASEPORT_4=47000
    CLUSTER_BASEPORT_5=48000
    ...

increasing the timeout is fine as well

deil87:
@michalkurka can we merge this now? Tests look fine.

michalkurka:
@deil87 this would be okay to merge to master but not `rel-yau` - merging to release branch is now not allowed

deil87:
@michalkurka 
> this would be okay to merge to master but not rel-yau - merging to release branch is now not allowed

is it allowed now, once we have released 3.26.0.11?

PR #4098, opened 2019-11-22
PUBDEV-7093: Treat Parquet boolean values as True/False categorical values

michalkurka:


h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/ChunkConverter.java
```diff
@@ -134,6 +134,8 @@ private PrimitiveConverter newConverter(int colIdx, byte vecType, PrimitiveType
       case Vec.T_TIME:
         if (OriginalType.TIMESTAMP_MILLIS.equals(parquetType.getOriginalType()) || parquetType.getPrimitiveTypeName().equals(PrimitiveType.PrimitiveTypeName.INT96)) {
           return new TimestampConverter(colIdx, _writer);
+        } else if (parquetType.getPrimitiveTypeName().equals(PrimitiveType.PrimitiveTypeName.BOOLEAN)) {
```
Pscheidl: In case of `Boolean`, the vector type is `T_CAT` only. Makes more sense to seal the contract in this switch and check for boolean only in `case Vec.T_CAT`:

```patch
Index: h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/ChunkConverter.java
IDEA additional info:
<+>UTF-8
--- h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/ChunkConverter.java	(revision c5512af07256509f367fb99d36a12da14dbdf0a8)
+++ h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/ChunkConverter.java	(date 1574517876127)
@@ -129,14 +129,15 @@
     switch (vecType) {
       case Vec.T_BAD:
       case Vec.T_CAT:
+        if (parquetType.getPrimitiveTypeName().equals(PrimitiveType.PrimitiveTypeName.BOOLEAN)) {
+          return new BooleanConverter(_writer, colIdx);
+        }
       case Vec.T_STR:
       case Vec.T_UUID:
       case Vec.T_TIME:
         if (OriginalType.TIMESTAMP_MILLIS.equals(parquetType.getOriginalType()) || parquetType.getPrimitiveTypeName().equals(PrimitiveType.PrimitiveTypeName.INT96)) {
           return new TimestampConverter(colIdx, _writer);
-        } else if (parquetType.getPrimitiveTypeName().equals(PrimitiveType.PrimitiveTypeName.BOOLEAN)) {
-          return new BooleanConverter(_writer, colIdx);
-        } else {
+        }  else {
           boolean dictSupport = parquetType.getOriginalType() == OriginalType.UTF8 || parquetType.getOriginalType() == OriginalType.ENUM;
           return new StringConverter(_writer, colIdx, dictSupport);
         }
```

Pscheidl: It's not a necessity. I had to ask myself why do we even consider converting Vec type of TIME and UUID into Boolean. Makes sense for CAT any maybe STR to me.

michalkurka: yes, doesn't make sense for Time and UUID, however, it does make sense for String - this is up to the user decision

Pscheidl: I'll leave this one up to you. Current status is fine to me.

Pscheidl: If it doesn't make sense for Time and UUID, then why not exclude them, but again, the user should know..as you said. The PR is approved.

michalkurka: thanks - fixed

PR #4101, opened 2019-11-24
[PUBDEV-7096] Add option which can stop cluster if nothing has touched the rest api for the specified timeout

jakubhava:
Initial implementation. I have tried to to find a single place which is called for all successful rest API calls, but couldn't find it, that is why I made this check only for 3/Cloud.

If you have better ideas, please let me know. I hope that at some point with complete switch to REST api on SW side, we can remove the watchdog client etc(maybe client as well), but that will take several major iterations still, so these 2 need to coexist now

### DIFF
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -753,7 +753,7 @@ final public class H2O {
     }
     
     if (ARGS.rest_api_ping_timeout < 0) {
-      parseFailed("rest_api_ping_timeout needs to be 0 or higher, was (" + ARGS.rest_api_ping_timeout + ")");
+      parseFailed(String.format("rest_api_ping_timeout needs to be 0 or higher, was (%d)", ARGS.rest_api_ping_timeout));
     }
 
     // Validate extension arguments
@@ -1667,8 +1667,8 @@ final public class H2O {
     }
 
     if (H2O.ARGS.rest_api_ping_timeout > 0) {
-      Log.info("Registering REST API Check Thread. If 3/Ping endpoint is not accessed during " + H2O.ARGS.rest_api_ping_timeout + "" +
-          " ms, the cluster will be terminated.");
+      Log.info(String.format("Registering REST API Check Thread. If 3/Ping endpoint is not" +
+          " accessed during %d ms, the cluster will be terminated.", H2O.ARGS.rest_api_ping_timeout));
       new RestApiPingCheckThread().start();
     }
     // Create the starter Cloud with 1 member
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -350,6 +350,11 @@ final public class H2O {
     /** -allow_clients, -allow_clients=true; Enable clients to connect to this H2O node - disabled by default */
     public boolean allow_clients = false;
 
+    /** If this timeout is set to non 0 value, stop the cluster if there hasn't been any rest api request to leader
+     * node after the given timeout. Unit is milliseconds.
+     */
+    public int rest_api_ping_timeout = 0;
+    
     /** specifies a file to write when the node is up */
     public String notify_local;
 
@@ -531,6 +536,9 @@ final public class H2O {
       }
       else if (s.matches("allow_clients")) {
         trgt.allow_clients = true;
+      } else if (s.matches("rest_api_ping_timeout")) {
+        i = s.incrementAndCheck(i, args);
+        trgt.rest_api_ping_timeout = s.parseInt(args[i]);
       }
       else if (s.matches("notify_local")) {
         i = s.incrementAndCheck(i, args);
@@ -742,6 +750,10 @@ final public class H2O {
       }
       if (ARGS.session_timeout <= 0)
         parseFailed("Invalid session timeout specification (" + ARGS.session_timeout + ")");
+    }
+    
+    if (ARGS.rest_api_ping_timeout < 0) {
+      parseFailed("rest_api_ping_timeout needs to be 0 or higher, was (" + ARGS.rest_api_ping_timeout + ")");
     }
 
     // Validate extension arguments
@@ -1654,6 +1666,11 @@ final public class H2O {
               "  2. Point your browser to " + NetworkInit.h2oHttpView.getScheme() + "://localhost:55555");
     }
 
+    if (H2O.ARGS.rest_api_ping_timeout > 0) {
+      Log.info("Registering REST API Check Thread. If 3/Ping endpoint is not accessed during " + H2O.ARGS.rest_api_ping_timeout + "" +
+          " ms, the cluster will be terminated.");
+      new RestApiPingCheckThread().start();
+    }
     // Create the starter Cloud with 1 member
     SELF._heartbeat._jar_md5 = JarHash.JARHASH;
     SELF._heartbeat._client = ARGS.client;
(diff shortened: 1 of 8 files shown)


jakubhava:
I will change the implementation  and introduce new endpoint for this

jakubhava:
I'm designing the new endpoint after this structure in Sparkling Water -> https://github.com/h2oai/sparkling-water/blob/features/dynamic-allocation/core/src/main/scala/org/apache/spark/h2o/ui/SparklingWaterHeartbeatEvent.scala

Sparkling Water needs to periodically ask for the following information:
-> cloud uptime
-> cloud healthy
-> free memory on nodes

By putting this info to the heartbeat endpoint we limit the number of traffic as currently, for only this 3 information we are pinging the `Cloud` endpoint

h2o-core/src/main/java/water/H2ONode.java
```diff
@@ -36,6 +36,7 @@
   transient short _unique_idx; // Dense integer index, skipping 0.  NOT cloud-wide unique.
   transient boolean _announcedLostContact;  // True if heartbeat published a no-contact msg
   transient public long _last_heard_from; // Time in msec since we last heard from this Node
+  transient public long _last_heard_from_rest; // Time in msec since somebody accessed the '3/Heartbeat' endpoint on this node
```
michalkurka: IMHO this is misleading because it only applies to SELF, it shouldn't be defined on H2ONode in my opinion.

h2o-core/src/main/java/water/api/HeartbeatHandler.java
```diff
@@ -0,0 +1,27 @@
...
+class HeartbeatHandler extends Handler {
+  
+  @SuppressWarnings("unused") // called through reflection by RequestServer
+  public HeartbeatV3 status(int version, HeartbeatV3 heartbeat) {
+
+    heartbeat.cloud_uptime_millis = System.currentTimeMillis() - H2O.START_TIME_MILLIS.get();
+    heartbeat.cloud_healthy = true;
+    H2ONode[] members = H2O.CLOUD.members();
+
+    H2ONode leader = H2O.CLOUD.leader();
```
michalkurka: unused?

h2o-core/src/main/java/water/api/HeartbeatHandler.java
```diff
@@ -0,0 +1,27 @@
+package water.api;
+
+import water.H2O;
+import water.H2ONode;
+import water.api.schemas3.HeartbeatV3;
+
+class HeartbeatHandler extends Handler {
```
michalkurka: is this all output info needed?

maybe we could simplify to as simple PingHandler :)

also, it would be great if this was actually implemented on SW side as an extension and the SW-specific jars were added to the runtime dynamically - IMHO that would help the codebase not just in this case but for other SW-specific extensions

h2o-core/src/main/java/water/api/HeartbeatHandler.java
jakubhava: yup, thanks

h2o-core/src/main/java/water/api/HeartbeatHandler.java
jakubhava: Yup, for sparkling water ping, we just need this info. We access also other endpoints, for for the ping this is sufficient. 

I can rename it :)

Regarding the extension -> I really like to idea! What about finalizing the major release with the current approach and then move SW specific endpoints toward sparkling water code-base with the run-time registration? We can create just new sub-module in SW where we define the extension points

h2o-core/src/main/java/water/H2ONode.java
jakubhava: Good point, will figure out better place

jakubhava:
Tried to apply the suggestions, please have a look @michalkurka

h2o-core/src/main/java/water/H2O.java
```diff
@@ -743,6 +751,10 @@ private static void validateArguments() {
       if (ARGS.session_timeout <= 0)
         parseFailed("Invalid session timeout specification (" + ARGS.session_timeout + ")");
     }
+    
+    if (ARGS.rest_api_ping_timeout < 0) {
+      parseFailed("rest_api_ping_timeout needs to be 0 or higher, was (" + ARGS.rest_api_ping_timeout + ")");
```
mn-mikke: nit: WDYT about using `string.format`

h2o-core/src/main/java/water/RestApiPingCheckThread.java
```diff
@@ -0,0 +1,33 @@
...
+public class RestApiPingCheckThread extends Thread {
+  
+  public RestApiPingCheckThread() {
+    this.setDaemon(true);
+  }
+
+  @Override
+  public void run() {
+    while (!Thread.currentThread().isInterrupted()) {
+      if (H2O.CLOUD._memary.length != 0 && H2O.SELF == H2O.CLOUD.leader()) {
```
mn-mikke: If the H2O node is not a loader, can't the thread just regularly finish?

h2o-core/src/main/java/water/H2O.java
```diff
@@ -1657,6 +1669,11 @@ private static void startLocalNode() {
               "  2. Point your browser to " + NetworkInit.h2oHttpView.getScheme() + "://localhost:55555");
     }
 
+    if (H2O.ARGS.rest_api_ping_timeout > 0) {
+      Log.info("Registering REST API Check Thread. If 3/Ping endpoint is not accessed during " + H2O.ARGS.rest_api_ping_timeout + "" +
```
mn-mikke: nit: WDYT about using `string.format`

h2o-core/src/main/java/water/RestApiPingCheckThread.java
jakubhava: @michalkurka can leader node change during H2O run-time? I think it can, but I guess that after cloud is locked the leader will remain same.

If that is the case, we can close the thread after cloud is locked in case we are not leader

h2o-core/src/main/java/water/H2O.java
jakubhava: 👍

h2o-core/src/main/java/water/H2O.java
jakubhava: 👍

h2o-core/src/main/java/water/RestApiPingCheckThread.java
michalkurka: the leader doesn't change after the cloud is locked

jakubhava: perfect!!

h2o-core/src/main/java/water/api/schemas3/PingV3.java
```diff
@@ -0,0 +1,35 @@
...
+  public static class NodeMemoryInfoV3 extends SchemaV3<Iced, NodeMemoryInfoV3> {
+    public NodeMemoryInfoV3() {
+    }
+
+    public NodeMemoryInfoV3(String ipPort, long freeMemory) {
+      ip_port = ipPort;
+      free_mem = freeMemory;
+    }
+
+
```
mn-mikke: nit: extra line

jakubhava: 👍

jakubhava:
Thank you for the prompt reviews!

PR #4102, opened 2019-11-25
PUBDEV-6796 - Spearman's correlation coeffiecient (Pearson part)

Pscheidl:
# PUBDEV-6796 ( Pearson correlation part)
- [JIRA PUBDEV-6796](https://0xdata.atlassian.net/browse/PUBDEV-6796)
- [H2O Support Portal Ticket 95038](https://support.h2o.ai/a/tickets/95038)

## Description

Goal of the original JIRA is to calculate a **Spearman's** correlation coefficient (SCC) of two numerical `Vec`s. I've decided to warm myself up and provide the less complex **Pearson's** correlation coefficient (PCC).

**Pearson's correlation coefficient** is calculated as follows:

pearson

where s<sub>x</sub> & s<sub>y</sub> are standard deviations of column `x` or `y` respectively.

### Role of mean values
Both SCC and PCC require mean values of the two colums/vecs `x` and `y`. However, the dataset might contain `NaN`s. The computation alone `drops` lines with `NaN`s. If there is a line with a `NaN` in at least one of the two columns PCC value is calculated for, that line is not included in the calculation.

This makes the `Vec.mean()` value (Mean pre-calculated from rollup stats) unusable, as such mean is calculated per each column/vector independetly and it might contain lines avoided by the PCC calculation algorithm. Therefore, `VecUtils.calculateMeansIgnoringNas` method has been introduced. Such method calculates means for `n > 0` numerical vecs, skipping over lines where any of the Vec has `NaN` in it. For PCC and SCC, `n = 2`, always. That's self understood, but the method was written in a more general way.

## PCC Implementation

The actual algorithm is exposed as rapids expression `AstPearson`. Example of a Rapids expression:

```
(pearson iris_pearson 'sepal_len' 'petal_len')
```

It is **tested** in `RapidsTest` class. The `testPearson` test has been manually calculated and verified using [Wolfram Alpha](https://www.wolframalpha.com/widgets/view.jsp?id=3038cb5ccf72f21a13801d9c78f70937). The other test, testing calculation of linear correlation on the **Iris** dataset named `testPearsonIris` is also verified using [Wolfram Alpha](https://www.wolframalpha.com/widgets/view.jsp?id=3038cb5ccf72f21a13801d9c78f70937) and using another [third party tool](https://www.socscistatistics.com/tests/pearson/default2.aspx).

## Questions

1. Provided we want to stick to using `double`, should we use [Iterative mean](http://www.heikohoffmann.de/htmlthesis/node134.html) to calculate the mean values in `VecUtils.calculateMeansIgnoringNas`, more specifically in `VecUtils.NaIgnoringMeanTask` ? I'm afraid about double overflow with current plain additive method used.
2. Do we want to expose PCC (and later SCC) using a dedicated function in Python and R, e.g. like `h2o.pearson()` ? Do we want to group those somehow, as the number of descriptive statistics might increase.

### Fun fact
Did you know `sepal length` and `petal length` have strong positive correlation according to Pearson's ?

### DIFF
--- a/h2o-core/src/main/java/water/rapids/Env.java
+++ b/h2o-core/src/main/java/water/rapids/Env.java
@@ -15,6 +15,7 @@ import water.rapids.ast.prims.operators.*;
 import water.rapids.ast.prims.reducers.*;
 import water.rapids.ast.prims.repeaters.*;
 import water.rapids.ast.prims.search.*;
+import water.rapids.ast.prims.statistics.AstPearson;
 import water.rapids.ast.prims.string.*;
 import water.rapids.ast.prims.time.*;
 import water.rapids.ast.prims.timeseries.*;
@@ -297,6 +298,9 @@ public class Env extends Iced {
     init(new AstRepLen());
     init(new AstSeq());
     init(new AstSeqLen());
+    
+    // Statistics
+    init(new AstPearson());
 
     // Custom (eg. algo-specific)
     for (AstPrimitive prim : PrimsService.INSTANCE.getAllPrims())
--- a/h2o-core/src/main/java/water/rapids/ast/prims/statistics/AstPearson.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/statistics/AstPearson.java
@@ -1,0 +1,163 @@
+package water.rapids.ast.prims.statistics;
+
+import water.DKV;
+import water.MRTask;
+import water.Value;
+import water.fvec.Chunk;
+import water.fvec.Frame;
+import water.fvec.Vec;
+import water.rapids.Env;
+import water.rapids.Val;
+import water.rapids.ast.AstPrimitive;
+import water.rapids.ast.AstRoot;
+import water.rapids.vals.ValNum;
+import water.util.VecUtils;
+
+import java.util.Objects;
+
+/**
+ * Pearson's correlation coefficient (PCC) calculation
+ *
+ * Note: If one of the observation's columns used to calculate Pearson's corr. coef. contains a NaN value,
+ * the whole line is skipped.
+ */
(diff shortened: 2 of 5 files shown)


h2o-core/src/main/java/water/rapids/Env.java
```diff
@@ -297,6 +298,9 @@ static void init(AstPrimitive ast, String name) {
     init(new AstRepLen());
     init(new AstSeq());
     init(new AstSeqLen());
+    
+    // Statistics
+    init(new AstPearson());
```
michalkurka: This already exist in AstCor

Pscheidl: Fail. I could've looked at it a little bit more. Would we still consider using parts of my code ? Some time in future? Or should I close it ?

- It's API is design to work on two columns only for simplicity, but it can easily be adapted for `n` columns, just like the original `AstCorrelation`.
- It doesn't need to duplicate any data to fight NaNs present in the dataset - the `array` method of `AstCorrelation` does that for NAs. It calculates means without NaN rows in a distributed manner.
- My code could be optimized to use `Vec.mean` when there are no NAs in the vec, which is an information give "for free" from vec's stats.

wendycwong: Good point Michalk.  It is easy to miss since there is no Pearson in the title of the rapids call.

wendycwong: I agree with Pavel with extending it to do multiple correlations instead of two columns can be useful.

wendycwong:
I finished the review.  It is quite an elegant implementation.  If you can showcase that this implementation is faster, I think we can replace the original implementation.  What do you think @michalk?

PR #4104, opened 2019-11-25
PUBDEV-7057: add JSON option of PrintMojo

honzasterba:
- necessary to convert GBM models to ONNX

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
@@ -8,10 +8,7 @@ import hex.genmodel.tools.PrintMojo;
 import hex.genmodel.utils.GenmodelBitSet;
 
 import java.io.PrintStream;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.BitSet;
-import java.util.Objects;
+import java.util.*;
 
 /**
  * Node in a tree.
@@ -42,8 +39,6 @@ public class SharedTreeNode implements INode<double[]>, INodeStat {
   // When a column is categorical, levels that are reachable to this node.
   // This in particular includes any earlier splits of the same colId.
   private BitSet inclusiveLevels;
-
-  private Logger logger = LoggerFactory.getLogger(SharedTreeNode.class);
 
   /**
    * Create a new node.
@@ -451,6 +446,37 @@ public class SharedTreeNode implements INode<double[]>, INodeStat {
     }
   }
 
+  public Map<String, Object> toJson() {
+    Map<String, Object> json = new LinkedHashMap<>();
+    json.put("nodeNumber", nodeNumber);
+    if (!Float.isNaN(weight)) json.put("weight", weight);
+    if (isLeaf()) {
+      if (!Float.isNaN(predValue)) json.put("predValue", predValue);
+    } else {
+      json.put("colId", colId);
+      json.put("colName", colName);
+      json.put("leftward", leftward);
+      json.put("isCategorical", isBitset());
+      json.put("inclusiveNa", inclusiveNa);
+      if (!Float.isNaN(splitValue)) {
+        json.put("splitValue", splitValue);
+      }
+    }
+    if (inclusiveLevels != null) {
+      List<Integer> matchedDomainValues = new ArrayList<>();
+      for (int i = inclusiveLevels.nextSetBit(0); i >= 0; i = inclusiveLevels.nextSetBit(i+1)) {
+        matchedDomainValues.add(i);
+      }
+      json.put("matchValues", matchedDomainValues);
+      json.put("inclusiveNa", inclusiveNa);
+    }
+    if (!isLeaf()) {
+      json.put("rightChild", rightChild.toJson());
+      json.put("leftChild", leftChild.toJson());
+    }
+    return json;
+  }
+  
   public SharedTreeNode getParent() {
     return parent;
   }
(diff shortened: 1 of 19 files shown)


h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
```diff
@@ -443,6 +440,35 @@ void printDotEdges(PrintStream os, int maxLevelsToPrintPerEdge, float totalWeigh
     }
   }
 
+  public Map<String, Object> toJson() {
+    Map<String, Object> json = new HashMap<>();
+    json.put("nodeNumber", nodeNumber);
```
michalkurka: is the bitset iself included? should the "json" representation be less raw and geared towards the visualization? eg. include what categorical values go which way?

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -153,9 +176,6 @@ private void parseArgs(String[] args) {
             nPlaces = Integer.parseInt(s);
             break;
 
-          case "--raw":
```
michalkurka: should be backward compatible, please keep this option - it can be removed from help but it should keep working

michalkurka:
How are categorical values handled? If we expose it to the users it should be complete in the sense users can interpret it even for categorical values.

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
```diff
@@ -443,6 +440,37 @@ void printDotEdges(PrintStream os, int maxLevelsToPrintPerEdge, float totalWeigh
...
+    json.put("colId", colId);
+    json.put("colName", colName);
+    json.put("leftward", leftward);
+    json.put("naVsRest", naVsRest);
+    json.put("inclusiveNa", inclusiveNa);
+    json.put("isCategorical", isBitset());
+    if (!Float.isNaN(predValue)) json.put("predValue", predValue);
+    if (!Float.isNaN(squaredError)) json.put("squaredError", squaredError);
+    if (domainValues != null && bs != null) {
+      List<String> matchedDomainValues = new ArrayList<>();
```
michalkurka: just keep in mind that the domain can be huge and so can be the json files

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
```diff
@@ -549,6 +577,19 @@ public final int getSplitIndex() {
   public final int next(double[] value) {
     final double d = value[colId];
 
+    if (
```
michalkurka: is this just reformatted code from the few lines bellow?

honzasterba: yes, this is just me tryinng to undrestand things

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
honzasterba: is there something we can do about that? use the bitset representation somehow?

michalkurka: the easiest thing to do would be to refer to the values using numerical indices and keep the mapping in a separate file

decoding the bitset would also be possible

I think we do not have to overcomplicate it for now - just something to keep in mind for the future

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
honzasterba: no, see domainValues handling below

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
honzasterba: i find this notation more readable and would like to keep it

michalkurka: sure

honzasterba:
regarding categoricals: a split node either has a `splitValue` which is a float or it contains the list of matching categorical values
`"matchValues": [
            "Iris-setosa"
          ]`

h2o-py/h2o/h2o.py
```diff
@@ -2120,6 +2122,41 @@ def upload_mojo(mojo_path):
...
+        cmd += ["--tree", str(tree_index)]
+    output_file = tempfile.mkstemp(prefix="mojo_output")[1]
+    try:
+        with open(output_file, 'w+') as stdout:
+            return_code = subprocess.call(cmd, stdout=stdout)
+            stdout.seek(0)
+            output = stdout.read()
+    except OSError as e:
+        traceback = getattr(e, "child_traceback", None)
+        raise H2OError("Unable to print MOJO: %s" % e, traceback)
```
seb-h2o: you don't need to handle Py2.7 special case here ("child_traceback").
Official supports ends in 3 weeks...

seb-h2o: https://pythonclock.org/

h2o-py/h2o/h2o.py
```diff
@@ -2120,6 +2122,41 @@ def upload_mojo(mojo_path):
...
+        format = "json"
+    cmd = [java, "-cp", jar, "hex.genmodel.tools.PrintMojo", "--input", mojo_path, "--format", format]
+    if tree_index is not None:
+        cmd += ["--tree", str(tree_index)]
+    output_file = tempfile.mkstemp(prefix="mojo_output")[1]
+    try:
+        with open(output_file, 'w+') as stdout:
+            return_code = subprocess.call(cmd, stdout=stdout)
+            stdout.seek(0)
+            output = stdout.read()
```
seb-h2o: this is all Python 2.x way of handling subprocesses, most of it is deprecated.
If we don't have any generic impl in h2o that handles subprocesses for both Python 3 and 2, then we should  change that.

seb-h2o: especially we should ensure that the subprocess is always terminated or killed.

h2o-py/h2o/h2o.py
honzasterba: yes, but accoriding to @michalkurka our customers might be stuck on py2 for some more time

h2o-py/h2o/h2o.py
honzasterba: same as above, we probably need to support py2 for some more time

seb-h2o: sure, not saying we should drop support.
But I'll create ticket to ensure that we move towards more Python3 friendliness (= hide the deprecated, sometimes buggy, Py2-specific methods or techniques).

PR #4108, opened 2019-11-25
[PUBDEV-7099] Remove dead code in UDPClientEvent

jakubhava:
Not causing any bugs, but it is a dead code, not used anywhere


h2o-core/src/main/java/water/H2O.java
```diff
@@ -1574,10 +1574,7 @@ public static String DEFAULT_FLOW_DIR() {
   /* A static list of acceptable Cloud members passed via -flatfile option.
    * It is updated also when a new client appears. */
   private static Set<H2ONode> STATIC_H2OS = null;
-
-  /* List of all clients that ever connected to this cloud. Keys are IP:PORT of these clients */
-  private static Map<String, H2ONode> CLIENTS_MAP = new ConcurrentHashMap<>();
```
jakubhava: unused

PR #4109, opened 2019-11-26
PUBDEV-7073 extended logging possibilities

koniecsveta:
when creating a hex.genmodel.easy.EasyPredictModelWrapper enableContributions parameter is false by default already.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -83,6 +83,7 @@ ext {
       project(':h2o-ext-jython-cfunc'),
       project(':h2o-hive'),
       project(':h2o-security'),
+      project(':h2o-logger'),
     ]
 
     javaProjects = [
@@ -115,6 +116,7 @@ ext {
       project(':h2o-ext-jython-cfunc'),
       project(':h2o-hive'),
       project(':h2o-security'),
+      project(':h2o-logger'),      
     ]
 
     scalaProjects = [
--- a/h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeMojoModel.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeMojoModel.java
@@ -6,6 +6,8 @@ import hex.genmodel.algos.gbm.GbmMojoModel;
 import hex.genmodel.attributes.VariableImportances;
 import hex.genmodel.utils.ByteBufferWrapper;
 import hex.genmodel.utils.GenmodelBitSet;
+import water.logging.Logger;
+import water.logging.LoggerFactory;
 
 import java.util.Arrays;
 import java.util.HashMap;
@@ -21,6 +23,8 @@ public abstract class SharedTreeMojoModel extends MojoModel implements SharedTre
     private static final int NsdLeft = NaSplitDir.Left.value();
 
     private ScoreTree _scoreTree;
+    
+    private static Logger logger = LoggerFactory.getLogger(SharedTreeMojoModel.class);
 
     /**
      * {@code _ntree_groups} is the number of trees requested by the user. For
@@ -640,17 +644,17 @@ public abstract class SharedTreeMojoModel extends MojoModel implements SharedTre
         weight_ok = (Math.abs(node.getWeight() - sum) < 1e-5 * (node.getWeight() + sum));
         ok &= weight_ok;
       }
-      if (!ok) {
-        System.err.println("\nTree inconsistency found:");
-        if (node.depth == 1 && !weight_ok) {
-          System.err.println("Note: this is a known issue for DRF and Isolation Forest models, " +
+      if (!ok && logger.isErrorEnabled()) {
+        logger.error("\nTree inconsistency found:");
+        if (node.depth == 1 && !weight_ok) { 
+            logger.error("Note: this is a known issue for DRF and Isolation Forest models, " +
                   "please refer to https://0xdata.atlassian.net/browse/PUBDEV-6140");
         }
-        node.print(System.err, "parent");
-        node.leftChild.print(System.err, "left child");
-        node.rightChild.print(System.err, "right child");
-        System.err.println("Auxiliary tree info:");
-        System.err.println(auxInfo.toString());
+        logger.error(node.getPrintString("parent"));
+        logger.error(node.leftChild.getPrintString("left child"));
+        logger.error(node.rightChild.getPrintString("right child"));
+        logger.error("Auxiliary tree info:");
+        logger.error(auxInfo.toString());
       }
     }
 
--- a/h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
@@ -2,6 +2,8 @@ package hex.genmodel.algos.tree;
 
 import ai.h2o.algos.tree.INode;
 import ai.h2o.algos.tree.INodeStat;
+import water.logging.Logger;
+import water.logging.LoggerFactory;
 import hex.genmodel.tools.PrintMojo;
 import hex.genmodel.utils.GenmodelBitSet;
 
@@ -40,6 +42,8 @@ public class SharedTreeNode implements INode<double[]>, INodeStat {
   // When a column is categorical, levels that are reachable to this node.
   // This in particular includes any earlier splits of the same colId.
   private BitSet inclusiveLevels;
+
+  private Logger logger = LoggerFactory.getLogger(SharedTreeNode.class);
 
   /**
    * Create a new node.
@@ -252,19 +256,23 @@ public class SharedTreeNode implements INode<double[]>, INodeStat {
   }
 
   public void print(PrintStream out, String description) {
-    out.println("        Node " + nodeNumber + (description != null ? " (" + description + ")" : ""));
-    out.println("            weight:      " + weight);
-    out.println("            depth:       " + depth);
-    out.println("            colId:       " + colId);
-    out.println("            colName:     " + ((colName != null) ? colName : ""));
-    out.println("            leftward:    " + leftward);
-    out.println("            naVsRest:    " + naVsRest);
-    out.println("            splitVal:    " + splitValue);
-    out.println("            isBitset:    " + isBitset());
-    out.println("            predValue:   " + predValue);
-    out.println("            squaredErr:  " + squaredError);
-    out.println("            leftChild:   " + ((leftChild != null) ? leftChild.getName() : ""));
-    out.println("            rightChild:  " + ((rightChild != null) ? rightChild.getName() : ""));
+    out.print(this.getPrintString(description));
+  }
+
+  public String getPrintString(String description) {
+    return "        Node " + nodeNumber + (description != null ? " (" + description + ")" : "") +
+            "\n            weight:      " + weight +
+            "\n            depth:       " + depth +
+            "\n            colId:       " + colId +
+            "\n            colName:     " + ((colName != null) ? colName : "") +
+            "\n            leftward:    " + leftward +
+            "\n            naVsRest:    " + naVsRest +
+            "\n            splitVal:    " + splitValue +
+            "\n            isBitset:    " + isBitset() +
+            "\n            predValue:   " + predValue +
+            "\n            squaredErr:  " + squaredError +
+            "\n            leftChild:   " + ((leftChild != null) ? leftChild.getName() : "") +
+            "\n            rightChild:  " + ((rightChild != null) ? rightChild.getName() : "");
   }
 
   void printEdges() {
--- a/h2o-logger/build.gradle
+++ b/h2o-logger/build.gradle
@@ -1,0 +1,15 @@
+apply plugin: 'java'
+
+sourceCompatibility = 1.7
+targetCompatibility = 1.7
+
+compileJava {
+    options.debug = true
+}
+
+
+dependencies {
+    testCompile group: 'junit', name: 'junit', version: '4.12'
+    compileOnly "org.slf4j:slf4j-api:1.7.29"
+    testCompile "org.slf4j:slf4j-api:1.7.29"
+}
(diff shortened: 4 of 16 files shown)


h2o-logger/src/main/java/logging/ConsoleLogger.java
```diff
@@ -0,0 +1,43 @@
+package logging;
```
honzasterba: should be either water, (public) or hex. (internal)

h2o-logger/src/main/java/logging/ConsoleLogger.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.io.PrintStream;
+
+public class ConsoleLogger implements ILogger {
```
honzasterba: do we really need this? the default slfj implementation I think does pretty much the same, so I would just use slf4j in genmodel and be done with it, @michalkurka WDYT?

koniecsveta: in http://www.slf4j.org/manual.html there is "if no binding is found on the class path, then slf4j-api will default to a no-operation implementation discarding all log requests."

honzasterba: oh right, thats a bit of a conundrum, then we probably need this class

h2o-logger/src/main/java/logging/ConsoleLogger.java
koniecsveta: fixed

h2o-logger/build.gradle
```diff
@@ -0,0 +1,17 @@
+plugins {
+    id 'java'
+}
+
+group 'ai.h2o'
+version '3.26.0'
+
+sourceCompatibility = 1.8
```
honzasterba: missing targetCompatibility

h2o-logger/build.gradle
```diff
@@ -0,0 +1,17 @@
+plugins {
+    id 'java'
+}
+
+group 'ai.h2o'
```
honzasterba: remove these two lines, this is done automagically

h2o-logger/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+version '3.26.0'
+
+sourceCompatibility = 1.8
+
+repositories {
```
honzasterba: remove, unnecessary

h2o-logger/build.gradle
koniecsveta: fixed.

h2o-logger/build.gradle
koniecsveta: fixed, thx.

h2o-logger/build.gradle
koniecsveta: fixed.

build.gradle
```diff
@@ -115,6 +115,7 @@ ext {
       project(':h2o-ext-jython-cfunc'),
       project(':h2o-hive'),
       project(':h2o-security'),
+      project(':h2o-logger'),
```
michalkurka: please don't forget to add to published projects as well

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
```diff
@@ -267,6 +269,33 @@ public void print(PrintStream out, String description) {
     out.println("            rightChild:  " + ((rightChild != null) ? rightChild.getName() : ""));
   }
 
+  public void log(LoggingLevel loggingLevel, String description) {
```
michalkurka: this is a relatively expensive function - perhaps the message shouldn't be constructed if won't be logged in the end

h2o-logger/src/main/java/water/logging/ConsoleLogger.java
```diff
@@ -0,0 +1,43 @@
+package water.logging;
+
+import java.io.PrintStream;
+
+public class ConsoleLogger implements ILogger {
+
+    @Override
+    public void trace(String paramString) {
```
michalkurka: could be called message instead?

h2o-logger/src/test/java/LoggerTest.java
```diff
@@ -0,0 +1,49 @@
...
+    public void noSlf4JAvailable() {
+        ILogger logger = Logger.getLogger(LoggerTest.class);
+        assertTrue(logger instanceof ConsoleLogger);
+    }
+
+    @Test
+    public void Slf4JAvailable() throws NoSuchMethodException, InvocationTargetException, IllegalAccessException, IOException {
+        ClassLoader classLoader = LoggerTest.class.getClassLoader();
+        
+        URL apiUrl = classLoader.getResource("slf4j-api-1.7.29.jar");
```
michalkurka: I would just do set JUnit tests instead of this one

if you had a LoggerFactory, you wouldn't need to have the Logger class all static

if it was not static you can parametrize the SLF4J class that is being checked (and you can use any dummy class) to get ConsoleLogger instead of manipulating the classpath

build.gradle
koniecsveta: added

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
koniecsveta: added default

h2o-logger/src/main/java/water/logging/ConsoleLogger.java
koniecsveta: renamed

h2o-logger/src/test/java/LoggerTest.java
koniecsveta: Logger class parametrized

h2o-logger/src/main/java/water/logging/Logger.java
```diff
@@ -0,0 +1,38 @@
+package water.logging;
+
+
+
+/**
+ * Logger provides SLF4J logging if SLF4J is on the classpath and otherwise provides logging straight to the console output.
+ */
+public class Logger <T extends Class> implements ILogger{
```
honzasterba: Is seems this class is not needed any more and we can replace it with ILogger (it can be just the interface

scripts/jenkins/Makefile.jenkins
```diff
@@ -220,6 +220,8 @@ test-package-java:
 		h2o-algos/build/libs/h2o-algos-*.jar \
 		h2o-automl/src/ \
 		h2o-automl/build/libs/h2o-automl-*.jar \
+		h2o-logger/src/ \
+        h2o-logger/build/libs/h2o-logger-*.jar \
```
honzasterba: fix indent

h2o-logger/src/test/java/water/logging/LoggerTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.Test;
+
+import java.lang.reflect.Field;
+
+
+import static org.junit.Assert.*;
+
+public class LoggerTest {
+    @Test
+    public void FactoryTest() {
```
honzasterba: even test methods should start with lowe-case

h2o-logger/src/test/java/water/logging/LoggerTest.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+
+    @Test (expected = ClassCastException.class)
+    public void ConstructorTest() throws IllegalAccessException, NoSuchFieldException {
+        
+        if (!isSlf4JAvailable()) {
+            Logger logger = new Logger(ConsoleLogger.class, LoggerTest.class);
+            Field f = logger.getClass().getDeclaredField("logger");
+            ConsoleLogger consoleLogger = (ConsoleLogger) f.get(logger);
```
honzasterba: its not clear which one of these is supposed to throw the CCEx, this thest IMO will always pass

buildSrc/src/main/java/H2OBuildVersion.java
```diff
@@ -129,6 +129,7 @@ private String calcLastCommitHash() {
       return line;
     }
     catch (Exception e) {
+      e.printStackTrace();
```
honzasterba: can be removed

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeMojoModel.java
```diff
@@ -641,16 +646,16 @@ static void checkConsistency(AuxInfo auxInfo, SharedTreeNode node) {
...
+        if (node.depth == 1 && !weight_ok) { 
+            logger.error("Note: this is a known issue for DRF and Isolation Forest models, " +
                   "please refer to https://0xdata.atlassian.net/browse/PUBDEV-6140");
         }
-        node.print(System.err, "parent");
-        node.leftChild.print(System.err, "left child");
-        node.rightChild.print(System.err, "right child");
-        System.err.println("Auxiliary tree info:");
-        System.err.println(auxInfo.toString());
+        node.log(LoggingLevel.ERROR, "parent");
```
honzasterba: static import would be nicer

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeNode.java
honzasterba: also I dont like the duplication of this code, we construct the same string as we do in print(), personally I would just have `print() { out.print(this.getPrintString(); }` and then getPrintString() method with the current contents of log() method, and in sharedtreemojoModel I would have 
`if (logger.isEnabled(ERROR) { logger.error(node.getPrintString()); }`

h2o-logger/src/main/java/water/logging/LoggerFactory.java
```diff
@@ -15,7 +20,21 @@
      * @return Slf4JLogger if SLF4J is on the classpath, ConsoleLogger if not
      */
     public static Logger getLogger(Class<?> clazz) {
```
honzasterba: this method is not covered by tests, thats why having the static LoggerFactory INSTANCE is the best solution

h2o-logger/src/main/java/water/logging/LoggerFactory.java
```diff
@@ -15,7 +20,21 @@
...
+        }
+    }
+
+    /**
+     * Returns new logger for each invocation depending on whether custom class is available.
+     *
+     * @param clazz class from which getLogger() is called
+     * @return Slf4JLogger if SLF4J is on the classpath, ConsoleLogger if not
+     */
+    Logger getCustomLogger(Class<?> clazz) {
```
honzasterba: this method is tested but not used anywhere in the code, therefore should be removed

h2o-logger/src/main/java/water/logging/LoggerFactory.java
```diff
@@ -0,0 +1,54 @@
...
+
+    
+    /**
+     * Returns new logger for each invocation.  Logger logs to console / SLF4J (custom wrapper around SLF4J) depending on whether SLF4J is available.
+     *
+     * @param clazz class from which getLogger() is called
+     * @return Slf4JLogger if SLF4J is on the classpath, ConsoleLogger if not
+     */
+    public Logger getCustomLogger(Class<?> clazz) {
+        if (isSlf4JAvailable()) {
```
honzasterba: I think we should do this check only once at init-time, doing class.forName for every time a logger is created can be expensive I think

h2o-logger/src/main/java/water/logging/LoggerFactory.java
```diff
@@ -0,0 +1,54 @@
+package water.logging;
+
+
+
+public class LoggerFactory {
+    private static final String DEFAULT_CLASS_TO_CHECK = "org.slf4j.LoggerFactory";
+    private static final LoggerFactory INSTANCE = new LoggerFactory(DEFAULT_CLASS_TO_CHECK);
+    
+    private String classToCheck;
```
honzasterba: slf4jClassName would be a better name I think

h2o-logger/src/main/java/water/logging/LoggerFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+    
+    /**
+     * Returns new logger for each invocation.  Logger logs to console / SLF4J (custom wrapper around SLF4J) depending on whether SLF4J is available.
+     *
+     * @param clazz class from which getLogger() is called
+     * @return Slf4JLogger if SLF4J is on the classpath, ConsoleLogger if not
+     */
+    public Logger getCustomLogger(Class<?> clazz) {
+        if (isSlf4JAvailable) {
```
michalkurka: can we make the decision just once?

IMHO better for the MOJO use case

michalkurka:
@koniecsveta looks good, please change the base branch to master

also I realized: we should also delegate to `water.util.Log` if the MOJO code is running inside H2O. Parts of genmodel are used by h2o runtime - in that case the logging should go to `water.util.Log` instead of system out/err.

@koniecsveta @honzasterba WDYT?

h2o-logger/src/main/java/water/logging/LoggerFactory.java
honzasterba: this is done once per class static-init so one if() is not a big issue I think

h2o-logger/src/main/java/water/logging/LoggerFactory.java
```diff
@@ -0,0 +1,56 @@
+package water.logging;
+
+
+
+public class LoggerFactory {
+    private static final String DEFAULT_CLASS_TO_CHECK = "org.slf4j.LoggerFactory";
+    private static final LoggerFactory INSTANCE = new LoggerFactory(DEFAULT_CLASS_TO_CHECK);
+    
+    private final String slf4jClassName;
```
honzasterba: we could extend this with waterClassName and look for water.util.Log if slf4j is not found

koniecsveta: making it this way we get circular dependency between modules with different language levels assigned: genmodel(7)->core(8)->logger(7)->genmodel(7).

PR #4110, opened 2019-11-26
[PUBDEV-7102] Ensure H2O does not prevent stopping when embedded inside another application

jakubhava:
https://0xdata.atlassian.net/browse/PUBDEV-7102


h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
```diff
@@ -41,7 +44,12 @@
     Server createJettyServer(String ip, int port) {
         System.setProperty("org.eclipse.jetty.server.Request.maxFormContentSize", Integer.toString(Integer.MAX_VALUE));
 
-        final Server jettyServer = new Server();
+        QueuedThreadPool pool = new QueuedThreadPool();
```
michalkurka: Do we need to do the same thing for Jetty 8?

jakubhava: I have manually tested Jetty 8 in context of Sparkling Water and it worked fine, the thread did not appear.

I can't also find this class in our Jetty 8 and from their git it seems like it wasn't put into jetty 8 https://github.com/eclipse/jetty.project/commit/fa412b4668e48590e89ad364541795fa73ba9ddc#diff-058fa4e5cdc53ae1c61956cddc416416

michalkurka: I meant `QueuedThreadPool` - it is part of Jetty 8 codebase

jakubhava: We do it, but just not in the constructor -> https://github.com/h2oai/h2o-3/pull/4110/files#diff-8fb7d504641963d01ccfd35dff8b56aeR55 Different jetty api

michalkurka: oh ok - missed that file, thanks

jakubhava:
@michalkurka do you think we are good to go with this change?

michalkurka:
@jakubhava all client mode tests failed, can you look into that?

jakubhava:
Yup, I think I know the reason, pushed a fix

jakubhava:
I was thinking about this today and I think it will be something in our tests ( how R is starting the client node), as in Sparkling Water the client node (running in JVM as embedded app) is working correctly

jakubhava:
Ah, I know the reason: when passing `-flatfile` option we do not have any long-running threads open ( all daemon) and therefore main finishes right away.

I will fix this + will also check for open threads in the multicast variant

h2o-core/src/main/java/water/H2O.java
```diff
@@ -2195,8 +2198,22 @@ public static void main( String[] args ) {
     Log.debug("    Start network services: " + (time10 - time9) + "ms");
     Log.debug("    Cloud up: " + (time11 - time10) + "ms");
     Log.debug("    Start GA: " + (time12 - time11) + "ms");
+
+    if (!ARGS.isEmbeddedApp) {
+      keepRunning();
```
jakubhava: @michalkurka When we ensures that all long-running threads were daemon, H2O actually stopped when it reached end of `main` method. This ensures that we keep h2o running.

Also it ensures that for all existing users of H2O, also the ones who uses H2O as embedded app, will not be affected.

Sparkling Water will set `embedded = true` to ensure we don't do this wait.

h2o-core/src/main/java/water/H2O.java
```diff
@@ -2195,8 +2201,24 @@ public static void main( String[] args ) {
     Log.debug("    Start network services: " + (time10 - time9) + "ms");
     Log.debug("    Cloud up: " + (time11 - time10) + "ms");
     Log.debug("    Start GA: " + (time12 - time11) + "ms");
+
+    if (!ARGS.isEmbeddedApp) {
+      keepH2OAliveThread.start();
```
jakubhava: @michalkurka When we ensures that all long-running threads were daemon, H2O actually stopped when it reached end of `main` method. This ensures that we keep h2o running.

Also it ensures that for all existing users of H2O, also the ones who uses H2O as embedded app, will not be affected.

Sparkling Water will set `embedded = true` to ensure we don't do this wait.

jakubhava:
There should no long-running threads in both multicast & flatfile case now. 

Also ensure the backwards compatibility, so no users should be affected.

h2o-core/src/main/java/water/H2O.java
```diff
@@ -2195,8 +2201,24 @@ public static void main( String[] args ) {
...
+    }
   }
+  private static Thread keepH2OAliveThread = new Thread("keepH2OAliveThread") {
 
+    @Override
+    public void run() {
+      while (true) {
```
jakubhava: Not a daemon thread on purpose to simulate how H2O worked before

jakubhava:
@michalkurka should be good now, tests are green :)

michalkurka:
@jakubhava I pushed one more commit to mimic the previous behavior - this would be safer for the release branch

jakubhava:
Looks good, thank you Michal!

PR #4111, opened 2019-11-27
[PUBDEV-7106] Report full flatfile in h2odriver

jakubhava:
Just proposal

### DIFF
--- a/h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
+++ b/h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
@@ -107,6 +107,7 @@ public class h2odriver extends Configured implements Tool {
   static String network = null;
   static boolean disown = false;
   static String clusterReadyFileName = null;
+  static String clusterFlatfileName = null;
   static String hadoopJobId = "";
   static String applicationId = "";
   static int cloudFormationTimeoutSeconds = DEFAULT_CLOUD_FORMATION_TIMEOUT_SECONDS;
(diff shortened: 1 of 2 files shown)


michalkurka:
lgtm - I would just fix the handling of the file resource but apart of that - I don't see an issue

jakubhava:
Perfect, let me have a look on this! Thank you!

h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
```diff
@@ -586,6 +599,7 @@ public void registerNode (String ip, int port, int attempt, Socket s) {
           flatfile += val;
           flatfile += "\n";
         }
+        createFlatFile(flatfile);
```
michalkurka: I would probably move this to somewhere higher level after the clouding is successfully completed.

If everything succeeds `flatfileContent` is available in a static field

jakubhava: Good idea, done

jakubhava:
@michalkurka please have a look. I have moved the creation & ensured we handle the IO in the same way as the cluster notification file

jakubhava:
Thanks for the changes @michalkurka, are we good to go?

michalkurka:
@jakubhava almost - I will need to run Hadoop smoke tests on this branch

I will merge when they pass

jakubhava:
Thank you!

PR #4113, opened 2019-11-27
PUBDEV-6876: added HGLM Gaussian/Gaussian to H2O-3

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6875

1. Algorithm implementation
2. R API
3. Model Metrics calculation
4. Python API

Still need to add the following:
5. More tests in R/Python/Java
6. documentation for new algo and also for my team mates to be better able to help me review.

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/ComputationState.java
+++ b/h2o-algos/src/main/java/hex/glm/ComputationState.java
@@ -16,6 +16,7 @@ import water.MemoryManager;
 import water.util.ArrayUtils;
 import water.util.Log;
 import water.util.MathUtils;
+import water.fvec.Frame;
 
 import java.util.Arrays;
 import java.util.Comparator;
@@ -31,6 +32,7 @@ public final class ComputationState {
   double [] _z;
   boolean _allIn;
   int _iter;
+  int _iterHGLM_GLMMME; // keep track of iterations used in estimating fixed/random coefficients
   private double _lambda = 0;
   private double _lambdaMax = Double.NaN;
   private GLMGradientInfo _ginfo; // gradient info excluding l1 penalty
@@ -39,6 +41,15 @@ public final class ComputationState {
   private DataInfo _activeData;
   private BetaConstraint _activeBC = null;
   private double[] _beta; // vector of coefficients corresponding to active data
+  private double[] _ubeta;  // HGLM, store coefficients of random effects;
+  private double[] _psi; // HGLM, psi
+  private double[] _phi; // HGLM, size random columns levels
+  private double _tau; // HGLM for ei
+  private double _correction_HL; // HGLM
+  double[] _sumEtaSquareConvergence;  // HGLM: sotre sumEtaSquare, convergence
+  double[] _likelihoodInfo; // HGLM: stores 4 elements: hlik, pvh, pbvh, cAIC
+  public String[] _randCoeffNames; // store random coefficient names
+  private Frame _priorw_wpsi;  // weight calculated for psi
   final DataInfo _dinfo;
   private GLMGradientSolver _gslvr;
   private final Job _job;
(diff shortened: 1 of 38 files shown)


michalkurka:
@wendycwong this is a masterpiece!

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -291,7 +301,7 @@ public void validate(GLM glm) {
         glm.hide("_nlambdas", "only applies if lambda search is on.");
         glm.hide("_early_stopping","only applies if lambda search is on.");
       }
-      if (_family == Family.ordinal) {
+      if (_family.equals(Family.ordinal)) {
```
honzasterba: == is fine for Enums

h2o-algos/src/main/java/hex/util/LinearAlgebraUtils.java
```diff
@@ -146,6 +268,70 @@ public BMulTask(Key<Job> jobKey, DataInfo dinfo, double[][] yt) {
...
+    }
+    
+    @Override public void map(Chunk[] xChunk) {
+      Chunk[] ychunk = new Chunk[_y.numCols()];
+      for (int ychunkInd=0; ychunkInd < _nyChunks; ychunkInd++) {
+        for (int chkIndex =0 ; chkIndex < _yColNum; chkIndex++) // grab a y chunk
+          ychunk[chkIndex] = _y.vec(chkIndex).chunkForChunkIdx(ychunkInd);
+        mulResultPerYChunk(xChunk, ychunk);
+      }
+/*      fetchChunkInfo =  getCorrectChunk(_y, 0, startRow, ychunk, null, fetchChunkInfo);
```
honzasterba: commented code with no explanation or TODO

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -23,6 +23,7 @@
 import java.lang.reflect.Method;
 import java.net.URL;
 import java.util.*;
+import java.io.*;
```
honzasterba: only added import

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -58,6 +69,73 @@ public ComputationState(Job job, GLMParameters parms, DataInfo dinfo, BetaConstr
...
+      _sumEtaSquareConvergence = new double[2];
+      if (_parms._calc_like)
+        _likelihoodInfo = new double[4];
+    }
+  }
+  public void set_sumEtaSquareConvergence(double[] sumInfo) {
+    _sumEtaSquareConvergence = sumInfo;
+  }
+  
+  public void set_beta_HGLM(double[] beta, int startIdx, int len, boolean interceptFirst) {
```
Pscheidl: `len` is length of what ? By looking at this method and seeing `  _beta = new double[len];`, I'm super,confused. 

Please add JavaDoc here or use better naming. JavaDoc is fine.

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -539,6 +618,34 @@ protected double updateState(double [] beta,GLMGradientInfo ginfo){
...
+    _randCoeffNames = copyArray(randCoeffNames, _randCoeffNames);
+    _psi = copyArray(psi, _psi);
+    _phi = copyArray(phi, _phi);
+    _correction_HL = hlcorrection;
+    _tau = tau;
+    _priorw_wpsi = wpsi;  // store prior_weight and calculated wpsi value for coefficients of random columns
+    _iterHGLM_GLMMME = 0;
+  }
+  
+  private double[] copyArray(double[] source, double[] dest) {
```
Pscheidl: For these two methods, it looks like we could use built-in `Arrays.copyOf`. If not and you really need a method to copy into an existing instance,  may I suggest moving these methods to one of `ArrayUtils` classes we have in H2O.

As a matter of personal preference, if I'd have been allocating primitive array types, I'd use `MemoryManager` - the more H2O memory manager knows about, the better I think.

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -166,6 +169,20 @@ public void revertCoeffNames(String[] colNames, int nclass, TwoDimTableV3 coeffs
           coeffs_table.columns[tableIndex+nclass].name = new String(colNames[tableIndex-1+nclass]);
       }
     }
+    
+    public TwoDimTable build2DTable(double[] ubeta, String[] randomColNames) {
```
Pscheidl: To strangers, this methods looks like it can build any 2D table. In fact, it contains the random coefficients. For searching in the code, it'd be appreciated to have this method named something like `buildRandomCoefficientsTable`.

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -225,6 +242,9 @@ public GLMModelOutputV3 fillFromImpl(GLMModel.GLMOutput impl) {
         }
       }
       coefficients_table.fillFromImpl(tdt);
+      if (impl.ubeta() != null)
```
Pscheidl: Are we testing `if (impl.ubeta() != null)` twice ? Here and on line `200` ? It could all be done in one `if` block.

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -225,6 +242,9 @@ public GLMModelOutputV3 fillFromImpl(GLMModel.GLMOutput impl) {
...
+        random_coefficients_table.fillFromImpl(build2DTable(impl.ubeta(), impl.randomcoefficientNames()));
```
Pscheidl: What if `impl.randomcoefficientNames()` is null ?

h2o-algos/src/main/java/hex/util/LinearAlgebraUtils.java
```diff
@@ -123,6 +220,50 @@ else if (!dinfo._catMissing[col])
...
+        this._maxIndex = other._maxIndex;
+      else if (this._maxIndex > other._maxIndex)
+        this._maxIndex = other._maxIndex; 
+    }
+  }
+  
+  public static class CopyQtoQMatrix extends MRTask<CopyQtoQMatrix> {
+    @Override public void map(Chunk[] cs) {
+      int totColumn = cs.length;  // all columns in cs.
+      int halfColumn = totColumn/2; // start of Q matrix
```
Pscheidl: What if totColumn is an odd number ?

h2o-algos/src/main/java/hex/util/LinearAlgebraUtils.java
```diff
@@ -123,6 +220,50 @@ else if (!dinfo._catMissing[col])
...
+      int totRows = cs[0].len();
```
Pscheidl: Are we 100% sure all chunks have the same length ? As this is not guaranteed for some Frames. I'm only asking, don't have deep insight ATM.

michalkurka: MRTasks can only run on frames that have the same layout => all chunks do have the same length.

wendycwong:
@michalk:  Masterpiece huh?  stupid thing is not even passing any tests!

wendycwong:
Okay, I have finished part of the documentation that can be helpful with the code review.  Let me attach it here.  
[H2OHGLMDocumentation.pdf](https://github.com/h2oai/h2o-3/files/3928501/H2OHGLMDocumentation.pdf)

h2o-algos/src/main/java/hex/glm/ComputationState.java
wendycwong: Added the description in JavaDoc.  Thanks.

h2o-algos/src/main/java/hex/glm/ComputationState.java
wendycwong: Agree on both points.  Done.

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: Fixed.

h2o-algos/src/main/java/hex/util/LinearAlgebraUtils.java
wendycwong: here, the totColumn is always an even number because the chunk cs is a repeat of [frame A frame A].

h2o-algos/src/main/java/hex/glm/GLMMetricBuilder.java
```diff
@@ -38,33 +41,36 @@
...
-          ((MetricBuilderMultinomial)_metricBuilder)._priorDistribution = ymu;
-          break;
-        case ordinal:
-          _metricBuilder = new MetricBuilderOrdinal(domain.length,domain);
-          ((MetricBuilderOrdinal)_metricBuilder)._priorDistribution = ymu;
-          break;
-        default:
-          _metricBuilder = new MetricBuilderRegression();
-          break;
+      if (domain!=null && domain.length==1 && domain[0].contains("HGLM")) {
```
michalkurka: @wendycwong is there a better way to detect that we are running HGLM?

wendycwong: Michalk:  I don't have access to parms and I don't want to break it for other GLM....

PR #4040, opened 2019-11-04
PUBDEV-7037: Support for handling interdependent hyper parameters for grid search

deil87:
WIP

This PR is supposed to add support for filtering out grid items for existing Cartesian and Random lazily evaluated grid searches.

Current state of PR support filtering only for Random walker. Once concept is reviewed and approved I can add support for Cartesian walker as well. 

Note: once parallel grid search is finished PUBDEV-6774 https://github.com/h2oai/h2o-3/pull/3990, it should be double checked that current functionality is not broken.

PR #4114, opened 2019-11-28
PUBDEV-7110: Early stopping logic should not be located within HyperSpaceIterator's hasNext method

deil87:
Need for this changes has arisen from within https://github.com/h2oai/h2o-3/pull/4040

These changes separate iteration logic from logic for early stopping. This also will enable us to support ability to nest multiple `FilteredWalker`s by properly implementing `hasNext` method.

### DIFF
--- a/h2o-algos/src/test/java/hex/grid/GridTest.java
+++ b/h2o-algos/src/test/java/hex/grid/GridTest.java
@@ -4,6 +4,7 @@ import hex.Model;
 import hex.genmodel.utils.DistributionFamily;
 import hex.tree.gbm.GBMModel;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -311,10 +312,7 @@ public class GridTest extends TestUtil {
       params._train = trainingFrame._key;
       params._response_column = "species";
 
-      GridSearch.SimpleParametersBuilderFactory simpleParametersBuilderFactory = new GridSearch.SimpleParametersBuilderFactory();
-      HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria hyperSpaceSearchCriteria = new HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria();
-
-      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms, simpleParametersBuilderFactory, hyperSpaceSearchCriteria, 1);
+      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms);
       Scope.track_generic(gs);
       final Grid errGrid = gs.get();
       Scope.track_generic(errGrid);
@@ -498,7 +496,7 @@ public class GridTest extends TestUtil {
         put("_distribution", new DistributionFamily[]{DistributionFamily.multinomial});
         put("_ntrees", new Integer[]{5});
         put("_max_depth", new Integer[]{2});
-        put("_min_rows", new Integer[]{10}); // Invalid hyperparameter, causes model training to fail
+        put("_min_rows", new Integer[]{10});
         put("_learn_rate", new Double[]{.7});
       }};
 
@@ -506,10 +504,7 @@ public class GridTest extends TestUtil {
       params._train = trainingFrame._key;
       params._response_column = "species";
 
-      GridSearch.SimpleParametersBuilderFactory simpleParametersBuilderFactory = new GridSearch.SimpleParametersBuilderFactory();
-      HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria hyperSpaceSearchCriteria = new HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria();
-
-      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms, simpleParametersBuilderFactory, hyperSpaceSearchCriteria, 1);
+      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms);
       Scope.track_generic(gs);
       final Grid errGrid = gs.get();
       Scope.track_generic(errGrid);
@@ -526,7 +521,7 @@ public class GridTest extends TestUtil {
 
       // Train a second model with modified parameters to forcefully produce a new model
       hyperParms.put("_learn_rate", new Double[]{0.5});
-      gs = GridSearch.startGridSearch(errGrid._key, params, hyperParms, simpleParametersBuilderFactory, hyperSpaceSearchCriteria, 1);
+      gs = GridSearch.startGridSearch(errGrid._key, params, hyperParms);
       Scope.track_generic(gs);
       final Grid grid = gs.get();
       Scope.track_generic(grid);
@@ -560,7 +555,7 @@ public class GridTest extends TestUtil {
     );
   }
 
-  @Test // this test is fine as with Cartesian we don't really have early stopping based on max_models ( always train whole hyper space)
+  @Test
   public void testParallelCartesian() {
     try {
       Scope.enter();
@@ -669,6 +664,100 @@ public class GridTest extends TestUtil {
       Scope.track_generic(grid1);
 
       assertEquals(custom_max_model, grid1.getModelCount());
+    } finally {
+      Scope.exit();
+    }
+  }
+
+  @Test
+  public void testParallelRandomSearchWithDefaultMaxModels_MultiRun() {
+    try {
+      Scope.enter();
+      final Frame trainingFrame = parse_test_file("smalldata/iris/iris_train.csv");
+      Scope.track(trainingFrame);
+
+      // Setup random hyperparameter search space
+      HashMap<String, Object[]> hyperParms = new HashMap<String, Object[]>() {{
+        put("_distribution", new DistributionFamily[]{DistributionFamily.multinomial});
+        put("_ntrees", new Integer[]{5});
+        put("_max_depth", new Integer[]{2});
+        put("_min_rows", new Integer[]{10,11,12,13,14});
+        put("_learn_rate", new Double[]{.7});
+      }};
+
+      GBMModel.GBMParameters params = new GBMModel.GBMParameters();
+      params._train = trainingFrame._key;
+      params._response_column = "species";
+
+      GridSearch.SimpleParametersBuilderFactory simpleParametersBuilderFactory = new GridSearch.SimpleParametersBuilderFactory();
+      HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria hyperSpaceSearchCriteria = new HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria();
+
+      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms, simpleParametersBuilderFactory, hyperSpaceSearchCriteria, 6);
+      Scope.track_generic(gs);
+      final Grid grid1 = gs.get();
+      Scope.track_generic(grid1);
+
+      assertEquals(5, grid1.getModelCount());
+
+      // Train a grid with new hyper parameters
+      hyperParms.put("_learn_rate", new Double[]{0.5});
+      gs = GridSearch.startGridSearch(grid1._key, params, hyperParms, simpleParametersBuilderFactory, hyperSpaceSearchCriteria, 6);
+      Scope.track_generic(gs);
+      final Grid grid = gs.get();
+      Scope.track_generic(grid);
+
+
+      int expectedNumberOfModelsFromTwoGrids = 10;
+      assertEquals(expectedNumberOfModelsFromTwoGrids, grid.getModelCount());
+
+    } finally {
+      Scope.exit();
+    }
+  }
+
+
+  @Test
+  public void test_parallel_random_search_with_custom_max_models_multirun() {
+    try {
+      Scope.enter();
+      final Frame trainingFrame = parse_test_file("smalldata/iris/iris_train.csv");
+      Scope.track(trainingFrame);
+
+      // Setup random hyperparameter search space
+      HashMap<String, Object[]> hyperParms = new HashMap<String, Object[]>() {{
+        put("_distribution", new DistributionFamily[]{DistributionFamily.multinomial});
+        put("_ntrees", new Integer[]{5});
+        put("_max_depth", new Integer[]{2});
+        put("_min_rows", new Integer[]{10,11,12,13,14});
+        put("_learn_rate", new Double[]{.7});
+      }};
+
+      GBMModel.GBMParameters params = new GBMModel.GBMParameters();
+      params._train = trainingFrame._key;
+      params._response_column = "species";
+
+      GridSearch.SimpleParametersBuilderFactory simpleParametersBuilderFactory = new GridSearch.SimpleParametersBuilderFactory();
+      HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria hyperSpaceSearchCriteria = new HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria();
+      int custom_max_model = 3;
+      hyperSpaceSearchCriteria.set_max_models(custom_max_model);
+
+      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms, simpleParametersBuilderFactory, hyperSpaceSearchCriteria, 2);
+      Scope.track_generic(gs);
+      final Grid grid1 = gs.get();
+      Scope.track_generic(grid1);
+
+      assertEquals(custom_max_model, grid1.getModelCount());
+
+      // Train a grid with new hyper parameters
+      hyperParms.put("_learn_rate", new Double[]{0.5});
+      gs = GridSearch.startGridSearch(grid1._key, params, hyperParms, simpleParametersBuilderFactory, hyperSpaceSearchCriteria, 2);
+      Scope.track_generic(gs);
+      final Grid grid = gs.get();
+      Scope.track_generic(grid);
+
+      int expectedNumberOfModelsFromTwoGrids = 2*custom_max_model;
+      assertEquals(expectedNumberOfModelsFromTwoGrids, grid.getModelCount());
+
     } finally {
       Scope.exit();
     }
--- a/h2o-algos/src/test/java/hex/grid/GridTest.java
+++ b/h2o-algos/src/test/java/hex/grid/GridTest.java
@@ -4,7 +4,6 @@ import hex.Model;
 import hex.genmodel.utils.DistributionFamily;
 import hex.tree.gbm.GBMModel;
 import org.junit.BeforeClass;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -16,6 +15,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.HashMap;
 
 import static org.junit.Assert.*;
@@ -27,7 +27,7 @@ public class GridTest extends TestUtil {
 
   @BeforeClass
   public static void setUp() {
-    stall_till_cloudsize(1);
+    stall_till_cloudsize(2);
   }
 
   @Test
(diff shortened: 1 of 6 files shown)


h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -177,9 +177,8 @@ private void gridSearch(Grid<MP> grid) {
     try {
       // Get iterator to traverse hyper space
       HyperSpaceWalker.HyperSpaceIterator<MP> it = _hyperSpaceWalker.iterator();
-      // Number of traversed model parameters
-      int counter = grid.getModelCount();
-      while (it.hasNext(model)) {
+      int numberOfBuiltModels = grid.getModelCount();
+      while (it.hasNext(model) && (it.max_models() == 0 || numberOfBuiltModels < it.max_models())) {
```
honzasterba: is this a breaking change? for cartesian walker this check did not used to be here

deil87:
@Pscheidl could you please take a look at failing test http://mr-0xc1:8080/job/h2o-3-pipeline/job/PR-4114/2/testReport/junit/hex.glrm/GLRMGridTest/Java_8_JUnit___testGridAppend/

It fails due to an assumption that grid could be reused. I'm not sure whether we are allowed to do so. If yes then we need to make sure that all early stopping related properties are reseted somehow.  

For example in the mentioned above test if you use Cartesian walker then `max_models` will be set to the size of hyper space. Once grid is traversed ( 4  items returned) grid will not allow to evaluate another grid of size 2 as it has remembered that `max_models` had been already reached. We need to set it to 6 ( for that we need another constructor where you can specify max_models for search criteria)

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @honzasterba for Cartesian max_models is set to a size of the hyperSpace https://github.com/h2oai/h2o-3/blob/39ace2cf92ea362f52cf13ad3d82cfab762e9c6c/h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java#L400

h2o-algos/src/test/java/hex/grid/GridTest.java
```diff
@@ -108,7 +221,10 @@ public void testFailedParamsRetention() {
       params._train = trainingFrame._key;
       params._response_column = "species";
 
-      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms);
+      GridSearch.SimpleParametersBuilderFactory simpleParametersBuilderFactory = new GridSearch.SimpleParametersBuilderFactory();
+      HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria hyperSpaceSearchCriteria = new HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria();
```
deil87: @Pscheidl I did this to kind of fix the test. With `RandomDiscreteValueSearchCriteria` we have default `max_models=0`(means no early stopping) whereas with `CartesianSearchCriteria` we have it set to hyperSpaceSize size by default ( and this would prevent us from evaluating extra grid items)

deil87:
@seb-h2o with these PR's changes `AutoMLTest.testTestAlgosHaveDefaultParametersEnforcingReproducibility` would also fail as we are also reusing same key for the three DL grids. The reason is that `maxAssignedModels` is being computed to `1` and due to re-usage of the grid early stopping would not allowed to compute more then one model( two last grids will be skipped). So are we ok to provide different keys for DL grids in `AutiML.defaultSearchDL` method? Before this change each grid had its own  counter for number of build models and early stopping logic was not taking into account what was happening in grids with the same key.

Pscheidl:
> @Pscheidl could you please take a look at failing test http://mr-0xc1:8080/job/h2o-3-pipeline/job/PR-4114/2/testReport/junit/hex.glrm/GLRMGridTest/Java_8_JUnit___testGridAppend/
> 
> It fails due to an assumption that grid could be reused. I'm not sure whether we are allowed to do so. If yes then we need to make sure that all early stopping related properties are reseted somehow.
> 
> For example in the mentioned above test if you use Cartesian walker then `max_models` will be set to the size of hyper space. Once grid is traversed ( 4 items returned) grid will not allow to evaluate another grid of size 2 as it has remembered that `max_models` had been already reached. We need to set it to 6 ( for that we need another constructor where you can specify max_models for search criteria)

As far as I know, the user can specify grid ID/name and then, when the ID is set to be the same for n > 2 grid search runs, the next run will just add more models to the grid, if any. There are even checks for the model already being present in the grid, which is useful not only during single grid run, but among multiple runs. This means what you said is true, we need to count with multiple GS runs, resultin in actions you mentioned (resetting parameters etc).

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: May I suggest introducing new method, called e.g. `shouldEarlyStop()`, which would encapsulate early stopping for all the implementations, forcing developers to move the logic into one place ?

deil87:
@Pscheidl 
> As far as I know, the user can specify grid ID/name and then, when the ID is set to be the same for n > 2 grid search runs, the next run will just add more models to the grid, if any. There are even checks for the model already being present in the grid, which is useful not only during single grid run, but among multiple runs. This means what you said is true, we need to count with multiple GS runs, resultin in actions you mentioned (resetting parameters etc).

yes, I also got some evidences that it might be a normal use case. I'm thinking maybe about increasing `max_models` by new `max_models` in case we already have grid with such key.

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @Pscheidl i think that the idea of this PR is to move this responsibility outside of our walkers implementations. For `GridSearch` class we don't have hierarchy yet and there is no different implementations in that sense. Are you suggesting to introduce GridSearch interface? 
 I was considering to introduce this method but just to highlight where we do early stopping.

Pscheidl: No need to do this, I was thinking out loud. Let's skip it in this PR then, as it would only make things more difficult.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -231,13 +230,12 @@ private void gridSearch(Grid<MP> grid) {
...
             }
-            
+            numberOfBuiltModels--;
             grid.appendFailedModelParameters(model != null ? model._key : null, params, e);
           }
         } catch (IllegalArgumentException e) {
           Log.warn("Grid search: construction of model parameters failed! Exception: ", e);
-          // Model parameters cannot be constructed for some reason
-          it.modelFailed(model);
+          numberOfBuiltModels--;
```
seb-h2o: this legacy `catch (IllegalArgumentException e)` looks so nasty with the nested `catch (RuntimeException e)` inside.
I'm sure its scope should have been limited to `params = it.nextModelParameters(model);`, avoiding you to have to handle `numberOfBuiltModels--` multiple times.
In worst case scenario — let's imagine  `grid.appendFailedModelParameters` throws an `IllegalArgumentException` — the counter will be decremented twice.

seb-h2o:
@deil87  do you really want to put this on `rel-yau`?
`Need for this changes has arisen from within #4040` : this latter one is for `master`.
In master, you would not have the key issue with DL in AutoML.

I would reserve those changes only for `master` now that `rel-yau` is mainly focusing on security fixes + `Booking` features. cc: @michalkurka

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @seb-h2o I just don't want to expand a scope of the PR.  I wanted to focus on the issue/fix that will let me to proceed with  https://github.com/h2oai/h2o-3/pull/4040 . I can create jira for refactoring this later.

deil87:
@seb-h2o  
> do you really want to put this on rel-yau?

I was considering it as a bug, therefore have chosen rel-yau as a target. Once master has a fix and `rel-yau` is occupied with a security theme it makes sense to switch to a `master`. Thx!

h2o-core/src/main/java/hex/grid/GridSearch.java
seb-h2o: Please do, thanks.

deil87: @seb-h2o I can put `numberOfBuiltModels--` as a last line after `appendFailedModelParameters` just for now.

deil87: @seb-h2o https://0xdata.atlassian.net/browse/PUBDEV-7113

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -18,13 +19,31 @@
...
   }
 
   private final transient ParallelModelBuilderCallback _callback;
-  private final transient IcedAtomicInt _modelInProgressCounter = new IcedAtomicInt();
+
+  public AtomicInteger getModelInProgressCounterNative() {
+    return _modelInProgressCounterNative;
+  }
+
+  //  private final transient IcedAtomicInt _modelInProgressCounter = new IcedAtomicInt(); // TODO remove after tests. Should be fine not to use Iced version as ParallelModelBuilder itself is not going to be distributed
```
deil87: will fix it after CI tests

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -18,13 +19,31 @@
 
   public static abstract class ParallelModelBuilderCallback<D extends ParallelModelBuilderCallback> extends Iced<D> {
 
-    public abstract void onBuildSucces(final Model model, final ParallelModelBuilder parallelModelBuilder);
+    public abstract void onBuildSuccess(final Model model, final ParallelModelBuilder parallelModelBuilder);
```
deil87: fixed type

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -714,10 +737,6 @@ public MP build() {
     }
   }
 
-  /**
-   * Constant for adaptive parallelism level - number of models built in parallel is decided by H2O.
-   */
-  public static final int ADAPTIVE_PARALLELISM_LEVEL = 0;
```
deil87: have not found usages for this constant

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -57,12 +57,6 @@
      */
     double time_remaining_secs();
 
-    /**
-     * Inform the Iterator that a model build failed in case it needs to adjust its internal state.
-     * @param failedModel
-     */
-    void modelFailed(Model failedModel);
```
deil87: it is package private. As we moved  early stopping logic outside of the iterator we don't do anything in `modelFailed` method.

h2o-algos/src/test/java/hex/grid/GridTest.java
```diff
@@ -24,9 +25,200 @@
   @Rule
   public TemporaryFolder temporaryFolder = new TemporaryFolder();
 
-  @Before
-  public void setUp() {
-    TestUtil.stall_till_cloudsize(1);
+  @BeforeClass
+  public static void setUp() {
+    stall_till_cloudsize(2);
```
deil87: will fix 2->1

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -18,13 +19,29 @@
...
+
+    public void postBuildSuccess(final Model model, final ParallelModelBuilder parallelModelBuilder) {};
 
     public abstract void onBuildFailure(final ModelBuildFailure modelBuildFailure, final ParallelModelBuilder parallelModelBuilder);
+
+    public void postBuildFailure(final ModelBuildFailure modelBuildFailure, final ParallelModelBuilder parallelModelBuilder) {};
```
seb-h2o: do you really need both `postBuildSuccess` and `postBuildFailure`?
Given how they're used in `onModelSuccess` and `onModelFailure`, it looks like they actually have the same goal.
Especially, I don't understand why the callback would want to process the `failure` multiple times, once with `onBuildFailure` and another time with `postBuildFailure`.

I would suggest to merge these 2 `post___` methods to a single `afterCompletion`:
```
public void afterCompletion(final Optional<Model> model, final ParallelModelBuilder parallelModelBuilder);
```

seb-h2o: (also, I would avoid `post` prefix, at first I thought it was posting something)

deil87: @seb-h2o this is a quite common naming for a lifecycle kind of things. Let's say in Akka https://alvinalexander.com/scala/understand-methods-akka-actors-scala-lifecycle.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -205,23 +208,39 @@ public void onBuildFailure(final ParallelModelBuilder.ModelBuildFailure modelBui
...
+        int numberOfCompletedModels = parallelModelBuilder.getModelCompletedCounter().get();
 
-        parallelModelBuilder.run(Collections.singletonList(ModelBuilder.make(nextModelParams)));
-      } else {
-        parallelModelBuilder.noMoreModels();
+        if (hyperspaceIterator.hasNext(previousModel)
+                && isThereEnoughTime()
+                && (hyperspaceIterator.max_models() == 0 || numberOfCompletedModels + numberOfExpectedToCompleteModels < hyperspaceIterator.max_models())
+                && !_job.stop_requested()
+                && (previousModel == null || !_hyperSpaceWalker.stopEarly(previousModel, grid.getScoringInfos()))) {
```
seb-h2o: can't the `stopEarly` method handle the null case?

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -205,23 +208,39 @@ public void onBuildFailure(final ParallelModelBuilder.ModelBuildFailure modelBui
...
+      synchronized (hyperspaceIterator) {
+        int numberOfExpectedToCompleteModels = parallelModelBuilder.getModelInProgressCounter().get();
+        int numberOfCompletedModels = parallelModelBuilder.getModelCompletedCounter().get();
 
-        parallelModelBuilder.run(Collections.singletonList(ModelBuilder.make(nextModelParams)));
-      } else {
-        parallelModelBuilder.noMoreModels();
+        if (hyperspaceIterator.hasNext(previousModel)
+                && isThereEnoughTime()
+                && (hyperspaceIterator.max_models() == 0 || numberOfCompletedModels + numberOfExpectedToCompleteModels < hyperspaceIterator.max_models())
```
seb-h2o: can you please extract this long condition to a variable like `hasReachedmaxModels`?
There's a lot of conditions here, so it would be nice if it were easy to read:
```
hasNext() 
  && hasEnoughTime() 
  && !hasReachedMaxModels
  && !stopRequest() 
  && !stopEarly()
```

h2o-core/src/main/java/hex/ParallelModelBuilder.java
deil87: @seb-h2o 
> do you really need both postBuildSuccess and postBuildFailure? Given how they're used in onModelSuccess and onModelFailure, it looks like they actually have the same goal.

Not exact the same. Original implementation was dong two things in `onModelSuccess` handler: processing success and making attempt to ask for next model parameters. 

So what I wanted to achieve is to be able to update counters in between of these two actions because I need them to be updated before making decision about whether we want to return next model parameters ( max_models early stopping relies on these counters). There are other ways of how to achieve same result.... but it would also require changes in signatures.

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @seb-h2o will try to improve

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -205,23 +208,39 @@ public void onBuildFailure(final ParallelModelBuilder.ModelBuildFailure modelBui
...
-      final MP nextModelParams = getNextModelParams(hyperspaceIterator, previousModel, grid);
-      if (nextModelParams != null
-              && isThereEnoughTime()
-              && !_job.stop_requested()
-              && !_hyperSpaceWalker.stopEarly(previousModel, grid.getScoringInfos())) {
 
-        // FIXME: ModelFeeder's <MP> type parameter is hiding GridSearch's one. Ideally would have moved ModelFeeder outside but it is too coupled with GridSearch class.
-        reconcileMaxRuntime(grid._key, hyperspaceIterator, nextModelParams);
+      synchronized (hyperspaceIterator) {
+        int numberOfExpectedToCompleteModels = parallelModelBuilder.getModelInProgressCounter().get();
```
seb-h2o: please keep the `inProgress` name here. This just adds confusion as we don't know when those models are expected to complete... like total of expected models to complete? current expected models to complete? especially hen reading the boolean logic below.

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @seb-h2o this `previousModel == null` covers case when we have failed to evaluate particular model parameters and we can't make new decision about early stopping untill we get new data ( new models). Not sure if this logic could be moved inside `stopEarly`. Need to think about it.

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @seb-h2o I see your point. Agree

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @seb-h2o done

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @seb-h2o decided to move null case inside `stopEarly`

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @seb-h2o fixed

h2o-core/src/main/java/hex/ParallelModelBuilder.java
seb-h2o: > what I wanted to achieve is to be able to update counters in between of these two actions
I understand why you need both `onBuildSuccess` AND `postBuildSuccess`.
But I think you could merge both `postBuildSuccess` AND `postBuildFailure` into a more generic `afterCompletion`.

deil87: @seb-h2o ah, sorry, my bad. Misunderstood your message.  

`postBuildSuccess` and `postBuildFailure` have different signatures: one has Model parameter and another one has ModelBuilderFailure parameter. Don't think it makes sense to mix these cases together. Having two separate `post...` methods imo also good for a symmetry (otherwise for consistency we could also want generic `onCompletion` method).

seb-h2o: not convinced that symmetry is good here: as mentioned above, I think that errors (here as wrapped in `ModelBuildFailure`) should be handled only once, i.e. only in `onBuildFailure`, not in `postBuildFailure`, that's why I was suggesting a common `post___` method.
Anyway, don't want to block you more, although this are still potentially public hooks.

deil87:
@seb-h2o  It is not just about code symmetry.  It is about logic symmetry(similar to build success execution path) - we need to update counters after handling of the failure and before attempting next model. I have double checked this with my test right now `GridTest.test_parallel_grid_search_MaxModelIsSatisfiedWhenSomeModelsHaveFailed` and it would fail from time to time (if run N times) without `post____` approach. Thank you for your review!

seb-h2o:
@deil87 I don't think you understood my comment correctly, it was a follow-up to https://github.com/h2oai/h2o-3/pull/4114/files#r363393662

your default implementation of `postBuildFailure` is ignoring the failure, so I don't see how the presence of the `ModelBuildFailure` parameter can have any impact on the logic.
I just don't think it's a good idea to offer the possibility to handle an exception (the cause attribute of `ModelBuildFailure`) AFTER is was already handled (by `onBuildFailure`), that's all. And actually you're not doing anything with it, that's why I was originally suggesting to pass `Optional<Model>` instead and use the same post-processing method for both success and failure.

deil87:
@seb-h2o sorry,  I thought we agreed on "different signatures" argument, but apparently we were not :) Although I would prefer symmetry in that case I see benefits in your suggestion as well ( getting rid of unused exception in this particular implementation). Switched to your version, thanks! Waiting for tests before merging.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -464,7 +453,7 @@ public RandomDiscreteValueWalker(MP params,
     /** Based on the last model, the given array of ScoringInfo, and our stopping criteria should we stop early? */
     @Override
     public boolean stopEarly(Model model, ScoringInfo[] sk) {
-      return ScoreKeeper.stopEarly(ScoringInfo.scoreKeepers(sk),
+      return model !=null && ScoreKeeper.stopEarly(ScoringInfo.scoreKeepers(sk),
```
honzasterba: broken formatting

h2o-core/src/main/java/hex/grid/GridSearch.java
honzasterba: should be used in getParallelismLevel below

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -349,7 +372,7 @@ private void gridSearch(Grid<MP> grid) {
             scoringInfo.time_stamp_ms = System.currentTimeMillis();
 
             //// build the model!
-            model = buildModel(params, grid, ++counter, protoModelKey);
+            model = buildModel(params, grid, ++numberOfBuiltModels, protoModelKey);
```
honzasterba: should we increase numberOfBuiltModels only if a model was actually built?

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -205,26 +208,40 @@ public void onBuildFailure(final ParallelModelBuilder.ModelBuildFailure modelBui
...
+                && !hasReachedMaxModels(parallelModelBuilder)
+                && !_job.stop_requested()
+                && !_hyperSpaceWalker.stopEarly(previousModel, grid.getScoringInfos())) {
 
-        parallelModelBuilder.run(Collections.singletonList(ModelBuilder.make(nextModelParams)));
-      } else {
-        parallelModelBuilder.noMoreModels();
+          final MP nextModelParams = getNextModelParams(hyperspaceIterator, previousModel, grid);
+
+          // FIXME: ModelFeeder's <MP> type parameter is hiding GridSearch's one. Ideally would have moved ModelFeeder outside but it is too coupled with GridSearch class.
```
honzasterba: arent those the same?

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -94,7 +97,8 @@ private GridSearch(Key<Grid> gkey, HyperSpaceWalker<MP, ?> hyperSpaceWalker, fin
     // this function the grid object is in DKV and accessible.
     final Grid<MP> grid;
     Keyed keyed = DKV.getGet(_result);
-    if (keyed != null) {
+
+    if (_isMultiRunSearch = keyed != null) {
```
honzasterba: hidden assignment in if statemenet

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: will fix

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @honzasterba this is a valid usage of Java and I have seen examples in our codebase. I'm ok to change this but  I would like to ask team's opinion on this...whether we are ok with this syntax? Couple of times I run into situation when reviewers were suggesting opposite things about code conventions. I would prefer to discuss such cases and stick to our team's conventions. @michalkurka @seb-h2o @Pscheidl ( like if you are ok with the syntax and dislike if it is better to avoid it)

PR #4115, opened 2019-12-02
PUBDEV-7066: expose mae, rmsle, aucpr to GridSearch in Py client

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-7066

Added clients missing supports for MAE, RMSLE, AUCPR.

For consistency with stopping metric name, deprecated `pr_auc` functions on client side and forwarded implementation to `aucpr` equivalent.

On Python, improved generic support for deprecations and aliases.


h2o-py/h2o/backend/connection.py
```diff
@@ -191,7 +191,42 @@ def url(self):
...
+        shutdown_server=lambda self, prompt=True: self.cluster.shutdown(prompt),
+        make_url=lambda self, url_suffix, _rest_version=3: "/".join([self._base_url, str(_rest_version), url_suffix]),
+        get=lambda *args, **kwargs: _deprecated_get(*args, **kwargs),
+        post=lambda *args, **kwargs: _deprecated_post(*args, **kwargs),
+        delete=lambda *args, **kwargs: _deprecated_delete(*args, **kwargs),
+        get_json=lambda *args, **kwargs: _deprecated_get(*args, **kwargs),
+        post_json=lambda *args, **kwargs: _deprecated_post(*args, **kwargs),
+    )
+)
+class H2OConnection(h2o_meta()):
```
seb-h2o: I was also considering simply deleting all those old backwards incompatibilities:
`These methods are deprecated since July 2016. Please remove them if it's 2017 already...`
But we're still using some of those internally....
So will create a separate ticket for this.

h2o-py/tests/testdir_misc/pyunit_metric_accessors.py
```diff
@@ -108,6 +113,16 @@ def metric_accessors():
...
+    # using list from http://docs.h2o.ai/h2o/latest-stable/h2o-docs/performance-and-prediction.html#classification
+    # + common ones
+    for metric in ['gini', 'logloss', 'auc', 'aucpr', 'mse', 'rmse']:
+        val = getattr(gbm, metric)()
+        assert isinstance(val, float), "expected a float for metric {} but got {}".format(metric, val)
+
+    for metric in ['mcc', 'F1', 'F0point5', 'F2', 'accuracy', 'mean_per_class_error']:
+        val = getattr(gbm, metric)()[0][1]
+        assert isinstance(val, float), "expected a float for metric {} but got {}".format(metric, val)
+
```
seb-h2o: please admire the consistency of our metrics accessors...
Please +1 if I should create a JIRA for this, knowing that it will be difficult to make this backwards compatible.

h2o-py/tests/testdir_misc/pyunit_metric_accessors.py
```diff
@@ -420,9 +436,19 @@ def metric_accessors():
...
     predictors = ["displacement","power","weight","acceleration","year"]
-    gbm.distribution="multinomial"
+    gbm = H2OGradientBoostingEstimator(nfolds=3, distribution=distribution, fold_assignment="Random")
     gbm.train(x=predictors,y=response_col, training_frame=train, validation_frame=valid)
 
+    # using list from http://docs.h2o.ai/h2o/latest-stable/h2o-docs/performance-and-prediction.html#classification
+    # + common ones
+    for metric in ['logloss', 'mse', 'rmse', 'mean_per_class_error']:
+        val = getattr(gbm, metric)()
+        assert isinstance(val, float), "expected a float for metric {} but got {}".format(metric, val)
```
seb-h2o: please again admire how `model.mean_per_class_error()` returns a `float` for `multinomial`, but an array for `binomial`...

h2o-py/h2o/grid/grid_search.py
```diff
@@ -694,11 +706,33 @@ def gini(self, train=False, valid=False, xval=False):
...
+        :param bool valid: If valid is True, then return the aucpr value for the validation data.
+        :param bool xval:  If xval is True, then return the aucpr value for the validation data.
+
+        :returns: The AUCPR for the models in this grid.
+        """
+        return {model.model_id: model.aucpr(train, valid, xval) for model in self.models}
+
+    @deprecated(replaced_by=aucpr)
+    def pr_auc(self):
+        pass
```
seb-h2o: after discussion with @ledell , we decided to deprecate legacy accessor for `pr_auc` and make `aucpr` the new standard on client side.
Alternative (also now available), was to make it an alias.

h2o-py/tests/testdir_misc/pyunit_metric_accessors.py
wendycwong: This is an error, right?  You can actually get away with one value for binomial but for multinomial, it must be an array.  Please write a JIRA for this when you have a chance.

h2o-py/h2o/grid/grid_search.py
```diff
@@ -694,11 +706,33 @@ def gini(self, train=False, valid=False, xval=False):
...
+
```
wendycwong: seb:  The giniCoef() is supposed to be deprecated as well.  The new one is gini().  However, it was not done properly and is not working.  Would it be too painful for you to fix it as well?

wendycwong: Seb:

I gave up.  I checked R, the current version is h2o.giniCoef() while python wants gini().  Please ignore what I said before.  I will write a new JIRA for the gini.

h2o-py/h2o/backend/connection.py
michalkurka: thanks

h2o-py/h2o/grid/grid_search.py
michalkurka: @seb-h2o can you please run this by @NidhiMehta so that we know there is no issue for users? there were bunch of issues/confusion with aucpr recently

h2o-py/tests/testdir_misc/pyunit_metric_accessors.py
michalkurka: > please again admire how model.mean_per_class_error() returns a float for multinomial, but an array for binomial...

really? how does it return a single number for multinomial? what is that number?

michalkurka:
LGTM - we should just double-check with the support team that this is okay to do

h2o-py/h2o/backend/connection.py
seb-h2o: https://0xdata.atlassian.net/browse/PUBDEV-7114

h2o-py/h2o/grid/grid_search.py
seb-h2o: will tell her: note that the `@deprecated` decorator, leaves the `pr_auc` method visible and functional.
It just raises a Py warning on call and then forwards to the `replaced_by` function if provided, so there shouldn't be any backwards incompatibilities.

h2o-py/tests/testdir_misc/pyunit_metric_accessors.py
seb-h2o: @wendycwong, @michalkurka: there are various inconsistencies here:
- "single number for multinomial" is fine, as the metric is `mean_per_class_error`, not `mean_error_per_class`, a single float is what is expected here.
- although `mean_per_class_error` returns an array for `binomial` because there, the metric methods also support a `threshold` and we always return an array of [threshold, value] even if no threshold or a single threshold was provided, making the result at the same time unexpected but making sense (it also provides the threshold for which the metric is computed).
- then there is this inconsistency among binomial metrics themselves: the ones computed globally (auc, aucpr, logloss, gini...) return a single value, but the ones that accept an "optional" threshold (F1, accuracy, mcc, mean_per_class_error...) return an array or arrays as explained above.

This is not completely wrong or buggy, it is just that the API is not very user-friendly...

seb-h2o: I can write a ticket to improve the usability here., but I don't think this can be done without sacrificing a minimum backward compatibility.

wendycwong: seb:  I wrote a JIRA for fixing the Gini: https://0xdata.atlassian.net/browse/PUBDEV-6888.

Thank you in advance!

h2o-py/h2o/grid/grid_search.py
wendycwong: Here is the JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6888

seb-h2o:
all green tells me that I should merge...

PR #4119, opened 2019-12-03
[PUBDEV-7121] Parallel grid search does not handle max_runtime_secs == 0 for unlimited search

deil87:
https://0xdata.atlassian.net/browse/PUBDEV-7121

Target branch is `master` as we now are not allowed to merge to rel-yau

### DIFF
--- a/h2o-algos/src/test/java/hex/grid/GridTest.java
+++ b/h2o-algos/src/test/java/hex/grid/GridTest.java
@@ -9,6 +9,7 @@ import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 import water.*;
 import water.fvec.Frame;
+import water.util.Log;
 
 import java.io.*;
 import java.nio.file.Files;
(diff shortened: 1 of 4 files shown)


h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -231,7 +235,7 @@ private void constructScoringInfo(final Model model) {
     }
 
     private boolean isThereEnoughTime() {
-      final boolean enoughTime = hyperspaceIterator.max_runtime_secs() > 0 && hyperspaceIterator.time_remaining_secs() > 0;
+      final boolean enoughTime = hyperspaceIterator.max_runtime_secs() == 0 || ( hyperspaceIterator.max_runtime_secs() > 0 && hyperspaceIterator.time_remaining_secs() > 0);
```
deil87: this is where the main issue were - missing == 0 case

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -270,7 +274,9 @@ private void parallelGridSearch(final Grid<MP> grid) {
     final ParallelModelBuilder parallelModelBuilder = new ParallelModelBuilder(modelFeeder);
 
     List<ModelBuilder> startModels = new ArrayList<>();
-    final List<MP> mps = initialModelParameters(_parallelism, iterator);
+
+    int numberOfInitialModels = iterator.max_models() == 0 ? _parallelism : Math.min(_parallelism, iterator.max_models());
```
deil87: also had to take into account case where `_parallelism` is greater than `max_models()`

h2o-algos/src/test/java/hex/grid/GridTest.java
```diff
@@ -59,12 +61,14 @@ public void testParallelModelTimeConstraint() {
               hyperParms,
               new GridSearch.SimpleParametersBuilderFactory(),
               new HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria(), 2);
+
       Scope.track_generic(gridSearch);
       final Grid grid = gridSearch.get();
       Scope.track_generic(grid);
 
+      // According to javadoc of `RandomDiscreteValueSearchCriteria.max_runtime_secs`  each model will have one second to finish ( due to params._max_runtime_secs = 1D;) and there is no time limit for Grid
+      // It means if it is enough to finish one model with one second then we should expect whole space to be trained => assertion is probably incorrect
```
seb-h2o: `params._max_runtime_secs = 1D;`
please use `1d`,  it took me 5 looks to read `1D` instead of `10`.

deil87: @seb-h2o it is not my code.

seb-h2o: are you suggesting that the `assertNotEquals` below should be `assertEquals` and that's the reason you want to `@Ignore` the test?
If so, I agree with you, looks like it should be `assertEquals` and the test should be reenabled.

seb-h2o: @Pscheidl ?

deil87: @seb-h2o to be consistent with my current understanding of how max_runtime should be managed....yes, it should be `assertEquals`. But probably original intention of the author was different - I assume author wanted to show that 1 sec would not be enough to build all models

seb-h2o: > @seb-h2o it is not my code.
I know, but as you have the occasion, will change it... it's very minor but a bit disturbing

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -346,18 +342,7 @@ private void gridSearch(Grid<MP> grid) {
           // Sequential model building, should never propagate
           // exception up, just mark combination of model parameters as wrong
 
-          // Do we need to limit the model build time?
```
deil87: @seb-h2o please pay attention that logic was changed. It was not just moving.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -406,6 +391,32 @@ private void gridSearch(Grid<MP> grid) {
...
+    if (params._max_runtime_secs > 0 ) {
+      double was = params._max_runtime_secs;
+      params._max_runtime_secs = Math.min(params._max_runtime_secs, time_remaining_secs);
+      Log.info("Due to the grid time limit, changing model max runtime from: " + was + " secs to: " + params._max_runtime_secs + " secs.");
+    } else { // params._max_runtime_secs == 0
+      params._max_runtime_secs = time_remaining_secs;
+      Log.info("Due to the grid time limit, changing model max runtime to: " + params._max_runtime_secs + " secs.");
+    }
+  }
+
```
seb-h2o: re-re-reviewed and approved.

legacy `if (max_runtime_secs > 0) {` before checking the `params._max_runtime_secs` was preventing a correct application of the rule, thanks for the fix.
Must the reason why the test above was passing when it shouldn't have.

h2o-algos/src/test/java/hex/grid/GridTest.java
seb-h2o: Please fix the test instead of ignoring it, the fix should be part of the PR.

deil87: @seb-h2o sure, i just wanted to discuss it and then fix

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @seb-h2o 
> Must be the reason why the test above was passing when it shouldn't have

i think yes

h2o-algos/src/test/java/hex/grid/GridTest.java
Pscheidl: The intention was to test whether max runtime will be respected, the training will be stopped and the whole hyperparam space not walked.

Could also be tested by <`grid.modelCount`, instead of notEquals.

Pscheidl: The second issue is something I can't help with. On Linux, my IntelliJ IDEA uses DejaVu Sans Mono, which can properly distinguish among these characters:

Screenshot from 2019-12-08 21-40-47

If these basic fonts are not available to you on your Apple device, this one might help you with character readability: https://mplus-fonts.osdn.jp/about-en.html

It's an excellent choice.

seb-h2o: just for @Pscheidl 
without the comparison above, I quickly read 10, not 1D, I guess it's because I'm used, like most people, to see a digit after another digit if they're not separated by a space...), so font it not the only concern.
Also, you know as much as I do that  in Java the literal suffix is optional for doubles and ints, so no need to specify it if it can lead to confusion.
Besides, again to improve readability, for those suffixes, it's recommended to use the case that leads to less confusion:
- 111L, instead of 111l
- 555f and 555F both work but I think 1st version is more common
- 100d instead of 100D
when you read numbers you haven't written yourself, you just spend a fraction of a second reading those, so I think it's important to make them as readable as possible.

h2o-algos/src/test/java/hex/grid/GridTest.java
```diff
@@ -4,11 +4,13 @@
 import hex.genmodel.utils.DistributionFamily;
 import hex.tree.gbm.GBMModel;
 import org.junit.Before;
+import org.junit.Ignore;
```
michalkurka: Appears unused. Which is a good thing :)

michalkurka:
@deil87 @Pscheidl @seb-h2o Thanks everyone!

h2o-algos/src/test/java/hex/grid/GridTest.java
deil87: @michalkurka fixed

PR #4120, opened 2019-12-04
PUBDEV-6796 - Spearman's correlation

Pscheidl:


h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
```diff
@@ -0,0 +1,125 @@
...
+    // Sort by Y
+    final AstRankWithinGroupBy.SortnGrouby sortTaskY = new AstRankWithinGroupBy.SortnGrouby(rankTaskX._finalResult, new int[]{},
+            new int[]{vecIdY}, new int[]{1}, "rankY");
+    sortTaskY.doAll(sortTaskY._groupedSortedOut);
+
+    AstRankWithinGroupBy.RankGroups rankTaskY = new AstRankWithinGroupBy.RankGroups(sortTaskY._groupedSortedOut, sortTaskY._groupbyCols,
+            sortTaskY._sortCols, sortTaskY._chunkFirstG, sortTaskY._chunkLastG, sortTaskY._newRankCol)
+            .doAll(sortTaskY._groupedSortedOut);
+
+    final Vec rankX = rankTaskY._finalResult.vec("rankX");
```
Pscheidl: Currently, the ranks have two problems:

- Duplicated values are given different rank (they're sorted one after another correctly, but their rank is incremented)
- There are "gaps" and "holes" between the ranks. E.g. `1,2,5,6`.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
```diff
@@ -0,0 +1,125 @@
...
+
+public class AstSpearman extends AstPrimitive<AstSpearman> {
+  @Override
+  public int nargs() {
+    return 1 + 3; // Frame ID and ID of two numerical vectors to calculate SCC on.
+  }
+
+  @Override
+  public String[] args() {
+    return new String[0];
```
michalkurka: why is this empty?

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  @Override
+  public Val apply(Env env, Env.StackHelp stk, AstRoot[] asts) {
+    final Value value = DKV.get(asts[1].str());
```
michalkurka: you should do something like 

    asts[1].exec(env)).getFrame()

otherwise you would be working with an unevaluated Frame possibly

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
```diff
@@ -0,0 +1,125 @@
...
+    if (!value.isFrame()) {
+      throw new IllegalArgumentException(String.format("The given key '%s' is not a frame.", asts[1]));
+    }
+
+    final Frame originalUnsortedFrame = value.get(Frame.class);
+    final int vecIdX = originalUnsortedFrame.find(asts[2].str());
```
michalkurka: the values need to be evaluated

michalkurka:
@Pscheidl how does this work on categorical columns?

We can assume categorical columns already are ranked by the ordering of the domain

Pscheidl:
The POC did not really count with categoricals, but I've added support for them in https://github.com/h2oai/h2o-3/pull/4120/commits/a204876652b47a64c76dec6e25572f78e81a2df1. There is one downside - I test it on `iris` - correlation between type of the flower and `sepal_len`. And the ranking system makes so many "holes" the resulting pearson value is useless. When the ranking is corrected manually, it calculates the proper value.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
Pscheidl: Done.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
Pscheidl: Done.

Pscheidl: I accidentally deleted the comment about "evaluation" below.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
Pscheidl: Added param names.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
Pscheidl: When I'm thinking about it, the way we need duplicated values treated requires cross-chunk checking. For the computation form. There is alternative way, just by adjusting the equation, we might only need sorted values, not ranked.

Pscheidl:
Two observations:

- The Spearman correlation value have to be calculated manually, as most automated tools use the easier calculation formula, only providing approximation.
- Our current labeling used in this PR gives increasing labels to the same values in the dataset. This has to be fixed and is probably responsible for the small delta in case of Iris test.

Here is the spreadsheet I used for Iris: 

[iris_spearman.zip](https://github.com/h2oai/h2o-3/files/3932093/iris_spearman.zip)

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1747,6 +1747,11 @@ public Frame sort( int[] cols ) {
   public Frame sort(int[] cols, int[] ascending) {
     return Merge.sort(this, cols, ascending);
   }
+  
+  public void label(final String labelColumnName){
```
michalkurka: Such a method imho doesn't belong in the Frame class.

h2o-core/src/test/java/water/rapids/RapidsTest.java
```diff
@@ -28,18 +34,118 @@
...
+      assertEquals(-0.1720027734934786, pearson_sepal.getNum(), 1e-8);
+
+    } finally {
+      Scope.exit();
+    }
+  }
   
-  @Test public void bigSlice() {
+  @Test
+  public void testSort(){
```
Pscheidl: This test demonstrates why the existing sort is not usable for us - it doesn't keep the original order and was probably made for different purpose by @wendycwong.

We need the original dataframe (or a copy of it), but with the original order of rows, only labeled. This is mandatory for Spearman.

Pscheidl: We need two vectors from the original frame ranked, without affecting the order of rows.

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: Moved to FrameUtils, thanks.

Pscheidl:
Todo:
1. Add support for categorical columns
2. Add support for ordinals.
3. Add more test cases ?
4. Make the ranking process done in AstSpearman class distributed ?

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
```diff
@@ -0,0 +1,180 @@
...
+    final Vec yLabel = sortedY.vec("label") == null ? sortedY.vec(0) : sortedY.vec("label");
+    final Vec yValue = sortedY.vec(0);
+    Scope.track(xLabel);
+    Scope.track(yLabel);
+
+    double lastX = Double.NaN;
+    double lastY = Double.NaN;
+    long skippedX = 0;
+    long skippedY = 0;
+    for (int i = 0; i < orderX.length(); i++) {
```
Pscheidl: We can make this distributed, but it will take two passes instead of one according to my poor knowledge. 

Any advice @michalkurka please ? Thank you.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
michalkurka: Is this done? The code looks the same to me

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
michalkurka: Reading data this way is indeed very inefficient. Vec.Reader should be used instead at the bare minimum.

2-step MRTask seems like a good idea to me.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
```diff
@@ -0,0 +1,180 @@
...
+    final Value value = DKV.get(asts[1].str());
+    if (!value.isFrame()) {
+      throw new IllegalArgumentException(String.format("The given key '%s' is not a frame.", asts[1]));
+    }
+
+    final Frame originalUnsortedFrame = value.get(Frame.class);
+    final int vecIdX = originalUnsortedFrame.find(asts[2].exec(env).getStr());
+    final int vecIdY = originalUnsortedFrame.find(asts[3].exec(env).getStr());
+
+    Scope.enter();
```
michalkurka: Scope always needs to be used with try-finally

Pscheidl: Done. Moved the whole ranking logic to a separate method wrapped in try-finally block, so the scope is ensured to be exitted.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
Pscheidl: I used Vec Readers and Writers to do the bare minimum. Just in case.

I've been playing with the two-pass algorithm and found some obstacles in the rank calculation - I can't guarantee all the duplicates will be corrected properly in the second pass, unless some formula to calculate the offsets based on first pass (which is per-chunk sort). A trivial solution which would localize the computation is to do up to `n` passes, where `n` is the number of chunks.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
Pscheidl: It was not, now it is:
`final Frame originalUnsortedFrame = stk.track(asts[1].exec(env).getFrame());`

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -827,4 +829,80 @@ public void map(Chunk c, NewChunk nc) {
     }
   }
 
+  /**
+   * Calculates means of given numerical Vectors. Provided there is a NaN on a row in any of the give vectors,
```
michalkurka: What differentiates this method from say Vec.mean()?

Pscheidl: `Vec.mean` uses rollup stats.

```java
  /** Vecs's mean 
   *  @return Vec's mean */
  public double mean() { 
    return rollupStats()._mean; }
```

Those are not calculated for categorical columns. I'd have to duplicate the categorical vec and make it `T_NUM`, as we're unable to modify the original frame. Calculating it manually prevents us from copying the categorical data.

Also, if we'll ever implement the better version of Pearson (Perason can be implemented 100% without data copying in a very fast way), it will be reused there as well.

michalkurka: it seem the way this method works is very specific to your implementation (if there is NA in any item of a given row, the whole row is skipped) - should this be move closer to your implementation?

Pscheidl: Yes, why not. Consider it done. I'll wait for the tests before pushing again, as I tested it on a cluster of 5 locally, but want the latest results.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstSpearman.java
```diff
@@ -0,0 +1,230 @@
...
+      this._xyMul += mrt._xyMul;
+    }
+
+    @Override
+    protected void postGlobal() {
+      final double xStdDev = Math.sqrt(_xDiffSquared / _linesVisited);
+      final double yStdDev = Math.sqrt(_yDiffSquared / _linesVisited);
+
+      spearmanCorrelationCoefficient = (_xyMul - (_linesVisited * _xMean * _yMean))
+              / ((_linesVisited) * xStdDev * yStdDev);
```
Pscheidl: Note to self: Fix the forgotten brackets.

h2o-core/src/main/java/water/util/VecUtils.java
Pscheidl: Moved it to AstSpearman. If it will be reused elsewhere, I'll make the necessary modifications afterwards.

PR #4122, opened 2019-12-04
PUBDEV-7116: Python Docs: Add examples to PSVM

hannah-tillman:


h2o-py/h2o/estimators/psvm.py
```diff
@@ -311,6 +451,16 @@ def seed(self):
...
 
         Type: ``int``  (default: ``-1``).
+
+        :examples:
+
+        >>> splice = h2o.import_file("http://h2o-public-test-data.s3.amazonaws.com/smalldata/splice/splice.svm")
+        >>> svm = H2OSupportVectorMachineEstimator(gamma=0.1,
+        ...                                        rank_ratio=0.1,
+        ...                                        seed=1234,
+        ..                                        disable_training_metrics=False)
```
angela0xdata: missing .

angela0xdata:
So congrats on creating your first .py file!
I think this is good, but would like a dev set of eyes on this too.

PR #4124, opened 2019-12-05
PUBDEV-7074: AutoML leaderboard extensions

seb-h2o:
https://0xdata.atlassian.net/browse/PUBDEV-7074

This PR provides abstractions and a mechanism to extend the leaderboard on-demand.
The values for the extensions are lazy-loaded to avoid extra cost when creating the default leaderboard showing only metric scores.

The first extensions provided aim at model complexity assessment :
- `training_time_millis`
- `predict_time_per_row_millis` (as the value is a double usually < 0, we may want to consider exposing `predict_time_per_row_nanos` instead, although having the same unit for both columns is nice too...)
- `model_size_bytes` (logic not implemented yet, and won't be in this PR)

The PR also includes the following refactoring:
- moved all leaderboard components to `ai.h2o.automl.leaderboard` package.
- moved all eventLog components to `ai.h2o.automl.events` package for symmetry.
Those were not imported in Sparkling Water, so this doesn't break backward compatibility there.

Status:
- backend done.
- REST API done.
- Py+R integration done.
- Java tests.
- WONT DO HERE: Flow support...


h2o-automl/src/main/java/ai/h2o/automl/events/EventLogEntry.java
```diff
@@ -62,17 +63,17 @@ public Object parseObject(String source, ParsePosition pos) {
     }
   }
 
-  static final Format epochFormat = new SimpleFormat<Date>() {
+  public static final Format epochFormat = new SimpleFormat<Date>() {
```
michalkurka: is this okay to have it static?

I am refering to Format's javadoc:
``
 * Formats are generally not synchronized.
 * It is recommended to create separate format instances for each thread.
 * If multiple threads access a format concurrently, it must be synchronized
 * externally.
``

seb-h2o: @michalkurka Mmm, fair point! Never realized that about Format...
Will create a ticket to fix this: we haven't had any issue with this until now.

I think this could easily be fixed by just putting those in a ThreadLocal.

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/LeaderboardColumn.java
```diff
@@ -0,0 +1,16 @@
+package ai.h2o.automl.leaderboard;
+
+import hex.Model;
+import water.Freezable;
+import water.Key;
+
+public interface LeaderboardColumn<V, SELF extends LeaderboardColumn> extends Freezable<SELF> {
+    LeaderboardColumnDescriptor getDescriptor();
+
+    Key<Model> getModelId();
```
deil87: @seb-h2o not used

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/LeaderboardColumn.java
```diff
@@ -0,0 +1,16 @@
+package ai.h2o.automl.leaderboard;
+
+import hex.Model;
+import water.Freezable;
+import water.Key;
+
+public interface LeaderboardColumn<V, SELF extends LeaderboardColumn> extends Freezable<SELF> {
```
deil87: @seb-h2o don't see usages for SELF, planning to add something?

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/LeaderboardColumn.java
```diff
@@ -0,0 +1,16 @@
...
+    LeaderboardColumnDescriptor getDescriptor();
+
+    Key<Model> getModelId();
+    V getValue();
+    void setValue(V value);
+
+    default boolean isNA() { return getValue() == null; }
+    default V fetch() { return getValue(); }
```
deil87: @seb-h2o not clear when we should use `fetch` and when `getValue`. Maybe naming could be improved?

h2o-automl/src/test/java/ai/h2o/automl/AutoMLTest.java
```diff
@@ -138,6 +138,7 @@
       assertEquals("wrong amount of standard models", 3, count_non_se);
       assertEquals("wrong amount of SE models", 2, count_se);
       assertEquals(5, aml.leaderboard().getModelCount());
+      Log.info(aml.leaderboard().toTwoDimTable("training_time_millis", "predict_time_per_row_millis"));
```
deil87: @seb-h2o maybe enum for strings or marker interface with implementations in corresponding columns classes?

h2o-automl/src/main/java/water/automl/api/schemas3/EventLogV99.java
```diff
@@ -11,15 +12,18 @@
   @API(help="List of events produced during the AutoML run", direction=API.Direction.OUTPUT)
   public EventLogEntryV99[] events;
 
+  @API(help="A table representation of this event log, for easy rendering", direction=API.Direction.OUTPUT)
+  public TwoDimTableV3 table;
```
deil87: @seb-h2o does EventLogs have anything to do with leaderboard extensions?

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLV99.java
```diff
@@ -67,42 +67,42 @@ public AutoMLKeyV3(Key<AutoML> key) {
...
 
     // NOTE: don't return nulls; return an empty leaderboard/eventLog, to ease life for the client
     EventLog eventLog = autoML.eventLog();
     if (null == eventLog) {
       eventLog = new EventLog(autoML._key);
     }
-    this.event_log = new EventLogV99().fillFromImpl(eventLog);
-    this.event_log_table = new TwoDimTableV3().fillFromImpl(eventLog.toTwoDimTable());
+    event_log = new EventLogV99().fillFromImpl(eventLog);
+    event_log_table = event_log.table;  // for backwards compatibility
```
deil87: @seb-h2o does it mean that `event_log_table ` is is deprecated?

h2o-automl/src/main/java/water/automl/api/LeaderboardsHandler.java
```diff
@@ -42,30 +43,29 @@ public boolean filter(KeySnapshot.KeyInfo k) {
...
+      lb.table = new TwoDimTableV3().fillFromImpl(leaderboard.toTwoDimTable(input.extensions));
+    }
+    return lb;
   }
 
-  // TODO: almost identical to ModelsHandler; refactor
-  public static Leaderboard getFromDKV(String param_name, String key_str) {
-    return getFromDKV(param_name, Key.make(key_str));
+  private static Leaderboard getFromDKV(Key key) {
+      return getFromDKV(key, "(none)");
```
deil87: @seb-h2o looks like Java 8 Optional

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/LeaderboardColumn.java
seb-h2o: I'll still leave it for consistency and debuggability: leaderboard stores the columns in a map, that's why it's not used, but when I have a column, I want to be able to know to which model it is associated.
Also, if I decided to store them in a list, the modelId would be necessary: the interface should be self-contained and independent from the data structure I'll use to store the instances.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -55,6 +58,22 @@
...
+    final Key<AutoML> amlKey = automl._key;
+
+    return new LeaderboardExtensionsProvider() {
+      @Override
+      public LeaderboardColumn[] createExtensions(Model model) {
+        final AutoML aml = amlKey.get();
+        return new LeaderboardColumn[] {
+                new TrainingTime(model),
+                new ScoringTimePerRow(model, aml.getLeaderboardFrame(), aml.getTrainingFrame()),
+//                new ModelSize(model._key)
```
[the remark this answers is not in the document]
seb-h2o: will do

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/LeaderboardColumn.java
seb-h2o: it's necessary to forward it to `Freezable`, and allow the implementations to extend `Iced`.
That's why I used the name `SELF`, to suggest that the implementation should put its own class here.

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/LeaderboardColumn.java
seb-h2o: `getValue` is an accessor, by convention in Java, accessors are not supposed to do long running tasks, user should be safe to call them anytime.
`fetch` explicitly says that something needs to be retrieved.
I will add Javadoc, but I think the naming is quite standard.

h2o-automl/src/test/java/ai/h2o/automl/AutoMLTest.java
seb-h2o: This shouldn't have been committed, I'll write proper tests. Was just using this to check quickly.
And I won't add enum, each extension column class exposes its name.

h2o-automl/src/main/java/water/automl/api/schemas3/EventLogV99.java
seb-h2o: No. As explained in the `refactoring` section of the ticket description. I had to move `Leaderboard` to its package to isolate the leaderboard logic.
As there is some kind of structural symmetry between EventLog and Leaderboard design, I also moved the first one to its package, and applied the same small changes at API level.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLV99.java
seb-h2o: not yet, I don't want to change the clients yet.

h2o-automl/src/main/java/water/automl/api/LeaderboardsHandler.java
seb-h2o: I don't see a big benefit in using `Optional` here. It always needs a string for the exception...

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/LeaderboardColumn.java
seb-h2o: To say differently, a column entry without anything to identify its row index can quickly become unusable.

seb-h2o: I may rename the interface to `LeaderboardCell`, and `LeaderboardColumnDescriptor` to `LeaderboardColumn`.
This would make more sense indeed.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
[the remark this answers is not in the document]
seb-h2o: https://0xdata.atlassian.net/browse/PUBDEV-7124

h2o-automl/src/main/java/ai/h2o/automl/events/EventLogEntry.java
seb-h2o: https://0xdata.atlassian.net/browse/PUBDEV-7126

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/Leaderboard.java
```diff
@@ -25,6 +26,61 @@
  * TODO: make this robust against removal of models from the DKV.
  */
 public class Leaderboard extends Lockable<Leaderboard> {
+
```
seb-h2o: moved this bloc of static methods on top of the class, as it is common usage in Java. No change in impl.

h2o-automl/src/test/java/ai/h2o/automl/leaderboard/LeaderboardTest.java
```diff
@@ -0,0 +1,167 @@
...
+        model.deleteCrossValidationModels();
+        model.delete();
+      }
+      if( fr != null) {
+        fr.delete();
+      }
+    }
+  }
+
+  @Test
```
seb-h2o: only this test is new.
The rest of the test class was moved when moving `Leaderboard` to its dedicated package.

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/LeaderboardColumn.java
seb-h2o: @deil87 : client mode proved me that it was necessary.
The `IcedHashMap<Key, LeaderboardCell[]>` raised `ClassCastException` after deserialization in client mode, so I changed the storage as a simple `LeaderboardCell[]` and it now works :)

h2o-automl/src/main/java/water/automl/api/LeaderboardsHandler.java
deil87: @seb-h2o i was not actually suggesting to change anything, just was surprised by this implementation of the Option concept :)

h2o-automl/src/test/java/ai/h2o/automl/leaderboard/LeaderboardTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+      TwoDimTable lb_table_ext = lb.toTwoDimTable(LeaderboardExtensionsProvider.ALL);
+      assertEquals(7+2, lb_table_ext.getColDim());
+      assertArrayEquals(new String[] {"model_id", "auc", "logloss", "aucpr", "mean_per_class_error", "rmse", "mse", "training_time_ms", "predict_time_per_row_ms"},
+              lb_table_ext.getColHeaders());
+      assertArrayEquals(new String[] {"string", "double", "double", "double", "double", "double", "double", "long", "double"},
+              lb_table_ext.getColTypes());
+      assertArrayEquals(new String[] {"%s", "%.6f", "%.6f", "%.6f", "%.6f", "%.6f", "%.6f", "%s", "%.6f"},
+              lb_table_ext.getColFormats());
+      assertTrue(lb_table_ext.get(0, 7) instanceof Long);
```
deil87: @seb-h2o can I ask you to use self descriptive variables for these numbers ( 7, 8 etc) ? would make it easier to read test. Or comments, but imho variables is a better way to avoid repetitions of comments

h2o-py/h2o/automl/autoh2o.py
```diff
@@ -654,3 +666,23 @@ def get_automl(project_name):
...
+
+
+def get_leaderboard(aml, extensions=None):
+    """
+    Retrieve the leaderboard from the AutoML instance.
+    Contrary to the default leaderboard attached to the automl instance, this one can return columns other than the metrics.
+    :param H2OAutoML aml: the instance for which to return the leaderboard.
+    :param extensions: a string or a list of string specifying which optional columns should be added to the leaderboard. Defaults to None.
+        Currently supported extensions are:
+        - 'ALL': adds all columns below.
```
deil87: @seb-h2o I personally expect all string keys to be of the same case. Maybe all capitals?

h2o-r/h2o-package/R/automl.R
```diff
@@ -486,3 +498,32 @@ h2o.getAutoML <- function(project_name) {
...
+#' h2o.init()
+#' votes_path <- system.file("extdata", "housevotes.csv", package = "h2o")
+#' votes_hf <- h2o.uploadFile(path = votes_path, header = TRUE)
+#' aml <- h2o.automl(y = "Class", project_name="aml_housevotes",
+#'                   training_frame = votes_hf, max_runtime_secs = 30)
+#' lb_all <- h2o.getLeaderboard(aml, 'ALL')
+#' lb_custom <- h2o.getLeaderboard(aml, c('predict_time_per_row_ms', 'training_time_ms'))
+#' }
+#' @export
+h2o.getLeaderboard <- function(aml, extensions=NULL) {
```
deil87: @seb-h2o shouldn’t  we use underscore style `get_leaderboard` here?

h2o-py/h2o/automl/autoh2o.py
deil87: @seb-h2o but probably ALL is of a higher level semantic here

h2o-py/tests/testdir_algos/automl/pyunit_automl_leaderboard.py
```diff
@@ -293,6 +293,29 @@ def test_AUTO_stopping_metric_with_custom_sorting_metric():
     check_model_property(non_se, 'stopping_metric', True, "RMSE")
 
 
+def test_custom_leaderboard():
+    print("Check custom leaderboard")
+    ds = prepare_data('binomial')
+    aml = H2OAutoML(project_name="py_aml_custom_lb_test",
```
deil87: @seb-h2o I was thinking about whether we treat default columns as default extensions... Don't we want to sort based on extension columns?
```
aml = H2OAutoML(project_name="py_aml_custom_lb_test",
                    max_models=5, sort_metric = "training_time_ms",
                    seed=automl_seed)
get_leaderboard(aml, extensions=["training_time_ms"]).names == std_columns + ["training_time_ms"]
```
Maybe we can lazily keep sorting metric and only up on request return or fail depending on whether column exists in the leaderboard frame?

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/ModelSize.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public boolean isNA() {
+        return getValue() == null || getValue() < 0;
+    }
+
+    @Override
+    public Long fetch() {
+        if (getValue() == null) {
+            try {
+//                Model model = _modelId.get();
```
deil87: @seb-h2o TODO?

ledell:
I am putting the R interface and output here for discussion.  I don't think we want to make users use the `attr()` function this way.  Too awkward and hard to document.

```
> attr(aml@leaderboard, 'extended')()
                                             model_id mean_per_class_error
1                        GLM_1_AutoML_20191210_011713           0.02666667
2 StackedEnsemble_BestOfFamily_AutoML_20191210_011713           0.04666667
3                    XGBoost_2_AutoML_20191210_011713           0.06000000
4                        DRF_1_AutoML_20191210_011713           0.06000000
5                    XGBoost_3_AutoML_20191210_011713           0.06000000
6    StackedEnsemble_AllModels_AutoML_20191210_011713           0.06000000
     logloss      rmse        mse training_time_ms predict_time_per_row_ms
1 0.06742221 0.1483498 0.02200765              372                0.162997
2 0.28696076 0.2626580 0.06898924              226                0.129062
3 0.38684309 0.3295028 0.10857206               76                0.007728
4 0.32766183 0.1957416 0.03831478               91                0.024406
5 0.22489948 0.2342886 0.05489116               64                0.007669
6 0.26886025 0.2531024 0.06406083              215                0.072973
```


```
> print(attr(aml@leaderboard, 'extended')("training_time_ms"))
                                             model_id mean_per_class_error
1                        GLM_1_AutoML_20191210_011713           0.02666667
2 StackedEnsemble_BestOfFamily_AutoML_20191210_011713           0.04666667
3                    XGBoost_2_AutoML_20191210_011713           0.06000000
4                        DRF_1_AutoML_20191210_011713           0.06000000
5                    XGBoost_3_AutoML_20191210_011713           0.06000000
6    StackedEnsemble_AllModels_AutoML_20191210_011713           0.06000000
     logloss      rmse        mse training_time_ms
1 0.06742221 0.1483498 0.02200765              372
2 0.28696076 0.2626580 0.06898924              226
3 0.38684309 0.3295028 0.10857206               76
4 0.32766183 0.1957416 0.03831478               91
5 0.22489948 0.2342886 0.05489116               64
6 0.26886025 0.2531024 0.06406083              215
```

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/Leaderboard.java
```diff
@@ -95,58 +160,8 @@ public Leaderboard(String project_name, EventLog eventLog, Frame leaderboardFram
...
-   */
-  public static String idForProject(String project_name) { return "AutoML_Leaderboard_" + project_name; }
-
-  /**
-   * @param metric
-   * @return true iff the metric is a loss function
-   */
-  public static boolean isLossFunction(String metric) {
-    return metric != null && !Arrays.asList("auc", "aucpr").contains(metric.toLowerCase());
+  public void setExtensionsProvider(LeaderboardExtensionsProvider provider) {
```
deil87: @seb-h2o except for the docs what would suggest user to use this method? maybe add example to javadoc of Leaderboard as you did for Py/R?

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -55,6 +58,22 @@
           new StepDefinition(Algo.StackedEnsemble.name(), Alias.defaults),
   };
 
+  private static LeaderboardExtensionsProvider createLeaderboardExtensionProvider(AutoML automl) {
```
deil87: @seb-h2o is it a sort of default extension provider? if yes maybe it is better to reflect it in the naming?

h2o-py/h2o/automl/autoh2o.py
seb-h2o: 'all' also works, but for clarity in the doc, I prefer making a distinction between a "normal" column name (like "training_time_ms") and a constant with a different semantic like "ALL".

h2o-r/h2o-package/R/automl.R
seb-h2o: I've been hesitating too: problem is that automl module started to use `getAutoML`, so for consistency, using the same case here.
Another solution it to create an alias `get_automl` for `getAutoML` and have only `get_leaderboard` for this new function.
@ledell ?

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/Leaderboard.java
```diff
@@ -198,42 +213,39 @@ public Frame leaderboardFrame() {
...
 
   /**
    * @return list of models sorted by the given metric
    */
   public Model[] getModelsSortedByMetric(String metric) {
-    Key<Model>[] modelKeys = sortModels(metric);
-    if (modelKeys == null || 0 == modelKeys.length) return new Model[0];
-    return getModelsFromKeys(modelKeys);
+    if (getModelCount() == 0) return new Model[0];
+    return getModelsFromKeys(sortModels(metric));
```
deil87: @seb-h2o wdyt about changing `sortModels` to `getSortedModelKeys`? or at least `sortModelKeys`?

h2o-py/tests/testdir_algos/automl/pyunit_automl_leaderboard.py
seb-h2o: leaderboard is represented as a `H2OFrame` on client side, user can do the sorting by himself then: no point to complicate the backend logic for a functionality that you get for free on client side.

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/ModelSize.java
seb-h2o: no need, ticket was created for this. I can add it to the comment..

h2o-py/tests/testdir_algos/automl/pyunit_automl_leaderboard.py
deil87: @seb-h2o maybe, but we already have `sort_metric = "training_time_ms"` functionality for H2OAutoML. Just to follow existing contract I would expect to support this for extensions as well. Or we should at least for ourselves distinguish original metrics and extensions.

seb-h2o: I called those "extensions" to make it clear that they are distinct from "metrics", so I don't expect `sort_metric` to apply here.
Besides, it would complicate the implementation a lot given how sorting is done today on backend, for almost zero benefit.
If people really want it,  we can do it later, but currently it doesn't make sense imo.

Is it really difficult to do:
```
h2o.automl.get_leaderboard(aml, 'ALL').sort(by='training_time_ms')
```
I think duplicating features is the best way to create inconsistent APIs.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
seb-h2o: it's not a default in the sense that it would be possible to change it... it is the provider used by AutoML. dot.

h2o-automl/src/test/java/ai/h2o/automl/leaderboard/LeaderboardTest.java
seb-h2o: will do

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/Leaderboard.java
seb-h2o: adding more javadoc in various places now.

h2o-py/tests/testdir_algos/automl/pyunit_automl_leaderboard.py
deil87: It was not about duplicating features but about making original one to be working with new extensions ( which could be seen as columns we might want to sort by as well).  The only point I wanted to communicate is that we already have this feature and to support it we would want to support sorting based on extension columns. If we don't wan't to support we need to deprecate existing `sort_metric` and suggest to use sorting from Frames.  But as you have said those are "extensions"... and we distinguish between them... then we are fine. Thanks for clarifying!

deil87:
@seb-h2o description of PR says you are going to add more java tests. Have you done this already?

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/Leaderboard.java
seb-h2o: - `sortModels` is mainly delegating to `ModelMetrics.sortModelsByMetric`, so it's consistent internally at least.
- I tend to avoid `get___` for methods that do something instead of just returning an object stored somewhere. Otherwise, any method that doesn't return `void` would be potentially named `getSomethingAfterDoingSomethingElse`.
- `sortModels` is private, so I'm less picky about its naming.
- I actually prefer `sortModels` to `sortModelKeys` because that's really the models we're sorting somehow (we need the model to obtain its metric and so on...), if we were just sorting keys we would have choice between lexicographic order, and... it's quite limited.

h2o-automl/src/test/java/ai/h2o/automl/leaderboard/LeaderboardTest.java
seb-h2o: @deil87 to answer your question about Java tests: this big test should cover all the Leaderboard extensions feature.

h2o-r/h2o-package/R/automl.R
seb-h2o: problem solved: 
- exposing `h2o.get_automl` and `h2o.get_leaderboard`
- deprecated `h2o.getAutoML`: still functional though.

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/Leaderboard.java
deil87: @seb-h2o 
> I tend to avoid get___ for methods that do something instead of just returning an object stored somewhere.

does mention activity stays visible to the outside world? if no then I would say it should not matter whether we just retrieve from storage or compute something to prepare result. 

> I actually prefer sortModels to sortModelKeys because that's really the models we're sorting somehow (we need the model to obtain its metric and so on.

i think it is result(returning values) that defines naming.... we should not care **how** it is done

> sortModels is private, so I'm less picky about its naming

agree, but to care about developers as well we better be equally picky :)

seb-h2o: > we should not care how it is done
I agree with that, but that's not what I mean. I mean we are really sorting the models, not its keys. It's just that the keys are just what you need to access the models then...
> result defines naming
so many examples where this is not true, it's mainly true for accessors. But definitely not a requirement imo, especially with a strongly types language.

h2o-py/tests/testdir_algos/automl/pyunit_automl_leaderboard.py
ledell: It's true that we have `sort_metric` in AutoML and then `sort_by` in the get_grid code.   But since it's metric specific (and our new columns are not metrics) it seems less relevant.  It's not a pervasive convention either. I agree that people can sort a data frame on their own... the only non-trivial part is that the leaderboard is an H2OFrame and so users need to know how to sort H2OFrames, which unfortunately is not the same way that you can sort R data.frames or Python Pandas DataFrames.  So it's more awkward than it should be. http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-munging/sortcolumn.html

seb-h2o: @ledell this can be shown through examples, let me add a line for sorting in the `get_leaderboard` Py example.
Also, h2o frames can easily be converted to pandas df.
We may want to have a detailed example in the online doc.

ledell:
## Final client API

### Python
```
lb_all = h2o.automl.get_leaderboard(aml, 'ALL')
lb_custom = h2o.automl.get_leaderboard(aml, ['predict_time_per_row_ms', 'training_time_ms'])
lb_custom_sorted = lb_custom.sort(by='predict_time_per_row_ms')
```

### R
```
lb_all <- h2o.get_leaderboard(aml, 'ALL')
lb_custom <- h2o.get_leaderboard(aml, c('predict_time_per_row_ms', 'training_time_ms'))
```

Also this PR adds an alias in R for `h2o.getAutoML()` which is `h2o.get_automl()`.

h2o-py/tests/testdir_algos/automl/pyunit_automl_leaderboard.py
ledell: @seb-h2o yeah we just need more documentation

PR #4126, opened 2019-12-06
Grid: Make sure getModelKeys returns keys in a stable order

michalkurka:
getModelKeys is backed by IcedHashMap - which itself doesn't guarantee
stable ordering

### DIFF
--- a/h2o-core/src/main/java/hex/grid/Grid.java
+++ b/h2o-core/src/main/java/hex/grid/Grid.java
@@ -317,7 +317,9 @@ public class Grid<MP extends Model.Parameters> extends Lockable<Grid<MP>> {
    * @return list of model keys
    */
   public Key<Model>[] getModelKeys() {
-    return _models.values().toArray(new Key[_models.size()]);
+    Key<Model>[] keys = _models.values().toArray(new Key[_models.size()]);
+    Arrays.sort(keys);
+    return keys;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -317,7 +317,9 @@ void appendFailedModelParameters(final Key<Model> modelKey, final MP params, fin
    * @return list of model keys
    */
   public Key<Model>[] getModelKeys() {
```
Pscheidl: Small note. I personally try to make the JavaDoc of our project a little bit better. So I'd add an information to the Javadoc it is actually the intention and desired behavior to order the model keys lexically.

h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -317,7 +317,9 @@ void appendFailedModelParameters(final Key<Model> modelKey, final MP params, fin
    * @return list of model keys
```
Pscheidl:
```suggestion
   * @return list of model keys sorted lexically
```

Pscheidl: Small note. I personally try to make the JavaDoc of our project a little bit better. So I'd add an information to the Javadoc it is actually the intention and desired behavior to order the model keys lexically.

PR #4129, opened 2019-12-09
PUBDEV-7127: Don't perform internal self-check when converting trees in H2O

michalkurka:
Only keep in PrintMojo - which was the original intention of the check

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ConvertTreeOptions.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ConvertTreeOptions.java
@@ -1,0 +1,30 @@
+package hex.genmodel.algos.tree;
+
+public class ConvertTreeOptions {
+  static final ConvertTreeOptions DEFAULT = new ConvertTreeOptions();
+
+  final boolean _checkTreeConsistency;
+
+  public ConvertTreeOptions() {
+    this(false);
+  }
+  
+  private ConvertTreeOptions(boolean checkTreeConsistency) {
+    _checkTreeConsistency = checkTreeConsistency;
+  }
+
+  /**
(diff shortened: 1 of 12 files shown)


h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ConvertTreeOptions.java
```diff
@@ -0,0 +1,23 @@
+package hex.genmodel.algos.tree;
+
+public class ConvertTreeOptions {
+  static final ConvertTreeOptions DEFAULT = new ConvertTreeOptions();
```
honzasterba: I would very much like _checkTreeConsistency to be final, making this an immutable object, which in turn would make this static safe from modification

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ConvertTreeOptions.java
```diff
@@ -0,0 +1,23 @@
...
+
+  boolean _checkTreeConsistency = false;
+
+  /**
+   * Performs a self-check on each converted tree. Inconsistencies are reported in the log.
+   * @return self
+   */
+  public ConvertTreeOptions enableCheckTreeConsistency() {
```
honzasterba: if we make this immutable we can keep this method, only it would return a new object and would probably need a different name like `withTreeConsitencyCheckEnabled()`

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ConvertTreeOptions.java
michalkurka: that makes sense, good point @honzasterba - thank you

PR #4130, opened 2019-12-10
PUBDEV 7089 - tree visualization via jgrapht

koniecsveta:


h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -153,6 +168,12 @@ private void parseArgs(String[] args) {
             nPlaces = Integer.parseInt(s);
             break;
 
+          case "--direct":
```
honzasterba: why not `--format png`

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -189,26 +210,61 @@ private void validateArgs() {
...
+    DOTImporter<String, LabeledEdge> importer = new DOTImporter<>(vertexProvider, edgesProvider, componentUpdater);
+    DirectedMultigraph<String, LabeledEdge> result = new DirectedMultigraph<>(LabeledEdge.class);
+    try ( FileInputStream is = new FileInputStream(dotSourceFilePath.toFile())) {
+      importer.importGraph(result, new InputStreamReader(is));
+      JGraphXAdapter<String, LabeledEdge> graphAdapter = new JGraphXAdapter<String, LabeledEdge>(result);
+      mxIGraphLayout layout = new mxCompactTreeLayout(graphAdapter, false);
+      layout.execute(graphAdapter.getDefaultParent());
+      BufferedImage image = mxCellRenderer.createBufferedImage(graphAdapter, null, 2, Color.WHITE, true, null);
+      File imgFile = new File(outputPngName);
+      ImageIO.write(image, "PNG", imgFile);
```
honzasterba: alternatively this may be written to stdout

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
honzasterba: this may also work without -o with the PNG being written to stdout

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -21,10 +30,16 @@
   private static int treeToPrint = -1;
   private static int maxLevelsToPrintPerEdge = 10;
   private static boolean detail = false;
-  private static String outputFileName = null;
+  private String outputFileName = null;
```
honzasterba: I cannot say I like all these being static, but we should have them all static or none

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -21,10 +30,16 @@
...
   private static String optionalTitle = null;
   private static PrintTreeOptions pTreeOptions;
   private static boolean internal;
+  private String outputFormat = "dot";
```
honzasterba: @michalkurka  maybe we should first merge https://github.com/h2oai/h2o-3/pull/4104 and build this on top of it?

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -21,10 +30,16 @@
...
+  private static final String tmpOutputFileName = "tmpOutputFileName.gv";
+
+  private LabeledVertexProvider vertexProvider = new LabeledVertexProvider();
```
honzasterba: it looks like these are only needed when doing png exports and also are used in one method, do we realy need them as instance variables?

michalkurka:
Dependency report:

```
+--- net.sf.opencsv:opencsv:2.3
+--- com.google.code.gson:gson:2.6.2
+--- ai.h2o:h2o-tree-api:0.3.9
+--- ai.h2o:deepwater-backend-api:1.0.4
+--- org.jgrapht:jgrapht-core:1.3.1
|    \--- org.jheaps:jheaps:0.10
+--- org.jgrapht:jgrapht-ext:1.3.1
|    +--- org.jgrapht:jgrapht-core:1.3.1 (*)
|    \--- com.github.vlsi.mxgraph:jgraphx:3.9.8.1
\--- org.jgrapht:jgrapht-io:1.3.1
     +--- org.jgrapht:jgrapht-core:1.3.1 (*)
     +--- org.antlr:antlr4-runtime:4.7.2
     \--- org.apache.commons:commons-text:1.6
          \--- org.apache.commons:commons-lang3:3.8.1
```

IMHO we should definitely relocate at least `commons-lang` and `antl4-runtime` in order to avoid dependency clashes when the shadowed jar is embedded into 3rd party applications (or Spark).

michalkurka:
@koniecsveta I don't think you need to close this PR and lose the comment history, you an just rebase and change the target branch on Github if you want to.

PR #4134, opened 2019-12-11
Add a property that enabled verbose logging of stratified sampling

michalkurka:


h2o-algos/src/main/java/hex/tree/SharedTree.java
```diff
@@ -232,7 +232,8 @@ protected void validateRowSampleRate() {
                 throw new IllegalArgumentException("class_sampling_factors must have " + _train.lastVec().domain().length + " elements");
               trainSamplingFactors = _parms._class_sampling_factors.clone(); //clone: don't modify the original
             }
-            Frame stratified = water.util.MRUtils.sampleFrameStratified(_train, _train.lastVec(), _train.vec(_model._output.weightsName()), trainSamplingFactors, (long)(_parms._max_after_balance_size*_train.numRows()), _parms._seed, true, false);
+            boolean verboseSampling = Boolean.getBoolean(H2O.OptArgs.SYSTEM_PROP_PREFIX + "debug.sharedTree.sampleFrameStratified.verbose");
```
Pscheidl: Could be final

h2o-core/src/main/java/water/util/MRUtils.java
```diff
@@ -223,6 +223,9 @@ public static Frame sampleFrameStratified(final Frame fr, Vec label, Vec weights
       numrows += sampling_ratios[i] * dist[i];
     }
     if (Float.isNaN(numrows)) {
+      Log.err("Total number of sampled rows was NaN. " +
```
honzasterba: why not improve the exception message as well, this way the user has to dig through the logs to get the details which could easily be part of the exception

h2o-core/src/main/java/water/util/MRUtils.java
```diff
@@ -223,6 +223,9 @@ public static Frame sampleFrameStratified(final Frame fr, Vec label, Vec weights
       numrows += sampling_ratios[i] * dist[i];
```
honzasterba: also we could add detection of NaN here or even sooner where one of these factors becomes NaN

h2o-core/src/main/java/water/util/MRUtils.java
michalkurka: The number of classes can be large and including all in the message might cause issues on the client (large JSON).

We can add information about which class was the problematic one but this most likely won't be user-actionable, this is most likely caused by a numerical instability of our math.

h2o-core/src/main/java/water/util/MRUtils.java
michalkurka: yeah, we could - but IMHO not super useful, reporting ex-post seems enough to me - it is not expected this math will fail at this point

PR #4137, opened 2019-12-12
disabling XGBoost in multinode

seb-h2o:
disabling XGBoost in multinode and improved warning message when user is trying to use it in cluster environment.

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -323,7 +323,7 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
         boolean isMultinode = H2O.CLOUD.size() > 1;
         _eventLog.warn(Stage.ModelTraining,
                 isMultinode ? "AutoML: "+algo.name()+" is not available in multi-node cluster; skipping it."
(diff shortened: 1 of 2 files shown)


h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -320,7 +320,11 @@ void planWork() {
 
     for (Algo algo : Algo.values()) {
       if (!skippedAlgos.contains(algo) && !algo.enabled()) {
-        _eventLog.warn(Stage.ModelTraining, "AutoML: "+algo.name()+" is not available; skipping it.");
+        boolean isMultinode = H2O.CLOUD.size() > 1;
+        _eventLog.warn(Stage.ModelTraining,
+                isMultinode ? "AutoML: "+algo.name()+" is not available in multi-node cluster; skipping it." :
```
michalkurka: Maybe we can add info how to enable it manually? In such case we either link documentation or create a technote - eg. https://0xdata.atlassian.net/browse/TN-1

seb-h2o: yes, was wondering how to expose this... the link to doc should be fine. Adding it.

seb-h2o: Are technotes public? Good to know

PR #4138, opened 2019-12-12
PUBDEV-6966 kmeans compute training metrics on dataset from parms

maurever:
JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6966

_train parameter has no key by default, _parms.train() has key

### DIFF
--- a/h2o-algos/src/main/java/hex/kmeans/KMeans.java
+++ b/h2o-algos/src/main/java/hex/kmeans/KMeans.java
@@ -357,7 +357,7 @@ public class KMeans extends ClusteringModelBuilder<KMeansModel,KMeansModel.KMean
         vecs2[vecs2.length-1].remove();
 
         // Create metrics by scoring on training set otherwise scores are based on last Lloyd iteration
-        model.score(_train).delete();
+        model.score(_parms.train()).delete();
         model._output._training_metrics = ModelMetrics.getFromDKV(model,_parms.train());
 
         Log.info(model._output._model_summary);
--- a/h2o-algos/src/main/java/hex/kmeans/KMeans.java
+++ b/h2o-algos/src/main/java/hex/kmeans/KMeans.java
@@ -357,8 +357,8 @@ public class KMeans extends ClusteringModelBuilder<KMeansModel,KMeansModel.KMean
         vecs2[vecs2.length-1].remove();
 
         // Create metrics by scoring on training set otherwise scores are based on last Lloyd iteration
-        model.score(_train).delete();
-        model._output._training_metrics = ModelMetrics.getFromDKV(model,_train);
+        model.score(_parms.train()).delete();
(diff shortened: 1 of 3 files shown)


h2o-algos/src/main/java/hex/kmeans/KMeans.java
```diff
@@ -357,8 +357,8 @@ public void computeImpl() {
         vecs2[vecs2.length-1].remove();
 
         // Create metrics by scoring on training set otherwise scores are based on last Lloyd iteration
-        model.score(_train).delete();
-        model._output._training_metrics = ModelMetrics.getFromDKV(model,_train);
+        model.score(_parms.train()).delete();
```
michalkurka: I do not believe this is correct, scoring needs to be done on the adapted dataset, not the original dataset with extra columns

michalkurka: This comment is incorrect ^^^ please ignore. Model#score calls adaptTestToTrain - which makes this ok

h2o-algos/src/test/java/hex/kmeans/KMeansTest.java
```diff
@@ -662,4 +662,28 @@ public void testTimeColumnPubdev6264() {
...
+      parms._train = fr._key;
+      parms._k = 3;
+      parms._standardize = true;
+      parms._max_iterations = 1;
+      parms._init = KMeans.Initialization.Random;
+
+      KMeans job = new KMeans(parms);
+      kmeans = (KMeansModel) Scope.track_generic(job.trainModel().get());
+
+      assert  kmeans._output._training_metrics.frame()._key != null;
```
michalkurka: please use assertNotNull instead

h2o-algos/src/test/java/hex/kmeans/KMeansTest.java
```diff
@@ -662,4 +662,28 @@ public void testTimeColumnPubdev6264() {
...
+    Frame fr = null;
+    try {
+      fr = parse_test_file("smalldata/iris/iris_wheader.csv");
+
+      KMeansModel.KMeansParameters parms = new KMeansModel.KMeansParameters();
+      parms._train = fr._key;
+      parms._k = 3;
+      parms._standardize = true;
+      parms._max_iterations = 1;
+      parms._init = KMeans.Initialization.Random;
```
michalkurka: the test should probably have a seed

h2o-algos/src/test/java/hex/kmeans/KMeansTest.java
```diff
@@ -662,4 +662,28 @@ public void testTimeColumnPubdev6264() {
...
+
+      KMeans job = new KMeans(parms);
+      kmeans = (KMeansModel) Scope.track_generic(job.trainModel().get());
```
michalkurka: this test doest use Scope, please remove the call to track_generic

PR #4147, opened 2019-12-13
[PUBDEV-7064] Fix NPE in GBM when sample rate < 1

honzasterba:


h2o-algos/src/main/java/hex/tree/gbm/GBM.java
```diff
@@ -954,10 +954,6 @@ public HuberLeafMath(FrameMap frameMap, double huberDelta, Vec strata) {
 
     @Override
     public void map(Chunk[] cs) {
-      if (_strataMin < 0 || _strataMax < 0) {
-        Log.warn("No Huber math can be done since there's no strata.");
-        return;
-      }
```
maurever: What is the reason to remove these lines of code? I am just curious. :)

h2o-algos/src/main/java/hex/tree/gbm/GBM.java
```diff
@@ -954,10 +954,6 @@ public HuberLeafMath(FrameMap frameMap, double huberDelta, Vec strata) {
...
       final int nstrata = _strataMax - _strataMin + 1;
       Log.info("Computing Huber math for (up to) " + nstrata + " different strata.");
       _huberGamma = new double[nstrata];
```
maurever: I am surprised this line of code does not solve the problem with the initialization of `_huberGamma`...  What is the case this line does not initialize the _huberGamma? The zero-length chunk?

h2o-algos/src/main/java/hex/tree/gbm/GBM.java
honzasterba: its does not make any sense, the strata can be negative, even when it is negative its still useful

h2o-algos/src/main/java/hex/tree/gbm/GBM.java
honzasterba: you are partly right, let me improve this

maurever: Ok, looks more understandable. :+1:

PR #4150, opened 2019-12-13
PUBDEV-6929: Fix 1-hot encoder for POJO

michalkurka:


h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
```diff
@@ -115,18 +123,19 @@ Constraints emptyConstraints(Frame f) {
...
   }
 
-  @Test public void testOneHotExplicitWithPOJO() {
+  @Test public void testOneHotExplicitWithPOJO() throws Exception {
     try {
       Scope.enter();
       final String response = "CAPSULE";
-      Frame fr = parse_test_file("./smalldata/logreg/prostate.csv")
+      final String testFile = "./smalldata/logreg/prostate.csv";
+      Frame fr = parse_test_file(testFile)
```
Pscheidl: Minor: can also be final. Together with some other variables. Not something I insist on, yet I noticed it and want to mention it.

h2o-genmodel/src/main/java/hex/genmodel/tools/PredictCsv.java
```diff
@@ -453,7 +471,7 @@ private void checkMissingColumns(final String[] parsedColumnNamesArr) {
   private void parseArgs(String[] args) {
     try {
       String pojoMojoModelNames = ""; // store Pojo/Mojo/Model names
-      int loadType = 0; // 0: load pojo, 1: load mojo, 2: load model
+      int loadType = -1; // 0: load pojo, 1: load mojo, 2: load model
```
Pscheidl: Minor: Outdated comment ?

h2o-genmodel/src/main/java/hex/genmodel/easy/OneHotEncoderColumnMapper.java
```diff
@@ -18,11 +18,20 @@ public OneHotEncoderColumnMapper(GenModel m) {
     String[][] origDomainValues = _m.getOrigDomainValues();
     Map<String, Integer> columnMapping = new HashMap<>(origNames.length);
     int pos = 0;
+    // non-categorical
+    for (int i = 0; i < _m.getOrigNumCols(); i++) {
```
Pscheidl: We need to iterate twice, because the number of null `origDomainValues` is not known beforehand, correct ? Thus we can't do only one `for` cycle, writing categoricals with proper offset righ away ?

h2o-genmodel/src/main/java/hex/genmodel/easy/OneHotEncoderColumnMapper.java
```diff
@@ -18,11 +18,20 @@ public OneHotEncoderColumnMapper(GenModel m) {
...
+    // categorical
     for (int i = 0; i < _m.getOrigNumCols(); i++) {
-      String colName = origNames[i];
-      columnMapping.put(colName, pos);
       String[] domainValues = origDomainValues[i];
-      pos += domainValues != null ? domainValues.length + 1 : 1; 
+      if (domainValues == null)
+        continue;
+      columnMapping.put(origNames[i], pos);
+      pos += domainValues.length + 1;
```
Pscheidl: Why is `+1` needed ? Advancing for `domainValues.length` should be enough for the reader ?

michalkurka: The last one is for NA (missing).

Pscheidl: I see. Thanks.

h2o-genmodel/src/main/java/hex/genmodel/easy/OneHotEncoderColumnMapper.java
michalkurka: That is correct, it could be simplified but this code runs just once on load. IMHO not an issue to do 2 for-loops.

h2o-genmodel/src/main/java/hex/genmodel/tools/PredictCsv.java
michalkurka: you are right, will fix

PR #4155, opened 2019-12-16
[DOCKER] Use postgres as back end for hive metastore

honzasterba:
hopefully this will make hadoop tests more stable


docker/hadoop/common/startup/51_postgres_init
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -ex
+
+# init database for metastore
+
+#rm -rf /var/lib/mysql
```
michal-raska: Can we remove the commented lines please?

docker/hadoop/cdh/Dockerfile
```diff
@@ -89,11 +89,9 @@ RUN chmod 700 /usr/sbin/startup.sh && \
     chown -R hive:hive /opt/hive-scripts && \
     chmod +x /usr/sbin/install_hive.sh && \
     chmod +x /usr/sbin/install_ldap.sh && \
-    chmod +x /usr/sbin/mysql_configure.sh && \
     sync && \
     /usr/sbin/install_hive.sh && \
-    /usr/sbin/mysql_configure.sh && \
-    ln -sf /usr/share/java/mysql-connector-java.jar /usr/lib/hive/lib/mysql-connector-java.jar
+    ln -sf /usr/share/java/postgresql-jdbc4.jar /usr/lib/hive/lib/postgresql-jdbc4.jar
```
michal-raska: There's a similar linking in `docker/hadoop/hdp/scripts/sbin/install_hive.sh`. I would like if they will be handled in one place, but that's probably out of scope of this PR.

honzasterba: yes, the difference is that HIVE_HOME is not needed here

docker/hadoop/common/startup/51_postgres_init
honzasterba: good catch, forgot about those

PR #4156, opened 2019-12-16
Make hadoop tests more stable

honzasterba:


michalkurka:
@honzasterba how does this fix work?

Smaller blocks mean that H2O streams data in smaller batches from HDFS. However, smaller blocks also mean that the name node needs also more memory.

honzasterba:
since we only put very small files on HDFS it does not make much sense to have 64MB blocks, we have ran into issues with the whole job failing because of disk-space being too low, so this is an attempt to lower the disk-space requiremets since its calculated by a number of avialble blocks.

michalkurka:
@honzasterba I see, thanks for explanation - makes sense.

PR #4157, opened 2019-12-16
PUBDEV-7150: fixed race condition on 2dimtable creation for eventlog

seb-h2o:


h2o-automl/src/main/java/ai/h2o/automl/events/EventLog.java
```diff
@@ -71,9 +71,10 @@ private static String idForRun(Key<AutoML> runKey) {
   /** Add a EventLogEntry, but don't log. */
   public void addEvent(EventLogEntry event) {
```
seb-h2o: the thread safety of this method will have to be handled properly.
Currently works because of we create events only from the main AutoML thread, but this method should probably get `synchronized`

PR #4165, opened 2019-12-18
PUBDEV-7088 - Tree Class documentation is difficult to interpret. Add examples.

Pscheidl:
- [PUBDEV-7088](https://0xdata.atlassian.net/browse/PUBDEV-7088)
- [Support portal ticket 95917](https://support.h2o.ai/a/tickets/95917)

----------------

- Mentioned that our representation is based on Scikit directly.
- Explained how can the information about each node be accessed (using it's ID as an index)
- Added bits and pieces from my blogpost.
- Changed the constructor used in examples to explicitely name the parameters.

Do you think it is good enough ? Shall I add more ? Shall I create a dedicated section in the docs.h2o.ai ?

Thank you for your reviews and comments.

### DIFF
--- a/h2o-py/h2o/tree/tree.py
+++ b/h2o-py/h2o/tree/tree.py
@@ -5,13 +5,50 @@ from h2o.estimators import H2OXGBoostEstimator
 class H2OTree():
     """
     Represents a model of a Tree built by one of H2O's algorithms (GBM, Random Forest, XGBoost, Isolation Forest).
-
+    
+    The internal structure mimics behavior of Scikit's internal tree representation and contains all the information available about every node in the graph.
+    (https://scikit-learn.org/stable/auto_examples/tree/plot_unveil_tree_structure.html)
+    It provides both human-readable output and format suitable for machine processing.
+    
+    In the fetched object, there two representations of the graph contained:
+    - graph-oriented representation, starting with a root node with edges,
+    - array/vector representation, useful for quick machine processing of the tree’s structure.
+    
+    Every graph starts wit ha root node, which is in fact an instance of H2OSplitNode and is naturally the very beginning of all decision paths in graph.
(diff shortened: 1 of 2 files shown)


h2o-py/h2o/tree/tree.py
```diff
@@ -4,44 +4,46 @@
...
-    At tree level, the following information is provided:
-    - Number of nodes in the tree,
-    - Model the tree belongs to,
-    - Tree class (if applicable),
-    - Pointer to a root node for tree traversal (breadth-first, depth-first) and manual tree walking.
+    At the tree level, the following information is provided:
+
+    - Number of nodes in the tree
+    - Model the tree belongs to
+    - Tree class (if applicable,
```
neemah2o: Should there be a closing parentheses after applicable?

h2o-py/h2o/tree/tree.py
```diff
@@ -5,24 +5,63 @@
...
+    2. Prediction value (floating point number).
+    
     :param model: The model this tree is related to.
     :param tree_number: An integer representing the order in which the tree has been built in the model.
-    :param tree_class: A string representing the name of the tree's class. The number of tree classes equals the number of levels in categorical response column. As there is exactly one class per categorical level, the name of the tree's class is equal to the corresponding categorical level of the response column. In case of regression and binomial models, the name of the categorical level is ignored and can be omitted, as there is exactly one tree built in both cases.
+    :param tree_class: A string representing the name of the tree's class. Specifies the class of the tree requested. Required for multi-class classification. The number of tree classes equals the number of levels in categorical response column. As there is exactly one class per categorical level, the name of the tree's class is equal to the corresponding categorical level of the response column. In case of regression and binomial models, the name of the categorical level is ignored and can be omitted.
```
neemah2o: I am thinking:
:param tree_class: A string representing the name of the **target** class. Specifies the......

neemah2o: The reason I would like to avoid the words "tree's class" is that this object is called "H2O Tree Class" so it can be confusing to the reader using the same name.

PR #4166, opened 2019-12-18
PUBDEV-7089-tree-visualization

koniecsveta:


h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,261 @@
...
+    static String getFormat(String[] args) {
+        for (int i = 0; i < args.length; i++) {
+            if (args[i].equals("--format")) {
+                return args[++i];
+            }
+        }
+        return "";
+    }
+    
+    protected static void usage() {
```
honzasterba: is this copy-paste really necessary? it will be a pain to maintain

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -75,7 +77,7 @@ private static void usage() {
     System.out.println();
     System.out.println("    --input | -i    Input mojo file.");
     System.out.println();
-    System.out.println("    --output | -o   Output dot filename.");
+    System.out.println("    --output | -o   Output dot|json|raw filename.");
```
honzasterba: nice catch, would just leave "Output filename."

h2o-genmodel/src/main/java/hex/genmodel/tools/MojoPrinter.java
```diff
@@ -0,0 +1,6 @@
+package hex.genmodel.tools;
+
+public interface MojoPrinter {
```
honzasterba: why do we need this ifc? I think there is couple of possible better solutions
a)  make JgraphtPrintMojo extend printMojo 
b) refactor PringMojo to be extendable via service interface and maybe also extract JSON into separate service

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,261 @@
...
+                ImageIO.write(image, "PNG", System.out);
+            }
+        }
+    }
+
+    protected String getPngName(String treeName) {
+        if (treeName.isEmpty()) {
+            return outputFileName;
+        } else {
+            String infixTreeName = treeName.replaceAll("\\s+", "").replaceAll(",", "_");
```
honzasterba: this magic is unnecessary imho, I think in case we want to generate multiple files the outputFile parameter should be treated as directory (created if not existing) and we should create treeX.png files there

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
```diff
@@ -112,10 +113,10 @@ public void testMojoCategoricalPrint_limitedLevels() throws IOException {
       }
 
       final String treeDotz = FileUtils.readFileToString(treeOutput);
-      assertTrue(treeDotz.contains("\"SG_0_Node_0\" -> \"SG_0_Node_1\" [fontsize=14, label=\"[NA]\\n2 levels\\n\"]\n" +
-          "\"SG_0_Node_0\" -> \"SG_0_Node_4\" [fontsize=14, label=\"Iris-virginica\\n\"]\n" +
```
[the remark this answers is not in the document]
koniecsveta: Changed the way how we are generating those \ because jgrapht dot importer had problem with it.. tested whether graphviz is still able to read them and it is ok also after this change.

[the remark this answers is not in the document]
honzasterba: if the resulting PNG from dot looks the same than fine

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: will fix

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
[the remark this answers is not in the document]
koniecsveta: it looks

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
```diff
@@ -185,11 +186,30 @@ private void assertMojoJSONEqualsFixture(Model model, String fixtureFile) throws
...
+      fail("Expected PrintMojo to call System.exit()");
+    } catch (PreventedExitException e) {
+      // expected
+    }
+    Path treeOutputPath = treeOutput.toPath();
+    Assert.assertTrue(Files.exists(treeOutputPath));
+    Files.walk(treeOutputPath)
+                .sorted(Comparator.reverseOrder())
+                .map(Path::toFile)
+                .forEach(File::delete);
```
honzasterba: this is in temporary folder, not necessary to delete, what I would like here is some check that png files were created as expected

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
```diff
@@ -185,11 +186,30 @@ private void assertMojoJSONEqualsFixture(Model model, String fixtureFile) throws
...
+    Assert.assertFalse(Files.deleteIfExists(treeOutputPath));
```
honzasterba: its nicer to use static imports for asserts

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
```diff
@@ -185,11 +186,30 @@ private void assertMojoJSONEqualsFixture(Model model, String fixtureFile) throws
...
     final String expectedTreeJson = IOUtils.toString(getClass().getResourceAsStream(fixtureFile));
     assertEquals(
-        removeH2OVersion(expectedTreeJson), 
-        removeH2OVersion(treeJson)
+            removeH2OVersion(expectedTreeJson),
+            removeH2OVersion(treeJson)
     );
   }
   
+  private void assertMojoPngGenerated(Model model, String outputFileName) throws IOException {
```
honzasterba: missing a test when the tree number is specified and instead of directory this creates just a plain file

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    private void printPng(SharedTreeGraph trees) throws IOException, ImportException {
+        Path outputDirectoryPath = Paths.get(outputFileName);
+        int numberOfTrees = trees.subgraphArray.size();
+        if (Files.exists(outputDirectoryPath)) {
+            Files.walk(outputDirectoryPath)
+                    .sorted(Comparator.reverseOrder())
+                    .map(Path::toFile)
+                    .forEach(File::delete);
```
honzasterba: why do we need to delete everything in the output directory?

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,224 @@
...
+            JGraphXAdapter<LabeledVertex, LabeledEdge> graphAdapter = new JGraphXAdapter<LabeledVertex, LabeledEdge>(result);
+            mxIGraphLayout treeLayout = new mxCompactTreeLayout(graphAdapter, true);
+            mxIGraphLayout nonOverlappingEdgesLayout = new mxEdgeLabelLayout(graphAdapter);
+            treeLayout.execute(graphAdapter.getDefaultParent());
+            nonOverlappingEdgesLayout.execute(graphAdapter.getDefaultParent());
+            BufferedImage image = mxCellRenderer.createBufferedImage(graphAdapter, null, 2, Color.WHITE, true, null);
+            if (outputFileName != null) {
+                ImageIO.write(image, "PNG", new File(getPngName(numberOfTrees, treeName)));
+            } else {
+                ImageIO.write(image, "PNG", System.out);
```
honzasterba: we probably should validate that png output with multiple trees also sets outputFile, writing multiple PNGs to stdout is not very useful

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,224 @@
...
+            try (FileOutputStream fosTemp = new FileOutputStream(dotSourceFilePath.toFile()); PrintStream osTemp = new PrintStream(fosTemp)) {
+                tree.printDot(osTemp, maxLevelsToPrintPerEdge, detail, optionalTitle, pTreeOptions, true);
+                generateOutputPng(dotSourceFilePath, numberOfTrees, tree.name);
+            }
+            Files.delete(dotSourceFilePath);
+        }
+
+    }
+
+    private void generateOutputPng(Path dotSourceFilePath, int numberOfTrees, String treeName) throws ImportException, IOException {
```
honzasterba: numberOfTrees should not be a parameter for this method, only as litle change as moving call to getPngName into the caller would make this much cleaner

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: In general, when we create output by PrintMojo, previous output is rewritten (existing behaviour). When we consider output directory as an output and it is left there from previous run, filled with .pngs and we want to generate new output with new run, on the same path, this was intended to preserve existing behaviour: rewrite existing output (=old directory with its contents) by new output (= new output directory with .pngs from actual run). if we want to preserve old content, I think this could be done in a way that we will move old content (if there is any) to archive directory, as discussed. What do you think?

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
koniecsveta: will fix

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
koniecsveta: hm I see, will fix

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: will fix

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: fixed

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
koniecsveta: fixed

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: Saving to directory done, left .replaceAll("\\s+", "").replaceAll(",", "_") part because otherwise names like "Tree 1, Class NO.png" would be possible

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: not necessary with SPI approach

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
```diff
@@ -258,6 +276,78 @@ public void testMojoGBMJson() throws IOException {
     }
   }
 
+  @Test
+  public void testMojoCategoricalPng() throws IOException {
+    try {
+      Scope.enter();
+      String[] expectedPngNames = new String[1];
```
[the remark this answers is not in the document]
koniecsveta: fixed

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
koniecsveta: fixed

h2o-genmodel/src/main/java/hex/genmodel/tools/MojoPrinter.java
koniecsveta: did SPI approach, JgraphtPrintMojo extends PrintMojo and MojoPrinter is used as service interface

h2o-algos/src/test/java/hex/tree/PrintMojoTreeTest.java
koniecsveta: testMojoCategoricalPng() tests model with _ntrees = 1

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
honzasterba: the output directory can be any directory, we do not care about there being other files and we should not presume anything about the contents of the directory, we only create it if does not exsits, we should definity remove everything, just overwrite the files we need to

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,227 @@
...
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Comparator;
+import java.util.Map;
+
+public class JgraphtPrintMojo extends PrintMojo implements MojoPrinter {
+
+    public static void main(String[] args) {
```
honzasterba: this main method is unnecessary I think

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,227 @@
...
+            System.exit(2);
+        }
+        
+        // Success
+        System.exit(0);
+    }
+    
+    @Override
+    protected void validateArgs() {
+        if (genModel == null) {
```
honzasterba: this method is a copy of the overriden method, it should not be, the validation that output!=png for jdk7 can be done in differently

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
honzasterba: I dont see it

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -13,34 +13,53 @@
...
+  protected final String tmpOutputFileName = "tmpOutputFileName.gv";
 
   public static void main(String[] args) {
+    MojoPrinter mojoPrinter = null;
+    
+    if ("png".equals(getFormat(args))) {
+      if (JAVA_VERSION.isKnown() && JAVA_VERSION.getMajor() > 7) {
+        ServiceLoader<MojoPrinter> mojoPrinters = ServiceLoader.load(MojoPrinter.class);
+        for (MojoPrinter printer : mojoPrinters) {
+          if ("hex.genmodel.tools.JgraphtPrintMojo".equals(printer.getClass().getName())) {
```
honzasterba: this can break in case of refactoring, what about a supportsFormat(Format) method instead?
woudl also get rid of the if below
basically I would refactor this to a findPrinterForFormat() method that would fail on JDK7 for PNG this way you can get rid of the check in validateArgs

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,227 @@
...
+            try (FileOutputStream fosTemp = new FileOutputStream(dotSourceFilePath.toFile()); PrintStream osTemp = new PrintStream(fosTemp)) {
+                tree.printDot(osTemp, maxLevelsToPrintPerEdge, detail, optionalTitle, pTreeOptions, true);
+                generateOutputPng(dotSourceFilePath, getPngName(numberOfTrees, tree.name));
+            }
+            Files.delete(dotSourceFilePath);
+        }
+
+    }
+
+    private void generateOutputPng(Path dotSourceFilePath, String treeName) throws ImportException, IOException {
```
honzasterba: confusing argName, its actualy output file path

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: ok, so I understood it wrongly, will rework

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: it is not / it is a path to .dot|.gv file from which we generate .png graph

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: fixed

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: fixed

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: fixed

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
koniecsveta: fixed as discussed

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,201 @@
...
+
+    @Override
+    public void run() throws Exception {
+        validateArgs();
+     
+        if (genModel instanceof TreeBackedMojoModel){
+            TreeBackedMojoModel treeBackedModel = (TreeBackedMojoModel) genModel;
+            ConvertTreeOptions options = new ConvertTreeOptions().withTreeConsistencyCheckEnabled();
+            final SharedTreeGraph g = treeBackedModel.convert(treeToPrint, null, options);
+            printPng(g);
```
honzasterba: since this supports all formats it can be called with format other than png, for that case we need to delegate to super.run()

h2o-genmodel-extensions/jgrapht/build.gradle
```diff
@@ -0,0 +1,19 @@
+
+group 'ai.h2o'
+version '3.29.0'
```
michalkurka: version shouldn't be there, group doesn't need to be there (imho)

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/SharedTreeSubgraph.java
```diff
@@ -131,9 +131,9 @@ void print() {
     rootNode.printEdges();
   }
 
-  void printDot(PrintStream os, int maxLevelsToPrintPerEdge, boolean detail, String optionalTitle, PrintMojo.PrintTreeOptions treeOptions) {
```
michalkurka: I think the parameter name is too specific to the purpose we are using it for. I would call it "isDirected" or something like that.

Also, I would perhaps preserve the original signature of the function as well (in case MLI uses it).

michalkurka:
@koniecsveta Code looks good, minor comments posted.

Can you please include the dependency graph of h2o-genmodel to this PR. Do we need to relocate any dependencies to avoid clashes with users libraries they might have on the classpath?

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,204 @@
...
+    
+    private void printPng(SharedTreeGraph trees) throws IOException, ImportException {
+        Path outputDirectoryPath = Paths.get(outputFileName);
+        int numberOfTrees = trees.subgraphArray.size();
+        if (numberOfTrees > 1) { 
+            if (outputFileName == null) {
+                outputFileName = Paths.get("").toString();
+            }
+            if (Files.exists(outputDirectoryPath) && !Files.isDirectory(outputDirectoryPath)) {
+                Files.delete(outputDirectoryPath);
```
honzasterba: we should never delete anything, overwriting things withtout asking is bad enough.
main reason: what if output dir would be current working directory? than this would fail, but why couldnt the output be the working dir?

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,204 @@
...
+            }
+            if (!Files.exists(outputDirectoryPath)) {
```
honzasterba: also if the output needs to be a directory and it is not, we should just fail immediately with an error

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+        for (SharedTreeSubgraph tree : trees.subgraphArray) {
+            Path dotSourceFilePath = Files.createTempFile("", tmpOutputFileName);
+            try (FileOutputStream fosTemp = new FileOutputStream(dotSourceFilePath.toFile()); PrintStream osTemp = new PrintStream(fosTemp)) {
+                tree.printDot(osTemp, maxLevelsToPrintPerEdge, detail, optionalTitle, pTreeOptions, true);
+                generateOutputPng(dotSourceFilePath, getPngName(numberOfTrees, tree.name));
+            }
+            Files.delete(dotSourceFilePath);
+        }
+
```
honzasterba: extra whitespace

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: this will not delete directory (...&& !Files.isDirectory(outputDirectoryPath)). If we want to create directory of pictures of trees and there is already file named like the directory to be created then this file is deleted so that directory can be created - as an analogy to overwriting the old file with the new one. But can be redone to just fail with an error.

koniecsveta:
> @koniecsveta Code looks good, minor comments posted.
> 
> Can you please include the dependency graph of h2o-genmodel to this PR. Do we need to relocate any dependencies to avoid clashes with users libraries they might have on the classpath?

+--- project :h2o-genmodel
|    +--- project :h2o-logger
|    +--- net.sf.opencsv:opencsv:2.3
|    +--- com.google.code.gson:gson:2.6.2
|    +--- ai.h2o:h2o-tree-api:0.3.9
|    \--- ai.h2o:deepwater-backend-api:1.0.4
+--- org.jgrapht:jgrapht-core:1.3.1
|    \--- org.jheaps:jheaps:0.10
+--- org.jgrapht:jgrapht-ext:1.3.1
|    +--- org.jgrapht:jgrapht-core:1.3.1 (*)
|    \--- com.github.vlsi.mxgraph:jgraphx:3.9.8.1
\--- org.jgrapht:jgrapht-io:1.3.1
     +--- org.jgrapht:jgrapht-core:1.3.1 (*)
     +--- org.antlr:antlr4-runtime:4.7.2
     \--- org.apache.commons:commons-text:1.6
          \--- org.apache.commons:commons-lang3:3.8.1

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
honzasterba: got it, thanks

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
honzasterba: hmm, this might necesarily not be true, if the output is a regular file than we can ovewrite it with a directory, so this would be reasonable to do

michalkurka:
@koniecsveta jira is marked for 3.28.0.2, the PR should be against rel-yu - please change the base

PR #4168, opened 2019-12-19
PUBDEV-7177: Make sure h2odriver gets correct version of Hadoop dependencies

michalkurka:
The issue is that declaring hadoop-hdfs-client dependency causes to use default Hadoop libraries, not the specific Hadoop libraries for the given distribution

For now fixed for HDP 2.6

### DIFF
--- a/h2o-hadoop-2/assemblyjar.gradle
+++ b/h2o-hadoop-2/assemblyjar.gradle
@@ -6,12 +6,30 @@ description = 'H2O HDFS client shadowjar for Hadoop ' + hadoopVersion
 sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
+def hasCustomHdfsDep = ext.has("hdfsDependency") && (ext.get("hdfsDependency") != defaultHdfsDependency)
+
+if (hasCustomHdfsDep) {
+  configurations {
+    compile.exclude group: "org.apache.hadoop", module: "${defaultHdfsDependency}"
+  }
+}
+
 dependencies {
+  compile ("org.apache.hadoop:hadoop-client:$hadoopMavenArtifactVersion") {
+    force = true
+  }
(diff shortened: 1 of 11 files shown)


h2o-hadoop-2/h2o-hdp2.6-assembly/build.gradle
```diff
@@ -1,6 +1,7 @@
 ext {
   hadoopVersion = 'hdp2.6'
-  hadoopMavenArtifactVersion = '2.7.3.2.6.1.0-129'
+  hadoopMavenArtifactVersion = '2.7.3.2.6.5.196-1'
+  hdfsDependency = "org.apache.hadoop:hadoop-hdfs"
```
honzasterba: this should be just `hadoop-hdfs` no?
also maybe it deservers a comment?

h2o-hadoop-2/assemblyjar.gradle
```diff
@@ -6,12 +6,30 @@ description = 'H2O HDFS client shadowjar for Hadoop ' + hadoopVersion
 sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
+def hasCustomHdfsDep = ext.has("hdfsDependency") && (ext.get("hdfsDependency") != defaultHdfsDependency)
+
+if (hasCustomHdfsDep) {
+  configurations {
+    compile.exclude group: "org.apache.hadoop", module: "${defaultHdfsDependency}"
```
honzasterba: unnecesary "${}"

h2o-hadoop-2/h2o-hdp2.6-assembly/build.gradle
michalkurka: thanks for catching that, I will fix

PR #4170, opened 2019-12-19
[PUBDEV-6762] automate execution of testdir_hdfs tests

honzasterba:
step one: create an image and add the first test


docker/Jenkinsfile-build-hadoop-docker
```diff
@@ -132,3 +131,63 @@ parallel(pipelineContext.getBuildConfig().getSupportedHadoopDistributions().coll
...
+                             usernamePassword(credentialsId: "${params.dockerRegistry}", usernameVariable: 'REGISTRY_USERNAME', passwordVariable: 'REGISTRY_PASSWORD')]) {
+                dir("${imageName}.${version}") {
+                    pipelineContext.getUtils().unstashFiles(this, DOCKER_STASH)
+                    sh """
+                                    cd docker
+
+                                    docker build \
+                                        -t ${REGISTRY_PREFIX}/${imageName}:${version} \
+                                        -f hadoop/slave/Dockerfile \
+                                        --build-arg PATH_PREFIX=hadoop/slave \
```
michalkurka: I think "slave" is a misleading name, could it be edge_node instead or similar?

docker/hadoop/slave/conf/core-site.xml
```diff
@@ -0,0 +1,149 @@
+<!--Wed Dec  4 13:10:01 2019-->
```
michalkurka: I don't think this file should be in this repository, it is better to automatically download a fresh copy of the hadoop conf directory a mount it in the docker container in every single run and thus make it fully dynamic.

michalkurka:
@honzasterba I very much like this approach, the only thing I would suggest to change is to avoid including the Hadoop configuration in the git repo

docker/Jenkinsfile-build-hadoop-docker
honzasterba: was not happy about the name too, will fix

docker/hadoop/slave/conf/core-site.xml
honzasterba: sounds good, will do

honzasterba:
it works!
http://mr-0xc1:8080/view/H2O-3/job/h2o-hadoop-multinode-pipeline/

PR #4171, opened 2019-12-19
Prevent Jenkins build issues by allowing a dynamic port for client generator

michalkurka:
Currently Jenkins fails oftentimes with

On /127.0.0.1 some of the required ports 48000, 48001 are not available, change -port PORT and try again.

This prevents nightly builds from running

In this PR we allow the port to be dynamic and avoid port conflicts

### DIFF
--- a/scripts/run.py
+++ b/scripts/run.py
@@ -263,7 +263,7 @@ class H2OCloudNode(object):
 
     def __init__(self, is_client, allow_clients,
                  cloud_num, nodes_per_cloud, node_num, cloud_name, h2o_jar, ip, base_port,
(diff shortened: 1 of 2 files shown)


scripts/run.py
```diff
@@ -281,6 +281,7 @@ def __init__(self, is_client, allow_clients,
         :param ldap_config_path: path to LDAP config, if none, no LDAP will be used.
         :param jvm_opts: str with additional JVM options.
         :param flatfile: path to flatfile (optional) 
+        :param strict_port: interpret port as exact specification, otherwise as a base port (optional, default is strict)
```
honzasterba: do I understand it correctly that this change is a no-op and more will follow?

honzasterba: oh now I get it!

PR #4172, opened 2019-12-20
[PUBDEV-7029] Add xgboost offset column

honzasterba:


michalkurka:
@honzasterba this looks good but I don't see any changes in MOJOs

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/predict/XGboostJavaBigScoreChunkPredict.java
```diff
@@ -24,8 +24,6 @@ public XGboostJavaBigScoreChunkPredict(DataInfo di, XGBoostOutput output, double
 
   @Override
   public double[] score0(Chunk[] chks, double offset, int row_in_chunk, double[] tmp, double[] preds) {
-    if (offset != 0) throw new UnsupportedOperationException("Unsupported: offset != 0");
```
michalkurka: offset doesn't seem to be considered

h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostNativeMojoModel.java
```diff
@@ -38,19 +39,24 @@ private static Booster makeBooster(byte[] boosterBytes) {
...
-                                double _defaultThreshold, final boolean _sparse) {
-
+  public static double[] score0(
+      double[] doubles, double offset, double[] preds,
+      String _boosterType, int _ntrees, 
+      Booster _booster, int _nums, int _cats,
+      int[] _catOffsets, boolean _useAllFactorLevels,
+      int nclasses, double[] _priorClassDistrib,
+      double _defaultThreshold, final boolean _sparse, 
+      double _mojo_version
```
michalkurka: can we avoid making it a parameter? for in H2O execution, we don't really need to add the version

you can move the check to the non-static method (line 42) and throw the message there

honzasterba: 👍 this was lazy of me

honzasterba: fixed

PR #4173, opened 2019-12-20
PUBDEV-7165 - Expose Spearman's Rho to Py/R

Pscheidl:
[PUBDEV-7165](https://0xdata.atlassian.net/browse/PUBDEV-7165)

I've implemented Spearman's correlation coefficient before. However, the coefficient has been implemented as a simple Rapids expression, calculating SCC only for two columns.

This PR aims to integrate such functionality into existing `h2o.corr` Py/R function, which does correlation coefficients on `Frame` level.

- `AstSpearman is removed` - is this safe to do ? It will be part of AstCorrelation, yet reside in another class.

Testing the results against `numpy`, they're exactly the same

```
 NumPy:
         C1        C2        C3        C4
0  1.000000 -0.172003  0.882409  0.827453
1 -0.172003  1.000000 -0.328611 -0.306333
2  0.882409 -0.328611  1.000000  0.933388
3  0.827453 -0.306333  0.933388  1.000000
 H2O:
         C1        C2        C3        C4
0  1.000000 -0.172003  0.882409  0.827453
1 -0.172003  1.000000 -0.328611 -0.306333
2  0.882409 -0.328611  1.000000  0.933388
3  0.827453 -0.306333  0.933388  1.000000

```
Spearman's Rho is calculated such as:
rho

However, there is a simplified version used by many (e.g R) sensitive to repeating values:

rho_simplified

If there are repeated values (or heavily repeated), it might produce a slightly different result. In Python, the calculation is exact, thus the results are 100% the same. In R, there is a small delta. Results of `cor` function in R in Spearman mode are equal to manual calculation with the simplified formula.

In general, it's the same equation as Pearson's, using labeled columns instead of original values.


h2o-py/tests/testdir_munging/unop/pyunit_cor.py
```diff
@@ -9,22 +8,18 @@
...
 
 def cor_test():
-
-
-
     iris_h2o = h2o.import_file(path=pyunit_utils.locate("smalldata/iris/iris.csv"))
     iris_np = np.genfromtxt(pyunit_utils.locate("smalldata/iris/iris.csv"),
                             delimiter=',',
-                            skip_header=1,
+                            skip_header=-1,
```
Pscheidl: This was a bug. It skipped the first line. Therefore the `assert max < .006, "expected equal correlations"`. Now, the difference is 0 !

Pscheidl:
@michalkurka @koniecsveta Can be roughly reviewed to see how I did the implementation itself. We've our own implementation of Spearman, even though we could only rank the columns and use existing Pearson algorithm that is already implemented. However, Existing Pearson does unnecessary vec copying in vain. Therefore, it has been implemented in a separate MRTask. Can be used afterwards to speed up and make Pearson more effective as well.

h2o-py/h2o/frame.py
```diff
@@ -3247,7 +3247,7 @@ def sd(self, na_rm=False):
         return ExprNode("sd", self, na_rm)._eager_scalar()
 
 
-    def cor(self, y=None, na_rm=False, use=None):
+    def cor(self, y=None, na_rm=False, use=None, measure="Pearson"):
```
Pscheidl: Rename measure parameter to match R `cor` naming.

h2o-core/src/main/java/water/rapids/Env.java
```diff
@@ -211,7 +211,6 @@ static void init(AstPrimitive ast, String name) {
     init(new AstTable());
     init(new AstUnique());
     init(new AstVariance());
-    init(new AstSpearman());
```
honzasterba: this smells little bit like breaking api change? could someone be affected by removing Spearman support from rapids?

h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
```diff
@@ -1,72 +1,132 @@
...
+import java.util.Objects;
+
+public class SpearmanCorrelation {
+
+  public static Frame calculate(final Frame frameX, Frame frameY, final AstCorrelation.Mode mode) {
+    Objects.requireNonNull(frameX);
+    Objects.requireNonNull(frameY);
+
+    final Frame correlationMatrix = createCorrelationMatrix(frameX);
+    final boolean framesAreEqual = frameX == frameY; // TODO: Client APIs create new frames with no keys to compare
```
honzasterba: how is this todo related to this functionality?

h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
```diff
@@ -1,72 +1,132 @@
...
+
+    checkCorrelationDoable(frameX, frameY, mode);
+
+    for (int vecIdX = 0; vecIdX < frameX.numCols(); vecIdX++) {
+      for (int vecIdY = 0; vecIdY < frameX.numCols(); vecIdY++) {
```
honzasterba: should there be frameY.numCols, I know they are same, in which case I would make that more explicit, or maybe this is fine, just caught my eye as a bug at first look
maybe checkCorrelationDoable() can return ncols and that can be used here?

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstCorrelation.java
```diff
@@ -59,7 +62,40 @@ public Val apply(Env env, Env.StackHelp stk, AstRoot asts[]) {
...
+        return fry.numRows() == 1 ? scalar(frx, fry, mode) : array(frx, fry, mode);
+      case Spearman:
+        return spearman(frx, fry, mode);
+      default:
+        throw new IllegalStateException(String.format("Given method input'%s' is not supported. Available options are: %s",
+                method, Arrays.toString(Method.values())));
+    }
+  }
+
+  private static Method getMethodFromUserInput(final String methodUserInput) {
```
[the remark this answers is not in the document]
Pscheidl: Thank you for catching this. The whole method's logic is wrong. I wrote the method separately to match in a case-insensitive way, but never really "implemented" it. There are multiple problems with this method.

1. I ignored my own code in `EnumUtils`, specifically method `valueOfIgnoreCase`, which does what I need.
2. We really need to match in a case-insensitive way. In R, the `cor` function uses all lower-case, while in Python, there are different styles. One solution is to convert the string on the client (Py/R/Rapids written directly ?), but this is a simpler and more centralized solution. I don't see a reason to punish our users for case-related mistakes. No trimming, no guessing, just case-related mistakes. We do this on many places.



Using EnumUtils instead. Also, made `EnumUtills.valueOfIgnoreCase` return Optional, which required some minor "enchantments" in non-related code (three occurances basically).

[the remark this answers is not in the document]
Pscheidl: The method is now so small it could be removed and it's content used directly, IDK. Matter of code style I guess. Looks more obvious from the main method I guess.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstCorrelation.java
```diff
@@ -24,11 +25,13 @@
...
 
-  private enum Mode {Everything, AllObs, CompleteObs}
+  protected enum Mode {Everything, AllObs, CompleteObs}
+
+  private enum Method {Pearson, Spearman}
 
   @Override
   public int nargs() {
-    return 1 + 3; /* (cor X Y use) */
+    return 1 + 4; /* (cor X Y use) */
```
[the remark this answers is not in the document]
Pscheidl: Thank you. Also the method `args` is outdated.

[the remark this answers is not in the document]
Pscheidl: Corrected both.

h2o-core/src/main/java/water/rapids/Env.java
Pscheidl: 1. It has been there since `3.28.0.1`, where I implemented it only as Rapids. Which means if there is someone using Rapids directly, it could break their code without javac catching it (all plain strings).
2. As @michalkurka told me, users are not meant to write Rapids expressions on their own, those are assembled by our API. Which means to me this contract is not guaranteed, structure of Rapids is not guaranteed and it can be joined together with `cor` Rapids.

Pscheidl: @michalkurka WDYT ? I ask the very same question in the description of this PR. I'm for removing the old way myself.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
Pscheidl: Thank you. Fixed.

## Why the TODO ?
This piece of code was effectively not working. The  variable `framesAreEqual` is used for optimization on the diagonal of the correlation matrix - if the vectors compared are guaranteed to be the same, the correlation is automatically `1` without any calculations required. With one exception, which is `AstCorrelation.Mode.Everything` - this mode still requires to inspect all the observations and output NaN if there is an observation with a single NaN found.

Rapids will create two "distinct" frames (object address is distinct), but with the same content in case the correlation is created within one frame. Frame X and Frame Y will contain the same vecs, but object-wise, they'll be different frames. Therefore, `equals` or direct comparison with `==` won't work.

## Resolution

This has been resolved by introducing `framesContainSameVecs`, which compares the vectors in a Frame key-wise. And if they're the same, this optimization can happen.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
Pscheidl: Thank you, it would be useful indeed to be more expressive about that.

I chose the way of doing `frameY.numCols()` in the inner iteration.

Also added `assert frameX.numCols() == frameY.numCols();` at the very beginning of the method after null checks to make it clear what is expected.

h2o-core/src/main/java/water/api/ImportSQLTableHandler.java
```diff
@@ -20,7 +20,8 @@ public JobV3 importSQLTable(int version, ImportSQLTableV99 importSqlTable) {
     if (importSqlTable.fetch_mode == null) {
       sqlFetchMode = SqlFetchMode.DISTRIBUTED;
     } else {
-      sqlFetchMode = EnumUtils.valueOfIgnoreCase(SqlFetchMode.class, importSqlTable.fetch_mode);
+      sqlFetchMode = EnumUtils.valueOfIgnoreCase(SqlFetchMode.class, importSqlTable.fetch_mode)
```
honzasterba: I would do something similar to what you are doing in getMethodFromUserInput() i.e. include the invalid value in the exception message

h2o-core/src/main/java/water/rapids/Env.java
honzasterba: I think the core part here is `users are not meant to write Rapids expressions` if that is true than this is fine

h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
```diff
@@ -1,72 +1,155 @@
...
 import java.math.BigDecimal;
 import java.math.MathContext;
+import java.util.Objects;
+
+public class SpearmanCorrelation {
+
+  public static Frame calculate(final Frame frameX, Frame frameY, final AstCorrelation.Mode mode) {
+    Objects.requireNonNull(frameX);
+    Objects.requireNonNull(frameY);
+    assert frameX.numCols() == frameY.numCols();
```
honzasterba: this duplicates the assert in checkCorrelationDoable

h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
```diff
@@ -1,72 +1,155 @@
...
+   * @throws IllegalArgumentException When the {@link AstCorrelation.Mode} is set to AllObs and any of the vectors contains NaN
+   */
+  private static void checkCorrelationDoable(final Frame frameX, final Frame frameY, final AstCorrelation.Mode mode)
+          throws IllegalArgumentException {
+    if (!AstCorrelation.Mode.AllObs.equals(mode)) return;
+
+    final Vec[] vecsX = frameX.vecs();
+    final Vec[] vecsY = frameY.vecs();
+
+    assert vecsX.length == vecsY.length;
```
honzasterba: should this really be an assert? is this checked somewhere else so it will never fail here? if user input can make this assert fail it should be an if () with an exception explaining the error to the user

h2o-core/src/main/java/water/util/PojoUtils.java
```diff
@@ -524,8 +524,10 @@ else if (f.getType() == float.class && (value instanceof Number))
           Enum[] valuesCast = (Enum[]) Array.newInstance(f.getType().getComponentType(), valuesTyped.length);
           for (int i = 0; i < valuesTyped.length; i++) {
             String v = (String) valuesTyped[i];
-            Enum enu = EnumUtils.valueOfIgnoreCase((Class<Enum>)f.getType().getComponentType(), v);
-            if (enu == null) throw new IllegalArgumentException("Field = " + fieldName + " element cannot be set to value = " + value);
+            final IllegalArgumentException fieldNotSettablException = new IllegalArgumentException("Field = " + fieldName + " element cannot be set to value = " + value);
```
honzasterba: personally I am against instantiating exception when they are not really needed, I would move this inside the orElseThrow (same below)

Pscheidl: I'd need value to be final then, so I'll make a final copy of `value` locally.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
Pscheidl: Removed the one from the check.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
Pscheidl: Truth be told, this rule should not even be applied. Having a non-square matrix is perfectly fine here. I'll add a test.

Pscheidl: Resolved in https://github.com/h2oai/h2o-3/pull/4173/commits/43759b10b4a76748073720233d94c1591fe47837 - non-squared correlation matrix is a perfectly fine approach. Thank you.

Also:
- Moved the check `isCorrelationDoable` after the null checks not to allocate any resources before this check in vain.
- The `isCorrelationDoable` newly checks the frames given for correlation calculation are not empty (don't have 0 columns).

h2o-core/src/main/java/water/util/PojoUtils.java
Pscheidl: Luckily, there is `final Object[] valuesTyped = ((Object[]) value);`  just above, which can be made final. And as it's an array of Objects, I've used `Arrays.toString(valuesTyped)` at the end of the exception's message properly.

h2o-core/src/main/java/water/api/ImportSQLTableHandler.java
Pscheidl: Done. Changed to:

```java
.orElseThrow(() -> new IllegalArgumentException("Unrecognized SQL Fetch mode: " + importSqlTable.fetch_mode));
```

h2o-core/src/main/java/water/rapids/Env.java
michalkurka: this should be fine

Pscheidl:
There's a test failure. Fixing. http://mr-0xc1:8080/blue/organizations/jenkins/h2o-3-pipeline/detail/PR-4173/17/tests

Pscheidl:
Test ran & passed: http://mr-0xc1:8080/job/h2o-3-pipeline/job/PR-4173/19/artifact/r3.4-small-client-mode/h2o-3/h2o-r/tests/results/testdir_munging_unop_runit_cor_spearman.R.out.txt

Merging.

PR #4174, opened 2019-12-20
PUBDEV-7193: Fix a race condition in addVec

michalkurka:
Credit goes to https://github.com/SimonSchmid who found reported the issue and provided a reproducer. Thank you!

### DIFF
--- a/h2o-core/src/main/java/water/fvec/Vec.java
+++ b/h2o-core/src/main/java/water/fvec/Vec.java
@@ -1522,15 +1522,15 @@ public class Vec extends Keyed<Vec> {
      *  @author tomasnykodym   */
     private final static class AddVecs2GroupTsk extends TAtomic<VectorGroup> {
       final Key _key;
-      int _n;          // INPUT: Keys to allocate; OUTPUT: start of run of keys
+      final int _n; // INPUT: Keys to allocate; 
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/fvec/Vec.java
```diff
@@ -1539,7 +1539,7 @@ static private Key init_key() {
     public int reserveKeys(final int n) {
       AddVecs2GroupTsk tsk = new AddVecs2GroupTsk(_key, n);
       tsk.invoke(_key);
-      return tsk._n;
```
sebhrusen: @michalkurka changes make sense in terms of clarity, but still not 100% clear to me why it was racing here:  there's one `tsk` per thread, the only object that can be modified concurrently here is the `VecGroup`, correct?

michalkurka: The issue here is that the atomic operation should not mutate any state that is not purely used for output. In this case, `_n` is mutated (old._len) is stored, however, when the automatic operation fails and is re-run we lost the original `_n` and the task no longer know how many keys to allocate! Very ugly.

sebhrusen: > when the atomic operation fails (due to racy update) and is re-run...

Oh, I see, makes sense then, thanks!

PR #4176, opened 2019-12-31
PUBDEV-7174 Fix simple typo: wether -> whether

timgates42:
There is a small typo in h2o-py/h2o/frame.py.
Should read `whether` rather than `wether`.

Jira: PUBDEV-7174


angela0xdata:
@timgates42 Thank you!

PR #4177, opened 2020-01-02
Execute most of testdir_hdfs tests

honzasterba:


h2o-hadoop-common/tests/python/pyunit_connect_invalid.py
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 # -*- encoding: utf-8 -*-
-import sys, os
-sys.path.insert(1, os.path.join("..", "..", "..", "h2o-py"))
+import sys
+import os
+sys.path.insert(1, os.path.join("../../../h2o-py"))
```
Pscheidl: Aren't `"..","..",".."` instead of just doing `("../../../h2o-py")` used because python will insert either `\` or `/` depending on OS and thus keeping OS compatibility ?

https://docs.python.org/2/library/os.path.html
and specifically: https://docs.python.org/2/library/os.path.html#os.path.join

Judging from behavior observed.

honzasterba: fixed

PR #4180, opened 2020-01-03
PUBDEV-6993: Fix link to SW topic in the Java FAQ

angela0xdata:


h2o-docs/src/product/faq/java.rst
```diff
@@ -15,7 +15,7 @@ To see how the REST API is used with H2O:
 
 -  Write an R program for H2O using the H2O R package that uses ``h2o.startLogging()`` at the beginning. All REST API calls used are logged.
 
-The second way to use H2O with Java is to embed H2O within your Java application, similar to `Sparkling Water <https://github.com/h2oai/sparkling-water/blob/master/DEVEL.rst>`__.
+The second way to use H2O with Java is to embed H2O within your Java application, similar to `Sparkling Water <https://github.com/h2oai/sparkling-water/blob/master/doc/src/site/sphinx/configuration/configuration_properties.rst>`__.
```
maurever: The link is working now. However, I am not sure how it is useful for users. There is only the SW parameter description, but nothing more general how to use SW in Java. Is there any other SW page where embedding of SW to Java is better described?

angela0xdata: @maurever That's the closest I could find to anything describing embedding H2O within Java. I thought about just taking the link out.

maurever: Well, we can merge this PR and create a new JIRA to improve our doc about embedding h2o within Java. Some guys asked me for it, and sadly, there is no documentation/demo/instructions about it.

angela0xdata: Going with plan C to fix this. I'll remove that "similar to Sparkling Water" phrase and instead include an example.

PR #4181, opened 2020-01-03
PUBDEV-6142: Remove ignored_columns from list of available parameters in AutoML

angela0xdata:


angela0xdata:
will submit PR for rel-yu branch

PR #4186, opened 2020-01-06
[PUBDEV-7164] XGBoost perfomance fixes

honzasterba:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -213,7 +221,7 @@
   }
 
   @Override
-  public ModelBuilder getModelBuilder() {
+  public XGBoost getModelBuilder() {
```
honzasterba: just to fix warning about generics

PR #4191, opened 2020-01-07
PUBDEV-7110: Early stopping logic should not be located within HyperSpaceIterator's hasNext method

deil87:
This PR is a reopened version of https://github.com/h2oai/h2o-3/pull/4114 PR as it had been merged before all concerns were addressed

### DIFF
--- a/h2o-core/src/main/java/hex/ParallelModelBuilder.java
+++ b/h2o-core/src/main/java/hex/ParallelModelBuilder.java
@@ -19,7 +19,7 @@ public class ParallelModelBuilder extends ForkJoinTask<ParallelModelBuilder> {
 
   public static abstract class ParallelModelBuilderCallback<D extends ParallelModelBuilderCallback> extends Iced<D> {
 
-    public abstract void onBuildSuccess(final Model model, final ParallelModelBuilder parallelModelBuilder);
+    public abstract void onBuildSucces(final Model model, final ParallelModelBuilder parallelModelBuilder);
 
     public abstract void onBuildFailure(final ModelBuildFailure modelBuildFailure, final ParallelModelBuilder parallelModelBuilder);
 
@@ -67,7 +67,7 @@ public class ParallelModelBuilder extends ForkJoinTask<ParallelModelBuilder> {
     @Override
     public void onModelSuccess(Model model) {
       try {
-        _callback.onBuildSuccess(model, ParallelModelBuilder.this);
+        _callback.onBuildSucces(model, ParallelModelBuilder.this);
       } finally {
         _modelCompletedCounter.incrementAndGet();
         _modelInProgressCounter.decrementAndGet();
--- a/h2o-core/src/main/java/hex/ParallelModelBuilder.java
+++ b/h2o-core/src/main/java/hex/ParallelModelBuilder.java
@@ -2,10 +2,11 @@ package hex;
 
 import jsr166y.ForkJoinTask;
 import water.Iced;
-import water.util.IcedAtomicInt;
 
-import java.util.*;
+import java.util.Collection;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Dispatcher for parallel model building. Starts building models every time the run method is invoked.
@@ -18,13 +19,19 @@ public class ParallelModelBuilder extends ForkJoinTask<ParallelModelBuilder> {
 
   public static abstract class ParallelModelBuilderCallback<D extends ParallelModelBuilderCallback> extends Iced<D> {
 
-    public abstract void onBuildSucces(final Model model, final ParallelModelBuilder parallelModelBuilder);
+    public abstract void onBuildSuccess(final Model model, final ParallelModelBuilder parallelModelBuilder);
 
     public abstract void onBuildFailure(final ModelBuildFailure modelBuildFailure, final ParallelModelBuilder parallelModelBuilder);
+
+    public abstract void afterBuildProcessing(final ParallelModelBuilder parallelModelBuilder, final Model model);
   }
 
   private final transient ParallelModelBuilderCallback _callback;
-  private final transient IcedAtomicInt _modelInProgressCounter = new IcedAtomicInt();
+
+  private final transient AtomicInteger _modelInProgressCounter = new AtomicInteger();
+
+  private final transient AtomicInteger _modelCompletedCounter = new AtomicInteger();
+
   private final transient AtomicBoolean _completed = new AtomicBoolean(false);
   private final transient ParallelModelBuiltListener _parallelModelBuiltListener;
 
@@ -50,30 +57,37 @@ public class ParallelModelBuilder extends ForkJoinTask<ParallelModelBuilder> {
       }
   }
 
+  public boolean hasReachedMaxModels(int maxModels) {
+    return !(maxModels == 0 || _modelCompletedCounter.get() + _modelInProgressCounter.get() < maxModels);
+  }
+
 
   private class ParallelModelBuiltListener extends ModelBuilderListener<ParallelModelBuiltListener> {
 
     @Override
     public void onModelSuccess(Model model) {
       try {
-        _callback.onBuildSucces(model, ParallelModelBuilder.this);
+        _callback.onBuildSuccess(model, ParallelModelBuilder.this);
       } finally {
+        _modelCompletedCounter.incrementAndGet();
         _modelInProgressCounter.decrementAndGet();
+
+        _callback.afterBuildProcessing(ParallelModelBuilder.this, model);
       }
       attemptComplete();
     }
 
     @Override
     public void onModelFailure(Throwable cause, Model.Parameters parameters) {
+      ModelBuildFailure modelBuildFailure = new ModelBuildFailure(cause, parameters);
       try {
-        final ModelBuildFailure modelBuildFailure = new ModelBuildFailure(cause, parameters);
         _callback.onBuildFailure(modelBuildFailure, ParallelModelBuilder.this);
       } finally {
         _modelInProgressCounter.decrementAndGet();
       }
+      _callback.afterBuildProcessing(ParallelModelBuilder.this, null);
       attemptComplete();
     }
-    
   }
 
   /**
@@ -108,7 +122,6 @@ public class ParallelModelBuilder extends ForkJoinTask<ParallelModelBuilder> {
     if(!_completed.get() || _modelInProgressCounter.get() != 0) return;
     complete(this);
   }
-
 
   @Override
   public ParallelModelBuilder getRawResult() {
--- a/h2o-core/src/main/java/hex/grid/GridSearch.java
+++ b/h2o-core/src/main/java/hex/grid/GridSearch.java
@@ -182,7 +182,7 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
     }
 
     @Override
-    public void onBuildSuccess(final Model finishedModel, final ParallelModelBuilder parallelModelBuilder) {
+    public void onBuildSucces(final Model finishedModel, final ParallelModelBuilder parallelModelBuilder) {
       try {
         parallelSearchGridLock.lock();
         constructScoringInfo(finishedModel);
--- a/h2o-core/src/main/java/hex/grid/GridSearch.java
+++ b/h2o-core/src/main/java/hex/grid/GridSearch.java
@@ -121,7 +121,7 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
     long gridWork=0;
     if (gridSize > 0) {//if total grid space is known, walk it all and count up models to be built (not subject to time-based or converge-based early stopping)
       int count=0;
-      while (it.hasNext(model) && (it.max_models() == 0 || count++ < it.max_models())) { //only walk the first max_models models, if specified
+      while (it.hasNext(model) && (it.max_models() > 0 && count++ < it.max_models())) { //only walk the first max_models models, if specified
         try {
           Model.Parameters parms = it.nextModelParameters(model);
           gridWork += (parms._nfolds > 0 ? (parms._nfolds+1/*main model*/) : 1) *parms.progressUnits();
--- a/h2o-core/src/main/java/hex/grid/GridSearch.java
+++ b/h2o-core/src/main/java/hex/grid/GridSearch.java
@@ -121,7 +121,7 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
     long gridWork=0;
     if (gridSize > 0) {//if total grid space is known, walk it all and count up models to be built (not subject to time-based or converge-based early stopping)
       int count=0;
-      while (it.hasNext(model) && (it.max_models() > 0 && count++ < it.max_models())) { //only walk the first max_models models, if specified
+      while (it.hasNext(model) && (it.max_models() == 0 || count++ < it.max_models())) { //only walk the first max_models models, if specified
         try {
           Model.Parameters parms = it.nextModelParameters(model);
           gridWork += (parms._nfolds > 0 ? (parms._nfolds+1/*main model*/) : 1) *parms.progressUnits();
@@ -346,12 +346,9 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
       // Get iterator to traverse hyper space
       HyperSpaceWalker.HyperSpaceIterator<MP> it = _hyperSpaceWalker.iterator();
 
-      // This counter will be used as index to generate keys for corresponding `successful` models
-      int numberOfBuiltModels = grid.getModelCount();
+      int indexOfLastBuiltModel = grid.getModelCount();
 
-      int maxNumberOfModelsToBuild = numberOfBuiltModels + it.max_models();
-
-      while (it.hasNext(model) && (it.max_models() == 0 || grid.getModelCount() < maxNumberOfModelsToBuild)) {
+      while (it.hasNext(model) && (it.max_models() == 0 || grid.getModelCount() < it.max_models())) {
         if (_job.stop_requested()) throw new Job.JobCancelledException();  // Handle end-user cancel request
 
         MP params = null;
@@ -369,7 +366,7 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
             scoringInfo.time_stamp_ms = System.currentTimeMillis();
 
             //// build the model!
-            model = buildModel(params, grid, ++numberOfBuiltModels, protoModelKey);
+            model = buildModel(params, grid, ++indexOfLastBuiltModel, protoModelKey);
             if (model != null) {
               model.fillScoringInfo(scoringInfo);
               grid.setScoringInfos(ScoringInfo.prependScoringInfo(scoringInfo, grid.getScoringInfos()));
--- a/h2o-core/src/main/java/hex/grid/GridSearch.java
+++ b/h2o-core/src/main/java/hex/grid/GridSearch.java
@@ -176,9 +176,6 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
     private final Grid grid;
     private final Lock parallelSearchGridLock = new ReentrantLock();
 
-    // Is used to make sure that counters for inProgress and completed models are consistent with availability of model parameters in an iterator
-    private final Lock getNextModelParametersLock = new ReentrantLock();
-
     public ModelFeeder(HyperSpaceWalker.HyperSpaceIterator<MPF> hyperspaceIterator, Grid grid) {
       this.hyperspaceIterator = hyperspaceIterator;
       this.grid = grid;
@@ -218,7 +215,7 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
 
     private void attemptBuildNextModel(final ParallelModelBuilder parallelModelBuilder, final Model previousModel) {
 
-      getNextModelParametersLock.lock();
+      parallelSearchGridLock.lock();
       try {
         if (hyperspaceIterator.hasNext(previousModel)
                 && isThereEnoughTime()
@@ -235,7 +232,7 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
           parallelModelBuilder.noMoreModels();
         }
       } finally {
-        getNextModelParametersLock.unlock();
+        parallelSearchGridLock.unlock();
       }
     }
 
--- a/h2o-core/src/main/java/hex/grid/GridSearch.java
+++ b/h2o-core/src/main/java/hex/grid/GridSearch.java
@@ -10,7 +10,9 @@ import water.fvec.Frame;
 import water.util.Log;
 import water.util.PojoUtils;
 
-import java.io.*;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.util.*;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -168,19 +170,19 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
     return _hyperSpaceWalker.getMaxHyperSpaceSize();
   }
 
-  private class ModelFeeder<MP extends Model.Parameters, D extends ModelFeeder> extends ParallelModelBuilder.ParallelModelBuilderCallback<D>{
+  private class ModelFeeder<MPF extends Model.Parameters, D extends ModelFeeder> extends ParallelModelBuilder.ParallelModelBuilderCallback<D>{
 
-    private final HyperSpaceWalker.HyperSpaceIterator<MP> hyperspaceIterator;
+    private final HyperSpaceWalker.HyperSpaceIterator<MPF> hyperspaceIterator;
     private final Grid grid;
     private final Lock parallelSearchGridLock = new ReentrantLock();
 
-    public ModelFeeder(HyperSpaceWalker.HyperSpaceIterator<MP> hyperspaceIterator, Grid grid) {
+    public ModelFeeder(HyperSpaceWalker.HyperSpaceIterator<MPF> hyperspaceIterator, Grid grid) {
       this.hyperspaceIterator = hyperspaceIterator;
       this.grid = grid;
     }
 
     @Override
-    public void onBuildSucces(final Model finishedModel, final ParallelModelBuilder parallelModelBuilder) {
+    public void onBuildSuccess(final Model finishedModel, final ParallelModelBuilder parallelModelBuilder) {
       try {
         parallelSearchGridLock.lock();
         constructScoringInfo(finishedModel);
@@ -193,7 +195,6 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
       } finally {
         parallelSearchGridLock.unlock();
       }
-      attemptBuildNextModel(parallelModelBuilder, finishedModel);
     }
 
     @Override
@@ -205,28 +206,30 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
       } finally {
         parallelSearchGridLock.unlock();
       }
-      attemptBuildNextModel(parallelModelBuilder, null);
+    }
+
+    @Override
+    public void afterBuildProcessing(ParallelModelBuilder parallelModelBuilder, Model finishedModel) {
+      attemptBuildNextModel(parallelModelBuilder, finishedModel);
     }
 
     private void attemptBuildNextModel(final ParallelModelBuilder parallelModelBuilder, final Model previousModel) {
-      // Attempt to train next model
-      try {
-        parallelSearchGridLock.lock();
-        final MP nextModelParams = getNextModelParams(hyperspaceIterator, previousModel, grid);
-        if (nextModelParams != null
+
+      synchronized (hyperspaceIterator) {
+        if (hyperspaceIterator.hasNext(previousModel)
                 && isThereEnoughTime()
+                && !parallelModelBuilder.hasReachedMaxModels(hyperspaceIterator.max_models())
                 && !_job.stop_requested()
                 && !_hyperSpaceWalker.stopEarly(previousModel, grid.getScoringInfos())) {
 
-          // FIXME: ModelFeeder's <MP> type parameter is hiding GridSearch's one. Ideally would have moved ModelFeeder outside but it is too coupled with GridSearch class.
+          final MPF nextModelParams = getNextModelParams(hyperspaceIterator, previousModel, grid);
+
           reconcileMaxRuntime(grid._key, hyperspaceIterator, nextModelParams);
 
           parallelModelBuilder.run(Collections.singletonList(ModelBuilder.make(nextModelParams)));
         } else {
           parallelModelBuilder.noMoreModels();
         }
-      } finally {
-        parallelSearchGridLock.unlock();
       }
     }
 
@@ -249,12 +252,12 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
       return enoughTime;
     }
 
-    private MP getNextModelParams(final HyperSpaceWalker.HyperSpaceIterator<MP> hyperSpaceWalker, final Model model, final Grid grid){
-      MP params = null;
+    private MPF getNextModelParams(final HyperSpaceWalker.HyperSpaceIterator<MPF> hyperSpaceIterator, final Model model, final Grid grid){
+      MPF params = null;
 
       while (params == null) {
-        if (hyperSpaceWalker.hasNext(model)) {
-          params = hyperSpaceWalker.nextModelParameters(model);
+        if (hyperSpaceIterator.hasNext(model)) {
+          params = hyperSpaceIterator.nextModelParameters(model);
           final Key modelKey = grid.getModelKey(params.checksum(IGNORED_FIELDS_PARAM_HASH));
           if(modelKey != null){
             params = null;
@@ -280,7 +283,9 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
 
     List<ModelBuilder> startModels = new ArrayList<>();
 
-    int numberOfInitialModels = iterator.max_models() == 0 ? _parallelism : Math.min(_parallelism, iterator.max_models());
+    int maxHyperSpaceSize = (int) _hyperSpaceWalker.getMaxHyperSpaceSize();
+    int maxModelsBasedOnParallelism = iterator.max_models() == 0 ? _parallelism : Math.min(_parallelism, iterator.max_models());
+    int numberOfInitialModels = Math.min(maxHyperSpaceSize, maxModelsBasedOnParallelism);
     final List<MP> mps = initialModelParameters(numberOfInitialModels, iterator);
 
     for (int i = 0; i < mps.size(); i++) {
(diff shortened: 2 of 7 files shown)


deil87:
@honzasterba as a follow-up to your comment https://github.com/h2oai/h2o-3/pull/4114#discussion_r363746393 -> yes, `ADAPTIVE_PARALLELISM_LEVEL` should have been used in `getParallelismLevel` but first, there were no usages and second, there is a method `getAdaptiveParallelism` which dynamically computes level. My guess was that the method has substituted need for the constant. Was I correct @Pscheidl that we don't need this constant anymore?

Pscheidl:
My method `getParallelismLevel` could be improved such that `if (parallelism == 0) ` is changed to `if (parallelism == ADAPTIVE_PARALLELISM_LEVEL) ` - This was original meaning of the constant, only forgot to use it.

Otherwise no, we don't need it.

deil87:
@honzasterba as a follow-up to your comment https://github.com/h2oai/h2o-3/pull/4114#discussion_r363748065 -> no, they are not the same from the compiler's point of view.
In both `GridSearch` and `ModelFeeder` classes `MP` generic type parameter is being defined ( it is not being passed over from outer class to inner `ModelFeeder` class )
Here is also warning from IDE about this:
image

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -205,26 +208,40 @@ public void onBuildFailure(final ParallelModelBuilder.ModelBuildFailure modelBui
...
+                && !hasReachedMaxModels(parallelModelBuilder)
+                && !_job.stop_requested()
+                && !_hyperSpaceWalker.stopEarly(previousModel, grid.getScoringInfos())) {
 
-        parallelModelBuilder.run(Collections.singletonList(ModelBuilder.make(nextModelParams)));
-      } else {
-        parallelModelBuilder.noMoreModels();
+          final MP nextModelParams = getNextModelParams(hyperspaceIterator, previousModel, grid);
+
+          // FIXME: ModelFeeder's <MP> type parameter is hiding GridSearch's one. Ideally would have moved ModelFeeder outside but it is too coupled with GridSearch class.
```
Pscheidl: It's not hiding anything. Those are equal.

`class ModelFeeder<MP extends Model.Parameters, D extends ModelFeeder>` only defines it in the very same manner as the `GridSearch` class. If you want it to automatically change to whichever boundary is set to `MP` defined in `GridSearch`, you can just delete the `MP` generic parameter in `ModelFeeder` definition:

 `private class ModelFeeder<D extends ModelFeeder> `

Not sure if it helps anything.

deil87: @Pscheidl please see my answer above. Having same name probably a convenient thing here for readability but it makes it impossible to assign those 2 types to each other.

honzasterba: whats the point of the FIXME comment then? do you plan on fixing it, if yes, please add JIRA id to the comment, otherwise it will end up just like the other todos in the codebase

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -18,13 +19,27 @@
...
 
     public abstract void onBuildFailure(final ModelBuildFailure modelBuildFailure, final ParallelModelBuilder parallelModelBuilder);
+
+    public abstract void afterBuildProcessing(final ParallelModelBuilder parallelModelBuilder, final Model model);
   }
 
   private final transient ParallelModelBuilderCallback _callback;
-  private final transient IcedAtomicInt _modelInProgressCounter = new IcedAtomicInt();
+
+  public AtomicInteger getModelInProgressCounter() {
```
honzasterba: this exposes the internal state unecessarily, this method and the one below can return simple int

h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -18,13 +19,27 @@
...
+    return _modelInProgressCounter;
+  }
+
+  private final  AtomicInteger _modelInProgressCounter = new AtomicInteger();
```
honzasterba: whitespace after final
mixing field definitions with getters

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -205,26 +208,40 @@ public void onBuildFailure(final ParallelModelBuilder.ModelBuildFailure modelBui
...
+          reconcileMaxRuntime(grid._key, hyperspaceIterator, nextModelParams);
+
+          parallelModelBuilder.run(Collections.singletonList(ModelBuilder.make(nextModelParams)));
+        } else {
+          parallelModelBuilder.noMoreModels();
+        }
       }
     }
 
+    private boolean hasReachedMaxModels(ParallelModelBuilder parallelModelBuilder) {
```
honzasterba: just an idea: wouldnt this be better as a method on parallelModelBuilder with max_models parameter?

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @honzasterba I'm not going to fix this. The Jira like this would get forgotten I'm pretty sure. I just think that those todos and fixme things have a higher chance to be noticed by developers than forgotten Jiras. This is not a best practice for sure but our reality. I will just remove the comment.

deil87: will just rename ModelFeeder's type parameter

deil87:
@honzasterba as a follow up to https://github.com/h2oai/h2o-3/pull/4114#discussion_r363748699 -> 
 this is a valid usage of Java and I have seen examples in our codebase. I'm ok to change this but I would like to ask team's opinion on this...whether we are ok with this syntax? Couple of times I run into situation when reviewers were suggesting opposite things about code conventions. I would prefer to discuss such cases and stick to our team's conventions. @michalkurka @seb-h2o @Pscheidl ( like if you are ok with the syntax like this
``` 
if (_isMultiRunSearch = keyed != null) { 
```
 and dislike if it is better to avoid it)

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -464,7 +453,7 @@ public RandomDiscreteValueWalker(MP params,
     /** Based on the last model, the given array of ScoringInfo, and our stopping criteria should we stop early? */
     @Override
     public boolean stopEarly(Model model, ScoringInfo[] sk) {
-      return ScoreKeeper.stopEarly(ScoringInfo.scoreKeepers(sk),
+      return model != null && ScoreKeeper.stopEarly(ScoringInfo.scoreKeepers(sk),
```
honzasterba: I would split this into multiple lines since now the args are not alligned with the method call

h2o-core/src/main/java/hex/grid/GridSearch.java
honzasterba: some more reading on this subject: https://wiki.c2.com/?TodoCommentsConsideredHarmful

if even a jira will not make you fix it than just forget about it

michalkurka:
> @honzasterba as a follow up to [#4114 (comment)](https://github.com/h2oai/h2o-3/pull/4114#discussion_r363748699) ->
> this is a valid usage of Java and I have seen examples in our codebase. I'm ok to change this but I would like to ask team's opinion on this...whether we are ok with this syntax? Couple of times I run into situation when reviewers were suggesting opposite things about code conventions. I would prefer to discuss such cases and stick to our team's conventions. @michalkurka @seb-h2o @Pscheidl ( like if you are ok with the syntax like this
> 
> ```
> if (_isMultiRunSearch = keyed != null) { 
> ```
> 
> and dislike if it is better to avoid it)
> 
> I decided to use ^^^ syntax as otherwise i would write something like this:
> 
> ```
> _isMultiRunSearch = keyed != null
> if (_isMultiRunSearch) {
> ```

I don't think you need this flag at all. It doesn't matter if the search was restarted or not, you can always ask the grid to give you number of build models (if it was not restarted it will be 0).

h2o-core/src/main/java/hex/ParallelModelBuilder.java
michalkurka: yes, great point

why are the counters not transient anymore?

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @honzasterba with existing design `ParallelModelBuilder' does not have reference to `hyperspaceIterator`...it is working with `Collection<ModelBuilder>`.

deil87: @honzasterba sorry, with max_model parameter it could work. Will try

h2o-core/src/main/java/hex/ParallelModelBuilder.java
deil87: @honzasterba  Moving `hasReachedMaxModels` to ParallelModelBuilder would help to avoid exposure

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @honzasterba thanks, I know this. I just think that we always should consider situation at hand for a final decision. Anyway, fixed

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
deil87: @honzasterba see your point. Let's see if you will like new version.

h2o-core/src/main/java/hex/ParallelModelBuilder.java
deil87: @michalkurka will add transient keywords. Btw... how do we make sure that our copy of ForkJoinTask is up to date?

h2o-core/src/main/java/hex/ParallelModelBuilder.java
deil87: fixed

deil87:
@michalkurka 
> I don't think you need this flag at all. It doesn't matter if the search was restarted or not, you can always ask the grid to give you number of build models (if it was not restarted it will be 0).

yeah, I remember at some point it was needed but not anymore :) Thanks!

deil87:
@honzasterba @michalkurka I believe I have addressed all the comments. Are there any concerns left?

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -205,23 +207,30 @@ public void onBuildFailure(final ParallelModelBuilder.ModelBuildFailure modelBui
...
-      // Attempt to train next model
-      final MP nextModelParams = getNextModelParams(hyperspaceIterator, previousModel, grid);
-      if (nextModelParams != null
-              && isThereEnoughTime()
-              && !_job.stop_requested()
-              && !_hyperSpaceWalker.stopEarly(previousModel, grid.getScoringInfos())) {
 
-        // FIXME: ModelFeeder's <MP> type parameter is hiding GridSearch's one. Ideally would have moved ModelFeeder outside but it is too coupled with GridSearch class.
-        reconcileMaxRuntime(grid._key, hyperspaceIterator, nextModelParams);
+      synchronized (hyperspaceIterator) {
```
honzasterba: I still dont understand why this whole block needs to be synchronized instead of the old approach with null check

deil87: @honzasterba because old version was not dealing with max_models. 
The problem is that line
```
final MP nextModelParams = getNextModelParams(hyperspaceIterator, previousModel, grid);
```
takes next model parameters before all the checks in if statement. So as we are working in a parallel fashion there is a chance for a following scenario:
1) ThreadA initialises nextModelParameters before `if` checks
2) ThreadB in the meantime also takes model parameters and has them processed ( which increases counter of completed models). Let's assume counter was increased to the `max_models` value.
3) ThreadA continues and checking conditions in `if` statement. `hasReachedMaxModels` now will reject this model parameters and we will end up with `max_models - 1` number of built model parameters.

honzasterba: ok, sounds reasonable, could we then avoid this synchronized block by moving the check `!parallelModelBuilder.hasReachedMaxModels(hyperspaceIterator.max_models())` to `getNextModelParams` https://github.com/h2oai/h2o-3/pull/4191/files/bd981f568c5ca42afc37f088c4f7a008d3f8ca8a#diff-d12a4b90c0799eec2a3e8439901328e5R260

deil87: @honzasterba no, we can't do this as well.
Following scenario will lead to a case when we have built more then `max_models` models ( how much more depends on level of parallelisation).

if we do like I think you are suggesting:
```
while (params == null) {
        if (hyperSpaceWalker.hasNext(model) && !parallelModelBuilder.hasReachedMaxModels(hyperspaceIterator.max_models())) { ...
```
we will have a chance to get following scenario:

1) ThreadA passed `hasNext` check. ThreadA passed `hasReachedMaxModels` check
2) ThreadB passed `hasNext` check. ThreadB passed `hasReachedMaxModels` check
3) ThreadA retrieved model parameters, built a model and only now updated counter for completed models. Let's assume that counter now equals to `max_models`. Note that there are still plenty of grid permutations out there.
4) ThreadB ( as it has already passed all the checks before counter's update) retrieved model parameters, built a model. This model will be `max_model + 1`th model.

So we do need synchronisation between  checks and actual retrieval of the parameters. It is similar to the problem with `size()` method in let's say `ConcurrentHashMap` https://stackoverflow.com/questions/10313556/concurrent-hashmap-checking-size. In some concurrent scenarios it is not crucial to check size.. more important to have high concurrent performance. But once you need to control exact size - you have to synchronise write operations ( in our case it is equivalent to building a model).

honzasterba: I am looking at the old code and exactly the same situation was possible before, so this is something @Pscheidl did not catch originally?

deil87: @honzasterba your question is related to the origin of the PR. Early stopping in terms of number of models was embedded inside iterator's hasNext method. Pavel did not have checks for max models outside of the iterator.

deil87: @honzasterba the difference I think is that before there was only one counter `in progress models`. But in this PR I added completed counter. So there is a chance to have them in an inconsistent state when one was decreased and another was not yet increased.

honzasterba: I am sorry, but I do not understand what are you trying to say.

Nevertherless I have gone through the code again and I think I see your point in there being a need for synchronisation here. Are you sure that the synchronized block is as small as it needs to be? I think that at least `reconcileMaxRuntime`  and `parallelModelBuilder.run` do not need to be inside a synchronized block. 

Also it might be more readable to use a ReentrantLock as in other places in parallelmodelbuilder to make it more obvious for what we are locking?

deil87: @honzasterba I don't think that we can move `reconcileMaxRuntime` and `parallelModelBuilder.run` outside of the synchronisation as they are both dependant on `nextModelParams` variable

> Also it might be more readable to use a ReentrantLock as in other places in parallelmodelbuilder to make it more obvious for what we are locking?

I will try to improve readability

honzasterba: but you can call nextModelParams in a synchronized block and then do the rest outside, no?

deil87: @honzasterba sure, i just don't see how we can synchronise over only `if` conditions and  `getNextModelParams` call. Need to think about it more.

honzasterba: something like
```suggestion
     MPF nextModelParams = null;
      synchronized (hyperspaceIterator) {
         if(checks)
           nextModelParams = getNextModelParams()
       }
       if (nextModelParams != null) train
       else noMoreModels() 
```
could work

deil87: @honzasterba yes, something like this. But it would not work. Because we need this line from `parallelModelBuilder.run` to be synchronised:
```
_modelInProgressCounter.incrementAndGet();
```
Otherwise it is the same as your second suggestion where we end up with `max_models + 1` models. Not sure if it makes sense to increment counter separately from `attemptBuildNextModel` as 1) `reconcileMaxRuntime` should be called before `parallelModelBuilder.run` method and 2) `parallelModelBuilder.run` is a non-blocking training

deil87: I actually was trying, of course, to reduce scope of synchronisation (before holidays) and ended up with what I have in current version

Pscheidl: I did not check max models in my original implementation, that's true. I thought walkers are the ones who control max model amount, which I can see is not true right now.

Walker knows how many models it "gave out", would it be possible to have such logic there ?

honzasterba:
@michalkurka does this need to go to master?

Pscheidl:
So, this solves the max_size issue, right ?

deil87:
@honzasterba good question! there are two tests that would fail on master's codebase. `testMaxModelIsSatisfiedWhenSomeModelsHaveFailed` and its parallel vesrion `test_parallel_grid_search_MaxModelIsSatisfiedWhenSomeModelsHaveFailed`. Original goal was to just redesign/refactor the code but it turns out PR also fixes the cases when there are failed model parameters. So I would assume it should go into release branch (at least partially but do think that merging altogether is a simpler possible option)

deil87:
@michalkurka please let me know whether you are ok to merge it into release branch

michalkurka:
@deil87 @honzasterba 

It seems like the APIs are changed - for this to go to release branch I would recommend to make it backward compatible if possible.

Maybe just keeping "modelFailed" method in the interface (and providing default - empty -  implementation) would be enough. Any other incompatibilities?

deil87:
@michalkurka 
1) fixed "modelFailed". 
2) Regarding other incompatibilities - in `ParallelModelBuilderCallback` abstract class I has fixed typo `onBuildSucces` to `onBuildSuccess`. Probably not a big deal to revert it here as well. I can create tiny PR for master to fix typo. wdyt?

deil87:
@honzasterba I have replaced `synchronised` with `ReentrantLock` as we discussed

deil87:
@michalkurka I've changed a base of this PR to `rel-yu`.  Reverted fix for typo in API. So, probably there no other inconsistencies in API anymore.

deil87:
@michalkurka can we get back to this one? conflict is tiny and should not be an obstacle for a review

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -329,9 +345,13 @@ private void gridSearch(Grid<MP> grid) {
...
       // Get iterator to traverse hyper space
       HyperSpaceWalker.HyperSpaceIterator<MP> it = _hyperSpaceWalker.iterator();
-      // Number of traversed model parameters
-      int counter = grid.getModelCount();
-      while (it.hasNext(model)) {
+
+      // This counter will be used as index to generate keys for corresponding `successful` models
+      int numberOfBuiltModels = grid.getModelCount();
+
+      int maxNumberOfModelsToBuild = numberOfBuiltModels + it.max_models();
```
michalkurka: I do not understand this line. For example for the Cartesian walker, it.max_models is the size of the hyperspace. Adding any positive number to it results in a larger size than the size of the hyperspace. I am not sure what the `maxNumberOfModelsToBuild` really represents then.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -329,9 +345,13 @@ private void gridSearch(Grid<MP> grid) {
...
+
+      while (it.hasNext(model) && (it.max_models() == 0 || grid.getModelCount() < maxNumberOfModelsToBuild)) {
```
michalkurka: (it.max_models() == 0 || grid.getModelCount() < maxNumberOfModelsToBuild)

that seems like an always true because is defined as

    maxNumberOfModelsToBuild = numberOfBuiltModels + it.max_models()

It would make sense to me if `maxNumberOfModelsToBuild = it.max_models() - numberOfBuiltModels`

deil87: @michalkurka We need to take into an account situations when we proceed to build models from a new grid which happens to have the very same key as previous grid(s). Let me give you an example:
1) Cartesian Grid1 with hp space of size 6 was fully built. 
```
int numberOfBuiltModels = grid.getModelCount(); // will be 0 for Grid1

int maxNumberOfModelsToBuild = numberOfBuiltModels + it.max_models(); // will be 0 + 6 for Grid1

while (it.hasNext(model) && (it.max_models() == 0 || grid.getModelCount() < 6)) { .... // where grid.getModelCount() belongs to [0,1,2,3,4,5] depending on iteration
```
Now we want(for whatever reason) to train second Grid2 with the same key and with hp of size 20. 
Let's consider two options:

a) Grid2 is again Cartesian:
```
int numberOfBuiltModels = grid.getModelCount(); // will be 6 as Grid1 was fully built

int maxNumberOfModelsToBuild = numberOfBuiltModels + it.max_models(); // will be 6 + 20 for Grid2 = 26

while (it.hasNext(model) && (it.max_models() == 0 || grid.getModelCount() < 26)) { .... // where grid.getModelCount() belongs to [6,7,....25] depending on iteration
```
b) Grid2 is RandomDiscrete with max_models = 10 which is < hp space size:
```
int numberOfBuiltModels = grid.getModelCount(); // will be 6 as Grid1 was fully built

int maxNumberOfModelsToBuild = numberOfBuiltModels + it.max_models(); // will be 6 + 10 for max_models from Grid2

while (it.hasNext(model) && (it.max_models() == 0 || grid.getModelCount() < 16)) { .... // where grid.getModelCount() belongs to [6,7,....15] depending on iteration number
```

All these scenarios I believe are supported properly. `maxNumberOfModelsToBuild` means max number of models grid with this key can have.

h2o-core/src/main/java/hex/grid/GridSearch.java
deil87: @michalkurka please see my explanation here https://github.com/h2oai/h2o-3/pull/4191#discussion_r370112796

h2o-core/src/main/java/hex/grid/GridSearch.java
michalkurka: Ok, let me ask a question: What will your code do if I run the train method twice?

```
>>> search_criteria = {'strategy': 'RandomDiscrete', 'max_models': 2, 'seed': 1}
>>> gbm_grid1 = H2OGridSearch(model=H2OGradientBoostingEstimator,
...                           grid_id='gbm_grid1',
...                           hyper_params=gbm_params1,search_criteria=search_criteria)
>>> gbm_grid1.train(x=x, y=y,
...                 training_frame=train,
...                 validation_frame=valid,
...                 ntrees=100,
...                 seed=1)
gbm Grid Build progress: |████████████████████████████████████████████████████████████████████| 100%
>>> gbm_grid1.train(x=x, y=y,
...                 training_frame=train,
...                 validation_frame=valid,
...                 ntrees=100,
...                 seed=1)
gbm Grid Build progress: |████████████████████████████████████████████████████████████████████| 100%
>>> gbm_grid1
    col_sample_rate learn_rate max_depth sample_rate          model_ids             logloss
0               0.5        0.1         5         0.8  gbm_grid1_model_1  0.5552990538781676
1               0.5       0.01         3         1.0  gbm_grid1_model_2  0.6291962754031823
```

Is the result going to have 2 models or 4?

Also a reminder: one of the use cases is to stop grid search and then continue training. What is the behavior in that case?

michalkurka: One technical note (small one): for large hyperspaces max_models returns Integer.MAX_VALUE - your math thus has an integer overflow problem.

deil87: @michalkurka could you please provide `gbm_params1` in our code snippet so that I can answer your question correctly?

> One technical note (small one): for large hyperspaces max_models returns Integer.MAX_VALUE - your math thus has an integer overflow problem.

We can try to use Long type or BigDecimal :) , but honestly I believe we are quite far from abilities to compute hyperspaces of these sizes. For example if average time to compute single permutation from the grid takes 1 second - it will take 68 years to build a space of Integer.MAX_VALUE size and OOM will happen much much earlier according to our AutoML benchmark. I think we have many places in our project where we rely on a common sense from the user when he/she defines an input for the system. Anyway if user wants to train more then MAX_VALUE models he will run into an issue even without my math as .max_models() returns int right now.

michalkurka: > We can try to use Long type or BigDecimal :) , but honestly I believe we are quite far from abilities to compute hyperspaces of these sizes.

What do you mean by that? Are you saying that H2O won't handle a hyperspace of the size of Integer.MAX_VALUE with say max_models = 2?

deil87: @michalkurka What I was trying to say is that this is a quite unusual use case. And you were talking about case when max_models is not 2:

>for large hyperspaces max_models returns Integer.MAX_VALUE - your math thus has an integer overflow problem

We just can stretch this logic to any extent. Not even Long or BigDecimal type would be enough to use. What is your suggestion here then? we can take Math.min(x, Int.MAX_VALUE) for `maxNumberOfModelsToBuild`

michalkurka: Not really. Almost all the APIs functions return longs, when they do not return longs just like the max_value method they convert to Integer.MAX_VALUE:

https://github.com/h2oai/h2o-3/blob/master/h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java#L400

This tells us:
- the code assumes everything can be `long` (and it is the responsibility of that code to not to exceed long because that is the API contract)
- where things collapse into `ints` it is handled

Resolving my comment could be eg. changing `int maxNumberOfModelsToBuild` to `long maxNumberOfModelsToBuild`. No harm done and the users who has a theoretically giant search space and wants to build a reasonable number of models won't end up with no models because of integer overflow.

deil87: @michalkurka 
> @michalkurka could you please provide gbm_params1 in your code snippet so that I can answer your question correctly?

I will try to guess then. Probably your hyperspace was of a size 2. It is probably not 100% true as there is a chance that iterator from second grid returns  same permutations as a first grid. 
As far as I know we right now have a check whether models parameters were already built for a given permutation therefore we don't have repetitions in a built models within the grid.

Let's say our hyperspace is bigger than 2 ( for instance 20). And we run two grids with same key with `max_models = 2` for both grids. Then we will have from 2 to 4 built models as a results. But `maxNumberOfModelsToBuild` will be equal to 4 and we have an opportunity to skip 2 models due to this logic https://github.com/h2oai/h2o-3/blob/5478d2c6abc3c2b49c25176b762bcbe583724dee/h2o-core/src/main/java/hex/grid/GridSearch.java#L477

deil87: @michalkurka 
> Not really. Almost all the APIs functions return longs, when they do not return longs just like the max_value method they convert to Integer.MAX_VALUE:

this is great. And I also doing this type of handling when it is really makes things safe. 

> the code assumes everything can be long (and it is the responsibility of that code to not to exceed long because that is the API contract)

who will stop us from providing a Map of hyperparameters with up to Integer.MAX_VALUE of keys and where every key points to an array of Integer.MAX_VALUE elements. This will overflow long in `computeMaxSizeOfHyperSpace`. 

> the users who has a theoretically giant search space and wants to build a reasonable number of models won't end up

For cartesian case we always set max_models to a size of hp space. It means that is user wants to build many of such grids..... he will sooner or later overflow and even long. 

Note: my code would not overflow if user wants to build a reasonable number of models, as it would mean that .max_models() was set to a reasonable number and this:
```
int maxNumberOfModelsToBuild = numberOfBuiltModels + it.max_models();
``` 
would not overflow int even if hyperspace is giant

So my point was that with reasonable users even right now code is not going to overflow. We can do long - no problem. But still if you want you can break it with examples from above.

michalkurka: Just tried your code and it produces 4 models instead of 2 (current release):

```
>>> # GBM hyperparameters
... gbm_params1 = {'learn_rate': [0.01, 0.1],
...                 'max_depth': [3, 5, 9],
...                 'sample_rate': [0.8, 1.0],
...                 'col_sample_rate': [0.2, 0.5, 1.0]}
>>> 
>>> search_criteria = {'strategy': 'RandomDiscrete', 'max_models': 2, 'seed': 1}
>>> 
>>> # Train and validate a cartesian grid of GBMs
... gbm_grid1 = H2OGridSearch(model=H2OGradientBoostingEstimator,
...                           grid_id='gbm_grid1',
...                           hyper_params=gbm_params1, search_criteria=search_criteria)
>>> 
>>> gbm_grid1.train(x=x, y=y,
...                 training_frame=train,
...                 validation_frame=valid,
...                 ntrees=100,
...                 seed=1)
gbm Grid Build progress: |████████████████████████████████████████████████████████████████████| 100%
>>> gbm_grid1
    col_sample_rate learn_rate max_depth sample_rate          model_ids             logloss
0               0.5        0.1         5         0.8  gbm_grid1_model_1  0.5552990538781676
1               0.5       0.01         3         1.0  gbm_grid1_model_2  0.6291962754031823

>>> gbm_grid1.train(x=x, y=y,
...                 training_frame=train,
...                 validation_frame=valid,
...                 ntrees=100,
...                 seed=1)
gbm Grid Build progress: |████████████████████████████████████████████████████████████████████| 100%
>>> gbm_grid1
    col_sample_rate learn_rate max_depth sample_rate          model_ids             logloss
0               1.0        0.1         3         0.8  gbm_grid1_model_5  0.5537787805597868
1               0.5        0.1         5         0.8  gbm_grid1_model_1  0.5552990538781676
2               1.0        0.1         5         0.8  gbm_grid1_model_6  0.5567215084062702
3               0.5       0.01         3         1.0  gbm_grid1_model_2  0.6291962754031823
```

michalkurka: >  We can do long - no problem.
OK

deil87: @michalkurka and this result matches what I said:
>  Then we will have from 2 to 4 built models as a results.

Your second run did not choose the same permutations as your first run. That is why 2+2. But  you could get less if we skip some of them in the second run.

deil87: reduce your space to 3 elements. And you should get 3 models

deil87: @michalkurka my understanding that every time you call gbm_grid1.train() from the client you start new grid search but with the same key ( like in the unit tests here https://github.com/h2oai/h2o-3/blob/5478d2c6abc3c2b49c25176b762bcbe583724dee/h2o-algos/src/test/java/hex/grid/GridTest.java#L543 ). Everu new run creates new iterator we don't have a state. That is why we do skipping based on models stat are stored in DKV.

deil87: But from the client's perspective I agree it looks like max_models should be applied to all .train() calls. But we need to store state then ( max_models from the first run)

michalkurka: > Your second run did not choose the same permutations as your first run. That is why 2+2. But you could get less if we skip some of them in the second run.

The code is fully seeded and always produces 2 models on the current release.

```
import h2o
from h2o.estimators.gbm import H2OGradientBoostingEstimator

h2o.init()

# Import a sample binary outcome dataset into H2O
data = h2o.import_file("https://s3.amazonaws.com/erin-data/higgs/higgs_train_10k.csv")

x = data.columns
y = "response"
x.remove(y)

data[y] = data[y].asfactor()
test[y] = test[y].asfactor()

# Split data into train & validation
ss = data.split_frame(seed = 1)
train = ss[0]
valid = ss[1]

# GBM hyperparameters
gbm_params1 = {'learn_rate': [0.01, 0.1],
                'max_depth': [3, 5, 9],
                'sample_rate': [0.8, 1.0],
                'col_sample_rate': [0.2, 0.5, 1.0]}

search_criteria = {'strategy': 'RandomDiscrete', 'max_models': 2, 'seed': 1}

# Train and validate a cartesian grid of GBMs
gbm_grid1 = H2OGridSearch(model=H2OGradientBoostingEstimator,
                          grid_id='gbm_grid1',
                          hyper_params=gbm_params1, search_criteria=search_criteria)
gbm_grid1.train(x=x, y=y,
                training_frame=train,
                validation_frame=valid,
                ntrees=100,
                seed=1)
gbm_grid1.train(x=x, y=y,
                training_frame=train,
                validation_frame=valid,
                ntrees=100,
                seed=1)
gbm_grid1
```

Your code produces 4 models.

deil87: @michalkurka let me investigate then. Probably I was relying on java tests and behaviour is different there.

deil87: @michalkurka 
My solution was based on an assumption that we should append grids. Confusion started in this test https://github.com/h2oai/h2o-3/blob/9ab1cec03da7f674ace8fab9f56e077de0d8c98f/h2o-algos/src/test/java/hex/glrm/GLRMGridTest.java#L96 and particularly this assert here https://github.com/h2oai/h2o-3/blob/9ab1cec03da7f674ace8fab9f56e077de0d8c98f/h2o-algos/src/test/java/hex/glrm/GLRMGridTest.java#L154

This test actually means that we can run two grids with the same key and final number of models can be more then `max_models`. 
I even added parallel version of `testGridAppend` test here: https://github.com/h2oai/h2o-3/blob/5478d2c6abc3c2b49c25176b762bcbe583724dee/h2o-algos/src/test/java/hex/grid/GridTest.java#L441

We can reproduce this case in a client:
```
# GBM hyperparameters
      gbm_params1 = {'learn_rate': [0.01],
                     'max_depth': [3, 5],
                     'sample_rate': [0.8],
                     'col_sample_rate': [0.2]}
      gbm_params2 = {'learn_rate': [0.01],
                     'max_depth': [3, 5],
                     'sample_rate': [0.8],
                     'col_sample_rate': [0.1]} # difference is here

      search_criteria = {'strategy': 'Cartesian'}

      # Train and validate a cartesian grid of GBMs
      gbm_grid1 = H2OGridSearch(model=H2OGradientBoostingEstimator,
                                grid_id='gbm_grid1',
                                hyper_params=gbm_params1, search_criteria=search_criteria)
      gbm_grid1.train(x=x, y=y,
                      training_frame=train,
                      validation_frame=valid,
                      ntrees=100,
                      seed=1)

      gbm_grid1 = H2OGridSearch(model=H2OGradientBoostingEstimator,
                                grid_id='gbm_grid1',
                                hyper_params=gbm_params2, search_criteria=search_criteria)
      gbm_grid1.train(x=x, y=y,
                      training_frame=train,
                      validation_frame=valid,
                      ntrees=100,
                      seed=1)
      print(gbm_grid1)
```
will return 4 models even though `max_models = 2`.

So I think it is a bad design of iterators. Even though both iterators implements following method
```
/**
     * @return the maximum number of models to build
     */
    int max_models();
```
Cartesian implementation does not use it to stop anything. And also `HyperSpaceSearchCriteria` does not have `max_models` and `set_max_models`, as a consequence CartesianSearchCriteria does not have them as well....but Cartesian iterator does know about `max_models` for some reason.

But anyway, looks like it is how it is. I'm going to follow assumption that for RandomDiscrete case we should not get more models than `max_models` that was used in a constructor of the H2OGridSearch on a client(same for other clients). If `max_models` is being changed for different runs of the same grid key.... then we should not add them up but should use Math.max(max_models1, max_models_new,...)

deil87: @michalkurka we should use .max_model() = 0 for Cartesian instead of setting it to the size of hp space I think. Which will be handled here
```
(it.max_models() == 0 || grid.getModelCount() < maxNumberOfModelsToBuild)
```

deil87: @michalkurka We need to decide what will be an expected behaviour in case we run grid multiple times as there is some inconsistency in behaviours right now:
1) In your py snippet expectation is that we respect max_models regardless of number of times user runs grid.
2) In java unit tests, we can find test that assumes that we will append grids. [Here](https://github.com/h2oai/h2o-3/blob/46681848d670c197bc32dcd4cd4618b3b7cc8f9d/h2o-algos/src/test/java/hex/grid/GridTest.java#L412) for example we expect 2 models even though size of hp space is 1 ( max_model therefore also 1). 

I'm going to fix `testFailedParamsRetention` test with either Random search and corresponding max_models for two grids(for Cartesian search we should not expect second model to be added). Or, which is better I think, to make Cartesian's max_models return 0 ( which just disables early stopping)

PR #4202, opened 2020-01-10
PUBDEV-7069 - Protection against non-valid(BAD) types in DKVClearTask

Pscheidl:
DO NOT MERGE - Custom code for our customer.

https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7069?filter=myopenissues&orderby=priority%20DESC

### DIFF
--- a/h2o-core/src/main/java/water/DKVManager.java
+++ b/h2o-core/src/main/java/water/DKVManager.java
@@ -12,14 +12,15 @@ public class DKVManager {
   /**
    * Clears keys in all H2O nodes, except for the ones marked as retained.
    * Only Model and Frame keys are retained. If a key of any other type is provided, it will be removed as well.
-   * 
+   * <p>
    * Model's training and validation frames are retained automatically with the specified model. However, cross validation models are NOT retained.
    *
-   * @param retainedKeys Keys of {@link Frame}s and {@link Model}s to be retained.
+   * @param retainedKeys Keys of {@link Frame}s and {@link Model}s to be retained. Only Frame and Model keys are accepted.
    */
   public static void retain(final Key[] retainedKeys){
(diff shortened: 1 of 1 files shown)


h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -111,10 +112,17 @@ protected void setupLocal() {
         }
         if(value.isNull()) continue;
 
-        if (value.isFrame()) {
-          ((Frame) value.get()).retain(removalFutures, retainedKeys);
-        } else if (value.isModel()) {
-          ((Model)value.get()).remove(removalFutures);
+        try {
+          TypeMap.getTheFreezableOrThrow(value.type());
```
Pscheidl: It declares ClassNotFoundException and makes the catch block possible to compile. The exception itself may occur in the `value.isFrame()` or `value.isModel` calls below as well.

Pscheidl:
I'll try to merge this code once a reproducible test case is found.

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -111,10 +112,17 @@ protected void setupLocal() {
...
+          if (value.isFrame()) {
+            ((Frame) value.get()).retain(removalFutures, retainedKeys);
+          } else if (value.isModel()) {
+            ((Model) value.get()).remove(removalFutures);
+          }
+
+        } catch (ClassNotFoundException e) {
```
michalkurka: Please add a comment on why is this exception caught.

Pscheidl: Done-.

Pscheidl:
Also added a little bit more JavaDoc.

PR #4203, opened 2020-01-10
We don't need Hmisc library, remove it from test

michalkurka:


Pscheidl:
Should we remove it from docker image as well ?

michalkurka:
@Pscheidl absolutely, that is the next step

Pscheidl:
Cool. I'd wait for the test though.

PR #4208, opened 2020-01-13
PUBDEV-7135 download functions

koniecsveta:
functions for downloading binary model from R / Python client added:
https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7135


h2o-core/src/main/java/water/AutoBuffer.java
```diff
@@ -222,6 +222,19 @@ public AutoBuffer( ) {
     _persist = 0;               // No persistance
   }
 
+  /** Write to an ever-expanding byte[].  Instead of calling {@link #close()},
+   *  call {@link #buf()} to retrieve the final byte[].  */
+  public AutoBuffer(boolean persist) {
```
honzasterba: what is the reason for not using AutoBuffer( OutputStream os, boolean persist )?

h2o-py/tests/testdir_misc/pyunit_download_model.py
```diff
@@ -0,0 +1,34 @@
...
+    prostate_glm.train(x=["AGE","RACE","PSA","DCAPS"], y="CAPSULE", training_frame=prostate)
+    path = pyunit_utils.locate("results")
+
+    assert os.path.isdir(path), "Expected save directory {0} to exist, but it does not.".format(path)
+    model_path = h2o.download_model(prostate_glm, path=path)
+
+    assert os.path.isfile(model_path), "Expected load file {0} to exist, but it does not.".format(model_path)
+    the_model = h2o.load_model(model_path)
+
+    assert isinstance(the_model, H2OEstimator), "Expected and H2OBinomialModel, but got {0}".format(the_model)
```
honzasterba: the error message does not match the assert, H2OGeneralizedLinearEstimator should be expected

h2o-py/h2o/h2o.py
```diff
@@ -1405,6 +1405,31 @@ def save_model(model, path="", force=False):
     return api("GET /99/Models.bin/%s" % model.model_id, data={"dir": path, "force": force})["dir"]
 
 
+def download_model(model, path=""):
```
honzasterba: I think for save we have a convenience method model.save() we should have one for download as well I think

h2o-core/src/main/java/water/api/ModelsHandler.java
```diff
@@ -154,6 +154,15 @@ public StreamingSchema fetchMojo(int version, ModelsV3 s) {
     String filename = JCodeGen.toJavaId(s.model_id.key().toString()) + ".zip";
     return new StreamingSchema(model.getMojo(), filename);
   }
+  
+  @SuppressWarnings("unused") // called from the RequestServer through reflection
+  public StreamingSchema fetchBinaryModel(int version, ModelsV3 s) {
+    Model model = getFromDKV("key", s.model_id.key());
+    OutputStream os = null;
```
honzasterba: unused variable

h2o-core/src/main/java/water/AutoBuffer.java
koniecsveta: when I tried to use it, the buffer was overflowing and as a result I got only the end of binary file (only the content which was written to it after overflow)

michalkurka: @koniecsveta I think we do need to make the approach outlined by @honzasterba work if there is an issue - I can help with fixing that

koniecsveta:
thanks for review and help! pushed the remaining fixes in a4ffb6e

michalkurka:
@koniecsveta since we are adding download_model, should we also have upload_model?

h2o-py/h2o/model/model_base.py
```diff
@@ -859,6 +859,19 @@ def aucpr(self, train=False, valid=False, xval=False):
     def pr_auc(self, train=False, valid=False, xval=False):
         pass
 
+    def download_model(self, path=""):
+        """
+        Download an H2O Model object to disk.
+    
+        :param model: The model object to download.
+        :param path: a path to download the model at (hdfs, s3, local)
```
michalkurka: it seems to me it can only be `local`

koniecsveta:
> @koniecsveta since we are adding download_model, should we also have upload_model?
> 
> should we add another jira?

we already have h2o.load_model in python API and h2o.loadModel in R API which needs path to binary file. But we don't have anything like h2o.upload(/*bytes representing the model*/) for uploading model that is not stored as a file. Do we want to implement it?

honzasterba:
I think upload_model is a good idea, but since this looks pretty much done I would put it into separate PR

PR #4212, opened 2020-01-13
PUBDEV-6999: Fixed interaction bug for GLM Binomial

wendycwong:
This PR fixes the bug reported in https://0xdata.atlassian.net/browse/PUBDEV-6999?filter=-1.

I have done the following:
1. fix the bug with enum enum column interactions in method DataInfo.filterExpandedColumns where it is assumed that the interaction column is always the first numerical column.
2. fix the bug with enum and num column interactions in method DataInfo.getInteractionOffset.  I added the NA check before accessing the interaction column.  If there is NA, it should be replaced with the mode of the enum column.
3. Fixed bug in gradient calculation with enum and num interaction columns.
4.  Added detail pyunit test to check all possible column type interactions with validation_frame in the model building

### DIFF
--- a/h2o-r/tests/testdir_algos/glm/runit_GLM_interactions3.R
+++ b/h2o-r/tests/testdir_algos/glm/runit_GLM_interactions3.R
@@ -40,7 +40,7 @@ test.glm.interactions3 <- function() {
     print( paste0("H2O Coeff2: ",  h2o_coef2))
     print( paste0("R   Coeff: ",  R_coef))
     expect_true( abs( h2o_coef1-R_coef) < 1e-11 )
-    expect_true( abs( h2o_coef2-h2o_coef2 ) < 1e-15  )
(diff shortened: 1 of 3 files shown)


h2o-r/tests/testdir_algos/glm/runit_GLM_interactions3.R
```diff
@@ -40,7 +40,7 @@ test.glm.interactions3 <- function() {
     print( paste0("H2O Coeff2: ",  h2o_coef2))
     print( paste0("R   Coeff: ",  R_coef))
     expect_true( abs( h2o_coef1-R_coef) < 1e-11 )
-    expect_true( abs( h2o_coef2-h2o_coef2 ) < 1e-15  )
+    expect_true( abs( h2o_coef2-h2o_coef1 ) < 1e-15  )
```
michalkurka, line 43: good catch!

PR #4215, opened 2020-01-14
PUBDEV-6447 Constrained K-means improvements

maurever:
**Constrained K-means Improvements**
- add Python and R API - :heavy_check_mark: 
- create a demo and compare the running time with other implementation :no_entry_sign: (reason: no suitable open source implementation for big data)
- create a demo where Aggregator is used to reducing big dataset size, compare the result with pure constrained K-means :heavy_check_mark: 
- add tests :heavy_check_mark: 
- add doc :heavy_check_mark:


angela0xdata:
@maurever would it be possible to include R and Python tests for this? Or is that in a different PR?

maurever:
> @maurever would it be possible to include R and Python tests for this? Or is that in a different PR?

Yes, I will add tests and documentation within this PR. However, first is important to figure out if there is some way to calculate Constrained K-Means for big data.

maurever:
Thank you, @angela0xdata, for updates and your review!

h2o-py/tests/testdir_algos/kmeans/pyunit_constrained_kmeans.py
```diff
@@ -0,0 +1,54 @@
...
+        [1, 40, 100],
+        [1, 1, 1],
+        [1, 1, 148],
+        [147, 1, 1],
+        [1, 148, 1],
+        [1, 1, 1],
+        [50, 50, 50]
+    ]
+
+    for i in list(range(len(constraints))):
```
honzasterba: the list() call is unnecessary

h2o-algos/src/main/java/hex/kmeans/KMeansSimplexSolver.java
```diff
@@ -374,56 +345,82 @@ public Frame assignClusters() {
...
         return false;
     }
 
+    /**
+     * Check if the constraints are satisfied. 
+     * If yes, the algorithm can continue as standard K-means and save time. Useful when constraints are small numbers
+     * @return 
+     */
+    public boolean areConstraintsSatisfied() {
+        for(long i = 2; i < _secondLayerSize + 2; i++) {
```
michalkurka: What is the size of the layer? Sequential reads benefit the most from using Vec.Reader.

h2o-algos/src/main/java/hex/kmeans/KMeansSimplexSolver.java
```diff
@@ -753,3 +754,91 @@ public void reduce(FindMinimalWeightTask mrt) {
...
+        cs[_newAssignmentIndex].set(row, clusterIndex);
+        // distances
+        cs[_distanceAssignmentIndex].set(row, cs[_dataStopIndex + (_hasWeightsColumn ? 1 : 0)  + clusterIndex].at8(row));
+        numberOfPointsInClusterLocal[clusterIndex]++;
+    }
+
+    @Override
+    public void map(Chunk[] cs) {
+        int[] numberOfPointsInClusterLocal = new int[_constraintsLength];
+        for(int i = 0; i < _constraintsLength; i++){
```
michalkurka: not needed

h2o-algos/src/main/java/hex/kmeans/KMeansSimplexSolver.java
```diff
@@ -753,3 +754,91 @@ public void reduce(FindMinimalWeightTask mrt) {
...
+                }
+            }
+        }
+        this._numberOfPointsInCluster = numberOfPointsInClusterLocal;
+    }
+
+    @Override
+    public void reduce(AssignClusterTask mrt) {
+        for (int i = 0; i < _constraintsLength; i++){
+            this._numberOfPointsInCluster[i] += mrt._numberOfPointsInCluster[i];
```
michalkurka: ArrayUtils has a helper function for this

h2o-algos/src/main/java/hex/kmeans/KMeansSimplexSolver.java
```diff
@@ -753,3 +754,91 @@ public void reduce(FindMinimalWeightTask mrt) {
...
+                boolean assigned = false;
+                for (int j = 0; j < _constraintsLength; j++) {
+                    if (cs[_flowIndexStart + j].at8(i) == 1) {
+                        assignCluster(cs, i, j, numberOfPointsInClusterLocal);
+                        assigned = true;
+                        break;
+                    }
+                }
+                if(!assigned){
+                    double minDistance = Double.MAX_VALUE;
```
michalkurka: Can a floating-point math cause eg. infinities to be present?

h2o-algos/src/test/java/hex/kmeans/KmeansConstrainedTest.java
```diff
@@ -259,4 +260,34 @@ public void testWeatherChicagoConstrained() {
         }
     }
 
+    @Test @Ignore
```
michalkurka: why ignored?

h2o-py/tests/testdir_algos/kmeans/pyunit_constrained_kmeans.py
maurever: Of course, thank you.

h2o-algos/src/main/java/hex/kmeans/KMeansSimplexSolver.java
maurever: Good point. The size of layer equals `k`. It is not usually a huge number. However, the reader can help. Fixed.

h2o-algos/src/main/java/hex/kmeans/KMeansSimplexSolver.java
maurever: You are right.  :+1:

h2o-algos/src/main/java/hex/kmeans/KMeansSimplexSolver.java
maurever: Fixed. Thank you.

h2o-algos/src/main/java/hex/kmeans/KMeansSimplexSolver.java
maurever: I think this can't happen. But I changed to code to be 100% safe.

h2o-algos/src/test/java/hex/kmeans/KmeansConstrainedTest.java
maurever: It uses big data, and it is really slow. I used it for internal testing. Should I remove it?

honzasterba: since it can be useful for development, I would just add a comment

michalkurka: it is fine to keep, however, usually these things go bad over time - many of the Ignored tests don't work anymore

maybe (and really just an idea) - could we turn it into a benchmark - since you are still improving performance?

PR #4216, opened 2020-01-14
PUBDEV-7204: GridSearch over TE parameters is not working

deil87:
This PR fixes two issues:
1) Implementation of TargetEncoderBuilder does not take into account the way how GridSearch is working. Instead of using TargetEncoderBuilder's constructor which sets passed model parameters, GridSearch is using startupOnce constructor with default parameters so that it can later reassign them with values from the grid. Even though initialisation is being called, training frame is not available  in default model parameters

2) Wrong key was passed to TargetEncoderModel
```new TargetEncoderModel(_result, ...   ```. Instead of `_result` from ModelBuilder the one from Job was used. See for more details my comment https://github.com/h2oai/h2o-3/pull/4216#discussion_r368741184

### DIFF
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
@@ -2,6 +2,7 @@ package ai.h2o.targetencoding;
 
 import hex.ModelBuilder;
 import hex.ModelCategory;
+import water.DKV;
 import water.Scope;
 import water.fvec.Frame;
 import water.fvec.Vec;
@@ -36,7 +37,11 @@ public class TargetEncoderBuilder extends ModelBuilder<TargetEncoderModel, Targe
   private class TargetEncoderDriver extends Driver {
     @Override
     public void computeImpl() {
+      // In case ModelBuilder is registered with default model parameters ( i.e. without training frame etc.) we want to init it at this stage as well
+      if(_startUpOnceModelBuilder) init(false);
+
       final int numColsRemoved = hasFoldCol() ? 2 : 1; // Response is always at the last index, fold column is on the index before.
+
       final String[] encodedColumns = Arrays.copyOf(train().names(), train().names().length - numColsRemoved);
       TargetEncoder tec = new TargetEncoder(encodedColumns);
 
@@ -54,10 +59,8 @@ public class TargetEncoderBuilder extends ModelBuilder<TargetEncoderModel, Targe
 
       disableIgnoreConstColsFeature();
       TargetEncoderModel.TargetEncoderOutput output = new TargetEncoderModel.TargetEncoderOutput(TargetEncoderBuilder.this, _targetEncodingMap, priorMean);
-      _targetEncoderModel = new TargetEncoderModel(_job._result, _parms, output, tec);
-
-      _targetEncoderModel.write_lock(_job);
-      _targetEncoderModel.unlock(_job);
+      _targetEncoderModel = new TargetEncoderModel(_result, _parms, output, tec);
+      DKV.put(_targetEncoderModel);
     }
 
     private void disableIgnoreConstColsFeature() {
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
@@ -29,38 +29,43 @@ public class TargetEncoderBuilder extends ModelBuilder<TargetEncoderModel, Targe
     super(new TargetEncoderModel.TargetEncoderParameters(), startupOnce);
   }
 
-  @Override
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: h2o-core/src/main/java/water/Lockable.java [hunk 1]
```diff
@@ -113,7 +113,7 @@ public final Futures delete(Key<Job> job_key, Futures fs, boolean cascade) {
     @Override public Lockable atomic(Lockable old) {
       _old = old;
       if( old != null ) {       // Prior Lockable exists?
-        assert !old.is_wlocked(_job_key) : "Key "+_key+" already locked (or deleted); lks="+Arrays.toString(old._lockers); // No double locking by same job
+        if(old.is_wlocked(_job_key)) return old;
```
deil87: @michalkurka can we not fail if Lockable is already locked with the same key? This way we can probably include `write_lock` inside `update`. I don't know if there are situations when we want to update without prior locking. Maybe local updates do not require distributed locks( then it is better not to include write_lock inside update to give flexibility) .
Of course without failing developers are free to abuse this ability and use write_lock unnecessary number of times. But still it is easier than trying to understand why some execution paths have different lock story. I run into such situation : for GBM model parameters process of building a model does not release lock but for TargetEncoder model parameters it does. But it is hard to understand who and where releases the lock in TE path.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
```diff
@@ -55,9 +60,8 @@ public void computeImpl() {
       disableIgnoreConstColsFeature();
       TargetEncoderModel.TargetEncoderOutput output = new TargetEncoderModel.TargetEncoderOutput(TargetEncoderBuilder.this, _targetEncodingMap, priorMean);
       _targetEncoderModel = new TargetEncoderModel(_job._result, _parms, output, tec);
-
-      _targetEncoderModel.write_lock(_job);
-      _targetEncoderModel.unlock(_job);
```
deil87: @michalkurka  What is the purpose of these two lines? Does it make sure that `_targetEncoderModel` is updated globally?

h2o-extensions/target-encoder/src/test/java/ai/h2o/targetencoding/strategy/TargetEncoderRGSTest.java
```diff
@@ -52,4 +56,82 @@ public void getTargetEncodingMapByTrainingTEBuilder() {
...
+      Scope.track_generic(grid);
+
+      assertEquals(54, grid.getModelCount());
+
+    } finally {
+      Scope.exit();
+    }
+  }
+
+  @Test // Probably could be resolved with redesign/fix PUBDEV-7110
```
deil87: This parallel version of test is not fixed yet. Will try to merge in https://github.com/h2oai/h2o-3/pull/4191/files to see if it fixes issue.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
```diff
@@ -54,10 +59,8 @@ public void computeImpl() {
 
       disableIgnoreConstColsFeature();
       TargetEncoderModel.TargetEncoderOutput output = new TargetEncoderModel.TargetEncoderOutput(TargetEncoderBuilder.this, _targetEncodingMap, priorMean);
-      _targetEncoderModel = new TargetEncoderModel(_job._result, _parms, output, tec);
-
-      _targetEncoderModel.write_lock(_job);
-      _targetEncoderModel.unlock(_job);
+      _targetEncoderModel = new TargetEncoderModel(_result, _parms, output, tec);
```
deil87: ``` job._result``` was a wrong key to pass to a Model. It was causing key to be removed by Scope. Needed key that is stored in `ModelBuilder`'s `_result` field first, matches keys that are used to retrieve models from the grid ( this way we don't get N null models in the end) and second, this key is being created outside of (before) Scope.enter() and is not being unnecessary deleted from DKV.

h2o-core/src/main/java/water/Lockable.java
deil87: this is outdated. But still interesting to know.

michalkurka:
I will review this first thing tomorrow

michalkurka:
@deil87 tests are great, thank you

Code changes in TargetEncoderModel do not respect how initialization should be done and do not lock model while it is building, my attempt to address these issues: https://github.com/h2oai/h2o-3/pull/4245

deil87:
@michalkurka I also noticed that `parallelGridSearch()` method does not wrap computations in a try-catch block as `gridSearch()` does. Maybe we can move this logic up an something like this:
```
try {
          if (_parallelism == 1) {
            gridSearch(grid);
          } else if (_parallelism > 1) {
            parallelGridSearch(grid);
          } else {
            throw new IllegalArgumentException(String.format("Grid search parallelism level must be >= 1. Give value is '%d'.",
                    _parallelism));
          }
        } finally {
          grid.unlock(_job);
        }
```
?

PR #4217, opened 2020-01-14
[PUBDEV-7203] Expose Individual Chunks via REST API

mn-mikke:
Lots of the code is extracted from [`ExternalFrameReaderBackend`](https://github.com/h2oai/h2o-3/blob/8b9882de0701538be731a0b4110dbec8fb67bf09/h2o-core/src/main/java/water/ExternalFrameReaderBackend.java) and [`ExternalFrameReaderClient`](https://github.com/h2oai/h2o-3/blob/914448614409b50b096893affa1b4a45905e0c12/h2o-core/src/main/java/water/ExternalFrameReaderClient.java) to minimalize changes. Once the whole initiative of REST transportation logic is done, all `ExternalFrame*`classes can be removed.

### DIFF
--- a/h2o-core/src/main/java/water/ChunkAutoBufferReader.java
+++ b/h2o-core/src/main/java/water/ChunkAutoBufferReader.java
@@ -20,61 +20,61 @@ public final class ChunkAutoBufferReader implements Closeable  {
     }
 
     public boolean readBoolean() {
-        boolean data = buffer.getZ();
+        final boolean data = buffer.getZ();
         isLastNA = ExternalFrameUtils.isNA(buffer, data);
         return data;
     }
 
     public byte readByte() {
-        byte data = buffer.get1();
+        final byte data = buffer.get1();
         isLastNA = ExternalFrameUtils.isNA(buffer, data);
         return data;
     }
 
     public char readChar() {
-        char data = buffer.get2();
+        final char data = buffer.get2();
         isLastNA = ExternalFrameUtils.isNA(buffer, data);
         return data;
     }
 
     public short readShort() {
-        short data = buffer.get2s();
+        final short data = buffer.get2s();
         isLastNA = ExternalFrameUtils.isNA(buffer, data);
         return data;
     }
 
     public int readInt() {
-        int data = buffer.getInt();
+        final int data = buffer.getInt();
         isLastNA = ExternalFrameUtils.isNA(buffer, data);
         return data;
     }
 
     public long readLong() {
-        long data = buffer.get8();
+        final long data = buffer.get8();
         isLastNA = ExternalFrameUtils.isNA(buffer, data);
         return data;
     }
 
     public float readFloat() {
-        float data = buffer.get4f();
+        final float data = buffer.get4f();
         isLastNA = ExternalFrameUtils.isNA(data);
         return data;
     }
 
     public double readDouble() {
-        double data = buffer.get8d();
+        final double data = buffer.get8d();
         isLastNA = ExternalFrameUtils.isNA(data);
         return data;
     }
 
     public String readString() {
-        String data = buffer.getStr();
+        final String data = buffer.getStr();
         isLastNA = ExternalFrameUtils.isNA(buffer, data);
         return data;
     }
 
     public Timestamp readTimestamp() {
-        Timestamp data = new Timestamp(buffer.get8());
+        final Timestamp data = new Timestamp(buffer.get8());
         isLastNA = ExternalFrameUtils.isNA(buffer, data);
         return data;
     }
--- a/h2o-core/src/main/java/water/ChunkAutoBufferReader.java
+++ b/h2o-core/src/main/java/water/ChunkAutoBufferReader.java
@@ -1,7 +1,6 @@
 package water;
 
 import java.io.Closeable;
-import java.io.IOException;
 import java.io.InputStream;
 import java.sql.Timestamp;
 
@@ -88,7 +87,7 @@ public final class ChunkAutoBufferReader implements Closeable  {
     }
     
     @Override
-    public void close() throws IOException {
+    public void close() {
         buffer.close();
     }
 }
--- a/h2o-core/src/main/java/water/ChunkAutoBufferReader.java
+++ b/h2o-core/src/main/java/water/ChunkAutoBufferReader.java
@@ -1,0 +1,94 @@
+package water;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.InputStream;
+import java.sql.Timestamp;
+
+public final class ChunkAutoBufferReader implements Closeable  {
+    
+    private final AutoBuffer buffer;
+    private final int numRows;
+    private boolean isLastNA = false;
+    
+    public ChunkAutoBufferReader(InputStream inputStream) {
+        this.buffer = new AutoBuffer(inputStream);
+        this.numRows = readInt();
+    }
+
+    public int getNumRows() {
+        return numRows;
+    }
+
+    public boolean readBoolean() {
+        boolean data = buffer.getZ();
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public byte readByte() {
+        byte data = buffer.get1();
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public char readChar() {
+        char data = buffer.get2();
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public short readShort() {
+        short data = buffer.get2s();
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public int readInt() {
+        int data = buffer.getInt();
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public long readLong() {
+        long data = buffer.get8();
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public float readFloat() {
+        float data = buffer.get4f();
+        isLastNA = ExternalFrameUtils.isNA(data);
+        return data;
+    }
+
+    public double readDouble() {
+        double data = buffer.get8d();
+        isLastNA = ExternalFrameUtils.isNA(data);
+        return data;
+    }
+
+    public String readString() {
+        String data = buffer.getStr();
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public Timestamp readTimestamp() {
+        Timestamp data = new Timestamp(buffer.get8());
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    /**
+     * This method is used to check if the last received value was marked as NA by H2O backend
+     */
+    public boolean isLastNA() {
+        return isLastNA;
+    }
+    
+    @Override
+    public void close() throws IOException {
+        buffer.close();
+    }
+}
--- a/h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
+++ b/h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
@@ -6,7 +6,6 @@ import water.fvec.Frame;
 import water.parser.BufferedString;
 
 import java.io.Closeable;
-import java.io.IOException;
 import java.io.OutputStream;
 import java.util.UUID;
 
@@ -22,13 +21,13 @@ public final class ChunkAutoBufferWriter implements Closeable  {
     }
 
     public void writeChunk(String frameName, int chunkId, byte[] expectedTypes, int[] selectedColumnIndices) {
-        Frame frame = DKV.getGet(frameName);
-        Chunk[] chunks = ChunkUtils.getChunks(frame, chunkId);
-        int numberOfRows = chunks[0]._len;
+        final Frame frame = DKV.getGet(frameName);
+        final Chunk[] chunks = ChunkUtils.getChunks(frame, chunkId);
+        final int numberOfRows = chunks[0]._len;
         ExternalFrameUtils.sendInt(buffer, numberOfRows);
 
         // buffered string to be reused for strings to avoid multiple allocation in the loop
-        BufferedString valStr = new BufferedString();
+        final BufferedString valStr = new BufferedString();
         for (int rowIdx = 0; rowIdx < numberOfRows; rowIdx++) {
             for(int i = 0; i < selectedColumnIndices.length; i++){
                 if (chunks[selectedColumnIndices[i]].isNA(rowIdx)) {
--- a/h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
+++ b/h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
@@ -1,0 +1,89 @@
+package water;
+
+import water.fvec.Chunk;
+import water.fvec.ChunkUtils;
+import water.fvec.Frame;
+import water.parser.BufferedString;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.UUID;
+
+import static water.ExternalFrameUtils.*;
+import static water.ExternalFrameUtils.EXPECTED_STRING;
+
+public final class ChunkAutoBufferWriter implements Closeable  {
+    
+    private final AutoBuffer buffer;
+    
+    public ChunkAutoBufferWriter(OutputStream outputStream) {
+        this.buffer = new AutoBuffer(outputStream, false);
+    }
+
+    public void writeChunk(String frameName, int chunkId, byte[] expectedTypes, int[] selectedColumnIndices) {
+        Frame frame = DKV.getGet(frameName);
+        Chunk[] chunks = ChunkUtils.getChunks(frame, chunkId);
+        int numberOfRows = chunks[0]._len;
+        ExternalFrameUtils.sendInt(buffer, numberOfRows);
+
+        // buffered string to be reused for strings to avoid multiple allocation in the loop
+        BufferedString valStr = new BufferedString();
+        for (int rowIdx = 0; rowIdx < numberOfRows; rowIdx++) {
+            for(int i = 0; i < selectedColumnIndices.length; i++){
+                if (chunks[selectedColumnIndices[i]].isNA(rowIdx)) {
+                    ExternalFrameUtils.sendNA(buffer, expectedTypes[i]);
+                } else {
+                    final Chunk chnk = chunks[selectedColumnIndices[i]];
+                    switch (expectedTypes[i]) {
+                        case EXPECTED_BOOL:
+                            ExternalFrameUtils.sendBoolean(buffer, (byte)chnk.at8(rowIdx));
+                            break;
+                        case EXPECTED_BYTE:
+                            ExternalFrameUtils.sendByte(buffer, (byte)chnk.at8(rowIdx));
+                            break;
+                        case EXPECTED_CHAR:
+                            ExternalFrameUtils.sendChar(buffer, (char)chnk.at8(rowIdx));
+                            break;
+                        case EXPECTED_SHORT:
+                            ExternalFrameUtils.sendShort(buffer, (short)chnk.at8(rowIdx));
+                            break;
+                        case EXPECTED_INT:
+                            ExternalFrameUtils.sendInt(buffer, (int)chnk.at8(rowIdx));
+                            break;
+                        case EXPECTED_FLOAT:
+                            ExternalFrameUtils.sendFloat(buffer, (float)chnk.atd(rowIdx));
+                            break;
+                        case EXPECTED_LONG:
+                            ExternalFrameUtils.sendLong(buffer, chnk.at8(rowIdx));
+                            break;
+                        case EXPECTED_DOUBLE:
+                            ExternalFrameUtils.sendDouble(buffer, chnk.atd(rowIdx));
+                            break;
+                        case EXPECTED_TIMESTAMP:
+                            ExternalFrameUtils.sendTimestamp(buffer, chnk.at8(rowIdx));
+                            break;
+                        case EXPECTED_STRING:
+                            String string = null;
+                            if (chnk.vec().isCategorical()) {
+                                string = chnk.vec().domain()[(int) chnk.at8(rowIdx)];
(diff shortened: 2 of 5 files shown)


mn-mikke:
Should i also add something like [DownloadDataHandler](https://github.com/h2oai/h2o-3/blob/43f8ab952a69a8bc9484bd0ffac909b6e3e820ca/h2o-core/src/main/java/water/api/DownloadDataHandler.java) and related things?

michalkurka:
@mn-mikke why do you unpack the Chunk and then sent row-by-row? This means you are transferring much the data uncompressed, is it necessary to transfer row-by-row?

Wouldn't it be better to send compressed chunks and then reconstruct the rows (for iteration) on the client side (in SW)?

mn-mikke:
@michalkurka Row-by-row data transfer would enable Sparkling Water to load only rows that are needed for processing. Chunk-by-chunk data transfer would enforce SW to load the whole partition into memory first and start with processing. To be honest with you, the original row-by-row data transfer loads the whole partition into memory anyway. I just wanted to say that sending data row by row could be the way how to optimize memory consumption in SW.

Is the chunk compression ratio significant?

h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
```diff
@@ -0,0 +1,89 @@
...
+public final class ChunkAutoBufferWriter implements Closeable  {
+    
+    private final AutoBuffer buffer;
+    
+    public ChunkAutoBufferWriter(OutputStream outputStream) {
+        this.buffer = new AutoBuffer(outputStream, false);
+    }
+
+    public void writeChunk(String frameName, int chunkId, byte[] expectedTypes, int[] selectedColumnIndices) {
+        Frame frame = DKV.getGet(frameName);
```
Pscheidl:
```suggestion
        final Frame frame = DKV.getGet(frameName);
```

Can be final

h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
```diff
@@ -0,0 +1,89 @@
...
+        Chunk[] chunks = ChunkUtils.getChunks(frame, chunkId);
```
Pscheidl:
```suggestion
        final Chunk[] chunks = ChunkUtils.getChunks(frame, chunkId);
```

h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
```diff
@@ -0,0 +1,89 @@
...
+        int numberOfRows = chunks[0]._len;
```
Pscheidl:
```suggestion
        final int numberOfRows = chunks[0]._len;
```

h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
```diff
@@ -0,0 +1,89 @@
...
+        ExternalFrameUtils.sendInt(buffer, numberOfRows);
+
+        // buffered string to be reused for strings to avoid multiple allocation in the loop
+        BufferedString valStr = new BufferedString();
```
Pscheidl:
```suggestion
        final BufferedString valStr = new BufferedString();
```

h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
```diff
@@ -0,0 +1,89 @@
...
+
+public final class ChunkAutoBufferWriter implements Closeable  {
+    
+    private final AutoBuffer buffer;
+    
+    public ChunkAutoBufferWriter(OutputStream outputStream) {
+        this.buffer = new AutoBuffer(outputStream, false);
+    }
+
+    public void writeChunk(String frameName, int chunkId, byte[] expectedTypes, int[] selectedColumnIndices) {
```
Pscheidl: Do you really need to set the vectors types manually ? 

Introducing a function

```java
    private static byte[] vectorTypes(final Frame frame, final int[] vectors){
        final byte[] vectorTypes = new byte[vectors.length];

        for (int i = 0; i < vectors.length; i++) {
            vectorTypes[i] = frame.vec(i).get_type();
        }
        
        return vectorTypes;
    }
```

and then adding one line somewhere at the beginning of this method:

```java
final byte[] expectedTypes = vectorTypes(frame, selectedColumnIndices);
```

and then removing the expectedTypes would do the trick.

Pscheidl: Oh, I see, those are part of the client request. Do they serve for type checking, or for conversion as well ? If yes, what to do with incovertible types then ?

Do we check the types inside the frame are equal to the types requested ? Something like that.

```java
    private static boolean verifyVectorTypes(final Frame frame, final int[] vectors, final byte[] expectedTypes){

        for (int i = 0; i < vectors.length; i++) {
            if (expectedTypes[i] != frame.vec(vectors[i]).get_type()) return false;
        }
        
        return true;
    }
```

Pscheidl: Should we check the `expectedTypes` really match the types of vectors red from the Frame ? The type mapping has already been done by @jakubhava in `ExternalFrameUtils.vecTypesFromExpectedTypes`.

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+  @Override
+  protected void doGet(HttpServletRequest request, HttpServletResponse response) {
+    String uri = ServletUtils.getDecodedUri(request);
+    try {
+      RequestParameters parameters = extractRequestParameters(request);
+      response.setContentType("application/octet-stream");
+      ServletUtils.setResponseStatus(response, HttpServletResponse.SC_OK);
+      OutputStream outputStream = response.getOutputStream();
+      ChunkAutoBufferWriter writer = new ChunkAutoBufferWriter(outputStream);
+      try {
```
Pscheidl: I suggest using try-with-resources block. 

```java
try(ChunkAutoBufferWriter writer = new ChunkAutoBufferWriter(response.getOutputStream())) {
```
1. The finally black may be deleted
2. The following lines above the try-catch block as well

```java
    OutputStream outputStream = response.getOutputStream();
      ChunkAutoBufferWriter writer = new ChunkAutoBufferWriter(outputStream);
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    buffer.get(selectedColumnIndices);
+            
+    return new RequestParameters(frameName, chunkId, expectedTypes, selectedColumnIndices); 
+  }
+  
+  @Override
+  protected void doGet(HttpServletRequest request, HttpServletResponse response) {
+    String uri = ServletUtils.getDecodedUri(request);
+    try {
+      RequestParameters parameters = extractRequestParameters(request);
```
Pscheidl:
```suggestion
      final RequestParameters parameters = extractRequestParameters(request);
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    IntBuffer buffer = ByteBuffer.wrap(selectedColumnsBytes).order(ByteOrder.BIG_ENDIAN).asIntBuffer();
+    int[] selectedColumnIndices = new int[buffer.remaining()];
+    buffer.get(selectedColumnIndices);
+            
+    return new RequestParameters(frameName, chunkId, expectedTypes, selectedColumnIndices); 
+  }
+  
+  @Override
+  protected void doGet(HttpServletRequest request, HttpServletResponse response) {
+    String uri = ServletUtils.getDecodedUri(request);
```
Pscheidl:
```suggestion
    final String uri = ServletUtils.getDecodedUri(request);
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    public RequestParameters(String frameName, int chunkId, byte[] expectedTypes, int[] selectedColumnIndices) {
+      this.frameName = frameName;
+      this.chunkId = chunkId;
+      this.expectedTypes = expectedTypes;
+      this.selectedColumnIndices = selectedColumnIndices;
+    }
+  }
+  
+  private String getParameterAsString(HttpServletRequest request, String parameterName) {
+    String result = request.getParameter(parameterName);
```
Pscheidl:
```suggestion
    final String result = request.getParameter(parameterName);
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    if (result == null) {
+      throw new RuntimeException(String.format("Cannot find value for parameter \'%s\'", parameterName));
+    }
+    return result;
+  }
+  
+  private RequestParameters extractRequestParameters(HttpServletRequest request) {
+    String frameName = getParameterAsString(request, "frame_name");
```
Pscheidl:
```suggestion
    final String frameName = getParameterAsString(request, "frame_name");
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    
+    String chunkIdString = getParameterAsString(request, "chunk_id");
```
Pscheidl:
```suggestion
    final String chunkIdString = getParameterAsString(request, "chunk_id");
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    int chunkId = Integer.parseInt(chunkIdString);
+    
+    String expectedTypesString = getParameterAsString(request, "expected_types");
```
Pscheidl:
```suggestion
    final String expectedTypesString = getParameterAsString(request, "expected_types");
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    byte[] expectedTypes = Base64.getDecoder().decode(expectedTypesString);
+
+    String selectedColumnsString = getParameterAsString(request, "selected_columns");
```
Pscheidl:
```suggestion
    final String selectedColumnsString = getParameterAsString(request, "selected_columns");
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    byte[] selectedColumnsBytes = Base64.getDecoder().decode(selectedColumnsString);
+    IntBuffer buffer = ByteBuffer.wrap(selectedColumnsBytes).order(ByteOrder.BIG_ENDIAN).asIntBuffer();
```
Pscheidl:
```suggestion
    final IntBuffer buffer = ByteBuffer.wrap(selectedColumnsBytes).order(ByteOrder.BIG_ENDIAN).asIntBuffer();
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+            parameters.chunkId, 
+            parameters.expectedTypes,
+            parameters.selectedColumnIndices);
+      } finally {
+        writer.close();
+      }
+    } catch (Exception e) {
+      ServletUtils.sendErrorResponse(response, e, uri);
+    } finally {
+      ServletUtils.logRequest("GET", request, response);
```
Pscheidl:
```suggestion
      ServletUtils.logRequest(request.getMethod(), request, response);
```

h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
```diff
@@ -0,0 +1,89 @@
...
+
+import static water.ExternalFrameUtils.*;
+import static water.ExternalFrameUtils.EXPECTED_STRING;
+
+public final class ChunkAutoBufferWriter implements Closeable  {
+    
+    private final AutoBuffer buffer;
+    
+    public ChunkAutoBufferWriter(OutputStream outputStream) {
+        this.buffer = new AutoBuffer(outputStream, false);
```
Pscheidl:
```suggestion
        Objects.requireNonNull(outputStream);
        this.buffer = new AutoBuffer(outputStream, false);
```

h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
```diff
@@ -0,0 +1,89 @@
+package water;
+
+import water.fvec.Chunk;
+import water.fvec.ChunkUtils;
+import water.fvec.Frame;
+import water.parser.BufferedString;
+
+import java.io.Closeable;
+import java.io.IOException;
```
Pscheidl:
```suggestion
```

Unused import.

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+ * This servlet class handles GET requests for the path /3/Chunk
+ * It requires 4 get parameters
+ * - frame_name - a unique string identifier of H2O Frame
+ * - chunk_id - a unique identifier of the chunk within the H2O Frame
+ * - expected_type - byte array encoded in Base64 encoding. The types corresponds to the `selected_columns` parameter
+ * - selected_columns - selected columns indices encoded int Base64 encoding.
+ * The result is represented as a stream of binary data. Data are encoded to AutoBuffer row by row. 
+ * The data stream starts with the integer representing the number of rows.
+ */
+public class ChunkServlet extends HttpServlet {
```
Pscheidl:
```suggestion
public final class ChunkServlet extends HttpServlet {
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    
+    public RequestParameters(String frameName, int chunkId, byte[] expectedTypes, int[] selectedColumnIndices) {
+      this.frameName = frameName;
+      this.chunkId = chunkId;
+      this.expectedTypes = expectedTypes;
+      this.selectedColumnIndices = selectedColumnIndices;
+    }
+  }
+  
+  private String getParameterAsString(HttpServletRequest request, String parameterName) {
```
Pscheidl:
```suggestion
  private String getParameterAsString(final HttpServletRequest request, final String parameterName) {
```

h2o-core/src/main/java/water/api/ChunkServlet.java
```diff
@@ -0,0 +1,91 @@
...
+    String result = request.getParameter(parameterName);
+    if (result == null) {
+      throw new RuntimeException(String.format("Cannot find value for parameter \'%s\'", parameterName));
+    }
+    return result;
+  }
+  
+  private RequestParameters extractRequestParameters(HttpServletRequest request) {
```
Pscheidl:
```suggestion
  private RequestParameters extractRequestParameters(final HttpServletRequest request) {
```

h2o-core/src/main/java/water/ChunkAutoBufferReader.java
```diff
@@ -0,0 +1,93 @@
...
+        this.buffer = new AutoBuffer(inputStream);
+        this.numRows = readInt();
+    }
+
+    public int getNumRows() {
+        return numRows;
+    }
+
+    public boolean readBoolean() {
+        boolean data = buffer.getZ();
```
Pscheidl:
```suggestion
        final boolean data = buffer.getZ();
```

h2o-core/src/main/java/water/ChunkAutoBufferReader.java
```diff
@@ -0,0 +1,93 @@
...
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public byte readByte() {
+        byte data = buffer.get1();
```
Pscheidl:
```suggestion
        final byte data = buffer.get1();
```

h2o-core/src/main/java/water/ChunkAutoBufferReader.java
```diff
@@ -0,0 +1,93 @@
...
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public char readChar() {
+        char data = buffer.get2();
```
Pscheidl:
```suggestion
        final char data = buffer.get2();
```

h2o-core/src/main/java/water/ChunkAutoBufferReader.java
```diff
@@ -0,0 +1,93 @@
...
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public short readShort() {
+        short data = buffer.get2s();
```
Pscheidl:
```suggestion
        final short data = buffer.get2s();
```

h2o-core/src/main/java/water/ChunkAutoBufferReader.java
```diff
@@ -0,0 +1,93 @@
...
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public int readInt() {
+        int data = buffer.getInt();
```
Pscheidl:
```suggestion
        final int data = buffer.getInt();
```

h2o-core/src/main/java/water/ChunkAutoBufferReader.java
```diff
@@ -0,0 +1,93 @@
...
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public long readLong() {
+        long data = buffer.get8();
```
Pscheidl:
```suggestion
        final long data = buffer.get8();
```

h2o-core/src/main/java/water/ChunkAutoBufferReader.java
```diff
@@ -0,0 +1,93 @@
...
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public float readFloat() {
+        float data = buffer.get4f();
```
Pscheidl:
```suggestion
        final float data = buffer.get4f();
```

h2o-core/src/main/java/water/ChunkAutoBufferReader.java
```diff
@@ -0,0 +1,93 @@
...
+        isLastNA = ExternalFrameUtils.isNA(data);
+        return data;
+    }
+
+    public double readDouble() {
+        double data = buffer.get8d();
```
Pscheidl:
```suggestion
        final double data = buffer.get8d();
```

h2o-core/src/main/java/water/ChunkAutoBufferReader.java
```diff
@@ -0,0 +1,93 @@
...
+        isLastNA = ExternalFrameUtils.isNA(data);
+        return data;
+    }
+
+    public String readString() {
+        String data = buffer.getStr();
```
Pscheidl:
```suggestion
        final String data = buffer.getStr();
```

h2o-core/src/main/java/water/ChunkAutoBufferReader.java
```diff
@@ -0,0 +1,93 @@
...
+        isLastNA = ExternalFrameUtils.isNA(buffer, data);
+        return data;
+    }
+
+    public Timestamp readTimestamp() {
+        Timestamp data = new Timestamp(buffer.get8());
```
Pscheidl:
```suggestion
        final Timestamp data = new Timestamp(buffer.get8());
```

h2o-core/src/main/java/water/ChunkAutoBufferWriter.java
mn-mikke: Great idea, thanks!

h2o-core/src/main/java/water/api/ChunkServlet.java
mn-mikke: Thanks, will do.

mn-mikke:
@Pscheidl thanks for your valuable comments, I've tried to address all of them. I also added validation of input parameters so may I ask you to take a look at the PR one more time?

jakubhava:
@michalkurka we had sync up with @mn-mikke and Marek was explaining that you would prefer keeping the rest api only used by SW on Sparkling Water side.

This makes great sense, however then we would need to keep extending the jars which is the main thing we would like to get rid of. Ideally, the users should be able to use h2o drivers without any enhancements in this client-less scenarios.

Maybe we can do some compromise? We could create a separate repo, something like `sparkling-water-rest-api-extensions` where we would put these endpoints and H2O-3 assembly would depend on it, what do you think?

That way we would ensure H2O-3 does not need to maintain our specific endpoints, but it is included in the final builds of H2O

jakubhava:
Just note to @michalkurka comment about unpacking the chunk: If we would send the chunk as they are, compressed, we could also remove our optimizations when sending the data (brings another complexity) and we could read data from chunks directly using methods available on chunks.

michalkurka:
> @michalkurka we had sync up with @mn-mikke and Marek was explaining that you would prefer keeping the rest api only used by SW on Sparkling Water side.
> 
> This makes great sense, however then we would need to keep extending the jars which is the main thing we would like to get rid of. Ideally, the users should be able to use h2o drivers without any enhancements in this client-less scenarios.
> 
> Maybe we can do some compromise? We could create a separate repo, something like `sparkling-water-rest-api-extensions` where we would put these endpoints and H2O-3 assembly would depend on it, what do you think?
> 
> That way we would ensure H2O-3 does not need to maintain our specific endpoints, but it is included in the final builds of H2O

@jakubhava You could use `-libjars` to add this functionality dynamically and build the SW extension jar for H2O (wich all the SW specifics). For users launching manually, they would need to download SW distribution add add this jar (this is not common). For all other users - SW could add this automatically.

michalkurka:
The other option is to move these SW extensions out of the h2o-core into a module of the h2o-3 repository.

jakubhava:
@michalkurka I see, I forgot that actually Sparkling Water is the one who is responsible for creating the `hadoop jar` command.

I like this approach!. If @mn-mikke agrees, I would go with this solution as it will give us also more flexibility when it comes to developing changes related to external backend.

mn-mikke:
The logic was moved to https://github.com/h2oai/sparkling-water/pull/1752. Closing the PR.

PR #4219, opened 2020-01-15
PUBDEV-7140 - Checkponting works with both model and string key in py…

Pscheidl:
…thon

https://0xdata.atlassian.net/browse/PUBDEV-7140

### DIFF
--- a/h2o-docs/src/booklets/v2_2015/source/DeepLearning_Vignette_code_examples/deeplearning_checkpoint.py
+++ b/h2o-docs/src/booklets/v2_2015/source/DeepLearning_Vignette_code_examples/deeplearning_checkpoint.py
@@ -2,8 +2,9 @@
 # using the `checkpoint` argument
 model_chkp = H2ODeepLearningEstimator(
         checkpoint=model,
+        activation="RectifierWithDropout",
         distribution="multinomial",
-        hidden=model.actual_params["hidden"],
+        hidden=[32,32,32],
         input_dropout_ratio=0.2,
         sparse=True,
         l1=1e-5,
--- a/h2o-docs/src/booklets/v2_2015/source/DeepLearning_Vignette_code_examples/deeplearning_checkpoint.py
+++ b/h2o-docs/src/booklets/v2_2015/source/DeepLearning_Vignette_code_examples/deeplearning_checkpoint.py
@@ -3,8 +3,7 @@
 model_chkp = H2ODeepLearningEstimator(
         checkpoint=model,
         distribution="multinomial",
-        activation="RectifierWithDropout",
-        hidden=[32,32,32],
+        hidden=model.actual_params["hidden"],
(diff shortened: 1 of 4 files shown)


michalkurka:
@Pscheidl this fix will work however getter will return a different value (key) than what user passes in the setter

do you know why the automatic conversion fails?

https://github.com/h2oai/h2o-3/blob/master/h2o-py/h2o/backend/connection.py#L702

Pscheidl:
I'll see why it fails, did not know about the automatic conversion. Tried to find it, did not expect it in connection.py.

My fix is not correct then.

Pscheidl:
Ok @michalkurka, I think I'm talking about something completely different.

When the model training is issued by calling `h2o.api` and the `data` payload with `parms` is converted, this actually checks for `elif isinstance(value, dict):`- H2OEstimator is not a dict. We could convert all H2O estimators to KeyV3 in `connection.py`, however I can't foresee the consequences and would rather avoid such a big leap.

As for the getter, now it returns `str` in both cases, where the key is represented as a String. This makes sense, as the value is internally converted to String.

Pscheidl:
@michalkurka I have completely reverted the previous changes, only kept the test, but modified it.

Currently, the key string is extracted before training is invoked, so that backend gets the key only. This only applies to checkpoints, therefore nothing else is potentially broken. Also, the getters now return exactly the value given.

Pscheidl:
@michalkurka In R, it seems the `@model_id` slot is used and there is no problem in keeping it that way. H2O even does validation in `.h2o.checkAndUnifyModelParameters` to ensure even checkpoint is a `character` and not `H2OModel`.

Pscheidl:
Related test failure:

http://mr-0xc1:8080/job/h2o-3-pipeline/job/PR-4219/3/artifact/py2.7-booklets/h2o-3/h2o-docs/src/booklets/v2_2015/source/results/failed/pybooklet.deeplearning.vignette.py.out.txt

This test never worked as intended, as `deeplearning_checkpoint.py` defines `        hidden=[32,32,32],` and the checkpointed model `[32,16,8]` - however, this was never checked, as there was no checkpoint and the DL model just started from scratch.

Pscheidl:
Found another bug - for some reason, the `hidden` getter does not return the actual value used if there was none explicitely set before. Interesting.

Pscheidl:
It turns out our Python estimators have getters which look into `self._parms`, instead of `self.params` for actual value. Not sure if this is wanted or not, but it seems like they all do that. Honestly, I did not know users have to look into `.params`, or alternatively into `.actual_params` to get the actual value. The fix is straightforward then.

Pscheidl:
Finally fixed, but the checkpointed model changes few other parameters that are not allowed to be changed, e.g. activation. Fixed, waiting for tests. Might do this a couple times more.

PR #4220, opened 2020-01-15
PUBDEV-7201: sklearn wrappers should expose wrapped estimator as a public property

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-7201

the property used to be "python-hidden". As some editors like Jupyter notebook use this convention to remove those hidden properties from auto-completion hints, exposing it here as a proper property.

Also, minor fix on some internal property that couldn't be pickled correctly.


ledell:
Thanks!

PR #4221, opened 2020-01-15
PUBDEV-6460: We can use NM_HOST to find out IP of H2O node

michalkurka:
This won't be directly exposed via a parameter, instead users can add an
option to H2O driver like this

    hadoop jar h2odriver.jar -Dh2o.ip.envvar=NM_HOST -mapperXmx 4g -n 10

This will make H2O lookup the hostname (of the associated node manager) and
use this IP instead of the built-in heuristics.

If H2O is running containerized in Docker, this approach might fail but we
might be able to get the IP from a different property (this is why we are
leaving the setting up to the user).

### DIFF
--- a/h2o-mapreduce-generic/src/main/java/water/hadoop/h2omapper.java
+++ b/h2o-mapreduce-generic/src/main/java/water/hadoop/h2omapper.java
@@ -6,6 +6,7 @@ import org.apache.hadoop.mapreduce.Counter;
 import org.apache.hadoop.mapreduce.Mapper;
 import water.H2O;
 import water.util.Log;
+import water.util.StringUtils;
 
 import java.io.*;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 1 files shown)


michalkurka:
@bilcus I think we could add this `-Dh2o.ip.envvar=NM_HOST` to Steam by default (but there should also be an option to override this by the user).

h2o-mapreduce-generic/src/main/java/water/hadoop/h2omapper.java
```diff
@@ -243,6 +209,54 @@ private int run2(Context context) throws IOException {
...
+      // Need to modify this config here as we don't know the destination dir for keys when generating it
+      if("default-security.config".equals(basename)) {
+        modifyKeyPath(fileName, ice_root);
+      }
+    }
+
+    String ipEnvVar = conf.get(H2O_IP_ENVVAR);
+    if (!StringUtils.isNullOrEmpty(ipEnvVar)) {
+      argsList.add("-ip");
+      argsList.add(System.getenv(ipEnvVar));
```
honzasterba: Should we check the env var values us nit empty?

michalkurka: Check added - probably to report an explicit error. Thanks.

PR #4222, opened 2020-01-15
PUBDEV-7194: warn user if validation frame is set with CV enabled

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-7194


ledell:
Thanks!

PR #4227, opened 2020-01-16
PUBDEV-7212 - Rank vectors of Spearman correlation might have differe…

Pscheidl:
…nt chunk layout

https://0xdata.atlassian.net/browse/PUBDEV-7212

The crucial stacktrace is 

```
java.lang.IllegalArgumentException
 [1] "java.lang.IllegalArgumentException: Vec C2 is not compatible with the rest of the frame"                     
 [2] "    water.fvec.Frame.checkCompatibility(Frame.java:273)"                                                     
 [3] "    water.fvec.Frame.<init>(Frame.java:180)"                                                                 
 [4] "    water.fvec.Frame.<init>(Frame.java:138)"                                                                 
 [5] "    water.fvec.Frame.<init>(Frame.java:133)"                                                                 
 [6] "    water.MRTask.doAll(MRTask.java:378)"                                                                     
 [7] "    water.MRTask.doAll(MRTask.java:377)"                                                                     
 [8] "    water.rapids.ast.prims.advmath.SpearmanCorrelation.calculateMeans(SpearmanCorrelation.java:343)"         
 [9] "    water.rapids.ast.prims.advmath.SpearmanCorrelation.calculate(SpearmanCorrelation.java:45)"               
[10] "    water.rapids.ast.prims.advmath.AstCorrelation.spearman(AstCorrelation.java:85)"                          
[11] "    water.rapids.ast.prims.advmath.AstCorrelation.apply(AstCorrelation.java:71)"                             
```

## Cause

When two different frames (different vecs with different keys) are parsed and then sent to `AstCor` for Spearman's Rho calculation, the vectors might have different chunk layout. This is not an issue until the columns with ranks are created. Those effectively copy the chunk layout of their predecessors - which is wanted (might be e.g. categorical - no need to caulcuate ranks for categoricals). However, such vectors are incompatible when `Frame.isCompatible` method is invoked before the beginning of `MeanTask`, which is an MRTask originally invoked in the following way:

```
      return new MeanTask()
              .doAll(vecs)._means;
```
The vectors passed as an argument to `doAll(vecs)` method are checked for compatibility. Otherwise, it would be hard to iterate over them in `map` phase.

## Solution

One solution is to make the vectors compatible. Which, in the worst case, might end up by effectively duplicating the data.

A better solution is to check vector compatibility. If vectors are compatible, then the means are calculated in one MRTask named `MeanTask`. If not, then the `MeanTask` is executed exactly twice instead of once, as there are always two columns with ranks compared.

To introduce a general vector compatibility check with decent O(n), I've created `public static boolean areVectorsMutuallyCompatible(Vec... vectors) ` in `VecUtils`. This check is then used in `SpearmanCorrelation.calculateMeans` method.

### DIFF
--- a/h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
@@ -8,6 +8,7 @@ import water.fvec.Frame;
 import water.fvec.Vec;
 import water.rapids.Merge;
 import water.util.FrameUtils;
+import water.util.VecUtils;
 
 import java.math.BigDecimal;
 import java.math.MathContext;
@@ -326,8 +327,8 @@ public class SpearmanCorrelation {
       throw new IllegalArgumentException("There are no vectors to calculate means from.");
     }
 
+    // Validate vectors types and length
     final long referenceVectorLength = vecs[0].length();
-
     for (int i = 0; i < vecs.length; i++) {
       if (!vecs[i].isCategorical() && !vecs[i].isNumeric()) {
         throw new IllegalArgumentException(String.format("Given vector '%s' is not numerical or categorical.",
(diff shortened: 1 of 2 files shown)


Pscheidl:
Please don't review, this solution is bad. Only fixes it for the means calculation, but fails afterwards on Spearman MR Task.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/SpearmanCorrelation.java
```diff
@@ -339,8 +340,22 @@ public double getSpearmanCorrelationCoefficient() {
       }
     }
 
-    return new MeanTask()
-            .doAll(vecs)._means;
+    if (VecUtils.areVectorsMutuallyCompatible(vecs)) {
+      return new MeanTask()
+              .doAll(vecs)._means;
+    } else {
```
michalkurka: this is not necessary

new Frame().add(null, vecs) will make the vecs compatible

Pscheidl: 1. It would be best for me to return the vecs already compatible from `rankedVectors` method.
2. (note to self) Need to return the vecs from inside the Frame, as it might end up duplicating the vector given internally, creating a new one.

Pscheidl:
Please review not @michalkurka and others.

Pscheidl:
The first version of the fix did not help at all. Trying to fix it by adding first vector via constructor and then adapting the second one to it by calling `add`. Hope this helps.

Pscheidl:
That fixed it - running tests once again to be sure it's not random.

Pscheidl:
All green - merging.

PR #4229, opened 2020-01-16
PUBDEV-7214: Add offset_column to XGBoost

hannah-tillman:


angela0xdata:
Ugh, so the link in this description is no longer valid. Can you remove the "For more information, refer to the following link" sentence from the offset_column description in each supported algorithm?

angela0xdata:
>> from the offset_column description in each supported algorithm?
This change still needs to be made where it appears in other algorithms.

PR #4234, opened 2020-01-19
Make sure XGBoost featuremap file gets cleaned-up after model training

michalkurka:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -412,7 +417,9 @@ private String createFeatureMapFile(String featureMap) {
       OutputStream os = null;
       try {
         File tmpModelDir = java.nio.file.Files.createTempDirectory("xgboost-model-" + _result.toString()).toFile();
```
honzasterba: it seems to me we do not need to create this directory at all, the file-name should be quite unique by itself, or we can add the the _result the the file-name, rather than creating unnecessary directories

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -412,7 +417,9 @@ private String createFeatureMapFile(String featureMap) {
       OutputStream os = null;
```
honzasterba: I would also modify this method to modify the return String that is then directly assigned to featureMapFileAbsolutePath is unnecesary and unredable, also I would add a check here that would throw illegal state exceptio when this method would be called and featureMapFileAbsolutePath != null or altenatively delete the file when != null before creating new one?
or maybe this needs bit more thought.

michalkurka: Thanks, the code needed a clean-up - please see my changes

PR #4236, opened 2020-01-20
[PUBDEV-7192] fix adaptTestToTrain for OneHotExplicit

honzasterba:


h2o-core/src/main/java/hex/Model.java
```diff
@@ -1248,7 +1248,7 @@ public double lift_top_group() {
 
       // still have work to do below, make sure we set the names/domains to the original user-given values
       // such that we can do the int->enum mapping and cat. encoding below (from scratch)
-      if (match) {
+      if (match && !catEncoded) {
```
sebhrusen: Change make sense, although I'm wondering if we could not skip the `checkCategoricals` logic altogether in the nested call: L1225: `checkCategoricals = !catEncoded && ...`

Also, all the tests below are passing without this change.
Could you please add a failing test for this? thx

honzasterba: I was thinking about the same lines, and its probably equal, will look into it and the tests as well.

PR #4238, opened 2020-01-21
PUBDEV-6745 - H2O Frame does not escape double quotes for CSV format

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6745
## Problem
H2O Frame, when serialized into CSV, does not escape double quotes. We only use double quotes to enclose the characters, therefore single quotes do not need to be escaped.

In CSV - all quotes not enclosing a token should be escaped by putting another quote in front of it.

## Solution

Make `Frame.CSVStream` class escape double quotes. :smile:

Introduced `static final Pattern doubleQuotePattern = Pattern.compile("\"{1}");` to compile the pattern only once. Then, all potential Strings are sent to `escapeQuotesForCsv` method that will escape the double quotes using a `Matcher` obtained from the `Pattern`.

### DIFF
--- a/h2o-core/src/main/java/water/fvec/Frame.java
+++ b/h2o-core/src/main/java/water/fvec/Frame.java
@@ -1646,16 +1646,12 @@ public class Frame extends Lockable<Frame> {
         if (!_curChks[i].isNA(_chkRow)) {
           if (v.isCategorical()) {
             final String escapedString = escapeQuotesForCsv(v.factor(_curChks[i].at8(_chkRow)));
-            sb.append('"')
-                    .append(escapedString)
-                    .append('"');
+            sb.append('"').append(escapedString).append('"');
           } else if (v.isUUID()) sb.append(PrettyPrint.UUID(_curChks[i].at16l(_chkRow), _curChks[i].at16h(_chkRow)));
           else if (v.isInt()) sb.append(_curChks[i].at8(_chkRow));
           else if (v.isString()) {
             final String escapedString = escapeQuotesForCsv(_curChks[i].atStr(tmpStr, _chkRow).toString());
-            sb.append('"')
-                    .append(escapedString)
-                    .append('"');
+            sb.append('"').append(escapedString).append('"');
           } else {
             double d = _curChks[i].atd(_chkRow);
             // R 3.1 unfortunately changed the behavior of read.csv().
@@ -1676,7 +1672,7 @@ public class Frame extends Lockable<Frame> {
       return StringUtils.bytesOf(sb);
     }
 
-    static final Pattern doubleQuotePattern = Pattern.compile("\"{1}");
+    static final Pattern doubleQuotePattern = Pattern.compile("\"");
     /**
      * Escapes  double-quotes (ASCII 34) in a String.
      *
--- a/h2o-core/src/main/java/water/fvec/Frame.java
+++ b/h2o-core/src/main/java/water/fvec/Frame.java
@@ -1659,6 +1659,9 @@ public class Frame extends Lockable<Frame> {
 
         if (escapingRequired) {
           localEscapedCategoricalVecDomains[i] = escapedDomain;
+        } else {
+          // If the domain does not need escaping, simply link to the original domain and drop the escaped array
+          localEscapedCategoricalVecDomains[i] = originalDomain;
         }
       }
 
@@ -1680,7 +1683,7 @@ public class Frame extends Lockable<Frame> {
         if (i > 0) sb.append(_parms._separator);
         if (!_curChks[i].isNA(_chkRow)) {
           if (v.isCategorical()) {
-            final String escapedString = getEscapedCategoricalLevel(i, (int) _curChks[i].at8(_chkRow), v);
+            final String escapedString = _escapedCategoricalVecDomains[i][(int) _curChks[i].at8(_chkRow)];
             sb.append('"').append(escapedString).append('"');
           } else if (v.isUUID()) sb.append(PrettyPrint.UUID(_curChks[i].at16l(_chkRow), _curChks[i].at16h(_chkRow)));
           else if (v.isInt()) sb.append(_curChks[i].at8(_chkRow));
@@ -1705,25 +1708,6 @@ public class Frame extends Lockable<Frame> {
       }
       sb.append('\n');
       return StringUtils.bytesOf(sb);
-    }
-
-    /**
-     * Returns a factor escaped for CSV format, if required.
-     *
-     * @param vectorId              Id of the categorical vector
-     * @param categoricalLevelIndex Index of the categorical level in the original domain
-     * @param vec                   Vector with it's domain
-     * @return A {@link String} representation of the categorical level, with double-quotes escaped for CSV.
-     */
-    private String getEscapedCategoricalLevel(final int vectorId, final int categoricalLevelIndex, final Vec vec) {
-      final String[] escapedDomain = _escapedCategoricalVecDomains[vectorId];
-
-      if (escapedDomain == null) {
-        return vec.factor(categoricalLevelIndex);
-      } else {
-        return escapedDomain[categoricalLevelIndex];
-      }
-
     }
 
     private static final Pattern DOUBLE_QUOTE_PATTERN = Pattern.compile("\"");
--- a/h2o-core/src/main/java/water/fvec/Frame.java
+++ b/h2o-core/src/main/java/water/fvec/Frame.java
@@ -1597,6 +1597,7 @@ public class Frame extends Lockable<Frame> {
     Chunk[] _curChks;
     int _lastChkIdx;
     public volatile int _curChkIdx; // used only for progress reporting
+    private final Map<Integer, String[]> escapedCategoricalLevels = new HashMap<>();
 
     public CSVStream(Frame fr, CSVStreamParams parms) {
       this(firstChunks(fr), parms._headers ? fr.names() : null, fr.anyVec().nChunks(), parms);
@@ -1621,13 +1622,42 @@ public class Frame extends Lockable<Frame> {
       StringBuilder sb = new StringBuilder();
       if (names != null) {
         sb.append('"').append(names[0]).append('"');
-        for(int i = 1; i < names.length; i++)
+        for (int i = 1; i < names.length; i++)
           sb.append(_parms._separator).append('"').append(names[i]).append('"');
         sb.append('\n');
       }
       _line = StringUtils.bytesOf(sb);
       _chkRow = -1; // first process the header line
       _curChks = chks;
+      escapeCategoricalVecDomains(_curChks);
+    }
+
+    /**
+     * Escapes categorical levels of vectors and puts them in a map of escaped categorical levels.
+     * Only the domains with at least one level with an escaped quote are saved. If a domain does not need
+     * any escaping, it is considered better practice to reach to the `vec.domain()` method itself and not duplicate entries
+     * in memory here.
+     * @param chunks
+     */
+    private void escapeCategoricalVecDomains(final Chunk[] chunks) {
+
+      for (int i = 0; i < chunks.length; i++) {
+        final Vec vec = chunks[i].vec();
+        if (!vec.isCategorical()) continue;
+
+        final String[] originalDomain = vec.domain();
+        final String[] escapedDomain = new String[originalDomain.length];
+
+        boolean noEscapingRequired = true;
+        for (int level = 0; level < originalDomain.length; level++) {
+          escapedDomain[level] = escapeQuotesForCsv(originalDomain[level]);
+          noEscapingRequired = noEscapingRequired && escapedDomain[level].equals(originalDomain[level]);
+        }
+
+        if (!noEscapingRequired) {
+          escapedCategoricalLevels.put(i, escapedDomain);
+        }
+      }
     }
 
     public int getCurrentRowSize() throws IOException {
@@ -1635,7 +1665,7 @@ public class Frame extends Lockable<Frame> {
       assert av > 0;
       return _line.length;
     }
-    
+
 
     byte[] getBytesForRow() {
       StringBuilder sb = new StringBuilder();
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1665,7 +1676,20 @@ public int getCurrentRowSize() throws IOException {
       return StringUtils.bytesOf(sb);
     }
 
-    @Override public int available() throws IOException {
+    static final Pattern doubleQuotePattern = Pattern.compile("\"{1}");
```
honzasterba: {1} is unnecessary

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1633,19 +1635,28 @@ public int getCurrentRowSize() throws IOException {
...
-          if( v.isCategorical() ) sb.append('"').append(v.factor(_curChks[i].at8(_chkRow))).append('"');
-          else if( v.isUUID() ) sb.append(PrettyPrint.UUID(_curChks[i].at16l(_chkRow), _curChks[i].at16h(_chkRow)));
-          else if( v.isInt() ) sb.append(_curChks[i].at8(_chkRow));
-          else if (v.isString()) sb.append('"').append(_curChks[i].atStr(tmpStr, _chkRow)).append('"');
-          else {
+        if (i > 0) sb.append(_parms._separator);
+        if (!_curChks[i].isNA(_chkRow)) {
+          if (v.isCategorical()) {
+            final String escapedString = escapeQuotesForCsv(v.factor(_curChks[i].at8(_chkRow)));
+            sb.append('"')
```
honzasterba: I know this si knitpiccking, but these would look fine on a single line (same below)

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: Yes of course, did that for readability (an escaped quote in two quotes). Removed `{1}`.

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: My "fault", originally, this was formatted in a single line. Fixed.

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1633,19 +1635,24 @@ public int getCurrentRowSize() throws IOException {
...
-        if(!_curChks[i].isNA(_chkRow)) {
-          if( v.isCategorical() ) sb.append('"').append(v.factor(_curChks[i].at8(_chkRow))).append('"');
-          else if( v.isUUID() ) sb.append(PrettyPrint.UUID(_curChks[i].at16l(_chkRow), _curChks[i].at16h(_chkRow)));
-          else if( v.isInt() ) sb.append(_curChks[i].at8(_chkRow));
-          else if (v.isString()) sb.append('"').append(_curChks[i].atStr(tmpStr, _chkRow)).append('"');
-          else {
+        if (i > 0) sb.append(_parms._separator);
+        if (!_curChks[i].isNA(_chkRow)) {
+          if (v.isCategorical()) {
+            final String escapedString = escapeQuotesForCsv(v.factor(_curChks[i].at8(_chkRow)));
```
michalkurka: Can we avoid escaping the same categorical value each time? Or maybe - since it is unlikely that categoricals will have quote characters at all, can we just have a per-column flag that would say that the domain doesn't have any categoricals that need escaping (the most common case)?

michalkurka:
Looks good, improvement suggested for the categorical case

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: 1. This very case is recognized by H2O as categorical vector. And it has quotes :smile: 
2. Excellent suggestion. Will do. Thank you.

Pscheidl: Implemented your suggestion.

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1595,6 +1597,7 @@ public CSVStreamParams setSeparator(byte separator) {
     Chunk[] _curChks;
     int _lastChkIdx;
     public volatile int _curChkIdx; // used only for progress reporting
+    private final Map<Integer, String[]> escapedCategoricalLevels = new HashMap<>();
```
honzasterba: incosistent naming

honzasterba: also I think this could be more defensive and this Map could be Immutable and as such should be a return value from escapeCategoricalVecDomains and initialized in constructor

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1619,13 +1622,42 @@ public CSVStream(Chunk[] chks, String[] names, int nChunks, CSVStreamParams parm
...
+    }
+
+    /**
+     * Escapes categorical levels of vectors and puts them in a map of escaped categorical levels.
+     * Only the domains with at least one level with an escaped quote are saved. If a domain does not need
+     * any escaping, it is considered better practice to reach to the `vec.domain()` method itself and not duplicate entries
+     * in memory here.
+     * @param chunks
+     */
+    private void escapeCategoricalVecDomains(final Chunk[] chunks) {
```
honzasterba: unnecessary argument, its a field

h2o-core/src/main/java/water/fvec/Frame.java
michalkurka: Question: Is this being serialized anywhere? Can we make it transient?

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1619,13 +1622,42 @@ public CSVStream(Chunk[] chks, String[] names, int nChunks, CSVStreamParams parm
...
+        final String[] escapedDomain = new String[originalDomain.length];
+
+        boolean noEscapingRequired = true;
+        for (int level = 0; level < originalDomain.length; level++) {
+          escapedDomain[level] = escapeQuotesForCsv(originalDomain[level]);
+          noEscapingRequired = noEscapingRequired && escapedDomain[level].equals(originalDomain[level]);
+        }
+
+        if (!noEscapingRequired) {
+          escapedCategoricalLevels.put(i, escapedDomain);
```
honzasterba: I wonder since this is indexed by column index if an simple  String[][] owuld not be faster and cheaper

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1619,13 +1622,42 @@ public CSVStream(Chunk[] chks, String[] names, int nChunks, CSVStreamParams parm
...
+        final String[] originalDomain = vec.domain();
+        final String[] escapedDomain = new String[originalDomain.length];
+
+        boolean noEscapingRequired = true;
+        for (int level = 0; level < originalDomain.length; level++) {
+          escapedDomain[level] = escapeQuotesForCsv(originalDomain[level]);
+          noEscapingRequired = noEscapingRequired && escapedDomain[level].equals(originalDomain[level]);
+        }
+
+        if (!noEscapingRequired) {
```
honzasterba: double negative, would be nicer as just if (escapingRequired)

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1665,7 +1702,40 @@ public int getCurrentRowSize() throws IOException {
...
+    /**
+     * Returns a factor escaped for CSV format, if required.
+     *
+     * @param vectorId              Id of the categorical vector
+     * @param categoricalLevelIndex Index of the categorical level in the original domain
+     * @param vec                   Vector with it's domain
+     * @return A {@link String} representation of the categorical level, with double-quotes escaped for CSV.
+     */
+    private String getEscapedCategoricalLevel(final int vectorId, final int categoricalLevelIndex, final Vec vec) {
+      String[] domain = escapedCategoricalLevels.get(vectorId);
```
honzasterba: would rename this to escapedDomain to make it more readable

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1665,7 +1702,40 @@ public int getCurrentRowSize() throws IOException {
...
+
+      if (domain == null) {
+        return vec.factor(categoricalLevelIndex);
+      } else {
+        return domain[categoricalLevelIndex];
+      }
+
+    }
+
+    static final Pattern doubleQuotePattern = Pattern.compile("\"");
```
honzasterba: constant should be named as ALL_CAPS, also can be private

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: The field is set only line line above this method is called. If someone moves it somewhere else for any reason, it could have nasty consequences which are not obvious. Therefore I have the argument in place, to be 100% sure the chunks are required.

But I'm ok with changing it.

honzasterba: good point, for generators called from constructor this is always a tough call

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: Cheaper in terms of memory - slightly. Faster ? Both are accessible in constant time, yet array would be faster. Tiny bit faster. But I don't know how to make it immutable inside, which I can do with a map. I'll use immutable map, WDYT ? Because of the immutability. The performance will be good anyway.

honzasterba: yes, just thinking out loud, another option is immutable list, but since its sparse, map is probably best

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: Went through usages, doesn't seem like it. `ExportTaskDriver` uses it in `compute2` method, locally. Not serializing it. The same applies to `FrameUtils.EstimateTaskSize`, which uses this as a local instance in `map` method. This is safe to be marked as transient.

michalkurka: thank you, please make it transient then

Pscheidl: Done. Thank you.

Pscheidl: Aligned the naming to the corresponding method. Also made the method return immutable set. Thank you.

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: Yeah, that's nasty. Corrected.

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: Oops. Fixed.

h2o-core/src/main/java/water/fvec/Frame.java
michalkurka: I like the String[][] suggestion, the issue with this approach is unnecessary autoboxing just to lookup the value

Pscheidl: As @michalkurka pointed out, there is implicit conversion from int to Integer involved. This makes me choose array over Map.

Pscheidl: I'll still keep the method and it will still return `String[][]` - to make it obvious the method sets the array of encoded categoricals in the constructor. One reference copy more.

Pscheidl: Done.

h2o-core/src/main/java/water/fvec/Frame.java
Pscheidl: Fixed.

h2o-core/src/main/java/water/fvec/Frame.java
honzasterba: i ment the convention to use _ as prefix for instance vars, I dont like it and for new classes I dont honor it, but for old classes I think we should keep them consitent and either remove all _ or add new fields with _

Pscheidl: Have the same approach personally. I just forgot to add it in this case. Done, it's there.

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1619,13 +1622,49 @@ public CSVStream(Chunk[] chks, String[] names, int nChunks, CSVStreamParams parm
...
+        boolean escapingRequired = false;
+        for (int level = 0; level < originalDomain.length; level++) {
+          escapedDomain[level] = escapeQuotesForCsv(originalDomain[level]);
+          escapingRequired = escapingRequired || !escapedDomain[level].equals(originalDomain[level]);
+        }
+
+        if (escapingRequired) {
+          localEscapedCategoricalVecDomains[i] = escapedDomain;
+        } else {
+          localEscapedCategoricalVecDomains[i] = null;
```
honzasterba: unnecessary - but I like that it sort of explains the code below, but still unecessary.

Pscheidl: That's exactly why it's there, otherwise it's of course initialized to null from the very beginning. I'll get rid of my personal style and just remove it.

Pscheidl: Done.

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1619,13 +1622,47 @@ public CSVStream(Chunk[] chks, String[] names, int nChunks, CSVStreamParams parm
...
+      final String[][] localEscapedCategoricalVecDomains = new String[chunks.length][];
+
+      for (int i = 0; i < chunks.length; i++) {
+        final Vec vec = chunks[i].vec();
+        if (!vec.isCategorical()) continue;
+
+        final String[] originalDomain = vec.domain();
+        final String[] escapedDomain = new String[originalDomain.length];
+
+        boolean escapingRequired = false;
```
michalkurka: This adds complexity and doesn't IMHO address any issue. You can simply keep the original domain and avoid switching between escaped/original in method getEscapedCategoricalLevel.

Pscheidl: Got rid of the null checking in `getEscapedCategoricalLevel` - this is definitely unnecessary complexity.

But I can't see how to drop the `escapingRequired` - at the end, we need to either drop the escaped domain if it's the same and use reference to the original `vec.domain()` or use the escaped domain. And we need something to base the decision on, and that is the `escapingRequired`. 

Doing `final String[] escapedDomain = vec.domain()` and then modifying it would mean the matcher would modify the original domain. Or we'd need to duplicate it, which goes against our aim to save memory.

michalkurka: Looks good to me now, thank you

PR #4242, opened 2020-01-22
[PUBDEV-6502] fix XGBoost multinode issues

honzasterba:
- auto-limit the number of CPU used when running multiple H2O nodes on one machine
- fix cleanup issues when XGBoost training job is canceled
- enable multi-node xgboost back for AutoML
- some minor code improvements

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
@@ -310,7 +310,6 @@ public class XGBoost extends ModelBuilder<XGBoostModel,XGBoostModel.XGBoostParam
         model._output._sparse = isTrainDatasetSparse();
       }
 
-      XGBoostSetupTask setupTask = null;
       File featureMapFile = null;
       try {
         XGBoostSetupTask.FrameNodes trainFrameNodes = XGBoostSetupTask.findFrameNodes(_train);
@@ -336,37 +335,21 @@ public class XGBoost extends ModelBuilder<XGBoostModel,XGBoostModel.XGBoostParam
         if (_parms.hasCheckpoint()) {
           checkpointBytes = model.model_info()._boosterBytes;
         }
-        setupTask = new XGBoostSetupTask(model, _parms, boosterParms, checkpointBytes, getWorkerEnvs(rt), trainFrameNodes).run();
+        XGBoostSetupTask setupTask = new XGBoostSetupTask(model, _parms, boosterParms, checkpointBytes, getWorkerEnvs(rt), trainFrameNodes).run();
         try {
-          // initial iteration
           XGBoostUpdateTask nullModelTask = new XGBoostUpdateTask(setupTask, 0).run();
           BoosterProvider boosterProvider = new BoosterProvider(model.model_info(), featureMapFile, nullModelTask);
-
-          // train the model
           scoreAndBuildTrees(setupTask, boosterProvider, model);
-
-          // shutdown rabit & XGB native resources
-          XGBoostCleanupTask.cleanUp(setupTask);
-          setupTask = null;
-
-          waitOnRabitWorkers(rt);
         } finally {
+          XGBoostCleanupTask.cleanUp(setupTask);
           stopRabitTracker(rt);
         }
       } catch (XGBoostError xgBoostError) {
-        xgBoostError.printStackTrace();
         throw new RuntimeException("XGBoost failure", xgBoostError);
       } finally {
         if (featureMapFile != null) {
           if (! featureMapFile.delete()) {
             Log.warn("Unable to delete file " + featureMapFile + ". Please do a manual clean-up.");
-          }
-        }
-        if (setupTask != null) {
-          try {
-            XGBoostCleanupTask.cleanUp(setupTask);
-          } catch (Exception e) {
-            Log.err("XGBoost clean-up failed - this could leak memory!", e);
           }
         }
         // Unlock & save results
@@ -461,19 +444,9 @@ public class XGBoost extends ModelBuilder<XGBoostModel,XGBoostModel.XGBoostParam
       }
     }
 
-    // RT should not be started for 1 node clouds
-    private void waitOnRabitWorkers(RabitTrackerH2O rt) {
-      if(H2O.CLOUD.size() > 1) {
-        rt.waitFor(0);
-      }
-    }
-
-    /**
-     *
-     * @param rt Rabit tracker to stop
-     */
     private void stopRabitTracker(RabitTrackerH2O rt) {
       if(H2O.CLOUD.size() > 1) {
+        rt.waitFor(0);
         rt.stop();
       }
     }
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
@@ -23,9 +23,8 @@ import water.util.Log;
 import water.util.SBPrintStream;
 
 import java.io.*;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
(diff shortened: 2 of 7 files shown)


honzasterba:
There were two issues that this PR adresses:

First when running multiple nodes on one physical hosts OMP can fail to allocate native threads (since every xgboost process tryes to take all the available cores), fix this by checking how many nodes are running on one host and set nthread accordingly.

Second when the first issue was fixed sometimes when the XGBoost training was canceled by AutoML this would cause the RabitTracker thread to be interrupted and socket connections of rabit network being dropped which in turn led to the JVM being hard-exited by xgboost native code. Executing the full rabit shutdown on train interrupt fixes this issue.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -336,39 +335,23 @@ final void buildModelImpl() {
...
           BoosterProvider boosterProvider = new BoosterProvider(model.model_info(), featureMapFile, nullModelTask);
-
-          // train the model
           scoreAndBuildTrees(setupTask, boosterProvider, model);
-
-          // shutdown rabit & XGB native resources
-          XGBoostCleanupTask.cleanUp(setupTask);
-          setupTask = null;
-
-          waitOnRabitWorkers(rt);
```
michalkurka: Why is this not needed?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -440,7 +446,24 @@ protected XGBoostModel deepClone(Key<XGBoostModel> result) {
   }
   
   private static int getMaxNThread() {
-    return Integer.getInteger(SYSTEM_PROP_PREFIX + "xgboost.nthread", H2O.ARGS.nthreads);
+    if (System.getProperty(PROP_NTHREAD) != null) {
```
michalkurka: What will happen when there are other H2O clusters running on the same physical server? Would that be an issue?

honzasterba: yes, that can cause issues

michalkurka:
@honzasterba code looks good, can you please add an explanation of what happens when we don't 

- auto-limit the number of CPU used when running multiple H2O nodes on one machine

? Do we hit some OS limits?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitTrackerH2O.java
```diff
@@ -213,13 +213,16 @@ public int waitFor(long timeout) {
...
+                Thread.currentThread().interrupt();
             }
         }
         return 0;
     }
 
     @Override
     public void uncaughtException(Thread t, Throwable e) {
+        Log.err("Uncaught exception occurred on Rabit tracker thread " + t.getName(), e);
+        Log.err(e);
```
sebhrusen: second log looks useless, doesn't it?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -440,7 +446,24 @@ protected XGBoostModel deepClone(Key<XGBoostModel> result) {
...
+        if (!checkedNodes.contains(nodeHost)) {
+          checkedNodes.add(nodeHost);
+          long cnt = Stream.of(H2O.CLOUD.members()).filter(h -> h.getIpHostString().equals(nodeHost)).count();
+          if (cnt > maxNodesPerHost) {
+            maxNodesPerHost = (int) cnt;
+          }
+        }
+      }
+      int maxThreads = H2O.ARGS.nthreads;
+      return Math.max(1, maxThreads / maxNodesPerHost);
```
sebhrusen: it's probably a corner case or a bad practice to use multiple nodes per host (outside testing), but why dividing by `maxNodesPerHost` and not by `nodesOnCurrentHost`? is it just to ensure that all the nodes have the same config, i.e. same `nthreads`?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -279,7 +281,12 @@ public static BoosterParms createParams(XGBoostParameters p, int nClasses, Strin
       params.put("eta", p._learn_rate);
     }
     params.put("max_depth", p._max_depth);
-    params.put("silent", p._quiet_mode);
+    if (System.getProperty(PROP_VERBOSITY) != null) {
+      params.put("verbosity", System.getProperty(PROP_VERBOSITY));
+    } else {
+      params.put("silent", p._quiet_mode);
+    }
+    params.put("verbosity", "3");
```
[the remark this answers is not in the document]
honzasterba: was removed

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
honzasterba: the problem is that with hadoop this can happen quite easily
the config is global so thats why maxNodesPerHost, but the more I think about this I think it can be improved, thanks for the nudge

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitTrackerH2O.java
honzasterba: the first one does not log the stacktrace, so its not.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -440,7 +446,24 @@ protected XGBoostModel deepClone(Key<XGBoostModel> result) {
...
+        String nodeHost = node.getIpHostString();
+        if (!checkedNodes.contains(nodeHost)) {
+          checkedNodes.add(nodeHost);
+          long cnt = Stream.of(H2O.CLOUD.members()).filter(h -> h.getIpHostString().equals(nodeHost)).count();
+          if (cnt > maxNodesPerHost) {
+            maxNodesPerHost = (int) cnt;
+          }
+        }
+      }
+      int maxThreads = H2O.ARGS.nthreads;
```
Pscheidl:
```suggestion
      final int maxThreads = H2O.ARGS.nthreads;
```

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
michalkurka: The issue here is that it won't most likely make anything faster because the weakest link will always slow down the others in the all-reduce setting.

I think keeping the cluster uniform is maybe better, memory allocation will be more predictable. You can make it configurable and see what is behaving better.

michalkurka: You don't see the other cluster and what they are doing, however, the single worst case gives you a glimpse of what might be going on the other nodes. For deployments where the resource sharing could happen - I would prefer stability over (a marginal) gain of speed. If the gain is significant - that is another question :)

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/rabit/RabitTrackerH2O.java
sebhrusen: oh! I'm so much used to `commons.logging` and/or `log4j` that I didn't noticed that our logger was incompatible with those: such a shame...

honzasterba: true that brother. https://0xdata.atlassian.net/browse/PUBDEV-7046

PR #4243, opened 2020-01-22
PUBDEV-7190 - Add H2O import folder support to GCS

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7190


h2o-app/build.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
   if (project.hasProperty("doIncludeMojoPipeline") && project.doIncludeMojoPipeline == "true") {
     compile project(":h2o-ext-mojo-pipeline")
   }
+  compile project(":h2o-persist-gcs")
```
Pscheidl: Intentional. Let's keep it there for easier development, may I ?

h2o-persist-gcs/build.gradle
```diff
@@ -4,7 +4,7 @@ description = "H2O Persist GCS"
 
 dependencies {
     compile project(":h2o-core")
-    compile 'com.google.cloud:google-cloud-storage:1.19.0'
+    compile 'com.google.cloud:google-cloud-storage:1.103.0'
```
Pscheidl: Upgraded version, let's provide the latest fixes.

h2o-persist-gcs/src/main/java/water/persist/PersistGcs.java
```diff
@@ -157,11 +158,18 @@ public void importFiles(String path,
...
     } else {
       try {
-        Blob blob = storageProvider.getStorage().get(bk[0], bk[1]);
-        final GcsBlob gcsBlob = GcsBlob.of(blob.getBlobId());
-        final Key k = GcsFileVec.make(path, blob.getSize());
-        keys.add(k.toString());
-        files.add(path);
+        Iterable<Blob> values = storageProvider.getStorage().list(bk[0], Storage.BlobListOption.prefix(bk[1])).getValues();
+        values.forEach(blob -> {
+          final String blobPath = new StringBuilder("gs://")
```
honzasterba: if you would do just plain string concat here javac will translate it to this exact code

h2o-app/build.gradle
honzasterba: what about the other persists, should they be here as well? is there a reason they are not and are added only to assembly?

Pscheidl: I have no idea about the reason. @michalkurka please ?

h2o-persist-gcs/src/main/java/water/persist/PersistGcs.java
Pscheidl: Ok, cool. didn't know javac will output the same bytecode. Heard of it.

Done. Thank you.

h2o-app/build.gradle
Pscheidl: For now, I removed it from the dependencies.

h2o-hadoop-common/tests/python/pyunit_gcs_import.py
```diff
@@ -1,18 +0,0 @@
-#! /usr/env/python
-
-import sys
-import os
-sys.path.insert(1, os.path.join("../../../h2o-py"))
-from tests import pyunit_utils
-import h2o
-
-def gcs_import():
-    # Just test the import works - no class clashes, no exception
```
honzasterba: I think we should keep running this test on hdp, why not?

Pscheidl: Oh my god, I did not intend to move it.

Pscheidl: Moved back.

PR #4244, opened 2020-01-22
PUBDEV-7226: Adds support for registering custom servlets

michalkurka:


h2o-core/src/main/java/water/server/ServletService.java
```diff
@@ -0,0 +1,35 @@
...
+    _loader = ServiceLoader.load(ServletProvider.class);
+  }
+
+  /**
+   * Locates all available non-core primitives of the Rapid language.
+   * @return list of Rapid primitives
+   */
+  public LinkedHashMap<String, Class<? extends HttpServlet>> getAllServlets() {
+    List<ServletProvider> providers;
+    synchronized (_loader) {
```
mn-mikke: Why the block is synchronized if `providers` is just a variable with the method.

h2o-core/src/test/java/water/webserver/EchoServletProvider.java
```diff
@@ -0,0 +1,43 @@
...
+    protected void doGet(HttpServletRequest request, HttpServletResponse response) {
+      String message = request.getQueryString();
+      String uri = ServletUtils.getDecodedUri(request);
+      try {
+        response.setContentType("text/plain");
+        ServletUtils.setResponseStatus(response, HttpServletResponse.SC_OK);
+        try (OutputStream os = response.getOutputStream()){
+          os.write(message.getBytes());
+        }
+        ServletUtils.setResponseStatus(response, HttpServletResponse.SC_OK);
```
mn-mikke: This line seems to be duplicated.

michalkurka: Thank you, fixed

h2o-core/src/main/java/water/server/ServletService.java
michalkurka: Because iterating over iterator provided by service loader is not thread-safe (and this class is a singleton).

h2o-core/src/test/java/water/webserver/CustomServletTest.java
```diff
@@ -0,0 +1,32 @@
...
+public class CustomServletTest extends TestUtil {
+
+  @BeforeClass()
+  public static void setup() {
+    stall_till_cloudsize(1);
+  }
+
+  @Test
+  public void testEchoServletWorking() throws Exception {
+    URL echoURL = new URL(H2O.getURL(H2O.ARGS.jks != null ? "https" : "http") + "/99/Echo?test-query");
```
Pscheidl:
```suggestion
    final URL echoURL = new URL(H2O.getURL(H2O.ARGS.jks != null ? "https" : "http") + "/99/Echo?test-query");
```

h2o-core/src/test/java/water/webserver/CustomServletTest.java
```diff
@@ -0,0 +1,32 @@
...
+    try (InputStream is = echoURL.openStream()) {
+      String result = new BufferedReader(new InputStreamReader(is)).lines().collect(Collectors.joining("\n"));
```
Pscheidl:
```suggestion
      final String result = new BufferedReader(new InputStreamReader(is)).lines().collect(Collectors.joining("\n"));
```

h2o-core/src/test/java/water/webserver/EchoServletProvider.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.servlet.http.HttpServletResponse;
+import java.io.OutputStream;
+import java.util.LinkedHashMap;
+
+@Ignore
+public class EchoServletProvider implements ServletProvider {
+
+  @Override
+  public LinkedHashMap<String, Class<? extends HttpServlet>> servlets() {
+    LinkedHashMap<String, Class<? extends HttpServlet>> s = new LinkedHashMap<>();
```
Pscheidl:
```suggestion
    final LinkedHashMap<String, Class<? extends HttpServlet>> s = new LinkedHashMap<>();
```

h2o-core/src/test/java/water/webserver/EchoServletProvider.java
```diff
@@ -0,0 +1,42 @@
...
+    s.put("/99/Echo", EchoServlet.class);
+    return s;
+  }
+
+  public static class EchoServlet extends HttpServlet {
+    @Override
+    protected void doGet(HttpServletRequest request, HttpServletResponse response) {
+      String message = request.getQueryString();
```
Pscheidl:
```suggestion
      final String message = request.getQueryString();
```

h2o-core/src/test/java/water/webserver/EchoServletProvider.java
```diff
@@ -0,0 +1,42 @@
...
+      String uri = ServletUtils.getDecodedUri(request);
```
Pscheidl:
```suggestion
      final String uri = ServletUtils.getDecodedUri(request);
```

h2o-core/src/test/java/water/webserver/EchoServletProvider.java
```diff
@@ -0,0 +1,42 @@
...
+      try {
+        response.setContentType("text/plain");
+        try (OutputStream os = response.getOutputStream()){
```
Pscheidl:
```suggestion
        try (final OutputStream os = response.getOutputStream()) {
```

h2o-core/src/test/java/water/webserver/H2OHttpViewImplTest.java
```diff
@@ -30,4 +32,11 @@ protected boolean isAcceptingRequests() {
 
     verify(response).setStatus(405);
   }
+
+  @Test
+  public void customServletRegistered() {
+    H2OHttpView httpView = new H2OHttpViewImpl(null);
```
Pscheidl:
```suggestion
    final H2OHttpView httpView = new H2OHttpViewImpl(null);
```

h2o-core/src/main/java/water/server/ServletProvider.java
```diff
@@ -0,0 +1,24 @@
...
+public interface ServletProvider {
+
+  /**
+   * Provides a collection of Servlets that should be registered.
+   * @return a map of context path to a Servlet class
+   */
+  LinkedHashMap<String, Class<? extends HttpServlet>> servlets();
+
+  /**
+   * Provider priority, providers with higher priority will be used first. H2O Core Provider will be used last and
```
Pscheidl: What if two have the same priority ? What happens then ? From the impl, if it's not core, it'll get overriden in the order the services are discovered, which is not predictable behavior. Should we document it ?

h2o-core/src/main/java/water/server/ServletProvider.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.servlet.http.HttpServlet;
+import java.util.LinkedHashMap;
+
+public interface ServletProvider {
+
+  /**
+   * Provides a collection of Servlets that should be registered.
+   * @return a map of context path to a Servlet class
+   */
+  LinkedHashMap<String, Class<? extends HttpServlet>> servlets();
```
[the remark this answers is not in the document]
michalkurka: I need a map that gives a predictable order. There are other ways to do it if you think this is an issue.

h2o-core/src/main/java/water/server/ServletService.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private ServletService() {
+    _loader = ServiceLoader.load(ServletProvider.class);
+  }
+
+  /**
+   * Locates all available non-core primitives of the Rapid language.
+   * @return list of Rapid primitives
+   */
+  public LinkedHashMap<String, Class<? extends HttpServlet>> getAllServlets() {
```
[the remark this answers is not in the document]
michalkurka: This doesn't look simpler to me - just my opinion ;)

h2o-core/src/main/java/water/server/ServletProvider.java
michalkurka: sure, more documentation is always better

h2o-core/src/main/java/water/server/ServletService.java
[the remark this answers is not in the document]
Pscheidl: It stays in one stream and collects directly to LinkedHashMap via flatmapping the internal Maps into a stream of entries. Does not interrupt the stream flow by creating intermediate collection to be absorbed by a LinkedHashMap afterwards. Also, the design of stream API forces you to solve key collisions explicitely (`(val1, val2) -> val2`) - this is the biggest value for me.

I don't think it's mandatory.

h2o-core/src/main/java/water/server/ServletProvider.java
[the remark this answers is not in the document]
Pscheidl: Not an issue if this is truly required. I know why you're doing this. I have the context of this PR. The future reader will probably look at this interface first. It should be mentioned why there is a LinkedHashMap as a return value.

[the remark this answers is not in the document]
michalkurka: I changed the interface, please check again if it is better now

h2o-core/src/main/java/water/server/ServletService.java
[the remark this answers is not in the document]
michalkurka: incorporated your solution with a tiny change

h2o-core/src/main/java/water/server/ServletProvider.java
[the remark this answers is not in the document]
Pscheidl: Fantastic. Thanks.

PR #4245, opened 2020-01-22
PUBDEV-7204: Fix grid search for Target Encoder

michalkurka:


h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
```diff
@@ -29,38 +29,43 @@ public TargetEncoderBuilder(final boolean startupOnce) {
...
+        double priorMean = tec.calculatePriorMean(_targetEncodingMap.entrySet().iterator().next().getValue());
+
+        for(Map.Entry<String, Frame> entry: _targetEncodingMap.entrySet()) {
+          Frame frameWithEncodingMap = entry.getValue();
+          Scope.untrack(frameWithEncodingMap.keys());
+        }
+
+        disableIgnoreConstColsFeature();
+        _targetEncoderModel._output = new TargetEncoderModel.TargetEncoderOutput(TargetEncoderBuilder.this, _targetEncodingMap, priorMean);
+        _job.update(1);
```
deil87: @michalkurka I have a concern here. It looks like we update same _job twice for every model in the grid. Here, we pass Job<Grid> into `trainModelNested` method. https://github.com/h2oai/h2o-3/blob/97624413c5bd154076716fc15622d8193933c9b8/h2o-core/src/main/java/hex/grid/GridSearch.java#L529. Then  it is being assigned to Job<M> here https://github.com/h2oai/h2o-3/blob/5a3dcffe0e0098f16686b07dd88b3f2350c1d601/h2o-core/src/main/java/hex/ModelBuilder.java#L173.
As a result same job is being updated two times:
- first time in computeImpl() of a Driver
- second time in a GridSearch loop here https://github.com/h2oai/h2o-3/blob/97624413c5bd154076716fc15622d8193933c9b8/h2o-core/src/main/java/hex/grid/GridSearch.java#L381

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
```diff
@@ -29,38 +29,43 @@ public TargetEncoderBuilder(final boolean startupOnce) {
...
+        init(true);
+
+        final int numColsRemoved = hasFoldCol() ? 2 : 1; // Response is always at the last index, fold column is on the index before.
+        final String[] encodedColumns = Arrays.copyOf(train().names(), train().names().length - numColsRemoved);
+        TargetEncoder tec = new TargetEncoder(encodedColumns);
+
+        TargetEncoderModel.TargetEncoderOutput emptyOutput =
+                new TargetEncoderModel.TargetEncoderOutput(TargetEncoderBuilder.this, new IcedHashMapGeneric<>(), Double.NaN);
+        TargetEncoderModel model = new TargetEncoderModel(dest(), _parms, emptyOutput, tec);
+        _targetEncoderModel = model.delete_and_lock(_job); // and clear & write-lock it (smashing any prior)
```
deil87: @michalkurka thanks! Now I see that model was not locked. I was confused by lock on a grid here https://github.com/h2oai/h2o-3/blob/97624413c5bd154076716fc15622d8193933c9b8/h2o-core/src/main/java/hex/grid/GridSearch.java#L114 . I thought that it is locking models as well but now I see how naive this  thought was :)

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBuilder.java
michalkurka: @deil87 I think I need to set progressUnits on TargetEncoderParams to 1 and then it will be fine

https://github.com/h2oai/h2o-3/blob/master/h2o-core/src/main/java/hex/grid/GridSearch.java#L125

Next step then is to update the job progress after each encoded columns. That would give a smoother progress bar.

deil87: @michalkurka my understanding is that `gridWork` will be equal to number of model parameters ( if progressUnits is equal to 1). This number I guess is equivalent to a 100%. But we right now updating _job.update() twice for a single model parameters... so progress bar should show `completed` when we only have built 50% of the model parameters.

michalkurka: I think you are right, the GridSearch code doesn't seem correct to me (it doesn't account for the extra +1 that GridSearch adds after a model is built). This is probably not noticeable because the amount of work for usual models is the number of iterations, here it would be more noticeable because the model is trained in one iteration (one unit of work).

I think this PR is now correct but GridSearch needs to be fixed. WDYT?

deil87: @michalkurka agree, this should be fixed in a separate PR

deil87: @michalkurka I will create a Jira and PR. I just still think that it is not N+1 but it is N+N and should be noticeable. update is in the while loop.

michalkurka: yes, +1 for each model: https://0xdata.atlassian.net/browse/PUBDEV-7224

deil87:
@michalkurka merging it soon into my PR then

michalkurka:
@deil87 no need, I will do that once all the tests are green

PR #4246, opened 2020-01-22
PUBDEV-7222: custom_metric_func not supported in GLM

hannah-tillman:


h2o-docs/src/product/data-science/glm.rst
```diff
@@ -163,8 +163,6 @@ Defining a GLM Model
 
 -  **obj_reg**: Specifies the likelihood divider in objective value computation. This defaults to 1/nobs.
 
--  `custom_metric_func <algo-params/custom_metric_func.html>`__: Optionally specify a custom evaluation function.
-
 -  `upload_custom_metric <algo-params/upload_custom_metric.html>`__: Upload a custom metric into a running H2O cluster.
```
angela0xdata: upload_custom_metric is also not supported in GLM.

PR #4247, opened 2020-01-23
PUBDEV-7223: Add an option to specify JDBC URL to get Delegation Token

michalkurka:
User can now specify custom JDBC URL to retrieve Hive Delegation token, eg.:

    h2odriver ... -hiveJdbcUrlPattern "jdbc:hive2://hostname:10000/core;ssl=true;sslTrustStore=/path/to/file.jks;principal={{principal}}" -hivePrincipal "hive/HOST@DOMAIN"

`hiveJdbcUrlPattern` is a pattern where parts {{host}} and {{principal}} can be substituted with values given in parameter `-hiveHost` and `-hivePrincipal`

This substitution is optional, a full URL can be given as well:

    h2odriver ... -hiveJdbcUrlPattern "jdbc:hive2://hostname:10000/core;ssl=true;sslTrustStore=/path/to/file.jks;principal=hive/HOST@DOMAIN" -hivePrincipal "hive/HOST@DOMAIN"

`hivePrincipal` needs to be present on the command line (it is used not just to generate the URL).

### DIFF
--- a/h2o-mapreduce-generic/src/main/java/water/hadoop/HiveTokenGenerator.java
+++ b/h2o-mapreduce-generic/src/main/java/water/hadoop/HiveTokenGenerator.java
@@ -1,7 +1,6 @@
 package water.hadoop;
 
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapreduce.Job;
 import org.apache.hadoop.mapreduce.security.token.delegation.DelegationTokenIdentifier;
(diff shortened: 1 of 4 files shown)


h2o-mapreduce-generic/src/main/java/water/hadoop/HiveTokenGenerator.java
```diff
@@ -24,60 +23,75 @@
...
-      String hiveHost = conf.get(H2O_HIVE_HOST);
+      String hiveJdbcUrl = conf.get(H2O_HIVE_JDBC_URL);
       String hivePrincipal = conf.get(H2O_HIVE_PRINCIPAL);
-      return make(hiveHost, hivePrincipal);
+      return make(hiveJdbcUrl, hivePrincipal);
     }
     
   }
+
+  private static boolean isPresent(String value) {
```
honzasterba: candidate for stringutils or replace with StringUtils.isNullOrEmpty

PR #4251, opened 2020-01-24
HEXDEV-748: Fix replica invalidation

michalkurka:
H2O doesn't expect to see more than 10 "other nodes" after the cluster forms. Other nodes can be clients or anything that is trying to talk to h2o. When this limit is exceeded ArrayIndexOutOfBounds is thrown.

We optimize the fix for the usual scenario when there is just a single client connected to the cluster. We keep the limit of 10 nodes but permit other nodes to connect as well. For these nodes, we are doing speculative key invalidation - we do not keep track which node used which key, we just invalidate on all the extra nodes (regardless if they asked for the key or not).

This whole mechanism will go away with client-less approach.

### DIFF
--- a/h2o-core/src/main/java/water/Value.java
+++ b/h2o-core/src/main/java/water/Value.java
@@ -492,16 +492,16 @@ public final class Value extends Iced implements ForkJoinPool.ManagedBlocker {
     // up, but the replica list does not account for the new replica.  However,
     // the rwlock cannot go down until an ACKACK is received, and the ACK
     // (hence ACKACK) doesn't go out until after this function returns.
-    
-    markHotReplica(h2o._unique_idx);
+    markHotReplica(h2o);
     // Both rwlock taken, and replica count is up now.
     return true;
   }
 
-  private void markHotReplica(short idx) {
+  private void markHotReplica(H2ONode n) {
+    n.markLocalDKVAccess();
     byte[] r = replicas();
-    if (idx < r.length)
-      r[idx] = 1; 
+    if (n._unique_idx < r.length)
+      r[n._unique_idx] = 1; 
   }
 
   /** Atomically lower active GET and Invalidate count */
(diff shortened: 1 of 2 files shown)


jakubhava:
Note: We also need to automatically add the jar to `-libjars` in case of automatic mode of external backend and mention this in the documentation

h2o-core/src/main/java/water/Value.java
```diff
@@ -485,11 +492,18 @@ boolean setReplica( H2ONode h2o ) {
...
-    replicas()[h2o._unique_idx] = 1;
+    markHotReplica(h2o);
     // Both rwlock taken, and replica count is up now.
     return true;
   }
 
+  private void markHotReplica(H2ONode n) {
+    n.markLocalDKVAccess();
+    byte[] r = replicas();
+    if (n._unique_idx < r.length)
```
jakubhava: If I understand correctly, this is the main fix where we don't set replica to active on unseen nodes, is that correct?

h2o-core/src/main/java/water/Value.java
```diff
@@ -498,7 +512,7 @@ void lowerActiveGetCount( H2ONode h2o ) {
       int old = _rwlock.get(); // Read the lock-word
       assert old > 0;      // Since lowering, must be at least 1
       assert old != -1;    // Not write-locked, because we are an active reader
-      assert (h2o==null) || (_replicas!=null && _replicas[h2o._unique_idx]==1); // Self-bit is set
+      assert (h2o==null) || (_replicas!=null && (h2o._unique_idx >= _replicas.length || _replicas[h2o._unique_idx]==1)); // Self-bit is set
```
jakubhava: I'm having hard time understanding this change here `(h2o._unique_idx >= _replicas.length` cold you please explain it a little?

h2o-core/src/main/java/water/Value.java
michalkurka: The fix does 2 things:
- instead of considering `CLOUD.size()` to make the replica array, it uses `H2ONode.IDX.length` - this is the correct way because any node that tries to communicate with H2O during cloud formation will be recorded in H2ONode.IDX array.
- we don't grow the array, instead we assume that in a regular deployment there will be just one client (but we keep buffer for 10), if the assumption is not valid we do not remember the replicas for the rest of the clients. Instead of invalidating per-value, we always invalidate regardless whether the node actually touched the value or not. We just remember whether the node touched the DKV at all instead.

h2o-core/src/main/java/water/Value.java
michalkurka: For "too many" nodes we stop registering replicas. We just invalidate every key if the node ever touched DKV.

PR #4255, opened 2020-01-25
PUBDEV-6917 Introduce HDP 3.2 H2O driver to support Azure Data Lake S…

Pscheidl:
…torage Gen 2 (abfs:// protocol)

No code changes required in H2O itself.

https://0xdata.atlassian.net/browse/PUBDEV-6917

Works with H2O just by supplying the proper driver. Since `hadoop-azure` version `3.2.0`, there is `org.apache.hadoop.fs.azurebfs.AzureBlobFileSystem` present. Which provides out of the box support for Azure `abfs[s]://` protocol. Previous versions support `wasb` only. However, this requires hadoop `3.2` ideally. Doesn't work with versions 2.x, as there are classes missing on the classpath.

Hint: there is a newer version `3.2.1` and this one should probably be used: https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-azure

Below is the configuration used, make sure to replace the `<storage_account_name_here>` with real storage account names. Plus fill the secret shared key. The configuration comes from: https://hadoop.apache.org/docs/stable/hadoop-azure/abfs.html

```xml
<?xml version="1.0"?>
<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
<configuration>
<property>
  <name>fs.azure.account.auth.type.<storage_account_name_here>.dfs.core.windows.net</name>
  <value>SharedKey</value>
  <description>
  </description>
</property>
<property>
  <name>fs.azure.account.key.<storage_account_name_here>.dfs.core.windows.net</name>
<value>INSERT_YOUR_KEY_HERE</value>
  <description>
  The secret password. Never share these.
  </description>
</property>
<property>
    <name>fs.abfs.impl</name>
    <value>org.apache.hadoop.fs.azurebfs.AzureBlobFileSystem</value>
</property>
<property>
    <name>fs.abfss.impl</name>
    <value>org.apache.hadoop.fs.azurebfs.SecureAzureBlobFileSystem</value>
</property>
</configuration>
```

As this requires `hadoop 3.2`, this PR aims to provide H2O driver support for this new Hadoop.  Maybe the driver will work with previous Hadoop versions as well, but that is not guaranteed. You might give it a try @JHerna17 

Screenshot from 2020-01-25 10-58-52

## Hadoop smoke pipeline
http://mr-0xc1:8080/view/H2O-3/job/h2o-3-hadoop-smoke-pipeline/job/pavel%252Fpubdev-6917/

### DIFF
--- a/h2o-hadoop-3/h2o-hdp3.2-assembly/build.gradle
+++ b/h2o-hadoop-3/h2o-hdp3.2-assembly/build.gradle
@@ -1,0 +1,15 @@
+ext {
+  hadoopVersion = 'hdp3.2'
+  hadoopMavenArtifactVersion = '3.2.1'
+  orcSupported = true
+  orcHiveExecVersion = "3.1.2"
+}
+
+apply from: '../assemblyjar.gradle'
+
+dependencies {
+  compile (project(":h2o-jetty-9")) {
+    exclude module: "servlet-api"
+    exclude group: "javax.servlet", module: "javax.servlet-api"
+  }
+}
--- a/docker/hadoop/hdp/scripts/sbin/add_hdp_repo.sh
+++ b/docker/hadoop/hdp/scripts/sbin/add_hdp_repo.sh
@@ -15,6 +15,11 @@ function print_green {
 }
 
 case "${1}" in
+  3.2)
+    major_version="3"
+    hdp_version="3.2.0.0"
+    ubuntu_repo_version="14"
+    ;;
   3.1)
     major_version="3"
     hdp_version="3.1.0.0"
--- a/h2o-dist/buildinfo.json
+++ b/h2o-dist/buildinfo.json
@@ -134,6 +134,11 @@
             "zip_file_path" : "h2o-SUBST_PROJECT_VERSION-hdp3.1.zip"
         },
         {
+            "distribution" : "hdp3.2",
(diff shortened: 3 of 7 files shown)


h2o-hadoop-3/h2o-hdp3.2-assembly/build.gradle
```diff
@@ -0,0 +1,15 @@
+ext {
+  hadoopVersion = 'hdp3.2'
+  hadoopMavenArtifactVersion = '3.2.1'
+  orcSupported = true
```
Pscheidl: False ? @michalkurka

Pscheidl:
Looks like hortonWorks doesn't have 3.2 version out, and we need it.
https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-common?repo=hortonworks-releases

http://mr-0xc1:8080/blue/organizations/jenkins/docker-images%2Fh2o-3-hadoop-docker-build/detail/h2o-3-hadoop-docker-build/62/pipeline/90/

michalkurka:
@Pscheidl this is okay, but to resolve the original jira please add to a different PR:

- test demonstrating the support
- work with @angela0xdata on documenting how to enable ABFS(S) with H2O

PR #4258, opened 2020-01-27
PUBDEV-7227: unsupported Python bytecode instructions when calling `apply` on H2OFrame

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-7227

`lambda_to_expr` implementation doesn't support some bytecode instructions introduced in Python 3.7.

Also added support for some instructions introduced with 3.6, especially `CALL_FUNCTION_EX`, even if they don't seem that useful in the context of `H2OFrame.apply`, but the greater support, the better.

Note that our jenkins pipeline is still not running tests using Python 3.7, so this will be officially supported only once https://github.com/h2oai/h2o-3/pull/3560 is fixed.

### DIFF
--- a/h2o-py/h2o/astfun.py
+++ b/h2o-py/h2o/astfun.py
@@ -39,6 +39,8 @@ BYTECODE_INSTRS = {
     # bject to call is on the stack. Pops all function arguments, and the function
     # itself off the stack, and pushes the return value.
     "CALL_FUNCTION": "",
+    "CALL_FUNCTION_VAR": "",  # Py <= 3.5
+    "CALL_FUNCTION_VAR_KW": "",  # Py <= 3.5
     # from https://docs.python.org/3/library/dis.html#opcode-CALL_FUNCTION_KW :
     # Calls a function. argc indicates the number of arguments (positional and keyword).
     # The top element on the stack contains a tuple of keyword argument names.
@@ -60,7 +62,7 @@ BYTECODE_INSTRS = {
     # CALL_FUNCTION_EX pops all arguments and the callable object off the stack,
     # calls the callable object with those arguments,
     # and pushes the return value returned by the callable object.
-    "CALL_FUNCTION_EX": "",
+    "CALL_FUNCTION_EX": "",  # Py >= 3.6
     # from https://docs.python.org/3/library/dis.html#opcode-CALL_METHOD :
     # Calls a method. argc is the number of positional arguments.
     # Keyword arguments are not supported. This opcode is designed to be used with LOAD_METHOD.
@@ -68,7 +70,7 @@ BYTECODE_INSTRS = {
     # Below them, the two items described in LOAD_METHOD are on the stack
     # (either self and an unbound method object or NULL and an arbitrary callable).
     # All of them are popped and the return value is pushed.
-    "CALL_METHOD": "",
+    "CALL_METHOD": "",  # Py >= 3.7
 }
 
 
@@ -90,10 +92,16 @@ def is_func(instr):
 def is_func_kw(instr):
     return "CALL_FUNCTION_KW" == instr
 
-def is_func_ex(instr):  # since 3.6
+def is_func_ex(instr):  # Py >= 3.6
     return "CALL_FUNCTION_EX" == instr
 
-def is_method_call(instr):
+def is_func_var(instr): # Py <= 3.5
+    return "CALL_FUNCTION_VAR" == instr
+
+def is_func_var_kw(instr): # Py <= 3.5
+    return "CALL_FUNCTION_VAR_KW" == instr
+
+def is_method_call(instr): # Py >= 3.7
     return "CALL_METHOD" == instr
 
 def is_callable(instr):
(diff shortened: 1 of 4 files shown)


h2o-py/h2o/astfun.py
```diff
@@ -7,15 +7,15 @@
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from opcode import *  # an undocumented builtin module
+import dis
```
sebhrusen: using "public" `dis` module here to avoid depending on non-official modules that may not be available if not using `Cython` implementation.

h2o-py/tests/testdir_misc/pyunit_apply.py
```diff
@@ -114,6 +115,37 @@ def pyunit_apply_n_to_n_ops():
...
+        scale_with_kwarg=lambda x: x.scale(center=False, scale=False),
+        scale_with_argkwarg=lambda x: x.scale(False, scale=False),
+        scale_with_global_arg=(lambda x: x.scale(false, scale=false)) if not PY2 else None,
+        scale_with_args=(lambda x: x.scale(*args)) if not PY2 else None,
+        scale_with_kwargs=(lambda x: x.scale(**kwargs)) if not PY2 else None,
+        scale_with_partial_args=(lambda x: x.scale(False, *partial_args)) if not PY2 else None,
+        scale_with_partial_kwargs=(lambda x: x.scale(False, **partial_kwargs)) if not PY2 else None,
+        scale_with_args_and_kwargs=(lambda x: x.scale(*partial_args, **partial_kwargs)) if not PY2 else None,
+        scale_with_all_kind_args=(lambda x: x.scale(False, *partial_args, **partial_kwargs)) if not PY2 else None,  # strangely this works because our signature verification is not that strict, but it's fine... at least behaves as expected.
+    )
```
st-pasha: It seems like all these tests are for the `CALL_METHOD` op. Can we also have tests for `CALL_FUNCTION_EX` (i.e. call a regular function, not a method).

st-pasha: Also, in Py3.9 there will be added a new C function for calling python functions without any arguments. I don't know whether they'd be adding a new opcode for this, but it's possible. Thus, we should also add a test case for calling a function/method with no arguments.

sebhrusen: actually depending on Python version, those tests will use either `CALL_FUNCTION`, `CALL_METHOD`, `CALL_FUNCTION_KW` or `CALL_FUNCTION_EX`...
from 3.6, it's using `CALL_FUNCTION_EX` as soon as there is some unpacking needed (*args or **kwargs). 
Actually testing with `3.5`, not sure all of those are supported there.

sebhrusen: for the call without args, there are other tests in this file

sebhrusen:
@st-pasha the new unpacking support now also works on older Py version, including `2.7` to `3.5` so that I could get a green build.
Any additional comments on your side?

michalkurka:
Does this mean we can also test for 3.8 and 3.9 (just released)?

sebhrusen:
@michalkurka you can only test with those versions locally for now; for jenkins, we still need new images: https://github.com/h2oai/h2o-3/pull/3560
the only problem with testing with Py 3.7+ in the ticket above is due to incompatibilities with some of our test dependencies: h2o-py itself works fine.
Btw, `3.9` is just in alpha currently, and it as some new bytecode instructions to extend lists and dicts: didn't try it yet, but I hope it's not using those instructions to build the function params...

sebhrusen:
@michalkurka ok to merge this? 
FYI, after a complete refresh of my Py setup today, I ran the `pyunit_apply.py` test against following versions:
`2.7.17`, `3.5.9`,  `3.6.10`, `3.7.6`, `3.8.1`.
I couldn't run it against `3.9-dev`: too early in development stage I believe as I couldn't install some of the dependencies right now.

h2o-py/h2o/expr.py
```diff
@@ -149,7 +150,12 @@ def _get_ast_str(self, top):
             return self._cache._id  # Data already computed under ID, but not cached
         assert isinstance(self._children,tuple)
         exec_str = "({} {})".format(self._op, " ".join([ExprNode._arg_to_expr(ast) for ast in self._children]))
-        gc_ref_cnt = len(gc.get_referrers(self))
+        # explicit call to gc.collect as recommended before calling gc.get_referrers to obtain only live referrers;
+        # otherwise, this creates subtle bugs for Py < 3.7 as we have logic below based on the ref count.
+        # Also filtering out the frames for the referrers (they appear until 3.6) and change the behaviour depending on the execution path.
+        gc.collect()
+        referrers = gc.get_referrers(self)
+        gc_ref_cnt = len([r for r in referrers if not inspect.isframe(r)])
```
sebhrusen: the logic below relying on gc_ref_cnt is a complete nightmare: different versions of Py don't behave the same regarding `gc`, causing very annoying bugs when switching versions.
I added the `gc.collect()` that optimizes this for Py 3.7+.
Also filtering out calling frames, as they're kept in Py 3.6-, leading to creating of  duplicate tmp frames below.

Need more work (different task) to see if we could completely remove this `MAGIC_REF_COUNT` limit.

michalkurka:
@sebhrusen thank you!

PR #4262, opened 2020-01-28
[PUBDEV-7248] Remove UDPClientEvent from the codebase

jakubhava:


h2o-core/src/main/java/water/Paxos.java
```diff
@@ -71,23 +71,7 @@ static synchronized int doHeartbeat( H2ONode h2o ) {
...
-      if (!H2O.ARGS.client && h2o._heartbeat._client && !H2O.isNodeInFlatfile(h2o)) {
-        // A new client was reported to this node so we propagate this information to all nodes in the cluster, to this
-        // as well
-        UDPClientEvent.ClientEvent.Type.CONNECT.broadcast(h2o);
-      } else if (H2O.ARGS.client && !H2O.isNodeInFlatfile(h2o)) {
-        // This node is a client and using a flatfile to figure out a topology of the cluster. The flatfile passed to the
-        // client is always modified at the start of H2O to contain only a single node. This node is used to propagate
-        // information about the client to the cluster. Once the nodes have the information about the client, then propagate
-        // themselves via heartbeat to the client
-        H2O.addNodeToFlatfile(h2o);
```
jakubhava: Client's flatfile contains all the worker nodes at the time Sparkling Water is started

h2o-core/src/main/java/water/Paxos.java
```diff
@@ -71,23 +71,7 @@ static synchronized int doHeartbeat( H2ONode h2o ) {
       ListenerService.getInstance().report("different_cloud", h2o);
       return 0;
     }
-
-
-    // Update manual flatfile in case of flatfile is enabled
-    if (H2O.isFlatfileEnabled()) {
-      if (!H2O.ARGS.client && h2o._heartbeat._client && !H2O.isNodeInFlatfile(h2o)) {
-        // A new client was reported to this node so we propagate this information to all nodes in the cluster, to this
```
jakubhava: Since Client's flatfile contains all the worker nodes we can be sure client will start pinging all of the worker nodes

michalkurka:
client tests failed though :(

jakubhava:
It might be because in the client tests the client does not have full flatfile?

Can I modify them to simulate how Sparkling Water works? With full flatfile?

michalkurka:
IMHO the client should have the full flatfile: https://github.com/h2oai/h2o-3/blob/master/scripts/run.py#L587

jakubhava:
Oki oki, let me have a look on the issue then

jakubhava:
Pushed fixed, I removed a little bit too much :)

We still need to add the client to the worker's node flatfile which was previously handled after in the handling mathod of the broadcast call

PR #4264, opened 2020-01-28
PUBDEV-6818: Fix Unsupported Exception

jakubhava:
Fixes 

```
Exception in thread "TCP-SMALL-SEND-/192.168.3.113:54323" java.lang.UnsupportedOperationException
	at java.nio.ByteBuffer.array(ByteBuffer.java:994)
	at water.H2ONode$SmallMessagesSendThread.run(H2ONode.java:590)
```
From java documentation:
 > UnsupportedOperationException - If this buffer is not backed by an accessible array

Still trying to find the cause

### DIFF
--- a/h2o-core/src/main/java/water/AutoBuffer.java
+++ b/h2o-core/src/main/java/water/AutoBuffer.java
@@ -421,7 +421,7 @@ public final class AutoBuffer {
           // For small-packet write, send via UDP.  Since nothing is sent until
           // now, this close() call trivially orders - since the reader will not
           // even start (much less close()) until this packet is sent.
-          if( _bb.limit() < MTU) return udpSend();
+          if( _bb.position() < MTU) return udpSend();
           // oops - Big Write, switch to TCP and finish out there
         }
       }
@@ -1359,7 +1359,7 @@ public final class AutoBuffer {
 
     put1(8);                    // Ship as full longs
     int sofar = x;
-    if ((y-sofar)*8 > _bb.remaining()) expandByteBuffer(ary.length*8);
+    if ((y-sofar)*8 > _bb.remaining()) expandByteBuffer(nzlen*8);
     while( sofar < y ) {
       LongBuffer lb = _bb.asLongBuffer();
       int len = Math.min(y - sofar, lb.remaining());
--- a/h2o-core/src/main/java/water/AutoBuffer.java
+++ b/h2o-core/src/main/java/water/AutoBuffer.java
@@ -421,7 +421,7 @@ public final class AutoBuffer {
           // For small-packet write, send via UDP.  Since nothing is sent until
           // now, this close() call trivially orders - since the reader will not
           // even start (much less close()) until this packet is sent.
-          if( _bb.position() < MTU) return udpSend();
+          if( _bb.limit() < MTU) return udpSend();
           // oops - Big Write, switch to TCP and finish out there
         }
       }
--- a/h2o-core/src/main/java/water/TimeLine.java
+++ b/h2o-core/src/main/java/water/TimeLine.java
@@ -184,7 +184,7 @@ public class TimeLine extends UDP {
   // possible to the same point in time.
   static long[][] SNAPSHOT;
   static long TIME_LAST_SNAPSHOT = 1;
-  static private H2O CLOUD;      // Cloud instance being snapshotted
+  static private H2O CLOUD = H2O.CLOUD;      // Cloud instance being snapshotted
   public static H2O getCLOUD(){return CLOUD;}
   static public long[][] system_snapshot() {
     // Now spin-wait until we see all snapshots check in.
--- a/h2o-core/src/main/java/water/TimeLine.java
+++ b/h2o-core/src/main/java/water/TimeLine.java
@@ -184,7 +184,7 @@ public class TimeLine extends UDP {
   // possible to the same point in time.
   static long[][] SNAPSHOT;
   static long TIME_LAST_SNAPSHOT = 1;
-  static private H2O CLOUD = H2O.CLOUD;      // Cloud instance being snapshotted
+  static private H2O CLOUD;      // Cloud instance being snapshotted
   public static H2O getCLOUD(){return CLOUD;}
   static public long[][] system_snapshot() {
     // Now spin-wait until we see all snapshots check in.
@@ -221,6 +221,9 @@ public class TimeLine extends UDP {
 
   // Send our most recent timeline to the remote via TCP
   @Override AutoBuffer call( AutoBuffer ab ) {
+    if (CLOUD == null) {
+      return null;
+    }
     long[] a = snapshot();
     if( ab._h2o == H2O.SELF ) {
       synchronized(TimeLine.class) {
@@ -236,6 +239,9 @@ public class TimeLine extends UDP {
 
   // Receive a remote timeline
   static void tcp_call( final AutoBuffer ab ) {
+    if (CLOUD == null) {
+      return;
+    }
     ab.getPort();
     long[] snap = ab.getA8();
     int idx = CLOUD.nidx(ab._h2o);
--- a/h2o-core/src/test/java/water/AutoBufferTest.java
+++ b/h2o-core/src/test/java/water/AutoBufferTest.java
@@ -6,6 +6,7 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 
 import java.io.Serializable;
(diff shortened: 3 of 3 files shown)


h2o-core/src/main/java/water/H2ONode.java
```diff
@@ -581,13 +581,20 @@ private boolean keepSending() {
               _outgoingMsgQ.put(bb); // put back and give someone else a chance to deliver
               break; // terminate
             }
-            while( bb != null ) {         // while have an BB to process
-              assert !bb.isDirect() : "Direct BBs already got recycled";
+            while( bb != null) {         // while have an BB to process
```
honzasterba: this is seems to be a whitespace-only change and its not improving anything (talking about the while only)

h2o-core/src/main/java/water/H2ONode.java
```diff
@@ -581,13 +581,20 @@ private boolean keepSending() {
               _outgoingMsgQ.put(bb); // put back and give someone else a chance to deliver
               break; // terminate
             }
-            while( bb != null ) {         // while have an BB to process
-              assert !bb.isDirect() : "Direct BBs already got recycled";
```
honzasterba: this does not seem like the right thing to do, according to the assert this case should have been handled above. did some other modification break this assert?

michalkurka: To me this means we cannot work with direct BBs here because at the point we get here, they are already returned to the object pool in AutoBuffer because this is an asynchronous process.

michalkurka:
It seems to me the cause is that the TimeLine snapshot can become too big to be sent through the small message channel.

h2o-core/src/main/java/water/H2ONode.java
jakubhava: Yup, thx

h2o-core/src/main/java/water/H2ONode.java
jakubhava: I have actually tried to put Runtime Exception to be thrown if the `bb` is direct and it was thrown

michalkurka: yep, that is right, a direct byte buffer can definitely find its way in - the behavior you are describing is permissible

What I am saying is that the message of the assert means something like: "we cannot work with direct BBs here because at the point we get here, they are already returned to the object pool in AutoBuffer because this is an asynchronous process"

jakubhava: I see, makes sense

jakubhava:
Was debugging a bit more, discovered that in `close` method in `AutoBuffer`, condition `if( _bb.position() < MTU) {` always holds, but `_bb.limit() < MTU` does not.

Most of the messages are small and have limit of 1024 < 1500(MTU), but some messages are big and exceeds the MTU, but because of the condition, these messages are still send as small ones.

Also the messages which are "big" are always backed by direct byte buffers

jakubhava:
@michalkurka please have a look, the change to `_bb.limit` is actually solving the issue, so far verified locally, waiting for the full test suite

michalkurka:
@jakubhava code change looks good

We should keep an eye on the performance impacts. Do you think it can have any? Can there be cases when "larger" messages were sent through the small messages channel without an error? If so this could definitely affect performance.

jakubhava:
@michalkurka regarding performance -> I believe that everytime larger message got it in, it was backed by the direct byte buffer, which was already pretty rare situation. From my observation all the messages were actually small except the outlier

jakubhava:
@honzasterba I will try to find a way how to test it, thanks! But not sure if I want write unit-tests for all AutoBuffer methods(currently Autobuffer has just 2 simple tests

h2o-core/src/main/java/water/AutoBuffer.java
```diff
@@ -421,7 +421,7 @@ public final int close() {
           // For small-packet write, send via UDP.  Since nothing is sent until
           // now, this close() call trivially orders - since the reader will not
           // even start (much less close()) until this packet is sent.
-          if( _bb.position() < MTU) return udpSend();
+          if( _bb.limit() < MTU) return udpSend();
```
michalkurka: @jakubhava please see the implementation of `udpSend`:
```
  private int udpSend() throws IOException {
    assert _chan == null;
    TimeLine.record_send(this,false);
    _size = _bb.position();
    assert _size < AutoBuffer.BBP_SML._size;
    _bb.flip();                 // Flip for sending
```

the assumption here is that `size == position`, also `bb.flip` does set `limit = position` - IMHO this is therefore not correct way to fix this

jakubhava: We can go to `!__bb.isDirect() && _bb.position() < MTU` I was thinking about it and I thin it actually descrives the invariant more explicitly, what do you think?

michalkurka: that might be better, but the invariant that `_bb.position == size` is still used later 

and we really shouldn't have a Direct BB that is small

there must be a bug elsewhere,

didn't you say that the size of the data is 65k? it is suspiciously close to

static BBPool BBP_BIG = new BBPool(64*1024);

h2o-core/src/main/java/water/AutoBuffer.java
```diff
@@ -1359,7 +1359,7 @@ public AutoBuffer putA8( long[] ary ) {
 
     put1(8);                    // Ship as full longs
     int sofar = x;
-    if ((y-sofar)*8 > _bb.remaining()) expandByteBuffer(ary.length*8);
+    if ((y-sofar)*8 > _bb.remaining()) expandByteBuffer(nzlen*8);
```
jakubhava: @michalkurka method `putA8` tries to optimize the communication and trims the input array by leading and trailing zeros. New array size is `nzlen` and we should expand the byte buffer according to that

jakubhava:
This fixes the specific issue we were hitting, but I still think I am still missing something as I see some different errors after this change,

such as:

```
java.lang.ArrayIndexOutOfBoundsException: -1
	at water.H2O.submitTask(H2O.java:1328)
	at water.TCPReceiverThread.basic_packet_handling(TCPReceiverThread.java:348)
	at water.TCPReceiverThread$SmallMessagesReaderThread.run(TCPReceiverThread.java:257)
```

or

```
IO error
	at water.TimeLine.tcp_call(TimeLine.java:258)
	at water.TCPReceiverThread$TCPReaderThread.run(TCPReceiverThread.java:171)
```

jakubhava:
Oki, the NPE is fixed. It happened now for the first time as the Timeline message was successfully sent because of the first fix (changing size to which bytebuffer should expand to trimmed array).

The NPE happened because in Timeline the CLOUD is initialized to `null` and is set to non-null value during a call `system_snapshot`. But if this method hasn't been  called yet and the node already received a timeline message -> NPE

h2o-core/src/main/java/water/AutoBuffer.java
michalkurka: @jakubhava great investigation, thank you!

h2o-core/src/main/java/water/TimeLine.java
```diff
@@ -184,7 +184,7 @@ public static InetAddress inet( long[] tl, int idx ) {
   // possible to the same point in time.
   static long[][] SNAPSHOT;
   static long TIME_LAST_SNAPSHOT = 1;
-  static private H2O CLOUD;      // Cloud instance being snapshotted
+  static private H2O CLOUD = H2O.CLOUD;      // Cloud instance being snapshotted
```
michalkurka: Is this needed just in tests or also for regular execution?

michalkurka: See this code
```
        for( int i=0; i<CLOUD._memary.length; i++ )
          if( CLOUD._memary[i]==H2O.SELF )
            SNAPSHOT[i] = a;
        TimeLine.class.notify();
```

the code works only because the initial CLOUD doesn't have any members, otherwise, there would be NPE because snapshot is not initialized. This means the code will not actually doing anything.

Both `call` and `tcp_call` methods don't seem to do anything in that case. I think this should be made more obvious by handling CLOUD != null in those methods and returning without action.

@jakubhava @honzasterba WDYT?

jakubhava: @michalkurka It happens also in normal exeution. Agree ^ Will have a look

michalkurka:
Thank you @jakubhava

PR #4267, opened 2020-01-28
[PUBDEV-7246] make EasyPredictModelWrapper support xgboost with offset

honzasterba:
this is a quick and dirty solution and I dont like it, WDYT?

### DIFF
--- a/h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostMojoModel.java
+++ b/h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostMojoModel.java
@@ -70,7 +70,7 @@ public abstract class XGBoostMojoModel extends MojoModel implements SharedTreeGr
   public void postReadInit() {}
 
   @Override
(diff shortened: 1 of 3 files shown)


honzasterba:
@michalkurka I think an ideal solution was to add a similar flag to all other offse-enabled MOJOs and make the behaviour consistent, that would mean bumping their version because the behaviour for new models would be different than for old ones.

honzasterba:
I have refactored to a 3-state enum to better describe whats going on, we can postopne modifying ohter MOJOs for later

michalkurka:
@honzasterba, to be honest, I think the original solution was better (clearer) - I am not sure what we will do with this new infrastructure code in the future

XGBoost is the only algorithm that really does _require offset_ when the model was trained with offset - because the behavior is actually different if the offset is given or not (predictions for calling score0 with offset=0 and without the offset parameter are different). No other algos behave like that (and it really doesn't make sense why they would). This is why we built-in the protection in the score0 logic.

For everything else, score0(x, offset=0) and score0(x) is the same thing.

This is why I like the previous solution better, it correctly captures that XGBoost is the only algo that is different. WDYT?

PR #4272, opened 2020-01-29
AutoML: simplifying logic for default stopping tolerance

sebhrusen:
Existing logic was confusing, splitting the logic in 2 different places, and giving the impression that default `stopping_tolerance` was `0.001`, when it is actually is an adaptative value, computed from some properties of the training frame.

### DIFF
--- a/h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
+++ b/h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
@@ -49,6 +49,26 @@ def test_invalid_project_name():
         assert "1nvalid" in str(e)
 
 
+def test_early_stopping_defaults():
+    print("Check default early stopping params")
+    ds = import_dataset()
+    aml = H2OAutoML(project_name="py_aml_early_stopping_defaults", max_models=max_models)
+    aml.train(y=ds['target'], training_frame=ds['train'])
+    stopping_criteria = aml._build_resp['build_control']['stopping_criteria']
+    print(stopping_criteria)
+
+    from math import sqrt
+    auto_stopping_tolerance = (lambda fr: min(0.05, max(0.001, 1/sqrt((1 - sum(fr.nacnt()) / (fr.ncols * fr.nrows)) * fr.nrows))))(ds['train'])
+
+    assert stopping_criteria['stopping_rounds'] == 3
+    assert stopping_criteria['stopping_tolerance'] == auto_stopping_tolerance
+    assert stopping_criteria['stopping_metric'] == 'AUTO'
+    assert stopping_criteria['max_models'] == max_models
+    assert stopping_criteria['max_runtime_secs'] == 0
+    assert stopping_criteria['max_runtime_secs_per_model'] == 0
+
+
+
 def test_early_stopping_args():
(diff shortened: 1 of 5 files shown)


h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
```diff
@@ -312,8 +332,8 @@ def test_no_time_limit_if_max_models_is_provided():
     aml.train(y=ds['target'], training_frame=ds['train'])
     max_runtime = aml._build_resp['build_control']['stopping_criteria']['max_runtime_secs']
     max_models = aml._build_resp['build_control']['stopping_criteria']['max_models']
-    assert max_runtime == 0
-    assert max_models == 1
+    assert max_models == 1, max_models
```
deil87: @sebhrusen have not seen this type of syntax. Could you please share with me what does it mean when we add comma after comparison?

sebhrusen: it's just the message printed on failed assertion

sebhrusen: https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

h2o-r/h2o-package/R/automl.R
```diff
@@ -317,6 +317,7 @@ h2o.automl <- function(x, y, training_frame,
   # GET AutoML object
   aml <- h2o.get_automl(project_name = res$job$dest$name)
   attr(aml, "id") <- res$job$dest$name
+  attr(aml, '_build_json') <- res # hidden attribute for debugging/testing
```
deil87: @sebhrusen Just trying to understand couple of things: 
1) so we did not have _build_resp analog in R? In Py do we use `_build_resp` for non testing purposes? 
2) why it is called `_build_json` here. _build_resp is derived from the same json but parsed into object and in case of R we don't have it parsed, right?  
3) Are we extending API just for testing purposes?

sebhrusen: in Python, variables prefixed with `_` are not considered as API: it's a naming convention to say they are "private" and using them in consumer code is not safe.
in R, it's a bit similar with attributes, they're not considered as API (they don't appear in class description), but rather some kind of metadata.
I'll rename to `_build_resp` for consistency with Py

h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
deil87: @sebhrusen thanks, this was my number one guess, just usually Failing messages contain some text for context.

h2o-r/h2o-package/R/automl.R
deil87: @sebhrusen good then, thanks for the info!

h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
sebhrusen: Python error also prints the code line raising the assertion, so it's not as necessary as in Java to be verbose in description

PR #4273, opened 2020-01-29
PUBDEV-6965: adding new ensemble metalearners

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-6965

Internal refactorings (moved code had only `package` visibility) to allow discovery of metalearners from extensions.
`XGBoost` metalearner is defined in its extension module.
`NaiveBayes` and `PSVM` are using defaults.


h2o-py/h2o/h2o.py
```diff
@@ -993,25 +991,12 @@ def get_model(model_id):
...
-    elif algo == "xgboost":      m = H2OXGBoostEstimator()
-    elif algo == "word2vec":     m = H2OWord2vecEstimator()
-    elif algo == "generic": m = H2OGenericEstimator()
-    elif algo == "deeplearning":
-        if model_json["output"]["model_category"] == "AutoEncoder":
-            m = H2OAutoEncoderEstimator()
-        else:
-            m = H2ODeepLearningEstimator()
-    elif algo == "stackedensemble": m = H2OStackedEnsembleEstimator()
-    elif algo == "isolationforest": m = H2OIsolationForestEstimator()
```
sebhrusen: this list was not really up-to-date (missing `psvm`, `coxph`, maybe others) + created a hard dependency on classes from submodule.
Replaced by auto-registration of estimators.

sebhrusen:
@ledell @michalkurka : this one has been waiting for a while now

h2o-py/h2o/h2o.py
sebhrusen: if you want me to move those changes in another branch, it's easy: just did it here because some new tests were failing.

angela0xdata:
@sebhrusen FYI, I made some minor updates in this PR to the User Guide to reflect the new metalearner_algorithm option available as a result of this change.

sebhrusen:
@angela0xdata perfect, thanks

h2o-py/h2o/h2o.py
michalkurka: LGTM, I don't see an issue - thank you

h2o-algos/src/main/java/hex/ensemble/MetalearnerProvider.java
```diff
@@ -0,0 +1,12 @@
+package hex.ensemble;
+
+import hex.Model.Parameters;
+
+public interface MetalearnerProvider<M extends Metalearner, P extends Parameters> {
+
+    String getName();
+
+    P newParameters();
```
michalkurka: You are using `ModelBuilder.make` to create the metalearner instance (eg. GLM), this method also automatically supplies a new (and modifiable) instance of the default parameter object. Can we simplify the API and remove the need to have `newParameters`?

sebhrusen: @michalkurka this would not work for "AUTO" for example or any additional metalearner that would not be named exactly like its algo: we can imagine having more than one metalearner per algo (with different default params).

This `newParameters` method is called by `StackedEnsembleParameters`, so at a time where we don't havecreated the metalearner instance yet...

I could also suggest to keep the `newParameters` but make it a default interface method:
```
default P newParameters() {
   return newInstance().createBuilder()._parms;
}
```
not sure it is very efficient, nor elegant to have this in an interface though.
wdyt?

michalkurka: > @michalkurka this would not work for "AUTO" for example or any additional metalearner that would not be named exactly like its algo: we can imagine having more than one metalearner per algo (with different default params).

My thought was to keep the `newInstance` method (for "AUTO"...) but remove `newParameters`, `newInstance` already provides the parameters. In the returned metalearner.

> This newParameters method is called by StackedEnsembleParameters, so at a time where we don't have created the metalearner instance yet...

https://github.com/h2oai/h2o-3/pull/4273/files#diff-91d255cd89e95ed4a240332df7d9002bR82

IMHO a special case, you could address by calling `newInstance().createBuilder()._parms` there.

I am also fine with keeping as-is.

sebhrusen: sold! less is more..

sebhrusen: @michalkurka  waiting for the build results and then merging

sebhrusen:
Build status 99% green (1 XGB test failure) -> merging

ledell:
After finding a [bug with PSVM](https://0xdata.atlassian.net/browse/PUBDEV-7314) we removed "psvm" as a metalearner in a follow-up PR: https://github.com/h2oai/h2o-3/pull/4341

PR #4276, opened 2020-01-30
PUBDEV-7189: ensure that AutoML works even if XGBoost extension is missing on the classpath

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-7189

fully decoupled algorithms implementation logic from AutoML classes, especially for XGBoost so that AutoML will work even if XGB extension is not in the classpath.

Tested locally: no proper way to test this using UTs. Ideally, would need a separate `xgboost-automl` module to ensure this: not sure it's worth the additional complexity for a single class though...

### DIFF
--- a/h2o-automl/build.gradle
+++ b/h2o-automl/build.gradle
@@ -4,12 +4,13 @@ dependencies {
   compile project(":h2o-genmodel")
   compile project(":h2o-core")
   compile project(":h2o-algos")
-  compile project(":h2o-ext-xgboost")
+  compileOnly project(":h2o-ext-xgboost")
 
   // Test dependencies only
   testCompile "junit:junit:${junitVersion}"
   testCompile "com.github.stefanbirkner:system-rules:1.18.0"
   testCompile project(path: ":h2o-core", configuration: "testArchives")
+  testCompile project(":h2o-ext-xgboost")
   testRuntimeOnly project(":${defaultWebserverModule}")
 }
 
--- a/h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
@@ -1,0 +1,188 @@
+package ai.h2o.automl.modeling;
+
+import ai.h2o.automl.Algo;
+import ai.h2o.automl.AutoML;
+import ai.h2o.automl.ModelingStep;
+import ai.h2o.automl.ModelingSteps;
+import hex.genmodel.utils.DistributionFamily;
+import hex.grid.Grid;
+import hex.tree.xgboost.XGBoostModel;
+import hex.tree.xgboost.XGBoostModel.XGBoostParameters;
+import water.Job;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static ai.h2o.automl.ModelingStep.GridStep.DEFAULT_GRID_TRAINING_WEIGHT;
+import static ai.h2o.automl.ModelingStep.ModelStep.DEFAULT_MODEL_TRAINING_WEIGHT;
+
+public class XGBoostSteps extends ModelingSteps {
+
+    static XGBoostParameters prepareModelParameters(AutoML aml, boolean emulateLightGBM) {
+        XGBoostParameters xgBoostParameters = new XGBoostParameters();
+
+        if (emulateLightGBM) {
+            xgBoostParameters._tree_method = XGBoostParameters.TreeMethod.hist;
+            xgBoostParameters._grow_policy = XGBoostParameters.GrowPolicy.lossguide;
+        }
+
+        // setDistribution: no way to identify gaussian, poisson, laplace? using descriptive statistics?
+        xgBoostParameters._distribution = aml.getResponseColumn().isBinary() && !(aml.getResponseColumn().isNumeric()) ? DistributionFamily.bernoulli
+                : aml.getResponseColumn().isCategorical() ? DistributionFamily.multinomial
+                : DistributionFamily.AUTO;
+
+        xgBoostParameters._score_tree_interval = 5;
+        xgBoostParameters._stopping_rounds = 5;
+
+        xgBoostParameters._ntrees = 10000;
+        xgBoostParameters._learn_rate = 0.05;
+//            xgBoostParameters._min_split_improvement = 0.01f;
+
+        return xgBoostParameters;
+    }
+
+    static abstract class XGBoostModelStep extends ModelingStep.ModelStep<XGBoostModel> {
+
+        boolean _emulateLightGBM;
+
+        XGBoostModelStep(String id, int weight, AutoML autoML, boolean emulateLightGBM) {
+            super(Algo.XGBoost, id, weight, autoML);
+            _emulateLightGBM = emulateLightGBM;
+        }
+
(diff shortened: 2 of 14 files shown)


h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostStepsProvider.java
```diff
@@ -193,7 +18,19 @@ public String getName() {
...
     @Override
     public XGBoostSteps newInstance(AutoML aml) {
-        return new XGBoostSteps(aml);
+        return Algo.XGBoost.enabled() ? new XGBoostSteps(aml) : null;
+    }
+
+    @Override
+    public Model.Parameters newDefaultParameters() {
+        if (Algo.XGBoost.enabled()) {
+            try {
```
michalkurka: There is a generic way of constructing new ModelBuilder instance and parameters, you could do 

```
   ModelBuilder.make("xgboost", null, null)._parms;
```

h2o-automl/build.gradle
```diff
@@ -4,12 +4,13 @@ dependencies {
   compile project(":h2o-genmodel")
   compile project(":h2o-core")
   compile project(":h2o-algos")
-  compile project(":h2o-ext-xgboost")
+  compileOnly project(":h2o-ext-xgboost")
```
michalkurka: perfect, this will help some deployments @navdeep-G

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostStepsProvider.java
sebhrusen: Is it worth it? Creating a builder instance just to get the params?

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
```diff
@@ -0,0 +1,188 @@
...
+                        searchParams.put("_min_rows", new Double[]{0.01, 0.1, 1.0, 3.0, 5.0, 10.0, 15.0, 20.0});  // = _min_child_weight
+                    }
+
+                    searchParams.put("_sample_rate", new Double[]{0.6, 0.8, 1.0}); // = _subsample
+                    searchParams.put("_col_sample_rate" , new Double[]{ 0.6, 0.8, 1.0}); // = _colsample_bylevel"
+                    searchParams.put("_col_sample_rate_per_tree", new Double[]{ 0.7, 0.8, 0.9, 1.0}); // = _colsample_bytree: start higher to always use at least about 40% of columns
+//                    searchParams.put("_learn_rate", new Double[]{0.01, 0.05, 0.08, 0.1, 0.15, 0.2, 0.3, 0.5, 0.8, 1.0}); // = _eta
+//                    searchParams.put("_min_split_improvement", new Float[]{0.01f, 0.05f, 0.1f, 0.5f, 1f, 5f, 10f, 50f}); // = _gamma
+//                    searchParams.put("_tree_method", new XGBoostParameters.TreeMethod[]{XGBoostParameters.TreeMethod.auto});
+                    searchParams.put("_booster", new XGBoostParameters.Booster[]{ //gblinear crashes currently
```
michalkurka: gblinear doesn't crash anymore

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostStepsProvider.java
sebhrusen: checked the impl: yes, it's probably better, will switch. Thx

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
sebhrusen: thx for info, I'll reconsider this

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
```diff
@@ -0,0 +1,188 @@
+package ai.h2o.automl.modeling;
```
sebhrusen: nothing new here: just extracted this class for decoupling (it contains dependencies to XGBoost module) as it used to be internal to `XGBoostStepsProvider`

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
michalkurka: maybe it could be a different steps (since the parameters are differetn)

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
```diff
@@ -0,0 +1,188 @@
...
+
+                    searchParams.put("_sample_rate", new Double[]{0.6, 0.8, 1.0}); // = _subsample
+                    searchParams.put("_col_sample_rate" , new Double[]{ 0.6, 0.8, 1.0}); // = _colsample_bylevel"
+                    searchParams.put("_col_sample_rate_per_tree", new Double[]{ 0.7, 0.8, 0.9, 1.0}); // = _colsample_bytree: start higher to always use at least about 40% of columns
+//                    searchParams.put("_learn_rate", new Double[]{0.01, 0.05, 0.08, 0.1, 0.15, 0.2, 0.3, 0.5, 0.8, 1.0}); // = _eta
+//                    searchParams.put("_min_split_improvement", new Float[]{0.01f, 0.05f, 0.1f, 0.5f, 1f, 5f, 10f, 50f}); // = _gamma
+//                    searchParams.put("_tree_method", new XGBoostParameters.TreeMethod[]{XGBoostParameters.TreeMethod.auto});
+                    searchParams.put("_booster", new XGBoostParameters.Booster[]{ //gblinear crashes currently
+                            XGBoostParameters.Booster.gbtree, //default, let's use it more often
+                            XGBoostParameters.Booster.gbtree,
```
michalkurka: does GridSearch skip duplicate parameters? eg. ntrees=5, booster=gbtree and second ntrees=5, booster=gbtree'

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
sebhrusen: https://0xdata.atlassian.net/browse/PUBDEV-7254

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
sebhrusen: damn, you're right, random search keeps track of indices combinations, not of parameters combinations...
I guess the harm is not big today, as the hyperspace is large, but I need to reconsider this: I wish there was a way to give weights to some param, but as you suggested above, maybe the best here is to have a separate grid, though I'm not sure it's worth having a full grid with similar search space for each booster.
Will definitely revisit this.

sebhrusen: https://0xdata.atlassian.net/browse/PUBDEV-7289

sebhrusen:
@michalkurka @navdeep-G @deil87 Any other concern about this PR? Checks are good otherwise.

PR #4277, opened 2020-01-30
PUBDEV-7230 - Force connections to FlowUI on port 54321 to TLS1.2

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7230

This is probably the way we should go:
- Keep Jetty updated as much as possible (security)
- Disable old TLS once and for all. Disable 1.2 once 1.3 support comes out as well ?

### DIFF
--- a/h2o-jetty-9/build.gradle
+++ b/h2o-jetty-9/build.gradle
@@ -1,10 +1,10 @@
 dependencies {
     compile project(":h2o-webserver-iface")
-    compile 'org.eclipse.jetty:jetty-server:9.4.26.v20200117'
-    compile "org.eclipse.jetty:jetty-servlets:9.4.26.v20200117"
-    compile 'org.eclipse.jetty:jetty-jaas:9.4.26.v20200117'
-    compile 'org.eclipse.jetty:jetty-proxy:9.4.26.v20200117'
-    compile 'org.eclipse.jetty:jetty-servlet:9.4.26.v20200117'
+    compile 'org.eclipse.jetty:jetty-server:9.3.20.v20170531'
+    compile "org.eclipse.jetty:jetty-servlets:9.3.20.v20170531"
+    compile 'org.eclipse.jetty:jetty-jaas:9.3.20.v20170531'
+    compile 'org.eclipse.jetty:jetty-proxy:9.3.20.v20170531'
+    compile 'org.eclipse.jetty:jetty-servlet:9.3.20.v20170531'
     testCompile group: "junit", name: "junit", version: "4.12"
     testCompile "org.mockito:mockito-core:2.23.0"
 }
--- a/h2o-jetty-9/build.gradle
+++ b/h2o-jetty-9/build.gradle
@@ -1,5 +1,6 @@
 dependencies {
     compile project(":h2o-webserver-iface")
+    compile project(":h2o-security")
     compile 'org.eclipse.jetty:jetty-server:9.3.20.v20170531'
     compile "org.eclipse.jetty:jetty-servlets:9.3.20.v20170531"
     compile 'org.eclipse.jetty:jetty-jaas:9.3.20.v20170531'
--- a/h2o-jetty-9/build.gradle
+++ b/h2o-jetty-9/build.gradle
@@ -1,10 +1,10 @@
 dependencies {
     compile project(":h2o-webserver-iface")
-    compile 'org.eclipse.jetty:jetty-server:9.3.20.v20170531'
-    compile "org.eclipse.jetty:jetty-servlets:9.3.20.v20170531"
-    compile 'org.eclipse.jetty:jetty-jaas:9.3.20.v20170531'
-    compile 'org.eclipse.jetty:jetty-proxy:9.3.20.v20170531'
-    compile 'org.eclipse.jetty:jetty-servlet:9.3.20.v20170531'
+    compile 'org.eclipse.jetty:jetty-server:9.4.26.v20200117'
+    compile "org.eclipse.jetty:jetty-servlets:9.4.26.v20200117"
+    compile 'org.eclipse.jetty:jetty-jaas:9.4.26.v20200117'
+    compile 'org.eclipse.jetty:jetty-proxy:9.4.26.v20200117'
+    compile 'org.eclipse.jetty:jetty-servlet:9.4.26.v20200117'
     testCompile group: "junit", name: "junit", version: "4.12"
     testCompile "org.mockito:mockito-core:2.23.0"
 }
--- a/h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
+++ b/h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
@@ -2,14 +2,7 @@ package water.webserver.jetty9;
 
 import ai.h2o.org.eclipse.jetty.security.authentication.SpnegoAuthenticator;
 import org.eclipse.jetty.jaas.JAASLoginService;
-import org.eclipse.jetty.security.Authenticator;
-import org.eclipse.jetty.security.ConstraintMapping;
-import org.eclipse.jetty.security.ConstraintSecurityHandler;
-import org.eclipse.jetty.security.DefaultIdentityService;
-import org.eclipse.jetty.security.HashLoginService;
-import org.eclipse.jetty.security.IdentityService;
-import org.eclipse.jetty.security.LoginService;
-import org.eclipse.jetty.security.SpnegoLoginService;
+import org.eclipse.jetty.security.*;
 import org.eclipse.jetty.security.authentication.BasicAuthenticator;
 import org.eclipse.jetty.security.authentication.FormAuthenticator;
 import org.eclipse.jetty.server.*;
@@ -60,8 +53,8 @@ class Jetty9Helper {
 
         final ServerConnector connector;
         if (isSecured) {
-            final SslContextFactory.Server sslContextFactory = new SslContextFactory.Server();
-            sslContextFactory.addExcludeProtocols("TLSv1.1", "TLSv1");
+            final SslContextFactory sslContextFactory = new SslContextFactory(config.jks);
+            sslContextFactory.setIncludeCipherSuites("TLSv1.2"); // TLS 1.2 is prioritized
             sslContextFactory.setKeyStorePassword(config.jks_pass);
             if (config.jks_alias != null) {
                 sslContextFactory.setCertAlias(config.jks_alias);
@@ -155,10 +148,10 @@ class Jetty9Helper {
 
         final SessionHandler sessionHandler = new SessionHandler();
         if (config.session_timeout > 0) {
-            sessionHandler.setMaxInactiveInterval(config.session_timeout * 60);
+            sessionHandler.getSessionManager().setMaxInactiveInterval(config.session_timeout * 60);
         }
         sessionHandler.setHandler(security);
-        jettyServer.setSessionIdManager(sessionHandler.getSessionIdManager());
+        jettyServer.setSessionIdManager(sessionHandler.getSessionManager().getSessionIdManager());
 
         // Pass-through to H2O if authenticated.
         jettyServer.setHandler(sessionHandler);
--- a/h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
+++ b/h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
@@ -15,6 +15,7 @@ import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
 import org.eclipse.jetty.util.thread.ScheduledExecutorScheduler;
 import org.eclipse.jetty.util.thread.Scheduler;
+import water.network.SecurityUtils;
 import water.webserver.iface.H2OHttpConfig;
 import water.webserver.iface.H2OHttpView;
 import water.webserver.iface.LoginType;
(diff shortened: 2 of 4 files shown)


h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
```diff
@@ -154,10 +155,10 @@ HandlerWrapper authWrapper(Server jettyServer) {
 
         final SessionHandler sessionHandler = new SessionHandler();
         if (config.session_timeout > 0) {
-            sessionHandler.getSessionManager().setMaxInactiveInterval(config.session_timeout * 60);
+            sessionHandler.setMaxInactiveInterval(config.session_timeout * 60);
```
Pscheidl: Required because of the new Jetty version.

h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9Helper.java
```diff
@@ -154,10 +155,10 @@ HandlerWrapper authWrapper(Server jettyServer) {
...
         }
         sessionHandler.setHandler(security);
-        jettyServer.setSessionIdManager(sessionHandler.getSessionManager().getSessionIdManager());
+        jettyServer.setSessionIdManager(sessionHandler.getSessionIdManager());
```
Pscheidl: #heretoo !

Pscheidl:
FYI the Jetty upgrade is not needed.

h2o-jetty-9/build.gradle
```diff
@@ -1,10 +1,10 @@
 dependencies {
     compile project(":h2o-webserver-iface")
-    compile 'org.eclipse.jetty:jetty-server:9.3.20.v20170531'
-    compile "org.eclipse.jetty:jetty-servlets:9.3.20.v20170531"
-    compile 'org.eclipse.jetty:jetty-jaas:9.3.20.v20170531'
-    compile 'org.eclipse.jetty:jetty-proxy:9.3.20.v20170531'
-    compile 'org.eclipse.jetty:jetty-servlet:9.3.20.v20170531'
+    compile 'org.eclipse.jetty:jetty-server:9.4.26.v20200117'
```
michalkurka: Just curious: Is there a reason to upgrade the version?

Pscheidl: No, there is not, besides fixes & possible security fixed on the way: https://github.com/eclipse/jetty.project/releases/tag/jetty-9.4.26.v20200117

kurkami: In that case because the upgrade includes an API change, I suggest to

- separate this fix from the upgrade,
- put the upgrade only in master - we don't want h2o fix release upgrade to break stuff for users who are embedding h2o into their apps,
- make sure all Hadoop and Spark tests pass.

Pscheidl:
I don't think this PR is right. In fact, it could be harmful.

We can of course set this programatically, limiting Jetty to a very  ciphers & protocols only. But why'd we want to do that in the first place ?  Disabling certain cryptography algos could mean inability to connect from some older browsers. And I've personally seen users who are stuck to old IE due to company policy.

## Defining protocols used in runtime environment
As defined in [JSSE Reference guide](https://docs.oracle.com/javase/7/docs/technotes/guides/security/jsse/JSSERefGuide.html#InstallationAndCustomization), protocols, ciphers and crypto algos in general are determined by Java runtime environment used an can be configured on the environment level. Our customer in the [support ticket](https://support.h2o.ai/a/tickets/96300) even found that out. We should not strictly limit Jetty into excluding older algorithms and ciphers and including only the specified ones. This should be defined by the runtime environment nad configured such as: 

`java ‑Djdk.tls.client.protocols="TLSv1.2" -Djdk.tls.disabledAlgorithms="SSLv2Hello, SSLv3, TLSv1, TLSv1.1" myApp` (the order means priority).



Good explanation is to be found here: https://www.java.com/en/configure_crypto.html

Jetty itself is only limited to what the runtime environment provides + programmatic configuration (or any other configuration available, e.g. xml). This is how our version of Jetty works in a  nutshell:

## Prioritize what Jetty is using
We could make Jetty prioritize TLS1.2, if available.

[Jetty documentation](https://www.eclipse.org/jetty/documentation/9.4.x/configuring-ssl.html#configuring-sslcontextfactory) 

Internally, our version of Jetty ,as well as all current 9.x Jettys work as follows in `SslContextFactory` class:

```java
    /**
     * Select protocols to be used by the connector
     * based on configured inclusion and exclusion lists
     * as well as enabled and supported protocols.
     *
     * @param enabledProtocols   Array of enabled protocols
     * @param supportedProtocols Array of supported protocols
     */
    public void selectProtocols(String[] enabledProtocols, String[] supportedProtocols)
    {
        Set<String> selected_protocols = new LinkedHashSet<>();

        // Set the starting protocols - either from the included or enabled list
        if (!_includeProtocols.isEmpty())
        {
            // Use only the supported included protocols
            for (String protocol : _includeProtocols)
            {
                if (Arrays.asList(supportedProtocols).contains(protocol))
                    selected_protocols.add(protocol);
                else
                    LOG.info("Protocol {} not supported in {}", protocol, Arrays.asList(supportedProtocols));
            }
        }
        else
            selected_protocols.addAll(Arrays.asList(enabledProtocols));

        // Remove any excluded protocols
        selected_protocols.removeAll(_excludeProtocols);

        if (selected_protocols.isEmpty())
            LOG.warn("No selected protocols from {}", Arrays.asList(supportedProtocols));

        _selectedProtocols = selected_protocols.toArray(new String[0]);
    }
```
The LinkedHashSet seems to prioritize `_includedProtocols` by adding those first on top of the list. During the negotiation, they should match first. However, there is no official contract. Only Jetty [documentation mentions this](https://www.eclipse.org/jetty/documentation/9.4.x/configuring-ssl.html#configuring-sslcontextfactory): 

```
The SslContextFactory is responsible for:
....omitted...
- Managing order of Ciphers offered (important for TLS/1.2 and HTTP/2 support)
....omitted...
```


And the prioritization is done by setting `Include Protocols`, which is programatically reflected by setting any values `_includeProtocols`. All other available protocols are still listed and used, only after the `_includeProtocols` (ass seen in the code above) is set to some value, those are checked first. Jetty will also check for us whether those are available and LOG a warning (INFO level) about the given protocol not being available.

**Therefore, for both Jetty 8 and 9, we should prioritize TLS1.2.** by using:
```java
sslContextFactory.setIncludeCipherSuites("TLSv1.2"); // TLS 1.2 is prioritized
```

Pscheidl:
Running Hadoop tests against this branch. http://mr-0xc1:8080/view/H2O-3/job/h2o-3-hadoop-smoke-pipeline/job/pavel%252Fpubdev-7230/

Pscheidl:
```
[2020-01-31 16:18:27] Connect to h2o on IP: 192.168.254.8 PORT: 54321

Using basic auth with jenkins user name
  File "/home/jenkins/slave_dir_from_mr-0xc1/workspace/smoke-pipeline_pavel_pubdev-7230/cdh-5.4---standalone/h2o-3/h2o-py/scripts/h2o-py-test-setup.py", line 186, in <module>
    h2o_test_setup(sys.argv)
  File "/home/jenkins/slave_dir_from_mr-0xc1/workspace/smoke-pipeline_pavel_pubdev-7230/cdh-5.4---standalone/h2o-3/h2o-py/scripts/h2o-py-test-setup.py", line 165, in h2o_test_setup
    h2o.connect(ip=_H2O_IP_, port=_H2O_PORT_, verbose=False, auth=auth, **_H2O_EXTRA_CONNECT_ARGS_)
  File "/home/jenkins/slave_dir_from_mr-0xc1/workspace/smoke-pipeline_pavel_pubdev-7230/cdh-5.4---standalone/h2o-3/h2o-py/h2o/h2o.py", line 103, in connect
    verbose=verbose)
  File "/home/jenkins/slave_dir_from_mr-0xc1/workspace/smoke-pipeline_pavel_pubdev-7230/cdh-5.4---standalone/h2o-3/h2o-py/h2o/backend/connection.py", line 384, in open
    conn._cluster = conn._test_connection(retries, messages=_msgs)
  File "/home/jenkins/slave_dir_from_mr-0xc1/workspace/smoke-pipeline_pavel_pubdev-7230/cdh-5.4---standalone/h2o-3/h2o-py/h2o/backend/connection.py", line 684, in _test_connection
    % (self._base_url, max_retries, "\n".join(errors)))
h2o.exceptions.H2OConnectionError: Could not establish link to the H2O cloud https://192.168.254.8:54321 after 5 retries
[18:27.92] H2OConnectionError: Unexpected HTTP error: HTTPSConnectionPool(host='192.168.254.8', port=54321): Max retries exceeded with url: /3/Cloud (Caused by SSLError(SSLError(1, u'[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure
```

http://mr-0xc1:8080/job/h2o-3-hadoop-smoke-pipeline/job/pavel%252Fpubdev-7230/1/artifact/cdh-5.4---standalone/h2o-3/h2o-hadoop-common/tests/python/results/failed/pyunit_PUBDEV_6758_glrm.py.out.txt I'll look into  this. It's not blocking the user.

Pscheidl:
# Setting priority of algorithms inside JDK

As stated above, security capabilities of a Java runtime depend on the specific implementation used. Usage of certain cryptographic capabilities are often restricted to certain areas around the globe, which is the primary reason of [JCE Unlimited Strength's](https://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html) existence.

To set a priority to TLSv1.2 by default inside H2O's embedded Jetty sounds like a good idea, yet there are multiple reasons why it's not. The strongest reason is Java runtime already has fbuilt-in priority settings](https://www.java.com/en/configure_crypto.html) - in the `Understanding the priority order for cipher suites used by TLS 1.2 on JDK 8 ` section.

Also, there are multiple variables:
- `jdk.tls.disabledAlgorithms`
- `jdk.tls.legacyAlgorithms`

which default to different values per each runtime distributor/vendor. H2O should not do any assumptions. Different runtimes might, for any reason, treat the priorities differently.  From the documentation, it is clear it's the role of the environment to `inject` the cryptographic capabilities and settings, not vice versa. I can see **no strong reason** why we should make any assumptions or override environment  settings from inside H2O.

## Disabling TLS properly

In order to disable TLS properly, it should simply be listed in the `jdk.tls.disabledAlgorithms` variable. An example on how to do this with Oracle JDK 8 is to be found [on Oracle's website](https://www.java.com/en/configure_crypto.html). This might vary among certain JDKs and is the responsibility of the environment creator to do these settings properly.

It should be noted these variables like `jdk.tls.disabledAlgorithms` **do have defaults**. E.g. on Oracle JDK 8, the [documentation](https://www.java.com/en/configure_crypto.html) in section `Disable TLS 1.0 and TLS 1.1 ` mentions the following default.

```
jdk.tls.disabledAlgorithms=SSLv3, RC4, DES, MD5withRSA, DH keySize < 1024, \
   EC keySize < 224, 3DES_EDE_CBC, anon, NULL 
```

In order to disable TLS on JDK 8, the previous set of values should be complemented with two new values, namely `{TLSv1, TLSv1.1}`.

```
jdk.tls.disabledAlgorithms=SSLv3, RC4, DES, MD5withRSA, DH keySize < 1024, \
   EC keySize < 224, 3DES_EDE_CBC, anon, NULL, TLSv1, TLSv1.1
```

## Enforcing TLSv1.2 in Jetty & why the tests failed

Our jetty enforced TLS1.2 only by settings the `_includeCipherSuites` inside `SslContextFactory` from inside Jetty.

### Jetty 8

```java
    public String[] selectProtocols(String[] enabledProtocols, String[] supportedProtocols)
    {
        Set<String> selected_protocols = new LinkedHashSet<String>();

        // Set the starting protocols - either from the included or enabled list
        if (!_includeProtocols.isEmpty())
        {
            // Use only the supported included protocols
            for (String protocol : _includeProtocols)
                if(Arrays.asList(supportedProtocols).contains(protocol))
                    selected_protocols.add(protocol);
        }
        else
            selected_protocols.addAll(Arrays.asList(enabledProtocols));


        // Remove any excluded protocols
        if (_excludeProtocols != null)
            selected_protocols.removeAll(_excludeProtocols);

        return selected_protocols.toArray(new String[selected_protocols.size()]);
    }
```

### Jetty 9
```java
    protected void selectCipherSuites(String[] enabledCipherSuites, String[] supportedCipherSuites)
    {
        List<String> selected_ciphers = new ArrayList<>();

        // Set the starting ciphers - either from the included or enabled list
        if (_includeCipherSuites.isEmpty())
            selected_ciphers.addAll(Arrays.asList(enabledCipherSuites));
        else
            processIncludeCipherSuites(supportedCipherSuites, selected_ciphers);

        removeExcludedCipherSuites(selected_ciphers);

        if (selected_ciphers.isEmpty())
            LOG.warn("No supported ciphers from {}", Arrays.asList(supportedCipherSuites));

        Comparator<String> comparator = getCipherComparator();
        if (comparator != null)
        {
            if (LOG.isDebugEnabled())
                LOG.debug("Sorting selected ciphers with {}", comparator);
            Collections.sort(selected_ciphers, comparator);
        }

        _selectedCipherSuites = selected_ciphers.toArray(new String[0]);
    }
```

The following piece is crucial:
```java
        // Set the starting ciphers - either from the included or enabled list
        if (_includeCipherSuites.isEmpty())
            selected_ciphers.addAll(Arrays.asList(enabledCipherSuites));
        else
            processIncludeCipherSuites(supportedCipherSuites, selected_ciphers);
```

where the `processIncludeCipherSuites` method is defined as follows:

```java
    protected void processIncludeCipherSuites(String[] supportedCipherSuites, List<String> selected_ciphers)
    {
        for (String cipherSuite : _includeCipherSuites)
        {
            Pattern p = Pattern.compile(cipherSuite);
            boolean added = false;
            for (String supportedCipherSuite : supportedCipherSuites)
            {
                Matcher m = p.matcher(supportedCipherSuite);
                if (m.matches())
                {
                    added = true;
                    selected_ciphers.add(supportedCipherSuite);
                }

            }
            if (!added)
                LOG.info("No Cipher matching '{}' is supported", cipherSuite);
        }
    }
```

This code's decision flow is simple: If there are no `_includeCipherSuites` defined, then include all enabled. If `_includeCipherSuites` is defined, then include everything stated inside `_includeCipherSuites` that is at the same time supported. Nothing else. This means, by the recent change, I've **effectively restricted H2O to TLS1.2**. Which is not desired. Such behavior **is shared in both Jetty 8 and Jetty 9**.

If we really wanted to prioritize `TLSv1.2` by force inside H2O, which I don't see any reason to do based on the evidence in the first section of this commentary, we'd need to collect all the enabled algorithms and change the order to make `TLSv1.2` first.

Pscheidl:
Closing, as no changes are required.

PR #4283, opened 2020-02-01
[PUBDEV-7258] add test for synonymous XGBoost params

honzasterba:
- also remove n_estimators param since it was not exposed to clients

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/schemas/XGBoostV3.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/schemas/XGBoostV3.java
@@ -90,8 +90,6 @@ public class XGBoostV3 extends ModelBuilderSchema<XGBoost,XGBoostV3,XGBoostV3.XG
 
     @API(help="(same as n_estimators) Number of trees.", gridable = true)
     public int ntrees;
-    @API(help="(same as ntrees) Number of trees.", gridable = true)
-    public int n_estimators;
 
     @API(help="Maximum tree depth.", gridable = true)
     public int max_depth;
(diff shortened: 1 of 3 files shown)


h2o-extensions/xgboost/src/main/java/hex/schemas/XGBoostV3.java
```diff
@@ -90,8 +90,6 @@
 
     @API(help="(same as n_estimators) Number of trees.", gridable = true)
     public int ntrees;
-    @API(help="(same as ntrees) Number of trees.", gridable = true)
-    public int n_estimators;
```
honzasterba: I think we should either remove it or expose it to clients, @michalkurka WDYT?

michalkurka: it is okay to remove from here if it was not exposed

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -265,15 +264,8 @@ public void dump(String format) {
   public static BoosterParms createParams(XGBoostParameters p, int nClasses, String[] coefNames) {
     Map<String, Object> params = new HashMap<>();
 
+    params.put("nround", p._ntrees);
     // Common parameters with H2O GBM
-    if (p._n_estimators != 0) {
```
michalkurka: if `_n_estimators` worked before we shouldn't remove it in a fix release, this is okay to do in master (and make a note in changelog - or migration document)

PR #4284, opened 2020-02-01
PUBDEV-6614: predict OOME when parsing and stop if in danger

honzasterba:


h2o-core/src/main/java/water/FrameSizeMonitor.java
```diff
@@ -0,0 +1,211 @@
...
+    private static final String SAFE_COEF_PROP = "util.frameSizeMonitor.safetyCoefficient";
+    private static final String SAFE_FREE_MEM_DEFAULT_COEF = "0.2";
+
+    private static final int LOG_LEVEL;
+    private static final boolean ENABLED;
+    private static final float SAFE_FREE_MEM_COEF;
+    private static final int SLEEP_MS = 100;
+    private static final int MB = 1024 * 1024;
+    
+    private static final Map<Key<Job>, FrameSizeMonitor> registry = new HashMap<>();
```
Pscheidl: It's not serialized anywhere (currently), but I'd rather mark it as transient. To make the message clear. Otherwise It'd have to be a different implementation anyway.

h2o-core/src/main/java/water/FrameSizeMonitor.java
```diff
@@ -0,0 +1,211 @@
...
+    
+    private static void finish(Key<Job> jobKey) {
+        synchronized (registry) {
+            registry.remove(jobKey);
+        }
+    }
+
+    @Override
+    protected void map(int id) {
+        float nextProgress = 0.02f;
```
Pscheidl: Can we name the start and the progress step in some way please, so it's more explanatory ? Start is at `0.02f` and the step seems to be `0.01f` and then `0.1f`.

h2o-core/src/main/java/water/FrameSizeMonitor.java
```diff
@@ -0,0 +1,211 @@
...
+                    nextProgress = currentProgress + 0.01f;
+                } else {
+                    nextProgress = currentProgress + 0.1f;
+                }
+            } else if (Log.isLoggingFor(LOG_LEVEL)) {
+                Log.log(LOG_LEVEL, "FrameSizeMonitor: waiting for progress " + currentProgress + " to jump over " + nextProgress);
+            }
+            synchronized (this) {
+                try {
+                    wait(SLEEP_MS);
```
Pscheidl: I can imagine the next step to be dynamically making this interval longer or shorter (shorter is probably more important) as H2O is getting closer to the memory limit.

Not a scope of this PR I'd say. Just an observation.

h2o-core/src/main/java/water/FrameSizeMonitor.java
```diff
@@ -0,0 +1,211 @@
...
+            }
+            float currentProgress = job.progress();
+            if (currentProgress >= nextProgress) {
+                if (isMemoryUsageOverLimit() && isFrameSizeOverLimit(currentProgress, job)) {
+                    job.fail(new RuntimeException("Aborting due to projected memory usage too high."));
+                    break;
+                } else if (nextProgress < 0.1f) {
+                    nextProgress = currentProgress + 0.01f;
+                } else {
+                    nextProgress = currentProgress + 0.1f;
```
Pscheidl: I can imagine making this interval smaller for large datasets, checking per every 10 percent parsed could go wrong. Based on how much data was parsed in the first 10 percent of progress maybe ? If the data initially parsed take up more than 10% of available memory, it may be reasonable to make this interval shorter.

h2o-core/src/main/java/water/FrameSizeMonitor.java
```diff
@@ -0,0 +1,219 @@
...
+    private final long totalMemory = getTotalMemory();
+
+    private long committedMemory = 0;
+    
+    FrameSizeMonitor(Key<Job> jobKey) {
+        this.jobKey = jobKey;
+    }
+    
+    public static void get(Key<Job> jobKey, Consumer<FrameSizeMonitor> c) {
+        synchronized (registry) {
```
michalkurka: Can we get away without using this global synchronized block?

IMHO you could just use a ConcurrentMap for registry and synchronize operation on the FrameSizeMonitor level.

Also with this mechanism disabled, do we need to enter the synchronized block at all?

h2o-core/src/main/java/water/FrameSizeMonitor.java
honzasterba: I agree this approach has its weaknesses, but there is a balance of checking too often and not checking often enough. This code is not meant to be bullet-proof, but has proven by my testing to work quite well.
Concrete suggestions on how to improve this are welcome.

PR #4287, opened 2020-02-03
Revert back to Gradle 4

michalkurka:
Also use older version of dev images (18)

Gradle 5.x doesn't support Java 13

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -240,7 +240,13 @@ subprojects {
             }
         }
     }
-    
+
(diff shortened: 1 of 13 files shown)


jakubhava:
Instead of downgrading, could we upgrade to Gradle 6.x which has the support in already? 

2 reasons -> We have already ported some code to be Gradle 5 friendly + it will stop working with bulding against Sparkling Water again

PR #4290, opened 2020-02-04
[PUBDEV-7260] XGBoost monotonicity fixes and tests

honzasterba:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -215,6 +215,9 @@
       error("_grow_policy", "must use tree_method=hist for grow_policy=lossguide");
 
     if ((_train != null) && (_parms._monotone_constraints != null)) {
+      if (_parms._tree_method == XGBoostModel.XGBoostParameters.TreeMethod.approx) {
```
michalkurka: I would just future proof this code by adding an assert that the set of possible tree methods is the current one. Then we can catch that a new tree method was added and that we might need to revisit this condition.

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostMonotoneTest.java
```diff
@@ -0,0 +1,86 @@
...
+import water.H2O;
+import water.Scope;
+import water.TestUtil;
+import water.fvec.Frame;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class XGBoostMonotoneTest  extends TestUtil  {
```
Pscheidl: Commercial break: you can use `H2ORunner` :) But it's definitely not necessary.

PR #4292, opened 2020-02-04
PUBDEV-7271 - Sequential DKV manager

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7271'

H2O is not completely safe for parallel usage. The internal DKV store does not handle parallel removal of a single value well.

The current removal mechanism is prone to errors due to possible removal of several cross-validation models multiple times.

### DIFF
--- a/h2o-core/src/main/java/water/DKVManager.java
+++ b/h2o-core/src/main/java/water/DKVManager.java
@@ -3,7 +3,7 @@ package water;
 import hex.Model;
 import water.fvec.Frame;
 import water.fvec.Vec;
-import water.util.Log;
+import water.util.ArrayUtils;
 
 import java.util.*;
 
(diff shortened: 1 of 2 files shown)


Pscheidl:
Please review but do not merge yet.

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -17,14 +17,81 @@
...
+  /**
+   * Collects a {@link Set} of keys available cluster-wide
+   *
+   * @return
+   */
+  private static final Set<Value> collectUniqueKeys() {
+    final Value[] collectedValues = new CollectValuesTask()
+            .doAllNodes()
+            ._collectedValues;
+    final Set<Value> clusterValues = new HashSet<>(collectedValues.length);
```
michalkurka: Why is this necessary?

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -17,14 +17,81 @@
...
+   * @param removedValues Values to be removed
+   * @param retainedKeys  A {@link Set} of keys to be retained
+   */
+  private static void removeValues(final Set<Value> removedValues, final Set<Key> retainedKeys) {
+
+    for (final Value value : removedValues) {
+      if (retainedKeys.contains(value._key)) {
+        continue;
+      }
+      final Futures removalFutures = new Futures();
```
michalkurka: This all deletes the metadata (container objects like Frame / Model) sequentially, however the contained objects are still deleted in parallel.

michalkurka:
I don't think it is feasible to collapse whole DKV to a single node. If you need to work sequentially then you can delete node by node (the simplest what I can think of).

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: It's not, you're right. Removed. Now there is an array.

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: The containing objects should delete distinct objects - problem is parallel or even sequential invocation of deletion of a Model with default remove_impl - as this might cause nasty errors.

Pscheidl:
Due to the change of modus operandi, closing this PR and creating the proper version over here: https://github.com/h2oai/h2o-3/pull/4311

PR #4300, opened 2020-02-06
PUBDEV-7280_demo_setup

koniecsveta:


h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
```diff
@@ -622,6 +625,21 @@ public SharedTreeSubgraph getSharedTreeSubgraph(final int tidx, final int cls) {
...
   }
 
+  /**
+   * Provides a visualization of tree based models when called from Flow.
+   * @param output output file name
+   */
+  @Override
+  public void visualize(String output) {
+    JgraphtPrintMojo jgraphtPrintMojo = new JgraphtPrintMojo();
+    List<SharedTreeSubgraph> subgraphArray = new ArrayList<>();
```
honzasterba, line 635: confusing variable name, its not an array, would rename to subgraphs, or subgraphList

h2o-core/build.gradle
```diff
@@ -7,6 +7,7 @@ description = "H2O Core"
 
 dependencies {
   // Dependency on shared model support
+  compile project(":h2o-genmodel-ext-jgrapht")
```
honzasterba, line 10: would pu this after h2o-genmodel, just to have them together

h2o-core/src/main/java/hex/Model.java
```diff
@@ -166,7 +166,14 @@ else if (_output.nclasses() > 1)
   }
 
   public ToEigenVec getToEigenVec() { return null; }
-
+  
+  /**
+   * Intended for overriding in SharedTreeModel for tree visualization.
+   */
+  public void visualize(String output) {
```
honzasterba, line 173: I think it would be better to have an interface for this method, or we can add it to sharedtreemodel or simmilar place and check with instanceof and return an error when user tries to visualise an usupported model

h2o-core/src/main/java/water/util/ModelPngWriter.java
```diff
@@ -0,0 +1,81 @@
...
+                byte[] buf = new byte[8192];
+                int c;
+                while ((c = is.read(buf, 0, buf.length)) > 0) {
+                    os.write(buf, 0, c);
+                    os.flush();
+                }
+                os.close();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
honzasterba, line 36: add an error message saying what went wrong - was it reading the png and writing it to stream?

h2o-core/src/main/java/water/util/ModelPngWriter.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Creates new ZIP archive containing file or directory represented by passed {@code path}. If {@code path}
+     * represents directory, all its files and sub-directories are recursively added to the resulting archive.
+     *
+     * @param path file or directory to be included in new ZIP archive
+     * @return {@link Path} of newly created ZIP file
+     */
+    private Path writeToZip(Path path) throws IOException {
```
honzasterba, line 47: I think we have code like this around the codebase already, don't we? maybe we should reuse it?

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
```diff
@@ -42,18 +42,28 @@ public void run() throws Exception {
             }
         } 
     }
+    
+    public void runFromFlow(SharedTreeGraph g, String outputPath){
+        try {
+            destination = outputPath;
```
honzasterba, line 48: I dont like this side-effect, maybe add a constructor that will initialize this property

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -28,11 +28,14 @@
   protected int treeToPrint = -1;
   protected int maxLevelsToPrintPerEdge = 10;
   protected boolean detail = false;
-  protected String outputFileName = null;
+  protected String destination = null;
   protected String optionalTitle = null;
   protected PrintTreeOptions pTreeOptions;
   protected boolean internal;
   protected final String tmpOutputFileName = "tmpOutputFileName.gv";
+  protected int nPlaces = -1;
```
honzasterba, line 36: these are never used, only in runFromFlow which is a subclass, I think it would be better to create default PrintTreeOptions into a new constructor

h2o-core/src/main/java/water/util/ModelPngWriter.java
koniecsveta: will look for it

h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
koniecsveta: they are used (and can be set to something else by the user) in PrintMojo.java#parseArgs() in version for terminal

h2o-core/src/main/java/hex/Model.java
koniecsveta: I need to use this from core and in core sharedtreemodel is not visible so I cant use instanceof. I think I can do interface 👍

h2o-genmodel-extensions/jgrapht/src/main/java/hex/genmodel/tools/JgraphtPrintMojo.java
koniecsveta: now I am initializing in constructor and calling printPng() from run()

h2o-core/src/main/java/hex/Model.java
honzasterba: why do you need to call it on model? you can make ModelPngWriter make to accept not Model but lets say ModelWithVisualisation

koniecsveta: i see, did that interface

h2o-core/src/main/java/water/util/ModelPngWriter.java
koniecsveta: What we have is not directly usable - it is made specifically for logs archiving (LogArchiveWriter.java). Probably it would be possible to extend existing ZipUtil with archive creation functionality and try to use them in both places. At this time, there are only some methods for zip archive reading. Nevertheless, I would rather create separate task for it and do it independently of this.

h2o-core/src/main/java/water/api/ModelsHandler.java
```diff
@@ -209,7 +209,16 @@ public ModelsV3 deleteAll(int version, ModelsV3 models) {
     if( missing.size() != 0 ) throw new H2OKeysNotFoundArgumentException("(none)", missing.toArray(new String[missing.size()]));
     return models;
   }
-
+  
+  @SuppressWarnings("unused") // called through reflection by RequestServer
+  public StreamingSchema exportTreeVisualization(int version, ModelExportV3 mexport) {
+    Model model = getFromDKV("model_id", mexport.model_id.key());
+    if (!(model instanceof ModelWithVisualization))
+      throw new RuntimeException("Trying to visualize model that does not support it.");
```
honzasterba, line 217: IllegalArgumentException would fit nicely here

h2o-core/src/main/java/water/util/ModelPngWriter.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public void writeTo(OutputStream os) {
+        try {
+            Path pngtemp = Files.createTempFile("", "visualization.png");
+            model.visualize(pngtemp.toAbsolutePath().toString());
+
+            try (InputStream is = new FileInputStream(writeToZip(pngtemp).toFile())) {
+                byte[] buf = new byte[8192];
```
honzasterba, line 27: use IOUtils.copyStream

PR #4158, opened 2019-12-16
[WIP] Integration of Target Encoding into AutoML (new)

deil87:
This PR is a reincarnation of an old one https://github.com/h2oai/h2o-3/pull/3183

DO NOT REVIEW! Still needs lots of refactoring due to changes in AutoML and TargetEncoding.

Working on moving PipelineModel related logic into a separate PR. 

WIP

PR #4303, opened 2020-02-07
[DESIGN] Design for data preprocessing integration into AutoML

deil87:
See markdown file: https://github.com/h2oai/h2o-3/blob/design_for_data_preprocessing_integration/Design%20of%20data%20preprocessing%20in%20AutoML.md


Design of data preprocessing in AutoML.md
```diff
@@ -0,0 +1,93 @@
+### Goal
+Provide a way to support feature preprocessing within AutoML pipeline.
```
michalkurka: The goal is to integrate Target Encoding, not necessarily a generic feature preprocessing. Generic feature preprocessing is one way to do it but is not a necessity.

Design of data preprocessing in AutoML.md
```diff
@@ -0,0 +1,93 @@
...
+- when we extend functionality of ModelBuilder by adding ability to have references to proprocessing models
+
+Not sure which approach is better but probably we can avoid introduction of the PipelineModelBuilder class. But it will increase number of responsibilities of the ModelBuilder class.
+
+#### Pros:
+- We do not hide any functionality behind PipelineModel wrapper
+- All models could be equally seen as pipelines ( which could also be a cons.)
+
+#### Cons:
+- conceptually abusing notion of the Model. Model ( what is also known as Transformer in other frameworks) is supposed to be only trained on data and after that be able to transform similar data. Nothing about data preprocessing is expected here. See Concerns section for more details.
```
michalkurka: Only if you focus on generic preprocessing. If we stay in the context of Target Encoding, it fits the Model very well because that is what the Model does for all other types of encodings.

Also having it built-in has some inherent advantages. The encoding can be applied just to the columns the model actually decides to use.

Design of data preprocessing in AutoML.md
```diff
@@ -0,0 +1,93 @@
...
+- new functionality will be exposed not only to AutoML scope. All algos will be "aware" about data preprocessing from now on. With first approach we would just add notion of ability to be a Proxy model and not all the models will have to realise this interface. Though, absense of the preprocessing models could be seen as "no-impact" on Model, still from code perspective there will a bigger impact. Hopefully preprocessing functionality will be generic enough without depenedencies on a particular types of the models.
```
michalkurka: > new functionality will be exposed not only to AutoML scope

All algos are aware of Categorical Encodings already, I don't see it as an issue. Also, it doesn't need to increase complexity of the Model class at all. The implementation could simply allow passing an external encoder (interface) instead of coupling the Model with Target Encoder capabilities. This could be available to both Java/Python/R.

Design of data preprocessing in AutoML.md
```diff
@@ -0,0 +1,93 @@
...
+For comparison I will be using SparkML framework:
+
+![SparkML vs H2O.ai](sparkml_vs_h2oai.png)
+
+ref. https://www.youtube.com/watch?v=jEyahxFp3ak
+
+Model is being overflowed with responsibilities imho. Transformation logic is being spread over many places in different forms.
+
+### Summary
+- we tend to keep adding functionality into Model class. We might be going to add pipeline related aspect into Model
```
michalkurka: All encodings are handled on the Model level, from this point of view it is internally consistent.

Design of data preprocessing in AutoML.md
```diff
@@ -0,0 +1,93 @@
...
+- lack of symmetry between production and non-production code ( Model vs MojoPipeline )
```
michalkurka: not necessarily, MOJOs can (and should) be adapted to include the TE as one of the options as well

Design of data preprocessing in AutoML.md
```diff
@@ -0,0 +1,93 @@
...
+- Some functionality that could be represented via existing approach of Models(transformer) is hardcoded. Lack of consistency. ( not too much related to a current task)
+
+
+Given current state of h2o-3's design and given we are ready to put responsibility of data preprocessing onto a Model class I think that second approach should be fine.
+But ideally I would want to see
+ - a PipelineModel related classes to be responsible for pipelining/data preprocessing
+ - and hardcoded transformations to be represented as models. ( not too much related to a current task)
+
+### Other approaches to consider ???
+
```
michalkurka: - H2OAssembly - largely abandoned at this point but needs to be mentioned becae "Assemblies are essentially Pipelines"

michalkurka:
Other things to consider:

- using a ModelPipeline gives the flexibility to apply just to a subset of columns, other categorical encodings are applied to all columns (could be changed...)
- presenting ModelPipeline to users will mean additional work - on the client-side, Flow
- an API to design a ModelPipeline will eventually be necessary - AutoML shouldn't be the only way to define the pipeline
- having pipelines done right is harder than just building the functionality in (nothing needs to be invented, we would just be following existing patterns)

Having pipelines seems ultimately a more generic approach opening up possibilities for the future. On the other hand, TE is still _just_ categorical encoding which is a well-established mechanism that is already build-in the Model class. There is thus no immediate need to implement TE using Pipelines.

To me, user experience is a key factor in deciding which approach to adopt. It seems to me the build-in approach seems little bit more straightforward given current H2O design and implementation.

I am open to the PipelineModel approach, however, it should be demonstrated - how does the user take advantage of the specificities of the particular models. And how is this sustainable with the generic nature of PipelineModels (they can become anything, not just 2 step processes with TE).

Design of data preprocessing in AutoML.md
deil87: @michalkurka I think that here is an example of our main issue. We should not just follow these 3 steps:

1. Design
2. Review design
3. Implement

 as you mentioned in h2o-3-auto-ml channel  https://h2oai.slack.com/archives/C0E0ADTM1/p1580849237085900?thread_ts=1580839939.083900&cid=C0E0ADTM1

Steps instead should be:
1.  Discuss the goal with all interested parties (automl team, h2o core team, etc)
2. Design
3. Review design
4. Implement

Goal is seriously narrowing down space of possible solutions. In most cases it defines solution in 1:1 fashion.

In current situation we have two different goals for a single task;
1) AutoML team:  Integrate TE into AutoML and make it **in generic way**  ( here is a discussion with @sebhrusen https://github.com/h2oai/h2o-3/pull/4158#issuecomment-579401097 )
2) Core team:   Integrate TE into AutoML **in not necessarily generic way** ( here is a statement https://github.com/h2oai/h2o-3/pull/4303#discussion_r376519702)

So no one can "sit on multiple chairs". But I consider myself as a member of AutoML team so I should probably be in favour our AutoML team's goals. It is not clear to me whether we should always agree with core teams goals. But anyway let's make a |discussion of the goal" a first step. 
This is the main issue and not implementing without reviewed design(but of course this is a good step as well)
@ledell @sebhrusen  @mmalohlava

Design of data preprocessing in AutoML.md
deil87: @michalkurka 
>because that is what the Model does for all other types of encodings.

this was explained further in Concerns section. I was talking here in the most general sense. Models are not supposed to do encoding or any preprocessing on their own. But in current state of h2o-3's design  Model class does these kind of things. In that sense, yes, hardcoding TE inside Model will fits well. In my opinion it is not an optimal long-term direction. Also embedding TE in a non generic way(like other existing encoders) will be a contribution to an already existing debt as hardcoded solutions are not scalable ones.

> Also having it built-in has some inherent advantages. The encoding can be applied just to the columns the model actually decides to use.

 If by "model decides" you mean which columns are going to be excluded due to "bad columns" and "ignore columns" mechanisms I would say that having them first encoded will prevent Model from "deciding" to ignore those columns. In short, model's decisions should come after data preprocessing steps.

Anyway I think that both approaches actually are in equal position in that concern.

Design of data preprocessing in AutoML.md
deil87: I was talking in terms of new TE functionality. Even though Models, as you just have said, are aware of Categorical Encoding, they will not be aware of TE in case we use PipelineModel. So I think my point still valid.

But agree, knowing that Models already have embedded preprocessing logic like Categorical Encodings means we can add TE in the same fashion and it will not increase complexity in terms of types of responsibilities ( as we already have preprocessing type of activity)

Design of data preprocessing in AutoML.md
deil87: @michalkurka it would be consistent internally, true. But I want to highlight here that I separate Transformation aspect from Pipelining aspect. We already have both of them hardcoded inside Model. And my opinion is that these aspects could be handled separately.

Design of data preprocessing in AutoML.md
sebhrusen: @michalkurka @deil87 I would suggest that put both goals here.
1. One is narrow and more short-term: TE integration.
2. A broader one to fulfil AutoML extension needs. As Andrey mentioned, AutoML plans to add more feature engineering in a near future: feature extraction, selection and probably more...
If AutoML simply wanted to apply PCA today, it would not be possible to provide a usable model to the user: so on our side, the absence of such pipeline abstraction is impeding us.

So, if I agree that the current integration deals only with TE, we should not forget the bigger picture on AutoML side, and for such level of abstraction, I think it's reasonable to add it to the goal.
The fact that TE is currently both an encoding step and a model (like PCA) probably adds to the confusion between those 2 goals...

Design of data preprocessing in AutoML.md
sebhrusen: Yeah, we also mentioned it in our discussions.
Agree that it should be mentioned, with its limitations: abandoned, supports only `Transformer`s, doesn't provide any single model feature.

Design of data preprocessing in AutoML.md
deil87: @michalkurka probably misunderstanding here. I meant that on MOJO side we have separate entity MojoPipeline. If we were to integrate preprocessing into just Model that would contribute against symmetry.

Design of data preprocessing in AutoML.md
deil87: @michalkurka yes, thank you. @sebhrusen told me about existence of `Assembly` in our automl discussion. I forgot to add it to the list.

Design of data preprocessing in AutoML.md
michalkurka: @sebhrusen I agree that the bigger picture should always be considered.

However, the solution should IMHO be designed to perfectly solve the specific problem. It might happen that a generic solution is _the_ perfect solution we are looking for.

I am not concerned much with how things are implemented in the code (that can always change) but more of what is presented to the users. What does the solution take away and what are the advantages.

deil87:
@michalkurka 
> using a ModelPipeline gives the flexibility to apply just to a subset of columns, other categorical encodings are applied to all columns (could be changed...)

This is what I think is a Pipeline aspect. It as hardcoded now as well as data preprocessing.

> presenting ModelPipeline to users will mean additional work - on the client-side, Flow

As I suggested on the diagram, we can just use `getMainModel` before returning PipelineModel to the clients. Maybe not that much work needed. Maybe just add list of keys of the preprocessing models to the response. I think we would need similar kind of changes anyway as (as you mentioned before) users like to look into the models and they would want to see TE model.  

> an API to design a ModelPipeline will eventually be necessary - AutoML shouldn't be the only way to define the pipeline

if ModelPipeline is a Model -> on the backend it would be available for free. For defining model from clients, agree. It will require work.

> TE is still just categorical encoding which is a well-established mechanism that is already build-in the Model class.
sorry, don't see your point here. Everything we implementing could be seen as just some encoding with well established mechanism and the fact that we have TE Model does not give us an answer whether to put it inside Model or inside PipelineModel.

Design of data preprocessing in AutoML.md
deil87: @sebhrusen 
> The fact that TE is currently both an encoder and a model (like PCA) probably adds to the confusion between those 2 goals...

Could you please elaborate on what do you mean it is both? It is just Model like PCA. The fact that there is a helper TargetEncoder should not contribute to confusions. Goals imho are different just because what is long-term from h2o-core perspective is short-term from automl team's perspective. 

I also thought that storing references to other models in a cornerstone Model class should be considered as  not an ideal solution.

deil87:
@michalkurka @sebhrusen 

Let's summarise:
My current task will be only short-term goal. Let's describe it in more details here.
1) According to 
>The implementation could simply allow passing an external encoder (interface) instead of coupling the Model with Target Encoder capabilities.

I can use `Model` interface as we have `TargetEncoderModel`. It means that I will store reference to the TE model somewhere within Model class(or it's output). No need for collections of preprocessing models, I can be as specific as `addTEModel(Model model)`

2) User experience will be changed in a way that on clients' sided user in some way will be able to access used TE model. Maybe name of the main model should reflect (eg by means of suffix) that some preprocessing took place.

3) ?

Design of data preprocessing in AutoML.md
deil87: @michalkurka @sebhrusen To formalise what was discussed here I have created jira for our long-term goal on Pipelines https://0xdata.atlassian.net/browse/PUBDEV-7291 .

PR #4304, opened 2020-02-10
PUBDEV-6917 - Enable Azure ABFSS by passing whole Hadoop configuration

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6917

Everything is described in the SW PR: https://github.com/h2oai/sparkling-water/pull/1808

### DIFF
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -367,6 +367,9 @@ final public class H2O {
     /** -hdfs_config=hdfs_config; configuration file of the HDFS */
     public String hdfs_config = null;
 
+    /** Whole hadoop configuration to be used by H2O **/
+    public String hadoop_config = null;
+
     /** -hdfs_skip=hdfs_skip; used by Hadoop driver to not unpack and load any HDFS jar file at runtime. */
     public boolean hdfs_skip = false;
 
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/H2O.java
```diff
@@ -367,6 +367,9 @@ private KeyValueArg(String key, String value) {
     /** -hdfs_config=hdfs_config; configuration file of the HDFS */
     public String hdfs_config = null;
 
+    /** Whole hadoop configuration to be used by H2O **/
+    public String hadoop_config = null;
```
michalkurka: How is Hadoop config different from hdfs_config? It does the exact same thing. The reason why `hdfs_config` is called `hdfs` is because it configures the persistence layer (it doesn't have to be just core-site.xml - it can be anything).

Pscheidl: `hdfs_config` is meant to be specialized for HDFS only. For this PR, it'd be sufficient to push the whole hadoopConfig as `hdfs_config` - that's true. Having a property with the whole config to distinct a specialized `hdfs_config` from the whole hadoop configuration is just a matter if  opinion guess. I'm fine with both.

michalkurka: It is unclear which one to use (for the user). And since there is no actual difference in what it does - there is no reason to confuse the users IMHO.

Pscheidl: Ok, I'll rework it.

Pscheidl:
This change is not required, as hdfs_config will be reused.

PR #4309, opened 2020-02-11
Avoid array copying in Model.clearModelMetrics

Pscheidl:
Found this piece of code during implementation of a different task. It's a tiny small refactor I'd like to propose.

1. The array is enlarged every time a new metric is removed. We know beforehand how many models will be removed.
2. It is searched in the Array of **not removed** keys sequentially, we might use Set to be speed it up a little bit :)

### DIFF
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -840,26 +840,24 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
     public boolean isAutoencoder() { return false; } // Override in DeepLearning and so on.
 
     public synchronized Key<ModelMetrics>[] clearModelMetrics(boolean keepModelTrainingMetrics) {
-      final List<Key<ModelMetrics>> removed;
+      Key<ModelMetrics>[] removed;
       if (keepModelTrainingMetrics) {
-        final Set<Key> kept = new HashSet(0);
-        if (_training_metrics != null) kept.add(_training_metrics._key);
-        if (_validation_metrics != null) kept.add(_validation_metrics._key);
-        if (_cross_validation_metrics != null) kept.add(_cross_validation_metrics._key);
+        Key<ModelMetrics>[] kept = new Key[0];
+        if (_training_metrics != null) kept = ArrayUtils.append(kept, _training_metrics._key);
+        if (_validation_metrics != null) kept = ArrayUtils.append(kept, _validation_metrics._key);
+        if (_cross_validation_metrics != null) kept = ArrayUtils.append(kept, _cross_validation_metrics._key);
 
-        removed = new ArrayList<>(_model_metrics.length - kept.size());
+        removed = new Key[0];
         for (Key<ModelMetrics> k : _model_metrics) {
-          if (!kept.contains(k)) {
-            removed.add(k);
-          }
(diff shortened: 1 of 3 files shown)


h2o-core/src/main/java/hex/Model.java
```diff
@@ -842,24 +842,26 @@ public ModelCategory getModelCategory() {
...
 
     public synchronized Key<ModelMetrics>[] clearModelMetrics(boolean keepModelTrainingMetrics) {
-      Key<ModelMetrics>[] removed;
+      final List<Key<ModelMetrics>> removed;
       if (keepModelTrainingMetrics) {
-        Key<ModelMetrics>[] kept = new Key[0];
-        if (_training_metrics != null) kept = ArrayUtils.append(kept, _training_metrics._key);
-        if (_validation_metrics != null) kept = ArrayUtils.append(kept, _validation_metrics._key);
-        if (_cross_validation_metrics != null) kept = ArrayUtils.append(kept, _cross_validation_metrics._key);
+        final Set<Key> kept = new HashSet(0);
```
michalkurka: Hmm, we have "up to" 3 items populated in this Set. I don't think you will get any performance benefit compared to the original approach (even though we are calling `ArrayUtils.contains` on the 3 item array).

I would keep the array here - the nice thing is that it sets the `_model_metrics` in a nice predictable order (training/valid/cv). Which is good to have.

Pscheidl: But what about the comparisons/lookups done afterwards ?

h2o-core/src/main/java/hex/Model.java
```diff
@@ -842,24 +842,26 @@ public ModelCategory getModelCategory() {
...
+        if (_training_metrics != null) kept.add(_training_metrics._key);
+        if (_validation_metrics != null) kept.add(_validation_metrics._key);
+        if (_cross_validation_metrics != null) kept.add(_cross_validation_metrics._key);
 
-        removed = new Key[0];
+        removed = new ArrayList<>(_model_metrics.length - kept.size());
```
michalkurka: +1

michalkurka:
I don't see a test ;) refactoring without a test = suicide :P

h2o-core/src/main/java/hex/Model.java
michalkurka: lookups in 3 item array = no issue

michalkurka: btw, the appends - that can be avoidedI like your fix there... (see the +1)

sebhrusen: agree with @michalkurka : everything below looks good, the `Set` doesn't look useful in this context though.

Pscheidl: Ok, roger.

Pscheidl:
@michalkurka @sebhrusen Added a basic test. See if this style suits your needs. Abused DummyModel. It'd be nice to have DummyModel outside of the test class, as now it's used by Segmentation models as well. Did not do it in this PR for simplicity, but I can do it. Just expecting anyone to have a strong opinion about that, thus waiting for it.

Looks like the main advantage of this PR would be a proper test of this method. I can also see how DummyModel and its related classes can be turned into a "testing framework" for model basics (let's not do it here and now).
Thanks you.

PR #4311, opened 2020-02-12
PUBDEV-7271 - Per-node sequential DKV manager

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7271

### DIFF
--- a/h2o-core/src/main/java/water/DKVManager.java
+++ b/h2o-core/src/main/java/water/DKVManager.java
@@ -23,9 +23,9 @@ public class DKVManager {
     // Frames and models have multiple nested keys. Those must be extracted and kept from deletion as well.
     extractNestedKeys(retainedSet);
     final Key[] allRetainedkeys = retainedSet.toArray(new Key[retainedSet.size()]);
+    final NodeKeysRemovalTask nodeKeysRemovalTask = new NodeKeysRemovalTask(allRetainedkeys);
 
     for (final H2ONode node : H2O.CLOUD.members()) {
-      final NodeKeysRemovalTask nodeKeysRemovalTask = new NodeKeysRemovalTask(allRetainedkeys);
       H2O.runOnH2ONode(node, nodeKeysRemovalTask);
     }
   }
--- a/h2o-core/src/main/java/water/DKVManager.java
+++ b/h2o-core/src/main/java/water/DKVManager.java
@@ -22,9 +22,10 @@ public class DKVManager {
     retainedSet.addAll(Arrays.asList(retainedKeys));
     // Frames and models have multiple nested keys. Those must be extracted and kept from deletion as well.
     extractNestedKeys(retainedSet);
+    final Key[] allRetainedkeys = retainedSet.toArray(new Key[retainedSet.size()]);
 
     for (final H2ONode node : H2O.CLOUD.members()) {
-      final NodeKeysRemovalTask nodeKeysRemovalTask = new NodeKeysRemovalTask(retainedSet.toArray(new Key[retainedSet.size()]));
+      final NodeKeysRemovalTask nodeKeysRemovalTask = new NodeKeysRemovalTask(allRetainedkeys);
       H2O.runOnH2ONode(node, nodeKeysRemovalTask);
     }
   }
--- a/h2o-core/src/main/java/water/DKVManager.java
+++ b/h2o-core/src/main/java/water/DKVManager.java
@@ -1,9 +1,9 @@
 package water;
 
 import hex.Model;
+import hex.ModelMetrics;
 import water.fvec.Frame;
 import water.fvec.Vec;
-import water.util.Log;
 
 import java.util.*;
 
@@ -17,14 +17,47 @@ public class DKVManager {
    *
    * @param retainedKeys Keys of {@link Frame}s and {@link Model}s to be retained. Only Frame and Model keys are accepted.
    */
-  public static void retain(final Key[] retainedKeys){
+  public static void retain(final Key... retainedKeys) {
     final Set<Key> retainedSet = new HashSet<>(retainedKeys.length);
     retainedSet.addAll(Arrays.asList(retainedKeys));
     // Frames and models have multiple nested keys. Those must be extracted and kept from deletion as well.
     extractNestedKeys(retainedSet);
 
-    new ClearDKVTask(retainedSet.toArray(new Key[retainedSet.size()]))
-            .doAllNodes();
+    for (final H2ONode node : H2O.CLOUD.members()) {
+      final NodeKeysRemovalTask nodeKeysRemovalTask = new NodeKeysRemovalTask(retainedSet.toArray(new Key[retainedSet.size()]));
+      H2O.runOnH2ONode(node, nodeKeysRemovalTask);
+    }
+  }
+
+  /**
+   * Exctracts keys a {@link Model} points to.
+   *
+   * @param retainedKeys A set of retained keys to insert the extracted {@link Model} keys to.
+   * @param model        An instance of {@link Model} to extract the keys from
+   */
+  private static void extractModelKeys(final Set<Key> retainedKeys, final Model model) {
+    Objects.requireNonNull(model);
+    if (model._parms._train != null) {
+      retainedKeys.add(model._parms._train);
+      extractFrameKeys(retainedKeys, model._parms._train.get());
+    }
+    if (model._parms._valid != null) {
+      retainedKeys.add(model._parms._valid);
+      extractFrameKeys(retainedKeys, model._parms._valid.get());
+    }
+
+    if (model._output != null && model._output.getModelMetrics() != null) {
+      for (final Key<ModelMetrics> modelMetric : model._output.getModelMetrics()) {
+        retainedKeys.add(modelMetric);
+      }
+    }
+
+    if (model._output != null && model._output._cross_validation_models != null) {
+      for (final Key xValModel : model._output._cross_validation_models) {
+        retainedKeys.add(xValModel);
+      }
+    }
+
   }
 
   /**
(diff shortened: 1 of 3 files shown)


h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -71,73 +104,39 @@ private static void extractFrameKeys(final Set<Key> retainedkeys, final Frame fr
...
+
+      for (final Key ignoredKey : _ignoredKeys) {
+        ignoredSet.add(ignoredKey);
       }
-      removalFutures.blockForPending();
-    }
 
+      for (final Key key : values) {
+        if (ignoredSet.contains(key)) continue; // Do not perform DKV.get at all if the key is to be ignored
+        if (key.isChunkKey()) continue; // Chunks are deleted with vectors & are not Keyed
```
Pscheidl: Deleting a `Chunk` via `Keyed.remove` is not possible, as it's not `Keyed`. I wonder if it'd be more futureproof to just test whether the underlying `Value` object has class assignable from Keyed.

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -71,73 +104,39 @@ private static void extractFrameKeys(final Set<Key> retainedkeys, final Frame fr
...
+
+        final Value value = DKV.get(key);
+        if (value == null || value.isVecGroup() || value.isNull()) continue;
```
Pscheidl: VecGroups are meanth to be deleted as a part of `Vec` only. It'll `H2O.fail()` on an attempt to `remove_impl` a VecGroup. Not sure whether that's the best thing to do, but let's respect it for now, unless there is a strong reason to change the behavior.

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -71,73 +104,39 @@ private static void extractFrameKeys(final Set<Key> retainedkeys, final Frame fr
...
 
+        if (value.isFrame()) {
+          final Frame frame = value.get();
+          frame.retain(futures, ignoredSet);
+        } else {
+          Keyed.remove(key, futures, true);
```
Pscheidl: Maybe we could only delete frames and models as well.

michalkurka: Yes, I would just do Frame & Models, there are various other things in DKV you don't want to remove (including some obscure ones like user defined functions, or instances of DecryptionTool).

h2o-core/src/main/java/water/DKVManager.java
michalkurka: If we just focus on Models & Frames we don't have to answer this question ;)

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: Exactly. Roger that.

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: Exactly. Deleting models & Frames. It should cascade well. :)

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -17,14 +17,47 @@
...
+  public static void retain(final Key... retainedKeys) {
     final Set<Key> retainedSet = new HashSet<>(retainedKeys.length);
     retainedSet.addAll(Arrays.asList(retainedKeys));
     // Frames and models have multiple nested keys. Those must be extracted and kept from deletion as well.
     extractNestedKeys(retainedSet);
 
-    new ClearDKVTask(retainedSet.toArray(new Key[retainedSet.size()]))
-            .doAllNodes();
+    for (final H2ONode node : H2O.CLOUD.members()) {
+      final NodeKeysRemovalTask nodeKeysRemovalTask = new NodeKeysRemovalTask(retainedSet.toArray(new Key[retainedSet.size()]));
```
honzasterba: do you really need to do retainedSet.toArray every time?

Pscheidl: Of course not. Own code blindness.

Pscheidl: Should be resolved.

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -71,73 +105,39 @@ private static void extractFrameKeys(final Set<Key> retainedkeys, final Frame fr
...
-      Futures removalFutures = new Futures();
-      for (final Value value : storeKeys) {
-        if (retainedKeys.contains(value._key)) {
-          continue;
+    public void run() {
+      final Set<Key> values = H2O.localKeySet();
+      final Set<Key> ignoredSet = new HashSet<>();
+      final Futures futures = new Futures();
+
+      for (final Key ignoredKey : _ignoredKeys) {
```
honzasterba: may be knitpicking but the for-loop mithgt be replaced with the same code used above
`retainedSet.addAll(Arrays.asList(retainedKeys));`

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -17,14 +17,48 @@
    *
    * @param retainedKeys Keys of {@link Frame}s and {@link Model}s to be retained. Only Frame and Model keys are accepted.
    */
-  public static void retain(final Key[] retainedKeys){
+  public static void retain(final Key... retainedKeys) {
     final Set<Key> retainedSet = new HashSet<>(retainedKeys.length);
     retainedSet.addAll(Arrays.asList(retainedKeys));
     // Frames and models have multiple nested keys. Those must be extracted and kept from deletion as well.
     extractNestedKeys(retainedSet);
```
michalkurka: Why do we need to do this for Models? If someone wants to keep the training Frame used to train a Model - it should be in "retainedKeys"

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -17,14 +17,48 @@
...
+      extractFrameKeys(retainedKeys, model._parms._train.get());
+    }
+    if (model._parms._valid != null) {
+      retainedKeys.add(model._parms._valid);
+      extractFrameKeys(retainedKeys, model._parms._valid.get());
+    }
+
+    if (model._output != null && model._output.getModelMetrics() != null) {
+      for (final Key<ModelMetrics> modelMetric : model._output.getModelMetrics()) {
+        retainedKeys.add(modelMetric);
```
michalkurka: Why do we need to worry about non-Frame / non-Model objects?

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -71,73 +105,39 @@ private static void extractFrameKeys(final Set<Key> retainedkeys, final Frame fr
...
+      }
+
+      for (final Key key : values) {
+        if (ignoredSet.contains(key)) continue; // Do not perform DKV.get at all if the key is to be ignored
+        if (key.isChunkKey()) continue; // Chunks are deleted with vectors & are not Keyed
+
+        final Value value = DKV.get(key);
+        if (value == null || value.isNull()) continue;
+        if (value.isModel()) {
+          Keyed.remove(key, futures, true);
```
michalkurka: Why cascade=True? This is why need to extract the frames from the models in the first place. If cascade = false then we don't need to do https://github.com/h2oai/h2o-3/pull/4311/files#diff-6d64c422677c3739279b45f7e56d72abL24

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: This is an unfortunate leftover from the previous implementation, where this was needed, as we were iterating over all the keys and deleting whatever came our way.

I believe this can be avoided completely now. The whole `extractModelKeys` method.

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: I believe we don't need to do this at all now.

Pscheidl: It was assumed the model as a whole should be retained as functional. It is debatable whether the training frame should be part of it, but validation/xval models will be deleted as well if we don't extract them. I'm not sure whether users would expect such behavior.

On the other hand, they can always extract the keys themselves, making other layers responsible, giving other more control over the behavior of this method.

**I've decided removed this behavior for Models completely.**, as you propose. This means users will have to extract model's internal keys themselves, otherwise the internal sub-models won't survive.

Pscheidl: I demonstrated this behavior in `testCrossValidationModelsNotRetained` test.

Pscheidl:
Let's sum up.

1. Iterate over one key at a time, node after node.
1. Make sure DKV.get() is called per each key to get the current value.
1. Only `Model`s and `Frame`s are deleted. Models without cascade. Frames are checked for shared vecs - if a vec of a frame is shared with other vec, it is not retained. This behavior is tested.
1. Cross-validation models and any other "sub-models" of a `Model` are deleted, unless explicitly stated to be retained.

h2o-core/src/main/java/water/DKVManager.java
```diff
@@ -17,14 +17,17 @@
...
     final Set<Key> retainedSet = new HashSet<>(retainedKeys.length);
     retainedSet.addAll(Arrays.asList(retainedKeys));
     // Frames and models have multiple nested keys. Those must be extracted and kept from deletion as well.
     extractNestedKeys(retainedSet);
+    final Key[] allRetainedkeys = retainedSet.toArray(new Key[retainedSet.size()]);
 
-    new ClearDKVTask(retainedSet.toArray(new Key[retainedSet.size()]))
-            .doAllNodes();
+    for (final H2ONode node : H2O.CLOUD.members()) {
+      final NodeKeysRemovalTask nodeKeysRemovalTask = new NodeKeysRemovalTask(allRetainedkeys);
```
honzasterba: NodeKeysRemovalTask seems immutable, so you only need one instance I guess

h2o-core/src/main/java/water/DKVManager.java
Pscheidl: False !

Pscheidl:
Rebased on top of current master, re-running tests, as there was a failure due to cloud size (previously, the tests were fine).

Pscheidl:
Tests are good @michalkurka . We may merge. Waiting for your approval.

PR #4313, opened 2020-02-12
Proposal: Attempt at simplifying contract and logic of Hyperspace iterator

sebhrusen:
This is a simplification proposal due to the following observations after planning to add more iterator for AutoML:
1. Lack of clear separation of concerns between `HyperspaceWalker` and `HyperspaceIterator`, leading to more issues (see following).
2. `HyperspaceIterator` complexity is way too high: it should focus on iteration, i.e. providing new model parameters.
3. Our implementations of `HyperspaceIterator` modify the state of their parent walker, making it unsafe to create multiple iterators (somehow an heresy to the concept of iterators), hence the legacy introduction of (still heretic) `reset` method on iterator, when a consumer should just be able to create a new one safely.
4. Confusion between `GridSearch` and `Iterator` about who is responsible to handle termination logic (esp. based on number of models or on runtime): instead, leveraging job timing properties for that.

Additional suggestions
- break up this file: 
   1. walker+iterator interfaces in one file.
   2. base/abstract classes in one file.
   3. implementations in their dedicated file.
   4. walker factory in dedicated file.
 this generates backwards incompatibility only for consumers not using the `GridSearch.startGridSearch` static methods, and for those implementing their own walker: probably very few of them, we can help them using the migration guide.
- `hasNext` and `nextModelParameters` should take either NO argument, or the entire `Grid`: currently the `model` param is ignored in all our implementations, and if we want to add some decision logic based on training history, the next params should be picked based on the entire history, not on the last model built, especially when the grid is trained in parallel.

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
@@ -322,9 +322,7 @@ public abstract class ModelingStep<M extends Model> extends Iced<ModelingStep> {
             setCustomParams(baseParms);
 
             AutoMLBuildSpec buildSpec = aml().getBuildSpec();
-            RandomDiscreteValueSearchCriteria searchCriteria =
-                    (RandomDiscreteValueSearchCriteria) buildSpec.build_control.stopping_criteria.getSearchCriteria().clone();
-            searchCriteria.set_stopping_criteria((StoppingCriteria) searchCriteria.stopping_criteria().clone());
+            RandomDiscreteValueSearchCriteria searchCriteria = buildSpec.build_control.stopping_criteria.getSearchCriteria().deepClone();
 
             Work work = getAllocatedWork();
             double maxAssignedTimeSecs = aml().timeRemainingMs() * getWorkAllocations().remainingWorkRatio(work) / 1e3;
--- a/h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
@@ -322,7 +322,7 @@ public abstract class ModelingStep<M extends Model> extends Iced<ModelingStep> {
             setCustomParams(baseParms);
 
             AutoMLBuildSpec buildSpec = aml().getBuildSpec();
-            RandomDiscreteValueSearchCriteria searchCriteria = buildSpec.build_control.stopping_criteria.getSearchCriteria().deepClone();
+            RandomDiscreteValueSearchCriteria searchCriteria = (RandomDiscreteValueSearchCriteria)buildSpec.build_control.stopping_criteria.getSearchCriteria().clone();
 
             Work work = getAllocatedWork();
             double maxAssignedTimeSecs = aml().timeRemainingMs() * getWorkAllocations().remainingWorkRatio(work) / 1e3;
--- a/h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
@@ -12,7 +12,9 @@ import hex.ScoreKeeper.StoppingMetric;
 import hex.ensemble.StackedEnsembleModel;
 import hex.grid.Grid;
 import hex.grid.GridSearch;
+import hex.grid.HyperSpaceSearchCriteria;
 import hex.grid.HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria;
+import hex.grid.HyperSpaceSearchCriteria.StoppingCriteria;
 import water.Iced;
 import water.Job;
 import water.Key;
@@ -322,6 +324,8 @@ public abstract class ModelingStep<M extends Model> extends Iced<ModelingStep> {
             AutoMLBuildSpec buildSpec = aml().getBuildSpec();
             RandomDiscreteValueSearchCriteria searchCriteria =
                     (RandomDiscreteValueSearchCriteria) buildSpec.build_control.stopping_criteria.getSearchCriteria().clone();
+            searchCriteria.set_stopping_criteria((StoppingCriteria) searchCriteria.stopping_criteria().clone());
+
             Work work = getAllocatedWork();
             double maxAssignedTimeSecs = aml().timeRemainingMs() * getWorkAllocations().remainingWorkRatio(work) / 1e3;
             // predicate can be removed if/when we decide to include SEs in the max_models limit
--- a/h2o-core/src/main/java/hex/grid/GridSearch.java
+++ b/h2o-core/src/main/java/hex/grid/GridSearch.java
@@ -32,7 +32,7 @@ import java.util.concurrent.locks.ReentrantLock;
  * possible parameters, it launches a separated model building job. The launch of jobs is sequential
  * and blocking. So after finish the last model, whole grid search job is done as well.
  *
- * By default, the grid search invokes cartezian grid search, but it can be
+ * By default, the grid search invokes cartesian grid search, but it can be
  * modified by passing explicit hyper space walk strategy via the
  * {@link #startGridSearch(Key, HyperSpaceWalker)} method.
  *
@@ -106,8 +106,7 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
         throw new H2OIllegalArgumentException("training_frame", "grid", "Cannot append new models to a grid with different training input");
       grid.write_lock(_job);
     } else {
-      grid =
-              new Grid<>(_result,
+      grid = new Grid<>(_result,
                       _hyperSpaceWalker.getParams(),
                       _hyperSpaceWalker.getHyperParamNames(),
                       _hyperSpaceWalker.getParametersBuilderFactory().getFieldNamingStrategy());
@@ -118,8 +117,7 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
     HyperSpaceWalker.HyperSpaceIterator<MP> it = _hyperSpaceWalker.iterator();
     long gridWork=0;
     if (gridSize > 0) {//if total grid space is known, walk it all and count up models to be built (not subject to time-based or converge-based early stopping)
-      int count=0;
-      while (it.hasNext(model) && (it.max_models() > 0 && count++ < it.max_models())) { //only walk the first max_models models, if specified
+      while (it.hasNext(model)) {
         try {
           Model.Parameters parms = it.nextModelParameters(model);
           gridWork += (parms._nfolds > 0 ? (parms._nfolds+1/*main model*/) : 1) *parms.progressUnits();
@@ -153,7 +151,7 @@ public final class GridSearch<MP extends Model.Parameters> extends Keyed<GridSea
         Log.warn("GridSearch job "+_job._description+" completed with exception: "+ex);
         return true;
       }
-    }, gridWork, it.max_runtime_secs());
+    }, gridWork, maxRuntimeSecs());
   }
 
   /**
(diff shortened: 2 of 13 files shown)


h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -387,54 +366,38 @@ public boolean hasNext(Model previousModel) {
...
-        public double max_runtime_secs() { return Double.MAX_VALUE; }
-
-        public int max_models() { return _maxHyperSpaceSize > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int)_maxHyperSpaceSize; }
-
         @Override
-        public void modelFailed(Model failedModel) {
-          // nada
+        public void onModelFailure(Model failedModel, Consumer<Object[]> withFailedModelHyperParams) {
+          // FIXME: when using parallel grid search, there's no good reason to think that the current hyperparam indices where the ones used for the failed model
+          withFailedModelHyperParams.accept(hypers(_currentHyperparamIndices));
```
sebhrusen: here, suggesting to keep a map of {model_params_checksum -> hyperparam_indices} instead of `_currentHyperparamIndices`, the latter being relevant only for grids run sequentially.

sebhrusen:
@michalkurka : if you feel like looking at this proposal.

deil87:
@sebhrusen there is a PR which moves early stopping outside of the iterator https://github.com/h2oai/h2o-3/pull/4191 . Maybe we should merge it first.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -118,8 +117,7 @@ private GridSearch(Key<Grid> gkey, HyperSpaceWalker<MP, ?> hyperSpaceWalker, fin
     HyperSpaceWalker.HyperSpaceIterator<MP> it = _hyperSpaceWalker.iterator();
     long gridWork=0;
     if (gridSize > 0) {//if total grid space is known, walk it all and count up models to be built (not subject to time-based or converge-based early stopping)
-      int count=0;
-      while (it.hasNext(model) && (it.max_models() > 0 && count++ < it.max_models())) { //only walk the first max_models models, if specified
```
michalkurka: Please keep in mind grid search restarts, a similar thing was already discussed in https://github.com/h2oai/h2o-3/pull/4191#discussion_r370252892

Is the behavior the same as before?

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -373,9 +384,8 @@ private void gridSearch(Grid<MP> grid) {
         } catch (IllegalArgumentException e) {
           Log.warn("Grid search: construction of model parameters failed! Exception: ", e);
           // Model parameters cannot be constructed for some reason
-          it.modelFailed(model);
-          Object[] rawParams = it.getCurrentRawParameters();
-          grid.appendFailedModelParameters(model != null ? model._key : null, rawParams, e);
+          final Model failedModel = model;
+          it.onModelFailure(failedModel, failedHyperParams -> grid.appendFailedModelParameters(failedModel != null ? failedModel._key : null, failedHyperParams, e));
```
michalkurka: This seems unusual to me, imho it should be responsibility of GridSearch to append the failed paramters regardless of what the hyperspacewalker does.

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
```diff
@@ -322,6 +324,8 @@ protected Work getAllocatedWork() {
             AutoMLBuildSpec buildSpec = aml().getBuildSpec();
             RandomDiscreteValueSearchCriteria searchCriteria =
                     (RandomDiscreteValueSearchCriteria) buildSpec.build_control.stopping_criteria.getSearchCriteria().clone();
```
michalkurka: it seems like you want to "deep clone", function `set_stopping_criteria` seems unnecessary in the API IMHO

deil87:
> @deil87 I dunno, have the feeling that my changes are more fundamental as they get rid of the persistent iterating state in the walker (now in the iterator), which was probably the root of the issue on grid reruns.
> Now walker is attached to the grid instance, but the iterator object is guaranteed to be renewed for each run.
> If the goal of your PR was to support this, I'll try to run your tests in my branch and see.

My PR originated from this comment https://github.com/h2oai/h2o-3/pull/4040#discussion_r351502644 . This comment is for more context https://github.com/h2oai/h2o-3/pull/4040#issuecomment-559534499 . In order to enable FilterWalkers to be wrappable into each other we need early stopping logic to be outside of the iterator. 
I'm totally fine if your changes will also solve goal of my PR. If no, it will of course complicate things for my PR. Let's see how my tests will be doing in your branch.

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
sebhrusen: yes, that was my original intention, but unfortunately `clone` is final in `Iced` class, so I have to create an ad-hoc `deepClone` method only for `HyperSpaceSearchCriteria` which I don't find completely elegant either... anyway it's still probably preferable to this double clone, so switching to a deepClone impl. and removing the then unnecessary setter.

h2o-core/src/main/java/hex/grid/GridSearch.java
sebhrusen: it is still GridSearch that is doing the appending, through the lambda/callback.
The contract of `onModelFailure` just say that it should call the callback, without knowing what it does.
Actually, I believe this approach being cleaner and less broken that the previous.

Previously, there were 2 methods, supposedly independent: `getCurrentRawParameters` and `modelFailed`, the first one retrieving some params supposedly stored by the iterator, and the second one handling the failure.
However, the params that we need here are the exact ones used to build the failed model (current impl. failed to see this because of the dissociation between the 2 methods and it is therefore broken when running grid in parallel), so those 2 methods should fundamentally be merged into one, hence the callback.

h2o-core/src/main/java/hex/grid/GridSearch.java
sebhrusen: It is just a refactoring, I rely on existing tests to check that the legacy behaviour is not broken.
If we have expectations that are nowhere in tests, then they are barely features.

Will check though, but really, if we don't trust our tests, then we can't touch anything.
Even the expectations in the discussion linked above don't seem always clear to everyone.

sebhrusen: @michalkurka FYI the code example you linked above still produces 2 models for each grid, as expected.
I got confused by you naming this "grid restart" as I though you wanted some kind of `pause/continue` behaviour that we're not currently supporting afaik.

sebhrusen:
@michalkurka, @Pscheidl FYI, it's fully green now: maybe it's worth some consideration then: only planning the change for `master`.

Pscheidl:
I like the changes. As discussed personally in the office. I especially like the separation of concerns done.

h2o-core/src/main/java/hex/grid/HyperSpaceSearchCriteria.java
```diff
@@ -89,36 +104,38 @@ public static double default_stopping_tolerance_for_frame(Frame frame) {
...
+      _stopping_criteria._stopping_metric = stopping_metric;
     }
 
     public void set_stopping_tolerance(double stopping_tolerance) {
-      this._stopping_tolerance = stopping_tolerance;
+      _stopping_criteria._stopping_tolerance = stopping_tolerance;
+    }
+
+    @Override
+    public RandomDiscreteValueSearchCriteria deepClone() {
```
honzasterba: I dont like this method, i think since we have separate SearchCriteria class it should be immutable, in which case cloning is not necessary, rather than having copy-modifier method like withStoppingTolerance(value) 
also since we are refactoring this could we change the setters to camelCase?

sebhrusen: @honzasterba I'm not a big fan either to be honest (see https://github.com/h2oai/h2o-3/pull/4313#discussion_r380367954), but the only alternative is "constructor copy", e.g.
```
SearchCriteria(SearchCriteria original) {
   //loop through fields
}
```
AutoML (and a custom grid could do that as well), adaptively changes some of the search criteria, and for this we need to copy the original and change the new ones (for example `max_runtime_secs`) so immutability is not possible here... or it is if we simply create a constructor taking all the criteria... I'm open to this.
Note though that `H2O-3` makes heavy use of reflection, and therefore immutability is never 100% certain.

> also since we are refactoring this could we change the setters to camelCase?
Not here, this is part of public API.

honzasterba: a copy-constructor came to my mind as well, even that I like better than deepClone

sebhrusen: OK, some can be converted into camelCase indeed.

sebhrusen: I may have a solution to get rid of the deepClone as well, and make `StoppingCriteria` not completely immutable, but at least partially.

sebhrusen: @honzasterba removed, now cloning `StoppingCriteria` internally when setting the values in `RandomSearchCriteria`

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -333,12 +260,52 @@ protected int integerHash(int[] ar) {
         hashMe[i] = ar[i] * _hyperParams.get(_hyperParamNames[i]).length;
       return Arrays.deepHashCode(hashMe);
     }
+
+    private void validateParams() {
+      // if a parameter is specified in both model parameter and hyper-parameter, this is only allowed if the
+      // parameter value is set to be default.  Otherwise, an exception will be thrown.
```
sebhrusen: I think that this restriction is unnecessary and overly complicated.
Why don't we just say that hyper-parameters take precedence over model parameters. Then everything is simple, people can specify a param in both if they want, we'll just use the one in the hyperparams... anyway user can access the params from the trained model later, no cheating.
@michalkurka  any thoughts?

michalkurka: I think it is good to have this fail-safe mechanism. People make mistakes all the time and experiment with different parameter combinations. It is better to avoid surprises.

h2o-core/src/main/java/hex/grid/GridSearch.java
michalkurka: we are supporting grid pause & restart - see the linked discussion

sebhrusen: @michalkurka  I was really surprised by this, so tried and confirm that it resumes grids.
First thought that it was because the `walker` was keeping the state of the iterator and then reused in the `resume` run, but it doesn't seem to happen this way: on reruns, there's actually a new walker created each time, but when building model, there is an "optimization": grid search looks up in the DKV for a model with the same params, and if found, it adds it to the grid.
So, what I guess is happening (I tried with a seed to be sure it doesn't rebuild the same models multiple times), is that it is still iterating from the start, but when finds the previous model, it re-adds it to the grid with its old key.
See `GridSearch#buildModel` for details.

I'll try it with this branch to confirm, but if my hypothesis is correct, it should still work.

sebhrusen: Will also try to find an existing test for this and create one if needed.

michalkurka: Yes, it is designed to be restartable. It works by looking up the models and checking checksums on parameters and frames. This way it can identify they belong to the grid and avoid rebuilding them again.

michalkurka: (the issue we discussed with Andrey in the other PR is that max_models should be respected - it shouldn't exceed the original max_models specification)

sebhrusen: to test this, I created a method on Python to cancel jobs (needed this for an AutoML test): you just need to call it in a subprocess after x seconds, restart the grid, recancel, compare the grid state.

PR #4315, opened 2020-02-12
Google Cloud H2O-3 Cluster Launcher doc update

JHerna17:
Doc update for latest version of the H2O-3 Cluster Launcher on GCP


SherZhou:
Sorry, I cannot find H2O-3 Cluster in the market of Google Cloud Platform right now, is it still available on GCP?

PR #4316, opened 2020-02-13
[PUBDEV-7279] exact tree method not supported in multinode

honzasterba:
- add error message

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
@@ -215,7 +215,8 @@ public class XGBoost extends ModelBuilder<XGBoostModel,XGBoostModel.XGBoostParam
       error("_learn_rate", "learn_rate must be between 0 and 1");
     if( !(0. < _parms._col_sample_rate && _parms._col_sample_rate <= 1.0) )
       error("_col_sample_rate", "col_sample_rate must be between 0 and 1");
-    if (_parms._grow_policy== XGBoostModel.XGBoostParameters.GrowPolicy.lossguide && _parms._tree_method!= XGBoostModel.XGBoostParameters.TreeMethod.hist)
+    if (_parms._grow_policy== XGBoostModel.XGBoostParameters.GrowPolicy.lossguide && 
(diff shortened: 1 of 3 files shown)


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -226,6 +226,10 @@
       TreeUtils.checkMonotoneConstraints(this, _train, _parms._monotone_constraints);
     }
 
+    if ((_train != null) && (_parms._tree_method == XGBoostModel.XGBoostParameters.TreeMethod.exact) || H2O.CLOUD.size() > 1) {
+      error("_tree_method", "exact is not supported in distributed environment");
```
michalkurka: This should provide more details on why it is not supported (possibly a link to Technote, jira project TN)

If this is an XGBoost bug the code comment should have a link to the bug as well.

michalkurka:
DRF and GBM have an option `build_tree_one_node`: http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/algo-params/build_tree_one_node.html

We could add this option to XGBoost as well and thus allow to run a small XGBoost with exact even on distributed clusters. The easiest way is to rebalance the Frame and bring it to the memory of the local node before calling https://github.com/h2oai/h2o-3/blob/master/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java#L323, this could be done the same way as deeplearning does it (for a different purpose): https://github.com/h2oai/h2o-3/blob/master/h2o-algos/src/main/java/hex/deeplearning/DeepLearning.java#L179 - simply rebalance the (small) dataset to a single chunk.

It is not always practical for our users to start a single node cluster, we should imho provide a way to run them with `exact`. This will become more important once the parallel per-segment model building is implemented.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
honzasterba: this is not an xgboost bug, seems like a limitation of their implementation

https://github.com/dmlc/xgboost/issues/5304

honzasterba:
created a JIRA for build_tree_one_node https://0xdata.atlassian.net/browse/PUBDEV-7303

PR #4317, opened 2020-02-13
[PUBDEV-7294] install docs deps only for py3

honzasterba:


angela0xdata:
Thank you, @honzasterba. 
So when I re-merge the doc updates for master, I’ll also update the h2o-py/docs-requirements.txt file with the newer version of sphinx and the new extensions. 
:crossed_fingers:

honzasterba:
@angela0xdata yes please keep the docs deps only in the new docs-requirements.txt

PR #4318, opened 2020-02-13
[PUBDEV-7292] add check to verify xgboost respect monotonicity constr…

honzasterba:
…aints

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
@@ -444,7 +444,12 @@ public class XGBoost extends ModelBuilder<XGBoostModel,XGBoostModel.XGBoostParam
       }
 
       Map<String, Integer> monotoneConstraints = _parms.monotoneConstraints();
-      if ((!monotoneConstraints.isEmpty()) && constraintCheckEnabled()) {
+      if (!monotoneConstraints.isEmpty() &&
+          _parms._booster != XGBoostModel.XGBoostParameters.Booster.gblinear &&
+          constraintCheckEnabled()
+      ) {
+        _job.update(0, "Checking monotonicity constraints on the final model");
+        boosterProvider.updateBooster();
         checkConstraints(model.model_info(), monotoneConstraints);
       }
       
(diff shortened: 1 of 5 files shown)


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -438,13 +442,94 @@ private void scoreAndBuildTrees(final XGBoostSetupTask setupTask, final BoosterP
           break;
         }
       }
+
+      Map<String, Integer> monotoneConstraints = _parms.monotoneConstraints();
+      if ((!monotoneConstraints.isEmpty()) && constraintCheckEnabled()) {
+        checkConstraints(model.model_info(), monotoneConstraints);
```
michalkurka: It would be nice and reassuring to have a Log message that says that the constraints were successfully checked.

michalkurka: Don't you need to use the `boosterProvider` to get the booster in this phase of training?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -438,13 +442,94 @@ private void scoreAndBuildTrees(final XGBoostSetupTask setupTask, final BoosterP
...
       _job.update(_parms._ntrees-model._output._ntrees);
     }
 
+    private boolean constraintCheckEnabled() {
+      return Boolean.parseBoolean(getSysProperty("xgboost.monotonicity.checkEnabled", "true"));
+    }
+
+    private void checkConstraints(XGBoostModelInfo model_info, Map<String, Integer> monotoneConstraints) {
+      GradBooster booster = XGBoostJavaMojoModel.makePredictor(model_info._boosterBytes).getBooster();
+      if (!(booster instanceof GBTree)) {
```
michalkurka: This is not ideal because the check can stop working over the time, instead
- we should not invoke monotonicity check for `gblinear`,
- change this into IllegalStateException.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
honzasterba: added _job.update

honzasterba: true, added call to get the latest booster

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
honzasterba: fixed

PR #4319, opened 2020-02-14
PUBDEV-7138 Extended Isolation Forest

valenad1:
Implementation of Extended Isolation Forest algorithm.

https://0xdata.atlassian.net/browse/PUBDEV-7138

The production version of product will be finished here: https://github.com/h2oai/h2o-3/pull/5246


h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,113 @@
...
+            for (double [] row : task.right) {
+                System.out.println(Arrays.toString(row));
+            }
+        }
+    }
+
+    private class SplitFrameMRTask extends MRTask<SplitFrameMRTask> {
+        private Vec mul;
+
+        private double [][] left;
```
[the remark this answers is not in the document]
valenad1: A different approach will be used.

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForestParameters.java
```diff
@@ -0,0 +1,23 @@
...
+    public String fullName() {
+        return "Extended Isolation Forest";
+    }
+
+    @Override
+    public String javaName() {
+        return ExtendedIsolationForestModel.class.getName();
+    }
+    
+    public int _extension_level;
```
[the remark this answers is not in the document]
valenad1: Fixed

valenad1:
I refactor vector creation as we talk about except uniformDistrFromFrameMR where I need to use reduce method.

maurever:
Please, add a description to this PR and link to the JIRA ticket. Thanks!

h2o-algos/src/main/java/hex/psvm/psvm/MatrixUtils.java
```diff
@@ -95,4 +140,41 @@ public void reduce(ProductMtvTask mrt) {
     }
   }
 
+  static class SubMtvTask extends MRTask<SubMtvTask> {
```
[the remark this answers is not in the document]
valenad1: Yes, it's better

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,151 @@
...
+    public ModelCategory[] can_build() {
+        return new ModelCategory[0];
+    }
+
+    @Override
+    public boolean isSupervised() {
+        return false;
+    }
+
+    private static class PrintRowsMRTask extends MRTask<PrintRowsMRTask> {
```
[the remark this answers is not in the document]
valenad1: Only for logging, it won't be part of final PR.

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,151 @@
...
+            // create first split
+            Vec p = VecUtils.uniformDistrFromFrameMR(tr, _parms._seed);
+            Vec n = VecUtils.makeGaussianVec(tr.numCols(), tr.numCols() - _parms.extensionLevel - 1, _parms._seed);
+            System.out.println(Arrays.toString(FrameUtils.asDoubles(n)));
+            System.out.println(n.length());
+            Frame res = MatrixUtils.subtractionMtv(tr, p);
+            Vec mul = MatrixUtils.productMtv2(res, n);
+            System.out.println(Arrays.toString(FrameUtils.asDoubles(mul)));
+            Frame left = new FilterFrameLeftTask(mul).doAll(tr.types(), tr).outputFrame();
+            Frame right = new FilterFrameRightTask(mul).doAll(tr.types(), tr).outputFrame();
```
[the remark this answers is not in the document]
valenad1: Please see for example: 6ae1b0f9f4

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        @Override
+        public void map(Chunk[] cs, NewChunk[] ncs) {
+            for (int row = 0; row < cs[0]._len; row++) {
+                double num = mul.at(cs[0].start() + row);
+                if (num <= 0) {
+                    // left
+                    for (int column = 0; column < cs.length; column++) {
+                        ncs[column].addNum(cs[column].atd(row));
```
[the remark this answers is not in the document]
valenad1: Good point, I have to think of it and ask about the details.

h2o-algos/src/test/java/hex/tree/isoforextended/ExtendedIsolationForestTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+            ExtendedIsolationForestParameters p = new ExtendedIsolationForestParameters();
+            p._train = train._key;
+            p._seed = 0xDECAF;
+            p._ntrees = 7;
+            p.extensionLevel = train.numCols() - 1;
+            p._min_rows = 1;
+//            p._sample_size = 5;
+
+            ExtendedIsolationForestModel model = new ExtendedIsolationForest(p).trainModel().get();
```
maurever: You need to track the model in this scope not to leak keys. 

`Scope.track_generic(model);`

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,321 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    private class Node {
+        private Frame frame;
```
maurever: Try to avoid saving the whole `Frame.` Save the key of the frame instead. When you need to use the frame, call something like `DKV.get(key)`.

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,321 @@
...
+                if (num > value) {
+                    for (int column = 0; column < cs.length; column++) {
+                        ncs[column].addNum(cs[column].atd(row));
+                    }
+                }
+            }
+        }
+    }
+
+    private class Node {
```
maurever: The `Node` should extend `Iced`.

valenad1: Done

h2o-algos/src/test/java/hex/tree/isoforextended/ExtendedIsolationForestTest.java
valenad1: Done

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
valenad1: Done

h2o-algos/src/test/java/hex/psvm/psvm/MatrixUtilsTest.java
```diff
@@ -41,4 +41,361 @@ public void productMtDM() {
...
+                    .withDataForCol(0, ard(1.0, 1.0, 1.0))
+                    .withDataForCol(1, ard(1.0, 1.0, 1.0))
+                    .build();
+            Scope.track(m);
+            Vec v = Vec.makeVec(ard(1.0, 1.0), Vec.newKey());
+
+            Frame res = MatrixUtils.subtractionMtv(m, v);
+
+            // check result
+            Assert.assertEquals(2, res.vecs().length);
```
maurever, line 59: Please, add a short message to all asserts.

h2o-algos/src/test/java/hex/tree/isoforextended/ExtendedIsolationForestTest.java
```diff
@@ -0,0 +1,264 @@
...
+            p._ntrees = 100;
+            p.extension_level = train.numCols() - 1;
+
+            ExtendedIsolationForest eif = new ExtendedIsolationForest(p);
+            ExtendedIsolationForestModel model = eif.trainModel().get();
+            Scope.track_generic(model);
+            assertNotNull(model);
+
+            Frame test = Scope.track(parse_test_file("smalldata/anomaly/single_blob.csv"));
+            Frame out = model.score(test);
```
maurever, line 43: Do you expect some specific output to test it? For example, in comparison to the implementation of the paper?

h2o-algos/src/test/java/hex/tree/isoforextended/ExtendedIsolationForestTest.java
```diff
@@ -0,0 +1,264 @@
...
+        assertEquals(11.583643521303037,
+                ExtendedIsolationForest.IsolationTree.averagePathLengthOfUnsuccesfullSearch(500), 1e-5);
+        assertEquals(1, ExtendedIsolationForest.IsolationTree.averagePathLengthOfUnsuccesfullSearch(2), 1e-5);
+        assertEquals(0, ExtendedIsolationForest.IsolationTree.averagePathLengthOfUnsuccesfullSearch(1), 1e-5);
+        assertEquals(0, ExtendedIsolationForest.IsolationTree.averagePathLengthOfUnsuccesfullSearch(0), 1e-5);
+        assertEquals(0, ExtendedIsolationForest.IsolationTree.averagePathLengthOfUnsuccesfullSearch(-1), 1e-5);
+    }
+
+    @Test
+    public void FilterLtTaskTest() {
```
maurever, line 176:
```suggestion
    public void testFilterLtTask() {
```

h2o-algos/src/test/java/hex/tree/isoforextended/ExtendedIsolationForestTest.java
```diff
@@ -0,0 +1,264 @@
...
+            assertEquals(1, resVec.numRows());
+            assertEquals(4, resVec.numCols());
+            DKV.remove(resVec._key);
+        } finally {
+            Scope.exit();
+        }
+    }
+
+    @Test
+    public void FilterLtTaskCategoricalDataTest() {
```
maurever, line 199:
```suggestion
    public void testFilterLtTaskCategoricalData() {
```

h2o-algos/src/test/java/hex/tree/isoforextended/ExtendedIsolationForestTest.java
```diff
@@ -0,0 +1,264 @@
...
+            assertEquals(3, resVec.numRows());
+            assertEquals(3, resVec.numCols());
+            DKV.remove(resVec._key);
+        } finally {
+            Scope.exit();
+        }
+    }
+
+    @Test
+    public void FilterGteTaskTest() {
```
maurever, line 221:
```suggestion
    public void testFilterGteTask() {
```

h2o-algos/src/test/java/hex/tree/isoforextended/ExtendedIsolationForestTest.java
```diff
@@ -0,0 +1,264 @@
...
+            assertEquals(1, resVec.numRows());
+            assertEquals(4, resVec.numCols());
+            DKV.remove(resVec._key);
+        } finally {
+            Scope.exit();
+        }
+    }
+
+    @Test
+    public void FilterGteTaskCategoricalDataTest() {
```
maurever, line 244:
```suggestion
    public void testFilterGteTaskCategoricalData() {
```

h2o-genmodel/src/main/java/hex/genmodel/algos/isoforextended/ExtendedIsolationForestMojoModel.java
```diff
@@ -0,0 +1,26 @@
...
+
+public final class ExtendedIsolationForestMojoModel extends SharedTreeMojoModel {
+
+  public ExtendedIsolationForestMojoModel(String[] columns, String[][] domains, String responseColumn) {
+    super(columns, domains, responseColumn);
+  }
+
+  @Override
+  public double[] unifyPreds(double[] row, double offset, double[] preds) {
+    return new double[0];
```
maurever, line 13: Should be correctly implemented.

h2o-py/tests/testdir_algos/isoforextended/pyunit_isoforextended_saveload.py
```diff
@@ -0,0 +1,34 @@
...
+    eif_model = H2OExtendedIsolationForestEstimator(ntrees=7, seed=12, sample_size=5)
+    eif_model.train(training_frame=train)
+
+    path = pyunit_utils.locate("results")
+
+    assert os.path.isdir(path), "Expected save directory {0} to exist, but it does not.".format(path)
+    model_path = h2o.save_model(eif_model, path=path, force=True)
+
+    assert os.path.isfile(model_path), "Expected load file {0} to exist, but it does not.".format(model_path)
+    reloaded = h2o.load_model(model_path)
```
maurever, line 23: Try to predict before saving a model and then compare the prediction after reloading the model.

h2o-py/tests/testdir_algos/isoforextended/pyunit_isoforextended_smoke.py
```diff
@@ -0,0 +1,22 @@
...
+
+def extended_isolation_forest():
+    print("Extended Isolation Forest Smoke Test")
+
+    train = h2o.import_file(pyunit_utils.locate("smalldata/anomaly/single_blob.csv"))
+
+    eif_model = H2OExtendedIsolationForestEstimator(ntrees=7, seed=12, sample_size=5)
+    eif_model.train(training_frame=train)
+
+    print(eif_model)
```
maurever, line 17: Make a prediction and compare it for example with some values from the source paper.

h2o-r/tests/testdir_algos/isoforextended/runit_isoforextended_smoke.R
```diff
@@ -0,0 +1,19 @@
...
+    set.seed(1234)
+    N = 1e4
+    random_data <- data.frame(
+        x = c(rnorm(N, 0, 0.5), rnorm(N*0.05, -1.5, 1)),
+        y = c(rnorm(N, 0, 0.5), rnorm(N*0.05,  1.5, 1))
+    )
+    random_data.hex <- as.h2o(random_data)
+    
+    exisofor.model <- h2o.extendedIsolationForest(training_frame = random_data.hex)
+    expect_equal(as.character(class(exisofor.model)), "H2OAnomalyDetectionModel")
```
maurever, line 16: The same comment as in the Python test, try to compare result values from the `predict` method.

h2o-algos/src/main/java/hex/psvm/psvm/MatrixUtils.java
```diff
@@ -30,15 +32,111 @@ static LLMatrix productMtDM(Frame m, Vec diagonal) {
...
    */
-  static double[] productMtv(Frame m, Vec v) {
+  public static double[] productMtv(Frame m, Vec v) {
     Vec[] vecs = ArrayUtils.append(m.vecs(), v);
     return new ProductMtvTask().doAll(vecs)._result;
   }
 
+  /**
+   * Previous Mtv method is for v of dimension (m x 1) here it is for (n x 1)
+   * <p>
```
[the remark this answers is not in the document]
valenad1: It was generated by Idea.
It mens that in the output will be two separate paragraphs. 
https://stackoverflow.com/questions/5260368/which-tag-should-be-used-as-paragraph-separator-in-javadoc

h2o-algos/src/main/java/hex/psvm/psvm/MatrixUtils.java
```diff
@@ -30,15 +32,111 @@ static LLMatrix productMtDM(Frame m, Vec diagonal) {
...
+   * Calculates matrix-vector product M'v
+   *
+   * @param m Frame representing matrix M (m x n)
+   * @param v Vec representing vector v (n x 1)
+   * @return m-element array representing the result of the product
+   */
+  public static Vec productMtv2(Frame m, Vec v) {
```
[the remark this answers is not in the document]
valenad1: I replaced 2 with "Math" because the difference to previous is that this method presumes Math correct matrix-vector dimmensions

h2o-algos/src/main/java/hex/psvm/psvm/MatrixUtils.java
```diff
@@ -30,15 +32,111 @@ static LLMatrix productMtDM(Frame m, Vec diagonal) {
...
+    if (v.length() != m.numCols()) {
+      throw new UnsupportedOperationException("Vector elements number must be the same as matrix column number");
+    }
+    Vec result = new ProductMtvTask2(v).doAll(Vec.T_NUM, m).outputFrame().vecs()[0];
+    return result;
+  }
+
+  /**
+   * Same as productMtv2 but here is array.
+   * <p>
```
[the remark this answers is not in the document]
valenad1: Was removed

h2o-algos/src/test/java/hex/tree/isoforextended/ExtendedIsolationForestTest.java
valenad1: No, I can test the output values itself but it depends on seed and branching implementation.

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,298 @@
...
+        }
+
+        @Override
+        protected void initializeModelSpecifics() {
+            iTrees = new IsolationTree[_parms._ntrees];
+        }
+
+    }
+
+    public static class IsolationTree extends Iced<IsolationTree> {
```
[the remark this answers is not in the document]
valenad1: Done

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,298 @@
...
+//                    + extensionLevelMax + "] but it is " + _parms.extension_level);
+//        }
+//        long sampleSizeMax = _parms.train().numRows();
+//        if (_parms._sample_size < 0 || _parms._sample_size > sampleSizeMax) {
+//            throw new IllegalStateException("Parameter sample_size must be in interval [0, "
+//                    + sampleSizeMax + "] but it is " + _parms._sample_size);
+//        }        
+//    }                
+
+    @Override
```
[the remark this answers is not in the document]
valenad1: not used in the current implementation of Extended Isolation Forest

h2o-genmodel/src/main/java/hex/genmodel/algos/isoforextended/ExtendedIsolationForestMojoModel.java
valenad1: It is not used in the current implementation of Extended Isolation Forest

h2o-algos/src/main/java/hex/psvm/psvm/MatrixUtils.java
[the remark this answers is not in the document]
maurever: We don't use it anywhere, but why not. It is up to you. :)

h2o-genmodel/src/main/java/hex/genmodel/algos/isoforextended/ExtendedIsolationForestMojoModel.java
maurever: So then you should return `H2O.unimpl()`, or something like that.

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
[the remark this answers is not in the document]
maurever: So then you should return `H2O.unimpl()`, or something like that.

h2o-core/src/main/java/water/util/ArrayUtils.java
```diff
@@ -219,7 +276,12 @@ public static double l2norm(double [] x, boolean skipLast){
 
   public static double[] add(double[] a, double[] b) {
     if( a==null ) return b;
-    for(int i = 0; i < a.length; i++ ) a[i] += b[i];
```
honzasterba, line 222: this changes the contract of the method, since the original only modified a now it can modify b as well and deal with different length arrays

h2o-py/docs/modeling.rst
```diff
@@ -67,12 +67,6 @@ Supervised
     :show-inheritance:
     :members:
 
-:mod:`H2OXGBoostEstimator`
```
honzasterba, line 70: why is this removed?

maurever: > smalldata is not kept in git, we can add these to our s3 storage

The dataset was already added and the Jenkins job is running, thanks to @Pscheidl .

h2o-algos/src/main/java/hex/psvm/psvm/MatrixUtils.java
```diff
@@ -95,4 +197,115 @@ public void reduce(ProductMtvTask mrt) {
...
 
+  /**
+   * Task for Matrix-Vector subtraction.
+   */
+  static class SubtractionMtvTask extends MRTask<SubtractionMtvTask> {
+
+    private Vec v;
+
+    /**
+     * @param v vector size is the same as number of matrix columns
```
michalkurka, line 208: I think it might be more efficient then to either use double array instead of a Vec or avoid using Vec#at which is expensive (requires a lookup of chunk every time).

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,324 @@
...
+import water.util.ArrayUtils;
+import water.util.MathUtils;
+import water.util.VecUtils;
+
+/**
+ * Extended isolation forest implementation. Algorithm comes from https://arxiv.org/pdf/1811.02141.pdf paper.
+ *
+ * @author Adam Valenta
+ */
+public class ExtendedIsolationForest extends SharedTree<ExtendedIsolationForestModel,
```
michalkurka, line 20: what capabilities do we get from SharedTree?

h2o-r/tests/testdir_algos/isoforextended/runit_isoforextended_smoke.R
valenad1: Done

maurever:
POC is working. The next steps are:
- improve parallelization of the algorithm based on `k` parameter
- test this improvement of the performance against H2O Isolation Forest implementation
- add document page

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForestModel.java
```diff
@@ -0,0 +1,137 @@
...
+    protected double[] score0(double[] data, double[] preds, double offset, int ntrees) {
+        super.score0(data, preds, offset, ntrees);
+        if (ntrees >= 1) preds[1] = preds[0] / ntrees;
+        
+        // compute score for given point
+        double pathLength = 0;
+        for (ExtendedIsolationForest.IsolationTree iTree : _output.iTrees) {
+            double iTreeScore = iTree.computePathLength(data);
+            pathLength += iTreeScore;
+//            System.out.println("iTreeScore " + iTreeScore);
```
maurever, line 52: Dont forget to remove these print lines or you can use Log.debug instead.

h2o-core/src/main/java/water/util/ArrayUtils.java
maurever: @valenad1 Is this resolved?

h2o-docs/src/product/data-science/eif.rst
```diff
@@ -0,0 +1,122 @@
...
+  - ``one_hot_explicit`` or ``OneHotExplicit``: N+1 new columns for categorical features with N levels
+  - ``binary`` or ``Binary``: No more than 32 columns per categorical feature
+  - ``eigen`` or ``Eigen``: *k* columns per categorical feature, keeping projections of one-hot-encoded matrix onto *k*-dim eigen space only
+  - ``label_encoder`` or ``LabelEncoder``:  Convert every enum into the integer of its index (for example, level 0 -> 0, level 1 -> 1, etc.)
+  - ``sort_by_response`` or ``SortByResponse``: Reorders the levels by the mean response (for example, the level with lowest response -> 0, the level with second-lowest response -> 1, etc.). This is useful in GBM/DRF, for example, when you have more levels than ``nbins_cats``, and where the top level splits now have a chance at separating the data with a split. Note that this requires a specified response column.
+
+- **extension_level**: (Required) The number in range :math:`[0, P-1]`; where P is the number of features. The minimum value of the hyperparameter is 0, which corresponds to Isolation Forest behavior. The maximum is :math:`P-1` and stands for a full extension.
```
maurever, line 70: Could you explain the extension level in more detail, please? Why it could be useful etc. You should also create a new doc page for this parameter and refer to it here.

h2o-docs/src/product/data-science/eif.rst
```diff
@@ -0,0 +1,122 @@
...
+
+.. math::
+    (x - p) � n < 0
+
+Tutorials and Blogs
+~~~~~~~~~~~~~~~~~~~
+
+The following tutorials are available that describe how to use Extended Isolation Forest. 
+
+- todo link to notebooks? Thesis? External Sources?
```
maurever, line 36: You can add links to IPython notebooks here and the link to the thesis too. Or if you find some interesting blog posts about EIF, you can also leave the links here.

h2o-genmodel/src/main/java/hex/genmodel/algos/isoforextended/ExtendedIsolationForestMojoModel.java
maurever: The same question here, @michalkurka, what is your opinion? :)

h2o-py/docs/modeling.rst
maurever: Maybe some accident? @valenad1, can you resolve this?

h2o-py/h2o/estimators/extended_isolation_forest.py
```diff
@@ -0,0 +1,558 @@
...
+        return self._parms.get("sample_size")
+
+    @sample_size.setter
+    def sample_size(self, sample_size):
+        assert_is_type(sample_size, None, int)
+        self._parms["sample_size"] = sample_size
+
+
+    @property
+    def extension_level(self):
```
maurever, line 348: Add an example here, please. :)

h2o-py/tests/testdir_algos/isoforextended/pyunit_isoforextended_smoke.py
```diff
@@ -0,0 +1,31 @@
...
+    train = h2o.import_file(pyunit_utils.locate("smalldata/anomaly/single_blob.csv"))
+
+    eif_model = H2OExtendedIsolationForestEstimator(ntrees=7, seed=12, sample_size=5)
+    eif_model.train(training_frame=train)
+    anomaly_score = eif_model.predict(train)
+    anomaly = anomaly_score['anomaly_score'].as_data_frame(use_pandas=True)["anomaly_score"]
+
+    print(eif_model)
+    
+    assert (abs(anomaly[0] - 0.54)) <= 0.01, "Not expected output: Expected value is about 0.54"
```
maurever, line 21: Leave comment with link to the source paper here, for others to know where the values come from.

maurever (changes requested):
Looks good! Minor changes are requested.

h2o-core/src/main/java/water/util/ArrayUtils.java
valenad1: I have implemented new addSave method

maurever: Cool, thanks!

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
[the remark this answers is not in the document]
michalkurka: is this method ever called? from `Score`? it would be safer to throw an exception here

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,378 @@
...
+        @Override
+        protected void scoreAndBuildTrees(boolean oob) {
+            IsolationTreeForkJoinTask [] iTreeTasks = new IsolationTreeForkJoinTask[_parms._ntrees];
+            for (int t = 0; t < _parms._ntrees; t++) {
+                iTreeTasks[t] = new IsolationTreeForkJoinTask();
+                H2O.submitTask(iTreeTasks[t]);
+            }
+            for (int t = 0; t < _parms._ntrees; t++) {
+                _iTrees[t] = iTreeTasks[t].getResult();
+                _job.update(1);
```
michalkurka, line 137: This will work in most cases but it is not 100% correct. If for whatever reason the first tree takes much longer than others the progress bar will not be smooth, it will have jumps. It would be better if the IsolationTreeForkJoinTask updated the job progress as the tree is build.

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,378 @@
...
+        protected boolean doOOBScoring() {
+            return true;
+        }
+
+        @Override
+        protected void scoreAndBuildTrees(boolean oob) {
+            IsolationTreeForkJoinTask [] iTreeTasks = new IsolationTreeForkJoinTask[_parms._ntrees];
+            for (int t = 0; t < _parms._ntrees; t++) {
+                iTreeTasks[t] = new IsolationTreeForkJoinTask();
+                H2O.submitTask(iTreeTasks[t]);
```
michalkurka, line 133: What is going to happen when we have eg. 10k trees? Do we need to think about this?

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
michalkurka: this question is even more relevant now since scoreAndBuildTrees is currently not used

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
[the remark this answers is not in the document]
valenad1: I am not quite sure what is the real purpose of this method but I am sure that I am not calling it in my Scoring.

I will throw the H2O.unimpl()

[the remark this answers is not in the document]
valenad1: I think the method is used to some "Column relevance" estimation in a supervised version of SharedTree algorithm.

h2o-algos/src/main/java/hex/schemas/ExtendedIsolationForestModelV3.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class ExtendedIsolationForestModelV3 extends SharedTreeModelV3<ExtendedIsolationForestModel,
+        ExtendedIsolationForestModelV3,
+        ExtendedIsolationForestModel.ExtendedIsolationForestParameters,
+        ExtendedIsolationForestV3.ExtendedIsolationForestParametersV3,
+        ExtendedIsolationForestModel.ExtendedIsolationForestOutput,
+        ExtendedIsolationForestModelV3.ExtendedIsolationForestModelOutputV3> {
+
+  public static final class ExtendedIsolationForestModelOutputV3 extends SharedTreeModelV3.SharedTreeModelOutputV3<ExtendedIsolationForestModel.ExtendedIsolationForestOutput, ExtendedIsolationForestModelOutputV3> {
+      // nothing
```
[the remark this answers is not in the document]
valenad1: Original Isolation Forest has a `public TwoDimTableV3 variable_splits;` as an output. I don't have the splits prepared the same way. I think there is nothing to provide for users so far...

h2o-genmodel/src/main/java/hex/genmodel/algos/isoforextended/ExtendedIsolationForestMojoModel.java
valenad1: @michalkurka Maybe the purpose of this method is convert anomaly_score into Anomaly/NotAnomaly prediction?

h2o-docs/src/product/data-science/eif.rst
valenad1: I added a link to the thesis and notebook about IF and EIF comparison by the authors of the algorithm. I am not sure if the notebooks [here](https://github.com/valenad1/h2o-3/tree/PUBDEV-7138-extended-isolation-forest/h2o-py/demos/extisofor) are good enough to be in the documentation.

h2o-py/docs/modeling.rst
valenad1: It is weird, when I build the master branch from 21.8 then the H2OXGBoostEstimator is again removed from the file and from h2o-py/h2o/estimators/__init__.py. Because the files are generated I told my self that maybe someone forgot to commit them. If you say that H2OXGBoostEstimator should be in the file then it is probably something wrong with my dev environment.

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
valenad1: I don't know how exactly the tasks are distributed around the cluster. No dought that the clever way would be to generate ntrees/numberOfClusters trees in each task. If I am right then the clusters can have different computing power so the logic here can be better than it is now but I don't know how big are the consequences to leaving it as it is.

h2o-docs/src/product/data-science/eif.rst
valenad1: I added a whole page about extension_level and here add only "As the extension_level is increased the bias of standard Isolation Forest is reduced."

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
valenad1: The major reason is to be able to follow Isolation Forest as a referential implementation and lack of knowledge of the codebase. Therefore I used the same parent class. Currently, I am not using any of the capabilities besides the interface that allows me to save, load model, and generate interfaces the same way as it is for Isolation Forest. Do you think it would be doable and better to extend Model directly?

valenad1: The major reason is to be able to follow Isolation Forest as a referential implementation and lack of knowledge of the codebase. Therefore I used the same parent class. Currently, I am not using any of the capabilities besides the interface that allows me to save, load model, and generate interfaces the same way as it is for Isolation Forest. Do you think it would be doable and better to extend Model directly?

maurever (approved):

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,377 @@
...
+        return false;
+    }
+
+    @Override
+    public boolean havePojo() {
+        return false;
+    }
+
+    @Override
+    public boolean haveMojo() {
```
honzasterba, line 104: since MOJO was not really implemented I think its better to return false here and remove the empty MOJOWriter class

valenad1:
Link: https://github.com/valenad1/h2o-3/blob/PUBDEV-7138-extended-isolation-forest-leak-fix/h2o-py/demos/extisofor/extended-isolation-forest-benchmark.ipynb

## Extended Isolation Forest Benchmark - Training stage

In the tests, H2O cloud is initialized before each run and shutdown each every run. The algorithm is firstly tested on Training performance and compared to the Isolation Forest run time.

* IF = Isolation Forest
* EIF = Extended Isolation Forest
* N = number of rows
* P = number of columns
* ntrees = number of trees to train (always 100)
* sample_size = how many rows will be used to build a tree
* max_depth = only for IF, how big is the depth of the tree, in EIF is always set on math.ceil(math.log(sample_size, 2)) and max_depth is always depended on sample_size in the benchmark.

#### Computer parameters:
 * Lenovo ThinkPad P53,
 * MS Windows 10 Pro x64,
 * Intel Core i7-9850H CPU @ 2.60GHz,
 * 6 cores and 12 threads,
 * 96.0 GB RAM.
 
### Training process:

eif-parallel

The algorithm runtime performance depends on _ntrees_ and _sample_size_ parameters. Parameter _ntrees_ is fixed on 100 for all runs. Setting of _sample_size_ and the amount of data is varying in runs. I focus to put values that switch ON/OFF the MRTask.

## Results

### 1.1 Toy data
### 1.1.1 Small data and small dimension

h2o-scale-perf_256_500_2_0 5679054737091065

### 1.1.2 Small data and high dimension

h2o-scale-perf_256_5000_30_1 6009737014770509

### 1.1.3 Big data - small dimension, small sample_size

h2o-scale-perf_256_1500000_37 48055565357208

### 1.1.4 Big data - high dimension, small sample_size

h2o-scale-perf_256_100000_11 623167252540588

### 1.1.5 Big data - small dimension, big sample_size

h2o-scale-perf_15000_1500000_57 57910377979279

### 1.1.6 Big data - high dimension, big sample_size

* N = 100_000
* P = 30
* sample_size = 10_000

Finish with error (More information in conclusion)

`H2OConnectionError: Unexpected HTTP error: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))`

### 1.2 Real Credit Card Fraud Detection Data

### 1.2.1 Test with samle_size = 256 (default value)

h2o-scale-perf_256_284807_30_26 633357524871826

### 1.2.2 Real data - sample_size = 1% (2 848 rows) of data

Finish with error (More information in conclusion)

`H2OConnectionError: Unexpected HTTP error: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))`


## Conclusion
 * The biggest problem I see across various runs is the sensitivity on _P_ parameter no matter what the _sample_size_ setting is. EIF is significantly slower than IF even for small data with 500 rows (section 1.1.2). If N grows and sample_size is small, then EIF finishes slower (sections 1.1.4, 1.2.1). The real problem comes with large data with big sample_size where EIF is not able to finish most probably because of memory requirements(sections 1.1.6, 1.2.2, 1.2.3). When I test it in Java, all trees are started to be prepared at once and save their Nodes and temporary Frames which leads to following warning message. With the same settings, the IF finish without any problem with good performance.
 
`WARN water.default: Unblock allocations; cache below desired, but also OOM: OOM, (K/V:139,7 MB + POJO:19,27 GB + FREE:1,87 GB == MEM_MAX:21,27 GB), desiredKV=2,66 GB OOM!`

 * If _P_ parameter is small and sample_size is small, then EIF performs equally or better than IF.
 * If _P_ parameter is small and sample_size is big, then EIF performs equally to IF.

Solution attempt:
 
I have tried to solve it with better distribution IsolationTree building. I hardcode to build 144 trees (12 on each thread) but with no difference in run time. Even if it works, I don't think it would be usable for a cluster where the performance of each node can and most probably will vary. That is why I only submitting tasks to the H2O cluster (one task per one tree). Trees are operating with their own sub-sample of data and can be built totally independently. But build all trees at one time seems to be memory and time-consuming.

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        private int rightChildIndex(int i) {
+            return 2 * i + 2;
+        }        
+
+        /**
+         * Helper method. Print nodes size of the tree.
+         */
+        public void print() {
```
honzasterba, line 238: would rename to logNodes() and logLeight()

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForestMojoWriter.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.io.IOException;
+
+import static hex.tree.isofor.IsolationForestModel.IsolationForestOutput;
+import static hex.tree.isofor.IsolationForestModel.IsolationForestParameters;
+
+/**
+ * Mojo definition for Extended Isolation Forest model.
+ */
+public class ExtendedIsolationForestMojoWriter extends SharedTreeMojoWriter<ExtendedIsolationForestModel, ExtendedIsolationForestModel.ExtendedIsolationForestParameters, ExtendedIsolationForestModel.ExtendedIsolationForestOutput> {
```
honzasterba, line 14: no reason to commit this NOOP class until MOJO is really implemented

h2o-algos/src/main/java/hex/tree/isoforextended/SubSampleTask.java
```diff
@@ -0,0 +1,37 @@
...
+import water.fvec.Chunk;
+import water.fvec.NewChunk;
+import water.util.RandomUtils;
+
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Task to create random sub-sample of given Frame with defined subSampleSize
+ * TODO avalenta make subsampling randomized and guarantee given size for small datasets
```
honzasterba, line 13: do you plan on fixing this?

h2o-core/src/main/java/water/util/MathUtils.java
```diff
@@ -773,4 +773,14 @@ public static double logloss(double err) {
     }
     return transposed;
   }
+
+  /**
+   * Compute harmonic number estimated by natural logarithm
+   * Constant here: Euler–Mascheroni constant (also called Euler's constant)
```
honzasterba, line 779: instead of a comment I would put the constant into a public static final field with an explanatory name

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,377 @@
...
+            this._nodes = new Node[maxNumNodesInTree]; 
+        }
+
+        /**
+         * Implementation of Algorithm 2 (iTree) from paper.
+         * Algorithm not need exact row vectors in scoring (evaluation) stage but just number of rows in node.
+         * Therefore nodeFrames are removed from DKV after filtering.
+         */
+        public void buildTree() {
+            Frame frame = DKV.get(_frameKey).get();
```
honzasterba, line 186: seems unnecessary, can be moved to Node constructor and just do getFrame().numRows()

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,377 @@
...
+     * IsolationTree class implements Algorithm 2 (iTree) and Algorithm 3 (pathLength) 
+     * from the Extended Isolation Forest paper.
+     * 
+     * The binary tree recursive implementation is replaced by array implementation.
+     * It is the most simple array implementation of binary tree.
+     */
+    public static class IsolationTree extends Iced<IsolationTree> {
+        private Node[] _nodes;
+
+        private Key<Frame> _frameKey;
```
honzasterba, line 165: I think most of these could be final

maurever:
I am sorry @honzasterba and @michalkurka, I sent you a new request for review, but I would like to know just your opinion on the benchmark right now. 

But thank you @honzasterba for your new review. We agreed with your previous comments about removing MOJO and extension from SharedTreeModel. @valenad1 just has not implemented it yet, but it is on his todo list.

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
valenad1: Resolved in 1a9d25b2d41ce5d77d8a84e221adbba5a5c23e3a

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
valenad1: Resolved in ff3dcb86a04ea08f4b6373fc1dcf2abcd751f13e on PR: https://github.com/h2oai/h2o-3/pull/4907

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
valenad1: Will be done in https://github.com/h2oai/h2o-3/pull/4907

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForestMojoWriter.java
valenad1: Implemented in 1a9d25b2d41ce5d77d8a84e221adbba5a5c23e3a on PR: https://github.com/h2oai/h2o-3/pull/4907

h2o-algos/src/main/java/hex/tree/isoforextended/SubSampleTask.java
valenad1: Will be resolved on https://github.com/h2oai/h2o-3/pull/4907

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
```diff
@@ -0,0 +1,377 @@
...
+                Vec[] subFrame = ArrayUtils.subarray(_train.vecs(), 0, _train.numCols() - 4);
+                String[] subNames = ArrayUtils.subarray(_train.names(), 0, _train.numCols() - 4);
+                String[][] subDomains = ArrayUtils.subarray2DLazy(_train.domains(), 0, _train.numCols() - 4);
+
+                Frame subSample = new SubSampleTask(_parms._sample_size, _parms._seed + randomUnit)
+                        .doAll(subTypes, subFrame).outputFrame(Key.make(), subNames, subDomains);
+
+                iTree = new IsolationTree(subSample._key, heightLimit, _parms._seed + randomUnit, _parms.extension_level);
+                iTree.buildTree();
+                if (Log.isLoggingFor(Log.DEBUG)) {
```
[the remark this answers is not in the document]
valenad1: Implemented in PR: https://github.com/h2oai/h2o-3/pull/4907

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
valenad1: Implemented in PR: https://github.com/h2oai/h2o-3/pull/4907

h2o-core/src/main/java/water/util/MathUtils.java
valenad1: Implemented in PR: https://github.com/h2oai/h2o-3/pull/4907

h2o-genmodel/src/main/java/hex/genmodel/algos/isoforextended/ExtendedIsolationForestMojoModel.java
valenad1: Mojo is postponed.

h2o-py/docs/modeling.rst
valenad1: It's a general problem on windows, will be fixed in https://github.com/h2oai/h2o-3/pull/4907

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
valenad1: Will be fixed in https://github.com/h2oai/h2o-3/pull/4907

h2o-algos/src/main/java/hex/tree/isoforextended/ExtendedIsolationForest.java
valenad1: Will be further discussed in https://github.com/h2oai/h2o-3/pull/4907

valenad1:
The production version of algorithm will be finished here: https://github.com/h2oai/h2o-3/pull/5246

PR #4321, opened 2020-02-14
PUBDEV-7286: Update docs to specify that mtries=-2 gives all features

hannah-tillman:


h2o-docs/src/product/data-science/algo-params/mtries.rst
```diff
@@ -9,11 +9,13 @@ Description
 
 Use this option to specify the number of columns to randomly select at each level. 
 
-This value defaults to -1, and the range for this option is -1 to >=1. If the default value is used, then the number of variables is:
+This value defaults to -1, and the range for this option is -2 to >=1. If a value other that -1 or -2 is used, then the number of variables is:
```
neemah2o: If a value other that -1 or -2 -> If a value other **than** -1 or -2

angela0xdata: I made a small update to the value range and changed "that" to "than" in the process.

PR #4324, opened 2020-02-17
[xgboost] fix nthread limiting via prop

honzasterba:


h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostModelTest.java
```diff
@@ -9,6 +9,14 @@
 
 public class XGBoostModelTest {
   
+  @Test
+  public void shouldGetCorrectMaxThread() {
+    assertEquals(Runtime.getRuntime().availableProcessors(), XGBoostModel.getMaxNThread());
+    System.setProperty("sys.ai.h2o.xgboost.nthreadMax", "2");
+    assertEquals(2, XGBoostModel.getMaxNThread());
+    System.clearProperty("sys.ai.h2o.xgboost.nthreadMax");
```
Pscheidl: Proper, fault-tolerant cleanup required.

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostModelTest.java
```diff
@@ -9,6 +9,14 @@
 
 public class XGBoostModelTest {
   
+  @Test
+  public void shouldGetCorrectMaxThread() {
+    assertEquals(Runtime.getRuntime().availableProcessors(), XGBoostModel.getMaxNThread());
+    System.setProperty("sys.ai.h2o.xgboost.nthreadMax", "2");
```
Pscheidl: There is a dedicated JUnit library named [System rules](https://stefanbirkner.github.io/system-rules/) which will help you avoid all the problems associated with testing based on system properties. We already have the dependency (which you can update optionally) and it is already used in H2O.

If nothing else, it makes sure the environment variables are restored properly. See `SQLManagerTest` for example.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -35,7 +35,7 @@
         implements SharedTreeGraphConverter, Model.LeafNodeAssignment, Model.Contributions {
 
   private static final String PROP_VERBOSITY = H2O.OptArgs.SYSTEM_PROP_PREFIX + ".xgboost.verbosity";
-  private static final String PROP_NTHREAD = SYSTEM_PROP_PREFIX + "xgboost.nthread";
+  private static final String PROP_NTHREAD = SYSTEM_PROP_PREFIX + "xgboost.nthreadMax";
```
Pscheidl: This appears to require a change in documentation, yet there is no JIRA tied with this PR.

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostModelTest.java
honzasterba: thanks, used it

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
honzasterba: the property is in experimental mode and has not been documented yet

PR #4325, opened 2020-02-17
PUBDEV-7305: missing model timing properties in client mode

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-7305

Ensuring that model is updated to DKV after setting those properties.
GLM being the usual bad boy, had to change the lock/update logic to make it work consistently like other models, ie. not leaving the CV models locked until the main model is being built, but by acquiring the lock only when needed.

Added a R test for this as only R tests are run in Client Mode...
There is already a Py test checking this: h2o-py/tests/testdir_jira/pyunit_pubdev_5829_pubdev_6091_get_model_timestamp.py

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -172,79 +172,82 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
       return levelOneFrame;
     }
 
-  /**
-   * Prepare a "level one" frame for a given set of models and actuals.
-   * Used for preparing validation frames for the metalearning step, and could also be used for bulk predictions for a StackedEnsemble.
-   */
-  private Frame prepareLevelOneFrame(String levelOneKey, Key<Model>[] baseModelKeys, Frame actuals, boolean isTraining) {
-    List<Model> baseModels = new ArrayList<>();
-    List<Frame> baseModelPredictions = new ArrayList<>();
+    /**
+     * Prepare a "level one" frame for a given set of models and actuals.
+     * Used for preparing validation frames for the metalearning step, and could also be used for bulk predictions for a StackedEnsemble.
+     */
+    private Frame prepareLevelOneFrame(String levelOneKey, Key<Model>[] baseModelKeys, Frame actuals, boolean isTraining) {
+      List<Model> baseModels = new ArrayList<>();
+      List<Frame> baseModelPredictions = new ArrayList<>();
 
-    for (Key<Model> k : baseModelKeys) {
-      Model aModel = DKV.getGet(k);
-      if (null == aModel)
-        throw new H2OIllegalArgumentException("Failed to find base model: " + k);
+      for (Key<Model> k : baseModelKeys) {
+        Model aModel = DKV.getGet(k);
+        if (null == aModel)
+          throw new H2OIllegalArgumentException("Failed to find base model: " + k);
 
-      Frame predictions = getPredictionsForBaseModel(aModel, actuals, isTraining);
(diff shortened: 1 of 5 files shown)


h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -172,79 +172,82 @@ private Frame prepareLevelOneFrame(String levelOneKey, Model[] baseModels, Frame
...
+    private Key<Frame> buildPredsKey(Key model_key, long model_checksum, Key frame_key, long frame_checksum) {
+      return Key.make("preds_" + model_checksum + "_on_" + frame_checksum);
+    }
 
-  protected Key<Frame> buildPredsKey(Model model, Frame frame) {
-    return frame == null || model == null ? null : buildPredsKey(model._key, model.checksum(), frame._key, frame.checksum());
-  }
+    protected Key<Frame> buildPredsKey(Model model, Frame frame) {
+      return frame == null || model == null ? null : buildPredsKey(model._key, model.checksum(), frame._key, frame.checksum());
+    }
```
sebhrusen: I recommend to hide whitespaces in the diff as this file has been reindented (indentation was broken).

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -271,6 +276,10 @@ private void setFinalState() {
     if (res != null && res._output != null) {
       res._output._job = _job;
       res._output.stopClock();
+//      res.unlock(_job == null ? null : _job._key, false); // last resort: dirty way to force unlock to be able to reacquire lock
+      res.write_lock(_job);
+      res.update(_job);
+      res.unlock(_job);
```
sebhrusen: @michalkurka I fund this very frustrating that by simply adding this lock-update-unlock after model completion, I got a lot of tests failing due to a large diversity in how algos implementations handle locks:
- no abstraction or lock-update-unlock at high level, for example around `init` or `computeImpl`.
- unlocks not systematically applied in a `finally` block, so that the model can remain locked if an exception is thrown.
- the laurels go to GLM, where the locking logic was different for CV models and for the main model...

it was supposed to be a small fix, and turns into a pretty risky change (relying fully on our test coverage).
What do you think about coming up with better encapsulation for this locking logic, at least for model training?

michalkurka: > What do you think about coming up with better encapsulation for this locking logic, at least for model training?

That seems like a good idea.

> unlocks not systematically applied in a finally block, so that the model can remain locked if an exception is thrown.

People really struggle with this, another common issue is assuming that the lock will always be acquired and then always trying to unlock the model in `finally`.

Better to do smaller steps - fix GLM and separately improve the framework.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -172,79 +172,82 @@ private Frame prepareLevelOneFrame(String levelOneKey, Model[] baseModels, Frame
       return levelOneFrame;
     }
 
-  /**
```
sebhrusen: I recommend to hide whitespaces in the diff as this file has been reindented (indentation was broken).

sebhrusen:
@michalkurka , @wendycwong could you please have a look at this? With the changes in the locks, I'd like to have people run it locally for a while before next release, thx/

h2o-core/src/main/java/hex/ModelBuilder.java
wendycwong: @seb:  GLM with CV was not implemented using the conventional method.  I am not sure why Tomas did it his way though.  Thanks you for fixing it.

PR #4326, opened 2020-02-18
printStackTrace is evil

honzasterba:
and should not be used to replace logging or error handling

### DIFF
--- a/h2o-algos/src/main/java/hex/tree/SharedTree.java
+++ b/h2o-algos/src/main/java/hex/tree/SharedTree.java
@@ -780,23 +780,6 @@ public abstract class SharedTree<
     // nothing by default - can be overridden in subclasses
   }
 
-  static int counter = 0;
-  // helper for debugging
-  @SuppressWarnings("unused")
-  static protected void printGenerateTrees(DTree[] trees) {
-    for( DTree dtree : trees )
-      if( dtree != null ) {
-        try {
-          PrintWriter writer = new PrintWriter("/tmp/h2o-3.tree" + ++counter + ".txt", "UTF-8");
-          writer.println(dtree.root().toString2(new StringBuilder(), 0));
-          writer.close();
-        } catch (FileNotFoundException|UnsupportedEncodingException e) {
-          throw new RuntimeException(e);
-        }
-        System.out.println(dtree.root().toString2(new StringBuilder(), 0));
-      }
-  }
-
   protected TwoDimTable createScoringHistoryTable() {
     O out = _model._output;
     return createScoringHistoryTable(out, out._scored_train, out._scored_valid, _job,
--- a/h2o-algos/src/main/java/hex/tree/SharedTree.java
+++ b/h2o-algos/src/main/java/hex/tree/SharedTree.java
@@ -791,7 +791,7 @@ public abstract class SharedTree<
           writer.println(dtree.root().toString2(new StringBuilder(), 0));
           writer.close();
         } catch (FileNotFoundException|UnsupportedEncodingException e) {
-          e.printStackTrace();
+          throw new RuntimeException(e);
         }
         System.out.println(dtree.root().toString2(new StringBuilder(), 0));
       }
--- a/h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
+++ b/h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
@@ -680,28 +680,32 @@ public abstract class SharedTreeModel<
       fileCtx.add(new CodeGenerator() {
         @Override
         public void generate(JCodeSB out) {
-          // Generate a class implementing a tree
-          out.nl();
-          toJavaForestName(out.ip("class "), mname, treeIdx).p(" {").nl().ii(1);
-          out.ip("public static void score0(double[] fdata, double[] preds) {").nl().ii(1);
-          for (int c = 0; c < nclass; c++) {
-            if (_output._treeKeys[treeIdx][c] == null) continue;
-            if (!(binomialOpt() && c == 1 && nclass == 2)) // Binomial optimization
-              toJavaTreeName(out.ip("preds[").p(nclass == 1 ? 0 : c + 1).p("] += "), mname, treeIdx, c).p(".score0(fdata);").nl();
-          }
-          out.di(1).ip("}").nl(); // end of function
-          out.di(1).ip("}").nl(); // end of forest class
-
-          // Generate the pre-tree classes afterwards
-          for (int c = 0; c < nclass; c++) {
-            if (_output._treeKeys[treeIdx][c] == null) continue;
-            if (!(binomialOpt() && c == 1 && nclass == 2)) { // Binomial optimization
-              String javaClassName = toJavaTreeName(new SB(), mname, treeIdx, c).toString();
-              CompressedTree ct = _output.ctree(treeIdx, c);
-              SB sb = new SB();
-              new TreeJCodeGen(SharedTreeModel.this, ct, sb, javaClassName, verboseCode).generate();
-              out.p(sb);
+          try {
+            // Generate a class implementing a tree
+            out.nl();
+            toJavaForestName(out.ip("class "), mname, treeIdx).p(" {").nl().ii(1);
+            out.ip("public static void score0(double[] fdata, double[] preds) {").nl().ii(1);
+            for (int c = 0; c < nclass; c++) {
+              if (_output._treeKeys[treeIdx][c] == null) continue;
+              if (!(binomialOpt() && c == 1 && nclass == 2)) // Binomial optimization
+                toJavaTreeName(out.ip("preds[").p(nclass == 1 ? 0 : c + 1).p("] += "), mname, treeIdx, c).p(".score0(fdata);").nl();
             }
+            out.di(1).ip("}").nl(); // end of function
+            out.di(1).ip("}").nl(); // end of forest class
+
+            // Generate the pre-tree classes afterwards
+            for (int c = 0; c < nclass; c++) {
+              if (_output._treeKeys[treeIdx][c] == null) continue;
+              if (!(binomialOpt() && c == 1 && nclass == 2)) { // Binomial optimization
+                String javaClassName = toJavaTreeName(new SB(), mname, treeIdx, c).toString();
+                CompressedTree ct = _output.ctree(treeIdx, c);
+                SB sb = new SB();
+                new TreeJCodeGen(SharedTreeModel.this, ct, sb, javaClassName, verboseCode).generate();
+                out.p(sb);
+              }
+            }
+          } catch (Exception e) {
+            throw new RuntimeException("Internal error creating the POJO.", e);
           }
         }
       });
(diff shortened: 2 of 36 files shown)


h2o-algos/src/main/java/hex/deeplearning/DeepLearningModel.java
```diff
@@ -2087,7 +2087,7 @@ static void checkIfParameterChangeAllowed(final DeepLearningParameters oldP, fin
                     throw new H2OIllegalArgumentException("Cannot change parameter: '" + fBefore.getName() + "': " + fBefore.get(oldP) + " -> " + fAfter.get(newP));
                   }
                 } catch (IllegalAccessException e) {
-                  e.printStackTrace();
```
michalkurka: I would say this was intentional

h2o-core/src/main/java/water/H2O.java
```diff
@@ -820,8 +820,8 @@ public static void notifyAboutCloudSize(InetAddress ip, int port, InetAddress le
         output.write(':');
         output.write(Integer.toString(API_PORT));
         output.flush();
-      } catch ( IOException e ) {
-        e.printStackTrace();
+      } catch (IOException e) {
```
michalkurka: inconsistent error handling - H2O.exit vs RuntimeException in the same method

h2o-core/src/main/java/water/RPC.java
```diff
@@ -175,64 +175,59 @@ private void handleCompleter( CountedCompleter cc ) {
 
     // Keep a global record, for awhile
     if( _target != null ) _target.taskPut(_tasknum,this);
-    try {
```
michalkurka: I would keep as-is

h2o-core/src/main/java/water/api/KillMinus3Handler.java
```diff
@@ -33,8 +33,7 @@ public KillMinus3V3 killm3(int version, KillMinus3V3 u) {
           } catch( java.io.IOException ioe ) {
             // Silently ignore if, e.g. /bin/kill does not exist on windows
           } catch (Exception xe) {
-            xe.printStackTrace();
-            throw new H2OIllegalArgumentException("");
```
michalkurka: Why?

h2o-core/src/main/java/water/api/RapidsHandler.java
```diff
@@ -32,17 +32,8 @@ public RapidsSchemaV3 exec(int version, RapidsSchemaV3 rapids) {
       RapidsHandler.SESSIONS.put(rapids.session_id, ses);
     }
 
-    Val val;
```
michalkurka: Keep as-is

h2o-core/src/main/java/water/parser/ZipUtil.java
```diff
@@ -81,7 +82,7 @@ static boolean isZipDirectory(ByteVec bv) {
         return isDir;
       }
     } catch (IOException e) {
-      e.printStackTrace();
```
michalkurka: Intentional

h2o-core/src/main/java/water/parser/ZipUtil.java
```diff
@@ -106,7 +107,7 @@ static boolean isZipDirectory(ByteVec bv) {
         }
         zipFile.close();
       } catch (IOException e) {
-        e.printStackTrace();
```
michalkurka: Intentionally ignored

h2o-core/src/main/java/water/parser/ZipUtil.java
```diff
@@ -140,7 +141,7 @@ static float getDecompressionRatio(ByteVec bv) {
...
```
michalkurka: Intentional

h2o-core/src/main/java/water/parser/ZipUtil.java
```diff
@@ -180,13 +181,12 @@ else if (cpr == Compression.ZIP) {
           return bits.length / zips.length;
         }
       } catch (IOException e) {
-        e.printStackTrace();
```
michalkurka: Intentional

h2o-core/src/main/java/water/parser/ZipUtil.java
```diff
@@ -268,14 +268,15 @@ else if (cpr == Compression.ZIP) {
         }
       }
     } catch (IOException e) {
-      e.printStackTrace();
```
michalkurka: Intentional

h2o-core/src/main/java/water/util/IcedHashMapBase.java
```diff
@@ -150,10 +150,6 @@ public final AutoBuffer write_impl( AutoBuffer ab ) {
         default:
           return ab.put(null);
       }
-    } catch(Throwable t){
-      System.err.println("Iced hash map serialization failed! " + t.toString() + ", msg = " + t.getMessage());
-      t.printStackTrace();
-      throw H2O.fail("Iced hash map serialization failed!" + t.toString() + ", msg = " + t.getMessage());
```
michalkurka: Seems incorrect to remove

h2o-core/src/main/java/water/util/IcedHashMapGeneric.java
```diff
@@ -152,8 +152,6 @@ else if (isDoubleVal(mode))
       }
       ab.put1(-1);
     } catch(Throwable t){
-      System.err.println("Iced hash map serialization failed! " + t.toString() + ", msg = " + t.getMessage());
-      t.printStackTrace();
```
michalkurka: Stacktrace lost

h2o-core/src/main/java/water/RPC.java
honzasterba: this change has 2 parts:
1) remove try..catch which IMHO is a NOOP
t.printStackTrace();
      throw Log.throwErr(t); // this prints the exception again and re throws it as runtimeexception
2) whitespace change because of remove block

h2o-core/src/main/java/water/api/KillMinus3Handler.java
honzasterba: fxied

h2o-core/src/main/java/water/api/RapidsHandler.java
honzasterba: why? whats the point of re-throwing some errors and rethrowing and logging (twice) some other errors? functionally this code does nothing.

h2o-core/src/main/java/water/util/IcedHashMapBase.java
honzasterba: ok, will just clean up

h2o-core/src/main/java/water/util/IcedHashMapGeneric.java
honzasterba: fixed

h2o-core/src/main/java/water/RPC.java
michalkurka: ad (1) - it is not NOOP, it is attempting to make sure the Throwable is written at least to stdout and logged. This is a desperate attempt to preserve the exception (even such as OutOfMemoryError). This is done on purpose - I would leave as-is.

h2o-core/src/main/java/water/api/RapidsHandler.java
honzasterba: reverted

michalkurka:
@honzasterba thanks for the changes, please wait with merging - I will go over it one more time

h2o-core/src/main/java/water/init/Linpack.java
```diff
@@ -72,7 +72,7 @@ public void run() {
       try {
         threads[t].join();
       } catch (InterruptedException e) {
-        e.printStackTrace();
+        Thread.currentThread().interrupt();
```
jakubhava: 👍

h2o-algos/src/main/java/hex/tree/SharedTree.java
```diff
@@ -791,7 +791,7 @@ static protected void printGenerateTrees(DTree[] trees) {
           writer.println(dtree.root().toString2(new StringBuilder(), 0));
           writer.close();
         } catch (FileNotFoundException|UnsupportedEncodingException e) {
-          e.printStackTrace();
+          throw new RuntimeException(e);
```
michalkurka: I think we can just delete this whole method

maybe we can add `toDebugString()` to `DTree` - it could do `dtree.root().toString2(new StringBuilder(), 0)` instead

h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
```diff
@@ -680,33 +680,28 @@ public SharedTreeSubgraph getSharedTreeSubgraph(final int tidx, final int cls) {
...
+            if (!(binomialOpt() && c == 1 && nclass == 2)) { // Binomial optimization
+              String javaClassName = toJavaTreeName(new SB(), mname, treeIdx, c).toString();
+              CompressedTree ct = _output.ctree(treeIdx, c);
+              SB sb = new SB();
+              new TreeJCodeGen(SharedTreeModel.this, ct, sb, javaClassName, verboseCode).generate();
+              out.p(sb);
             }
-          } catch (Throwable t) {
-            t.printStackTrace();
-            throw new IllegalArgumentException("Internal error creating the POJO.", t);
```
michalkurka: I think it is useful to preserve this error message instead of sending back an error that the POJO cannot be compile and the syntax error (which is what is most often happens).

h2o-core/src/main/java/hex/DMatrix.java
```diff
@@ -253,7 +253,6 @@ public VecTsk(H2OCountedCompleter cmp, Key progressKey, double [] y){
         final Chunk xChunk = chks[i];
         for (int k = xChunk.nextNZ(-1); k < res.length; k = xChunk.nextNZ(k))
           try { res[k] += yVal * xChunk.atd(k);} catch(Throwable t) {
```
michalkurka: This doesn't really make much sense, I would just get rid off the whole try - catch

h2o-core/src/main/java/hex/Model.java
```diff
@@ -879,7 +879,7 @@ public void printTwoDimTables(StringBuilder sb, Object o) {
             f.setAccessible(true);
             if (t != null) sb.append(t.toString(1,false /*don't print the full table if too long*/));
           } catch (IllegalAccessException e) {
-            e.printStackTrace();
+            throw new RuntimeException(e);
```
michalkurka: I don't think an issue with "printer"/toString method needs halt the algorithm - it is better to log an error and add a note about the error to the string output as well

h2o-core/src/main/java/hex/ModelMojoWriter.java
```diff
@@ -75,7 +75,7 @@ public ModelMojoWriter(M model) {
       writeTo(zos);
       zos.close();
     } catch (IOException e) {
-      e.printStackTrace();
```
michalkurka: omg, that was a bad bad thing - thx for fixing

PR #4329, opened 2020-02-18
AutoML test cleanup

sebhrusen:
removed test abusing resources and replaced by proper args test


sebhrusen:
@ledell FYI. merging

PR #4333, opened 2020-02-19
PUBDEV-7158: R Docs: Add Examples to R Reference Guide

hannah-tillman:


angela0xdata:
Some of the algorithms also need examples added. For example:
source: https://github.com/h2oai/h2o-3/blob/master/h2o-bindings/bin/custom/R/gen_drf.py
doc: http://docs.h2o.ai/h2o/latest-stable/h2o-r/docs/reference/h2o.randomForest.html

hannah-tillman:
> Some of the algorithms also need examples added. For example:
> source: https://github.com/h2oai/h2o-3/blob/master/h2o-bindings/bin/custom/R/gen_drf.py
> doc: http://docs.h2o.ai/h2o/latest-stable/h2o-r/docs/reference/h2o.randomForest.html

h2o-r/h2o-package/R/frame.R
```diff
@@ -2864,6 +3163,13 @@ h2o.ischaracter <- function(x) {
 #' @name h2o.asfactor
 #' @param x An H2OFrame object.
 #' @seealso \code{\link[base]{as.factor}} for the base R implementation.
```
angela0xdata: this "see also" points to itself. Can you change this since you're already working in this file :)
>#' @seealso \code{\link[base]{as.numeric}} for the base R implementation.

h2o-r/h2o-package/R/frame.R
```diff
@@ -2864,6 +3163,13 @@ h2o.ischaracter <- function(x) {
...
+#' @examples 
+#' \dontrun{
+#' library(h2o)
+#' h2o.init()
+#' > h2o <- h2o.importFile("https://s3.amazonaws.com/h2o-public-test-data/smalldata/junit/cars_20mpg.csv")
```
angela0xdata: Remove the starting ">" in line 3170

h2o-r/h2o-package/R/models.R
```diff
@@ -488,6 +505,18 @@ h2o.predict <- function(object, newdata, ...){
 #' @param model A trained model representing the transformation strategy
 #' @param ... Transformation model-specific parameters
 #' @return Returns an H2OFrame object with data transformed.
+#' @examples
```
angela0xdata: Hmm.. this h2o.transform example doesn't really make sense. (There are no words for the word2vec model, and the function is "h2o.transform_word2vec()" rather than "h2o.transform()".) Can you take this out for now?

h2o-r/h2o-package/R/models.R
```diff
@@ -3439,6 +3785,20 @@ h2o.cross_validation_holdout_predictions <- function(object) {
 #'
 #' @param object An \linkS4class{H2OModel} object.
 #' @return Returns a list of H2OFrame objects
+#' @examples
```
angela0xdata: This cross_validation_predictions example is missing the opening and closing { }

PR #4335, opened 2020-02-19
PUBDEV-6615: DeepLearning MOJOs should be thread-safe

michalkurka:


wendycwong:
@michalkurka:  Another issue with Deeplearning.  Mojo (autoencoder) does not work with categorical columns.  The reasons are DL treat categorical columns differently than the other algos.  In your test, may want to avoid that.

michalkurka:
> Another issue with Deeplearning. Mojo (autoencoder) does not work with categorical columns. The reasons are DL treat categorical columns differently than the other algos. In your test, may want to avoid that.

Thank you Wendy, i am using numerical-only for the purpose of simplicity of this test. @wendycwong do we have a jira issue for the problem you described?

PR #4336, opened 2020-02-19
Adding transform_word2vec to R docs

angela0xdata:


angela0xdata:
@Pscheidl I'm not sure why the R3.5 CMD Check as CRAN test failed

h2o-r/h2o-package/R/models.R
```diff
@@ -567,10 +567,10 @@ setMethod("h2o.transform", signature("H2OTargetEncoderModel"), function(model, d
 #'
 #' # Transform words to vectors without aggregation
 #' sentences <- as.character(as.h2o(c("b", "c", "a", NA, "b")))
-#' h2o.transform(w2v_model, sentences) # -> 5 rows total, 2 rows NA ("c" is not in the vocabulary)
+#' h2o.transform_word2vec(w2v_model, sentences) # -> 5 rows total, 2 rows NA ("c" is not in the vocabulary)
```
Pscheidl: The method defines below is `h2o.transform` not `h2o.transform_word2vec`. That's why the tests fail I guess.

I don't understand why we need such an example above `h2o.transform` method. This should probably be in `h2o.transform_word2vec` in `w2vutils.R` file. I can make the change if you agree.

angela0xdata: Sure! You're always welcome to make changes. 
And the goal is to include examples for each R function. I guess I'm still confused about transform vs. transform_word2vec :flushed:

angela0xdata:
@michalkurka I may have misunderstood the thread in h2o-3-r-client channel yesterday, which started as a request for "what does h2o.transform() do?"

Though one change in this PR is to actually add the transform_word2vec() function to the docs (where it is not visible now).

michalkurka:
> Though one change in this PR is to actually add the transform_word2vec() function to the docs (where it is not visible now).

Is there no place that would show how to use transform with word2vec?

I think @sebhrusen introduced this change - what effect does it have on documentation? Do we need to export the function?

Pscheidl:
No, it was me who introduced this change, @sebhrusen only merged it manually I believe. It was part of TE changes.

There is a `h2o.transform` method for word2Vec in `models.R` as well.

```
setMethod("h2o.transform", signature("H2OWordEmbeddingModel"), function(model, words, aggregate_method = c("NONE", "AVERAGE")) {
```

Pscheidl:
I think the issue is that in `w2vutils.R` there is the original method `transform_word2vec` (it was moved there).

However, this very function has `h2o.transform`  method call in the API examples.

sebhrusen:
correct, I just had to do a manual merge due to merge conflicts.
As discussed in Slack, I don't think we should merge this PR: `transform_word2vec` should rather be fully deprecated.

angela0xdata:
okay, I can close if transform_word2vec is deprecated.

PR #4340, opened 2020-02-20
PUBDEV-7267 - Terminal nodes recognized by lack of children. Also, split column not filled for terminal nodes in case of XGBoost.

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7262

Please note, marked for `3.28.0.5`, which is not today's release. Feel free to move it to current release of course (@michalkurka).

- XGBoost uses 0 as split column index on terminal nodes, which means nothing. We should not add split column if the nodes is terminal for XGBoost models (fixed).
- Adjusted leaf-detection mechanism inside H2OTreeHandler as well - if the node does not have any children and has empty split column, it is considered to be a leaf node. Until now, only the presence of split column name was tested - this step is not needed for the fix.

- Leaf nodes did not have prediction values in description for some reason. I added them. This has nothing to do with the original issue, but it is convenient to add it as well.

### DIFF
--- a/h2o-algos/src/main/java/hex/tree/TreeHandler.java
+++ b/h2o-algos/src/main/java/hex/tree/TreeHandler.java
@@ -182,7 +182,7 @@ public class TreeHandler extends Handler {
         int[] nodeLevels = node.getParent().isBitset() ? extractNodeLevels(node) : null;
         nodeDescriptionBuilder.append("Node has id ");
         nodeDescriptionBuilder.append(node.getNodeNumber());
-        if (node.getColName() != null && node.getRightChild() == null && node.getLeftChild() == null) {
+        if (node.getColName() != null && node.isLeaf()) {
             nodeDescriptionBuilder.append(" and splits on column '");
             nodeDescriptionBuilder.append(node.getColName());
             nodeDescriptionBuilder.append("'. ");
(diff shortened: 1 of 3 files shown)


h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -182,7 +182,7 @@ private static void fillnodeDescriptions(final SharedTreeNode node, final String
         int[] nodeLevels = node.getParent().isBitset() ? extractNodeLevels(node) : null;
         nodeDescriptionBuilder.append("Node has id ");
         nodeDescriptionBuilder.append(node.getNodeNumber());
-        if (node.getColName() != null) {
+        if (node.getColName() != null && node.getRightChild() == null && node.getLeftChild() == null) {
```
michalkurka: `SharedTreeNode` has a `isLeaf` method

Pscheidl: Thanks, will fix ASAP.

Pscheidl: Fixed in https://github.com/h2oai/h2o-3/pull/4340/commits/7895038b96793e2991f9a8b87e85725f177bba9a

Pscheidl:
Tests passed. We might as well have this in the release @michalkurka .

PR #4341, opened 2020-02-20
PSVM shouldn't be used a SE Metalearner

michalkurka:
Since PSVM has a bug with metrics (NaNs for some), let's remove it as a metalearner in SE: https://0xdata.atlassian.net/browse/PUBDEV-7314


sebhrusen:
thx @michalkurka, I'll add metrics checks in the SE tests later

PR #4344, opened 2020-02-21
[PUBDEV-6862] XGBoost off cluster POC

honzasterba:
**THIS IS ONLY PoC**

current state:
it works - it builds a complete xgboost with multiple scoring rounds
data transmission is very rudimentary a probably will have issues for large boosters


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/exec/RemoteXGBoostExecutor.java
```diff
@@ -0,0 +1,123 @@
...
+        assert modelKey.equals(resp.key.key());
+    }
+    
+    private XGBoostExecRespV3 post(Key key, XGBoostExecReqV3.Type type, XGBoostExecReq reqContent) {
+        XGBoostExecReqV3 req = new XGBoostExecReqV3(key, type, reqContent);
+        HttpPost httpReq = new HttpPost(uri);
+        httpReq.setEntity(new StringEntity(req.toJsonString(), UTF_8));
+        httpReq.setHeader(CONTENT_TYPE, ContentType.APPLICATION_JSON.getMimeType());
+        Log.info("Remote XGBoost: Request " + type + " " + reqContent);
+        try (CloseableHttpClient client = HttpClientBuilder.create().build();
```
Pscheidl, line 69: Connection pooling is not essential to us, as we don't stand and fall with REST API calls performance. It's worth thinking of it. Apache HTTP has connection pooling built-in. Moving this to another layer with reasonable connection pool timeouts (all the way down to absolute zero) sounds correct

https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html

If H2O had JAX-RS or similar, with some reasonable implementation like Jersey (https://eclipse-ee4j.github.io/jersey/), our life would be easier.

honzasterba: thanks will add to the list of things to do in the final impl

PR #4347, opened 2020-02-21
fix PUBDEV-6063 failing test (since changes in PUBDEV-7227)

sebhrusen:
removed the not-really-useful temporary `splitter` frame renaming in `split_frame` method as it causes the original key to remain until next Python GC collection (Python <=3.6 only), and causing `pyunit_pubdev_6603.py` to fail.

### DIFF
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -2706,7 +2706,6 @@ class H2OFrame(Keyed):
 
         splits = []
         tmp_runif = self.runif(seed)
-        tmp_runif.frame_id = "%s_splitter" % _py_tmp_key(h2o.connection().session_id)
 
         i = 0
         while i < num_slices:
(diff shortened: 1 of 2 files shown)


h2o-py/h2o/frame.py
```diff
@@ -2706,7 +2706,6 @@ def split_frame(self, ratios=None, destination_frames=None, seed=None):
 
         splits = []
         tmp_runif = self.runif(seed)
-        tmp_runif.frame_id = "%s_splitter" % _py_tmp_key(h2o.connection().session_id)
```
sebhrusen: should I revert this one? or do we leave this change as is?

michalkurka: I think we can keep it in the PR, I see no point giving something a pretty name if we are explicitly disposing later

PR #4343, opened 2020-02-21
PUBDEV-7312 - AWS Auth with session tokens

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7312

Marked for `3.28.0.4`

PR #4348, opened 2020-02-21
PUBDEV-7317: Add AWS Session Token to docs

angela0xdata:
This PR adds AWS Session Token to the docs. This is changed in dev in this PR:
https://github.com/h2oai/h2o-3/pull/4343

### DIFF
--- a/h2o-docs/src/product/cloud-integration/ec2-and-s3.rst
+++ b/h2o-docs/src/product/cloud-integration/ec2-and-s3.rst
@@ -18,8 +18,51 @@ AWS Standalone Instance
 '''''''''''''''''''''''
 
 When running H2O in standalone mode using the simple Java launch command, we can pass in the S3 credentials in three ways.
+H2O supports both AWS Credentials (pair consisting of AWS SECRET KEY and AWS SECRET ACCESS KEY) and temporary authentication using Session token
+(a triplet consisting of AWS SECRET KEY, AWS SECRET ACCESS KEY and AWS SESSION TOKEN).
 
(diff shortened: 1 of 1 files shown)


michalkurka:
I don't know if this is correct - @Pscheidl please review

Should we have separate sections for "regular" aws credentials and temporary ones? I could image someone could be thinking we require the session token.

angela0xdata:
@Pscheidl @michalkurka should I just mark the new session token as optional in the docs?

Pscheidl:
That's one good way to solve it, instead of explaining it all over and over again @michalkurka. 

I did a small push with one commit that splits it into two distinct sections. Please have a look @angela0xdata and just revert it if you feel like it's crap.

h2o-docs/src/product/cloud-integration/ec2-and-s3.rst
```diff
@@ -59,7 +61,53 @@ When running H2O in standalone mode using the simple Java launch command, we can
...
+         <value>[AWS SECRET KEY]</value>
+       </property>
+
+       <property>
+         <name>fs.s3.awsSecretAccessKey</name>
+         <value>[AWS SECRET ACCESS KEY]</value>
+       </property>
+
+       <property>
+         <name>fs.s3.awsSessionToken</name>
```
michalkurka: @Pscheidl can you confirm this actually works?

Pscheidl: No, the configuration has to be altered. But it can actually work. Created a fix PR here: https://github.com/h2oai/h2o-3/pull/4352

Thank you.

PR #4355, opened 2020-02-22
[PUBDEV-6900] xgboost 1.0.0

honzasterba:


docker/xgb/ubuntu/common/sbin/install_packages
```diff
@@ -11,5 +11,10 @@ add-apt-repository -y ppa:deadsnakes/ppa
 apt-get update
 apt-get install -y \
     wget gcc python3.5-dev git python-pip python-virtualenv \
-    libmysqlclient-dev unzip tar \
-    nodejs npm libkrb5-dev
+    mariadb-client \
```
michal-raska: just curious, why are we moving towards mariadb?

h2o-genmodel-extensions/xgboost/build.gradle
```diff
@@ -7,7 +7,7 @@ sourceCompatibility = 1.7
 targetCompatibility = 1.7
 
 ext {
-    xgboost4jVersion = "0.90.6"
+    xgboost4jVersion = "1.0.0.8-honza-1.0.0-SNAPSHOT"
```
michal-raska: is the SNAPSHOT here ok?

docker/xgb/ubuntu/common/sbin/install_packages
honzasterba: because its impossible to isntall libmysqlclient-dev and node at the same time in ubuntu

h2o-genmodel-extensions/xgboost/build.gradle
honzasterba: is temporary

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/util/BoosterHelper.java
```diff
@@ -11,7 +13,7 @@
 public class BoosterHelper {
 
   public static Booster loadModel(InputStream in) throws XGBoostError, IOException {
-    return Booster.loadModel(in);
```
honzasterba: was using a package-private method for no reason since XGBoost.loadModel exposes it

PR #4359, opened 2020-02-24
PUBDEV-7325: Fixing Data Manipulation section in Py doc

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-7325

### DIFF
--- a/h2o-py/h2o/targetencoder.py
+++ b/h2o-py/h2o/targetencoder.py
@@ -18,7 +18,8 @@ __all__ = ("TargetEncoder", )
 class TargetEncoder(object):
 
     """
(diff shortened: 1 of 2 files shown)


h2o-py/h2o/targetencoder.py
```diff
@@ -18,7 +18,8 @@
 class TargetEncoder(object):
 
     """
-    Deprecated API. Please use H2OTargetencoderEstimator instead.
+    :deprecated:
+    Use :func:`h2o.estimators.targetencoder.H2OTargetEncoderEstimator` instead.
```
sebhrusen: proper documentation for deprecation in Python, generating link to the replacement class.

PR #4360, opened 2020-02-25
[PUBDEV-7327] Remove code for external backend from H2O

jakubhava:
Sparkling Water is no longer using code for external backend in H2O, we can remove 🎉


michalkurka:
@jakubhava merges to rel-yu will not be propagated to any release, please revert the commits on rel-yu and make a new PR against master. Thank you!

michalkurka:
@jakubhava, in that case, please just revert the commit from rel-yu to keep the state as it was before

jakubhava:
Reverted https://github.com/h2oai/h2o-3/commit/187c3f32030630091c0502074a958bfddf6e3faf

PR #4362, opened 2020-02-25
PUBDEV-7319 Add Target Encoding as yet another categorical encoding supported by a Model

deil87:
As it was discussed here https://github.com/h2oai/h2o-3/pull/4303 as a first short-term version we want to have TE embedded in a Model itself ( similar to other existing encoders OHE, LabelEncoder etc). Long-term plan is to use Pipelines.

This PR covers:
1) support for attaching TE model to a Model/ModelBuilder/Model.Parameters class and ability to score on non-encoded data
~~2) support for clients (Py, R, Flow?)~~

<p><del>Last attempt was to use same approach as non-trainable encoders and introduce `Model.Parameters.CategoricalEncodingScheme.TargetEncoder` but looks like there are more differences than similarities and probably TE should be applied in a separate fashion.
</del></p>
<p><del>
The main difference as it is seen now is that TE could be trained and applied to specified categorical columns ( especially in the context of AutoML integration where we have strategies that decide to which categorical columns it is worth to apply TE ) , whereas encoders from `CategoricalEncodingScheme` operate on all categorical columns at once. Given that we need to apply encoders in turn: first TE to sub-set of cat columns and then apply encoder defined by _parms._categorical_encoding</del></p>


h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_add_te_model_to_a_regular_model.py
```diff
@@ -0,0 +1,82 @@
...
+    assert not pyunit_utils.compare_frames_local(preds, preds_without_te, 1, tol=1e-10, returnResult=True), "Predictions should be different"
+
+testList = [
+    test_target_encoder_attached_as_preprocessor
+]
+
+if __name__ == "__main__":
+    for test in testList: pyunit_utils.standalone_test(test)
+else:
+    for test in testList: test()
```
sebhrusen: fyi, please avoid this in tests, we had an issue with this pattern during last release: when running tests as script of from IDE, it's using the first ( __name__ == "main"), but from our test runner (in Jenkins), it was using the second....
I fixed the test runner since, but this pattern is still nasty.

I recommend to use only following instead:
```
pyunit_utils.run_tests(testList)
```

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1313,14 +1335,25 @@ else if (_parms._weights_column != null && _weights != null && !_weights.isBinar
     // Toss out extra columns, complain about missing ones, remap categoricals
     Frame va = _parms.valid();  // User-given validation set
     if (va != null) {
-      _valid = adaptFrameToTrain(va, "Validation Frame", "_validation_frame", expensive);
+      //  TODO    we are using in `adaptFrameToTrain`   _train._names which was not yet encoded
+      Frame adaptedValid = adaptFrameToTrain(va, "Validation Frame", "_validation_frame", expensive);
+      setValid(adaptedValid);
```
sebhrusen: is it to be extended in future commit? don't see any change here otherwise.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1329,7 +1362,10 @@ else if (_parms._weights_column != null && _weights != null && !_weights.isBinar
         separateFeatureVecs(); //fix up the pointers to the special vecs
       }
       if (_valid != null) {
-        _valid = encodeFrameCategoricals(_valid, ! _parms._is_cv_model /* for CV, need to score one more time in outer loop */);
+        Frame encodedVa = FrameUtils.applyTargetEncoder(teModel, va);
+        _toDelete.put(encodedVa._key, Arrays.toString(Thread.currentThread().getStackTrace()));
```
sebhrusen: logic similar as for training frame + you also need to ensure that `teModel != null` here, so maybe it's worth having a private method similar to `encodeFrameCategoricals`.

h2o-py/h2o/estimators/gbm.py
```diff
@@ -35,7 +35,7 @@ class H2OGradientBoostingEstimator(H2OEstimator):
                    "col_sample_rate_change_per_level", "col_sample_rate_per_tree", "min_split_improvement",
                    "histogram_type", "max_abs_leafnode_pred", "pred_noise_bandwidth", "categorical_encoding",
                    "calibrate_model", "calibration_frame", "custom_metric_func", "custom_distribution_func",
-                   "export_checkpoints_dir", "monotone_constraints", "check_constant_response"}
+                   "export_checkpoints_dir", "monotone_constraints", "check_constant_response", "te_model_key"}
```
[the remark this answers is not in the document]
sebhrusen: this file is auto-generated, you should not edit it manually.
For each model, the generator requests the schema for the corresponding algo and updates the `param_names` and generates the property accessors accordingly (same for the R client), so you don't need the accessors in `estimator_base.py` either.

However, for each type of model supporting TE (which means probably all of them...), you'll need to expose the property in the static `fields` property in parameters schema, see `GBMParametersV3` for example.
We should have found a way to avoid copy pasting those names, but that's the way it is today.
Although sth like this would probably work:
```
public static String[] fields = ArrayUtils.append(ModelParameters.fields, new String[] {... });
```
it would also bring some consistency to the order of those parameters (I suspect this order is then used in Flow...);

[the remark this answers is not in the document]
deil87: @sebhrusen 
>this file is auto-generated, you should not edit it manually 

thanks!  missed that

>public static String[] fields = ArrayUtils.append(ModelParameters.fields, new String[] {... });

was thinking about similar thing on client side...but once it is being autogenerated it could be probably done on backend

h2o-core/src/main/java/hex/ModelBuilder.java
deil87: @sebhrusen yes, I was going to introduce one after discussion about which approach to use `enum` vs `separate`.  This would automatically solve missing `teModel == null` check in case I'm lucky enough to copy from right place :) Thanks!

deil87: @sebhrusen fixed and covered with the tests. will introduce private method once other things are settled

h2o-core/src/main/java/hex/ModelBuilder.java
deil87: @sebhrusen artefact from "enum" approach". Will fix

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_add_te_model_to_a_regular_model.py
deil87: @sebhrusen thanks for info, good to know. Noticed that IDE will not recognise starting point ( no green triangle on the side) with new approach - fortunately with right-click context menu it is still working.

sebhrusen: yes, 'run' and 'debug' shortcuts also working afaik

h2o-py/h2o/estimators/gbm.py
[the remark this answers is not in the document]
deil87: @sebhrusen 
> Although sth like this would probably work:
public static String[] fields = ArrayUtils.append(ModelParameters.fields, new String[] {... });

did you mean `ModelParametersSchemaV3.fields` ? Field `fields` is not static and also class `ModelParametersSchemaV3` itself is in essence abstract - so I believe we don't have `shared` part of parameters, i.e. every `xxxParametersV3` defines all the parameters itself.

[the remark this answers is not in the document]
deil87: @sebhrusen noticed TargetEncoderV3 has `extractDeclaredApiParameters`, maybe I can do something like this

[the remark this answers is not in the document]
sebhrusen: Honestly, I would add it manually for now and propose this code cleanup in another ticket: whatever the approach you'll be using if you try to make this fields generation more generic, it will change the order of the fields, which is very likely to change the order of the fields in all Python generated code and maybe others too.
Without mentioning the potential impact on fields layout in Flow.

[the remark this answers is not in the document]
sebhrusen: > did you mean ModelParametersSchemaV3.fields 
yes, was pseudo-code, made the assumption that we would declare common fields in a static property there.
Relying on the introspection and `@API` annotation like it's done with `extractDeclaredApiParameters` makes more sense imo, but to ensure that the fields in Flow appear exactly in the same order, this would probably require adding an additional attribute to this annotation (sth, like `order` or `position` maybe?).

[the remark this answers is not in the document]
deil87: @sebhrusen I definitely don't want to "touch" much of this in this PR. But I still think that adding fields manually on a backend will/might affect an order of fields on clients. Not sure how exactly they are being sorted, maybe just with accordance to a "natural" order in the code.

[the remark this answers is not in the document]
sebhrusen: I don't know either tbh, try one or two and see :)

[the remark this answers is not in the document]
deil87: @sebhrusen btw, I've executed gen_python.py and `H2OXGBoostEstimator` got removed from some files `__init__.py` and `modelling.rst`.... it looks suspicious, am I missing something?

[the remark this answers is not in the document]
deil87: @sebhrusen > btw, I've executed gen_python.py and `H2OXGBoostEstimator` got removed from some files

Probably I figured out why... it is just because XGBoost extension is not enabled in my local setup

h2o-py/h2o/estimators/aggregator.py
```diff
@@ -366,6 +366,21 @@ def export_checkpoints_dir(self, export_checkpoints_dir):
...
+        Key of TargetEncoderModel
+
+        Type: ``str``.
+        """
+        return self._parms.get("te_model_id")
+
+    @te_model_id.setter
+    def te_model_id(self, te_model_id):
+        assert_is_type(te_model_id, None, str, H2OEstimator)
+        self._parms["te_model_id"] = te_model_id
```
sebhrusen: if you need a more specific validation here, there's a hook in the code generator for this.
in `bin/custom/python/gen_defaults.py`, you can add:
```
overrides=dict(
    te_model_id=dict(
        setter="""
assert_is_type(te_model_id, None, str, H2OTargetEncoderEstimator)
self._parms["te_model_id"] = te_model_id.key if isinstance(te_model_id, H2OEstimator) else te_model_id
""")
)
```

the same should have been done for `checkpoints` instead of the hardcoded logic in the abstract `estimator_base`.

h2o-core/src/main/java/water/util/FrameUtils.java
```diff
@@ -111,6 +111,32 @@ public static Frame categoricalEncoder(Frame dataset, String[] skipCols, Model.P
...
+    String[] predictorColumns = ArrayUtils.difference(teModel._parms.train().names(), teModel._parms._ignored_columns);
+    final String[] encodedColumns = ArrayUtils.remove(ArrayUtils.remove(predictorColumns, teModel._parms._fold_column), teModel._parms._response_column);
+
+    // We need to rearrange columns ( as some logic relies on response being a last column etc)
+    final String TE_ENCODED_COLUMN_POSTFIX = "_te";
+    Arrays.stream(encodedColumns).forEach(ec -> {
+      transformed
+        .swap(transformed.find(ec), transformed.find(ec + TE_ENCODED_COLUMN_POSTFIX));
+      DKV.put(transformed);
+      transformed.remove(ec).remove();
```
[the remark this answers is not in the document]
deil87: @sebhrusen 
>doesn't remove completely remove the Vec from DKV?

which one? there is a chain of two `remove` methods :)

I guess you are talking about the last one. In my case we call `score` method https://github.com/h2oai/h2o-3/pull/4362/files/20a2985fefc9319ede45d3b8ca5a53a8c3fe143e#diff-d63531c6c93a765957616530e4121ef1R115 which uses BigScore task and looks like it always returns totally new frame. I can add check in tests that original frame was not affected.

h2o-core/src/main/java/water/util/FrameUtils.java
```diff
@@ -111,6 +111,32 @@ public static Frame categoricalEncoder(Frame dataset, String[] skipCols, Model.P
...
+    // following manipulations with columns is a responsibility of a future pipeline concept. Support for input output columns for the models would be helpful as well.
+    String[] predictorColumns = ArrayUtils.difference(teModel._parms.train().names(), teModel._parms._ignored_columns);
+    final String[] encodedColumns = ArrayUtils.remove(ArrayUtils.remove(predictorColumns, teModel._parms._fold_column), teModel._parms._response_column);
+
+    // We need to rearrange columns ( as some logic relies on response being a last column etc)
+    final String TE_ENCODED_COLUMN_POSTFIX = "_te";
+    Arrays.stream(encodedColumns).forEach(ec -> {
+      transformed
+        .swap(transformed.find(ec), transformed.find(ec + TE_ENCODED_COLUMN_POSTFIX));
+      DKV.put(transformed);
```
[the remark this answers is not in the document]
deil87: @sebhrusen 
I see your concerns. Let's discuss and see what could be improved here.
**The main reason why this logic is in utility class is that we need to call it from two different places : ModelBuilder ( when training) and Model ( when scoring).**

Step by step:
1) score creates the frame copy -  I assume we are not going to change it.
2) applyTEIfModelAvailable calls applyTargetEncoder with the copy - yes, but just shallow copy.
3) 
> (what started to track them?) 

I agree, this Scope.track thing is hard to track :)
My understanding is that in non-CV case we are covered by L261 ModelBuilder (see comment L256). In case of CV we enter at L498 and exit at L547 of ModelBuilder. 

But the main "source of inspiration" is a call of `encodeFrameCategoricals` and especially comment at L1371 ModelBuilder. It is using Scope.track and assumes it should be available. The reason why it uses `toDelete` is different - we don't want to delete data too early( but we can, theoretically ). Since `FrameUtils.applyTargetEncoder` calls are analogous to `encodeFrameCategoricals` I assumed we can use Scope as well. 

Also L1173 of Model.java says "Created Vecs have to be deleted by the caller (e.g. Scope.enter/exit)" but I still not sure about having `Scope.untrack(transformed.keys());` in case we call it from Model.java as there is no Scope.enter in place. I will try to think more about this case.

> puts the transformed frame in DKV (or was it just a DKV update after the swaps?)

yes
4) applyTEIfModelAvailable tracks the transformed keys for deletion - i think this is fine. We could have used Scope.track but we can't guaranty that in all use cases Scope.enter is being called.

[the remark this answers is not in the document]
deil87: > can't you update the DKV only once after the loop?

@sebhrusen yeap, thanks, we can do that

h2o-py/h2o/estimators/aggregator.py
deil87: @sebhrusen My understanding is that ModelKeyV3 on backend results in generated client's code for  assertion only against H2OEstimator. As you said, for more specific validation ( against  H2OTargetEncoderEstimator let say) we can use this hook that you mentioned. I can't understand a couple of things:
1) I tried to add snippet to `gen_defaults.py` but execution of `gen_python.py` would not generate any new code. Shouldn't it? I was not sure about a proper syntax in `gen_defaults.py` and just made some changes in existing string literals.... rerun generator and nothing changed. What am I missing here?
2) What would be an alternative for R side? Do we have hooks there?
3) Actually I don't see  even more generic validation to be working when I'm passing dummy string instead of key:
```
H2OGradientBoostingEstimator(...., te_model_id = "absdefg")

@te_model_id.setter
    def te_model_id(self, te_model_id):
        assert_is_type(te_model_id, None, str, H2OEstimator)
```
is not failing for me.

sebhrusen: @deil87 I had to apply a small fix to `gen_python.py`: do you want me to push it your branch?
For R, there's no equivalent unfortunately.

sebhrusen: or I can give you a patch

deil87: @sebhrusen I will take a patch :)  if there is no equivalent for R, probably it is better to leave python version without a hook as well.

deil87:
@sebhrusen 
> We're trying to keep feature parity and get the APIs as close as possible, but for the internals, it's a different thing.

I thought that validation is not something that could be seen as internal. Users will have different experiences depending on client then.

sebhrusen:
@deil87 I don't understand your concern, it's not as if the contract was different in both clients if you can add proper validation only on Python client.
What will happen is user is setting a wrong `te_model_id` is that the Python client will fail early with a relevant message, whereas R client will handle only basic cases and the backend will fail later with a not-so-nice message.
But the contract remains the same in both clients.

I don't understand why you would prevent Python users from having a better early validation just because you can't provide it on R client.

deil87:
> @deil87 I don't understand your concern, it's not as if the contract was different in both clients if you can add proper validation only on Python client.
> What will happen is user is setting a wrong `te_model_id` is that the Python client will fail early with a relevant message, whereas R client will handle only basic cases and the backend will fail later with a not-so-nice message.
> But the contract remains the same in both clients.
> 
> I don't understand why you would prevent Python users from having a better early validation just because you can't provide it on R client.

@sebhrusen I just prefer real parity - when behaviour is as close as possible. Also the way how this hook is done does not look too attractive to me( code in the string literal; will add complexity for newcomers). Anyway, given your opinion I will add validation for Py.

h2o-py/h2o/estimators/xgboost.py
```diff
@@ -742,6 +742,21 @@ def categorical_encoding(self, categorical_encoding):
...
+        """
+        Key of TargetEncoderModel
+
+        Type: ``str``.
+        """
+        return self._parms.get("te_model_id")
+
+    @te_model_id.setter
+    def te_model_id(self, te_model_id):
+        assert_is_type(te_model_id, None, str, H2OEstimator)
```
deil87: Need to fix xgboost manually as locally files are not being generated when xgboost is not available

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_add_te_model_to_a_regular_model.py
```diff
@@ -25,24 +26,118 @@ def test_target_encoder_model_summary_does_not_contain_fold_column():
...
+    myX = ["pclass", "sex", "age", "sibsp", "parch", "fare", "cabin", "embarked", "home.dest"]
+    titanic_gbm_model = H2OGradientBoostingEstimator(ntrees=50,
+                                             learn_rate=0.1,
+                                             score_tree_interval=10,
+                                             stopping_rounds=5,
+                                             stopping_metric="AUC",
+                                             stopping_tolerance=0.001,
+                                             distribution="multinomial",
+                                             categorical_encoding="enum_limited",
+                                             te_model_id = te,
```
deil87: @sebhrusen @ledell got a question. On a backend we expect `te_model_id` to be a key ( string), that is why autogenerated documentation in Py says type should be `str`:

```
 @property
    def te_model_id(self):
        """
        Key of TargetEncoderModel

        Type: ``str``.
        """
```
On the other hand, for validation against type of model (its class) on Python side we need to pass Model, not id. This might be confusing for users ( not sure if we have convention that whenever id of model is required we can pass model itself). 
 
So below, we do check of type even though docs say it could be only string.
```
assert_is_type(te_model_id, None, H2OTargetEncoderEstimator)
```
The only place (that I found) where we accept model itself and do assertion against class is in `grid_search.py` - but this file is not autogenerated one and docs do not contradict with code.

Probably we should not validate against class or otherwise we can try to change somehow gen_python.py logic. With `str` type we will not be able to validate on client side (only by checking suffix but not sure if we want to do it like this).

h2o-py/h2o/estimators/aggregator.py
```diff
@@ -7,6 +7,7 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from h2o.estimators.estimator_base import H2OEstimator
+from h2o.estimators.targetencoder import H2OTargetEncoderEstimator
```
deil87: @sebhrusen for hook to be working we need this import as well. But we don't need import for TargetEncoder estimator for example and, theoretically, for some other models (but for now I'm trying to make all estimators to accept te_model_id parameter). Will try to exclude te in gen_python script. But if we go away from validation against type, there will be no need for this at all

deil87: @sebhrusen yeah, excluding by algo name is not a problem.

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_add_te_model_to_a_regular_model.py
deil87: @sebhrusen I added a hook for doc as well but it appends not overrides. We need overrides for docs as well.

h2o-core/src/main/java/water/api/schemas3/ModelParametersSchemaV3.java
```diff
@@ -122,6 +122,9 @@
           help = "Encoding scheme for categorical features")
   public Model.Parameters.CategoricalEncodingScheme categorical_encoding;
 
+  @API(help="Key of H2OTargetEncoderEstimator or H2OTargetEncoderEstimator", direction=API.Direction.INOUT)
```
deil87: @sebhrusen should we use here TargetEncoderModel or H2OTargetEncoderEstimator naming? Given that generated code will be used in Py it is better to use xxxEstimator naming. wdyth?

h2o-bindings/bin/gen_python.py
```diff
@@ -251,25 +253,26 @@ def extend_schema_params(param):
...
         yield ""
-        yield bi.wrap(property_doc, indent=8*' ')
-        custom_property_doc = get_customizations_for(algo, "doc.{}".format(pname))
+        custom_property_doc = get_customizations_or_defaults_for(algo, "doc.{}".format(pname))
         if custom_property_doc:
-            yield ""
             yield reformat_block(custom_property_doc, 8)
-        property_examples = get_customizations_for(algo, "examples.{}".format(pname))
+        else:
+            yield bi.wrap(property_doc, indent=8*' ')
```
deil87: @sebhrusen this is a way how we can replace property_doc with custom_property_doc. Once we discuss what API will be given to a user, it will become clearer whether we need these hooks or not. At least there is a way to hook docs as well.

h2o-py/tests/testdir_algos/automl/target_encoding/pyunit_add_te_model_to_a_regular_model.py
deil87: @sebhrusen figured this out. It does not append.... there is also a help section that comes before docs.

sebhrusen: You're talking about getters or setters here?
Your `te_model_id` will always be a `str`, right?
Currently, for some unknown reason, there's no docstring generated for setters, but you can add one if you need it:
in `gen_python.py`,
just change 
``` 
custom_property_doc = get_customizations_for(algo, "doc.{}".format(pname))
```
to
```
        custom_property_doc = get_customizations_or_defaults_for(algo, "doc.getter.{}".format(pname))
```
and a few lines below, you can add a similar hook for setter.

sebhrusen: actually it may not even be necessary, you can just add the doc string in your existing setter hook.

sebhrusen: also, you were complaining about code being represented as strings.
Yes, that's how the code generator works and has always been working.
But for Python, it's a bit more clever now, this should work:
```
def te_model_id_setter():
    """ your docstring"""
    your_python_code

overrides=dict(
    te_model_id=dict(
          setter=te_model_id_setter
    )
)

```

sebhrusen: no, this shouldn't work out of the box, but it should if you call `code_as_str` :
```
property_getter = code_as_str(get_customizations_for(algo, "overrides.{}.getter".format(pname)))
```
same for setter

ledell: You could follow whatever conventions we have already established with frames.  For all `training_frame` etc arguments, you can pass either an H2OFrame or an ID... so it's the same idea here.

deil87: @ledell ok, as I thought. Will rename `te_model_id` to `te_model` then.

deil87: @sebhrusen 
> also, you were complaining about code being represented as strings.
Yes, that's how the code generator works and has always been working.

sure, but mostly we generate code dynamically and this hook is fixed/hardcoded and also it is not super easy to find from `gen_python.py`.  But anyway, that's fine.

> You're talking about getters or setters here?

I was talking about setter. If we allow only strings ( keys of models) then there is probably no need for hooks at all. But I will try to keep ability to pass model itself.

h2o-core/src/main/java/water/api/schemas3/ModelParametersSchemaV3.java
ledell:
```suggestion
  @API(help="Key of Target Encoder model", direction=API.Direction.INOUT)
```

ledell: These API descriptions should not be language specific -- they are used in Java, Pythom & R (and Flow?).  So given that, I'd just call it "Target Encoding model".

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoder.java
```diff
@@ -544,7 +544,7 @@ public Frame applyTargetEncoding(Frame data,
      * @param data dataset that will be used as a base for creation of encodings .
      * @param targetColumnName name of the column with respect to which we were computing encodings.
      * @param columnToEncodingMap map of the prepared encodings with the keys being the names of the columns.
-     * @param dataLeakageHandlingStrategy see TargetEncoding.DataLeakageHandlingStrategy //TODO use common interface for stronger type safety.
+     * @param dataLeakageHandlingStrategy see TargetEncoder.DataLeakageHandlingStrategy //TODO use common interface for stronger type safety.
```
ledell: @deil87 Is this a name change or just fixing a typo?

h2o-core/src/main/java/water/api/schemas3/ModelParametersSchemaV3.java
deil87: @ledell you are right, thanks! But looks like this api will go away

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoder.java
deil87: @ledell this is just fixing a typo

deil87:
@michalkurka could you please take a look at this one?

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -146,6 +146,23 @@ public static boolean haveMojo(String urlName) {
...
+  }
+
+  /**
+   *
+   * @return TE model's key if one was assigned to a model builder, null otherwise
+   */
+  public Key<Model> getTEModelKey() {
+    return _parms._te_model;
+  }
+
```
sebhrusen, line 165: as this is supposed to be temporary, can we make those methods "less public" as suggested by @michalkurka ?
same with `deleteTEModel` in `Model` class and the method in `FrameUtils`.
Suggestion is to use naming convention: prefix methods with "_" or "internal".

deil87: @sebhrusen  I thought we were talking about Py and R API's. But probably you are right and it extends to Java API as well. It's just the fact that we don't provide docs for Java api makes me feel that Java api is not for customers/users. Thanks, will change naming.

sebhrusen: totally agree, but still safer with an explicit naming convention.

sebhrusen: and some few customers use the Java API directly, without mentioning those who extend/fork the project (see internal projects at H2O).

h2o-core/src/main/java/water/api/schemas3/ModelOutputSchemaV3.java
```diff
@@ -61,6 +61,9 @@
   @API(help="Cross-validation model metrics summary", direction=API.Direction.OUTPUT, level=API.Level.critical)
   public TwoDimTableV3 cross_validation_metrics_summary;
 
+  @API(help="TargetEncoderModel key", direction=API.Direction.OUTPUT, level=API.Level.critical)
+  public KeyV3.ModelKeyV3 te_model;
```
deil87, line 65: @sebhrusen I assume that for output's fields we are ok without `internal` prefix

PR #4363, opened 2020-02-25
Experiment: cleaning up DKV after each test (instead of test class)

sebhrusen:


sebhrusen:
experiment didn't even pass the first checks: `GLMBasic___` tests share data between test methods, therefore a special rule is required for those cases.

PR #4365, opened 2020-02-25
Collect a set of local keys on each node before each test to compare against after the test ran. This isolates the test leaks & only one test fails.

Pscheidl:
This isolates the test failures by collecting a set of local keys before each test is actually ran on each node. The comparison afterwards is done against such keys. 

After each test class, the DKV is still completely cleared.

Screenshot from 2020-02-25 20-42-49

### DIFF
--- a/h2o-core/src/test/java/water/runner/H2ORunner.java
+++ b/h2o-core/src/test/java/water/runner/H2ORunner.java
@@ -1,6 +1,7 @@
 package water.runner;
 
 import org.junit.AfterClass;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Ignore;
 import org.junit.internal.AssumptionViolatedException;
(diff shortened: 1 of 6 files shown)


h2o-core/src/test/java/water/runner/H2ORunner.java
```diff
@@ -41,16 +42,11 @@ public H2ORunner(Class<?> klass) throws InitializationError {
...
-    protected Statement withBeforeClasses(Statement statement) {
-        List<FrameworkMethod> befores = getTestClass()
-                .getAnnotatedMethods(BeforeClass.class);
-        return befores.isEmpty() ? statement :
-                new H2ORunnerBefores(statement, befores, null);
+    @Override
+    protected Statement withBefores(FrameworkMethod method, Object target, Statement statement) {
+        List<FrameworkMethod> befores = getTestClass().getAnnotatedMethods(
+                Before.class);
+        return new H2ORunnerBefores(statement, befores, null);
```
sebhrusen: cz

sebhrusen: could be useful if we want to progressively switch to this runner instead of the messy-do-too-may-things `TestUtil`

Pscheidl: Yes, we can do that. What's the reason to isolate only per-class ? Btw. I belive both is doable - check leaks per test and then check per-class leaks (@Beforeclass intializes some keys and @Afterclass does not do the cleanup).

I have an idea how to easily to per-class leak detection, but it will take a while longer and I'm in the middle of more important task, this can be however done in few hours.

honzasterba: I think we should strive to make each test run in complete isolation and we should force ourlseves to do so. I dont think making it easier to have messy tests will help anything in the long run.

sebhrusen: @honzasterba not encouraging anyone to use "class level isolation", just thought it would make like easier if we plan to migrate legacy tests to this runner. Refactoring tests doesn't bring much value, but at least with an annotation or a rule, we would immediately see the isolation level of those legacy tests.

PR #4370, opened 2020-02-28
PUBDEV-6852 - Kubernetes support

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6852

**It is recommended to read the README.md** introduced in this PR.

Screenshot from 2020-02-29 14-12-26

Works even locally on minikube:

Screenshot from 2020-02-29 21-12-22

### Example output of a Pod
```
02-29 13:09:06.927 10.129.20.65:54321    10     main      INFO: ----- H2O started  -----
02-29 13:09:06.945 10.129.20.65:54321    10     main      INFO: Build git branch: pavel/pubdev-6852
02-29 13:09:06.945 10.129.20.65:54321    10     main      INFO: Build git hash: e6623b22c55e5f43359348efeaac94bfe2f4433d
02-29 13:09:06.945 10.129.20.65:54321    10     main      INFO: Build git describe: jenkins-3.28.0.4-11-ge6623b22c5-dirty
02-29 13:09:06.946 10.129.20.65:54321    10     main      INFO: Build project version: 3.28.0.99999
02-29 13:09:06.946 10.129.20.65:54321    10     main      INFO: Build age: -55 minutes
02-29 13:09:06.946 10.129.20.65:54321    10     main      INFO: Built by: 'pavel'
02-29 13:09:06.946 10.129.20.65:54321    10     main      INFO: Built on: '2020-02-29 14:04:48'
02-29 13:09:06.947 10.129.20.65:54321    10     main      INFO: Found H2O Core extensions: [XGBoost, KrbStandalone]
02-29 13:09:06.948 10.129.20.65:54321    10     main      INFO: Processed H2O arguments: []
02-29 13:09:06.948 10.129.20.65:54321    10     main      INFO: Java availableProcessors: 1
02-29 13:09:06.948 10.129.20.65:54321    10     main      INFO: Java heap totalMemory: 12.9 MB
02-29 13:09:06.949 10.129.20.65:54321    10     main      INFO: Java heap maxMemory: 123.8 MB
02-29 13:09:06.949 10.129.20.65:54321    10     main      INFO: Java version: Java 11.0.6 (from Ubuntu)
02-29 13:09:06.949 10.129.20.65:54321    10     main      INFO: JVM launch parameters: []
02-29 13:09:06.949 10.129.20.65:54321    10     main      INFO: JVM process id: 10@example-0
02-29 13:09:06.950 10.129.20.65:54321    10     main      INFO: OS version: Linux 4.18.0-147.5.1.el8_1.x86_64 (amd64)
02-29 13:09:06.950 10.129.20.65:54321    10     main      INFO: Machine physical memory: 31.41 GB
02-29 13:09:06.950 10.129.20.65:54321    10     main      INFO: Machine locale: en_US
02-29 13:09:06.951 10.129.20.65:54321    10     main      INFO: X-h2o-cluster-id: 1582981745644
02-29 13:09:06.951 10.129.20.65:54321    10     main      INFO: User name: '1002280000'
02-29 13:09:06.951 10.129.20.65:54321    10     main      INFO: IPv6 stack selected: false
02-29 13:09:06.951 10.129.20.65:54321    10     main      INFO: Possible IP Address: eth0 (eth0), fe80:0:0:0:80f0:dcff:fef8:1dcd%eth0
02-29 13:09:06.951 10.129.20.65:54321    10     main      INFO: Possible IP Address: eth0 (eth0), 10.129.20.65
02-29 13:09:06.952 10.129.20.65:54321    10     main      INFO: Possible IP Address: lo (lo), 0:0:0:0:0:0:0:1%lo
02-29 13:09:06.954 10.129.20.65:54321    10     main      INFO: Possible IP Address: lo (lo), 127.0.0.1
02-29 13:09:06.954 10.129.20.65:54321    10     main      INFO: H2O node running in unencrypted mode.
02-29 13:09:06.956 10.129.20.65:54321    10     main      INFO: Internal communication uses port: 54322
02-29 13:09:06.956 10.129.20.65:54321    10     main      INFO: Listening for HTTP and REST traffic on http://10.129.20.65:54321/
02-29 13:09:06.960 10.129.20.65:54321    10     main      INFO: Initializing H2O Kubernetes cluster
02-29 13:09:07.032 10.129.20.65:54321    10     main      INFO: Timeout for node discovery is set to 120 seconds.
02-29 13:09:07.033 10.129.20.65:54321    10     main      INFO: Desired cluster size is set to 3 nodes.
02-29 13:09:07.120 10.129.20.65:54321    10     main      INFO: New H2O pod with DNS record 'example-0.h2o-service.h2o-statefulset.svc.cluster.local./10.129.20.65' discovered.
02-29 13:09:21.199 10.129.20.65:54321    10     main      INFO: New H2O pod with DNS record 'example-1.h2o-service.h2o-statefulset.svc.cluster.local./10.131.10.165' discovered.
02-29 13:09:52.344 10.129.20.65:54321    10     main      INFO: New H2O pod with DNS record 'example-2.h2o-service.h2o-statefulset.svc.cluster.local./10.130.8.110' discovered.
02-29 13:09:53.345 10.129.20.65:54321    10     main      INFO: Using the following pods to form H2O cluster: [10.131.10.165,10.130.8.110,10.129.20.65]
02-29 13:09:53.346 10.129.20.65:54321    10     main      INFO: Dynamically loaded 'H2OKubernetesEmbeddedConfigProvider' as AbstractEmbeddedH2OConfigProvider.
02-29 13:09:53.352 10.129.20.65:54321    10     main      INFO: H2O cloud name: '1002280000' on /10.129.20.65:54321, discovery address /234.145.114.231:60049
02-29 13:09:53.353 10.129.20.65:54321    10     main      INFO: If you have trouble connecting, try SSH tunneling from your local machine (e.g., via port 55555):
02-29 13:09:53.353 10.129.20.65:54321    10     main      INFO:   1. Open a terminal and run 'ssh -L 55555:localhost:54321 1002280000@10.129.20.65'
02-29 13:09:53.353 10.129.20.65:54321    10     main      INFO:   2. Point your browser to http://localhost:55555
02-29 13:09:54.128 10.129.20.65:54321    10     main      INFO: Log dir: '/tmp/h2o-1002280000/h2ologs'
02-29 13:09:54.128 10.129.20.65:54321    10     main      INFO: Cur dir: '/'
02-29 13:09:54.136 10.129.20.65:54321    10     main      INFO: Subsystem for distributed import from HTTP/HTTPS successfully initialized
02-29 13:09:54.136 10.129.20.65:54321    10     main      INFO: HDFS subsystem successfully initialized
02-29 13:09:54.139 10.129.20.65:54321    10     main      INFO: S3 subsystem successfully initialized
02-29 13:09:54.209 10.129.20.65:54321    10     main      INFO: GCS subsystem successfully initialized
02-29 13:09:54.211 10.129.20.65:54321    10     main      INFO: Flow dir: '//h2oflows'
02-29 13:09:54.237 10.129.20.65:54321    10     main      INFO: Cloud of size 1 formed [example-0.h2o-service.h2o-statefulset.svc.cluster.local/10.129.20.65:54321]
02-29 13:09:54.237 10.129.20.65:54321    10     main      INFO: Created cluster of size 1, leader node IP is 'example-0.h2o-service.h2o-statefulset.svc.cluster.local/10.129.20.65'
02-29 13:09:54.251 10.129.20.65:54321    10     main      INFO: Registered parsers: [GUESS, ARFF, XLS, SVMLight, AVRO, PARQUET, CSV]
02-29 13:09:54.255 10.129.20.65:54321    10     main      INFO: XGBoost extension initialized
02-29 13:09:54.256 10.129.20.65:54321    10     main      INFO: KrbStandalone extension initialized
02-29 13:09:54.312 10.129.20.65:54321    10     main      INFO: Registered 2 core extensions in: 475ms
02-29 13:09:54.313 10.129.20.65:54321    10     main      INFO: Registered H2O core extensions: [XGBoost, KrbStandalone]
02-29 13:09:54.932 10.129.20.65:54321    10     main      INFO: Found XGBoost backend with library: xgboost4j_minimal
02-29 13:09:54.932 10.129.20.65:54321    10     main      WARN: Your system supports only minimal version of XGBoost (no GPUs, no multithreading)!
02-29 13:09:55.233 10.129.20.65:54321    10     main      INFO: Registered: 187 REST APIs in: 920ms
02-29 13:09:55.233 10.129.20.65:54321    10     main      INFO: Registered REST API extensions: [Amazon S3, XGBoost, Algos, AutoML, Core V3, TargetEncoder, Core V4]
02-29 13:09:55.707 10.129.20.65:54321    10     main      INFO: Registered: 279 schemas in 474ms
02-29 13:09:55.707 10.129.20.65:54321    10     main      INFO: H2O started in 50056ms
02-29 13:09:55.707 10.129.20.65:54321    10     main      INFO: 
02-29 13:09:55.708 10.129.20.65:54321    10     main      INFO: Open H2O Flow in your web browser: http://10.129.20.65:54321
02-29 13:09:55.708 10.129.20.65:54321    10     main      INFO: 
02-29 13:09:57.764 10.129.20.65:54321    10     FJ-126-1  INFO: Cloud of size 3 formed [example-0.h2o-service.h2o-statefulset.svc.cluster.local/10.129.20.65:54321, example-2.h2o-service.h2o-statefulset.svc.cluster.local/10.130.8.110:54321, example-1.h2o-service.h2o-statefulset.svc.cluster.local/10.131.10.165:54321]
02-29 13:09:57.765 10.129.20.65:54321    10     FJ-126-1  INFO: Created cluster of size 3, leader node IP is 'example-0.h2o-service.h2o-statefulset.svc.cluster.local/10.129.20.65'

```


settings.gradle
```diff
@@ -100,3 +101,5 @@ if (System.getProperty("user.name").equals("jenkins")
     }
   }
 }
+include 'h2o-k8s'
```
pkozelka: isn't this a duplicate?

Pscheidl:
Memory limits should be handled on JVM level and are the container's responsibility ... `java -jar h2o.jar -Xmx...`

h2o-core/src/main/java/water/H2O.java
```diff
@@ -793,8 +794,61 @@ private static void validateArguments() {
...
    * Register embedded H2O configuration object with H2O instance.
    */
-  public static void setEmbeddedH2OConfig(AbstractEmbeddedH2OConfig c) { embeddedH2OConfig = c; }
-  public static AbstractEmbeddedH2OConfig getEmbeddedH2OConfig() { return embeddedH2OConfig; }
+  public static void setEmbeddedH2OConfig(AbstractEmbeddedH2OConfig c) {
+    embeddedH2OConfig = c;
+  }
+
+  /**
+   * Returns an inststance of {@link AbstractEmbeddedH2OConfig}. The origin of the embedded config might be either
```
jakubhava: NIT: typo -> inststance

h2o-core/src/main/java/water/H2O.java
```diff
@@ -793,8 +794,61 @@ private static void validateArguments() {
...
+   * from directly setting the embeddedH2OConfig field via setEmbeddedH2OConfig setter, or dynamically provided via
+   * service loader. Directly set {@link AbstractEmbeddedH2OConfig} is always prioritized. ServiceLoader lookup is only
+   * performed if no config is previously set.
+   * <p>
+   * Result of first ServiceLoader lookup is also considered final - once a service is found, dynamic lookup is not
+   * performed any further.
+   *
+   * @return An instance of {@link AbstractEmbeddedH2OConfig}, if set or dynamically provided. Otherwise null
+   * @author Michal Kurka
```
jakubhava: I do not want to take credits from @michalkurka :) But does not seem like patter we do in our codebase

h2o-core/src/main/java/water/H2O.java
```diff
@@ -793,8 +794,61 @@ private static void validateArguments() {
...
+      final EmbeddedConfigProvider embeddedConfigProvider = optionalEmbeddedConfigProvider.get();
+      Log.info(String.format("Dynamically loaded '%s' as AbstractEmbeddedH2OConfigProvider.", embeddedConfigProvider.getName()));
+      embeddedH2OConfig = embeddedConfigProvider.getConfig();
+    }
+    return embeddedH2OConfig;
+  }
+
+  /**
+   * Uses {@link ServiceLoader} to discover active instances of {@link EmbeddedConfigProvider}. Only one provider
+   * may be active at a time. If more providers are detected, {@link IllegalStateException is thrown.}
```
jakubhava: `}` should be after IllegalStateException

h2o-k8s/README.md
```diff
@@ -0,0 +1,103 @@
+# H2O Kubernetes integration
+
+The integration of Kubernetes and H2O is possible via `water.k8s.KubernetesEmbeddedConfigProvider` - to be found
+in this module. This implementation of `EmbeddedConfigProvider` is dynamically loaded on H2O start and remains inactive
+unless H2O running in a Docker container managed by Kubernetes is detected.
```
jakubhava: H2O is running?

h2o-k8s/README.md
```diff
@@ -0,0 +1,103 @@
+# H2O Kubernetes integration
+
+The integration of Kubernetes and H2O is possible via `water.k8s.KubernetesEmbeddedConfigProvider` - to be found
+in this module. This implementation of `EmbeddedConfigProvider` is dynamically loaded on H2O start and remains inactive
+unless H2O running in a Docker container managed by Kubernetes is detected.  
+
+## Running H2O in K8s - user's guide
+
+H2O Pods deployed on  Kubernetes cluster require a
```
jakubhava: NIT: two spaces after `on`

h2o-k8s/README.md
```diff
@@ -0,0 +1,103 @@
...
+metadata:
+  name: h2o-service
+spec:
+  type: ClusterIP
+  clusterIP: None
+  selector:
+    app: h2o-k8s
+  ports:
+  - protocol: TCP
+    port: 54321
```
jakubhava: What about internal H2O port? I'm sure it works, just curious :)

jakubhava: Ah, I see, this this the port exposed to the outside word

h2o-k8s/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+1. `H2O_KUBERNETES_SERVICE_DNS` - **[MANDATORY]** Crucial for the clustering to work. The format usually follows the
+ `<service-name>.<project-name>.svc.cluster.local` pattern. This setting enables H2O node discovery via DNS.
+  It must be modified to match the name of the headless service created. Also, pay attention to the rest of the address
+  to match the specifics of your Kubernetes implementation.
+1. `H2O_NODE_LOOKUP_TIMEOUT` - **[OPTIONAL]** Node lookup constraint. Time before the node lookup is ended. 
+1. `H2O_NODE_EXPECTED_COUNT` - **[OPTIONAL]** Node lookup constraint. Expected number of H2O pods to be discovered.
+
+If none of the optional lookup constraints is specified, a sensible default node lookup timeout will be set - currently
+defaults to 3 minutes. If any of the lookup constraints is defined, the H2O node lookup is terminated on whichever
```
jakubhava: are defined

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
```diff
@@ -0,0 +1,94 @@
...
+import water.k8s.lookup.LookupConstraintsBuilder;
+import water.util.Log;
+
+import java.util.Collection;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+/**
+ * A configuration provider for H2O running in Kubernetes cluster. It is able to detected it's ran in K8S
```
jakubhava: Maybe re-word a little bit? ` It is able to detected it's ran`

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
```diff
@@ -0,0 +1,94 @@
...
+            lookupConstraintsBuilder.withTimeoutSeconds(timeoutSeconds);
+        } catch (NumberFormatException e) {
+            Log.info(String.format("'%s' environment variable not set.", K8S_NODE_LOOKUP_TIMEOUT_KEY));
+        }
+
+        try {
+            final Integer desiredClusterSize = Integer.parseInt(System.getenv(K8S_DESIRED_CLUSTER_SIZE_KEY));
+            lookupConstraintsBuilder.withDesiredClusterSize(desiredClusterSize);
+        } catch (NumberFormatException e) {
+            Log.info(String.format("'%s' environment variable not set.", K8S_DESIRED_CLUSTER_SIZE_KEY));
```
jakubhava: we could mention that we are falling back to some default value

h2o-k8s/README.md
```diff
@@ -0,0 +1,103 @@
+# H2O Kubernetes integration
```
jakubhava: Would be good to provide utility to create docker image with H2O which can be used in Kuberentes pods and mention it in the docs, WDYT @Pscheidl?

If such image is released to official docker hub in each release of H2O, it would simplify using H2O a lot ( people won't need to build their images).

Plus Sparkling Water in external backend might just download the images and run the templates mentioned in the docs to start H2O on Kubernetes

h2o-k8s/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+If none of the optional lookup constraints is specified, a sensible default node lookup timeout will be set - currently
+defaults to 3 minutes. If any of the lookup constraints is defined, the H2O node lookup is terminated on whichever 
+condition is met first.
+
+### Exposing H2O cluster
+
+Exposing the H2O cluster is a responsibility of the Kubernetes administrator. By default, an
+ [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) can be created. Different platforms offer
+ different capabilities, e.g. OpenShift offers [Routess](https://docs.openshift.com/container-platform/4.3/networking/routes/route-configuration.html).
```
jakubhava: I couldn't find in the documentation a bit where actually h2o is started, am I missing something?

jakubhava:
Nice work!, mentioned few nits and proposals

jakubhava:
One more question: are all nodes accessible on the default port from the outside word? In Sparkling Water we need all nodes to be exposed (just the API port).

When running GET request `/3/Cloud` the requests returns the IP:Port of all the H2O nodes in the cluster, are these IP:Ports accessible from outside? ( This is how SW discovers the nodes in external backend).

Question 2: Could you please try with HTTPS? I guess it should work out of box, but would be good to also mention this in the documentation

h2o-k8s/README.md
jakubhava: What we could do in SW is just to enhance these images with sparkling water extension jars and publish ours as well

Pscheidl: This is the plan. Adressing in a separate main thread, as this is important.

h2o-core/src/main/java/water/H2O.java
Pscheidl: He actually wrote this piece of code and I stole it from him in order to have it in my codebase immediately. That's why I'm giving him the credit, as he is truly the author of these lines. :)

Pscheidl:
Thank you @jakubhava. I'll address the issues in documentation ASAP. More importantly, you raised **three important questions**. I feel like these should be addressed in the main thread. 

## Access to all pods by Sparkling Water 

There are two scenarios. I managed to draw them quickly. Please excuse me for not being exactly Pablo Picasso :rofl: :rofl: . There are two scenarios I can think of we may start the discussion with.

### Services accessing H2O outside of K8S cluster

This applies to all our clients (Py, R and whatever else we have). This scenario **might include Sparkling Water**. H2O is ran on Kubernetes cluster, but accessed from the outside.

[RB 2020-03-02 10.00.37.pdf](https://github.com/h2oai/h2o-3/files/4275074/RB.2020-03-02.10.00.37.pdf)

The correct way to do this is via an Ingress/Route, that publicly exposes the H2O Stateful set. Which will point to the internal service. The internal service will then route the traffic to one of the pods. If Sparkling water requires access to all pods and runs outside of the K8S cluster, then we might have a problem here. Kubernetes is about layers of abstraction. Individual pods are not visible outside the Kubernetes cluster - as far as I know. There might be a workaround over that named `ExternalIP`, but that is highly debatable, as it makes the advantage of K8S go away. If one node dies and all pods on it die as well, the service is no longer available. And many other disadvantages, like the lack of service discovery.

### Services running on the same K8S cluster as H2O

This is much easier scenario, as the headless service will give away Pod IPs (internal IPs inside the cluster, not reachable from the outside), Sparkling Water may then just connect directly to each H2O pod, if that is truly needed.

[RB 2020-03-02 10.16.50.pdf](https://github.com/h2oai/h2o-3/files/4275132/RB.2020-03-02.10.16.50.pdf)

Whether to scale H2O and Sparkling water in one pod is debatable. I'd definitely not do it in one container. That's what kubernetes is for. I do think Sparkling Water is a separate service inside the K8S cluster, not directly tied to H2O. But I might be deadly wrong on this one @mn-mikke @jakubhava - please do correct me if the real-world scenario is to have the same amount of Spark workers as H2O nodes. Then, it would be convenient to both H2O and SW in each pod - but there are others things to consider, e.g. RAM available on K8S nodes. Our recommendations for users are one thing, but the freedom of K8S cluster owners to create their own setup within certain technical limitations seems crucial to me. Let's not hard-wire things.

## Convenience Docker images

Currently, there is `pscheidl/h2o-k8s` in the documentation. This is a working image with changes in this PR, but it's of course only for my own purpose. It should and will be replaced with official H2O K8S image. My proposal is to release new image and push do docker hub with each new release. 

Currently, the image is really...well..not sophisticated:

```Dockerfile
FROM ubuntu:latest
MAINTAINER pavel@h2o.ai

RUN ls

RUN apt update && \
	apt install openjdk-11-jdk -y

COPY build/libs/* .

CMD java -jar h2o.jar
``` 

This discussion should be carried outside of this PR, unless there is a technical barrier.

### HTTPS

There are different ways to handle HTTPS and it really depends on the Kubernetes cluster settings. It is also related to the service layout discussed in the very first topic.

When H2O is running on K8S and exposed to the outer world via an Ingress/Route, then the TLS is really managed by the K8S implementation, not by H2O. Then, there is internal cluster security. Who can see the H2O pods and the related service(s) is a matter of configuration.

Obtaining a certificate inside K8S cluster should however be supported by H2O. Ways to obtain certificates trusted inside the Kubernetes cluster are described here: https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/#requesting-a-certificate

This would probably mean:
1. Create an external script that, on command (env variable present ?), will ask for a certificate and instruct H2O using existing flags to use that certificate.
1. Adjust H2O'S K8S plugin to instruct H2O to use the certificate (and obtain it first as well). This would result in some modification to existing AbstractEmbeddedH2OConfig, as the plugin would somehow need to access whole H2O configuration and modify it (the `jks*` options).
[RB 2020-03-02 10.00.37.pdf](https://github.com/h2oai/h2o-3/files/4275071/RB.2020-03-02.10.00.37.pdf)

jakubhava:
Great write up @Pscheidl!

Actually, Sparkling Water does not need to communicate with each of the H2O nodes directly. We always communicate with the H2O leader node except for the case when we uploading data from spark frame to h2o frame or downloading data from h2o frame back to spark frame. In this case we communicate with the nodes directly. 

I guess the headless service could somehow handle this traffic forwarding right? ( I'm not Kubernetes expert)

jakubhava:
As per the mixed approach -> I think we do not need to worry about that. We are moving towards external backend anyways + it will just work as both spark & h2o would be running inside Kubernetes

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
Pscheidl: Here, we are only saying environment variable is not set. The actual lookup constraint logic says exactly that. It will tell you if it falls back to a default timeout.

h2o-k8s/README.md
Pscheidl: Thanks, I'll add more explanation. H2O is started as soon as all pods cluster. Which means the headless service is created and the H2O pods are started. The ingress only makes H2O accessible from the outside. It may even change without affecting the underlying layers.

You may spawn H2O in one go by doing `kubectl apply -f h2o.yaml`. The YAML looks like this and it will work on your local minikube as well: 

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: h2o-k8s
  labels:
    app: h2o-k8s
  namespace: default
spec:
  containers:
    - name: h2o-k8s
      image: pscheidl/h2o-k8s
      ports:
        - containerPort: 54321
      env:
      - name: H2O_KUBERNETES_SERVICE_DNS
        value: h2o-service.default.svc.cluster.local
---
apiVersion: v1
kind: Service
metadata:
  name: h2o-service
spec:
  type: ClusterIP
  clusterIP: None
  selector:
    app: h2o-k8s
  ports:
  - protocol: TCP
    port: 54321
```

Pscheidl:
If we do not need to worry about access to all H2O nodes from the outside world, then we're good.

The headless service will provide you IPs of all the H2O nodes. Sparkling Water will have this information and it's also the responsibility of Sparkling Water to contact the nodes as needed. I guess there is a way via our Rest API to know which node is the leader node.

jakubhava:
Perfect, we do not need internal access. Just access via REST API to all nodes

h2o-k8s/README.md
```diff
@@ -0,0 +1,103 @@
...
+        app: h2o-k8s
+    spec:
+      terminationGracePeriodSeconds: 10
+      containers:
+        - name: h2o-k8s
+          image: 'pscheidl/h2o-k8s'
+          ports:
+            - containerPort: 54321
+              protocol: TCP
+          env:
```
nkpng2k: nit: may want to define some sample resources here. For cpu/memory.

h2o-k8s/README.md
```diff
@@ -0,0 +1,103 @@
...
+          - name: H2O_KUBERNETES_SERVICE_DNS
+            value: h2o-service.h2o-statefulset.svc.cluster.local
```
nkpng2k: QQ: I don't entirely remember how cluster ips work as opposed to loadbalancers but I would be concerned that the above will not always be the expected behavior. In fact I may recommend defining an `ingress` object https://kubernetes.io/docs/concepts/services-networking/ingress/

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
```diff
@@ -0,0 +1,94 @@
...
+    public AbstractEmbeddedH2OConfig getConfig() {
+        return kubernetesEmbeddedConfig;
+    }
+
+    /**
+     * @return True if there are environment variables indicating H2O is running inside a container managed by
+     * Kubernetes. Otherwise false.
+     */
+    private boolean isRunningOnKubernetes() {
+        final Pattern KUBERNETES_SERVICE_HOST = Pattern.compile(".*_SERVICE_HOST");
```
nkpng2k: QQ: are these guaranteed to exist in every k8s environment?

h2o-k8s/src/main/java/water/k8s/lookup/KubernetesDnsLookup.java
```diff
@@ -0,0 +1,159 @@
...
+ *             - containerPort: 54322
+ *               protocol: TCP
+ *           env:
+ *           - name: H2O_KUBERNETES_SERVICE_DNS
+ *             value: h2o-service.h2o-statefulset.svc.cluster.local
+ * </pre>
+ */
+public class KubernetesDnsLookup implements KubernetesLookup {
+
+    private static final String H2O_K8S_SERVICE_DNS_KEY = "H2O_KUBERNETES_SERVICE_DNS";
```
nkpng2k: nit: this is a bit confusing... at least to me...

Pscheidl: If it's confusing, then it's not nitpicking. I'll fix that !

h2o-k8s/src/main/java/water/k8s/lookup/KubernetesDnsLookup.java
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Discovery strategy on a DNS cluster leveraging the DNS record of a Kubernetes headless service present in the cluster.
+ * Kubernetes headless services, instead of load-balancing the requests onto one of the underlying pods, return the
+ * addresses of all the pods covered by the headless service.
+ * <p>
+ * Such pods can then be discovered via the above-mentioned DNS record. In order for H2O to know which service to query,
+ * it is mandatory for the K8S user to pass the name of the headless service to the H2O container, as follows:
+ *
+ * <pre>
+ * apiVersion: apps/v1
```
nkpng2k: nit: this is different from the readme. Specifically, open ports in container and also naming. Not sure if it matters, but this would maybe benefit from consistency.

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final String K8S_NODE_LOOKUP_TIMEOUT_KEY = "H2O_NODE_LOOKUP_TIMEOUT";
+    private static final String K8S_DESIRED_CLUSTER_SIZE_KEY = "H2O_NODE_EXPECTED_COUNT";
+
+    private boolean runningOnKubernetes = false;
+    private KubernetesEmbeddedConfig kubernetesEmbeddedConfig;
+
+    /**
+     * @return A Set
+     */
+    private static final Optional<Set<String>> resolveNodeIPs() {
```
nkpng2k: QQ: when we say `resolveNodeIPs` do we mean the internal ip address of each pod spun up for the h2o cluster in kubernetes or the actual physical kubernetes nodes. From a naming perspective this is potentially incorrect in the instance of the former, and could lead to confusion

Pscheidl: Yes, I'll rename it. The exact answer is: It only discovers pod IPs listed by the hadless service created. And the headless service only filters out (at least it should) pods with H2O. Therefore, this only discovers H2O nodes.

h2o-k8s/src/main/java/water/k8s/lookup/KubernetesDnsLookup.java
nkpng2k: well it made sense the second time I looked at it. So I figured it was a little nitpicky.

h2o-k8s/README.md
jakubhava: Cool, makes sense to to this separately. Thanks!

h2o-k8s/README.md
jakubhava: Nice!

Pscheidl:
No, it's not. The statement I wrote above negates this one. This is possible, but has to be done kind of manually by setting up a service and an ingress for every pod. And it takes away the K8S magic.

As I said above, H2O nodes are easily available once SW is running in the same K8S cluster.

jakubhava:
@Pscheidl so how python clients connect to the cluster?

Pscheidl:
Normally, I use `h2o.init()`.

```python
    # Not strictly checking versions, as there is an often changing dev build in the cluster
    h2o.init(url="http://test-serv-h2o-statefulset.apps.us-east-1.starter.openshift-online.com:80",
                   strict_version_check=False)
```
Works for me.

Screenshot from 2020-03-03 08-21-27

jakubhava:
Let's discuss it on the call later today. Not having this functionality is blocker for SW

Pscheidl:
Agreed. Let's have the call first. I guess & hope this is simply misunderstanding on both sides and it will be doable.

jakubhava:
Was thinking a bit more. It's true that in this setup SW won't work & I would actually like if we can work just with a single IP (ie: do not require direct access to all H2O nodes), just describing the current state.

We might be able to implement this, but would require change(probably small) of our internal conversion protocol @mn-mikke -> the requests would all come to the single node but we would send the uploadPlan as one of the REST parameters. So actually the node (let's say leader) would then decide where to put chunks. Not Sparkling Water. This way we can expect open Flow just from leader node.

All this could be doable on SW side

h2o-k8s/README.md
Pscheidl: This is highly individual. I'll mention the need to adjust these settings as well, but will add them.

Pscheidl:
To sum up:

- Sparkling water people are fine with current status @jakubhava @mn-mikke 
- Not only related to SW, but this is considered to be the basic building block. We should monitor the requirements of our customers and enhance the Kubernetes support.
- HTTPS communication from the outside  to internal cluster services (SW, H2O) can be secured - that is the responsibility of the cluster administration and we have no power over that. This is not to be done on our side.
- Internal secured communication is probably not required at the moment, but the requirements might come. There are two ways to handle it, both described in the [documentation section](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/).  First way is ship the certificate inside the container (or mount a volume with the container itself) and then configure K8S to trust the certificate. This can be done by the cluster administrator and requires their own packaging of H2O or other software into containers. Which might be the case for some users. Second is to teach H2O how to generate a certificate and  ask Kubernetes to sign it. This requires API access and probably role bindings set-up (not investigated any further). If this is required, I'd do it in a separate PR.

These are the notes I took during today's meeting. Is there anything I forgot @jakubhava @mn-mikke ?

If no, then we need to talk about the reference image @michalkurka - is that viable option or not ? Make it part of our builds ? Also, I'll ship the current "test" image in my public Dockerhub to a customer who is eager to try it out in their own OpenShift environment.

michalkurka:
> If no, then we need to talk about the reference image @michalkurka - is that viable option or not ? Make it part of our builds ? Also, I'll ship the current "test" image in my public Dockerhub to a customer who is eager to try it out in their own OpenShift environment.

it makes sense to have an image like that - but we do need to figure out how to test this properly as well (on an ongoing basis)

Pscheidl:
Time for the code to be reviewed I guess. @honzasterba please rip & tear ! :rofl:

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
Pscheidl: According to documentation, the answer is yes: https://kubernetes.io/docs/concepts/services-networking/service/#discovering-services

Pscheidl: This is very K8S-specific and looks like a good indicator that H2O is running on K8S cluster, unless a better one is found.

Pscheidl:
Following actions:

- Automatic image creation
- Testability in our Jenkins pipeline - two phases. First is to only smoke test the image - https://k3s.io/ should be tested. Second phase seems to be running all H2O tests against H2O pods spawned on K8S (https://rancher.com/ could be useful).

h2o-core/src/main/java/water/H2O.java
```diff
@@ -793,8 +794,61 @@ private static void validateArguments() {
...
+   * @author Michal Kurka
+   */
+  public static AbstractEmbeddedH2OConfig getEmbeddedH2OConfig() {
+    if (embeddedH2OConfig != null) {
+      return embeddedH2OConfig;
+    }
+
+    final Optional<EmbeddedConfigProvider> optionalEmbeddedConfigProvider = discoverEmbeddedConfigProvider();
+
+    if (optionalEmbeddedConfigProvider.isPresent()) {
```
honzasterba: would be nicer with map().orElse()

h2o-k8s/README.md
honzasterba: since the general recommendation is min 4GB per node I would stick to that here

h2o-k8s/README.md
```diff
@@ -0,0 +1,106 @@
...
+      app: h2o-k8s
+  template:
+    metadata:
+      labels:
+        app: h2o-k8s
+    spec:
+      terminationGracePeriodSeconds: 10
+      containers:
+        - name: h2o-k8s
+          image: 'pscheidl/h2o-k8s'
```
honzasterba: do you plan to update this to the h2oai/ namespaced image? also the image name should mention h2o-3

Pscheidl: Yes. That must be done.

h2o-core/src/main/java/water/H2O.java
michalkurka: no credits needed :)

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfig.java
```diff
@@ -0,0 +1,57 @@
...
+        return true;
+    }
+
+    @Override
+    public String fetchFlatfile() {
+        return flatfile;
+    }
+
+    @Override
+    public void exit(int status) {
```
michalkurka: Is this correct?

h2o-k8s/src/main/java/water/k8s/lookup/KubernetesDnsLookup.java
```diff
@@ -0,0 +1,164 @@
...
+     */
+    public Optional<Set<String>> lookupNodes(final Collection<LookupConstraint> lookupConstraints) {
+        final Set<String> lookedUpNodes = new HashSet<>();
+
+        while (lookupConstraints.stream().allMatch(lookupStrategy -> !lookupStrategy.isLookupEnded(lookedUpNodes))) {
+            try {
+                dnsLookup(lookedUpNodes);
+                Thread.sleep(ONE_SECOND);
+            } catch (NamingException e) {
+                continue;
```
michalkurka: can we produce a warning?

h2o-app/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
   compile project(":h2o-algos")
   compile project(":h2o-automl")
   compile project(":h2o-ext-target-encoder")
+  compile project(":h2o-k8s")
```
michalkurka: This should IMHO be in h2o-assemblies/main instead. Otherwise, the code would be bundled in Hadoop distributions as well. Not the worst thing (it should be inactive even if the Hadoop node was running on k8s) but still - if there is no need to bundle it, I would avoid it.

h2o-k8s/README.md
michalkurka: please, remove all mentions of `pscheidl/h2o-k8s` from the PR

michalkurka: (you can put there something like "<user your custom image>") for the time being

h2o-k8s/README.md
michalkurka: that is a good suggestion

h2o-core/src/main/java/water/H2O.java
Pscheidl: Changed. Nicely reveals how probably even this getter could output Optional - as it's often null. 

@michalkurka Wanted to incorporate some changes himself as well.

h2o-app/build.gradle
Pscheidl: Moved to main assembly. Removed from here.

h2o-k8s/README.md
Pscheidl: This is the port exposed to other K8S entities. The final exposure to the outside world is done by another layer named Ingress or Route (on OpenShift).

h2o-k8s/README.md
Pscheidl: Done, now there is `<someDockerImageWithH2OInside>`.

h2o-k8s/README.md
Pscheidl: Done. The request is now 4 GB, not the limit (limits, if not strictly enforced, are managed by K8S implementation). We'd only like to request 4 GB per memory. The rests is for the K8S admin to figure out,.

h2o-k8s/README.md
Pscheidl: The defined service is not a load balancer. It's a headless service. There is no load balancing done. It only returns internal IPS (cluster-internal) of all underlying pods (H2O pods in this case) and creates related DNS records. This is only used for H2O clustering.

This is desired behavior, because:

1. IPs of all pods in K8S are internal and should not be visible from the outside. This is not even desired. It's only used for clustering.
2. Making H2O visible from the outside is yet another layer in K8S world. It's a separation of concerns and proper layering principles that makes sure exposure to the outer world can be modified/deleted/created independently of internal cluster state. It's a pattern followed in general.

The DNS service address is there only for H2O to make clustering possible. Which should always be "internal process". Ingresses serve for exposure to the outer worlds, outside of K8S cluster.

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfig.java
Pscheidl: No actions required on exit. H2O process will be terminated, as well as the docker image. No other resources should be deleted. The services/ingresses/routes still live, ready to serve when the desired state of the underlying cluster of H2O pods goes live again.

If those are temporary and should be deleted with H2O, then this is the responsibility of the orechestration software, not H2O's.

E.g. when healthchecks are implemented (we should maybe do it in this PR), K8S `StatefulSet` will restart the whole cluster of H2O pods once any (even one) node fails - that's the point of having a StatefulSet, stateless applications can be restarted independently of each other and also load balanced. Once the cluster goes up, all the services/ingresses/routes/volumes etc. attached to it will continue to serve their purpose.

From H2O's perspective, do we miss something ? Maybe we should at least log some kind of message about termination ?

h2o-k8s/src/main/java/water/k8s/lookup/KubernetesDnsLookup.java
Pscheidl: Yes we can ! Done.

Beware: This might happen if the headless service is started at the same time as the H2O pods - During the very first iteration or a couple more iterations, this might actually happen, as the records are not there yet.

settings.gradle
Pscheidl: Yes it is !

Pscheidl: Done !

Pscheidl:
Next step is to do health checks in order to allow automatic cluster restarts when required (those are turned on and off by K8S cluster administrator or the orchestration software.).

Created a separate JIRA, as this functionality is definitely not required, it's a convenience feature. https://0xdata.atlassian.net/browse/PUBDEV-7359

Pscheidl:
Also, the REST API clustering could also be a convenient option for some. It has no huge advantages and it requires role bindings set up. I've created a JIRA for that as well, to keep track of the task. We can always postpone/decline the JIRA.

https://0xdata.atlassian.net/browse/PUBDEV-7360

Pscheidl:
Note to self: Detect K8S presence in any other way - sometimes we do NOT want the lookup to be triggered, even if running on K8S.

Pscheidl:
The commit https://github.com/h2oai/h2o-3/pull/4370/commits/95173548eb8382d30b78890884bee5caf3611744 makes it easy to disable this behavior. As the clustering will only work with a headless service set-up and the environment variable `H2O_KUBERNETES_SERVICE_DNS` present, the KubernetesEmbeddedConfigProvider simply detects presence of this environment variable. If it's there, it's a clear signal to do the clustering. If not, it ignores even the very fact that it's running on K8S.

This means freedom for the users. For example, it'd not serve us well in our test pipeline once it starts running inside K8S.

Pscheidl:
@michalkurka @mn-mikke This is it. Here, on my side, this is the final version, as I can't think of any improvements related to DNS-based clustering. 

Please review.

h2o-k8s/README.md
pkozelka: @Pscheidl `Routess` looks like a typo

Pscheidl: Fixed, thanks.

Pscheidl:
Thank you @pkozelka .

PR #4375, opened 2020-03-02
PUBDEV-6293: in AutoML, tune learning rate for best model found during exploration (XGB+GBM)

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-6293

Current approach is to split the AutoML training budget  into 3 stages:
- an exploration stage, which basically consists in the current state of AutoML: builds default models and performs HPO using random grid searches.
- an exploitation stage, trying to fine-tune/improve performance of the best models found during the first section.
- the final section just trains the stacked ensemble as before.

The main difficulty is to keep respecting the budget constraints (`max_runtime_secs`, `max_models`) with the introduction of this second section.

Adds new parameter to AutoML, `exploitation_ratio`, which defaults to 0 (for now).


h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -581,50 +609,40 @@ private void learn() {
...
-    return Key.make(algoName + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
-  }
-
-  Key<Grid> gridKey(String algoName, boolean with_counter) {
-    String counterStr = with_counter ? "_" + nextInstanceCounter(algoName, gridInstanceCounters) : "";
-    return Key.make(algoName + "_grid_" + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
+  Key makeKey(String algoName, String type, boolean with_counter) {
+    String counterStr = with_counter ? "_" + nextInstanceCounter(algoName, type) : "";
+    String prefix = StringUtils.isNullOrEmpty(type) ? algoName : algoName+"_"+type+"_";
+    return Key.make(prefix + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
```
sebhrusen: standardized all this `makeKey` logic so that we can always use the same methods for all steps.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -60,9 +61,24 @@ public AutoMLBuildControl() {
...
+      // reasonable defaults:
+      set_max_models(0); // no limit
+      set_max_runtime_secs(0); // no limit
+      set_max_runtime_secs_per_model(0); // no limit
+
+      set_stopping_rounds(3);
+      set_stopping_tolerance(AUTO_STOPPING_TOLERANCE);
+      set_stopping_metric(StoppingMetric.AUTO);
+    }
+
```
sebhrusen: just moved to respect usual Java convention:
- static fields
- fields
- constructor
- methods

h2o-automl/src/main/java/ai/h2o/automl/IAlgo.java
```diff
@@ -0,0 +1,9 @@
+package ai.h2o.automl;
+
+import java.io.Serializable;
+
+public interface IAlgo extends Serializable {
+  String name();
+  default String urlName() { return name().toLowerCase(); }
+  default boolean enabled() { return true; }
+}
```
sebhrusen: introduced this interface for 2 purposes:
1. being able to test `ModelingStep` internals, using a very fast Dummy algo.
2. offer the possibility to plug any algo to AutoML, including those not defined in the Algo enum.

h2o-automl/src/main/java/ai/h2o/automl/events/EventLog.java
```diff
@@ -19,16 +18,16 @@
  */
 public class EventLog extends Keyed<EventLog> {
 
-  public final Key<AutoML> _automl_id;
+  public final Key _runner_id;
```
sebhrusen: renamed this as it is not not used only for automl, for also for selection, the same applies with Leaderboard.
This is mainly because EventLog is a property of leaderboard, and I need a temporary leaderboard to apply the selection logic.

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/Leaderboard.java
```diff
@@ -436,11 +519,30 @@ public void addModels(final Key<Model>[] models) {
...
     boolean decreasing = !isLossFunction(metric);
-    List<Key<Model>> newModelsSorted =
-            ModelMetrics.sortModelsByMetric(metric, decreasing, Arrays.asList(models));
+    List<Key<Model>> newModelsSorted = ModelMetrics.sortModelsByMetric(metric, decreasing, Arrays.asList(models));
     return newModelsSorted.toArray(new Key[0]);
   }
 
+  /**
+   * Sort by metric on the leaderboard/test set or default model metrics.
+   */
```
sebhrusen: extracted the sorting logic that was in the big `addModels` method

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/Leaderboard.java
```diff
@@ -278,113 +289,161 @@ private void setDefaultMetrics(Model m) {
...
     _metrics = metrics;
+    update();
+    unlock();
+  }
+
+  private ModelMetrics getOrCreateModelMetrics(Key<Model> modelKey) {
+    return getOrCreateModelMetrics(modelKey, getExtensionsAsMap());
+  }
+
+  private ModelMetrics getOrCreateModelMetrics(Key<Model> modelKey, Map<Key<Model>, LeaderboardCell[]> extensions) {
```
sebhrusen: logic extracted from the big `addModels` method

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/Leaderboard.java
```diff
@@ -575,43 +677,55 @@ public TwoDimTable toTwoDimTable(String... extensions) {
...
+          );
+        }
+        return table;
+      } finally {
+        readLock.unlock();
+      }
+    } else {
+      return makeTwoDimTable(tableHeader, 0, getDefaultTableColumns().toArray(new LeaderboardColumn[0]));
+    }
+  }
```
sebhrusen: mainly reindentation due to the introduction of the readLock as the client polling can occur when leaderboard is updated.

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
```diff
@@ -172,6 +201,108 @@ XGBoostParameters prepareModelParameters() {
...
+                            new Object[] {       0.2   ,                   sti },
+                            new Object[] {       0.1   ,                 2*sti },
+                            new Object[] {       0.05  ,                 2*sti },
+                            new Object[] {       0.02  ,                 3*sti },
+                            new Object[] {       0.01  ,                 3*sti },
+                            new Object[] {       0.005 ,                 4*sti },
+                            new Object[] {       0.002 ,                 4*sti },
+                            new Object[] {       0.001 ,                 5*sti },
+                            new Object[] {       0.0005,                 5*sti },
+                    };
```
sebhrusen: I tried 2 strategies to keep the model training speed reasonnable when the learn_rate is diminishing:
1. increasing `stopping_rounds` when `learn_rate` is lowered, as first suggested in the Jira ticket: we keep scoring regulary (every 5 trees: this is hardcoded in XGBoostSteps as this can not be specified by the user), this way the training doesn't stop "too early" with low learn_rate, but it probably scores too often. Also, we modify a parameter specified by the user, which is not polite :)
2. keep the `stopping_rounds` constant and as specified by the user, but instead increase the `score_tree_interval` when `learn_rate`is lowered: this allows to score less often, but then we have more trees being built between each score, and for each `stopping_round`, so this also avoids stopping "too early" with low `learn_rate`.

h2o-core/src/main/java/hex/grid/SequentialWalker.java
```diff
@@ -0,0 +1,109 @@
...
+import hex.ScoreKeeper;
+import hex.ScoringInfo;
+import hex.grid.HyperSpaceSearchCriteria.SequentialSearchCriteria;
+import hex.grid.HyperSpaceSearchCriteria.StoppingCriteria;
+
+import java.util.Arrays;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+public class SequentialWalker<MP extends Model.Parameters> implements HyperSpaceWalker<MP, SequentialSearchCriteria> {
```
sebhrusen: This one is introduced for the exploitation phase, please review.

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
```diff
@@ -316,40 +377,188 @@ protected Work getAllocatedWork() {
...
                     baseParms,
                     searchParms,
-                    new GridSearch.SimpleParametersBuilderFactory<>(),
-                    searchCriteria,
-                    GridSearch.SEQUENTIAL_MODEL_BUILDING
+                    searchCriteria
             );
         }
     }
+
```
sebhrusen: Created the relatively complicated abstraction below for 2 reasons:
- it allows to have the exploitation steps for each algo focusing only on the exploitation logic, ie. define the model/search parameters.
- it is generic enough to be used for iterative exploration+selection (e.g. Bayesian Opt, although there are probably small changes needed).

h2o-automl/src/main/java/ai/h2o/automl/IAlgo.java
```diff
@@ -0,0 +1,9 @@
+package ai.h2o.automl;
+
+import java.io.Serializable;
+
+public interface IAlgo extends Serializable {
```
michalkurka: is there a way to avoid using Serializable? it will work but serializable is meant more of a fallback

h2o-automl/src/main/java/ai/h2o/automl/Models.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public Key<M>[] getModelKeys() {
+        return _modelKeys.clone();
+    }
+
+    @Override
+    public M[] getModels() {
+        return Arrays.stream(_modelKeys)
```
michalkurka: if you have many objects, it is better to call first `DKV.prefetch(key)` in a first loop - this will start transfer of all the objects in parallel

the second loop (DKV.get) then just waits for the transfer to finish

h2o-automl/src/main/java/ai/h2o/automl/Models.java
```diff
@@ -0,0 +1,64 @@
...
+import hex.ModelContainer;
+import water.*;
+import water.util.ArrayUtils;
+
+import java.lang.reflect.Array;
+import java.util.Arrays;
+
+public class Models<M extends Model> extends Lockable<Models<M>> implements ModelContainer<M> {
+
+    private final Class<M> _clz;
```
michalkurka: we usually avoid serializing the Class objects, instead we just pass around typeid - see eg. the constructor of the Job class for an example

h2o-automl/src/main/java/ai/h2o/automl/Models.java
sebhrusen: nice tip, will change my implementation, thx

h2o-automl/src/main/java/ai/h2o/automl/IAlgo.java
sebhrusen: for interfaces, this is the only possibility together with `Freezable`.
The problem is that implementing all Freezable methods each time we basically want to define a new string is a bit overkill...
Also, I'm not familiar enough with the Weaver internals, but the main implementation of this interface is the `Algo` enum, so I expect the Weaver to use the `enum` machinism to serialize the value and not the `Serializable` one: it all depends if given a class property, the Weaver uses the declared property type or the actual property type, this I don't know.

h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
ledell: @sebhrusen Second approach sounds more reasonable, computationally.

h2o-py/h2o/automl/autoh2o.py
```diff
@@ -65,6 +65,7 @@ def __init__(self,
                  project_name=None,
                  exclude_algos=None,
                  include_algos=None,
+                 exploitation_ratio=None,
```
```suggestion
                 exploitation_ratio=0,
```

h2o-r/h2o-package/R/automl.R
```diff
@@ -43,12 +43,13 @@
...
 #'        and the full list of options is: "DRF" (Random Forest and Extremely-Randomized Trees), "GLM", "XGBoost", "GBM", "DeepLearning" and "StackedEnsemble".
-#         Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
+#'        Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
 #' @param include_algos Vector of character strings naming the algorithms to restrict to during the model-building phase. This can't be used in combination with exclude_algos param.
-#         Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
+#'        Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
+#' @param exploitation_ratio The budget ratio (between 0 and 1) dedicated to the exploitation phase. Defaults to 0.
```
```suggestion
#' @param exploitation_ratio The budget ratio (between 0 and 1) dedicated to the exploitation (vs exploration) phase. Defaults to 0.
```

h2o-r/h2o-package/R/automl.R
```diff
@@ -95,6 +96,7 @@ h2o.automl <- function(x, y, training_frame,
                        exclude_algos = NULL,
                        include_algos = NULL,
                        modeling_plan = NULL,
+                       exploitation_ratio = NULL,
```
```suggestion
                       exploitation_ratio = 0.0,
```

h2o-automl/src/main/java/ai/h2o/automl/Algo.java
```diff
@@ -9,7 +9,7 @@
 // if we need to make the Algo list dynamic, we should just turn this enum into a class...
 // implementation of AutoML.algo can be safely removed once we get rid of this interface: current purpose
 // is to keep backward compatibility with {@link AutoML.algo}
-public enum Algo {
+public enum Algo implements IAlgo {
```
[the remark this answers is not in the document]
sebhrusen: it's old java naming convention for Interface

h2o-py/h2o/automl/autoh2o.py
```diff
@@ -113,6 +114,7 @@ def __init__(self,
         :param include_algos: List of character strings naming the algorithms to restrict to during the model-building phase.
           This can't be used in combination with `exclude_algos` param.
           Defaults to ``None``, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
+        :param exploitation_ratio: The budget ratio (between 0 and 1) dedicated to the exploitation phase. Defaults to 0.
```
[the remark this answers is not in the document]
sebhrusen: Adding a short section to our USer Guide.

h2o-automl/src/main/java/ai/h2o/automl/IAlgo.java
honzasterba: the only reason this integface seems to exist is testing, that can be acomplished with a mock library or simply not using enum for Algo, which I think is worth considering since this IInterface naming scheme is very ugly and quite confusing
also having an interfac where we never expect another implementation in prod is not a great idea

sebhrusen: @honzasterba https://github.com/h2oai/h2o-3/pull/4375#discussion_r398239638
Note that `ModelingStep` supports external definitions, and in that regard, any algo name, even those not listed in the enum should be supported...
I'm considering getting rid of the enum though, it's even documented in the Algo definition.
For info, agree that the interface naming is not pretty, but at least it's consistent (cf. `IStoppingMetric` for another enum...).

h2o-py/h2o/automl/autoh2o.py
```diff
@@ -113,6 +114,7 @@ def __init__(self,
         :param include_algos: List of character strings naming the algorithms to restrict to during the model-building phase.
           This can't be used in combination with `exclude_algos` param.
           Defaults to ``None``, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
+        :param exploitation_ratio: The budget ratio (between 0 and 1) dedicated to the exploitation (vs exploration) phase. Defaults to 0.
```
```suggestion
        :param exploitation_ratio: Specify the budget ratio (between 0 and 1) dedicated to the exploitation (vs exploration) phase. By default, the exploitation phase is disabled (exploitation_ratio=0) as this is still experimental; to activate it, it is recommended to try a ratio around 0.1. Note that the current exploitation phase only tries to fine-tune the best XGBoost and the best GBM found during exploration.
```

h2o-r/h2o-package/R/automl.R
```diff
@@ -43,12 +43,13 @@
...
 #'        and the full list of options is: "DRF" (Random Forest and Extremely-Randomized Trees), "GLM", "XGBoost", "GBM", "DeepLearning" and "StackedEnsemble".
-#         Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
+#'        Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
 #' @param include_algos Vector of character strings naming the algorithms to restrict to during the model-building phase. This can't be used in combination with exclude_algos param.
-#         Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
+#'        Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
+#' @param exploitation_ratio The budget ratio (between 0 and 1) dedicated to the exploitation (vs exploration) phase. Defaults to 0.
```
```suggestion
#' @param exploitation_ratio The budget ratio (between 0 and 1) dedicated to the exploitation (vs exploration) phase. By default, the exploitation phase is disabled (exploitation_ratio=0) as this is still experimental; to activate it, it is recommended to try a ratio around 0.1. Note that the current exploitation phase only tries to fine-tune the best XGBoost and the best GBM found during exploration.
```

PR #4378, opened 2020-03-03
PUBDEV-7328: Add to docs that can now control number of columns to display in dataframe in R client in Jupyter notebook

hannah-tillman:


h2o-r/h2o-package/R/frame.R
```diff
@@ -1926,10 +1926,15 @@ h2o.setLevels <- function(x, levels, in.place = TRUE) .newExpr("setDomain", chk.
 #' \dontrun{
 #' library(h2o)
 #' h2o.init(ip <- "localhost", port = 54321, startH2O = TRUE)
-#' australia_path <- system.file("extdata", "australia.csv", package = "h2o")
```
angela0xdata: No need to change the import commands. This method works in R and points to a csv that is packaged with the R client.

h2o-r/h2o-package/R/frame.R
```diff
@@ -1926,10 +1926,15 @@ h2o.setLevels <- function(x, levels, in.place = TRUE) .newExpr("setDomain", chk.
...
-#' australia <- h2o.uploadFile(path = australia_path)
-#' head(australia, 10)
-#' tail(australia, 10)
+#' australia <- h2o.importFile("https://s3.amazonaws.com/h2o-public-test-data/smalldata/extdata/australia.csv")
+#' # Return the first 10 rows and 6 columns
+#' h2o.head(australia, n = 10L, m = 6L)
+#' # Return the last 10 rows and 6 columns
+#' h2o.tail(australia, n = 10L, m = 6L)
+#' 
+#' # For Jupyter notebook with an R kernal,
```
angela0xdata: typo: kernel

PR #4380, opened 2020-03-04
PUBDEV-6938 TF-IDF

jancijen:
Implementation of measure TF-IDF used in information retrieval/text mining.

JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6938


h2o-algos/src/test/java/hex/tfidf/TfIdfPreprocessorTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class TfIdfPreprocessorTest extends TestUtil {
+
+    @BeforeClass()
+    public static void setup() { stall_till_cloudsize(1); }
+
+    @Test
+    public void testTfIdfPreprocessing() {
+        Frame fr = getSimpleTestFrame();
+        byte[] outputTypes = new byte[]{Vec.T_STR, Vec.T_NUM, Vec.T_NUM};
```
maurever: This line causes the error (`DistributedException from /192.168.0.164:54321: 'NewChunk has type String, but the Vec is of type Numeric'`). The third type (word tokens) should be `Vec.T_STR`.

jancijen: Thank you. Careless mistake.

h2o-r/h2o-package/R/tf-idf.R
```diff
@@ -0,0 +1,15 @@
...
+#' @param frame documents frame for which TF-IDF values should be computed.
+#'              Row format: documentID, documentString
+#'
+#' @return  resulting frame with TF-IDF values.
+#'          Row format: documentID, word, TF, IDF, TF-IDF
+h2o.tf_idf <- function(frame) {
+    if( is(frame, 'H2OFrame') ) {
+        .newExpr('tf-idf', frame)
+    } else {
+        warning(paste0("TF-IDF cannot be computed for class ", class(frame), ". Frame input is required."))
```
maurever, line 12:
```suggestion
        warning(paste0("TF-IDF cannot be computed for class ", class(frame), ". H2OFrame input is required."))
```

h2o-core/src/test/java/hex/tfidf/DocumentFrequencyTest.java
```diff
@@ -0,0 +1,153 @@
...
+            put("Z", 1L);
+        }};
+        int expectedDfValuesCnt = expectedDocumentFrequencies.size();
+
+        try {
+            DocumentFrequency df = new DocumentFrequency();
+            Frame outputFrame = df.compute(fr);
+            long outputRowsCnt = outputFrame.numRows();
+
+            Assert.assertEquals(expectedDfValuesCnt, outputRowsCnt);
```
maurever, line 45: Try to add messages to all asserts, please. :)

h2o-r/h2o-package/R/tf-idf.R
jancijen: Good point. I changed it.

h2o-core/src/test/java/hex/tfidf/DocumentFrequencyTest.java
jancijen: 👍 I added messages to all asserts.

h2o-algos/src/test/java/hex/tfidf/TfIdfTrainerTest.java
```diff
@@ -0,0 +1,71 @@
...
+                put("A", new int[]{1,3,0});
+                put("B", new int[]{1,0,1});
+                put("C", new int[]{1,0,3});
+                put("Z", new int[]{0,1,0});
+            }};
+
+            expWordsCounts.forEach((word, expCounts) -> {
+                Vec actualCounts = outputFrame.vec(wordsIndices.get(new BufferedString(word))._val);
+                for (int i = 0; i < expCounts.length; i++)
+                    assertEquals(expCounts[i], actualCounts.at8(i));
```
maurever, line 47: Please, add a message to all asserts. :)

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
```diff
@@ -0,0 +1,153 @@
...
+
+    @Override
+    public String str() {
+        return "tf-idf";
+    }
+
+    /**
+     * Final TF-IDF Map-Reduce task used to combine TF and IDF values together.
+     */
+    private static class TfIdf extends MRTask<TfIdf> {
```
maurever, line 114: I would recommend to name it as:
```suggestion
    private static class TfIdfTask extends MRTask<TfIdf> {
```

The same for your other MRTasks.

h2o-py/h2o/information_retrieval/tf_idf.py
```diff
@@ -0,0 +1,31 @@
...
+                        </p>
+                        <p>
+                        Row format when <code>preprocess = False</code>: documentID, word
+                        </p>
+    :param preprocess:  whether input frame should be pre-processed. Defaults to <code>True</code>.
+
+    :return:    resulting frame with TF-IDF values. 
+                Row format: documentID, word, TF, IDF, TF-IDF
+    """
+
```
maurever, line 28: Add check it is H2OFrame, please.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Override
+    public String[] args() {
+        return new String[]{ "frame", "preprocess" };
+    }
+
+    @Override
+    public Val apply(Env env, Env.StackHelp stk, AstRoot[] asts) {
+        Frame inputFrame = stk.track(asts[1].exec(env).getFrame());
```
maurever, line 47: What if a user will put a frame with more than two columns?

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
```diff
@@ -0,0 +1,153 @@
...
+            byte[] outputTypes = new byte[]{ Vec.T_NUM, Vec.T_STR };
+
+            TfIdfPreprocessor preprocessor = new TfIdfPreprocessor();
+            wordFrame = preprocessor.doAll(outputTypes, inputFrame).outputFrame(PREPROCESSED_FRAME_COL_NAMES, null);
+
+            documentsCnt = inputFrame.numRows();
+        } else {
+            wordFrame = inputFrame;
+            // TODO: Avoid converting to categorical Vec
+            documentsCnt = inputFrame.vec(0).toCategoricalVec().cardinality();
```
honzasterba, line 63: 1) AstUnique might by an alternative to converting to categorical
2) either or, you might want to track the newly created categorical vec so that it does not leak (I am not sure about this)

h2o-core/src/main/java/hex/tfidf/DocumentFrequency.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Computes document frequency values for given words in documents.
+     *
+     * @param wordFrame input frame of words for which document frequency
+     *                  values should be computed. For exact format
+     *                  see {@link TermFrequency}.
+     *
+     * @return frame containing document frequency values for given words.
+     */
+    public Frame compute(Frame wordFrame) {
```
honzasterba, line 25: can be static

h2o-core/src/test/java/hex/tfidf/DocumentFrequencyTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+        Set<String> doc1Tokens = new HashSet<>(tokens.subList(0, docSplitIndex));
+        Set<String> doc2Tokens = new HashSet<>(tokens.subList(docSplitIndex, tokensCnt));
+        
+        Map<String, Long> expectedDFs = doc1Tokens.stream().collect(Collectors.toMap(token -> token, token -> 1L));
+        doc2Tokens.forEach(token -> expectedDFs.merge(token, 1L, Long::sum));
+
+        System.out.println("Computing DF...");
+        try {
+            DocumentFrequency df = new DocumentFrequency();
```
honzasterba, line 91: unnecessary variable

h2o-core/src/test/java/hex/tfidf/DocumentFrequencyTest.java
```diff
@@ -0,0 +1,146 @@
...
+                .withChunkLayout(chunkLayout.stream().mapToLong(Integer::intValue).toArray())
+                .build();
+
+        Set<String> doc1Tokens = new HashSet<>(tokens.subList(0, docSplitIndex));
+        Set<String> doc2Tokens = new HashSet<>(tokens.subList(docSplitIndex, tokensCnt));
+        
+        Map<String, Long> expectedDFs = doc1Tokens.stream().collect(Collectors.toMap(token -> token, token -> 1L));
+        doc2Tokens.forEach(token -> expectedDFs.merge(token, 1L, Long::sum));
+
+        System.out.println("Computing DF...");
```
honzasterba, line 89: use Logging and debug (Log4j)

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        // DF
+        DocumentFrequency dfTaks = new DocumentFrequency();
```
honzasterba, line 67: unnecessary variable

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
```diff
@@ -0,0 +1,153 @@
...
+        dfOutFrame.remove(dfOutFrame.numCols() - 1);
+        dfOutFrame.add(IDF_COL_NAME, idfValues);
+
+        // TF
+        TermFrequency tfTask = new TermFrequency();
+        Frame tfOutFrame = tfTask.compute(wordFrame);
+        
+        // Intermediate frame containing both TF and IDF values
+        // TODO: Avoid converting a string column to categorical
+        tfOutFrame.replace(1, tfOutFrame.vecs()[1].toCategoricalVec());
```
honzasterba, line 83: could use AstUnique here as well, also I see potential leaks all over this method

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
```diff
@@ -0,0 +1,153 @@
...
+        // TF-IDF
+        int tfOutFrameColCnt = tfIdfIntermediate.numCols();
+        TfIdf tfIdfTask = new TfIdf(tfOutFrameColCnt - 2, tfOutFrameColCnt - 1);
+        Vec tfIdfValues = tfIdfTask.doAll(new byte[]{ Vec.T_NUM }, tfIdfIntermediate).outputFrame().anyVec();
+
+        // Construct final frame containing TF, IDF and TF-IDF values
+        tfIdfIntermediate.add(TF_IDF_COL_NAME, tfIdfValues);
+        tfIdfIntermediate._key = Key.make();
+
+        System.out.println(tfIdfIntermediate.toTwoDimTable().toString());
```
honzasterba, line 101: use logging ideally wrapped in isDebugEnabled

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
```diff
@@ -0,0 +1,153 @@
...
+
+        /**
+         * Computes TF-IDF value from given TF and IDF values.
+         * 
+         * @param tf    TF value to be used to compute TF-IDF.
+         * @param idf   IDF value to be used to compute TF-IDF.
+         *
+         * @return TF-IDF value from given TF and IDF values.
+         */
+        private static double tfIdf(final long tf, final double idf) {
```
honzasterba, line 149: unnecessary function

h2o-algos/src/test/java/hex/tfidf/TfIdfTrainerTest.java
jancijen: Sorry, this test is not used anymore. I removed it.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
jancijen: Sounds good. I renamed all of the `MRTask` in such way. Also, I renamed "tasks" which are not actual `MRTaks` - `DocumentFrequency`, `TermFrequency`... to be consistent.

h2o-py/h2o/information_retrieval/tf_idf.py
jancijen: Good point. Added.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
jancijen: I added check for the input types - `Num`, `Str` and number of columns (`>= 2`). Or would you prefer to have parameters to indicate which columns should be used?

I also added corresponding unit tests for this, and added check for empty input.

h2o-algos/src/test/java/hex/tfidf/Word2IndexTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Test
+    public void testSmallDataWordsIndexing() {
+        Frame fr = getSimpleTestFrame();
+
+        try {
+            Word2Index word2index = new Word2Index(" ").doAll(fr.vec(0));
+
+            String[] expUniqueWords = new String[] { "A", "B", "C", "Z" };
+            for (String expWord : expUniqueWords)
+                Assert.assertTrue(word2index._wordsIndices.containsKey(new BufferedString(expWord)));
```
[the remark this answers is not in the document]
jancijen: +1: I changed it in all test classes.

h2o-core/src/main/java/hex/tfidf/DocumentFrequency.java
```diff
@@ -0,0 +1,50 @@
...
+        
+        int[] groupByColumns = new int[]{ 1 };
+        
+        // Construct final frame with DF values
+        Frame dfFrame = new AstGroup().performGroupingWithAggregations(uniqueWordsPerDocFrame,
+                                                                       groupByColumns,
+                                                                       aggs).getFrame();
+        dfFrame.setNames(OUTPUT_FRAME_COL_NAMES);
+        
+        uniqueWordsPerDocFrame.remove();
```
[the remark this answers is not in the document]
jancijen: Good point. I added `Scope` to handle it.

h2o-core/src/main/java/hex/tfidf/DocumentFrequency.java
jancijen: Good point. I changed it for both `TermFrequency` and `DocumentFrequency` and in places where `compute` method is being used.

h2o-core/src/test/java/hex/tfidf/DocumentFrequencyTest.java
jancijen: Now solved by your previous suggestion (to make it static).

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
jancijen: Now solved by your previous suggestion (to make it static).

h2o-core/src/test/java/hex/tfidf/DocumentFrequencyTest.java
jancijen: +1: Good point. I replaced standard output prints with debug logs throughout the code.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
jancijen: 👍 Changed.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
jancijen: This was there just for the code readability, but I guess in this case it is quite obvious and strictly defined to be a product. I removed it.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
jancijen: Good idea. I replaced it with `AstUnique` and `Vec#length()`.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
jancijen: In this case I need to merge `tfOutFrame` and `dfOutFrame`, where term frequency values are "per word and document" and document frequency values which are "per word". Therefore, I need to add one value from `dfOutFrame` to multiple rows in `tfOutFrame`. How would you apply `AstUnique` to solve this?

I checked and it seems that long time ago there has been an implementation of `AstMerge` that worked with string values - see `mergetest.py` in https://github.com/h2oai/h2o-3/commit/f645ee8c6d86749845d033c4d1e4851bb9c96d52, but this no longer works.

But for now, it seems that the main reason why I was worried about conversion to `Categorical` vector - the `Categorical.MAX_CATEGORICAL_COUNT` does not apply for `String` vectors that are being transformed to `Categorical` vector. I tried to convert `String` vector containing `Categorical.MAX_CATEGORICAL_COUNT + 10` unique values to `Categorical` vector and it worked just fine.

Is there a better solution for this?

Also, I fixed the leaks by using `Scope`.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
maurever: I think, the parameter for indicating the columns should be the best way. WDYT @honzasterba  and @michalkurka ?

honzasterba: sounds good to me, something like this:

`tf_idf(frame, document_id=(str, int), document_word=(str,int), preprocess, case_sens)`

I think expecting that the frame has some pre-defined structure or column names is not something we do very often in H2O

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
honzasterba: Converting to categorical and doing merge is just hiding the problem. The problem is that you need to do a merge of two frames over a very large string-based key.
From my quick read of the code both tfOutFrame and dfOutFrame are computed from the same source wordsFrame, it might be possible to make this more efficient by combining both of the computations of tfOut and dfOut into a single series of MRTasks.
On the other hand how big can the domain really be? We are taking about text here, so the domains should not be bigger than couple 100k. So maybe the categorical+merge is not such a big deal.
@michalkurka any thoughts?

maurever:
Mr. Jenkins, please test!

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
jancijen: I know. Well, yes that is true, both `tfOutFrame` and `dfOutFrame` are built by using same input, but it is done via `GroupBy`, where `TermFrequency` is done on grouping by `docID` and `word`, and `DocumentFrequency` first groups by the same columns (I changed it in the last commit to expecte unique words per document, therefore reuse the output of `TermFrequencyTask` since it is not a publicly available API and it saves some computation.) and then do another groupby operation in order to count number of documents containing given word.

Well, to be honest it looks like that even `Categorical.MAX_CATEGORICAL_COUNT` shouldn't be a problem even for larger document corpora. But if there is a better way to solve this I would prefer that over converting it to a categorical vector.

h2o-core/src/main/java/water/rapids/ast/prims/advmath/AstTfIdf.java
jancijen: I changed it to expect column indices for `AstTfIdf` and for `R` and `Python` methods both column names and indices are allowed.

maurever:
So, we finished with @jancijen last updates of code to pass all tests. The last thing is failing and also fixed - tests are running right now (#4668 ). It is just a small issue in the Python 2.7 demo. So please, @honzasterba and @michalkurka, could you do a quick review again, that everything was implemented correctly, please? Thank you!

maurever:
The tests in the #4668 PR are green. Thank you, @jancijen, for your contribution to the H2O-3 platform! Thanks, @honzasterba, for your review!

PR #4381, opened 2020-03-04
PUBDEV-6599: Change the output of get_automl in Python from a dictionary to an object

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-6599


h2o-py/tests/testdir_jira/pyunit_pubdev_6599_h2o_automl_output.py
```diff
@@ -0,0 +1,41 @@
...
+        self.assertIsNotNone(self._aml_output['training_info'])
+        
+    def test_properties(self):
+        self.assertIsNotNone(self._aml_output.project_name)
+        self.assertIsNotNone(self._aml_output.leader)
+        self.assertIsNotNone(self._aml_output.leaderboard)
+        self.assertIsNotNone(self._aml_output.event_log)
+        self.assertIsNotNone(self._aml_output.training_info)
+        with self.assertRaises(KeyError):
+            self._aml_output['predict']
```
sebhrusen: in those 2 tests, could we compare `_aml_output` properties with `aml` properties instead of just checking that they're not `None`?

h2o-py/tests/testdir_jira/pyunit_pubdev_6599_h2o_automl_output.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_predict(self):
+        predictions = self._aml_output.predict(self._test_data)
+        self.assertTrue(float((self._test_data['species'] == predictions['predict']).mean()) > 0.95)
```
sebhrusen: same here, would rather compare that output predictions matches the `aml` predictions.

h2o-py/tests/testdir_jira/pyunit_pubdev_6599_h2o_automl_output.py
```diff
@@ -0,0 +1,41 @@
+import unittest
+import h2o
+from h2o.automl import H2OAutoML, get_automl
+from tests import pyunit_utils
+
+
+class H2OAutoMLOutputTest(unittest.TestCase):
```
sebhrusen: probably the first explicit usage of Py `unittest` in the entire project.
We should have started that long ago, thx!

h2o-py/h2o/automl/_h2o_automl_output.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        Download the leader model in AutoML in MOJO format.
+
+        :param path: the path where MOJO file should be saved.
+        :param get_genmodel_jar: if True, then also download h2o-genmodel.jar and store it in folder ``path``.
+        :param genmodel_name: Custom name of genmodel jar
+        :returns: name of the MOJO file written.
+        """
+
+        return ModelBase.download_mojo(self.leader, path, get_genmodel_jar, genmodel_name)
```
sebhrusen: looks like most of those methods with their documentation are  duplicates from the `H2OAutoML` class, any idea to share those? like a shared base class?

h2o-py/h2o/automl/_base.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        Download the leader model in AutoML in MOJO format.
+    
+        :param path: the path where MOJO file should be saved.
+        :param get_genmodel_jar: if True, then also download h2o-genmodel.jar and store it in folder ``path``.
+        :param genmodel_name: Custom name of genmodel jar
+        :returns: name of the MOJO file written.
+        """
+    
+        return ModelBase.download_mojo(self.leader, path, get_genmodel_jar, genmodel_name)
```
sebhrusen: I think this could include doc/contract for the shared properties: `leader`, `leaderboard` and so on, as soon as we delete the docstrings from the subclasses:
in mixin:
```
    @property
    def project_name(self):
        """
        Retrieve a string indicating the project_name of the automl instance to retrieve.
        :return: a string containing the project_name
        """
       pass
```

in impl:
```
    @property
    def project_name(self):
           return self._project_name
```
and 
```
    @property
    def project_name(self):
           return None if self._leader_id is None else h2o.get_model(self._leader_id)
```
this should work

sebhrusen:
Thank you @tomasfryda .
Feel free to merge after the PR build if the tests are passing.

tomasfryda:
Not all tests are passing - `Py3.6 Medium-large / pyunit_anomaly_deeplearning_large.py – r_suite` failed (seems to be a timeout issue - "Cancelling nested steps due to timeout"). And for some reason `continuous-integration/jenkins/pr-head`. I don't think that the failure is related to this PR (but I might be mistaken), should I try to rerun the tests (and possibly waste resources) or is it ok to merge?

sebhrusen:
@tomasfryda failures are unrelated, please merge.
For such a change, you can safely expect that only `Py3.5 Small AutoML` is relevant.

PR #4382, opened 2020-03-04
PUBDEV-7304: Add GBM POJO support for SortByResponse and EnumLimited encodings

koniecsveta:
https://0xdata.atlassian.net/browse/PUBDEV-7304

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/easy/EnumEncoderColumnMapper.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/easy/EnumEncoderColumnMapper.java
@@ -7,14 +7,18 @@ import java.util.Map;
 
 public class EnumEncoderColumnMapper {
 
-  private final GenModel _m;
+  protected final GenModel _m;
 
   public EnumEncoderColumnMapper(GenModel m) {
     _m = m;
   }
 
+  public String[] getModelColumnNames() {
+    return _m.getNames();
+  }
+  
   public Map<String, Integer> create() {
-    String[] modelColumnNames = _m.getNames();
+    String[] modelColumnNames = getModelColumnNames();
     Map<String, Integer> modelColumnNameToIndexMap = new HashMap<>(modelColumnNames.length);
     for (int i = 0; i < modelColumnNames.length; i++) {
       modelColumnNameToIndexMap.put(modelColumnNames[i], i);
--- a/h2o-genmodel/src/main/java/hex/genmodel/easy/EnumLimitedEncoder.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/easy/EnumLimitedEncoder.java
@@ -1,0 +1,56 @@
+package hex.genmodel.easy;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class EnumLimitedEncoder implements CategoricalEncoder {
+
+  private final String columnName;
+  private final int targetIndex;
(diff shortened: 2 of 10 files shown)


h2o-genmodel/src/main/java/hex/genmodel/easy/EnumLimitedEncoder.java
```diff
@@ -0,0 +1,57 @@
...
+public class EnumLimitedEncoder implements CategoricalEncoder {
+
+  private final String columnName;
+  private final int targetIndex;
+  private final Map<String, Integer> domainMap;
+  
+  EnumLimitedEncoder(String columnName, int targetIndex, String[] domainValues, String[] newDomainValues) {
+    this.columnName = columnName;
+    this.targetIndex = targetIndex;
+    domainMap = new HashMap<>(domainValues.length);
```
honzasterba: can be moved to Line#10

h2o-genmodel/src/main/java/hex/genmodel/easy/EnumLimitedEncoder.java
```diff
@@ -0,0 +1,57 @@
+package hex.genmodel.easy;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class EnumLimitedEncoder implements CategoricalEncoder {
```
honzasterba: this class is a great candidate for a unit test

h2o-genmodel/src/main/java/hex/genmodel/easy/EnumLimitedEncoderColumnMapper.java
```diff
@@ -0,0 +1,25 @@
+package hex.genmodel.easy;
+
+import hex.genmodel.GenModel;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class EnumLimitedEncoderColumnMapper {
```
honzasterba: this looks exactly the same as EnumEncoderColumnMapper we should not just copy-paste code

koniecsveta: there is a difference in line 17: EnumLimitedEncoderColumnMapper takes _m.getOrigNames() as modelColumnNames and EnumEncoderColumnMapper takes _m.getNames() as modelColumnNames

h2o-genmodel/src/main/java/hex/genmodel/easy/EnumLimitedEncoder.java
koniecsveta: added tests for all new classes

h2o-genmodel/src/main/java/hex/genmodel/easy/EnumEncoderColumnMapper.java
```diff
@@ -7,14 +7,19 @@
 
 public class EnumEncoderColumnMapper {
 
-  private final GenModel _m;
+  final GenModel _m;
+  String[] modelColumnNames;
```
honzasterba: there is no reason to make this a instance variable a simple protected getModelColumnNames() method would be enough, together with making _m protected

koniecsveta: oh i see, fixed

honzasterba: 👍

PR #4383, opened 2020-03-05
OPS: Create and use proxy for gradle repo

sebhrusen:


build.gradle
```diff
@@ -4,7 +4,7 @@ import org.apache.tools.ant.taskdefs.condition.Os
 // classpath
 buildscript {
     repositories {
-        maven { url 'https://plugins.gradle.org/m2/' }
+        maven {url "${localNexusLocation}/public" }
```
michalkurka: only use Local nexus in CI builds:

eg.:

        if (project.ext.isCi) {
            maven {
                url "${localNexusLocation}/public"
            }

gradle.properties
```diff
@@ -40,7 +40,8 @@ doUploadUBenchResults=false
 #
 # Internal Nexus location
 #
-localNexusLocation=http://nexus.0xdata.loc:8081/nexus/repository
```
michalkurka: this keeps changing back and forth, recently Honza changed it to the full URL, is there a reason not to keep it as-is?

sebhrusen: ok, will revert: thought this was old due to the domain name

build.gradle
sebhrusen: right, otherwise you can't build when not on VPN... fixing this, thx.

sebhrusen: @michalkurka should be fixed now, I ran 2 clean builds locally after cleaning gradle cache: once in CI mode and with VPN, one in dev mode and without VPN.

PR #4385, opened 2020-03-06
PUBDEV-6955: Python - modify StackEnsemble's metalearner property to …

tomasfryda:
…return the metalearner; R - add ensemble@model$metalearner_model containing the metalearner

Python
======
Metalearner property now returns metalearner object; to keep it partially backwards compatible it can be indexed to retrieve a "name".

R
=
Added property `metalearner_model` to the stacked ensemble containing the metalearner.

Link to JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6955

### DIFF
--- a/h2o-bindings/bin/custom/python/gen_stackedensemble.py
+++ b/h2o-bindings/bin/custom/python/gen_stackedensemble.py
@@ -43,11 +43,24 @@ def class_extensions():
         ...                                           seed=1,
         ...                                           keep_levelone_frame=True)
         >>> stack_blend.train(x=x, y=y, training_frame=train, blending_frame=blend)
-        >>> stack_blend.metalearner
+        >>> stack_blend.metalearner()
         """
+        def _get_item(self, key):
+            warnings.warn(
+                "The usage of stacked_ensemble.metalearner()['name'] will be deprecated. "
+                "Metalearner now returns the metalearner object. If you need to get the "
+                "'name' please use stacked_ensemble.metalearner().model_id",
+                DeprecationWarning
+            )
+            if key == "name":
+                return self.model_id
+            raise NotImplementedError
+
         model = self._model_json["output"]
         if "metalearner" in model and model["metalearner"] is not None:
-            return model["metalearner"]
+            metalearner = h2o.get_model(model["metalearner"]["name"])
+            metalearner.__class__.__getitem__ = _get_item
+            return metalearner
         print("No metalearner for this model")
 
     def levelone_frame_id(self):
@@ -117,12 +130,15 @@ def class_extensions():
 
 extensions = dict(
     __imports__="""
+import ast
+import json
+import warnings
+
+import h2o
+from h2o.base import Keyed
+from h2o.grid import H2OGridSearch
 from h2o.utils.shared_utils import quoted
 from h2o.utils.typechecks import is_type
-from h2o.grid import H2OGridSearch
-from h2o.base import Keyed
-import json
-import ast
 """,
     __class__=class_extensions
 )
--- a/h2o-bindings/bin/custom/R/gen_stackedensemble.py
+++ b/h2o-bindings/bin/custom/R/gen_stackedensemble.py
@@ -95,6 +95,7 @@ class(model@model$model_summary) <- "h2o.stackedEnsemble.summary"
 .h2o.fill_stackedensemble <- function(model, parameters, allparams) {
   # Store base models for the Stacked Ensemble in user-readable form
   model$base_models <- unlist(lapply(parameters$base_models, function (base_model) base_model$name))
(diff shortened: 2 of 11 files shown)


h2o-bindings/bin/custom/python/gen_stackedensemble.py
```diff
@@ -45,9 +45,22 @@ def metalearner(self):
         >>> stack_blend.train(x=x, y=y, training_frame=train, blending_frame=blend)
         >>> stack_blend.metalearner
```
sebhrusen: can you please fix the doc at the same time? This was wrong from the beginning, this is not a property but a function:
```
stack_blend.metalearner()
```

h2o-bindings/bin/custom/python/gen_stackedensemble.py
```diff
@@ -45,9 +45,22 @@ def metalearner(self):
...
+            if key == "name":
+                return self.model_id
+            raise NotImplementedError
+
         model = self._model_json["output"]
         if "metalearner" in model and model["metalearner"] is not None:
-            return model["metalearner"]
+            metalearner = h2o.get_model(model["metalearner"]["name"])
+            metalearner.__class__.__getitem__ = _get_item
+            return metalearner
```
sebhrusen: perfect, thx for the deprecation message!

h2o-bindings/bin/custom/python/gen_stackedensemble.py
```diff
@@ -117,10 +130,13 @@ def extend_parms(parms):
 
 extensions = dict(
     __imports__="""
+import ast
+import json
+import warnings
+
+import h2o
```
sebhrusen: I know what you tried, I also try to organize imports top down:
1. core python
2. external libraries
3. current project

however, as you see below, the generated file already imports h2o modules before... so was being pragmatic here and grouped import in generated file :)

h2o-bindings/bin/custom/python/gen_stackedensemble.py
tomasfryda: sure

h2o-bindings/bin/custom/python/gen_stackedensemble.py
tomasfryda: Ok,  I will revert that to the original ordering. Should I order the imports just alphabetically or should I order them as core python alphabetically, then external alphabetically?

I know it will result in something like this:

```
from h2o.estimators.estimator_base import H2OEstimator
from h2o.exceptions import H2OValueError
from h2o.frame import H2OFrame
from h2o.utils.typechecks import assert_is_type, Enum, numeric
import ...
import ...
import ...
from h2o.utils.shared_utils import quoted
from h2o.utils.typechecks import is_type
```

sebhrusen: > core python alphabetically, then external alphabetically
That's what I'd like to enforce (you won't see it applied in most legacy files).
No big deal, but makes it easier to see how much a module depends from this or that.

sebhrusen: you know what?
can you please just fix `gen_python.py`
and mode lines 196-197 above the h2o imports?

It will modify all generated algos, but I'll be nicer.

tomasfryda: Ok, I guess then I should also modify the gen_*algo*.py to have sorted imports, right?

stacked ensemble imports look now like this:
```

import ast
import json
import warnings

import h2o
from h2o.utils.shared_utils import quoted
from h2o.utils.typechecks import is_type
from h2o.estimators.estimator_base import H2OEstimator
from h2o.exceptions import H2OValueError
from h2o.frame import H2OFrame
from h2o.utils.typechecks import assert_is_type, Enum, numeric
```

Would this be ok?

sebhrusen: yes, better than it's ever been!

tomasfryda:
I am thinking about the `getBaseModels` and `getMetalearner`, we could create a new S4 class and fill it "import" from the backend. That should be quite simple if the class is supposed to have exactly one state, i.e., we get the ensemble and we are sure that the base models and meta learner don't change at the backend - as far as I know S4 slots don't have getters/setters.

Otherwise if we need to refresh the state, we could assign closure (function referencing the outer scope at the time of creation) in to the slot so it would be more like `ensemble@base_models()`; I don't think S4 slots support active binding so we wouldn't be able to lose the parenthesis (I think only R6 classes support that).

sebhrusen:
@tomasfryda for future PRs, don't hesitate to add @ledell as a reviewer.
We don't require all reviewers to approve as soon as there's no ongoing discussion, but it makes it easier to follow what's going on, thanks.

PR #4386, opened 2020-03-06
TargetEncoderModel's summary prints fold column as column that is going to be encoded by this model

deil87:
This PR addresses 2 issues in TargetEncoder's summary:

1) fold_column is being printed as if TE model was trained to encode it
Screenshot 2020-03-06 at 18 11 59

2) on backend skipping with `continue` does not decrease size of TwoDimensional table as a result we get extra empty lines in the summary's printout


Pscheidl:
Great fix, thanks.

deil87:
Tests are good. If there are no objections will merge it tomorrow.

PR #4387, opened 2020-03-06
PUBDEV-7358: changed order of parameter for TargetEncoder's Flow UI

deil87:
Originally user was being asked to select ignored_columns ( from empty set) and only then comes training_frame parameter

Screenshot 2020-03-06 at 20 53 09

### DIFF
--- a/h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java
+++ b/h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java
@@ -5,6 +5,7 @@ import ai.h2o.targetencoding.TargetEncoderModel;
 import water.api.API;
 import water.api.schemas3.ModelParametersSchemaV3;
 
+import java.util.ArrayList;
 import java.util.List;
 
 public class TargetEncoderV3 extends ModelBuilderSchema<TargetEncoderBuilder, TargetEncoderV3, TargetEncoderV3.TargetEncoderParametersV3> {
@@ -30,13 +31,14 @@ public class TargetEncoderV3 extends ModelBuilderSchema<TargetEncoderBuilder, Ta
   
     @Override
     public String[] fields() {
-      final List<String> params = extractDeclaredApiParameters(getClass());
+      final List<String> params = new ArrayList<>();
       params.add("model_id");
-      params.add("ignored_columns");
       params.add("training_frame");
       params.add("fold_column");
       params.add("response_column");
-  
+      params.add("ignored_columns");
+      params.addAll(extractDeclaredApiParameters(getClass()));
+
       return params.toArray(new String[0]);
     }
 
--- a/h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java
+++ b/h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java
@@ -11,19 +11,19 @@ import java.util.List;
 public class TargetEncoderV3 extends ModelBuilderSchema<TargetEncoderBuilder, TargetEncoderV3, TargetEncoderV3.TargetEncoderParametersV3> {
   public static class TargetEncoderParametersV3 extends ModelParametersSchemaV3<TargetEncoderModel.TargetEncoderParameters, TargetEncoderParametersV3> {
     
-    @API(help = "Blending enabled/disabled")
+    @API(help = "Blending enabled/disabled", level = API.Level.secondary)
     public boolean blending;
 
-    @API(help = "Inflection point. Used for blending (if enabled). Blending is to be enabled separately using the 'blending' parameter.")
+    @API(help = "Inflection point. Used for blending (if enabled). Blending is to be enabled separately using the 'blending' parameter.", level = API.Level.secondary)
     public double k;
 
-    @API(help = "Smoothing. Used for blending (if enabled). Blending is to be enabled separately using the 'blending' parameter.")
+    @API(help = "Smoothing. Used for blending (if enabled). Blending is to be enabled separately using the 'blending' parameter.", level = API.Level.secondary)
     public double f;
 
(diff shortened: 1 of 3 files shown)


h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java
```diff
@@ -30,13 +31,14 @@
...
       params.add("model_id");
-      params.add("ignored_columns");
       params.add("training_frame");
       params.add("fold_column");
       params.add("response_column");
-  
+      params.add("ignored_columns");
+      params.addAll(extractDeclaredApiParameters(getClass()));
+
       return params.toArray(new String[0]);
```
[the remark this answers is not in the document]
deil87: @Pscheidl as far as know your first version(current one) is more efficient ( please take a look at this answer on SO https://stackoverflow.com/a/29444594/1020472 ). wdyt?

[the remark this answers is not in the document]
Pscheidl: Interesting. Let's leave it this way then.

deil87:
There is a Cloud size 1 under 4 issue but it should be addressed by separate PR https://github.com/h2oai/h2o-3/pull/4401

h2o-r/h2o-package/R/targetencoder.R
```diff
@@ -38,14 +38,14 @@
 h2o.targetencoder <- function(x,
                               y,
                               training_frame,
+                              model_id = NULL,
```
deil87: @sebhrusen I had to generate files with accordance to backend changes ( git diff check would fail otherwise). I think that this should be considered as a API breaking changes right? it looks like in our tests we always use named parameters
```
target_encoder <- h2o.targetencoder(training_frame = data, x = encoded_columns, y = "survived")
```
instead of relying just on order of parameters. But it means someone who does rely on order of specifying... will be in trouble. Or is there a convention in R that you should use named parameters style?

sebhrusen: technically, it is a potential breaking change.
practically, I think that R users (like Python users) use positional params only for functions with very few params, or for the first 2-3 params max: common usage is to use named params.
@ledell to confirm?

h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java
```diff
@@ -30,13 +31,14 @@
...
+      final List<String> params = new ArrayList<>();
       params.add("model_id");
-      params.add("ignored_columns");
       params.add("training_frame");
       params.add("fold_column");
       params.add("response_column");
-  
+      params.add("ignored_columns");
+      params.addAll(extractDeclaredApiParameters(getClass()));
+
```
sebhrusen: so the goal is to keep "standard" params in their usual position for the TE widget in Flow?
If so, definitely good, I remember being confused by params last time I saw it: looked like the params were in  random order.

sebhrusen: maybe you can tweak the `level` in the API annotation as well to emphasize important params, and group others as secondary

Pscheidl: Great idea. I do think this PR is ready to be merged. Small enough change with a purpose. We might change this everywhere in another PR - it could be nice introduce an order in general. Why not expand the annotation interfaces with ordinals ? Also, we could introduce implicit ordering (depend on class where the parameter was defined - attributes from parent classes go first). Etc etc.

sebhrusen: @Pscheidl 
> Why not expand the annotation interfaces with ordinals ?
Also thought about it in the past, could be useful in some cases. But as you point out, logic can become tricky when there is inheritance, for example some subclass may want to add a frame and insert it right after`validation frame` (defined in parent class), and I guess it's  difficult to keep this simple without overriding the property to replace the annotation, but then you may get issues if the property is duplicated...
There's room for improvement though.

deil87: Agree with @sebhrusen that is could be tricky and we might end up with overriding some fields in some cases. As I put it in the corresponding jira "Probably it is done to have flexibility of specifying order on per algo basis just by hardcoding it in the list of fields.".  Unfortunately for developers, hardcoding is sometimes the best/cheap way to provide flexibility/usability for users. It would be great to improve this aspect one day.
I'm not sure if we have a specific idea/proposal and how to formalise it in a separate JIRA but created a JIRA for a reference in the future https://0xdata.atlassian.net/browse/PUBDEV-7375

h2o-r/h2o-package/R/targetencoder.R
ledell: In my opinion, it's ok to add this in the middle to be consistent with other modeling functions.  It's not generally advised to do this, because it will break code of people who relied only on the order of the arguments, but we probably have very few people using this function yet, considering our User Guide doesn't even have the up-to-date examples for TE.

h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderV3.java
deil87: @sebhrusen 
>maybe you can tweak the level in the API annotation for TE as well to emphasize important params, and group others as secondary

thanks, did some changes in levels of the fields

h2o-r/h2o-package/R/targetencoder.R
deil87: @ledell @sebhrusen ok then. Going to merge after tests are passed.

deil87:
Tests are all green. Will merge after today's release.

sebhrusen:
@deil87 reminder ;)
> Tests are all green. Will merge after today's release.

PR #4388, opened 2020-03-09
PUBDEV-6787: Populate base_models attribute in Stacked Ensemble in Python and R

tomasfryda:
In python, if it doesn't have any `base_models` in `_parms`, it uses `actual_params` to retrieve `base_models` and it contains the names of the models.

In R, it populates `ensemble@model$base_models` with a vector of base model names.

### DIFF
--- a/h2o-bindings/bin/custom/R/gen_stackedensemble.py
+++ b/h2o-bindings/bin/custom/R/gen_stackedensemble.py
@@ -56,6 +56,7 @@ if (!missing(metalearner_params))
 if (!missing(metalearner_params)) {
     model@parameters$metalearner_params <- list(fromJSON(model@parameters$metalearner_params))[[1]] #Need the `[[ ]]` to avoid a nested list
 }
+model@model <- .h2o.fill_stackedensemble(model@model, model@parameters, model@allparams)
 model@model$model_summary <- capture.output({
 
   print_ln <- function(...) cat(..., sep = "\\n")
@@ -86,6 +87,13 @@ model@model$model_summary <- capture.output({
 
 })
 class(model@model$model_summary) <- "h2o.stackedEnsemble.summary"
+""",
+    module="""
+.h2o.fill_stackedensemble <- function(model, parameters, allparams) {
+  # Store base models for the Stacked Ensemble in user-readable form
+  model$base_models <- unlist(lapply(parameters$base_models, function (base_model) base_model$name))
+  return(model)
+}
 """
 )
 
--- a/h2o-bindings/bin/custom/python/gen_stackedensemble.py
+++ b/h2o-bindings/bin/custom/python/gen_stackedensemble.py
@@ -134,6 +134,13 @@ if is_type(base_models, {ptype}):
 else:
     assert_is_type({pname}, None, [str])
     self._parms["{sname}"] = {pname}
+""",
+        getter="""
+base_models = self.actual_params.get("base_models", [])
+base_models = [base_model["name"] for base_model in base_models]
+if len(base_models) == 0:
+    base_models = self._parms.get("base_models")
+return base_models
 """
     ),
 
--- a/h2o-r/h2o-package/R/kvstore.R
+++ b/h2o-r/h2o-package/R/kvstore.R
@@ -210,6 +210,11 @@ h2o.getModel <- function(model_id) {
     }
   })
 
+  # Run model specific hooks
+  model_fill_func <- paste0(".h2o.fill_", json$algo)
+  if (exists(model_fill_func, mode="function")) {
+    model <- do.call(model_fill_func, list(model, parameters, allparams))
+  }
 
   # Convert ignored_columns/response_column to valid R x/y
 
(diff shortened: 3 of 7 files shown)


h2o-bindings/bin/custom/python/gen_stackedensemble.py
```diff
@@ -134,6 +134,12 @@ def extend_parms(parms):
 else:
     assert_is_type({pname}, None, [str])
     self._parms["{sname}"] = {pname}
+""",
+        getter="""
+base_models = self._parms.get("base_models")
+if base_models is None and self.actual_params.get("base_models"):
```
sebhrusen: I would reverse the logic (first check `actual_params`, and fallback to local `base_models` value): source of truth is always the backend, we never know what happens there...

sebhrusen: also, is `self.actual_params["base_models"]` returning a list of `ids` or a list of complex keys?
One of the goal of the ticket is to provide something directly usable, we want sth like:
```
> print(se.base_models)
> ['model_1', 'model_2', 'model_3', ...]
```

h2o-bindings/bin/custom/R/gen_stackedensemble.py
```diff
@@ -56,6 +56,7 @@ def update_param(name, param):
 if (!missing(metalearner_params)) {
     model@parameters$metalearner_params <- list(fromJSON(model@parameters$metalearner_params))[[1]] #Need the `[[ ]]` to avoid a nested list
 }
+model@model$base_models <- unlist(base_models)
```
sebhrusen: see comments for Py API below.
`model@model$base_models` should be fed from backend value if available, and formatted as a vector of model ids, so that it can directly be reused (for another SE for example).

h2o-bindings/bin/custom/python/gen_stackedensemble.py
tomasfryda: Thanks for the suggestion. That makes much more sense.

tomasfryda: I understood from the Jira that we want the model IDs. I will have to check tomorrow if we really get them now. When I tested the previous version with the fallback to `actual_params` it returned in both cases the model IDs.

tomasfryda: I added assertion to the test to expect list of strings as base models so the test would fail in case of different implementation `actual_params` in future.

tomasfryda:
Sure, I will do that.

Just to clarify it to me: Aren't the models in AutoML filled in by `h2o.get_model` in python (which is included in test)? To me it seems better to test the model retrieved by the `get_model` since it would test the same thing but without training more models. Or is there some other catch?

sebhrusen:
@tomasfryda you're absolutely right, testing this using `h2o.get_model` is even better.

tomasfryda:
Okay, so I updated the R part as well to test the model retrieved by `h2o.getModel` and modified the `h2o.getModel` function to populate the base models. I wonder if there is some other place to put model specific logic that would be used when importing using `h2o.getModel`.

h2o-r/h2o-package/R/kvstore.R
```diff
@@ -210,6 +210,10 @@ h2o.getModel <- function(model_id) {
     }
   })
 
+  # Store base models for the Stacked Ensemble in user-readable form
+  if (json$algo == "stackedensemble"){
+    model$base_models <- unlist(lapply(parameters$base_models, function (base_model) base_model$name))
+  }
```
sebhrusen: yeah, that's exactly what I was afraid of with this approach of adding properties to `model` slot.
It's opening the door to "put any specific logic here" and clutter the algo's logic this file and their dedicated one.

Do you feel like adding a hook here?
At this point, you know the `algo`, you have the `model`, the user params (`parameters`) and all params (`allParams`): what about checking if there is a private method `.h2o.fill_{algo}` and call it if present, e.g.
```
fill_func = paste0(".h2o.fill_", algo)
if (exists(fill_func, mode="function") {
   do.call(fill_func, list(model, parameters, allparams))
}
```
This way you can define 
```
.h2o.fill_stackedensemble <- function(model, user_params, all_params) { ... }
```
in the `stackedensemble.R` file without polluting the `kvstore.R`.

Didn't really try it, but would be nice if it worked

tomasfryda: I will add the hook there. There will be just a slight change:
```
fill_func = paste0(".h2o.fill_", algo)
if (exists(fill_func, mode="function") {
  model <- do.call(fill_func, list(model, parameters, allparams))
}
```

because AFAIK R likes its function arguments immutable. I could do "magic" to get the parent environment  to modify the values I want from the caller environment but this seems nicer. If you have better idea I will be happy to hear it because this solution has a problem of being able to modify only the model (which might not be all that we will want to do). I could also return all three arguments and update all of them but I don't know if that's what we want...

sebhrusen: ok, great. I think just returning `model` is fine for now, and should cover the majority of customizations we might want to do.

PR #4389, opened 2020-03-09
PUBDEV-4534: Add the possibility to use Grids as base model for StackedEnsemble

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-4534
```
Currently, the `base_models` argument only accepts a list of model ids.
Let's expand this (on the client side) to support the following:
* a single grid
* list of grids
* a mix of grids and models

The backend should turn each type of input into a list of model ids.
```

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -4,6 +4,7 @@ import hex.Model;
 import hex.ModelBuilder;
 import hex.ModelCategory;
 
+import hex.grid.Grid;
 import water.DKV;
 import water.Job;
 import water.Key;
@@ -13,6 +14,7 @@ import water.exceptions.H2OModelBuilderIllegalArgumentException;
 import water.fvec.Frame;
 import water.fvec.Vec;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -71,6 +73,28 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
     super.init(expensive);
 
     checkFoldColumnPresent(_parms._metalearner_fold_column, train(), valid(), _parms.blending());
+    validateAndExpandBaseModels();
+  }
+
+  /**
+   * Validates base models and if grid is provided instead of a model it gets expanded.
+   */
+  private void validateAndExpandBaseModels() {
+    List<Key> baseModels = new ArrayList<Key>();
+    for (Key baseModelKey : _parms._base_models) {
+      Object retrievedObject = DKV.getGet(baseModelKey);
+      if (retrievedObject instanceof Model) {
+        baseModels.add(baseModelKey);
+      } else if (retrievedObject instanceof Grid) {
+        Grid grid = (Grid) retrievedObject;
+        Collections.addAll(baseModels, grid.getModelKeys());
+      } else if (retrievedObject == null) {
+        throw new IllegalArgumentException(String.format("Specified id \"%s\" does not exist.", baseModelKey));
+      } else {
+        throw new IllegalArgumentException(String.format("Unsupported type \"%s\" as a base model.", retrievedObject.getClass().toString()));
+      }
+    }
+    _parms._base_models = baseModels.toArray(new Key[0]);
   }
 
   /**
--- a/h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
+++ b/h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
@@ -57,8 +57,9 @@ public class StackedEnsembleV99 extends ModelBuilderSchema<StackedEnsemble,Stack
 
     // Base models
     @API(level = API.Level.critical, direction = API.Direction.INOUT,
-            help = "List of models (or model ids) to ensemble/stack together. If not using blending frame, then models must have been cross-validated using nfolds > 1, and folds must be identical across models.", required = true)
-    public KeyV3.ModelKeyV3 base_models[];
+            help = "List of models or grids (or their ids) to ensemble/stack together. Grids are expanded to individual models. "
+                    + "If not using blending frame, then models must have been cross-validated using nfolds > 1, and folds must be identical across models.", required = true)
+    public KeyV3 base_models[];
 
     
     // Metalearner algorithm
--- a/h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
+++ b/h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
@@ -31,6 +31,7 @@ import water.util.Log;
 
 import java.util.*;
 
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.fail;
 
@@ -1098,5 +1099,64 @@ public class StackedEnsembleTest extends TestUtil {
     }
   }
 
+  @Test
+  public void testBaseModelsWorkWithGrid() {
+    GBMModel gbmModel = null;
+    try {
+      Scope.enter();
 
+      final Frame trainingFrame = TestUtil.parse_test_file("./smalldata/junit/weather.csv");
+      Scope.track(trainingFrame);
+      trainingFrame.toCategoricalCol("RainTomorrow");
+
+      GBMModel.GBMParameters parameters = new GBMModel.GBMParameters();
+      parameters._train = trainingFrame._key;
+      parameters._seed = 0xFEED;
+      parameters._response_column = "RainTomorrow";
+      parameters._ntrees = 1;
+      parameters._keep_cross_validation_predictions = true;
+
+      GBM gbm = new GBM(parameters);
+      gbmModel = gbm.trainModel().get();
+      assertNotNull(gbmModel);
+
+      final Integer[] maxDepthArr = new Integer[]{2, 3, 4};
+      HashMap<String, Object[]> hyperParms = new HashMap<String, Object[]>() {{
+        put("_distribution", new DistributionFamily[]{DistributionFamily.bernoulli});
+        put("_max_depth", maxDepthArr);
+      }};
+
+      GBMModel.GBMParameters params = new GBMModel.GBMParameters();
+      params._train = trainingFrame._key;
+      params._response_column = "RainTomorrow";
+      params._seed = 42;
+
+      Job<Grid> gs = GridSearch.startGridSearch(null, params, hyperParms);
+      Scope.track_generic(gs);
+      final Grid grid = gs.get();
+      Scope.track_generic(grid);
+
+      final StackedEnsembleParameters seParams = new StackedEnsembleParameters();
+      seParams._train = trainingFrame._key;
+      seParams._response_column = "RainTomorrow";
+      seParams._metalearner_algorithm = Algorithm.AUTO;
+      seParams._base_models = new Key[]{gbmModel._key, grid._key};
+      seParams._seed = 0xFEED;
+
+      final StackedEnsemble stackedEnsemble = new StackedEnsemble(seParams);
+      List<Key> expectedBaseModels = new ArrayList<Key>();
+      expectedBaseModels.add(gbmModel._key);
+      Collections.addAll(expectedBaseModels, grid.getModelKeys());
+
+      assertArrayEquals(expectedBaseModels.toArray(new Key[0]), stackedEnsemble._parms._base_models);
+    } finally {
+      Scope.exit();
+
+      if(gbmModel != null){
+        gbmModel.deleteCrossValidationModels();
+        gbmModel.deleteCrossValidationPreds();
+        gbmModel.remove();
+      }
+    }
+  }
 }
--- a/h2o-bindings/bin/custom/R/gen_stackedensemble.py
+++ b/h2o-bindings/bin/custom/R/gen_stackedensemble.py
@@ -23,21 +23,16 @@ if (is.null(training_frame)) training_frame <- blending_frame  # guarantee prese
     validate_params="""
 # Get the base models from model IDs (if any) that will be used for constructing model summary
 if(!is.list(base_models) && is.vector(x)) {
-   base_models <- as.list(base_models)
+  base_models <- if (inherits(base_models, "H2OGrid")) list(base_models) else as.list(base_models)
 }
-baselearners <- lapply(base_models, function(base_model) {
-  if (is.character(base_model))
-    base_model <- h2o.getModel(base_model)
-  base_model
-})
 
 # Get base model IDs that will be passed to REST API later
 if (length(base_models) == 0) stop('base_models is empty')
 
 # If base_models contains models instead of ids, replace with model id
 for (i in 1:length(base_models)) {
-  if (inherits(base_models[[i]], 'H2OModel')) {
-    base_models[[i]] <- base_models[[i]]@model_id
+  if (inherits(base_models[[i]], c('H2OModel', 'H2OGrid'))) {
+    base_models[[i]] <- h2o.keyof(base_models[[i]])
   }
 }
 """,
@@ -57,6 +52,14 @@ if (!missing(metalearner_params)) {
     model@parameters$metalearner_params <- list(fromJSON(model@parameters$metalearner_params))[[1]] #Need the `[[ ]]` to avoid a nested list
 }
 model@model <- .h2o.fill_stackedensemble(model@model, model@parameters, model@allparams)
+
+# Get the actual models (that were potentially expanded from H2OGrid on the backend)
+baselearners <- lapply(model@model$base_models, function(base_model) {
+  if (is.character(base_model))
+    base_model <- h2o.getModel(base_model)
+  base_model
+})
+
 model@model$model_summary <- capture.output({
 
   print_ln <- function(...) cat(..., sep = "\\n")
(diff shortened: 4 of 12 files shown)


h2o-r/tests/testdir_algos/stackedensemble/runit_stackedensemble_base_models_using_grid.R
```diff
@@ -0,0 +1,147 @@
...
+    expect_equal_unordered <- function (lhs, rhs) {
+        expect_equal(sort(lhs), sort(rhs))
+    }
+
+    train <- h2o.uploadFile(locate("smalldata/testng/higgs_train_5k.csv"),
+                            destination_frame = "higgs_train_5k")
+
+    valid <- h2o.uploadFile(locate("smalldata/testng/higgs_test_5k.csv"),
+                            destination_frame = "higgs_test_5k")
+
```
sebhrusen: please avoid this dataset for basic tests, `higgs_5k` is not that small, and we don't want the tests taking too long. Besides, most of the tests you have to write are for ensembles and automl, which means that the test has to train multiple models...

I'll rather recommend `prostate/prostate.csv`, `junit/cars.csv`, `junit/australia.csv`, `junit/iris.csv`, `junit/weather.csv`... let's say datasets < 50kB, and keep the larger ones to test distributed code, running on multiple nodes.

sebhrusen: I see that the Python counterpart is using the higgs dataset too: looks like I forgot to change it last time I modified this file. Feel free to change the dataset there as well, thx.

h2o-bindings/bin/custom/R/gen_stackedensemble.py
```diff
@@ -23,13 +23,12 @@ def update_param(name, param):
     validate_params="""
 # Get the base models from model IDs (if any) that will be used for constructing model summary
 if(!is.list(base_models) && is.vector(x)) {
-   base_models <- as.list(base_models)
+  if (inherits(base_models, "H2OGrid")) {
+    base_models <- list(base_models)
+  } else {
+    base_models <- as.list(base_models)
+  }
```
sebhrusen: can be written as a single expression, a bit like in Python: ensures that there'll never be any side-effect sneaking into this logic
```
base_models <- if (inherits(base_models, "H2OGrid")) list(base_models) else as.list(base_models)
```

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -71,6 +73,28 @@ public void init(boolean expensive) {
...
 
     checkFoldColumnPresent(_parms._metalearner_fold_column, train(), valid(), _parms.blending());
+    validateAndExpandBaseModels();
+  }
+
+  /**
+   * Validates base models and if grid is provided instead of a model it gets expanded.
+   */
+  private void validateAndExpandBaseModels() {
+    List<Key> base_models = new ArrayList<Key>();
```
sebhrusen: welcome to the mess of H2O naming "conventions" in Java.
Will try to sum it up:
1. use camelCase by default: local variables, method parameters, attributes, methods...
2. use snake_case for the fields exposed in Schemas (REST API uses snake_case).
3. fields in the POJOs mapped to the schema classe may use snake_case (this is because we use Java reflection by default to fill Schemas from POJOs and vice versa).
4. class attributes (and only those) are prefixed with underscore (except in Schemas, as it is the REST API).
5. test methods can be written camelCase or snake_case, this is because it is mush easier to read `test_stackedensemble_using_blending_with_invalid_model` than its camelCase equivalent..

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -71,6 +73,28 @@ public void init(boolean expensive) {
...
+    for (Key base_model: _parms._base_models) {
+      Object retrieved_object = DKV.getGet(base_model);
+      if (retrieved_object instanceof Model) {
+        base_models.add(base_model);
+      } else if (retrieved_object instanceof Grid) {
+        Grid grid = (Grid) retrieved_object;
+        Collections.addAll(base_models, grid.getModelKeys());
```
sebhrusen: as you're using an ArrayList, I don't think you get any benefit in using `Collections.addAll`, perf is probably even worse than a simple `list.addAll(...)`

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -71,6 +73,28 @@ public void init(boolean expensive) {
...
+      } else if (retrieved_object == null) {
+        continue;
```
sebhrusen: please throw exception here as well: principle of less surprise.
If user specifies an id that doesn't exist, we should tell him immediately.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -71,6 +73,28 @@ public void init(boolean expensive) {
...
+      } else {
+        throw new IllegalArgumentException(String.format("Don't know yet how to use %s as a base model.", retrieved_object.getClass()));
```
sebhrusen: Please keep it standard, sth like "Unsupported type %s as a base model."

h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
```diff
@@ -58,7 +58,7 @@ public AlgorithmValuesProvider() {
     // Base models
     @API(level = API.Level.critical, direction = API.Direction.INOUT,
             help = "List of models (or model ids) to ensemble/stack together. If not using blending frame, then models must have been cross-validated using nfolds > 1, and folds must be identical across models.", required = true)
```
sebhrusen: can you please update the `help` in annotation? This is used to generate doc for clients (including fields description in Flow).

h2o-bindings/bin/custom/R/gen_stackedensemble.py
```diff
@@ -38,6 +37,8 @@ def update_param(name, param):
 for (i in 1:length(base_models)) {
   if (inherits(base_models[[i]], 'H2OModel')) {
     base_models[[i]] <- base_models[[i]]@model_id
+  } else if (inherits(base_models[[i]], 'H2OGrid')) {
+    base_models[[i]] <- base_models[[i]]@grid_id
```
sebhrusen: can you please use the generic `h2o.keyof` instead? I created it for similar purposes, but apparently forgot to add it for grid.
cf `classes.R`:
```
setMethod("h2o.keyof", signature("H2OModel"), function(object) object@model_id)
```
you can add the equivalent for "H2OGrid" lower in the file, and simplify this:
```
if (inherits(base_models[[i]], c("H2OModel", "H2OGrid") 
   base_models[[i]] <- h2o.keyof(base_models[[i]])
```

h2o-bindings/bin/custom/R/gen_stackedensemble.py
```diff
@@ -57,6 +58,14 @@ def update_param(name, param):
     model@parameters$metalearner_params <- list(fromJSON(model@parameters$metalearner_params))[[1]] #Need the `[[ ]]` to avoid a nested list
 }
 model@model <- .h2o.fill_stackedensemble(model@model, model@parameters, model@allparams)
+
```
sebhrusen: now you see why it's better to trust the backend in your previous ticket about `base_models` ;)
I didn't even think about current ticket at that time.

h2o-bindings/bin/custom/python/gen_stackedensemble.py
```diff
@@ -128,11 +129,23 @@ def extend_parms(parms):
...
-    {pname} = [b.model_id for b in {pname}]
+def _get_id(something):
+    if hasattr(something, "model_id"):
+        return something.model_id
+    elif hasattr(something, "grid_id"):
+        return something.grid_id
+    elif is_type(something, str):
+        return something
+    else:
+        raise ValueError("Type {{}} is not supported as a base model.".format(type(something)))
```
sebhrusen: Same as for R, looks like I forgot to make `H2OGridSearch` a subclass of `Keyed`.
Please make the change (same as for `ModelBase`).

h2o-r/tests/testdir_algos/stackedensemble/runit_stackedensemble_base_models_using_grid.R
```diff
@@ -0,0 +1,147 @@
...
+                                  y=y,
+                                  training_frame = train,
+                                  base_models = as.list(c(grid1@model_ids, my_gbm@model_id, grid2@model_ids))
+    )
+
+    expect_equal_unordered(get_base_models(stack7), get_base_models(stack8))
+
+    # Check that we validate base models
+    # For some reason the error is printed so to make cleaner test logs I capture the output here as
+    # it could be confusing seeing exception in passing test
```
sebhrusen: good point, I'm sure some of our legacy tests would benefit from this...

h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_base_models.py
```diff
@@ -109,8 +111,94 @@ def test_base_models_are_populated():
...
+    # Single Grid
+    se1 = H2OStackedEnsembleEstimator(base_models=gs1.models, seed=seed)
+    se1.train(data.x, data.y, data.train)
+
+    se2 = H2OStackedEnsembleEstimator(base_models=gs1, seed=seed)
+    se2.train(data.x, data.y, data.train)
+
+    assert sorted(se1.base_models) == sorted(se2.base_models)
+
+    # List of Grids
```
sebhrusen: if possible, create a test method for each use-case.
Python fails at first failed assertion, so status is more clear when all tests are executed: also tests run in isolation this way.
It's a bit different with R test runner, but still, 1 use-case = 1 test is necessary for proper isolation.

h2o-bindings/bin/custom/R/gen_stackedensemble.py
tomasfryda: That's a good point. I haven't thought of that :)

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
tomasfryda: This is very helpful!

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
tomasfryda: Maybe I am missing something, but I see 3 options I have here - `getModelKeys` returns `Key<Model>[]`.

So either I use:
1. `Collections.addAll(baseModels, grid.getModelKeys());`, or
2. `baseModels.addAll(Arrays.asList(grid.getModelKeys()));`, or
3. for cycle adding one element at a time

Or is there some better way of doing it?

h2o-bindings/bin/custom/R/gen_stackedensemble.py
tomasfryda: Good to know about this method. 

I have question about R coding style. I can see that you don't use curly brackets here for the single line expression. Is that to prevent people subconsciously adding some more lines of code in the if expr.? Or is it just to keep the example shorter?

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
sebhrusen: are you concerned about the generic type? solutions #1 and #2 will bring an additional cost, just to be able to sweep the explicit cast under the rug.
Explicit cast should work:
```
baseModels.addAll((Key[])grid.getModelKeys());
```

sebhrusen: sorry, I though that `grid.getModelKeys()` was returning  a `List`

h2o-bindings/bin/custom/R/gen_stackedensemble.py
sebhrusen: > to keep the example shorter

personally I'm using brackets for `block` statements to distinguish clearly from expressions (same for inline functions).
Looks like it is compatible with tidyverse: https://style.tidyverse.org/syntax.html#indenting

h2o-bindings/bin/custom/python/gen_stackedensemble.py
```diff
@@ -128,11 +129,18 @@ def extend_parms(parms):
 overrides = dict(
     base_models=dict(
         setter="""
-if is_type(base_models, {ptype}):
-    {pname} = [b.model_id for b in {pname}]
+def _get_id(something):
+    if hasattr(something, "key"):
```
sebhrusen: strictly speaking, I'd prefer `if isinstance(something, Keyed)`, I know it's not following duck-typing rules, but `key` is a pretty common property, and only classes extending `Keyed` have it with the correct semantic.

h2o-bindings/bin/custom/python/gen_stackedensemble.py
```diff
@@ -128,11 +129,18 @@ def extend_parms(parms):
...
+        return something.key
+    return something
+
+if not is_type(base_models, list):
+    base_models = [base_models]
+if is_type(base_models, [H2OEstimator, H2OGridSearch, str]):
```
sebhrusen: having `str` there confuses me.
Is it now simpler to have:
```
if is_type(base_models, [Keyed]):
    {pname} = [b.key for b in {pname}]
    self._parms["{sname}"] = {pname}
else:
    assert_is_type({pname}, [None, str])
    self._parms["{sname}"] = {pname}
```

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
```diff
@@ -1098,5 +1099,66 @@ public void testInvalidFoldColumn_trainingFrame() {
...
+      parameters._response_column = "RainTomorrow";
+      parameters._ntrees = 1;
+      parameters._keep_cross_validation_predictions = true;
+
+      GBM gbm = new GBM(parameters);
+      gbmModel = gbm.trainModel().get();
+      assertNotNull(gbmModel);
+
+      final Integer[] ntreesArr = new Integer[]{5, 50, 7, 8, 9, 10};
+      final Integer[] maxDepthArr = new Integer[]{2, 3, 4};
```
sebhrusen: please keep it minimal for tests: 3-4 models in the grid is enough here, not need for 18.

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
```diff
@@ -1098,5 +1099,66 @@ public void testInvalidFoldColumn_trainingFrame() {
...
+      seParams._metalearner_algorithm = Algorithm.AUTO;
+      seParams._base_models = new Key[]{gbmModel._key, grid._key};
+      seParams._seed = 0xFEED;
+
+      final StackedEnsemble stackedEnsemble = new StackedEnsemble(seParams);
+      List<Key> expectedBaseModels = new ArrayList<Key>();
+      expectedBaseModels.add(gbmModel._key);
+      Collections.addAll(expectedBaseModels, grid.getModelKeys());
+
+      assertArrayEquals(expectedBaseModels.toArray(new Key[0]), stackedEnsemble._parms._base_models);
```
sebhrusen: am I missing sth here? 
this SE hasn't been trained, so `stackedEnsemble._parms._base_models` should be just equal to `seParams._base_models = new Key[]{gbmModel._key, grid._key}`.
then you can train the SE model, and check the `base_models` attribute on the SE, not on the `params`:
```
assertArrayEquals(expectedBaseModels.toArray(new Key[0]), stackedEnsemble.base_models);
```

h2o-bindings/bin/custom/python/gen_stackedensemble.py
tomasfryda: I was thinking about that but it wouldn't let the user to mix the `model_ids` and actual model objects in the `base_models`. Should I drop the use case with mixed `model_ids` and model objects?

sebhrusen: Oh, I see, good point; you can keep the mixed use case.
Just remove the assignment for `None` then:
`self._parms["{sname}"] = {pname}`
It's unnecessary now. Thx.

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
tomasfryda: I am probably missing here something. The test I wrote works because it expands the `_parms._base_models` during `StackedEnsemble.init`. I didn't find any attribute `StackedEnsemble.base_models` or `StackedEnsembleModel.base_models` etc. and wherever I looked the `_parms._base_models` was used so I used it as well.

The only `base_model` attribute I could find is in schemas and I don't know yet how to get to it after the model was trained. Any hints?

sebhrusen: forgot about it, must be tired... for a second I thought this was a client test...

h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_base_models.py
```diff
@@ -18,8 +20,8 @@
 
 
 def prepare_data(blending=False):
-    fr = h2o.import_file(path=pu.locate("smalldata/testng/higgs_train_5k.csv"))
-    target = "response"
+    fr = h2o.import_file(path=pu.locate("smalldata/junit/weather.csv"))
```
ledell: @tomasfryda Curious why you changed the test dataset... is it smaller/quicker to run the tests?

tomasfryda: yes, exactly it's a smaller dataset so it should be quicker to run the tests.

PR #4390, opened 2020-03-09
[PUBDEV-7229] Add CBOW to Word2Vec

honzasterba:


h2o-py/h2o/estimators/word2vec.py
```diff
@@ -96,9 +96,9 @@ def min_word_freq(self, min_word_freq):
     @property
     def word_model(self):
         """
-        Use the Skip-Gram model
+        The word model to use (SkipGram or CBOW)
 
-        One of: ``"skip_gram"``  (default: ``"skip_gram"``).
+        One of: ``"skip_gram"``, ``"cbow"``  (default: ``"skip_gram"``).
```
honzasterba: need to look into this if we recognize both SkipGram and skip_gram, the docs are inconsitent

honzasterba: all of above works, so lets leave it

PR #4392, opened 2020-03-10
PUBDEV-7293: Make sure we don't overload first node when parsing 40k+ Parquet files

michalkurka:
https://0xdata.atlassian.net/browse/PUBDEV-7293

### DIFF
--- a/h2o-core/src/main/java/water/persist/Persist.java
+++ b/h2o-core/src/main/java/water/persist/Persist.java
@@ -7,6 +7,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import water.*;
+import water.fvec.Vec;
 import water.util.Log;
 
 /** Abstract class describing various persistence targets.
@@ -24,6 +25,11 @@ public abstract class Persist {
 
   /** Load a previously stored Value */
   abstract public byte[] load(Value v) throws IOException;
+
+  public byte[] load(Key k, long skip, int max) throws IOException {
+    throw new UnsupportedOperationException(
+            "Persist Backend " + this.getClass().getSimpleName() + " doesn't support direct data read.");
+  }
 
   /** Reclaim space from a previously stored Value */
   abstract public void delete(Value v);
@@ -214,6 +220,37 @@ public abstract class Persist {
   }
 
   public InputStream open(String path) {
+    throw new RuntimeException("Not implemented");
+  }
+
+  /**
+   * Creates a seekable Hadoop implementation of InputStream (FSDataInputStream)
+   * 
+   * h2o-core doesn't depend on Hadoop libraries and can thus not declare the return type specifically
+   * 
+   * @param path any H2O-3 allowed path
+   * @return instance of FSDataInputStream
+   */
+  public InputStream openSeekable(String path) {
+    throw new RuntimeException("Not implemented");
+  }
+
+  /**
+   * Indicates whether this Persist backend can natively support Seekable InputStreams
+   * 
+   * @return true, if openSeekable can be safely called
+   */
+  public boolean isSeekableOpenSupported() {
+    return false;
+  }
+
+  /**
+   * Creates a Seekable InputStream for a given Vec
+   * 
+   * @param vec any Vec in theory typically a ByteVec/FileVec
+   * @return instance of FSDataInputStream
+   */
+  public InputStream wrapSeekable(Vec vec) {
     throw new RuntimeException("Not implemented");
   }
 
--- a/h2o-core/src/main/java/water/persist/PersistManager.java
+++ b/h2o-core/src/main/java/water/persist/PersistManager.java
@@ -5,6 +5,8 @@ import water.*;
 import water.api.FSIOException;
 import water.api.HDFSIOException;
 import water.exceptions.H2OIllegalArgumentException;
+import water.fvec.FileVec;
+import water.fvec.Vec;
 import water.parser.BufferedString;
 import water.util.FileUtils;
 import water.util.Log;
@@ -82,11 +84,15 @@ public class PersistManager {
   }
 
   private void validateHdfsConfigured() {
-    if (I[Value.HDFS] == null) {
+    if (!isHdfsConfigured()) {
       throw new H2OIllegalArgumentException("HDFS, S3, S3N, and S3A support is not configured");
     }
   }
 
+  private boolean isHdfsConfigured() {
+    return I[Value.HDFS] != null;
+  }
+  
   public boolean isGcsPath(String path) {
     return path.toLowerCase().startsWith("gs://");
   }
@@ -186,6 +192,13 @@ public class PersistManager {
   public byte[] load(int backend, Value v) throws IOException {
     stats[backend].load_count.incrementAndGet();
     byte[] arr = I[backend].load(v);
+    stats[backend].load_bytes.addAndGet(arr.length);
+    return arr;
+  }
+
+  public byte[] load(int backend, Key k, long skip, int max) throws IOException {
+    stats[backend].load_count.incrementAndGet();
+    byte[] arr = I[backend].load(k, skip, max);
     stats[backend].load_bytes.addAndGet(arr.length);
     return arr;
   }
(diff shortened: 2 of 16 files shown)


h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/VecParquetReader.java
```diff
@@ -115,7 +128,7 @@ public void close() throws IOException {
   public static byte[] readFooterAsBytes(Vec vec) {
     FSDataInputStream f = null;
     try {
-      f = new FSDataInputStream(new VecDataInputStream(vec));
+      f = (FSDataInputStream) H2O.getPM().openSeekable(vec);
```
Pscheidl: Let's make make sure the underlying PM implementation used for Parquest is seekable. Ideally introduce a test.

If it's not seekable, we should react accordingly (open a "normal" input stream ?).

h2o-core/src/main/java/water/persist/PersistManager.java
```diff
@@ -598,6 +607,25 @@ public InputStream open(String path) {
     }
   }
 
+  public InputStream openSeekable(Vec vec) {
```
Pscheidl: The contract of the method could be improved. This operation of opening a seekable stream can fail - in many ways. One of them is the underlying implementation does not support that. This should be somehow part of the contract, so the caller is able to protect itself. E.g see how it's used in `VecParquetReader`.

The fallback does `validateHdfsConfigured` which might throw `throw new H2OIllegalArgumentException("HDFS, S3, S3N, and S3A support is not configured");`- let's add some defense lines. I wouldn't walk the way of adding `Optional`, an exception to be caught is perfectly fine.

h2o-core/src/main/java/water/persist/Persist.java
```diff
@@ -217,6 +223,18 @@ public InputStream open(String path) {
     throw new RuntimeException("Not implemented");
   }
 
+  public InputStream openSeekable(String path) {
+    throw new RuntimeException("Not implemented");
```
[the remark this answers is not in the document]
Pscheidl: The Message could be improved as well, but that will be anyway I'm sure. It's about the class of the exception.

h2o-core/src/main/java/water/persist/PersistManager.java
Pscheidl:
```patch
Index: h2o-core/src/main/java/water/persist/PersistManager.java
IDEA additional info:
<+>UTF-8
--- h2o-core/src/main/java/water/persist/PersistManager.java	(revision e9e4b94cd40cb83d8bd81a96192a7aa526a02695)
+++ h2o-core/src/main/java/water/persist/PersistManager.java	(date 1583937314435)
@@ -82,6 +82,10 @@
     }
     return false;
   }
+  
+  private boolean isHdfsConfigured(){
+    return I[Value.HDFS] == null;
+  }
 
   private void validateHdfsConfigured() {
     if (I[Value.HDFS] == null) {
@@ -607,7 +611,7 @@
     }
   }
 
-  public InputStream openSeekable(Vec vec) {
+  public InputStream openSeekable(Vec vec) throws IOException {
     if (vec instanceof FileVec) {
       FileVec fileVec = (FileVec) vec;
       final String path = fileVec.getPath();
@@ -618,11 +622,16 @@
         Persist p = I[fileVec.getBackend()];
         if (p != null && p.isSeekableOpenSupported()) {
           return p.openSeekable(path);
+        } else {
+          Log.debug("Target path not seekable. Falling back to HDFS wrapper.");
         }
       }
     }
     // fallback
-    validateHdfsConfigured();
+    if(!isHdfsConfigured()){
+      throw new IllegalArgumentException(String.format("Target path '%s' not seekable. Unable to use seekable wrapper, HDFS not configured",
+              vec._key));
+    }
     return I[Value.HDFS].wrapSeekable(vec);
   }
   
```

Pscheidl:
```patch
Index: h2o-core/src/main/java/water/persist/PersistManager.java
IDEA additional info:
<+>UTF-8
--- h2o-core/src/main/java/water/persist/PersistManager.java	(revision e9e4b94cd40cb83d8bd81a96192a7aa526a02695)
+++ h2o-core/src/main/java/water/persist/PersistManager.java	(date 1583937559043)
@@ -82,6 +82,10 @@
     }
     return false;
   }
+  
+  private boolean isHdfsConfigured(){
+    return I[Value.HDFS] == null;
+  }
 
   private void validateHdfsConfigured() {
     if (I[Value.HDFS] == null) {
@@ -607,7 +611,14 @@
     }
   }
 
-  public InputStream openSeekable(Vec vec) {
+  /**
+   * Opens a {@link Vec} in seekable implementation of an {@link InputStream}.
+   * @param vec An instance of {@link Vec} to open
+   * @return A seekable instanceo of {@link InputStream}, never null.
+   * @throws IOException When the underlying resource does not allow seekable resource creation and all fallback solutions
+   * failed. Or on reading error.
+   */
+  public InputStream openSeekable(final Vec vec) throws IOException {
     if (vec instanceof FileVec) {
       FileVec fileVec = (FileVec) vec;
       final String path = fileVec.getPath();
@@ -618,11 +629,16 @@
         Persist p = I[fileVec.getBackend()];
         if (p != null && p.isSeekableOpenSupported()) {
           return p.openSeekable(path);
+        } else {
+          Log.debug("Target path not seekable. Falling back to HDFS wrapper.");
         }
       }
     }
     // fallback
-    validateHdfsConfigured();
+    if(!isHdfsConfigured()){
+      throw new IllegalArgumentException(String.format("Target path '%s' not seekable. Unable to use seekable wrapper, HDFS not configured",
+              vec._key));
+    }
     return I[Value.HDFS].wrapSeekable(vec);
   }
   
```

h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/VecParquetReader.java
michalkurka: yes, adding a test is a good idea - will do

> If it's not seekable, we should react accordingly (open a "normal" input stream ?).

I don't see how this could happen other than someone purposefully excluding persist-hdfs from the assembly - this code would fail but there are other places where it wouldn't even load the classes (since there are hard dependencies on that module)

h2o-core/src/main/java/water/persist/Persist.java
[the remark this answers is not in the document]
michalkurka: the current version is consistent with rest of the class, I think i prefer to keep it consistent

thank you for the suggestion

h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/VecParquetReader.java
Pscheidl: This comment is outdated now. Let's resolve it. Tests are always good anway.

michalkurka: okay, thank you Pavel

h2o-persist-hdfs/src/test/java/water/persist/PersistHdfsTest.java
```diff
@@ -28,11 +32,17 @@
   @BeforeClass
   public static void setup() { stall_till_cloudsize(1); }
```
Pscheidl: Commercial time / hint: Perfect opportunity to take advantage of the new and improved H2ORunner :)

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -179,24 +216,23 @@ public PersistHdfs(URI uri) {
...
+    }, true, max);
     end = System.currentTimeMillis();
     if (end-start > 1000) // Only log read that took over 1 second to complete
-      Log.debug("Slow Read: "+(end-start)+" millis to get bytes "+skip_ +"-"+(skip_+b.length)+" in HDFS read.");
+      Log.debug("Slow Read: "+(end-start)+" millis to get bytes "+skip +"-"+(skip+b.length)+" in HDFS read.");
 
     return b;
   }
-
+
```
Pscheidl:
```suggestion

```

PR #4394, opened 2020-03-11
PUBDEV-7275 -do not allow null keys for frames

robingood:


h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1161,7 +1161,7 @@ public void init(boolean expensive) {
     }
     Frame tr = _train != null?_train:_parms.train();
     if( tr == null ) { error("_train", "Missing training frame: "+_parms._train); return; }
-    setTrain(new Frame(null /* not putting this into KV */, tr._names.clone(), tr.vecs().clone()));
+    setTrain(new Frame(tr._key, tr._names.clone(), tr.vecs().clone()));
```
michalkurka: This is not a correct fix, everything should work with a key-less Frame, the Training frame should never be itself installed in DKV. The original Frame should also never be touched during model building.

robingood:
I will make a test then. Until then, here is a screenshot with the StackTrace when saving a frame to the K/V store although the frame_key is null. ModelMetrics -> buildKey method doesn't seem to care if the frame_key is null as long as the Frame is not null. 
on_null

michalkurka:
@robingood based on the screenshot it seems to me that the metrics in k-means are not created properly - the metrics should be associated to with the original, not the adapted Frame

robingood:
@michalkurka  thank you. I will direct my changes in that direction, after I will create the failing test.

michalkurka:
Closing due inactivity

PR #4396, opened 2020-03-11
PUBDEV-7371: Add an option to specify multiple Hadoop config files

michalkurka:
example:

    java -jar h2o.jar -hdfs_config core-site.xml -hdfs_config hdfs-site.xml

### DIFF
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -365,7 +365,7 @@ final public class H2O {
     // HDFS & AWS
     //-----------------------------------------------------------------------------------
     /** -hdfs_config=hdfs_config; configuration file of the HDFS */
-    public String hdfs_config = null;
+    public String[] hdfs_config = null;
 
     /** -hdfs_skip=hdfs_skip; used by Hadoop driver to not unpack and load any HDFS jar file at runtime. */
     public boolean hdfs_skip = false;
@@ -591,7 +591,7 @@ final public class H2O {
       }
       else if (s.matches("hdfs_config")) {
         i = s.incrementAndCheck(i, args);
-        trgt.hdfs_config = args[i];
+        trgt.hdfs_config = ArrayUtils.append(trgt.hdfs_config, args[i]);
       }
       else if (s.matches("hdfs_skip")) {
         trgt.hdfs_skip = true;
--- a/h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
+++ b/h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
@@ -48,17 +48,17 @@ public final class PersistHdfs extends Persist {
   };
 
   // Global HDFS initialization
-  // FIXME: do not share it via classes, but initialize it by object
   static {
-    Configuration conf = null;
-    if( H2O.ARGS.hdfs_config != null ) {
-      conf = new Configuration();
-      File p = new File(H2O.ARGS.hdfs_config);
-      if( !p.exists() ) H2O.die("Unable to open hdfs configuration file " + p.getAbsolutePath());
-      conf.addResource(new Path(p.getAbsolutePath()));
-      Log.debug("resource ", p.getAbsolutePath(), " added to the hadoop configuration");
+    final Configuration conf = new Configuration();
+    if (H2O.ARGS.hdfs_config != null && H2O.ARGS.hdfs_config.length > 0) {
+      for (String config : H2O.ARGS.hdfs_config) {
+        File p = new File(config);
+        if (!p.exists())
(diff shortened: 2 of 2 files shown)


michalkurka:
@mn-mikke does SW need to reflect this change?

h2o-core/src/main/java/water/H2O.java
```diff
@@ -591,7 +591,7 @@ else if (s.matches("nthreads")) {
       }
       else if (s.matches("hdfs_config")) {
         i = s.incrementAndCheck(i, args);
-        trgt.hdfs_config = args[i];
+        trgt.hdfs_config = ArrayUtils.append(trgt.hdfs_config, args[i]);
```
Pscheidl: Enlarging the array here is fine, as we don't know the number of HDFS confings in advance.

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -48,17 +48,17 @@
   };
 
   // Global HDFS initialization
-  // FIXME: do not share it via classes, but initialize it by object
```
Pscheidl: :+1:

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -48,17 +48,17 @@
...
-    if( H2O.ARGS.hdfs_config != null ) {
-      conf = new Configuration();
-      File p = new File(H2O.ARGS.hdfs_config);
-      if( !p.exists() ) H2O.die("Unable to open hdfs configuration file " + p.getAbsolutePath());
-      conf.addResource(new Path(p.getAbsolutePath()));
-      Log.debug("resource ", p.getAbsolutePath(), " added to the hadoop configuration");
+    final Configuration conf = new Configuration();
+    if (H2O.ARGS.hdfs_config != null && H2O.ARGS.hdfs_config.length > 0) {
+      for (String config : H2O.ARGS.hdfs_config) {
+        File p = new File(config);
```
Pscheidl:
```suggestion
        final File p = new File(config);
```

Tiny detail, as we started to use final 3 lines above.

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -48,17 +48,17 @@
...
+        if (!p.exists())
```
Pscheidl:
```suggestion
        if (!p.exists()) {
          H2O.die("Unable to open hdfs configuration file " + p.getAbsolutePath());
        }
```

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -48,17 +48,17 @@
...
+          H2O.die("Unable to open hdfs configuration file " + p.getAbsolutePath());
```
Pscheidl:
```suggestion
        if (!p.exists()) {
          H2O.die("Unable to open hdfs configuration file " + p.getAbsolutePath());
        }
```

Just a matter of style. Don't worry about it if you don't like it. This way, the boundary of the block is much more clearly defined.

mn-mikke:
@michalkurka IMHO, this shouldn't break anything in SW. The question is whether we should expose ability to set multiple files manually as well. https://github.com/h2oai/sparkling-water/blob/2b4dbeb5ade89605f4616c31f71bcc263069ca5c/core/src/main/scala/ai/h2o/sparkling/backend/internal/InternalBackendConf.scala#L65, but this is a rare case.

The feature that @pavel implemented for the internal backend dumps the whole spark hadoop configuration into one big file and passes that.

h2o-persist-hdfs/src/main/java/water/persist/PersistHdfs.java
```diff
@@ -48,17 +48,17 @@
...
+        conf.addResource(new Path(p.getAbsolutePath()));
+        Log.debug("resource ", p.getAbsolutePath(), " added to the hadoop configuration");
+      }
     } else {
-      conf = new Configuration();
```
Pscheidl: :+1:

Pscheidl:
> @michalkurka IMHO, this shouldn't break anything in SW. The question is whether we should expose ability to set multiple files manually as well. https://github.com/h2oai/sparkling-water/blob/2b4dbeb5ade89605f4616c31f71bcc263069ca5c/core/src/main/scala/ai/h2o/sparkling/backend/internal/InternalBackendConf.scala#L65, but this is a rare case.
> 
> The feature that @pavel implemented for the internal backend dumps the whole spark hadoop configuration into one big file and passes that.

I can only agree. 100% true.

PR #4400, opened 2020-03-11
fix R CMD Check failure

wendycwong:
@Michalk pointed out that R CMD Check and R CRAN check have been failing.

From the error message, it looks like that it does not like F and insist on using FALSE.  I went to the file and changed all T to TRUE and F to FALSE.


michalkurka:
thank you Wendy!

PR #4409, opened 2020-03-13
[PUBDEV-7347] implement leaf node assignments for XGBoost

honzasterba:
- do the same for IsoFor while we're at it


h2o-genmodel-extensions/xgboost/src/main/java/hex/genmodel/algos/xgboost/XGBoostMojoModel.java
```diff
@@ -100,6 +100,21 @@ public boolean requiresOffset() {
...
+  }
+
+  @Override
+  public int getNTreesPerGroup() {
+    return _nclasses > 2 ? _nclasses : 1;
+  }
+
+  @Override
+  public String[] getDecisionPath(double[] row) {
+    return new String[0];
```
Pscheidl: Is this correct ?

honzasterba: no! good catch!

h2o-genmodel/src/test/java/hex/genmodel/easy/EasyPredictModelWrapperTest.java
```diff
@@ -190,7 +187,6 @@ public void testSerializeWrapper() throws Exception {
     SupervisedModel rawModel = makeSupervisedModel();
     EasyPredictModelWrapper m = new EasyPredictModelWrapper(rawModel);
 
-    Assert.assertTrue(m instanceof Serializable);
```
michalkurka: does it cause any issue?

honzasterba: its always true

michalkurka: ok

PR #4413, opened 2020-03-16
PUBDEV-7288: Add dummy modeling step

tomasfryda:
I am not sure about labels for this one - no AutoML label unlike in JIRA, so please add appropriate labels.

link to JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7288

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -602,6 +602,11 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     return Key.make(algoName + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
   }
 
+  Key<Model> dummyKey(String algoName, boolean with_counter) {
+    String counterStr = with_counter ? "_" + nextInstanceCounter(algoName, algoInstanceCounters) : "";
+    return Key.make(algoName + "_dummy_" + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
+  }
+
   Key<Grid> gridKey(String algoName, boolean with_counter) {
     String counterStr = with_counter ? "_" + nextInstanceCounter(algoName, gridInstanceCounters) : "";
     return Key.make(algoName + "_grid_" + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
--- a/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
@@ -602,11 +602,6 @@ public final class AutoML extends Lockable<AutoML> implements TimedH2ORunnable {
     return Key.make(algoName + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
   }
 
-  Key<Model> dummyKey(String algoName, boolean with_counter) {
-    String counterStr = with_counter ? "_" + nextInstanceCounter(algoName, algoInstanceCounters) : "";
-    return Key.make(algoName + "_dummy_" + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
-  }
-
   Key<Grid> gridKey(String algoName, boolean with_counter) {
     String counterStr = with_counter ? "_" + nextInstanceCounter(algoName, gridInstanceCounters) : "";
     return Key.make(algoName + "_grid_" + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
--- a/h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
@@ -1,6 +1,7 @@
 package ai.h2o.automl;
 
 import ai.h2o.automl.AutoMLBuildSpec.AutoMLCustomParameters;
+import ai.h2o.automl.events.EventLogEntry;
 import ai.h2o.automl.events.EventLogEntry.Stage;
 import ai.h2o.automl.WorkAllocations.JobType;
 import ai.h2o.automl.WorkAllocations.Work;
@@ -13,6 +14,7 @@ import hex.ensemble.StackedEnsembleModel;
 import hex.grid.Grid;
 import hex.grid.GridSearch;
 import hex.grid.HyperSpaceSearchCriteria.RandomDiscreteValueSearchCriteria;
+import water.H2O;
 import water.Iced;
 import water.Job;
 import water.Key;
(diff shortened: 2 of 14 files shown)


sebhrusen:
@tomasfryda good idea: just created the label

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -602,6 +602,11 @@ private int nextInstanceCounter(String algoName, NonBlockingHashMap<String, Inte
     return Key.make(algoName + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
   }
 
+  Key<Model> dummyKey(String algoName, boolean with_counter) {
+    String counterStr = with_counter ? "_" + nextInstanceCounter(algoName, algoInstanceCounters) : "";
+    return Key.make(algoName + "_dummy_" + counterStr + "_AutoML_" + timestampFormatForKeys.format(_startTime));
+  }
+
```
sebhrusen: this should not appear here, please see comment below

h2o-py/tests/testdir_algos/automl/pyunit_automl_dummy_steps.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def test_modeling_plan_dummies_exist():
+    modeling_plan = [
+        dict(name='DeepLearning', steps=[dict(id='dummy_sleep', weight=10)]),  # sleep 1 secs
+        dict(name='DRF', steps=[dict(id='dummy_sleep', weight=10)]),
+        dict(name='GBM', steps=[dict(id='dummy_sleep', weight=10)]),
+        dict(name='GLM', steps=[dict(id='dummy_sleep', weight=10)]),
+        dict(name='StackedEnsemble', steps=[dict(id='dummy_sleep', weight=10)]),
+        dict(name='XGBoost', steps=[dict(id='dummy_sleep', weight=10)]),
```
sebhrusen: as you get the point of modeling plan, with my approach described above, you would just have the `DummyStepProvider` have `getName() -> "Dummy"`, so that you would use it (with simplified syntax) as:
```
modeling_plan = [
  ('Dummy', ['dummy_sleep', 'dummy_whatever_convenient_for_test'])
]
```

sebhrusen: also note that in the modeling plan, I think that if you need at least 3 models, you can repeat it multiple times:
```
modeling_plan = [
  ('Dummy', ['dummy_sleep', 'dummy_sleep', 'dummy_sleep'])
]
```

the alternative would be to provide a dummy grid generating max 10 models for example/

sebhrusen:
@tomasfryda reminder, please paste the JIRA link in the PR description.

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
```diff
@@ -352,4 +354,52 @@ protected Work getAllocatedWork() {
...
+
+        @Override
+        protected Key makeKey(String name, boolean withCounter) {
+            return aml().dummyKey(name, withCounter);
+        }
+
+        @Override
+        protected Job<M> startJob() {
+            // Sleep weight/10 seconds
+            long chunks = _weight;
```
sebhrusen: oh, I see you're using the weights here :) does make sense somehow.

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
```diff
@@ -352,4 +354,52 @@ protected Work getAllocatedWork() {
...
+            return job.start(new H2O.H2OCountedCompleter() {
+                @Override
+                public void compute2() {
+                    try {
+                        for (int i = 0; i < chunks; i++) {
+                            // Pretend some work
+                            Thread.sleep(100);
+                            // Update the status
+                            job.update(1);
+                        }
```
sebhrusen: ideally, you should also be able to have this dummy step providing a proper (but untrained) model, just use the constructor (e.g. DRFModel) and put it in DKV.
Also the idea is to use this to test automl logic for `max_runtime_secs`, making the assumption that the individual models will respect the `max_runtime_secs` param that we set for each of them.
What it means, is that the dummy step should sleep for the `max_runtime_secs` that is computed for it, by the logic in `ModelingStep`.
However, I'm afraid that the current logic there doesn't make it easy: I had to change this in my current branch though, so it will be easier once merged.

h2o-py/tests/testdir_algos/automl/pyunit_automl_dummy_steps.py
```diff
@@ -0,0 +1,64 @@
...
+import time
+
+sys.path.insert(1, os.path.join("..","..",".."))
+import h2o
+from tests import pyunit_utils as pu
+from h2o.automl import H2OAutoML
+
+"""This test suite checks the AutoML dummy modeling steps"""
+
+
```
sebhrusen: Idea is not to test the DummyStep, but use it for our tests relying on timeout logic.

deil87: @sebhrusen I think that having tests for our custom tool, that tests with timeout logic are going to rely on, is a necessary and a good thing. Within a scope of this PR I would expect exactly this kind of tests.

sebhrusen: having test is always fine, just saying that as the Jira ticket mentions a lot the question of testing, the goal is to use it for specific tests

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
tomasfryda: @sebhrusen I am struggling now with creating the DRFModel without training it I always get NullPointerException (see below). If I understand the nature of the problem, AutoML tries to get metrics from the untrained model. Any ideas how to deal with that?

```
        at ai.h2o.automl.leaderboard.Leaderboard.getMetrics(Leaderboard.java:461)
        at ai.h2o.automl.leaderboard.Leaderboard.addModels(Leaderboard.java:374)
        at ai.h2o.automl.leaderboard.Leaderboard.addModel(Leaderboard.java:395)
        at ai.h2o.automl.ModelingStepsExecutor.addModel(ModelingStepsExecutor.java:170)
        at ai.h2o.automl.ModelingStepsExecutor.monitor(ModelingStepsExecutor.java:147)
        at ai.h2o.automl.ModelingStepsExecutor.submit(ModelingStepsExecutor.java:66)
        at ai.h2o.automl.AutoML.learn(AutoML.java:574)
        at ai.h2o.automl.AutoML.run(AutoML.java:377)
        at ai.h2o.automl.H2OJob$1.compute2(H2OJob.java:33)
        at water.H2O$H2OCountedCompleter.compute(H2O.java:1470)
```

tomasfryda: @sebhrusen for now, I have implemented custom logic for the `max_runtime_secs`, mainly so I can test the implementation somehow.

It works like this: 
1. if `max_runtime_secs_per_model` is positive, use `min(max_runtime_secs_per_model, remaining_time)`
2. if not (1.) and `max_runtime_secs` is positive, use `min(max_runtime_secs, remaining_time)`
3. if not (1.) and not (2.), use `min(_weight, remaining_time)`

where `remaining_time` is `aml().timeRemainingMs()/1e3` and `max_runtime_secs{,_per_model}` are `aml().getBuildSpec().build_control.stopping_criteria.max_runtime_secs{,_per_model}()`.

h2o-py/tests/testdir_algos/automl/pyunit_automl_dummy_steps.py
tomasfryda: I agree that these tests are probably not necessary, I wrote them mainly for me as a sanity check. If necessary, I will remove them before merging but until this PR reaches the state when I will know I don't have to debug/modify DummySteps logic I would keep it here just to make it easier for me...

h2o-automl/src/main/java/ai/h2o/automl/modeling/DummyStepsProvider.java
```diff
@@ -0,0 +1,143 @@
...
+                params._class_sampling_factors = buildSpec.build_control.class_sampling_factors;
+                params._max_after_balance_size = buildSpec.build_control.max_after_balance_size;
+            }
+
+            params._keep_cross_validation_predictions = true;
+            params._keep_cross_validation_models = buildSpec.build_control.keep_cross_validation_models;
+            params._keep_cross_validation_fold_assignment = buildSpec.build_control.nfolds != 0 && buildSpec.build_control.keep_cross_validation_fold_assignment;
+            params._export_checkpoints_dir = buildSpec.build_control.export_checkpoints_dir;
+
+            return params;
```
sebhrusen: most of this logic is unnecessary. Actually I would completely remove it, as it overrides the default logic done in `ModelingStep` and would prevent testing the logic in that class.

h2o-automl/src/main/java/ai/h2o/automl/modeling/DummyStepsProvider.java
```diff
@@ -0,0 +1,143 @@
...
+                        .max_runtime_secs_per_model();
+                // if max_runtime_secs_per_model is not specified (or is negative) try to use max_runtime_secs_per_model
+                if (maxRuntimeSecs <= 0) {
+                    maxRuntimeSecs = aml().getBuildSpec().build_control.stopping_criteria.max_runtime_secs();
+                }
+                // if both max_runtime_secs{,_per_model} are not specified or negative use weight
+                if (maxRuntimeSecs <= 0) {
+                    maxRuntimeSecs = _weight;
+                }
+                maxRuntimeSecs = Math.min(maxRuntimeSecs, aml().timeRemainingMs() / 1e3);
```
sebhrusen: same here: the whole idea of `startJob` method is to focus ONLY on the parameters specific to that step, i.e. tuning algo parameters.
All the more generic params, stopping criteria, frames and columns, xval and so on are assigned in the parent class to avoid duplication and ensure consistency (and in case of `max_runtime_secs`, to ensure that it actually works).

PR #4415, opened 2020-03-16
PUBDEV-7380: Note in docs that "transform" is only for numeric data

angela0xdata:


ck37:
Thanks for working on this, appreciate it. I'm just wondering though: do binary variables count as numeric? Because for GLRM it's ambiguous: when using hinge loss they presumably should not be scaled (as shown in [correctForBinaryLoss](https://github.com/h2oai/h2o-3/pull/4415)), but otherwise they can be considered numeric rather than categorical.

wendycwong:
@ck37: When H2O parse a file, it uses a certain rule to determine whether a column is numeric, enum and others.  If you need a column to be read as a specific type especially the columns that can be interpreted as enum or numeric, please use the parse parameter col_types when you call import_file in Python.

However, if you have decided to make a column binary, it will be treated as binary.  It may be represented as a numeric though but it will not have numerical meanings less greater than or equal to and things like that.

ck37:
I appreciate the work on this but I think something has been lost in translation here. I have inserted the text of my question into https://0xdata.atlassian.net/browse/PUBDEV-7380 as well.

My original question remains: when specifying transform = "standardize" in GLRM, are binary variables (i.e. variables with a hinge or logistic loss) standardized or skipped?

The edits in this PR unfortunately do not seem to clarify this ambiguity, by my reading at least. Apologies if I'm just missing something here though.

PR #4418, opened 2020-03-17
PUBDEV-7234_return_cluster_info_in_2D_table

koniecsveta:
https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7234

### DIFF
--- a/h2o-py/h2o/backend/cluster.py
+++ b/h2o-py/h2o/backend/cluster.py
@@ -4,6 +4,7 @@ from __future__ import division, print_function, absolute_import, unicode_litera
 
 import sys
 import time
+import json
 
 import h2o
 from h2o.exceptions import H2OConnectionError, H2OServerError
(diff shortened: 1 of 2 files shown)


h2o-py/h2o/backend/cluster.py
```diff
@@ -257,6 +257,51 @@ def show_status(self, detailed=False):
...
+                    "swap_mem", "free_disk", "max_disk", "pid", "num_keys", "tcps_active", "open_fds", "rpcs_active"]
+            all_nodes_table = ["Node %d" % (i + 1) for i in range(len(self.nodes))]
+            node_table = [[k] for k in keys]
+            for j, node in enumerate(self.nodes):
+                for i, k in enumerate(keys):
+                    node_table[i].append(node[k])
+                all_nodes_table[j] = [all_nodes_table[j], node_table]
+            table = [table, all_nodes_table]
+        
+        return table
```
michalkurka: Should this be using H2OTwoDimTable for proper displaying?

koniecsveta: will use

PR #4420, opened 2020-03-17
Generic model predictions are compares with original model predictions. Scope management used test-wide.

Pscheidl:
- Predictions of generic models are compared with original model predictions as well, not only with pure GenModel/Mojo via testJavaScoring.
- Manual DKV management of each Keyed entry has been replaced by Scope management for every test method.

**No new functionality added in this PR.** Only test refactoring and enhancement.


Pscheidl:
Roger. Thanks.

PR #4426, opened 2020-03-17
[PUBDEV-7386] hive delegation token refresh igores runAsUser

honzasterba:


michalkurka:
@honzasterba I think our tests should reflect what Steam is doing as this is the most common form of deployment, can the smoke tests handle this scenario?

PR #4430, opened 2020-03-18
PUBDEV-7337: SQL Import: Fixes & perf improvements

michalkurka:


h2o-core/src/main/java/water/jdbc/SQLManager.java
```diff
@@ -725,52 +733,29 @@ public void map(Chunk[] cs, NewChunk[] ncs) {
...
+      else {
+        if (res instanceof Long || res instanceof Integer || res instanceof Short || res instanceof Byte)
+          nc.addNum(((Number) res).longValue(), 0);
+        else if (res instanceof Number)
+          nc.addNum(((Number) res).doubleValue());
+        else if (res instanceof Boolean)
+          nc.addNum(((boolean) res ? 1 : 0), 0);
+        else if (res instanceof String)
+          nc.addStr(res);
+        else if (res instanceof Date || res instanceof Time || res instanceof Timestamp)
```
honzasterba: '|| res instanceof Time || res instanceof Timestamp ' is unnecessary no? why not just do java.util.Date?

michalkurka: yes, thank you for the suggestion!

PR #4431, opened 2020-03-18
Added seed for algos when applicable.

wendycwong:
OMG:  I forgot to add seed to the appropriate algos.

### DIFF
--- a/h2o-py/tests/testdir_jira/pyunit_pubdev_4702_algo_max_runtime_secs_large.py
+++ b/h2o-py/tests/testdir_jira/pyunit_pubdev_4702_algo_max_runtime_secs_large.py
@@ -35,7 +35,7 @@ def algo_max_runtime_secs():
     training1_data = h2o.import_file(path=pyunit_utils.locate("smalldata/gridsearch/glrmdata1000x25.csv"))
     x_indices = list(range(training1_data.ncol))
     model = H2OGeneralizedLowRankEstimator(k=10, loss="Quadratic", gamma_x=0.3,
-                                           gamma_y=0.3, transform="STANDARDIZE")
+                                           gamma_y=0.3, transform="STANDARDIZE", seed=seed)
     grabRuntimeInfo(err_bound, 2.0, model, training1_data, x_indices)
     cleanUp([training1_data, model])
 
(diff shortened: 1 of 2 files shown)


h2o-py/tests/testdir_jira/pyunit_pubdev_4702_algo_max_runtime_secs_large.py
```diff
@@ -68,32 +68,32 @@ def algo_max_runtime_secs():
...
-    if H2ODeepWaterEstimator.available():
-        training1_data = h2o.import_file(path=pyunit_utils.locate("smalldata/gbm_test/ecology_model.csv"))
-        training1_data = training1_data.drop('Site')
-        training1_data['Angaus'] = training1_data['Angaus'].asfactor()
-        y_index = "Angaus"
-        x_indices = list(range(1, training1_data.ncol))
-        model = H2ODeepWaterEstimator(epochs=50, hidden=[4096, 4096, 4096], hidden_dropout_ratios=[0.2, 0.2, 0.2])
-        grabRuntimeInfo(err_bound, 2.0, model, training1_data, x_indices, y_index)
-        cleanUp([training1_data, model])
+    # # deepwater
```
maurever: I think we can remove these lines. :) @Pscheidl had already removed deepwater from h2o-3.

h2o-r/tests/testdir_jira/runit_pubdev_4702_algo_max_runtime_secs_large.R
```diff
@@ -77,19 +77,19 @@ function() {
...
-    training1_data <- training1_data.drop('Site')
-    training1_data['Angaus'] <- as.factor(training1_data['Angaus'])
-    hh <- h2o.deepwater(x=c(2:h2o.ncol(training1_data)), y="Angaus", training_frame=training1_data, epochs=50, hidden=c(4096, 4096, 4096), hidden_dropout_ratios=c(0.2, 0.2, 0.2))
-    hh2 <- h2o.deepwater(x=c(2:h2o.ncol(training1_data)), y="Angaus", training_frame=training1_data, epochs=50, hidden=c(4096, 4096, 4096), hidden_dropout_ratios=c(0.2, 0.2, 0.2), max_runtime_secs=hh@model$run_time/(1000.0*fact_red))
-    test_pass_fail <- c(test_pass_fail, eval_test_runtime(hh, hh2, err_bound, fact_red))
-    cleanUp(c(training1_data, hh, hh2))
-  } else {
-    print("*************  deepwater is skipped.  Not availabe.")
-  }
+  # # deepwater
```
maurever: Same here. Thank you!

h2o-py/tests/testdir_jira/pyunit_pubdev_4702_algo_max_runtime_secs_large.py
Pscheidl: Yes @maurever is right. This is kind-of tricky, as deepwater removal will be done in next major. So it's only in master. Everything related to DeepWater should be ignored now, as it gone.

wendycwong: Let me remove it then.

PR #4433, opened 2020-03-19
PUBDEV-7087 implement XGBoost AUC-PR

maurever:
Our implementation calculates AUC-PR using linear interpolation.

XGBoost uses more sofisticated calculation. 

Goal is to implement this calculation to h2o.

JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7087


h2o-core/src/main/java/hex/AUC2.java
```diff
@@ -278,6 +269,50 @@ private double compute_auc() {
...
+    if (isEmpty()) {
+      return Double.NaN;
+    }
+    checkRecallValidity();
+    
+    if (_fps[_nBins-1] == 0) return 1.0;
+    if (_tps[_nBins-1] == 0) return 0.0;
+    
+    double area = 0.0;
+    assert _p > 0 && _n > 0 : "AUC-PR calculation error, sum of positives and sum of negatives should be zero.";
```
Pscheidl:
```suggestion
    assert _p > 0 && _n > 0 : "AUC-PR calculation error, sum of positives and sum of negatives should be greater than zero.";
```

Maybe we could provide the information about how many positives and negatives are there as well, but it's optional.

h2o-core/src/main/java/hex/AUC2.java
```diff
@@ -278,6 +269,50 @@ private double compute_auc() {
...
+        if (tp == prevtp) {
+          a = 1.0;
+          b = 0.0;
+        } else {
+          h = (fp - prevfp) / (tp - prevtp);
+          a = 1.0 + h;
+          b = (prevfp - h * prevtp) / _p;
+        }
+        if (0.0 != b) {
+          area += (tp / _p - prevtp / _p -
```
Pscheidl: Tiny performance tweak: you may force to calculate `true positives / actual positives` aka `tp / _p` once. Same goes for ` prevtp / _p`

h2o-core/src/main/java/hex/AUC2.java
maurever: Good catch, thank you!

h2o-core/src/main/java/hex/AUC2.java
maurever: I think this is not a crucial performance part, but I edited it as you suggested. :) Thank you!

PR #4436, opened 2020-03-19
[PUBDEV-7386] fix impersonation for hive delegation tokens

honzasterba:


h2o-mapreduce-generic/src/main/java/water/hadoop/DelegationTokenRefresher.java
```diff
@@ -67,6 +67,7 @@ public DelegationTokenRefresher(
 
   public void start() {
     if (HiveTokenGenerator.isHiveDriverPresent()) {
+      refreshTokens();
```
honzasterba: this is unrelated, but improves user experience, since delegation token is acquired immediatelly after cluster startup

PR #4440, opened 2020-03-20
PUBDEV-7396: Parquet parser doesn't support arbitrary precision Decimal type

michalkurka:


h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/ParquetParser.java
```diff
@@ -277,7 +277,10 @@ private static ParquetPreviewParseWriter readFirstRecords(ParquetParseSetup init
           types[i] = OriginalType.TIMESTAMP_MILLIS.equals(parquetType.getOriginalType()) ? Vec.T_TIME : Vec.T_NUM;
           break;
         default:
-          types[i] = Vec.T_BAD;
+          if (OriginalType.DECIMAL.equals(parquetType.getOriginalType())) {
```
honzasterba: what is the getPrimitiveTypeName value here? should we rather convert the whole swithc to if-else if that is not usable?

h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/ext/DecimalUtils.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Note: this code 1-1 copy of https://github.com/apache/parquet-mr/blob/master/parquet-pig/src/main/java/org/apache/parquet/pig/convert/DecimalUtils.java
+ * All credit goes to original Parquet contributors
+ * 
+ * Conversion between Parquet Decimal Type to Java BigDecimal in Pig
+ * Code Based on the Apache Spark ParquetRowConverter.scala
+ */
+public class DecimalUtils {
+
+  public static BigDecimal binaryToDecimal(Binary value, int precision, int scale) {
```
honzasterba: unit test would make this much easier to read

h2o-parsers/h2o-parquet-parser/src/main/java/water/parser/parquet/ParquetParser.java
michalkurka: good point, @honzasterba please take another look

PR #4443, opened 2020-03-23
PUBDEV-7394: Fix base_models selection for StackedEnsemble

tomasfryda:
Fix for the base_models selection for SE consists of three minor changes:

1. fix in `StackedEnsemble.java` in method `validateAndExpandBaseModels` to not iterate `base_models` when they are `null`

2. fix in `Schema.java` in method `fillFromAny` to properly parse `KeyV3`

3. fix in H2O-Flow repo to `controls.coffee` to render model selector for `Key<Keyed>[]`

Link to JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7394

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -80,6 +80,9 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
    * Validates base models and if grid is provided instead of a model it gets expanded.
    */
   private void validateAndExpandBaseModels() {
+    // H2O Flow initializes SE with no base_models
+    if (_parms._base_models == null) return;
+
     List<Key> baseModels = new ArrayList<Key>();
     for (Key baseModelKey : _parms._base_models) {
       Object retrievedObject = DKV.getGet(baseModelKey);
--- a/h2o-docs/src/product/flow/packs/test-small/index.list
+++ b/h2o-docs/src/product/flow/packs/test-small/index.list
@@ -218,3 +218,4 @@ glm_junit_weights_all_ones.flow
 glm_junit_weights.flow
 grid_search_test.flow
 automl_prostate.flow
+stackedensemble_weather.flow
--- a/h2o-docs/src/product/flow/packs/test-small/stackedensemble_weather.flow
+++ b/h2o-docs/src/product/flow/packs/test-small/stackedensemble_weather.flow
@@ -1,0 +1,41 @@
+{
+  "version": "1.0.0",
+  "cells": [
+    {
+      "type": "cs",
+      "input": "importFiles [ \"../smalldata/junit/weather.csv\" ]"
+    },
+    {
+      "type": "cs",
+      "input": "setupParse paths: [ \"../smalldata/junit/weather.csv\" ]"
+    },
+    {
+      "type": "cs",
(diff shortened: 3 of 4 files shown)


h2o-core/src/main/java/water/api/Schema.java
```diff
@@ -513,8 +515,15 @@ public S fillFromAny(Object o) {
...
-      return KeyV3.make(fclz, Key.make(s.startsWith("\"") ? s.substring(1, s.length() - 1) : s)); // If the key name is in an array we need to trim surrounding quotes.
+      // If the key name is in an array we need to trim surrounding quotes.
+      s = s.startsWith("\"") ? s.substring(1, s.length() - 1) : s;
+      // guess if the s is a JSON encoded value
+      if (s.startsWith("{")) { // PUBDEV-7394
+        // s should contain here just the JSON encoded value ({"value":"xxxx"})
+        Properties parsedS = gson.fromJson(s, Properties.class);
+        s = parsedS.getProperty("value");
+      }
+      return KeyV3.make(fclz, Key.make(s));
```
sebhrusen: I don't undestand why you have to do this: is it just for Flow? If sth works for other clients, but not for Flow, we should fix it on Flow side.

sebhrusen: Also, don't like the ad-hoc parsing. If you obtain a json with a value attribute, what is it supposed to be mapped to?

h2o-core/src/main/java/water/api/Schema.java
```diff
@@ -500,7 +500,9 @@ public S fillFromAny(Object o) {
     }
 
     // Are we parsing an object from a string? NOTE: we might want to make this check more restrictive.
-    if (! fclz.isAssignableFrom(Schema.class) && s != null && s.startsWith("{") && s.endsWith("}")) {
+    if (! fclz.isAssignableFrom(Schema.class) &&
+        ! fclz.isAssignableFrom(KeyV3.class) && // PUBDEV-7394
```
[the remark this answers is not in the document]
sebhrusen: Also there's something fishy here: why this conditional block did suddenly start to match `KeyV3` so that you had to exclude it?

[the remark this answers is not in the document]
sebhrusen: I just want to ensure we're not adding incidental complexity here.

h2o-core/src/main/java/water/api/Schema.java
tomasfryda: I managed to find a place where I can fix it in the flow which would complete remove the need to modify schema.java...

Here is my proposed fix:
```
diff --git a/src/ext/modules/proxy.coffee b/src/ext/modules/proxy.coffee
index 04b3c92d..3375310f 100644
--- a/src/ext/modules/proxy.coffee
+++ b/src/ext/modules/proxy.coffee
@@ -135,6 +135,8 @@ exports.init = (_) ->
           if isNumber element
             return element
           if isObject element
+            if 'value' in element
+              return stringify element.value
             return stringify element
           return "\"#{element}\""
         "[#{mappedArray.join ','}]"
```

The problem is that it's quite hard to test flow and the `element` representing the `KeyV3` is just plain javascript object of type object containing just a single attribute called `value`.

sebhrusen: I don't think it's the right either

ledell:
Thanks @tomasfryda!

PR #4444, opened 2020-03-24
PUBDEV-5808: Add max_runtime_secs to Stacked Ensemble

tomasfryda:
This PR consists of:

- `max_runtime_secs` is exposed in R,  Python, and Flow.

- updated the `max_runtime_secs` page and the Stacked Ensemble page in the User Guide.

Link to Jira: https://0xdata.atlassian.net/browse/PUBDEV-5808

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
+++ b/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
@@ -37,7 +37,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
   protected P _metalearner_parameters;
   protected boolean _hasMetalearnerParams;
   protected long _metalearnerSeed;
-  protected long _remainingTime;
+  protected long _maxRuntimeSecs;
 
   void init(Frame levelOneTrainingFrame,
             Frame levelOneValidationFrame,
@@ -49,7 +49,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
             StackedEnsembleParameters parms,
             boolean hasMetalearnerParams,
             long metalearnerSeed,
-            long remainingTime) {
+            long maxRuntimeSecs) {
 
     _levelOneTrainingFrame = levelOneTrainingFrame;
     _levelOneValidationFrame = levelOneValidationFrame;
@@ -61,7 +61,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
     _parms = parms;
     _hasMetalearnerParams = hasMetalearnerParams;
     _metalearnerSeed = metalearnerSeed;
-    _remainingTime = remainingTime;
+    _maxRuntimeSecs = maxRuntimeSecs;
   }
 
   void compute() {
@@ -113,7 +113,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
     parms._train = _levelOneTrainingFrame._key;
     parms._valid = (_levelOneValidationFrame == null ? null : _levelOneValidationFrame._key);
     parms._response_column = _model.responseColumn;
-    parms._max_runtime_secs = _remainingTime;
+    parms._max_runtime_secs = _maxRuntimeSecs;
   }
 
   protected void setCrossValidationParams(P parms) {
--- a/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
+++ b/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
@@ -111,6 +111,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
     parms._train = _levelOneTrainingFrame._key;
     parms._valid = (_levelOneValidationFrame == null ? null : _levelOneValidationFrame._key);
     parms._response_column = _model.responseColumn;
+    parms._max_runtime_secs = _parms._max_runtime_secs;
   }
 
   protected void setCrossValidationParams(P parms) {
--- a/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
+++ b/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
@@ -37,6 +37,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
   protected P _metalearner_parameters;
   protected boolean _hasMetalearnerParams;
   protected long _metalearnerSeed;
+  protected long _remainingTime;
 
   void init(Frame levelOneTrainingFrame,
             Frame levelOneValidationFrame,
@@ -47,7 +48,8 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
             Job metalearnerJob,
             StackedEnsembleParameters parms,
             boolean hasMetalearnerParams,
-            long metalearnerSeed) {
+            long metalearnerSeed,
+            long remainingTime) {
 
     _levelOneTrainingFrame = levelOneTrainingFrame;
     _levelOneValidationFrame = levelOneValidationFrame;
@@ -59,7 +61,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
     _parms = parms;
     _hasMetalearnerParams = hasMetalearnerParams;
     _metalearnerSeed = metalearnerSeed;
-
+    _remainingTime = remainingTime;
   }
 
   void compute() {
@@ -111,7 +113,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
     parms._train = _levelOneTrainingFrame._key;
     parms._valid = (_levelOneValidationFrame == null ? null : _levelOneValidationFrame._key);
     parms._response_column = _model.responseColumn;
-    parms._max_runtime_secs = _parms._max_runtime_secs;
+    parms._max_runtime_secs = _remainingTime;
   }
 
   protected void setCrossValidationParams(P parms) {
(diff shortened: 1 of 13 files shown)


h2o-py/h2o/estimators/stackedensemble.py
```diff
@@ -550,6 +550,21 @@ def metalearner_params(self, metalearner_params):
             self._parms["metalearner_params"] = None
 
 
+    @property
+    def max_runtime_secs(self):
+        """
+        Maximum allowed runtime in seconds for model training. Use 0 to disable.
```
ledell:
```suggestion
        Maximum allowed runtime in seconds for metalearner model training. Use 0 to disable.
```

h2o-r/h2o-package/R/stackedensemble.R
```diff
@@ -34,6 +34,7 @@
 #'        classification problems. Must be one of: "AUTO", "Random", "Modulo", "Stratified".
 #' @param metalearner_fold_column Column with cross-validation fold index assignment per observation for cross-validation of the metalearner.
 #' @param metalearner_params Parameters for metalearner algorithm
+#' @param max_runtime_secs Maximum allowed runtime in seconds for model training. Use 0 to disable. Defaults to 0.
```
ledell:
```suggestion
#' @param max_runtime_secs Maximum allowed runtime in seconds for metalearner model training. Use 0 to disable the time limit. Defaults to 0.
```

h2o-docs/src/product/data-science/stacked-ensembles.rst
```diff
@@ -97,6 +97,9 @@ Defining a Stacked Ensemble Model
 
 -  **keep_levelone_frame**: (Optional) Keep the level one data frame that's constructed for the metalearning step. Defaults to False.
 
+-  `max_runtime_secs <algo-params/max_runtime_secs.html>`__: (Optional) Maximum allowed runtime in seconds for the metalearner model
+   training. Use 0 to disable.
```
ledell:
```suggestion
   training. Use 0 to disable the time limit..
```

tomasfryda:
@erin I am afraid that the changes you made for R and Python will disappear after a build (during the bindings generation). I will try to modify the `StackedEnsembleV99` schema to change the generated documentation (which will also change the documentation for Flow).

tomasfryda:
The `max_runtime_secs` is inherited from `ModelParametersSchemaV3` and it appears that I am unable to easily override a field in the schema with new doc string - basically I end up with 2 `max_runtime_secs` during the build and the build fails...

@sebhrusen do you have some idea how to approach this? I can think of only 2 ways both seem to me less than ideal: (1) somehow modify the API annotation of the `max_runtime_secs` property during the `StackedEnsembleParametersV99` construction, (2) modify `h2o-core/src/main/java/water/api/SchemaMetadata.java` to handle overriding the properties better.

For now I have modified the `gen_stackedensemble.py` files to provide the same docstring as @erin suggested, except for: in python the "Defaults to 0." is redundant as it is written on the next (non-blank) line and the "0." got wrapped on the next line if I would kept it. Maybe this would be enough but I don't like to have the same thing in multiple places...

tomasfryda:
Since this has been open for a long time and I grown a bit more comfortable with SE implementation and as @sebhrusen mentioned passing `max_runtime_secs` to metalearner would  not work as expected (e.g., with huge blending frames), I decided to try to make SE properly respect `max_runtime_secs`.

The change works like this: (if we specify `max_runtime_secs`)
1. start the timer (implicitly; done by ModelBuilder)
2. before evaluation of each base model check if we have time (or if a user cancelled the computation)
3. if we ran out of time we will build a metalearner only with what we computed so far
4. pass `max(1, remainingTime())` as `max_runtime_secs` to metalearner (if we ran out of time we don't want to pass 0 as it would mean unlimited time)

At first I tried to stop computation altogether but that yielded null ptr exception when running from R and it seems to me that other algos try to build at least some model rather than stop without any model built.

h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_max_runtime_secs.py
```diff
@@ -0,0 +1,88 @@
...
+
+    se = H2OStackedEnsembleEstimator(
+        base_models=gs1.model_ids,
+        max_runtime_secs=max_runtime_secs,
+        blending_frame=big_blending_frame)
+    se.train(data.x, data.y, data.train)
+
+    assert se.metalearner() is not None
+
+    # We should have SE with just one base model due to time constrains; intercept + base_model ==> 2
```
ledell, line 81: @tomasfryda I'm confused by this comment -- how does the metalearner time constraint impact the number of base models included in the Stacked Ensemble?

tomasfryda: The behaviour I implemented is as follows:

When we have a big blending frame, we  need to get predictions on that. If we run out of time during the predicting, we skip the rest of the base models. In the test, the situation is really just one extreme. 

I also considered another approach - if we ran out of time just fail. But from what I can tell about other algos they tend not to fail, rather they just "don't converge", e.g., DRF builds just the amount of trees it had time to etc.

I am open to suggestions how to handle this differently.

tomasfryda:
@ledell I am quite certain that I didn't manually edit those files. The changes in git are done by the automatic code generation. This line in `h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java` is responsible for those changes https://github.com/h2oai/h2o-3/pull/4444/files#diff-081f316cfe5acde20a43c9772f91fad8R46 .

Anyway, thank you for checking stuff like that as well!

h2o-algos/src/main/java/hex/ensemble/Metalearner.java
```diff
@@ -47,7 +48,8 @@ void init(Frame levelOneTrainingFrame,
             Job metalearnerJob,
             StackedEnsembleParameters parms,
             boolean hasMetalearnerParams,
-            long metalearnerSeed) {
+            long metalearnerSeed,
+            long remainingTime) {
```
sebhrusen, line 52: just to bother you:
classes should look like they are self-contained (even when it's not the case), but from a pure semantic perspective `remainingTime` systematically raises the question "remaining of what?".
The answer is simple: "remaining time from the caller perspective".
So, when naming things, we should try to forget (temporarily) the specific conditions in which this thing is being used: what is "remaining time" from SE point of view is just "max runtime" from metalearner perspective.

To your defense: almost all parameters in this class are badly named... 3 different case conventions, redundant use of word `metalearner` in a class named `Metalearner`... even the name `levelOne`... what if I want to do multi-layer stacking?

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -241,6 +242,18 @@ private Frame prepareLevelOneFrame(String levelOneKey, Key<Model>[] baseModelKey
       return levelOneFrame;
     }
 
+    private boolean shouldStop() {
+      if (stop_requested()) {
+        if (timeout()) {
+          Log.info("Stopping StackedEnsemble training because of timeout");
+          return true;
```
sebhrusen, line 249: why just breaking the loop in case of timeout, will it not keep the training with only a subset of base models?

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -62,6 +62,13 @@ protected boolean stop_requested() {
     return _job.stop_requested() || timeout();
   }
 
+  protected long remainingTimeSecs() {
+    long remainingTimeMillis = _build_model_countdown.remainingTime();
+    if (remainingTimeMillis == Long.MAX_VALUE)
+      return 0;
```
sebhrusen, line 68: not sure this is worth converting it back to 0, we're talking about a real lot of millis here... a few million years.
And when consuming this, you'll have to handle this 0 as a special case again.

h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_max_runtime_secs.py
sebhrusen: Gotcha, I understand the `shouldStop` in `StackedEnsemble` now.
I'm in favor of failing in that case for a simple reason: other algos do not fail if the training was started. As you say, they just "don't converge".
But in your case here, the metalearner training didn't even start, so you'll start a metalearner training with only a random subset of the models, and  besides, as it's timing out, you'll train it for a couple of seconds max: no point, really, better skip it completely.

h2o-algos/src/main/java/hex/ensemble/Metalearner.java
tomasfryda: No bother at all. I really appreciate these kind of remarks. I renamed it to `maxRuntimeSecs`  that should be clearer, I hope.

h2o-core/src/main/java/hex/ModelBuilder.java
tomasfryda: Fixed, thanks! I misread the behaviour of `_build_model_countdown.remainingTime()` in a way where I assume that if there is 0 or less remaining time, it would return `Long.MAX_VALUE`...

h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_max_runtime_secs.py
tomasfryda: Ok, I modified it back to just fail if there is not enough time. In R, I am throwing an error when I can't retrieve the metalearner saying it might be due to time out. In python, I don't do anything. The reason is that R uses functional interface and returns an object - something that we didn't build (as opposed to python's object oriented client interface).

I don't like that the two implementations behave differently. Should I also raise an exception from python? It didn't seem to me like something that other algos do.

sebhrusen: I see your concern now: other algos never (never?) fail on timeout, and starting the metalearner, even for 1s, avoided that as well.
I think raising an error on both R and Py clients make sense if metalearner is not available, you could even precise in your message that there was not enough time to start training the metalearner to inform user why this is failing instead of just "timing out".

PR #4445, opened 2020-03-24
PUBDEV-6976: Support parallel training (Segment Models) in R

michalkurka:


Pscheidl:
Re-ran the tests, R.35 CMD check was interrupted for some reason.

h2o-r/h2o-package/R/generic.R
```diff
@@ -42,3 +42,47 @@ h2o.generic <- function(model_id = NULL,
...
+#'
+#' Trains generic model for each segment of the training dataset.
+#'
+#' @param model_key Key to the self-contained model archive already uploaded to H2O.
+#' @param path Path to file with self-contained model archive.
+#' @param segment_columns A list of columns to segment-by. H2O will group the training (and validation) dataset by the segment-by columns
+#'        and train a separate model for each segment (group of rows).
+#' @param segment_models_id Identifier for the returned collection of Segment Models. If not specified it will be automatically generated.
+#' @export
+h2o.bulk_generic <- function(model_key = NULL,
```
Pscheidl: Generic models are not trained per se. Those are imported/wrapped models which are already **pre-trained**. This doesn't make sense to me.

Pscheidl: It also doesn't make sense to import MOJOs as a bulk (maybe for user convenience), because each of those models is a standalone model.

michalkurka: you are exactly right, this should be excluded from the API

michalkurka:
> Re-ran the tests, R.35 CMD check was interrupted for some reason.

This happens on master as well, it seems like addition of the examples caused this

PR #4446, opened 2020-03-25
[HIVE] move delegation token code to hive module

honzasterba:


h2o-hadoop-common/src/main/java/water/hadoop/common/HadoopUtils.java
```diff
@@ -0,0 +1,78 @@
+package water.hadoop.common;
```
michalkurka: I don't see anything Hadoop specific, can we move it to the h2o-core instead?

h2o-hive/src/main/java/water/hive/DelegationTokenRefresher.java
```diff
@@ -39,10 +44,10 @@ public static void setup(Configuration conf, String iceRoot) throws IOException
   }
   
   private static String writeKeytabToFile(String authKeytab, String iceRoot) throws IOException {
-    h2omapper.makeSureIceRootExists(iceRoot);
```
michalkurka: Is this correct in Sparkling Water's context? Or maybe - is this correct naming in a general context?

PR #4449, opened 2020-03-25
[PUBDEV-7385] use java based XGBoost scoring

honzasterba:


honzasterba:
benchmarks look good, no impact, @michalkurka @Pscheidl please review

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/predict/XGBoostNativeBigScoreChunkPredict.java
```diff
@@ -0,0 +1,128 @@
...
+        default:
+          throw new UnsupportedOperationException("Unsupported output type: " + outputType);
+      }
+      return preds == null ? new float[0][] : preds;
+    } catch (XGBoostError xgBoostError) {
+      throw new IllegalStateException("Failed to score with XGBoost.", xgBoostError);
+    } finally {
+      BoosterHelper.dispose(booster, data);
+      try {
+        Rabit.shutdown();
```
Pscheidl: Sounds like a too much of a paranoia, but what if BoosterHelper fails ? The rabit shutdown could still be a handful thing to do - therefore I'd put it into a finally block as well. Inception really, I know.

honzasterba: if the disposing of a booster fails something is very wrong and the whole cluster is probably unhealthy, I dont think we need to worry about shutting down rabit

PR #4451, opened 2020-03-25
PUBDEV-7405: Add "single_node_mode" for building CoxPH model

michalkurka:


h2o-algos/src/main/java/hex/coxph/CoxPHModel.java
```diff
@@ -51,6 +51,8 @@
 
     public boolean _calc_cumhaz = true; // support survfit
 
+    public boolean _single_node_mode = false;
```
honzasterba: seem unused

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
```diff
@@ -1,23 +1,21 @@
...
+import java.util.HashSet;
+import java.util.Set;
 
 import static org.junit.Assert.*;
 
 public class CoxPHTest extends TestUtil {
 
   @BeforeClass
-  public static void setup() { stall_till_cloudsize(1); }
+  public static void setup() { stall_till_cloudsize(2); }
```
honzasterba: intentional?

h2o-algos/src/main/java/hex/coxph/CoxPHModel.java
michalkurka: it is used all over the place :)

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
michalkurka: no! thanks - this is also causing the tests to fail

h2o-algos/src/main/java/hex/coxph/CoxPHModel.java
honzasterba: on sorry, i thought this was declared on model class, I was only looking the file name and did not realize this is on params

PR #4454, opened 2020-03-26
PUBDEV-7126: Make Date formatters ThreadLocal

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7126

I was also considering to use [DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html), which was introduced in Java 8 and is thread-safe, however, it would require quite a lot of changes (e.g., changing some variables from `java.util.Date` to `java.time.LocalDateTime`) or a lot of in-place conversions. The change would make sense to me if we would want to deal with time zones, which I don't think we want/need to do in this case but I might be wrong...


sebhrusen:
Please wait for creation of new release branch to change the base branch and merge.

PR #4457, opened 2020-03-26
Fix a race condition in SegmentModelsBuilder

michalkurka:
In SegmentModelBuilder each H2O node is generating a unique model ID based on
state of the local node (not a global state), this leads to collisions
because the model ids might not necessarily be globally unique.

The attached screenshot is showing an example of such model-key collisions.


michalkurka:
Screen Shot 2020-03-26 at 2 35 30 PM

PR #4460, opened 2020-03-27
PUBDEV-7413: Added parallel model building to Segment Models

michalkurka:


h2o-core/src/main/java/hex/segments/LocalSequentialSegmentModelsBuilder.java
```diff
@@ -54,8 +55,29 @@ void buildModels(SegmentModels segmentModels) {
...
+        if (result.isSuccessful())
+          stats._succeeded++;
+        else
+          stats._failed++;
       }
     }
+    return stats;
+  }
+
+  static class SegmentModelsStats extends Iced<SegmentModelsStats> {
```
Pscheidl: Could easily be normal class, not an inner static class. It's not only used in the `LocalSequentialSegmentModelsBuilder.java` builder, but also in `SegmentModelsBuilder` c lass.

h2o-core/src/main/java/hex/segments/LocalSequentialSegmentModelsBuilder.java
```diff
@@ -54,8 +55,29 @@ void buildModels(SegmentModels segmentModels) {
...
+    int _succeeded;
```
Pscheidl: As it's only local, we don't need to take care about atomicity of operations on top of these.

h2o-core/src/main/java/hex/segments/SegmentModelsBuilder.java
```diff
@@ -116,24 +127,51 @@ private Frame reorderColumns(Key<Frame> key) {
...
+        ).limit(_parallelism).collect(Collectors.toList());
+        try {
+          ExecutorService executor = Executors.newFixedThreadPool(_parallelism);
+          _stats = new SegmentModelsStats();
+          for (Future<SegmentModelsStats> f : executor.invokeAll(runnables)) {
+            _stats.reduce(f.get());
+          }
+        } catch (ExecutionException | InterruptedException e) {
+          throw new RuntimeException("Failed to build segment-models", e);
+        }
```
Pscheidl: This can be written as a stream in it's whole beauty. Optional of course. 

```suggestion
        ExecutorService executor = Executors.newFixedThreadPool(_parallelism);
        _stats = Stream.<Callable<SegmentModelsStats>>generate(
                () -> (() -> _builder.clone().buildModels(_segment_models)))
                .limit(_parallelism)
                .map(callable -> executor.submit(callable))
                .map(future -> {
                  try {
                    return future.get();
                  } catch (ExecutionException | InterruptedException e) {
                    throw new RuntimeException("Failed to build segment-models", e);
                  }
                })
                .reduce((a, b) -> {
                  a.reduce(b);
                  return a;
                }).get();
```

Pscheidl:
The WIP status LGTM !

h2o-core/src/main/java/hex/segments/LocalSequentialSegmentModelsBuilder.java
michalkurka: good point, will do!

h2o-core/src/main/java/hex/segments/SegmentModelsBuilder.java
michalkurka: nice trick with using submit instead of invokeAll to get rid of the outer InterruptedException

thank you @Pscheidl

PR #4463, opened 2020-03-30
Local smalldata path prefix

Pscheidl:
The original driver of this change was the inconvenience of JUnit tests in multinode. When H2O nodes are started via `H2OTestNodeStarter`, they're unable to find and parse the imported files, as the path is relative to H2O-3 working directory. One option was to enable downloading them from the S3 via `H2O_JUNIT_ALLOW_NO_SMALLDATA=TRUE` environment variable, but this might introduce unnecessary delays while downloading the data. Also, it is unnecessary to keep downloading the data over and over again. This variable is there for a different purpose.

A solution to define an environment variable with absolute path to the directory containing the test files (usually smalldata, but can also be bigdata and bigdata-laptop) arised. As it turned out (https://github.com/h2oai/h2o-3/pull/4463#issuecomment-606080773), such a solution can be generalized and whole H2O can be instructed via environment variable (do we want H2O arg as well ?) to search for the files in such a pre-defined location.

Screenshot from 2020-03-31 09-37-43

Makes running our JUnit tests in multinode locally easier. Without the need to modify the code.

### DIFF
--- a/h2o-core/src/main/java/water/util/FileUtils.java
+++ b/h2o-core/src/main/java/water/util/FileUtils.java
@@ -132,12 +132,12 @@ public class FileUtils {
   private static Optional<File> findFileInPredefinedPath(final String fileName) {
     Objects.requireNonNull(fileName);
 
-    final String smallDataPathPrefix = System.getenv("H2O_FILES_SEARCH_PATH");
-    if (smallDataPathPrefix == null) return Optional.empty();
+    final String searchPath = System.getenv("H2O_FILES_SEARCH_PATH");
+    if (searchPath == null) return Optional.empty();
 
-    final StringBuilder localizedFileNameBuilder = new StringBuilder(smallDataPathPrefix);
+    final StringBuilder localizedFileNameBuilder = new StringBuilder(searchPath);
 
-    if (!smallDataPathPrefix.endsWith("/")) {
+    if (!searchPath.endsWith("/")) {
       localizedFileNameBuilder.append('/');
     }
     
--- a/h2o-core/src/main/java/water/util/FileUtils.java
+++ b/h2o-core/src/main/java/water/util/FileUtils.java
@@ -9,6 +9,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URI;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * File utilities.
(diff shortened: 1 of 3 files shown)


h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -450,7 +452,35 @@ public static Frame parse_test_file( Key outputKey, String fname, ParseSetupTran
...
+    final String smallDataPathPrefix = System.getenv(TEST_DATA_PATH_PREFIX_KEY);
+    if (smallDataPathPrefix == null) return fileName;
+
+    final StringBuilder localizedFileNameBuilder = new StringBuilder(smallDataPathPrefix);
+
+    if (!smallDataPathPrefix.endsWith("/")) {
+      localizedFileNameBuilder.append('/');
+    }
+
+    if (fileName.startsWith("./")) {
```
Pscheidl: Must work with both

```java
            parse_test_file("./smalldata/testng/airlines_train.csv");
```

and

```
            parse_test_file("smalldata/testng/airlines_train.csv");

File: h2o-core/src/test/java/water/TestUtil.java [hunk 2]
```diff
@@ -450,7 +452,35 @@ public static Frame parse_test_file( Key outputKey, String fname, ParseSetupTran
...
+   */
+  private static String localTestDataPath(final String fileName) {
+    Objects.requireNonNull(fileName);
+
+    final String smallDataPathPrefix = System.getenv(TEST_DATA_PATH_PREFIX_KEY);
+    if (smallDataPathPrefix == null) return fileName;
+
+    final StringBuilder localizedFileNameBuilder = new StringBuilder(smallDataPathPrefix);
+
+    if (!smallDataPathPrefix.endsWith("/")) {
```
Pscheidl: Must work with both

```
H2O_TEST_DATA_PATH_PREFIX=/home/pavel/Work/h2o-3/
```

and

```
H2O_TEST_DATA_PATH_PREFIX=/home/pavel/Work/h2o-3
```

h2o-core/src/test/java/water/TestUtil.java
```diff
@@ -450,7 +452,35 @@ public static Frame parse_test_file( Key outputKey, String fname, ParseSetupTran
     return ParseDataset.parse(outputKey, new Key[]{nfs._key}, true, guessedSetup);
   }
 
-  protected Frame parse_test_file( String fname, String na_string, int check_header, byte[] column_types) {
+  private static final String TEST_DATA_PATH_PREFIX_KEY = "H2O_TEST_DATA_PATH_PREFIX";
```
honzasterba: since the constant is used only once I think its not necessary
if you want to keep it should be named SMALLDATA_PATH_PREFIX because thats what it does

h2o-core/src/test/java/water/TestUtil.java
honzasterba: this will work only if the property is set, is that intentional?

Pscheidl: Yes, no modification done unless it's not set.

h2o-core/src/test/java/water/TestUtil.java
Pscheidl: It's not just smalldata. If you sync bigdata or bigdata-laptop, it will work as well. 

But I agree and will remove the constant. Totally not worth it.

Pscheidl:
> you know you can change the working directory when launching H2O from idea and save that configuration and not worrry about it later

Honestly, no ! Tried to change it from `MODULE_WORKING_DIR` to project dir and it works of course.

It's good enough and makes my change look a little bit ... ehm ... not really needed ? But still, this is generic solution. It's also quite convenient. I export it **once**.

Therefore, I guess it's still a good change.

michalkurka:
@Pscheidl this is a good idea, WDYT about making this work on H2O-level instead?

you could introduce a variable (eg. "H2O_FILES_SEARCH_PATH") and use it in https://github.com/h2oai/h2o-3/blob/master/h2o-core/src/main/java/water/util/FileUtils.java#L99

this would make it work not just for junit tests but universally in H2O

Pscheidl:
What @michalkurka says is great and I'll make it so.

MNeE7

Expect an update tomorrow.

Pscheidl:
Updated the description. Did as @michalkurka said.

h2o-core/src/main/java/water/util/FileUtils.java
```diff
@@ -97,33 +101,74 @@ else if (file.isDirectory()) {
...
-  private static void check(boolean cond, String msg) throws IOException{
+  /**
+   * @param fileName File name/path to search for in pre-defined search path
+   * @return An {@link Optional} with the file inside, if the H2O_FILES_SEARCH_PATH is defined and the file exists.
+   * Otherwise an empty {@link Optional}. Never null.
+   */
+  private static Optional<File> findFileInPredefinedPath(final String fileName) {
+    Objects.requireNonNull(fileName);
+
+    final String smallDataPathPrefix = System.getenv("H2O_FILES_SEARCH_PATH");
```
honzasterba: confusing variable name

koniecsveta: maybe this can be renamed now to dataPathPrefix ?

h2o-core/src/test/java/water/util/FileUtilsTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public void testFindFileInPredefinedPath() throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+        final Method findFileInPredefinedPath = FileUtils.class.getDeclaredMethod("findFileInPredefinedPath",
+                String.class);
+        assertTrue(Modifier.isStatic(findFileInPredefinedPath.getModifiers()));
+        assertTrue(Modifier.isPrivate(findFileInPredefinedPath.getModifiers()));
+        try {
+            final File h2oHomeDir = new File(System.getProperty("user.dir")).getParentFile();
+            environmentVariables.set("H2O_FILES_SEARCH_PATH", h2oHomeDir.getAbsolutePath());
+
+            findFileInPredefinedPath.setAccessible(true);
```
honzasterba: you could leave the method as package private and have easier time testing it

h2o-core/src/test/java/water/util/FileUtilsTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+@RunWith(H2ORunner.class)
+@CloudSize(1)
+public class FileUtilsTest {
+
+    @Rule
+    public EnvironmentVariables environmentVariables = new EnvironmentVariables();
+
+    @Test
+    public void testFindFileInPredefinedPath() throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
```
honzasterba: the method has quite a few branches and a regex I would like to see some more unit tests

h2o-core/src/main/java/water/util/FileUtils.java
```diff
@@ -97,33 +101,74 @@ else if (file.isDirectory()) {
...
+    if (smallDataPathPrefix == null) return Optional.empty();
+
+    final StringBuilder localizedFileNameBuilder = new StringBuilder(smallDataPathPrefix);
+
+    if (!smallDataPathPrefix.endsWith("/")) {
+      localizedFileNameBuilder.append('/');
+    }
+    
+    // If the file starts with {"./", ".\"} (or multiple instances of these), strip it.
```
honzasterba: do you really need to strip these? I think that /path/./with/./dots/ is the equivalent to one withtout the /./ parts

h2o-core/src/test/java/water/util/FileUtilsTest.java
Pscheidl: I could, but this is easy enough. The method should stay private. Bending code because of JUnit test is only replacing one smell with another. The code that's actually used has proper visibility and the test is adjusted. I'd much rather write a slightly more complex test than exposing things with wrong visibility in production code.

h2o-core/src/main/java/water/util/FileUtils.java
Pscheidl: The true point is to strip `../` or windows `..\` and chains of these from the very beginning. Will pushed improved example. And not `/` only - as this indicates the path is absolute. The fact that it's able to make the resulting path and thus Frame keys a little bit more readable is a bonus. :grin:

Pscheidl: I'll also add tests for that. Good point.

h2o-core/src/main/java/water/util/FileUtils.java
Pscheidl: Yes, this is seriously wrong. Thanks. Copy & paste error.

Pscheidl: Copy & paste error.

h2o-core/src/test/java/water/util/FileUtilsTest.java
Pscheidl: Added. Please have a look now.

Pscheidl: Testing both positive (file found) and negative scenarios now. Both with "/" at the end of the directory-of-search path and without it.

PR #4468, opened 2020-03-31
PUBDEV-7350: Doc update for fractionalbinomial family in GLM

angela0xdata:
GLM now supports a fractionalbinomial family.
- Added this to list of families in the GLM chapter and in the "family" parameter appendix
- Updated the link/family table in the GLM chapter and in the "link" parameter appendix
- Added a "fractional logit regression" section to the user guide based on documented provided in jira ticket
- Updated booklet with these changes.
- Added Python and R examples to the booklet.


angela0xdata:
Made a small fix to the booklet examples after testing with latest nightly build. R and Python examples now both run correctly.

PR #4477, opened 2020-04-01
PUBDEV-7422: Segmented model training interface in R

ledell:
Work in progress: https://0xdata.atlassian.net/browse/PUBDEV-7422

### TO DO

- [x]  Forked `h2o.grid()` to create `h2o.segment_train()`
- [x] R example shows use
- [x]  Manually validated the output of the function against the existing `h2o.bulk_gbm()` function.
- [ ] Currently a lot of grid-related code is commented out; remove that.
- [ ] Remove all the `h2o.bulk_algorithm()` functions in R and all the Python scripts/code that generated those automatically.
- [ ] Look at User Guide to see if we need to add or change anything.
- [ ] Tests
- [ ] If we go with this proposal, I would also suggest that we use the same function/method name in Python (to change in a separate PR).  Currently it's `.bulk_train()` but "bulk" seems too vague and not specific to the fact that we are training models on segments of the training set.  The model type is also called `"H2OSegmentModel"` so it would be great to just use the word "segment".

### OPEN QUESTIONS
- What do you do with an object of type `"H2OSegmentModel"` once you have it?  It looks like this if you print it:  
```
> models
An object of class "H2OSegmentModels"
Slot "segment_models_id":
[1] "GBM_segment_models_R_1585732049703_4"
```
I am not sure how to grab the models inside of here by model ID (so you can look at hyperparams, performance, etc).  

- None of the tests seem to work for me (you can't apply `as.data.frame()` to an object of type H2OSegment model, so I don't know how these tests can work...) https://github.com/h2oai/h2o-3/blob/master/h2o-r/tests/testdir_algos/deeplearning/runit_deeplearning_bulk.R

### RELATED
- Created another [JIRA](https://0xdata.atlassian.net/browse/PUBDEV-7423) to add a proper print method for `"H2OSegmentModels"` in R:

### DIFF
--- a/h2o-r/h2o-DESCRIPTION.template
+++ b/h2o-r/h2o-DESCRIPTION.template
@@ -99,5 +99,6 @@ Collate:
          'w2vutils.R'
          'locate.R'
          'grid.R'
+         'segment.R'
          'predict.R'
          'zzz.R'
--- a/h2o-r/h2o-package/R/segment.R
+++ b/h2o-r/h2o-package/R/segment.R
@@ -1,0 +1,185 @@
+#'
+#' H2O Segmented-Data Bulk Model Training
+#'
+#' Provides a set of functions to train a group of models on different
+#' segments (subpopulations) of the training set. 
+
+#--------------------------------------------
+# Segmented-data bulk model training function
+#--------------------------------------------
+
+#'
+#' Launch grid search with given algorithm and parameters.
+#'
+#' @param algorithm  Name of algorithm to use in training segment models (gbm, randomForest, kmeans, glm, deeplearning, naivebayes, psvm,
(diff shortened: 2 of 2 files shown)


michalkurka:
@ledell thank you for the proposal, it is definitely better than the original solution

To your open question:
- The returned object is currently just a handle to the collection of models. This handle can be converted to a `data.frame` but in future updates will show some summary information about the models built. This API is rudimental in this respect.

I will also rename the functions in Python to follow the "segment_train" pattern.

I will continue the work in https://github.com/h2oai/h2o-3/pull/4480

ledell:
Closing this PR, as it was finished over here: https://github.com/h2oai/h2o-3/pull/4480

PR #4478, opened 2020-04-01
PUBDEV-7089: Fix naming convention in gradle settings

koniecsveta:


Pscheidl:
Wait, is my `h2o-genmodel-ext-deepwater` wrong @michalkurka @koniecsveta ? What is the current convention we use ?

koniecsveta:
> Wait, is my `h2o-genmodel-ext-deepwater` wrong @michalkurka @koniecsveta ? What is the current convention we use ?

yours is good, making mine to look like yours

Pscheidl:
I overlooked we're modifying the name inside the gradle script. Yes, it's allright. It's aligned now.

PR #4480, opened 2020-04-01
PUBDEV-7422: Adapt Segment Models to h2o.segment_train interface in R

michalkurka:
Continuation of Erin's work started in https://github.com/h2oai/h2o-3/pull/4477


michalkurka:
@ledell I tried to preserve the advantages of both approaches

1. singular API entry point provided by your proposal
2. autogenerated _private_ functions  strictly copying the semantics of the regular "train" function

(1) gives us a much better way to document the feature and will be better understood to the users, (2) is an implementation technicality - but it lowers the amount of maintenance we need to do on the "segment_train" function, all custom validations/extensions of the API are automatically reflected in the code without any manual changes due to the auto-generation and sharing code with normal "train"

please let me know if you consider it a good way to go

h2o-r/h2o-package/R/kvstore.R
```diff
@@ -287,10 +287,46 @@ h2o.getModel <- function(model_id) {
...
+#' h2o.init()
+#' iris_hf <- as.h2o(iris)
+#' h2o.segment_train(algorithm = "gbm",
+#'                   segment_columns = "Species", segment_models_id="models_by_species",
+#'                   x = c(1:3), y = 4, training_frame = iris_hf, ntrees = 5, max_depth = 4)
+#' models <- h2o.getSegmentModels("models_by_species")
+#' as.data.frame(models)
+#' }
+#' @export
 h2o.getSegmentModels <- function(segment_models_id) {
```
ledell, line 307:
```suggestion
h2o.get_segment_models <- function(segment_models_id) {
```

ledell: In the last year or more, we have dropped the camel case for `h2o.get*` functions in R and switched to snake case for all new functions (we are aspiring to modern R coding conventions), so I would suggest we do that here too.  At some point it might just make sense to make aliases for the remainder of the functions that still use camel case like `h2o.getModel()`, `h2o.getFrame()`, etc. but that's a discussion for another day!  We have at least one alias, `h2o.getAutoML()` (created three years ago) and now we also have `h2o.get_automl()`.

h2o-r/h2o-package/R/kvstore.R
```diff
@@ -287,10 +287,46 @@ h2o.getModel <- function(model_id) {
...
+#' @return Returns an object that is a subclass of \linkS4class{H2OSegmentModels}.
+#' @examples
+#' \dontrun{
+#' library(h2o)
+#' h2o.init()
+#' iris_hf <- as.h2o(iris)
+#' h2o.segment_train(algorithm = "gbm",
+#'                   segment_columns = "Species", segment_models_id="models_by_species",
+#'                   x = c(1:3), y = 4, training_frame = iris_hf, ntrees = 5, max_depth = 4)
+#' models <- h2o.getSegmentModels("models_by_species")
```
ledell, line 303:
```suggestion
#' models <- h2o.get_segment_models("models_by_species")
```

ledell: See comment below.

h2o-r/tests/testdir_misc/runit_bulk_gbm_happy.R
```diff
@@ -7,10 +7,11 @@ test.bulk_gbm <- function() {
     smtrees <- h2o.importFile(locate("smalldata/gbm_test/smtrees.csv"))
     smtrees$segment <- as.factor(smtrees$C1 < 6)
 
-    segment_models <- h2o.bulk_gbm(x=c("girth", "height"), y="vol", ntrees=3, max_depth=1, seed=42,
-                                   distribution="gaussian", min_rows=2,
-                                   training_frame=smtrees, segment_columns="segment", segment_models_id = "smtrees_sm")
-    
+    segment_models <- h2o.segment_train(x=c("girth", "height"), y="vol", ntrees=3, max_depth=1, seed=42,
```
ledell, line 10:
```suggestion
    segment_models <- h2o.segment_train(x = c("girth", "height"), y = "vol", ntrees = 3, max_depth = 1, seed = 42,
```

h2o-r/h2o-package/R/kvstore.R
michalkurka: sounds good, I was previously following the getModel pattern for consistency

ledell:
@michalkurka I just pushed a small commit to clean up the formatting in the R tests (proper spacing and order of arguments).

michalkurka:
> @michalkurka I just pushed a small commit to clean up the formatting in the R tests (proper spacing and order of arguments).

ouch! I just force pushed my fixes

ledell:
@michalkurka Oops... sorry I should have given you a heads up.  Is the PR ok now?

Other question -- Are you also working on the PR for changing the method name from `.bulk_train` to `.segment_train` in Python?

michalkurka:
I eradicated all use of "bulk" from R, I will do the same in a separate PR for Python. But I would prefer to do one at a time until we resolve this. Python should be easier for API changes, but still a lot of changes / renaming in tests. I want to rather keep those 2 things separate so that we don't miss anything.

michalkurka:
> @michalkurka Oops... sorry I should have given you a heads up. Is the PR ok now?

@ledell yes, I am not planning any more changes unless tests or review indicate a need for a change

ledell:
@michalkurka I was thinking about adding a [print method](https://0xdata.atlassian.net/browse/PUBDEV-7423) for H2OSegmentModels. Could I add it to this PR?  It should go in `segment.R`.

michalkurka:
@ledell please do, any improvements are welcome - optionally you can branch out my branch and make a new PR (in case I need to change things)

ledell:
@michalkurka I'll wait for you to merge this to master and make a follow-up PR for the print method.  Thanks!

PR #4488, opened 2020-04-02
Segment Models: Adopt "segment_train" terminology

michalkurka:
Get rid off all remaining "bulk" references


h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -114,11 +114,11 @@ def train(self, x=None, y=None, training_frame=None, offset_column=None, fold_co
                                  ignored_columns=ignored_columns, model_id=model_id, verbose=verbose)
         self._train(parms, verbose=verbose)
 
-    def bulk_train(self, x=None, y=None, training_frame=None, offset_column=None, fold_column=None,
-                   weights_column=None, validation_frame=None, max_runtime_secs=None, ignored_columns=None,
-                   segments=None, segment_models_id=None, parallelism=1, verbose=False):
+    def segment_train(self, x=None, y=None, training_frame=None, offset_column=None, fold_column=None,
```
sebhrusen, line 117: as you're about to rename this, shouldn't we prefer `train_segments` to `segment_train`?
I don't have strong opinion on this, just think that it's better for discovery: when you're looking for a specialized `train` method, you're more likely to search for sth that starts with `train`

michalkurka: that is a good point, wdyt @ledell?

PR #4489, opened 2020-04-02
Replace dead link

antonioalmeida:
- Referenced [link](http://www.idg.pl/mirrors/CRAN/web/packages/gbm/vignettes/gbm.pdf) on `offset_colum` is dead;
- Quick solution was to replace with h2o's [documentation](http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/algo-params/offset_column.html) on the parameter.


angela0xdata:
@antonioalmeida Thank you for finding this and for your contribution!
 
One suggestion is that this link should point within the current documentation help system rather than outside to avoid any possible version mismatch. For example, in the future if a user is reviewing this in an earlier version of the documentation, this link switches them the latest-stable docs.

Suggested change:
"For more information, refer to the `offset_column <offset_column.html>` documentation."

antonioalmeida:
Got it @angela0xdata 👍

angela0xdata:
@antonioalmeida Looks great. Thank you again!

PR #4491, opened 2020-04-02
PUBDEV-7428: Add documentation for single_node_mode to CoxPH

hannah-tillman:


h2o-py/h2o/backend/server.py
```diff
@@ -105,7 +105,7 @@ def start(jar_path=None, nthreads=-1, enable_assertions=True, max_mem_size=None,
         assert_is_type(log_dir, str, None)
         assert_is_type(log_level, str, None)
         assert_satisfies(log_level, log_level in [None, "TRACE", "DEBUG", "INFO", "WARN", "ERRR", "FATA"])
-        assert_is_type(max_log_file_size, str)
+        assert_is_type(max_log_file_size, str, None)
```
angela0xdata, line 108: What is this change?

h2o-docs/src/product/data-science/algo-params/single_node_mode.rst
```diff
@@ -0,0 +1,42 @@
+``single_node_mode``
+--------------------
+
+- Available in: CoxPH, Deep Learning
+- Hyperparameter: yes
```
angela0xdata, line 5: This is not a hyperparameter.
An option is a hyperparameter only if the schema shows it as "gridable=true", which is not the case with this option:
https://github.com/h2oai/h2o-3/pull/4451/files#diff-bf97f6406bbd14646fa7194138b45efe

h2o-py/h2o/backend/server.py
hannah-tillman: I have no idea--it's not in my github history...

h2o-docs/src/product/data-science/algo-params/single_node_mode.rst
```diff
@@ -0,0 +1,42 @@
...
+
+		library(h2o)
+		h2o.init()
+
+		# Import the heart dataset:
+		heart <- h2o.importFile("http://s3.amazonaws.com/h2o-public-test-data/smalldata/coxph_test/heart.csv")
+
+		# Build and train the model:
+		coxph.model <- h2o.coxph(x="age", event_column="event", start_column="start", stop_column="stop",training_frame=heart, single_node_mode=TRUE)
+
```
angela0xdata, line 26: Add a step to get the predictions:
pred <- h2o.predict(object=coxph.model, newdata=heart)

h2o-docs/src/product/data-science/algo-params/single_node_mode.rst
```diff
@@ -0,0 +1,44 @@
...
+		# Generate predictions <- h2o.predict(object=coxph.model, newdata=heart)
```
angela0xdata, line 27: This entire line is a comment. Separate the command from the comment line.

PR #4496, opened 2020-04-03
PUBDEV-6296: StackedEnsemble optimization: skip predictions for base models with coef 0 when using GLM metalearner

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-6296


ledell:
Thanks @tomasfryda!  

@sebhrusen What do you think of @michalkurka 's suggestion to generalize this to all zero-importance base learners (so it will be optimal for all Stacked Ensembles vs just the default GLM metalearner)?  We could also do this in two separate JIRA tickets...

tomasfryda:
I already started to work on that in this JIRA. I think it is better to do all in this one since the logic is the same... (but if needed I could still split it in multiple JIRAs)

So far my approach is adding a method `isFeatureUsed` to each model so the complexity of finding out if the feature is used is hidden from the end user. This method should be "conservative" - it can give false positives and it shouldn't influence anything but performance, i.e., if it is not implemented for a specific model it will return true for all the features that were present during the training (so if constant columns are ignored it will return false for them).

Or is there some other method/property that I could use to determine if a feature is used?

@sebhrusen  I am now thinking if I should limit this method only for models that we are now using as a meta learner (deeplearning,  drf, gbm, glm,  naivebayes,  xgboost) and create some interface that would have the `isFeatureUsed` method or if I should create the `isFeatureUsed` method in the  `Model` class and potentially implement it for all the models.

The former is easier since I wouldn't have to deal with encodings for categorical variables etc. (I found that during implementing the latter). What do you think?

Then there is also a question if I should ignore just the features with exactly 0 contribution or if I should relax on the exact zero and use some threshold like 0 +/- epsilon, i.e., the features would still have effectively zero contribution to the prediction would be used by the former approach. 
For example in ridge regression we are likely to have some coefficients close to zero or in Naive Bayes if we have a feature that is independent to the target the probabilities would be the same or almost the same (rounding errors) or I don't know if the first layer weights of neural network are likely to be exact zero if the feature is not useful for the prediction...

michalkurka:
@tomasfryda for simplicity you can look at the variable importance table and if the feature is not listed, it means it was not used at all.

I like your idea of `getUsedFeatures`, this could be part of the model output and be useful eg. in AutoDoc

@laurendiperna / @MartinBarus - do you guys somehow mark which features were actually used by the model and which were not?

tomasfryda:
@michalkurka Thanks, I was wondering about that at least for the tree algos and deep learning...

I might be wrong about this but it looks to me that not all models have implemented variable importance in java (e.g., NaiveBayes, GLM). I looked in python's `ModelBase.varimp()` it appears that GLM and GAM calculate theirs variable importance on the client's side. Or is there some other variable importance table?

Also, is it OK to return all the features as useful if we don't calculate variable importances (e.g., `H2ODeepLearningEstimator(variable_importances=False)`)?

h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
```diff
@@ -727,4 +727,38 @@ public void generate(JCodeSB out) {
     return (T) sb.p(mname).p("_Forest_").p(t);
   }
 
+  @Override
+  protected boolean isFeatureUsed(int featureIdx) {
+    // Maybe using variable importance would be enough? {
```
sebhrusen, line 732: I think so

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -166,7 +171,57 @@ protected Frame predictScoreImpl(Frame fr, Frame adaptFrm, String destination_ke
...
+    Model metalearner = _output._metalearner;
+    if (metalearner == null) return true; // training phase
+    if (modelCategory == ModelCategory.Multinomial) {
+      // Multinomial models output multiple columns and a base model
+      // might be useful just for one category...
+      int colsPerModel = metalearner._output.nfeatures() / _parms._base_models.length;
+      for (int i = index * colsPerModel; i < (index + 1) * colsPerModel; i++) {
+        if (metalearner.isFeatureUsed(metalearner._output._names[i]))
+          return true;
+      }
```
sebhrusen, line 189: just wondering if columns order is guaranteed during the whole process, and even so, if it's wise to rely on it.
It rather rely just on model keys and column names...
In case of multinomial, is there no way to get the probabilities column names for a given model?

sebhrusen:
>  if I should create the isFeatureUsed method in the Model class and potentially implement it for all the models.

I like it as it is, you don't have to implement it for all models in this PR though, this could be a follow up ticket. This is out of scope for this as soon as it is used only by SE for now.

> Then there is also a question if I should ignore just the features with exactly 0 contribution or if I should relax on the exact zero and use some threshold like 0 +/- epsilon

the metalearner should be more clever than any predefined threshold to decide which feature is relevant or not, I'm not convinced we should make this decision for it just for a small gain in computation time: @ledell wdyt?

MartinBarus:
@michalkurka I tried using this:

```
outputs = model._model_json["output"]
orig_features = outputs.get("original_names")
```

sometimes that property is not set however

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -166,7 +171,57 @@ protected Frame predictScoreImpl(Frame fr, Frame adaptFrm, String destination_ke
...
   }
 
+  /**
+   * Is the indexth model's prediction used in the metalearner?
+   *
+   * @param index
+   */
+  boolean isUsefulBaseModel(int index) {
+    Model metalearner = _output._metalearner;
+    if (metalearner == null) return true; // training phase
```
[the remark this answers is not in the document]
tomasfryda: Good point, thanks!

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -344,7 +344,11 @@ protected Frame getPredictionsForBaseModel(Model model, Frame actualsFrame, bool
           throw new H2OIllegalArgumentException("Failed to find the xval predictions frame. . .  Looks like keep_cross_validation_predictions wasn't set when building the models, or the frame was deleted.");
 
       } else {
-        fr = buildPredictionsForBaseModel(model, actualsFrame);
+        if (_model.isUsefulBaseModel(model)) {
+          fr = buildPredictionsForBaseModel(model, actualsFrame);
+        } else {
+          fr = _model.getDummyPredictions(actualsFrame);
```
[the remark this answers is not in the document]
tomasfryda: Thanks! I though I had to provide a "dummy" column. Somehow, I thought I had to provide a matrix with a same shape 🤦

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
tomasfryda: Thanks! This helped me realise that I could use column names in multiple other places.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -2808,4 +2808,22 @@ public ModelDescriptor modelDescriptor() {
     };
   }
 
+  /**
+   * Convenience method to find out if featureName is used for prediction, i.e., if it has beta == 0 in GLM,
+   * it is not considered to be used.
+   * This is mainly intended for optimizing prediction speed in StackedEnsemble.
+   * @param featureName
+   */
+  public boolean isFeatureUsed(String featureName) {
```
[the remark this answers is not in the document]
tomasfryda: Good point! I would probably prefer `isFeatureUsedInPredict`. Also, I added `getUsedFeatures` method that basically lists the features and uses the `isFeatureUsed` as a predicate; just in case it would be useful for someone as @michalkurka suggested. So I will* rename it to something like `getFeaturesUsedInPrediction`. Wdyt?

* Probably after easter; there is still some polishing needed in other parts of the code...

h2o-algos/src/test/java/hex/glm/GLMTest.java
```diff
@@ -2184,4 +2184,75 @@ public void testUnseenLevels(){
...
+    isFeatureUsedInPredictHelper(false, true);
+    isFeatureUsedInPredictHelper(true, true);
+  }
+
+  private void isFeatureUsedInPredictHelper(boolean ignoreConstCols, boolean multinomial) {
+    Scope.enter();
+    Vec target = Vec.makeRepSeq(100, 3);
+    if (multinomial) target = target.toCategoricalVec();
+    Vec zeros = Vec.makeCon(0d, 100);
+    Vec ones = Vec.makeCon(1e10d, 100);
```
sebhrusen, line 2202: big ones :)

tomasfryda:
> >Then there is also a question if I should ignore just the features with exactly 0 contribution or if I should relax on the exact zero and use some threshold like 0 +/- epsilon

> the metalearner should be more clever than any predefined threshold in deciding which feature is relevant or not, I'm not convinced we should make this decision for it just for a small gain in computation time. When contribution is null, it's obvious so we can remove it, when it's small, I'd say let's leave it to the metalearner to decide: @ledell wdyt?

@seb, @ledell I meant here extremely small. Why did I mention thresholding - imaginary situation: we are doing some iterative optimisation of loss function with quadratic penalty (Ridge regression) and we have one parameter b that has value b = 1e-165 then if we apply the quadratic penalty we should get 1e-330 but due to IEEE 754 (floating point specification) we will get 0 so the penalty is the same as if b=0. (This is very extreme since we would probably do a gradient base optimisation so the threshold would move closer to 1e-324 since we would by just doubling instead of squaring the number)

But after looking at Deep Learning I found out that the  contributions even of constant (non-informative) features are quite often really big (bigger than 1e-3). So I guess the thresholding I propose would just add complexity without helping much (maybe at all).

h2o-algos/src/test/java/hex/glm/GLMTest.java
tomasfryda: I will fix the naming here :D

ledell:
Thanks @tomasfryda

PR #4497, opened 2020-04-06
[HIVE] move token generator CLI into hive module

honzasterba:


h2o-hive/src/main/java/water/hive/GenerateHiveToken.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    private void validateArgs() {
+        ImpersonationUtils.validateImpersonationArgs(principal, keytabPath, runAsUser, this::error, this::warning);
+        if (hivePrincipal == null) {
+            error("hive principal name is required (use the '-hivePrincipal' option)");
+        }
+        if (hiveHost == null && hiveJdbcUrlPattern == null) {
+            error("delegation token generator requires Hive host to be set (use the '-hiveHost' or '-hiveJdbcUrlPattern' option)");
```
mn-mikke, line 90: nit: `requires Hive host to be set` -> `requires Hive host or JDBC URL pattern to be set`

PR #4501, opened 2020-04-07
PUBDEV-7441 - Target encoder does not unlock output Frame

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7441
Discovered as a part of https://0xdata.atlassian.net/browse/PUBDEV-7020 and should be merged together, therefore this PR targets the `PUBDEV-7020_extended_demo` branch, as this branch provides tests to reproduce this and therefore no effort is duplicated.

## The problem

Please see the following piece of code (https://github.com/h2oai/h2o-3/blob/master/h2o-core/src/main/java/water/fvec/Frame.java#L1832) demonstrating proper usage of locks:

```java
  public Frame toCategoricalCol(int columIdx){
    write_lock();
    replace(columIdx, vec(columIdx).toCategoricalVec()).remove();
    // Update frame in DKV
    update();
    unlock();
    return this;
  }
```

Notice how the locks are used. The `write_lock()` and `unlock()` is applied to the very same Frame object (with same object reference).

The original code in the `TargetEncoder` class looked as follows:

```java
    private void updateDomainGlobally(Frame fr, String teColumnName, String[] domain) {
      Lockable lock = fr.write_lock();
      Vec updatedVec = fr.vec(teColumnName);
      updatedVec.setDomain(domain);
      DKV.put(updatedVec);
      fr.update();
      lock.unlock();
    }
```
Please notice the `lock.unlock();` operation is not done on the original Frame (variable `fr`), but rather the Lockable instance (which is a Frame in fact). Problem is, by calling `Lockable lock = fr.write_lock();`, the `lock` variable does not contain reference to the same object as `fr` does. The `Lockable` is a completely new `Frame` wrapper object ! See the screenshot below:

Screenshot from 2020-04-07 09-16-51

`fr` and `lockable` are different objects. The `unlock` operation called on lockable is only propagated if an `update` operation is called on the original `Frame` after the `unlock` operation has been issued on the lock.

## The solution

The simplest of solutions is to call the `unlock` operation on the original frame, just like in the example posted above. Therefore, instead of `lockable.unlock()`, `fr.unlock();` is called. And that's it. This approach saves the need for one more update afterwards.

### DIFF
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoder.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoder.java
@@ -234,12 +234,12 @@ public class TargetEncoder extends Iced<TargetEncoder>{
     }
     
     private void updateDomainGlobally(Frame fr, String teColumnName, String[] domain) {
-      Lockable lock = fr.write_lock();
+      fr.write_lock();
       Vec updatedVec = fr.vec(teColumnName);
       updatedVec.setDomain(domain);
       DKV.put(updatedVec);
       fr.update();
-      lock.unlock();
+      fr.unlock();
     }
 
     Frame getOutOfFoldData(Frame encodingMap, String foldColumnName, long currentFoldValue)  {


deil87:
@Pscheidl I looked into the source code of `PriorWriteLock` and `Unlock` and it looks like logic depends on the `_key` values only and it should not matter whether java object is the same. But I might be wrong. There is a `_lockers`array in `Lockable` and looks like we just check whether key is in the array to say whether something is locked or not.

Pscheidl:
The key is different to me. See screenshot.

deil87:
@Pscheidl 

> The key is different to me. See screenshot.

Maybe there is some misunderstanding: I'm talking about DKV keys for highlighted rows:
```
fr =     _97a67970686cbc73e77c62875c2e4381
lock = _97a67970686cbc73e77c62875c2e4381
```

> the lock variable does not contain reference to the same object as fr does.

here you are probably talking about different references to java objects

Pscheidl:
I see. Yes, this is due to the nature of the Frame. It's just a thin wrapper.

michalkurka:
@Pscheidl thank you for fixing incorrect API use, good catch!

deil87:
@michalkurka @Pscheidl This way of locking was taken from this discussion https://github.com/h2oai/h2o-3/pull/3480#issuecomment-496975330 .  Unfortunately nobody noticed that it is an incorrect way to do it even though javadoc says that it returns old object. I think it is just because design is counterintuitive. Usually when you lock something you receive a handle.  Anyway I think that we still should write tests that reflects full use cases. This is the only way we are safe from mistakes.

deil87:
@Pscheidl I run build on Jenkins manually one more time because there were some failures ( probably unrelated, second run will give more confidence on this)

deil87:
> I see. Yes, this is due to the nature of the Frame. It's just a thin wrapper.

yes, I know they are. The point is that from my understanding locking is done via array of string keys :
```
private boolean is_locked(Key<Job> job_key) { 
    if( _lockers==null ) return false;
    for( int i=(_lockers.length==1?0:1); i<_lockers.length; i++ ) {
      Key k = _lockers[i];
      if( job_key==k || (job_key != null && k != null && job_key.equals(k)) ) return true;
    }
    return false;
  }
```
and in the `set_write_lock` I see that we only write to `_lockers` array.

The keys on the screenshot were the same. But at the same time we somehow (probably?) depend on java object ( wrapper which is mapped to the key) itself - this part I still don't understand, assuming we rely only on string keys. @michalkurka can you please help me understand this a bit better? @sebhrusen maybe you can help me as well?

PR #4502, opened 2020-04-07
[Investigation] Investigating SHAP values

deil87:
On the screenshot below we can see that values differ but at the same time additivity rule holds. I'm still looking for an explanation of this difference. I believe that additivity rule on its own is not enough  to say that values are correct ( but that is the only check that I see we check in our tests, might be wrong). Maybe it is just biasTerm( if it is analogous to expected value in shap library) is being incorporated ( distributed) over all SHAP values and at the same time we provide what this bias term was. Still biasTerm differs from expected value from shap ( maybe they are different in their purposes and should not be equal). 

image


michalkurka:
@deil87 thank you for investigating

> but that is the only check that I see we check in our tests

There are test that are checking correctness by comparing Naive ("brute force") Shapley implementation to the TreeSHAP implementation. In addition to that we compare the results to native XGBoost.

The difference is in BiasTeam is caused by the fact that H2O includes the null model prediction into the BiasTerm while XGBoost doesn't.

michalkurka:
I think something is not right, possibly after XGBoost/Predictor upgrade. I'm looking into it as well.

deil87:
> I think something is not right, possibly after XGBoost/Predictor upgrade. I'm looking into it as well.

I'm learning more about how we test against Naive approach and XGBoost. Thanks for reference. Don't know what exactly made you suspicious now but from my side I still wondering why shap values are quite different from those of shap library.

michalkurka:
The actual Shapley values seem to be the same in my tests (comparing with XGBoost), there is however a bug in the bias term calculation caused by a change in Predictor: https://github.com/h2oai/xgboost-predictor/blame/master/xgboost-predictor/src/main/java/biz/k11i/xgboost/Predictor.java#L51

michalkurka:
Do you get different results when you add `feature_perturbation="tree_path_dependent"` to explainer?

deil87:
> Do you get different results when you add `feature_perturbation="tree_path_dependent"` to explainer?

I was using KernelExplainer of the shap library as there is no (easy) way to feed h2o model into TreeExplainer ( they have type check).  KernelExplainer does not have such a parameter, don't know what it does by default.

deil87:
@michalkurka 

If there was a bug in computing `biasTerm` how come we did not spot it with following lines in the test:

```
//  we see that BiasTerm is part of contributions
assertEquals("BiasTerm", contributions.names()[contributions.names().length - 1]);
      
// basic sanity check - contributions should sum-up to predictions
Frame predsFromContributions = new CalcContribsTask().doAll(Vec.T_NUM, contributions).outputFrame();
...
assertVecEquals(expectedPreds.vec(2), predsFromContributions.vec(0), 1e-6);
```
Assuming predictions are fine, and sum is fine and given bias term was incorrect => sum of other contributions should be also incorrect/different. That is probably what I noticed during investigation. Or maybe I still missing something.

PR #4515, opened 2020-04-14
PUBDEV-7362: Fixed merge error

wendycwong:
This PR fixes the bugs (I hope) in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7362?filter=-1

I have performed the following:
1. During the conversion of long/BigInteger to int, if the result is negative, the casting will generate strange results.  I checked the results in long/BigInteger before casting it to integer.
2. For categorical merge columns, the use of id_maps is incorrect.  It should be associated with the right frame but was somehow used for the left frame.
3.The major work was to replace the previous bmerge_r with my own. 
4. I added a R unit test that include failed merges from Neema/Kuba and a customer. 
5. Added comments to various part of the code so that the next time I need to fix something, I can understand what is going on much faster.

### DIFF
--- a/h2o-core/src/main/java/water/rapids/BinaryMerge.java
+++ b/h2o-core/src/main/java/water/rapids/BinaryMerge.java
@@ -146,7 +146,7 @@ class BinaryMerge extends DTask<BinaryMerge> {
 
     // get left batches
     _leftKO.initKeyOrder(_leftSB._msb,/*left=*/true);
-    final long leftN = leftSortedOXHeader._numRows;
+    final long leftN = leftSortedOXHeader._numRows; // number of leftframe rows to fetch for leftMSB
     assert leftN >= 1;
 
     // get right batches
@@ -174,12 +174,13 @@ class BinaryMerge extends DTask<BinaryMerge> {
 
     // Find left and right MSB extents in terms of the key boundaries they represent
     // _riteSB._msb==-1 indicates that no right MSB should be looked at
-    final long leftMin = _leftSB.min();  // the first key possible in this bucket
-    final long leftMax = _leftSB.max();  // the last  key possible in this bucket
+    final long leftMin = _leftSB.min();  // the minimum possible key value in this bucket
+    final long leftMax = _leftSB.max();  // the maximum possible key value in this bucket
     // if _riteSB._msb==-1 then the values in riteMin and riteMax here are redundant and not used
-    final long riteMin = _riteSB._msb==-1 ? -1 : _riteSB.min();  // the first key possible in this bucket
-    final long riteMax = _riteSB._msb==-1 ? -1 : _riteSB.max();  // the last  key possible in this bucket
+    final long riteMin = _riteSB._msb==-1 ? -1 : _riteSB.min();  // the minimum possible key value in this bucket
+    final long riteMax = _riteSB._msb==-1 ? -1 : _riteSB.max();  // the maximum possible key value in this bucket
 
+    // _leftFrom and leftTo refers to the row indices to perform merging/search for each MSB value
     _leftFrom =   (_riteSB._msb==-1 || leftMin>=riteMin || (_allLeft && _riteSB._msb==0  )) ? -1    : bsearchLeft(riteMin, /*retLow*/true , leftN);
     long leftTo = (_riteSB._msb==-1 || leftMax<=riteMax || (_allLeft && _riteSB._msb==255)) ? leftN : bsearchLeft(riteMax, /*retLow*/false, leftN);
     // The (_allLeft && rightMSB==0) part is to include those keys in that
@@ -190,7 +191,7 @@ class BinaryMerge extends DTask<BinaryMerge> {
     // lowest right keys, because stitching assumes unique MSB/MSB pairs.
 
     long retSize = leftTo - _leftFrom - 1;   // since leftTo and leftFrom are 1 outside the extremes
-    assert retSize >= 0;
+    assert retSize >= 0; // retSize is number of rows to include in final merged frame
     if (retSize==0) { tryComplete(); return; } // nothing can match, even when allLeft
 
     _retBatchSize = 1048576;   // must set to be the same from RadixOrder.java
@@ -667,7 +668,7 @@ class BinaryMerge extends DTask<BinaryMerge> {
           // Fetch the left rows and mark the contiguous from-ranges each left row should be recycled over
           // TODO: when single node, not needed
           // TODO could loop through batches rather than / and % wastefully
-          long row = _leftKO.at8order(leftLoc);
+          long row = _leftKO.at8order(leftLoc); // global row number of matched row in left frame
           int chkIdx = _leftSB._vec.elem2ChunkIdx(row); //binary search in espc
           int ni = _leftSB._chunkNode[chkIdx]; // node index
           long pnl = perNodeLeftLoc[ni]++;    // pnl = per node location
@@ -678,10 +679,10 @@ class BinaryMerge extends DTask<BinaryMerge> {
         if (prevf == f && prevl == l)
           continue;  // don't re-fetch the same matching rows (cartesian). We'll repeat them locally later.
         prevf = f; prevl = l;
-        for (int r=0; r<l; r++) {
+        for (int r=0; r<l; r++) { // locate the corresponding matching row in right frame
           long loc = f+r-1;  // -1 because these are 0-based where 0 means no-match and 1 refers to the first row
           // TODO: could take / and % outside loop in cases where it doesn't span a batch boundary
-          long row = _riteKO.at8order(loc);
+          long row = _riteKO.at8order(loc); // right frame global row number that matches left frame
           // find the owning node for the row, using local operations here
           int chkIdx = _riteSB._vec.elem2ChunkIdx(row); //binary search in espc
           int ni = _riteSB._chunkNode[chkIdx];
@@ -694,7 +695,7 @@ class BinaryMerge extends DTask<BinaryMerge> {
     // TODO assert that perNodeRite and Left are exactly equal to the number
     // expected and allocated.
     Arrays.fill(perNodeLeftLoc ,0); // clear for reuse below
-    Arrays.fill(perNodeRightLoc,0);
+    Arrays.fill(perNodeRightLoc,0); // denotes number of rows fetched 
   }
 
   private void chunksGetRawRemoteRows(final long perNodeLeftRows[][][], final long perNodeRightRows[][][], 
--- a/h2o-core/src/main/java/water/rapids/Merge.java
+++ b/h2o-core/src/main/java/water/rapids/Merge.java
@@ -90,12 +90,11 @@ public class Merge {
     Frame rightFrame = naPresent ? new RemoveNAsTask(riteCols)
             .doAll(riteFrame.types(), riteFrame).outputFrame(riteFrame.names(), riteFrame.domains())
             : riteFrame;
-
-
+    
     // map missing levels to -1 (rather than increasing slots after the end)
     // for now to save a deep branch later
-    for (int i=0; i<id_maps.length; i++) {
-      if (id_maps[i] == null) continue;
+    for (int i=0; i<id_maps.length; i++) { // id_maps is for leftFrame.  
+      if (id_maps[i] == null) continue;    // id_maps -1 represent leftFrame levels not found in riteFrame
       assert id_maps[i].length >= leftFrame.vec(leftCols[i]).max()+1
               :"Left frame cardinality is higher than right frame!  Switch frames and change merge directions to get " +
               "around this restriction.";
@@ -103,7 +102,7 @@ public class Merge {
       int right_max = (int)rightFrame.vec(riteCols[i]).max();
       for (int j=0; j<id_maps[i].length; j++) {
         assert id_maps[i][j] >= 0;
-        if (id_maps[i][j] > right_max) id_maps[i][j] = -1;
+        if (id_maps[i][j] > right_max) id_maps[i][j] = -1;  // map enum levels of left frame to -1 if not found in rite frame
       }
     }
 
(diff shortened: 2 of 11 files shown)


h2o-r/tests/runitUtils/utilsR.R
```diff
@@ -936,12 +936,20 @@ compareFrames <- function(frame1, frame2, prob=0.5, tolerance=1e-6) {
   for (colInd in range(1, ncol(frame1))) {
     temp1=as.numeric(frame1[,colInd])
     temp2=as.numeric(frame2[,colInd])
+    notNumericCols = !(h2o.isnumeric(frame1[,colInd]) && h2o.isnumeric(frame2[,colInd]))
+    if (notNumericCols) {
```
honzasterba, line 940: move the above assignment into else branch

jakubhava:
@wendycwong I did not have time to summarize the information for the bug I found, but quick question. Is this change also fixing it? 

> Doing right marge ( allLeft=False, allRight=True) using radix method results in bad results

h2o-core/src/main/java/water/rapids/BinaryMerge.java
```diff
@@ -209,7 +210,8 @@ public void compute2() {
     // occur, they only happen for leftMSB 0 and 255, and will quickly resolve
     // to no match in the right bucket via bmerge
     t0 = System.nanoTime();
-    bmerge_r(_leftFrom, leftTo, -1, rightN);
+    binaryMergeOneMSB(_leftFrom, leftTo, -1, rightN);
+  //  bmerge_r(_leftFrom, leftTo, -1, rightN);  // search for match row in rite frame to left value
```
honzasterba, line 214: delete commented out code

h2o-core/src/main/java/water/rapids/BinaryMerge.java
```diff
@@ -369,15 +412,146 @@ private boolean leftKeyEqual(byte x[][], long xi, long yi) {
...
+        midSearchInd[1] = midSearchInd[0];
+      } else {  // iterate and binarysearch frame key match
+        return midSearchInd;
+      }      
+    }
+    midSearchInd[0] = -1;  // no match found
+    return midSearchInd;
+  }
+  
+  //*********  Not used, but DO NOT DELETE ***********
```
honzasterba, line 545: why not? you clearly replaced it with binaryMergeOneMSB

wendycwong:
@jakubhava Yes, my fix will fix your problem.  I made your code part of my test code in R.  Thank you.

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: Honza:  They have done something clever here that I do not want to lose until I prove that their timing is worse than mine.  Please indulge me.

h2o-r/tests/runitUtils/utilsR.R
wendycwong: Done.

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: I want to keep that old code here until the timing test is completed.

wendycwong:
In the original binary merge implementation of trying to match the sorted keys in the left and right frames, binary search is used on both the left and right frame keys.  This basically means we start looking at the left frame keys in the middle of the key array and then try to match it with a right frame key using binary search on the right frame keys.  Once we are done, we proceed to look at either the middle of the lower half or upper half of the left hand key arrays depending on the previous merge result.  However, there is a bug in the code so that it actually missed some left array keys in its jumping around.  I don't want to deal with this messy code and hence wrote my own.

wendycwong:
In my merge, I will go through the left key array one by one skipping over duplicated ones and then use binary search to find the match in the right key arrays.  I will continue to shrink the range of right key arrays to search through depending on the relative magnitude of the key arrays.  If all.x = true, I believe the time taken for both algos should be close.

wendycwong:
However, when all.x=all.y=False, I tried to optimize it a bit by going through the shorter key array and use binary search to jump through the longer key arrays.  In this case, I believe the old merge method should be faster.

wendycwong:
Here are my results from Hadoop run:

image

wendycwong:
Here is the merge code used to run the tests: 

nrow = 150000000
ncol = 2
iRange = nrow/10
frame1 = h2o.create_frame(rows=nrow, cols=ncol, integer_fraction=1, seed = 12345, integer_range = iRange, missing_fraction=0.0)
frame2 = h2o.create_frame(rows = nrow, cols = ncol, integer_fraction=1, seed = 54321, integer_range = iRange, missing_fraction=0.0)

frame1.set_names(["C1","C2"])
frame2.set_names(["C1","C3"])
time1 = time.time()
mergedLeft = frame1.merge(frame2,by_x=["C1"],by_y=["C1"], all_x=True)
print(mergedLeft[0,0])
leftMergeTime = time.time() - time1
print "Time taken to merge left is {0} seconds".format(leftMergeTime)
time2 = time.time()
mergedExact = frame1.merge(frame2, by_x=["C1"],by_y=["C1"], all_x=False, all_y=False)
print(mergedExact[0,0])
exactMergeTime = time.time() - time2
print "Time taken to merge with common keys are {0} seconds".format(exactMergeTime)
print "Done!"

wendycwong:
It turns out both timing are close enough.

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: I removed the code

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: Code removed.

h2o-core/src/main/java/water/rapids/BinaryMerge.java
```diff
@@ -357,145 +351,134 @@ private long bsearchLeft(long x, boolean returnLow, long upp)  {
...
+   * @param riteUppIn: number of rows in rite frame with specific MSB
+   *                  
+   */
+  private void binaryMergeOneMSB(long leftLowIn, long leftUppIn, long riteLowIn, long riteUppIn) {
+    if (!_allLeft && riteUppIn==0) return;  // no merging possible with empty rite frame here
+    boolean leftFrameIterate = _allLeft?true:((leftUppIn-leftLowIn)>(riteUppIn-riteLowIn)?false:true);
+    long iterIndex = leftFrameIterate?leftLowIn+1:riteLowIn+1;
+    long iterUppIn = leftFrameIterate?leftUppIn:riteUppIn;
+    long binarySearchUppIn = leftFrameIterate?riteUppIn:leftUppIn;
+    long iterHigh, binarySearchLow, binarySearchHigh, tempValue;  // temp variables to perform search
```
honzasterba, line 381: tempValue can be a local variable

h2o-core/src/main/java/water/rapids/BinaryMerge.java
```diff
@@ -357,145 +351,134 @@ private long bsearchLeft(long x, boolean returnLow, long upp)  {
...
+    long[] midSearchInd = new long[3]; // store match row, riteLow, riteUpp
```
honzasterba, line 382: there is no need for this to be an array, the comment is suggestion enough that this should be 3 variables or an object with 3 fields

h2o-core/src/main/java/water/rapids/BinaryMerge.java
```diff
@@ -357,145 +351,134 @@ private long bsearchLeft(long x, boolean returnLow, long upp)  {
...
+    
+    while (iterIndex < iterUppIn) { // for each left row, find matches in the right frame using binary search
+      midSearchInd = bsearchRiteMatch(iterIndex, midSearchInd, leftFrameIterate);  // try to find match for leftFrame at leftIndex in riteFrame
+      // extend index to include/skip over duplicates in iterate frame
+      iterHigh = iterIndex+1;
+      while (iterHigh<iterUppIn && frameKeyEqual(leftFrameIterate?_leftKO._key:_riteKO._key, iterIndex, iterHigh,
+              leftFrameIterate?_leftKO:_riteKO, leftFrameIterate?_leftSB:_riteSB)) {
+        iterHigh++; // find iterate frame duplicates
+      }
+      iterLen = (int) (iterHigh-iterIndex); // number of iterate frame rows to be included
```
honzasterba, line 395: it seems to me that iterlen is unnecessary and be replaced below with iterHigh-iterIndex

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: True.  Fixed.

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: Honza: 

I made an object with three values out of it.  It makes things clearer and I still can get multiple values returned from functions.

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: True.  fixed.

h2o-core/src/main/java/water/rapids/BinaryMerge.java
```diff
@@ -357,145 +351,147 @@ private long bsearchLeft(long x, boolean returnLow, long upp)  {
...
+   * 
+   * @param leftLowIn: left frame lowest row number minus 1
+   * @param leftUppIn: number of rows in left frame with specific MSB
+   * @param riteLowIn: rite frame lowest row number minus 1 with specific MSB
+   * @param riteUppIn: number of rows in rite frame with specific MSB
+   *                  
+   */
+  private void binaryMergeOneMSB(long leftLowIn, long leftUppIn, long riteLowIn, long riteUppIn) {
+    if (!_allLeft && riteUppIn == 0) return;  // no merging possible with empty rite frame here
+    boolean leftFrameIterate = _allLeft ? true : ((leftUppIn - leftLowIn) > (riteUppIn - riteLowIn) ? false : true);
```
honzasterba, line 377: leftFrameIterate can definitely be final, so can iterUppIn, maybe other as well?

h2o-core/src/main/java/water/rapids/BinaryMerge.java
```diff
@@ -357,145 +351,147 @@ private long bsearchLeft(long x, boolean returnLow, long upp)  {
...
-        tmpLow = mid;
-        tmpUpp = mid;
-        while (tmpLow < rUpp - 1) {
-          mid = tmpLow + (rUpp - tmpLow) / 2;
-          if (keycmp(_leftKO._key, lr, _riteKO._key, mid) == 0) tmpLow = mid;
-          else rUpp = mid;
+      if (binarySearchInfo._matchIndex >= 0) {  // match found in riteFrame, 
+        // find duplicates in binarysearch frame, could be lower or higher than midSearchInd[0]
+        binarySearchLow = binarySearchInfo._matchIndex;
+        long tempValue = binarySearchInfo._matchIndex - 1;
```
honzasterba, line 398: please find better name for tempValue

h2o-core/src/main/java/water/rapids/BinaryMerge.java
```diff
@@ -357,145 +351,147 @@ private long bsearchLeft(long x, boolean returnLow, long upp)  {
...
+   * Hence, in this case, we can search over either the left or rite frame depending on the number of rows to be 
+   * searched through.  
+   * 
+   * @param leftLowIn: left frame lowest row number minus 1
+   * @param leftUppIn: number of rows in left frame with specific MSB
+   * @param riteLowIn: rite frame lowest row number minus 1 with specific MSB
+   * @param riteUppIn: number of rows in rite frame with specific MSB
+   *                  
+   */
+  private void binaryMergeOneMSB(long leftLowIn, long leftUppIn, long riteLowIn, long riteUppIn) {
```
honzasterba, line 375: what if instead of doing leftFrameIterate ?: all over the place you just swap left-right values in the beginning and than just have something like iterating-frame and other-frame style variables?

h2o-core/src/main/java/water/rapids/BinaryMerge.java
```diff
@@ -357,145 +351,147 @@ private long bsearchLeft(long x, boolean returnLow, long upp)  {
...
+    long _upperSearchIndex;
+    
+    public MidSearchInfo(long matchI, long lowI, long highI) {
+      _matchIndex = matchI;
+      _lowSearchIndex = lowI;
+      _upperSearchIndex = highI;
+    }
+  }
+  
+  private void populate_ret1st_retLen(int iterLen, int binarySearchLen, long iterIndex, long matchIndex,
```
honzasterba, line 436: please try to stick to java naming conventions

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: yes, I added them.

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: Done.

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: There is a small problem with this.  Left and rite frames are not equivalent because of the following legal conditions:

all.x = true (left frame included), all.y = false;
all.x = false, all.y = false

x refers to left frame and y refers to right frame.

In terms of performing binary search, they are equivalent, you can do the binary search over the left frame keys or right frame keys.

However, the similarity ends there at the end of search.  I need to populate two data structure ret1st and retLen.  These data structure always refer to the left frame.  Hence, I cannot just swap the left and rite frame because in the end, I need to know which frame is the left frame and build up ret1st and retLen.

h2o-core/src/main/java/water/rapids/BinaryMerge.java
wendycwong: Good catch.  Fixed.

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -139,45 +139,48 @@ public static Frame merge(final Frame leftFrame, final Frame riteFrame, final in
...
-      if (allLeft) for (int leftMSB=(int)leftMSBto+1; leftMSB<=255; leftMSB++) {
-        BinaryMerge bm = new BinaryMerge(new BinaryMerge.FFSB(leftFrame,   leftMSB    ,leftShift,
-                leftIndex._bytesUsed,leftIndex._base),
-                new BinaryMerge.FFSB(rightFrame,/*rightMSB*/-1,riteShift,
-                        riteIndex._bytesUsed,riteIndex._base),
-                true);
-          bmList.add(bm);
-          fs.add(new RPC<>(SplitByMSBLocal.ownerOfMSB(leftMSB), bm).call());
-      }
+        if (allLeft) {  // not worthy restricting length here unless store column max.
```
honzasterba, line 173: indent seems off

wendycwong: Good catch.  I fixed it.

PR #4517, opened 2020-04-15
[PUBDEV-7046] use Log4j as backend for water.util.Log

honzasterba:
the main result of this PR is that you can now use Log4j Loogers in your code and change the logging level only for one component, which makes debugging from logs much easier


h2o-core/src/main/java/water/util/Log.java
```diff
@@ -97,105 +106,57 @@ public static void setLogLevel(String sLvl) {
...
-    case WARN: l4j.warn (sb); break;
-    case INFO: l4j.info (sb); break;
-    case DEBUG:l4j.debug(sb); break;
-    case TRACE:l4j.trace(sb); break;
-    default:
-      l4j.error("Invalid log level requested");
-      l4j.error(s);
+    String msg = msgBuff.toString();
+    if (_bufferMessages) { // Oops, need to buffer until we can do a proper header
+      INIT_MSGS.add(new BufferedMsg(lvl, msg, t));
```
michalkurka, line 134: does this preserve the information about the context? eg. the thread that logged the message?

h2o-core/src/main/java/water/parser/ParseDataset.java
```diff
@@ -1127,11 +1127,16 @@ public static void logParseResults(Frame fr) {
         }
       }
 
-      if (printLogSeparatorToStdout)
+      if (printLogSeparatorToStdout) {
         Log.info("Additional column information only sent to log file...");
```
jakubhava, line 1131: At some places you use LOG from log4j and at some places Log from water.util. Is it on purpose?

honzasterba: it is impossible to replace all uses of water.Log in one PR, this has to be done gradually, and for some places around h2o-core we might stick to water.Log because of its semantics (handling newlines for example)

h2o-core/src/main/java/water/util/Log.java
honzasterba: no, for such messages the context (time, thread) is lost, do you see that as an issue?

michalkurka: IMHO not a big issue, typically it will be logged by the main thread anyway

it could help with debugging what called an exit during the start-up process... but not a big issue I think

h2o-automl/src/main/java/ai/h2o/automl/ModelSelectionStrategies.java
```diff
@@ -2,15 +2,17 @@
...
 import water.util.ArrayUtils;
-import water.util.Log;
 
 import java.util.Arrays;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 public final class ModelSelectionStrategies {
+    
+    private static final Logger LOG = Logger.getLogger(ModelSelectionStrategies.class);
```
Pscheidl, line 15: :pray:

Pscheidl:
Great changes. How about we do one more thing and bring SLF4J to H2O ? We don't really need to do it now, but it'd be nice to have it ASAP if we decide to use it. 

http://www.slf4j.org/docs.html


Also, turns out SLF4J has **fluent logging API** that actually [improves the performance](https://google.github.io/flogger/benefits) and eliminates the need for `if(LOG.isDebugEnabled())`.

honzasterba:
@Pscheidl I was thinking about using SLF4J, but the point of SLF4J is flexibility, right now we are tied to log4j v1, event the migration to Log4j 2 is a big undertaking, so I am not sure that SLF4J makes sense.

Pscheidl:
This is the truth, unfortunately. These are the problems I currently see (from technical POV):

1. Hadoop driver (`h2odriver.java`) depends on Log4J 1.x and I have no idea ATM whether using other backend (talking generally), including LOG4J2 or Logback would break something. cc @michalkurka 
2. Initialization of LOG4J is done dynamically at runtime and is not enforced during `configureLogging` method in `H2O` class. For more information, have a look how and where`Log.setLog4jProperties` is called.

If the main method in `H2O.java` tried to obtain an SLF4J logger and log something directly after `H2O.configureLogging();` has been logged, nothing happens. Try it for yourself:

```patch
Index: h2o-core/src/main/java/water/H2O.java
IDEA additional info:
<+>UTF-8
--- h2o-core/src/main/java/water/H2O.java	(revision 169bfb878865d36459453ce2c7810b43e5e8a616)
+++ h2o-core/src/main/java/water/H2O.java	(date 1587116268835)
@@ -5,6 +5,7 @@
 import jsr166y.ForkJoinPool;
 import jsr166y.ForkJoinWorkerThread;
 import org.apache.log4j.LogManager;
+import org.apache.log4j.Logger;
 import org.apache.log4j.PropertyConfigurator;
 import water.UDPRebooted.ShutdownTsk;
 import water.api.LogsHandler;
@@ -2143,6 +2144,8 @@
   // --------------------------------------------------------------------------
   public static void main( String[] args ) {
    H2O.configureLogging();
+   Logger log4jLogger = Logger.getLogger(H2O.class);
+   log4jLogger.info("WILL NOT APPEAR.");
    extManager.registerCoreExtensions();
    extManager.registerListenerExtensions();
    extManager.registerAuthExtensions();

```

This way, embedded configs and other stuff initializing before H2O is fully started have no way to log using SLF4J, only our Log works, as the configuration has not yet been done.

To make this problem even **more complex**, there are args like `log_dir`, `log_level`, and newly `max_log_file_size` we have to respect. During parameter parsing, there is, again, no way to log using standard SLF4J, as the configuration has not yet been done. Solving this would require refactoring this part.

3. As for SLF4J, it is exactly right that we have no idea what we're going to end up with. And switching the backend would be much easier with SLF4J in place. But at the same time, I see it very easy to refactor from using Log4J directly to one of SLF4J's APIs.

h2o-py/h2o/backend/server.py
```diff
@@ -477,7 +477,7 @@ def _get_server_info_from_logs(self):
 
         :returns: (scheme, ip, port) tuple if the server has already started, None otherwise.
         """
-        searchstr = "INFO: Open H2O Flow in your web browser:"
+        searchstr = ": Open H2O Flow in your web browser:"
```
honzasterba, line 480: @michalkurka this means that the python package is not backwards compatible with previous H2O releases

michalkurka: ok, I think this should not go into the release branch then - we also need to make sure if Steam is able to handle it (it used to parse the same String... could be different now)

michalkurka:
On Java level - is this backwards compatible?

honzasterba:
Yes

PR #4521, opened 2020-04-16
Properly clear env vars in K8S Junit test.

Pscheidl:
Not sure how this happened, but the env variable is not cleared in one of the K8S Junit tests, causing it to fail. Fix is trivial.

### DIFF
--- a/h2o-k8s/src/test/java/water/k8s/KubernetesEmbeddedConfigProviderTest.java
+++ b/h2o-k8s/src/test/java/water/k8s/KubernetesEmbeddedConfigProviderTest.java
@@ -32,6 +32,7 @@ public class KubernetesEmbeddedConfigProviderTest {
     
     @Test
     public void testInactiveOnK8SEnvVariablesSet(){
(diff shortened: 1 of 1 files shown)


h2o-k8s/src/test/java/water/k8s/KubernetesEmbeddedConfigProviderTest.java
```diff
@@ -32,6 +32,7 @@ public void testActiveOnK8SEnvVariablesSet(){
     
     @Test
     public void testInactiveOnK8SEnvVariablesSet(){
+        environmentVariables.clear("H2O_KUBERNETES_SERVICE_DNS");
         kubernetesEmbeddedConfigProvider.init();
         assertFalse(kubernetesEmbeddedConfigProvider.isActive());
```
Pscheidl, line 37: We're testing for **inactivity** here. If the env var is not cleared, it will actually return true.

Pscheidl:
Fast review is appreciated.

Pscheidl:
:pray:

PR #4523, opened 2020-04-16
PUBDEV-7400

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-7400

added heuristic to exclude debug referrers from counting to avoid creation of self-referential tmp frames.

Should work at least with IDEs whose Python debugging integration is based on pydevd debugger: https://pypi.org/project/pydevd/

### DIFF
--- a/h2o-py/h2o/expr.py
+++ b/h2o-py/h2o/expr.py
@@ -12,6 +12,7 @@ import copy
 import gc
 import inspect
 import math
+import sys
 import time
 import traceback
 import numbers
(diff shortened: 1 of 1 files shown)


h2o-py/h2o/expr.py
```diff
@@ -149,12 +148,17 @@ def _get_ast_str(self, top):
         def is_ast_expr(ref):
             return isinstance(ref, list) and any(map(lambda r: isinstance(r, ASTId), ref))
 
+        def is_debug_ref(ref):
+            # if ref is a dict, then it is a `locals` scope: most of those are added in debug mode.
+            # However, keeping it if this scope refers to an H2OFrame (has `_ex` attribute`
```
sebhrusen, line 153:
```suggestion
            # However, keeping it if this scope refers to an H2OFrame (has `_ex` entry)
```

sebhrusen:
All Py tests passed.

PR #4524, opened 2020-04-16
PUBDEV-7463: Speed up training and prediction of Stacked Ensemble

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7463

This PR consists of two changes:

- Parallel scoring of base models
- Optionally using random subset of training data to calculate training metrics
  - I created `training_scoring_subsample_size` property so user can specify what size (s)he wants to use. -1 uses the whole training frame.

N.B. `training_scoring_subsample_size` probably isn't the best name so if you have better ideas I will be happy to change it I just can't think of anything better right now.

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -123,7 +123,9 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
     if (aModel._output.isBinomialClassifier()) {
       // GLM uses a different column name than the other algos
       Vec preds = aModelsPredictions.vec(2); // Predictions column names have been changed...
-      levelOneFrame.add(aModel._key.toString(), preds);
+      synchronized (levelOneFrame) {
+        levelOneFrame.add(aModel._key.toString(), preds);
+      }
     } else if (aModel._output.isMultinomialClassifier()) { //Multinomial
       //Need to remove 'predict' column from multinomial since it contains outcome
       Frame probabilities = aModelsPredictions.subframe(ArrayUtils.remove(aModelsPredictions.names(), "predict"));
@@ -132,14 +134,18 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
                       .map((name) -> aModel._key.toString().concat("/").concat(name))
                       .toArray(String[]::new)
       );
-      levelOneFrame.add(probabilities);
+      synchronized (levelOneFrame) {
+        levelOneFrame.add(probabilities);
+      }
     } else if (aModel._output.isAutoencoder()) {
       throw new H2OIllegalArgumentException("Don't yet know how to stack autoencoders: " + aModel._key);
     } else if (!aModel._output.isSupervised()) {
       throw new H2OIllegalArgumentException("Don't yet know how to stack unsupervised models: " + aModel._key);
     } else {
       Vec preds = aModelsPredictions.vec("predict");
-      levelOneFrame.add(aModel._key.toString(), preds);
+      synchronized (levelOneFrame) {
+        levelOneFrame.add(aModel._key.toString(), preds);
+      }
     }
   }
 
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -123,9 +123,7 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
     if (aModel._output.isBinomialClassifier()) {
       // GLM uses a different column name than the other algos
       Vec preds = aModelsPredictions.vec(2); // Predictions column names have been changed...
-      synchronized (levelOneFrame) {
-        levelOneFrame.add(aModel._key.toString(), preds);
-      }
+      levelOneFrame.add(aModel._key.toString(), preds);
     } else if (aModel._output.isMultinomialClassifier()) { //Multinomial
       //Need to remove 'predict' column from multinomial since it contains outcome
       Frame probabilities = aModelsPredictions.subframe(ArrayUtils.remove(aModelsPredictions.names(), "predict"));
@@ -134,18 +132,14 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
                       .map((name) -> aModel._key.toString().concat("/").concat(name))
                       .toArray(String[]::new)
       );
-      synchronized (levelOneFrame) {
-        levelOneFrame.add(probabilities);
-      }
+      levelOneFrame.add(probabilities);
     } else if (aModel._output.isAutoencoder()) {
       throw new H2OIllegalArgumentException("Don't yet know how to stack autoencoders: " + aModel._key);
     } else if (!aModel._output.isSupervised()) {
       throw new H2OIllegalArgumentException("Don't yet know how to stack unsupervised models: " + aModel._key);
     } else {
       Vec preds = aModelsPredictions.vec("predict");
-      synchronized (levelOneFrame) {
-        levelOneFrame.add(aModel._key.toString(), preds);
-      }
+      levelOneFrame.add(aModel._key.toString(), preds);
     }
   }
 
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
@@ -226,7 +226,16 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
 
   ModelMetrics doScoreMetricsOneFrame(Frame frame, Job job) {
     if (_parms._training_scoring_subsample_size > 0 && _parms._training_scoring_subsample_size < frame.numRows()) { 
-      frame =  MRUtils.sampleFrame(frame, _parms._training_scoring_subsample_size, _parms._seed);
+Frame scoredFrame = (_parms._training_scoring_subsample_size > 0 && _parms._training_scoring_subsample_size < frame.numRows()) 
+        ? MRUtils.sampleFrame(frame, _parms._training_scoring_subsample_size, _parms._seed)
+        : frame;
+try {
+      Frame pred = this.predictScoreImpl(frame, new Frame(scoredFrame), null, job, true, CFuncRef.from(_parms._custom_metric_func));
+      pred.delete();
+      return ModelMetrics.getFromDKV(this, frame);
+} finally {
+     if (scoredFrame != frame) scoredFrame.delete();
+}  
       Frame pred = this.predictScoreImpl(frame, new Frame(frame), null, job, true, CFuncRef.from(_parms._custom_metric_func));
       pred.delete();
       ModelMetrics metrics = ModelMetrics.getFromDKV(this, frame);
@@ -499,4 +508,3 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
     _output._metalearner.deleteCrossValidationFoldAssignment();
   }
 }
-
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
@@ -9,10 +9,13 @@ import water.exceptions.H2OIllegalArgumentException;
 import water.fvec.Frame;
 import water.udf.CFuncRef;
 import water.util.Log;
+import water.util.MRUtils;
 import water.util.ReflectionUtils;
 
 import java.lang.reflect.Field;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 
 import static hex.Model.Parameters.FoldAssignmentScheme.AUTO;
 import static hex.Model.Parameters.FoldAssignmentScheme.Random;
@@ -65,6 +68,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
     public String _metalearner_params = new String(); //used for clients code-gen only.
     public Model.Parameters _metalearner_parameters;
     public long _seed;
+    public long _training_scoring_subsample_size = 10_000;
 
     /**
      * initialize {@link #_metalearner_parameters} with default parameters for the current {@link #_metalearner_algorithm}.
@@ -116,22 +120,38 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
   protected Frame predictScoreImpl(Frame fr, Frame adaptFrm, String destination_key, Job j, boolean computeMetrics, CFuncRef customMetricFunc) {
     Frame levelOneFrame = new Frame(Key.<Frame>make("preds_levelone_" + this._key.toString() + fr._key));
 
-    // TODO: we should be able to parallelize scoring of base models
+    List<H2O.H2OCountedCompleter> tasks = new ArrayList<>();
+    final String seKey = this._key.toString();
+
+    // Run scoring of base models in parallel
     for (Key<Model> baseModelKey : _parms._base_models) {
       if (isUsefulBaseModel(baseModelKey)) {
-        Model baseModel = baseModelKey.get();
+        tasks.add(
+                H2O.submitTask(
+                        new H2O.H2OCountedCompleter() {
+                          @Override
+                          public void compute2() {
+                            Model baseModel = baseModelKey.get();
 
-        Frame basePreds = baseModel.score(
-                fr,
-                "preds_base_" + this._key.toString() + fr._key,
-                j,
-                false
-        );
+                            Frame basePreds = baseModel.score(
+                                    fr,
+                                    "preds_base_" + seKey + fr._key,
+                                    j,
+                                    false
+                            );
 
-        StackedEnsemble.addModelPredictionsToLevelOneFrame(baseModel, basePreds, levelOneFrame);
-        DKV.remove(basePreds._key); //Cleanup
-        Frame.deleteTempFrameAndItsNonSharedVecs(basePreds, levelOneFrame);
+                            StackedEnsemble.addModelPredictionsToLevelOneFrame(baseModel, basePreds, levelOneFrame);
+                            DKV.remove(basePreds._key); //Cleanup
+                            Frame.deleteTempFrameAndItsNonSharedVecs(basePreds, levelOneFrame);
+                            tryComplete();
+                          }
+                        }));
       }
+    }
+
+    // Wait for the completion of the base model scoring
+    for (H2O.H2OCountedCompleter task : tasks) {
+      task.join();
     }
 
     // Add response column to level one frame
@@ -205,10 +225,18 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
   }
 
   ModelMetrics doScoreMetricsOneFrame(Frame frame, Job job) {
+    if (_parms._training_scoring_subsample_size > 0 && _parms._training_scoring_subsample_size < frame.numRows()) { 
+      frame =  MRUtils.sampleFrame(frame, _parms._training_scoring_subsample_size, _parms._seed);
       Frame pred = this.predictScoreImpl(frame, new Frame(frame), null, job, true, CFuncRef.from(_parms._custom_metric_func));
-//      Frame pred = this.score(frame, null, job, true,  CFuncRef.from(_parms._custom_metric_func));
+      pred.delete();
+      ModelMetrics metrics = ModelMetrics.getFromDKV(this, frame);
+      frame.delete();
+      return metrics;
+    } else {
+      Frame pred = this.predictScoreImpl(frame, new Frame(frame), null, job, true, CFuncRef.from(_parms._custom_metric_func));
       pred.delete();
       return ModelMetrics.getFromDKV(this, frame);
+    }
   }
 
   void doScoreOrCopyMetrics(Job job) {
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
@@ -219,7 +219,7 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
     throw new UnsupportedOperationException("StackedEnsembleModel.makeMetricBuilder should never be called!");
   }
 
-  ModelMetrics doScoreMetricsOneFrame(Frame frame, Job job) {
+  private ModelMetrics doScoreTrainingMetrics(Frame frame, Job job) {
     Frame scoredFrame = (_parms._score_training_samples > 0 && _parms._score_training_samples < frame.numRows())
             ? MRUtils.sampleFrame(frame, _parms._score_training_samples, _parms._seed)
             : frame;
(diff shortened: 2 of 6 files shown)


h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
```diff
@@ -109,6 +110,11 @@ public AlgorithmValuesProvider() {
     @API(help = "Seed for random numbers; passed through to the metalearner algorithm. Defaults to -1 (time-based random number)", gridable = true)
     public long seed;
 
+    @API(help = "Size of training subsample used to compute training metrics. Use -1 to use the whole training frame.",
+            level = API.Level.secondary,
+            direction = API.Direction.INOUT)
+    public long training_scoring_subsample_size;
+
```
sebhrusen, line 117: wondering if it's worth exposing this to the API, really.
If user is not happy with the sample metric, then he can just run the following afterwards:
```
se.predict(train)  # to get metric on the entire frame
se.predict(train.split_frame([0.5])[0])  # to get metric on half of it
```
to me, not a useful addition to the API. Even the complicated name suggests "don't use me".
@erin, any opinion?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -205,10 +225,18 @@ boolean isUsefulBaseModel(Key<Model> baseModelKey) {
   }
 
   ModelMetrics doScoreMetricsOneFrame(Frame frame, Job job) {
+    if (_parms._training_scoring_subsample_size > 0 && _parms._training_scoring_subsample_size < frame.numRows()) { 
+      frame =  MRUtils.sampleFrame(frame, _parms._training_scoring_subsample_size, _parms._seed);
```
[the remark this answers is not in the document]
tomasfryda: Thanks!

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -205,10 +225,27 @@ boolean isUsefulBaseModel(Key<Model> baseModelKey) {
...
+      pred.delete();
+      ModelMetrics metrics = ModelMetrics.getFromDKV(this, frame);
+      frame.delete();
+      return metrics;
+    } else {
       Frame pred = this.predictScoreImpl(frame, new Frame(frame), null, job, true, CFuncRef.from(_parms._custom_metric_func));
-//      Frame pred = this.score(frame, null, job, true,  CFuncRef.from(_parms._custom_metric_func));
       pred.delete();
       return ModelMetrics.getFromDKV(this, frame);
+    }
```
sebhrusen, line 248: suggestion was committed in wrong place apparently... it couldn't be applied to whole block due to deleted lines, and beside I forgot to rename some `frame` into `scoreFrame`, should be sth like.
```
Frame scoredFrame = (_parms._training_scoring_subsample_size > 0 && _parms._training_scoring_subsample_size < frame.numRows()) 
        ? MRUtils.sampleFrame(frame, _parms._training_scoring_subsample_size, _parms._seed)
        : frame;
try {
      Frame pred = this.predictScoreImpl(scoredFrame, new Frame(scoredFrame), null, job, true, CFuncRef.from(_parms._custom_metric_func));
      pred.delete();
      return ModelMetrics.getFromDKV(this, scoredFrame);
} finally {
     if (scoredFrame != frame) scoredFrame.delete();
}  
```
better apply it locally first though.

tomasfryda: Thanks!

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -116,24 +120,40 @@ public void initMetalearnerParams(Metalearner.Algorithm algo) {
...
+                                    j,
+                                    false
+                            );
+
+                            StackedEnsemble.addModelPredictionsToLevelOneFrame(baseModel, basePreds, levelOneFrame);
+                            DKV.remove(basePreds._key); //Cleanup
+                            Frame.deleteTempFrameAndItsNonSharedVecs(basePreds, levelOneFrame);
+                            tryComplete();
+                          }
+                        }));
```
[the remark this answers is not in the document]
tomasfryda: I did some implementation using the MRTask but MRTask along with H2O's multi-node rpc/DKV are still new to me so I have some questions:

- Assumption: `Frame fr` is a lightweight object that contains references to `DKV` to its `Vec`s; `Vec`s contain `Chunk`s and `Chunk`s can be distributed on multiple nodes. 
  - When I run from local node `model.score(fr)` will it download the parts of the frame to local node and score it locally or does it broadcast the model to the nodes and evaluate the model in parallel for each part of the frame?
    - (The former isn't very practical (every node fetching the same data for the distributed base model scoring) so I am assuming that the latter is closer to the actual implementation.)
    - If the latter is true, would it really make a significant difference if I would score from single node or mutliple nodes? 

- If I run MR on one element, it seems to me that it won't run the reduce step. Is that correct? If so, I should "clean the dirt" in the map phase (current implementation).

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -103,6 +109,71 @@ public void initMetalearnerParams(Metalearner.Algorithm algo) {
...
+      >>>>> Key ='$04ff11000000ffffffff$nfs:/.././smalldata/testng/prostate_test.csv' value = 'adapted missing vectors' << This is leaked without the cleanup
+      >>>>> Key ='$04ff1c000000ffffffff$nfs:/.././smalldata/testng/prostate_test.csv' value = 'categorically adapted vec'
+      >>>>> Key ='$04ff1d000000ffffffff$nfs:/.././smalldata/testng/prostate_test.csv' value = 'categorically adapted vec'
+      >>>>> Key ='$04ff10000000ffffffff$nfs:/.././smalldata/testng/prostate_test.csv' value = 'adapted missing vectors' << This is leaked without the cleanup
+      */
+      // And leaks both of the "adapted missing vectors" on the slave[1] node.
+      //
+      // [1] By slave node I meant a node that wasn't the one that was directly invoked to run the test
+
+      FrameUtils.cleanUp(baseModel._toDelete);
```
tomasfryda, line 161: @sebhrusen could you look at this. It seems to fix the problem with leaking keys (at least locally) , but I am wondering if it shouldn't be fixed somewhere else.

When I run multi-node tests, `test_SE_scoring_with_missing_response_column` fails with leaked keys. The keys that leak are subset of the ones in `baseModel._toDelete` and AFAIK always the ones with "adapted missing vectors" label and also must be created on slave[1] node. The other keys that are contained in `baseModel._toDelete` but don't leak appear to be recreated on master[1] node (I guess we deterministically generate Vec key names) and deleted afterwards.

 (also please see the comment in the source code for more details)

[1] By master node I mean the one that was invoked to run the test and by slave I mean any other node in the cluster.

h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
```diff
@@ -45,6 +45,7 @@
       "metalearner_fold_column",
       "metalearner_params",
       "seed",
+      "training_scoring_subsample_size",
```
ledell, line 48: In Deep Learning, we call this parameter, `score_training_samples` set to 10,000 by default (we can decide what a good default for SE is).  So we should use the same parameter name.

```
score_training_samples: Specify the number of training set samples for scoring. The value must be >= 0. To use all training samples, enter 0.
```

h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
ledell: @sebhrusen I am ok either way, though (unlike AutoML), I don't worry about having too many parameters with Stacked Ensemble.  I wonder if we should tweak the default sample size (to something lower than 10,000?), or start with 10k and see if it's scalable.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -103,6 +109,71 @@ public void initMetalearnerParams(Metalearner.Algorithm algo) {
...
   }
 
+  public static class StackedEnsembleMRScorer extends MRTask<StackedEnsembleMRScorer> {
+    // IN
+    private Frame scoringFrame;
+    private String seKey;
+    private Job job;
+
+    // OUT
+    private Frame levelOneFrame;
```
sebhrusen, line 119: `MRTask` is a distributed task, therefore its attribute will be serialized, that's why we should be careful with those attributes types:
- no Frame, but use Key<Frame> instead: the `Frame` should be put in DKV, and retrieved the same way, and eventually removed after usage (for eample in reduce, you have to update the lhs in DKV, and you should remove the rhs).
- Job is also a `Keyed`, so even if it's a much smaller object than Frame, it's better to store just `Key<Job>` in those tasks as well (also note that job can be null).
- don't use String to store keys (`seKey`), but `Key<StackedEnsembleModel>`.

sebhrusen: also, naming conventions: I know they're not standard and can be annoying sometimes, but they are what they are... and let's be consistent in the same class: `_jobKey`, `_scoringFrameKey`, `_seKey`, `_levelOneFrameKey`

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -103,6 +109,71 @@ public void initMetalearnerParams(Metalearner.Algorithm algo) {
     public Key<Frame>[] _base_model_predictions_keys; 
   }
 
+  public static class StackedEnsembleMRScorer extends MRTask<StackedEnsembleMRScorer> {
```
sebhrusen, line 112: name is a bit misleading: strictly speaking, it doesn't "score" the model, it's preparing the level one frame that will be used to score the metalearner: what about `BaseModelsMRScorer`?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
sebhrusen: let's see if you still have leaking keys after using `Key<Frame>` instead of `Frame` above and managing them using DKV + frame update (write_lock, update, unlock)

h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
sebhrusen: 10k looks reasonable, the goal with such a fix is to prevent waste of computation time for something not really useful like `training_metrics`.

h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
sebhrusen: ok, if there's a parameter with the same semantic for DL, then let's expose it, and use the same name.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -103,6 +109,71 @@ public void initMetalearnerParams(Metalearner.Algorithm algo) {
...
+
+    @Override
+    public void reduce(StackedEnsembleMRScorer mrt) {
+      levelOneFrame.add(mrt.levelOneFrame);
+    }
+
+    public Frame getLevelOneFrame() {
+      // levelOneFrame can be uninitialized when we don't have any useful base-models, e.g., metalearner is using only intercept
+      levelOneFrame = levelOneFrame == null ? new Frame() : levelOneFrame;
+      levelOneFrame._key = Key.<Frame>make("preds_levelone_" + seKey + scoringFrame._key);
```
sebhrusen, line 172: if you change the key, you probably have to remove the old key from DKV and add this new one: this is a potential key leak.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -114,25 +185,11 @@ public void initMetalearnerParams(Metalearner.Algorithm algo) {
...
-        StackedEnsemble.addModelPredictionsToLevelOneFrame(baseModel, basePreds, levelOneFrame);
-        DKV.remove(basePreds._key); //Cleanup
-        Frame.deleteTempFrameAndItsNonSharedVecs(basePreds, levelOneFrame);
-      }
-    }
+    Key[] baseModels = Stream.of(_parms._base_models)
+            .filter(this::isUsefulBaseModel)
+            .toArray(Key[]::new);
+
+    Frame levelOneFrame = new StackedEnsembleMRScorer(fr, _key.toString(), j).doAll(baseModels).getLevelOneFrame();
```
sebhrusen, line 192: side note, wondering if we would profit from a similar approach when preparing levelOneFrame in `StackedEnsemble`, especially when using `blending` as in this case we need to explicitely score all base models against the blending frame.
Could be a future improvement.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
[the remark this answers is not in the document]
sebhrusen: sorry, I missed that comment:
> If I run MR on one element, it seems to me that it won't run the reduce step. Is that correct? If so, I should "clean the dirt" in the map phase (current implementation).

I think you're right. But the dirt you're creating is specific to the map phase so it makes sense to clean it there.

> When I run from local node model.score(fr) will it download the parts of the frame to local node and score it locally or does it broadcast the model to the nodes and evaluate the model in parallel for each part of the frame?

model scoring is implemented through the `BigScore` task which is an `MRTask`, so it has to be the latter.

> If the latter is true, would it really make a significant difference if I would score from single node or mutliple nodes?
very good point indeed... probably not...
Sorry to reply after your did the changes, but do you see any difference?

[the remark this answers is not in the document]
tomasfryda: I didn't test it well enough (just ~5 runs) to say if there is some difference - they were very close sometimes the first was faster sometimes the second solution was faster.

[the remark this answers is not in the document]
michalkurka: > If the latter is true, would it really make a significant difference if I would score from single node or mutliple nodes?

I think using a single node shouldn't be slower, maybe by just a little bit. There is an advantage of using a single node (in general):
1. are logs are in one place - not super important but can be convenient
2. model metrics are stable - eg. AUC will be slightly different depending on the node that initiates the scoring (the order of the reduces will be different and the histogram bins might be different as a result). In this case, you don't care about the metrics.
3. you can better control parallelism.

I would go for `LocalMR` implementation of the scoring, it will give you a better option of controlling how many parallel scoring tasks you want to invoke.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -116,24 +120,40 @@ public void initMetalearnerParams(Metalearner.Algorithm algo) {
   protected Frame predictScoreImpl(Frame fr, Frame adaptFrm, String destination_key, Job j, boolean computeMetrics, CFuncRef customMetricFunc) {
     Frame levelOneFrame = new Frame(Key.<Frame>make("preds_levelone_" + this._key.toString() + fr._key));
 
-    // TODO: we should be able to parallelize scoring of base models
+    List<H2O.H2OCountedCompleter> tasks = new ArrayList<>();
```
michalkurka, line 123: this is emulating what `Futures` are meant to do, I would rather use Futures or switch to MRLocal (see my other comment)

sebhrusen: good point, those `H2OCounterCompleter` are `Future`s

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -205,10 +225,16 @@ boolean isUsefulBaseModel(Key<Model> baseModelKey) {
   }
 
   ModelMetrics doScoreMetricsOneFrame(Frame frame, Job job) {
-      Frame pred = this.predictScoreImpl(frame, new Frame(frame), null, job, true, CFuncRef.from(_parms._custom_metric_func));
-//      Frame pred = this.score(frame, null, job, true,  CFuncRef.from(_parms._custom_metric_func));
+    Frame scoredFrame = (_parms._score_training_samples > 0 && _parms._score_training_samples < frame.numRows())
+            ? MRUtils.sampleFrame(frame, _parms._score_training_samples, _parms._seed)
```
michalkurka, line 229: Sampling will oftentimes be more expensive than using zero weights to eliminate unwanted rows.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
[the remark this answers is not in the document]
sebhrusen: thanks @michalkurka we also agreed on reverting back to local scoring

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -205,10 +225,16 @@ boolean isUsefulBaseModel(Key<Model> baseModelKey) {
...
+            : frame;
+    try {
+      Frame pred = this.predictScoreImpl(scoredFrame, new Frame(scoredFrame), null, job, true, CFuncRef.from(_parms._custom_metric_func));
```
michalkurka, line 232: This would be a use case for `Model#scoreMetrics` function. It doesn't have the custom metric function though - I don't think you care about the custom metrics here, WDYT?

ledell: @michalkurka For now, we don't support/need custom metrics.  In the future, we might, but we would change it then.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
ledell: @michalkurka Good to know... Do you know what we do in Deep Learning where we also score on just a sub-sample the training set?

tomasfryda: In Deep Learning, we use the sampleFrame method (`hex/deeplearning/DeepLearning.java:378`), that's where I found it.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
sebhrusen: @tomasfryda the switch to `scoreMetrics` can be done in a separate ticket and addressed later as it's not related with your current changes. Feel free to file one in Jira.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
tomasfryda: @michalkurka @sebhrusen I am trying to find a way to use zero weights, so far it seems to me that I have to create a new column with weights and then for each base model (and meta learner) change their property `_weights_column` and protected property `_hasWeights` and then score everything and finally return the properties to their original values. (I am not sure if that's safe)

I guess we can have some cached version of a model on each node so I should do the `write_lock`, `update`, `unlock`, right?

Or is there some better way to do it?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
tomasfryda: Ok, I created it: https://0xdata.atlassian.net/browse/PUBDEV-7478

sebhrusen:
sorry, forgot about the replacement of sampling with 0 weights

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -205,10 +220,16 @@ boolean isUsefulBaseModel(Key<Model> baseModelKey) {
   }
 
   ModelMetrics doScoreMetricsOneFrame(Frame frame, Job job) {
-      Frame pred = this.predictScoreImpl(frame, new Frame(frame), null, job, true, CFuncRef.from(_parms._custom_metric_func));
-//      Frame pred = this.score(frame, null, job, true,  CFuncRef.from(_parms._custom_metric_func));
+    Frame scoredFrame = (_parms._score_training_samples > 0 && _parms._score_training_samples < frame.numRows())
```
sebhrusen, line 223: can we apply this logic in `doScoreOrCopyMetrics`, that is where we know for sure that the frame is the training frame?

sebhrusen: or rename this method to ensure it can't be used for anything else than training frame.. even wondering why this is not private.

sebhrusen: please make it private and rename to `doScoreTrainingMetrics` or sth similar.

tomasfryda: Sure, good point. Thanks!

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
sebhrusen: agree that it looks fishy, would not recommend it.
@michalkurka any existing example showing how to do this safely? Or we can consider this as an improvement for a separate PR...

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -123,7 +123,9 @@ static void addModelPredictionsToLevelOneFrame(Model aModel, Frame aModelsPredic
     if (aModel._output.isBinomialClassifier()) {
       // GLM uses a different column name than the other algos
       Vec preds = aModelsPredictions.vec(2); // Predictions column names have been changed...
-      levelOneFrame.add(aModel._key.toString(), preds);
+      synchronized (levelOneFrame) {
```
michalkurka, line 126: I would be happier if this method didn't use synchronization. The semantics can be changed to instead return a Frame wrapping the columns that would be later stitched together to "levelOneFrame".

1. This method will return new columns
2. Columns will be cbinded together here: https://github.com/h2oai/h2o-3/pull/4524/files#diff-91d255cd89e95ed4a240332df7d9002bR150

WDYT? cbind is a cheap operation

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
michalkurka: I think given that we are making such a small sample (10k) rows any optimization doesn't make much sense. I didn't realize the sample is so "petite". Thanks for looking into it though.

michalkurka:
Great work, I would suggest getting rid off the synchronization. See inline comments.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
tomasfryda: @michalkurka Thank you for the comment!

I tried but didn't find a reasonable way how to collect results from `Futures` so I looked into using `LocalMR`. I came up with the following. It doesn't use `synchronized` but I kept the semantics of the method the same and call it sequentially since it's not the bottleneck (the method takes some columns, renames them, and cbinds them) and it might complicate cleanup.

Performance-wise, I didn't see any significant difference to the previous solution on my mini benchmark.

I didn't commit this change since I am not sure with the use of `LocalMR` and its `nthreads` argument which is often used the way I did - as a way to call `MrFun::map(i) for i in 0..nthreads`. Is that OK? Is that any better than the previous solution?

```java
    Model[] usefulBaseModels = Stream.of(_parms._base_models)
            .filter(this::isUsefulBaseModel)
            .map(Key::get)
            .toArray(Model[]::new);

    Frame[] baseModelPredictions = new Frame[usefulBaseModels.length];

    // Run scoring of base models in parallel
    H2O.submitTask(new LocalMR(new MrFun() {
      @Override
      protected void map(int id) {
        baseModelPredictions[id] = usefulBaseModels[id].score(
                fr,
                "preds_base_" + seKey + usefulBaseModels[id]._key + fr._key,
                j,
                false
        );
      }
    }, usefulBaseModels.length)).join();

    for (int i = 0; i < usefulBaseModels.length; i++) {
      StackedEnsemble.addModelPredictionsToLevelOneFrame(usefulBaseModels[i], baseModelPredictions[i], levelOneFrame);
      DKV.remove(baseModelPredictions[i]._key);
      Frame.deleteTempFrameAndItsNonSharedVecs(baseModelPredictions[i], levelOneFrame);
    }
```

PR #4525, opened 2020-04-16
Use a version of sphinx and sphinx_substitution_extensions that support Py3.5

angela0xdata:
Confirmed that other requirements listed here are also compatible.


angela0xdata:
Build passed: http://mr-0xc1:8080/view/H2O-3/job/docker-images/job/h2o-3-dev-images/68/

PR #4529, opened 2020-04-17
PUBDEV-7089 Add info message instead of NPE in case of wrong usage of P…

koniecsveta:
In case you try to execute PrintMojo with --format png by using h2o.jar instead of h2o-genmodel.jar, ServiceLoader doesn't see JgraphtPrintMojo class and hence PrintMojo is not able to find mojoPrinter which supports required .png format. This produces NPE and it looks like the tool doesn't work. After this change, instead of printing NPE it will print informative message: "No supported MojoPrinter for the format required found. Please make sure you are using h2o-genmodel.jar for executing this tool."

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
@@ -44,10 +44,14 @@ public class PrintMojo implements MojoPrinter {
           mojoPrinter = printer;
         }
       }
+      if (mojoPrinter == null) {
+        System.out.println("No supported MojoPrinter for format required found. Please make sure you are using h2o-genmodel.jar for executing this tool.");
+        System.exit(1);
+      }
     } else {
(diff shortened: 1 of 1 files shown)


h2o-genmodel/src/main/java/hex/genmodel/tools/PrintMojo.java
```diff
@@ -44,10 +44,14 @@ public static void main(String[] args) {
           mojoPrinter = printer;
         }
       }
+      if (mojoPrinter == null) {
+        System.out.println("No supported MojoPrinter for format required found. Please make sure you are using h2o-genmodel.jar for executing this tool.");
```
Pscheidl, line 48:
```suggestion
        System.out.println("No supported MojoPrinter for the format required found. Please make sure you are using h2o-genmodel.jar for executing this tool.");
```

koniecsveta: thx 👍

PR #4531, opened 2020-04-17
PUBDEV-7408: Add progress bar during Shap Contributions calculations

michalkurka:


h2o-core/src/main/java/water/api/ModelMetricsHandler.java
```diff
@@ -352,50 +352,52 @@ public JobV3 predictAsync(int version, final ModelMetricsListSchemaV3 s) {
...
+      throw new H2OIllegalArgumentException("deviances", "not supported for async", s.deviances_frame);
 
     final ModelMetricsList parms = s.createAndFillImpl();
-
-    //predict2 does not return modelmetrics, so cannot handle deeplearning: reconstruction_error (anomaly) or GLRM: reconstruct and archetypes
-    //predict2 can handle deeplearning: deepfeatures and predict
-
-    if (s.deep_features_hidden_layer > 0 || s.deep_features_hidden_layer_name != null) {
+    
+    long workAmount = -1;
```
honzasterba, line 360: why not set the default to parms._frame.anyVec().nChunks() and get rid of the tertiary op at the end?

michalkurka: :) no idea, thanks - fixed

PR #4532, opened 2020-04-17
[PUBDEV-7465] add option to run xgboost on secured cluster

honzasterba:


tomkraljevic:
cc @bilcus to look at the implementation.

h2o-core/src/main/java/water/H2O.java
```diff
@@ -700,6 +703,9 @@ else if (s.matches("internal_security_conf")) {
       else if (s.matches("internal_security_conf_rel_paths")) {
         trgt.internal_security_conf_rel_paths = true;
       }
+      else if (s.matches("allow_insecure_xgboost")) {
+        trgt.allow_insecure_xgboost = true;
```
Pscheidl, line 707: Should we print a warning to the console ?

PR #4533, opened 2020-04-19
[build] extract common test code into a separate module

honzasterba:


Pscheidl:
Damn, this breaks our tests: `/home/jenkins/slave_dir_from_mr-0xc1/workspace/h2o-3-pipeline_PR-4533@2/java-8-smoke/h2o-3/h2o-core/src/test/java/water/webserver/EchoServletProvider.java:8: error: package javax.servlet.http does not exist` 

http://mr-0xc1:8080/blue/organizations/jenkins/h2o-3-pipeline/detail/PR-4533/2/pipeline

We need to find other way around this.

honzasterba:
@Pscheidl I think I found the cause
`  testCompile project(path: ":h2o-core", configuration: "testArchives")`
this adds another JAR which also contains the servlet API, I think this is inherently the wrong thing to do and we should get rid of it - we should move the shared test code for h2o-core/src/main/test into something like h2o-test-support/src/main and use this as a testCompile dependency in place of the testArchives dependency

Pscheidl:
Interesting. In H2O-CORE, `testCompileOnly "javax.servlet:servlet-api:2.5"` is present, which means it should not be propagated into h2o-algos or any other depending module, right ?

Isn't the problem in IntelliJ's interpretation of testCompile project(path: ":h2o-core", configuration: "testArchives") ?

Pscheidl:
After some investigation, I think this is IntelliJ's misinterpretation and we should file a bug there. The correct solution is changing the scope of core-test to "Test" and not "Compile".

Screenshot from 2020-04-20 09-52-06

But it's just my best guess. This resolves all the problems.

gradle/cp.gradle
```diff
@@ -1,7 +1,7 @@
 task cpLibs(type:Sync) {
```
honzasterba, line 1: I am not really sure what is this needed for, any clues?

h2o-algos/build.gradle
```diff
@@ -47,13 +47,10 @@ dependencies {
...
 
-
   // Test dependencies only
+  testCompile project(":h2o-test-support")
   testCompile project(":h2o-genmodel-ext-jgrapht")
-  testCompile "junit:junit:${junitVersion}"
-  testCompile project(path: ":h2o-core", configuration: "testArchives")
-  testCompileOnly 'javax.servlet:servlet-api:2.5'
-  testRuntimeOnly project(":${defaultWebserverModule}")
+  testCompile project(":${defaultWebserverModule}")
```
honzasterba, line 53: so far this seems to be the only good solution for intellij

honzasterba:
@michalkurka please elaborate, which change exactly breaks what use-case?

PR #4534, opened 2020-04-20
BACKPORT rel-yule [PUBDEV-7465] add option to run xgboost on secured cluster (#4532)

honzasterba:


h2o-hadoop-common/tests/python/pyunit_xgboost_smoke.py
```diff
@@ -0,0 +1,24 @@
+#! /usr/env/python
+import sys
+import os
+sys.path.insert(1, os.path.join("../../../h2o-py"))
+from tests import pyunit_utils
+import h2o
+from h2o.estimators import H2OXGBoostEstimator
+
+def xgb_smoke():
+  df = h2o.import_file(path=pyunit_utils.locate("smalldata/logreg/prostate.csv"))
```
michalkurka, line 10: This check is fine now because the check on the backend is global, it ignores whether the file is just on single node or not.

I am not sure if it is an issue for the future. The user could say that with build_tree_one_node there is no need to run anything multinode and wonder why we still complain about internode communication. Maybe the check on  could change. But fine at least for now.

PR #4537, opened 2020-04-20
PUBDEV-7426 Grid threshold criteria

maurever:
JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7426


h2o-core/src/main/java/hex/AUC2.java
```diff
@@ -64,7 +64,12 @@
         double mcc = (tp*tn - fp*fn);
         if (mcc == 0) return 0;
         mcc /= Math.sqrt((tp+fp)*(tp+fn)*(tn+fp)*(tn+fn));
-        assert(Math.abs(mcc)<=1.) : tp + " " + fp + " " + fn + " " + tn;
+        // due tp and tn are double; the MCC could be slightly higher than 1. for example 1.000000000000002
+        double eps = 1e-10;
+        assert(Math.abs(mcc) <= 1. + eps) : tp + " " + fp + " " + fn + " " + tn;
+        if(mcc > 1){
```
michalkurka, line 70: should we handle the negative case as well?

h2o-core/src/test/java/hex/AUC2Test.java
```diff
@@ -31,4 +31,30 @@ public void checkRecallValidity() throws Exception {
...
+    AUC2.AUCBuilder bldr = new AUC2.AUCBuilder(2);
+    bldr._n = 2;
+    System.arraycopy(tps, 0, bldr._tps, 0, tps.length);
+    System.arraycopy(fps, 0, bldr._fps, 0, fps.length);
+    AUC2 auc2 = new AUC2(bldr);
+    // index = 0, because AUCBuilder switch the order of array
+    double mcc = AUC2.ThresholdCriterion.absolute_mcc.exec(auc2, 0);
+    double exact_mcc = (tp*tn - fp*fn);
+    exact_mcc /= Math.sqrt((tp+fp)*(tp+fn)*(tn+fp)*(tn+fn));
+    assert mcc == 1: "It should return MCC = 1, but it returns "+ mcc;
```
michalkurka, line 55: please use assertEquals

maurever: Why do you suggest to use assertEquals? Because of the delta? The result should be exactly 1.0, nothing else. The delta should be 0 in assertEquals and, it does not make other sense to me to prefer assertEquals.

h2o-core/src/main/java/hex/AUC2.java
maurever: Sure.

h2o-core/src/test/java/hex/AUC2Test.java
michalkurka: Assertions can be disabled. In our test framework, we enable them to always be on but in your IDE you have no such guarantee. In my IntelliJ `assert 1 == 0` will pass by default.

Another big reason is reporting - if the assertion fails you want to see the value, not just that it fails.

maurever: Ok, I will fix it! Thank you for your explanation.

PR #4541, opened 2020-04-21
[build] fix running unit tests in IDEA

honzasterba:
- uprade servlet-api dep to 3.0 works around the issue


Pscheidl:
I wonder why. Signature of the package ?

Pscheidl:
It is safe to upgrade, if both Jettys use them. JavaEE standards are made to be backwards compatible, so there is nothing removed. This is safe operation from my POV.

honzasterba:
This is still IDEA bug workaround, IDEA is putting both API jar and jetty-api-jar on CP and one of them is probably not signed or something like that

PR #4545, opened 2020-04-23
PUBDEV-7446 - Throw error when https is true on h2o.init() while starting local cluster

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7446

Only when the cluster does not exist and an attempt is made to locally start the cluster, throw an error if HTTPS is set to true.

### DIFF
--- a/h2o-py/h2o/h2o.py
+++ b/h2o-py/h2o/h2o.py
@@ -294,6 +294,8 @@ def init(url=None, ip=None, port=None, name=None, https=None, cacert=None, insec
         if not start_h2o: raise
         if ip and not (ip == "localhost" or ip == "127.0.0.1"):
             raise H2OConnectionError('Can only start H2O launcher if IP address is localhost.')
+        if https:
+            raise H2OConnectionError('Unable to start local server with https enabled. Consider disabling https.')
         hs = H2OLocalServer.start(nthreads=nthreads, enable_assertions=enable_assertions, max_mem_size=mmax,
                                   min_mem_size=mmin, ice_root=ice_root, log_dir=log_dir, log_level=log_level,
                                   max_log_file_size=max_log_file_size, port=port, name=name,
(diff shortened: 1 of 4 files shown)


h2o-py/h2o/h2o.py
```diff
@@ -294,6 +294,8 @@ def get_mem_size(mmint, mmgb):
         if not start_h2o: raise
         if ip and not (ip == "localhost" or ip == "127.0.0.1"):
             raise H2OConnectionError('Can only start H2O launcher if IP address is localhost.')
+        if https:
+            raise H2OConnectionError('Unable to start local server with https enabled. Consider disabling https.')
```
michalkurka, line 298: If someone needs HTTPS they can start H2O jar manually. I would provide a link to the documentation that shows how to do that.

Pscheidl: Roger.

Pscheidl: Done.

h2o-py/h2o/h2o.py
```diff
@@ -294,6 +294,9 @@ def get_mem_size(mmint, mmgb):
         if not start_h2o: raise
         if ip and not (ip == "localhost" or ip == "127.0.0.1"):
             raise H2OConnectionError('Can only start H2O launcher if IP address is localhost.')
+        if https:
+            raise H2OConnectionError('Unable to start local server with https enabled. Consider disabling https or'
```
honzasterba, line 298: I don't line the wording of the error message. To me it suggest that we tried to start but were unable because of some internal issue/bug not user/config error. But I am having hard time coming up with better wording. Please help @angela0xdata

neemah2o: How about "Starting local server is not available with https enabled"?

honzasterba: @neemah2o love it!

Pscheidl: Great, I'll make it so.

Pscheidl: Done. Thanks for the suggestions.

PR #4548, opened 2020-04-23
PUBDEV-7460: Added pyunit test to double check on various modelmetrics.

wendycwong:
The work in this PR is for JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/?filter=myopenissues

Apart from verifying that all the associated metrics work for GAM, I have also done the following:
1. Fixed bug in adaptedTestforTrain.
2.  Fixed standardized coefficients to be sorted in decreasing order of magnitudes.
3. Added support/test for validation frame.
4. Tests are added in Python and Java to make sure that my addition works.

I was not able to add the cross-validation in here but has opened an JIRA to make sure this one will make be installed.

### DIFF
--- a/h2o-algos/src/main/java/hex/gam/GAM.java
+++ b/h2o-algos/src/main/java/hex/gam/GAM.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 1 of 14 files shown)


Pscheidl:
The MOJO compatibility issues in tests are suspicious.

PR #4551, opened 2020-04-27
[PUBDEV-7484] test for token refresh working in supported scenarios

honzasterba:
- add tests for jenkins

### DIFF
--- a/h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
+++ b/h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
@@ -1,0 +1,45 @@
+package water.hive;
+
+import org.apache.hadoop.io.Text;
+import org.apache.hadoop.mapreduce.security.token.delegation.DelegationTokenIdentifier;
+import org.apache.hadoop.security.token.Token;
+import org.apache.hadoop.security.token.delegation.AbstractDelegationTokenIdentifier;
+
+import java.io.ByteArrayInputStream;
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.util.Date;
+
+public class DelegationTokenPrinter {
+
+    /**
+     * The actual class of the token might not be available to us. Use this to read the token data.
+     */
+    private static class PrintingDelegationTokenIdentifier extends AbstractDelegationTokenIdentifier {
+        public Text getKind() {
+            return new Text("PRINT");
+        }
+    }
+    
+    public static void printToken(String tokenString) {
+      try {
+        Token<DelegationTokenIdentifier> token = new Token<>();
+        token.decodeFromUrlString(tokenString);
+        AbstractDelegationTokenIdentifier identifier = new PrintingDelegationTokenIdentifier();
+        identifier.readFields(new DataInputStream(new ByteArrayInputStream(token.getIdentifier())));
+        System.out.println(
+            "token.kind: " + token.getKind() + ", " + 
+            "token.service: " + token.getService() + ", " +
+            "id.owner: " + identifier.getOwner() + ", " +
+            "id.renewer: " + identifier.getRenewer() + ", " +
+            "id.realUser: " + identifier.getRealUser() + ", " +
+            "id.issueDate: " + identifier.getIssueDate() + " (" + new Date(identifier.getIssueDate()) + "), " +
+            "id.maxDate: " + identifier.getMaxDate() + " (" + new Date(identifier.getMaxDate()) + "), " +
+            "id.validity: " + (identifier.getMaxDate() - System.currentTimeMillis()) / 3600_000 + " hours");
+      } catch (IOException e) {
+        System.out.println("Failed to decode token.");
+        e.printStackTrace();
+      }
+    }
+
+}
--- a/h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
+++ b/h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
@@ -21,11 +21,12 @@ public class DelegationTokenPrinter {
         }
     }
     
-    public static void printToken(String tokenString) {
+    public static void printToken(final String tokenString) {
+        Objects.requireNonNull(tokenString);
       try {
-        Token<DelegationTokenIdentifier> token = new Token<>();
+        final Token<DelegationTokenIdentifier> token = new Token<>();
         token.decodeFromUrlString(tokenString);
-        AbstractDelegationTokenIdentifier identifier = new PrintingDelegationTokenIdentifier();
+        final AbstractDelegationTokenIdentifier identifier = new PrintingDelegationTokenIdentifier();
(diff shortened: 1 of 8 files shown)


honzasterba:
This is just a proposal for now, the testing part we will definitely keep, we need to have a discussion whether we also need to keep the driver functionality.

honzasterba:
for now only improve testing and logging (print token info)

h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
```diff
@@ -0,0 +1,45 @@
...
+        public Text getKind() {
+            return new Text("PRINT");
+        }
+    }
+    
+    public static void printToken(String tokenString) {
+      try {
+        Token<DelegationTokenIdentifier> token = new Token<>();
+        token.decodeFromUrlString(tokenString);
+        AbstractDelegationTokenIdentifier identifier = new PrintingDelegationTokenIdentifier();
```
Pscheidl, line 28:
```suggestion
        final AbstractDelegationTokenIdentifier identifier = new PrintingDelegationTokenIdentifier();
```

Nitpick...

h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
```diff
@@ -0,0 +1,45 @@
...
+            "token.kind: " + token.getKind() + ", " + 
+            "token.service: " + token.getService() + ", " +
+            "id.owner: " + identifier.getOwner() + ", " +
+            "id.renewer: " + identifier.getRenewer() + ", " +
+            "id.realUser: " + identifier.getRealUser() + ", " +
+            "id.issueDate: " + identifier.getIssueDate() + " (" + new Date(identifier.getIssueDate()) + "), " +
+            "id.maxDate: " + identifier.getMaxDate() + " (" + new Date(identifier.getMaxDate()) + "), " +
+            "id.validity: " + (identifier.getMaxDate() - System.currentTimeMillis()) / 3600_000 + " hours");
+      } catch (IOException e) {
+        System.out.println("Failed to decode token.");
```
Pscheidl, line 40: Looks like our Logger is available in this package. Why don't we use it ?

h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * The actual class of the token might not be available to us. Use this to read the token data.
+     */
+    private static class PrintingDelegationTokenIdentifier extends AbstractDelegationTokenIdentifier {
+        public Text getKind() {
+            return new Text("PRINT");
+        }
+    }
+    
+    public static void printToken(String tokenString) {
```
Pscheidl, line 24:
```suggestion
    public static void printToken(final String tokenString) {
        Objects.requireNonNull(tokenString);
```

Nitpick

h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
```diff
@@ -0,0 +1,45 @@
...
+      try {
+        Token<DelegationTokenIdentifier> token = new Token<>();
```
Pscheidl, line 26:
```suggestion
        final Token<DelegationTokenIdentifier> token = new Token<>();
```

Nitpick

h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
honzasterba: 1) we cannot use water.Log because when running in h2odriver its not initialized and does not print to stdout
2) we cannot use Log4j because when running in mapper log4j does not print to stdout, this is better in master now and Log4j would be ideal to use here. will remember to improve

scripts/jenkins/groovy/hadoopCommands.groovy
```diff
@@ -4,7 +4,10 @@
  * @param stageConfig stage configuration to read mode and additional information from
  * @return the cmd used to start H2O in given mode
  */
-def call(final stageConfig) {
+def call(final stageConfig, final boolean getMakeTarget = false) {
+    if (getMakeTarget) {
+        return ""
```
michal-raska, line 9: q: what's the purpose of this please?

honzasterba: for some cases we want to execute another make target (to check that token refresh happended succesfully)

honzasterba: https://github.com/h2oai/h2o-3/pull/4551/files#diff-922fd441e2d3ead9c35f17e132b6daecR46

h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
```diff
@@ -0,0 +1,43 @@
...
+            return new Text("PRINT");
+        }
+    }
+
+    public static void printToken(final String tokenString) {
+        Objects.requireNonNull(tokenString);
+        try {
+            final Token<DelegationTokenIdentifier> token = new Token<>();
+            token.decodeFromUrlString(tokenString);
+            final AbstractDelegationTokenIdentifier identifier = new PrintingDelegationTokenIdentifier();
```
michalkurka, line 27: is there a reason you cannot just call `Token#decodeIdentifier`?

scripts/jenkins/Makefile.jenkins
```diff
@@ -445,6 +445,16 @@ test-kerberos-hadoop-3-hdp: test-hadoop-3-smoke-standalone test-hadoop-common-sm
...
 
+test-kerberos-verify-token-refresh: h2o_one_node
+	cat h2o_one_node | grep job | sed 's/job/application/g' > h2o_yarn_id
+	@if [ -z "`cat h2o_yarn_id`" ]; then\
+		echo "YARN Application ID not found."; \
+		exit 1; \
+	fi
+	@echo "YARN Application ID is `cat h2o_yarn_id`"
+	-yarn application -kill `cat h2o_yarn_id`
+	yarn logs -applicationId `cat h2o_yarn_id` | grep -e "TOKEN REFRESH: Updating credentials"
```
michalkurka, line 456: does it check anywhere that the pattern was found?

h2o-hive/src/main/java/water/hive/DelegationTokenPrinter.java
honzasterba: yes, for some distros/versions this call fails because the token class is not available in the jdbc driver JAR, thats why I am using my own Identifier class to load the data

scripts/jenkins/Makefile.jenkins
honzasterba: if the pattern was not found grep returns non-zero exit status which is detected by make as failure of the target

michalkurka: duh! thank you

PR #4554, opened 2020-04-27
PUBDEV-7469 - Expose leader node only on K8S

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7469

All traffic should be handled by H2O Leader node only (for reproducibility). A kubernetes-native way to achieve this are readiness probes: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ - do not mistake them with liveness probes. A container is free to change it's readiness state during its lifetime.

1. H2O Pods starts up - clustering is not yet finished. All pods mark themselves as ready - readiness probe returns 200 OK.
2. Clustering is finished, pods mark themselves as not ready to serve requests (readiness probe returns 404). With one exception, and that is the leader node.

**Why 404 ?** - could not find better fit. 404 is interpreted as "ready node not found". Suggestions are welcome. It has to be something else than 2xx.

:heavy_check_mark: README adjusted

:heavy_check_mark: Tested in our Kubernetes tests. 

:heavy_check_mark: Tested manually on OpenShift

Screenshot from 2020-04-28 10-32-55

Screenshot from 2020-04-28 10-33-01

Screenshot from 2020-04-28 10-34-16

### Distinct REST API for Kubernetes extension

There is a distinct rest API for Kubernetes extension implemented on top of extremely lightweight (few KB in size, the core is only 1 class) [nanoHTTPD](https://github.com/NanoHttpd/nanohttpd). H2O REST API starts serving requests after the clustering is finished, yet we need kubernetes-related services like probes (and possibly others in future) in place before that happens. Also, we want to prevent users to contact H2O directly and lock the cluster by running any tasks on it before the actual clustering is finished. Therefore, the Kubernetes REST API is distinct and resides on it's own port. The port is configurable and defaults to 8080.

See the `h2o-k8s/README.md` for a complete usage example.


h2o-k8s/src/main/java/water/k8s/lookup/KubernetesDnsLookup.java
```diff
@@ -108,13 +108,16 @@ private static String extractHost(final String server, final Pattern extractHost
...
-                Thread.sleep(ONE_SECOND);
             } catch (NamingException e) {
                 Log.warn(e.getMessage());
                 continue;
-            } catch (InterruptedException e) {
-                Log.err(e);
-                return Optional.empty();
+            } finally {
+                try {
+                    Thread.sleep(ONE_SECOND);
```
Pscheidl, line 116: This is a small bugfix. This makes sure the sleep is actually applied :)

h2o-k8s/build.gradle
```diff
@@ -8,6 +8,12 @@ repositories {
 
 dependencies {
     compile project(":h2o-core")
+    compile 'org.nanohttpd:nanohttpd:2.3.1'
```
mn-mikke, line 11: Is there any reason why we can't implement this feature via a Jetty servlet?

honzasterba: +1

h2o-k8s/src/main/java/water/k8s/api/KubernetesRestApi.java
```diff
@@ -0,0 +1,77 @@
...
+     * The REST API is bound to a default port of 8080, unless specified otherwise by H2O_K8S_API_PORT environment
+     * variable.
+     */
+    public KubernetesRestApi() {
+        super(getPort());
+        addMappings();
+    }
+
+    private static int getPort() {
+        final String customKubernetesPort = System.getenv(KUBERNETES_REST_API_PORT_KEY);
```
honzasterba, line 41: why not use the standard h2o property prefix?

h2o-k8s/build.gradle
Pscheidl: Explained in the description.

Pscheidl: Our jetty servlet starts a lot later and required H2O configured. Also, we don't want to let our users call the Jetty ensemble due to cloud locking issues. Also, what if the Jetty is secured by the user ? Could also happen on Kubernetes. The K8S API should stand aside.

h2o-k8s/src/main/java/water/k8s/api/KubernetesRestApi.java
Pscheidl: `H2O_K8S_API_PORT` hides under the `KUBERNETES_REST_API_PORT_KEY` static field. This is in-line with the rest of H2O K8S properties. A simple `H2O` prefix is used for K8S to save space in YAML files. It's easy and short and doesn't make heavily indented lines any longer with unnecessary prefixes like our system property prefix `public static final String SYSTEM_PROP_PREFIX = "sys.ai.h2o.";`

This style is also in-line with what K8S uses: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/

Let's compare the two styles:

```
          - name: H2O_KUBERNETES_SERVICE_DNS
            value: h2o-service.h2o-statefulset.svc.cluster.local
          - name: H2O_NODE_LOOKUP_TIMEOUT
            value: '180'
          - name: H2O_NODE_EXPECTED_COUNT
            value: '3'
          - name: H2O_K8S_API_PORT
            value: '8081'
```

SYS_AI prefix: 
```
          - name: SYS_AI_H2O_KUBERNETES_SERVICE_DNS
            value: h2o-service.h2o-statefulset.svc.cluster.local
          - name: SYS_AI_H2O_NODE_LOOKUP_TIMEOUT
            value: '180'
          - name: SYS_AI_H2O_NODE_EXPECTED_COUNT
            value: '3'
          - name: SYS_AI_H2O_K8S_API_PORT
            value: '8081'
```

From user's POV, what SYS_AI even means ? It's long and unnecessary. H2O properties are prefixed with H2O. This is simplest and most user friendly and meaningful.

Pscheidl: I'm not 100% against it, yet I don't think our K8S env vars need to follow the same long pattern as our sys props.

Pscheidl: Actually now, when I'm looking at it, I have H2O_KUBERNETES in one field and H2O_K8S in another. This must be aligned to H2O_KUBERNETES.

h2o-k8s/build.gradle
Pscheidl: For more information see `gateHandler` method of `H2OHttpViewImpl` class.

This method is waiting for
```
    ServletUtils.startRequestLifecycle();
    while (! isAcceptingRequests()) {
      try { Thread.sleep(100); }
      catch (Exception ignore) {}
    }
```

which is waiting for `H2O.startServingRestApi` call. 

In order to go start the REST API, we'd need to do the preparations for H2O core beforehand (which is possible to do) and start serving the REST API sooner. The effort would then be duplicated. Also, we'd need to filter out only certain requests (probes), otherwise H2O would start accepting requests which could lock the cluster prematurely. There are several potential deadlocks.

This abstracts away K8S REST API from H2O's API and acts as a separate service (which it is).

h2o-core/src/main/java/water/webserver/H2OHttpViewImpl.java
```diff
@@ -2,6 +2,7 @@
 
 import org.apache.commons.io.IOUtils;
 import water.ExtensionManager;
+import water.H2O;
```
michalkurka, line 5: forgotten import?

h2o-app/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
   compile project(":h2o-algos")
   compile project(":h2o-automl")
   compile project(":h2o-ext-target-encoder")
+  compile project(":h2o-k8s")
```
michalkurka, line 12: Should this really be added to h2o-app? Or should it only go to the main assembly? I would expect to see it only in assembly.

h2o-core/src/main/java/water/webserver/H2OHttpViewImpl.java
Pscheidl: Yes, thanks. Resolved in https://github.com/h2oai/h2o-3/pull/4554/commits/2e4abce3a2616c5b64bb68ca47d5762bbc7ad94f.

h2o-app/build.gradle
Pscheidl: Oh snap ! That's forgotten. It's part of the assembly of course. Removed in https://github.com/h2oai/h2o-3/pull/4554/commits/785505a19fea1cb894fc02ca48ece25836d8c062.

h2o-k8s/build.gradle
michalkurka: The different lifecycles of the endpoints and also different needs of authentication/security are IMHO good reasons to avoid relying on the embedded Jetty.

h2o-k8s/src/main/java/water/k8s/probe/KubernetesLeaderNodeProbeHandler.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public NanoHTTPD.Response.IStatus getStatus() {
+        throw new IllegalStateException(String.format("Method getMimeType should not be called on '%s'",
+                getClass().getName()));
+    }
+
+    @Override
+    public NanoHTTPD.Response get(RouterNanoHTTPD.UriResource uriResource, Map<String, String> urlParams, NanoHTTPD.IHTTPSession session) {
+        // All nodes report ready state until the clustering process is finished. Since then, only the leader node is ready.
+        if (H2O.SELF == null || H2O.SELF.isLeaderNode() || !KubernetesEmbeddedConfigProvider.isClustered()) {
```
michalkurka, line 34: H2O.SELF should theoretically never transition from a valid reference to null. However, to handle edge cases better I like to introduce a local variable to avoid possible NPEs:

H2ONode self = H2O.SELF;

self == null | self.isLeaderNode || ...

Pscheidl: Great, thanks. Fixed in https://github.com/h2oai/h2o-3/pull/4554/commits/22be4dc4fff2fbe83e14f4dc9b4de2718e30e275.

h2o-k8s/build.gradle
michalkurka: @mn-mikke @honzasterba WDYT?

mn-mikke: Ok, thanks @Pscheidl for your explanation! The library seems to be light-weight and doesn't have any other dependencies. it seems that SW/H2O assembly jar won't grow much, so I'm fine with that.

Pscheidl: It weights only a few tens of kilobytes in total.

PR #4556, opened 2020-04-28
Remove markup language from R docs

Pscheidl:
This should fix the R CMD check issues - there was a markup language in the R docs.


Pscheidl:
Had to push once again, as it seems the percentage sign was causing other problems. Works locally. Please wait for the tests and merge before release @michalkurka

michalkurka:
thank you for the fix @Pscheidl

jangorecki:
@Pscheidl yes, percentage has to be escaped with `\%`. Good reference on R manual files: https://developer.r-project.org/parseRd.pdf

PR #4559, opened 2020-04-29
PUBDEV-4916: add weights_column parameter to Stacked Ensemble

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-4916

I chose the second variant in the case when user doesn't set any weights_column in SE:

>  If all the base models use the same weights_column, then warn the user that they may want to also manually specify the weights_column in the Stacked Ensemble function.  By default, NULL means no weight column here.


tomasfryda:
Thanks @ledell for reminding me to check those tests! I had a bug there - since base models can be both models and grids and grids get expanded to the models on the backend (java) during model creation, so I have to replicate the expanding logic in clients to be able to provide the warning...

Or is there a way to send warning (not exception) from the backend to the clients so they can notify user without stopping the work?

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -78,7 +78,9 @@ public void init(boolean expensive) {
       throw new H2OIllegalArgumentException("Setting \"distribution\" to StackedEnsemble is unsupported. Please set it in \"metalearner_parameters\".");
     }
 
-    checkFoldColumnPresent(_parms._metalearner_fold_column, train(), valid(), _parms.blending());
+    checkColumnPresent("fold", _parms._metalearner_fold_column, train(), valid(), _parms.blending());
```
sebhrusen, line 81: is it not a good occasion to deprecate those `metalearner_fold_column`, `metalearner_nfolds`, `metalearner_fold_assignment` parameters here if we start accepting the param values that are set directly on the SE model?

It would really simplify API and avoid some confusions:
For example, with your current changes, if user does the following
```
H2OStackedEnsembleEstimator(fold_column="fcol", weights_column="wcol")
```
then it will use the weights correctly but ignore the fold column (actually raise an error in Python, but when used directly in Java, it will just ignore it).
so that user needs to do:
```
H2OStackedEnsembleEstimator(metalearner_fold_column="fcol", weights_column="wcol")
```
which looks really inconsistent.
I never really understood this `_metalearner_xxx` naming for those CV specific arguments and we should go towards SE being more consistent and behaving more like other algos.

Note: we have utilities to deprecate properties in Py. For arguments (Py and R) we could emit a standard warning when they're passed.

Idea of deprecation here also involves creation of ticket to remove completely support for those 3 properties in let's say version 3.36.
@ledell , @tomasfryda any thoughts?

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -104,22 +106,36 @@ private void validateAndExpandBaseModels() {
...
+
+    for (int i = 0; i < _parms._base_models.length; i++) {
+      Model baseModel = DKV.getGet(_parms._base_models[i]);
+
+      if ((i == 0) && (_parms._offset_column == null))
+        _parms._offset_column = baseModel._parms._offset_column;
+
+      if (((_parms._offset_column == null) && (baseModel._parms._offset_column != null)) ||
+              ((_parms._offset_column != null) && (!_parms._offset_column.equals(baseModel._parms._offset_column))))
+        throw new IllegalArgumentException("All base models have to have the same offset_column!");
```
sebhrusen, line 119: "must have" sounds nicer than "have to have", doesn't it? Not an expert though.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -104,22 +106,36 @@ private void validateAndExpandBaseModels() {
...
+
+
+    for (int i = 0; i < _parms._base_models.length; i++) {
+      Model baseModel = DKV.getGet(_parms._base_models[i]);
+
+      if ((i == 0) && (_parms._offset_column == null))
+        _parms._offset_column = baseModel._parms._offset_column;
+
+      if (((_parms._offset_column == null) && (baseModel._parms._offset_column != null)) ||
+              ((_parms._offset_column != null) && (!_parms._offset_column.equals(baseModel._parms._offset_column))))
```
sebhrusen, line 118:
```
if (!Objects.equals((_parms._offset_column, baseModel._parms._offset_column))
```
?

h2o-bindings/bin/custom/R/gen_stackedensemble.py
```diff
@@ -35,6 +35,32 @@ def update_param(name, param):
...
+    reference_weights_column <- weights_columns[[1]]
+    if (!is.null(reference_weights_column)) {
+      if (all(sapply(weights_columns, function(wc) !is.null(wc) && wc == reference_weights_column))) {
+        warning(paste0("All base models use weights_column=\\\"", reference_weights_column,
+                       "\\\" but Stacked Ensemble does not. If you want to use the same ",
+                       "weights_column for the meta learner, please specify it as an argument ",
+                       "in the h2o.stackedEnsemble call."))
+      }
+    }
+}
```
sebhrusen, line 63: Non-trivial validation logic should be done in backend instead of being replicated in each client.
Especially when like here, it requires loading all base modelson the client.
Also, this kind of validation won't work in Flow, and would need to be replicated in SparklingWater as well.

Instead, you should be able to use `ModelBuilder.warn` when doing the validation in `StackedEnsemble`.
This populates `  public ValidationMessageV3 messages[];` in `ModelBuilderV3`, which is then accessible through REST API.
For example, in Python, those messages should be accessible in `estimator_base.py`, method `_train`:
```
        model_builder_json = h2o.api("POST /%d/ModelBuilders/%s" % (rest_ver, self.algo), data=parms)
```
I've often been frustrated that we don't have this mechanism for AutoML, but it does exist for other algos, we should just capture the messages there and emit the warnings on the client.

h2o-bindings/bin/custom/python/gen_stackedensemble.py
```diff
@@ -111,6 +111,29 @@ def stacking_strategy(self):
...
+                        raise TypeError("Unsupported type of base model \"{}\". Should be either a model or a grid.".format(base_model))
+
+            weight_columns = [wc.get("column_name") if wc else wc for wc in weight_columns]
+            weight_column_ref = weight_columns[0]
+            if weight_column_ref is not None and all((wc == weight_column_ref for wc in weight_columns)):
+                warnings.warn(("All base models use weights_column=\"{}\" but Stacked Ensemble does not. "
+                               "If you want to use the same weights_column for the meta learner, "
+                               "please specify it as an argument in StackedEnsemble's constructor.").format(weight_column_ref))
+
+
```
sebhrusen, line 136: same comment as above (R)

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
tomasfryda: I agree that we should keep things consistent with the rest of the algos, that includes deprecating `metalearning_fold_column` etc.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
tomasfryda: Thank you!

ledell:
@tomasfryda @sebhrusen Anyone know the status of this? Thanks!

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
tomasfryda: @ledell I'd say that the status of this PR is dependent on the decision if we will want to deprecate `metalearner_*` params in this PR.

sebhrusen: it doesn't have to be in this PR, but it would be great if both are done in the same release because all those parameters (sometimes with, sometimes without prefix) are getting confusing.
Deprecations should be done only for `3.32.0.1`, so do we really need this in `zahradnik`?

sebhrusen: I'd suggest to consider merging this to `master` and add a follow-up ticket for the `metalearner` fold params deprecations and ensure it is done before `3.32`.

tomasfryda: I agree. I created a JIRA[1] for deprecation of the `metalearner_*` params. So if @ledell is ok with this being in `master` instead of `rel-zahradnik`, I will merge it.


[1] https://0xdata.atlassian.net/browse/PUBDEV-7675

PR #4563, opened 2020-04-30
[PUBDEV-7496] use java predictor for xgb var-imp

honzasterba:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -230,30 +230,6 @@ public XGBoostModel(Key<XGBoostModel> selfKey, XGBoostParameters parms, XGBoostO
     setDataInfoToOutput(dinfo);
     model_info = new XGBoostModelInfo(parms, dinfo);
   }
-
-  // useful for debugging
-  @SuppressWarnings("unused")
-  public void dump(String format) {
```
honzasterba, line 236: this code is already duplicated in BoosterDump.java

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostVarImpComparisonTest.java
```diff
@@ -0,0 +1,112 @@
...
+            XGBoostModel nativeModel = new XGBoost(parms).trainModel().get();
+            Scope.track_generic(nativeModel);
+
+            System.setProperty("sys.ai.h2o.xgboost.predict.native.enable", "false");
+            XGBoostModel javaModel = new XGBoost(parms).trainModel().get();
+            Scope.track_generic(javaModel);
+
+            if (nativeModel._output._varimp != javaModel._output._varimp) {
+                assertArrayEquals(nativeModel._output._varimp._names, javaModel._output._varimp._names);
+                assertArrayEquals(nativeModel._output._varimp._varimp, javaModel._output._varimp._varimp, 1e-2f);
```
michalkurka, line 101: Why do we need a large tolerance? What differs in their math compared to ours? float vs doubles?

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostVarImpComparisonTest.java
```diff
@@ -0,0 +1,112 @@
...
+    public String distribution;
+
+    @Parameterized.Parameter(2)
+    public String response;
+
+    @Test
+    public void testImportancesAreSame() {
+        Scope.enter();
+        try {
+            Frame tfr = Scope.track(parse_test_file("./smalldata/prostate/prostate.csv"));
```
michalkurka, line 69: there already is a version of the prostate dataset that is split to test and train, it is imho better for reproducibility of results

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/predict/XGBoostJavaVariableImportance.java
```diff
@@ -0,0 +1,58 @@
...
+    
+    public XGBoostJavaVariableImportance(XGBoostModelInfo modelInfo) {
+        predictor = PredictorFactory.makePredictor(modelInfo._boosterBytes, false);
+        dataInfo = modelInfo.dataInfo();
+    }
+
+    @Override
+    public Map<String, FeatureScore> getFeatureScores() {
+        Map<String, FeatureScore> featureScore = new HashMap<>();
+        if (!(predictor.getBooster() instanceof GBTree)) {
```
michalkurka, line 30: gblinear doesn't have feature scores?

michalkurka:
@honzasterba this is really nice, we could also get rid of the feature map file eventually

I have 2 questions:
- does gblinear not have feature importance?
- does parsing of the model to calculate variable importance affect performance? for large models? 

* how much work would it be to parse only the new trees (this would be generally helpful, not just in context of this task)

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostVarImpComparisonTest.java
honzasterba: float only has 6-7 digits for precision for numbers of 1234.56 there can be a difference on the 0.01 scale probably because of order of additions

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/predict/XGBoostJavaVariableImportance.java
honzasterba: no

michalkurka: oh, okay - thank you

honzasterba:
The old impl was based on parsing a test-based tree dump of the booster (the dump is ment to be human readable so the parsing is very cumbersome), thats why there is no feature importnace for gblinear (it has no trees).

I would suspect this would be much faster, but will measure.
Not sure how we would compute it only for new trees since we always load the full booster.

michalkurka:
> Not sure how we would compute it only for new trees since we always load the full booster.
yes, right now we do that, but we could fast forward in the byte array - just go to the location where the new trees start and go from there (it is not hard to make the jump - the code that parses the scoring trees could be used for that)

honzasterba:
I guess for very large boosters there could be a benefit to skipping some trees. I think adding on option to Predictor to skip some trees would be quite trivial.

PR #4566, opened 2020-05-01
PUBDEV-7491: Enable using quantile and other distributions in SE

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7491

### Proposal of solution / description
This PR should enable user to use different distributions/families in SE (metalearner).

As far as I can tell, the current solution infers task (Binomial/Multinomial Classification, Regression) from the first base model. It is possible to set "distribution" in `metalearner_params`, however, doing the same thing for "family" is not possible since GLM/Auto metalearner overrides it (without warning) based on the `ModelCategory`.

In this PR, I am trying to make SE behave the way I would expect it to behave as a user.
This means distribution/family resolution in the following order:

1. If user specifies "distribution" in `metalearner_params` use that (even if it means that metalearner fails with "Unsupported distribution"; user should know about it and change it manually).
1. If there is no "distribution" specification in `metalearner_params`, try to infer it from first base model (also check that all other base models have the same distribution + its params (see TODO))
1. If the "distribution" from the first base model isn't compatible with the metalearner, fallback to the current solution using default distribution per task (Binomial/Multinomial Classification, Regression).

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
+++ b/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
@@ -7,8 +7,12 @@ import hex.ensemble.StackedEnsembleModel.StackedEnsembleParameters;
 import water.DKV;
 import water.Job;
 import water.Key;
+import water.exceptions.H2OIllegalArgumentException;
 import water.fvec.Frame;
 import water.util.Log;
+
+import java.lang.reflect.Field;
+import java.util.Set;
 
 public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Model<M, P, ?>, P extends Model.Parameters> {
 
@@ -37,6 +41,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
   protected P _metalearner_parameters;
   protected boolean _hasMetalearnerParams;
   protected long _metalearnerSeed;
+  protected Set<String> _metalearnerParamsUserOverride;
 
   void init(Frame levelOneTrainingFrame,
             Frame levelOneValidationFrame,
@@ -47,7 +52,8 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
             Job metalearnerJob,
             StackedEnsembleParameters parms,
             boolean hasMetalearnerParams,
-            long metalearnerSeed) {
+            long metalearnerSeed,
+            Set<String> metalearnerParamsUserOverride) {
 
     _levelOneTrainingFrame = levelOneTrainingFrame;
     _levelOneValidationFrame = levelOneValidationFrame;
@@ -59,7 +65,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
     _parms = parms;
     _hasMetalearnerParams = hasMetalearnerParams;
     _metalearnerSeed = metalearnerSeed;
-
+    _metalearnerParamsUserOverride = metalearnerParamsUserOverride;
   }
 
   void compute() {
@@ -99,6 +105,22 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
   }
 
   abstract B createBuilder();
+
+  /**
+   * This method sets parms.field=value if it wasn't explicitly set by a user.
+   * @param parms
+   * @param field
+   * @param value
+   * @param <T>
+   */
+  protected <T> void setOverridableParm(P parms, String field, T value) {
+    if (_metalearnerParamsUserOverride.contains(field)) return;
+    try {
+      parms.getClass().getField("_".concat(field)).set(parms, value);
+    } catch (IllegalAccessException | NoSuchFieldException e) {
+      throw new H2OIllegalArgumentException("Unable to set field \"".concat(field).concat("\"."));
+    }
+  }
 
   protected void setCommonParams(P parms) {
     if (parms._seed == -1) { //use _metalearnerSeed only as legacy fallback if not set on metalearner_parameters 
--- a/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
+++ b/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
@@ -73,8 +73,8 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
       setCommonParams(builder._parms);
       setCrossValidationParams(builder._parms);
       setCustomParams(builder._parms);
-
       validateParams(builder._parms);
+      prepareModel(builder._parms);
 
       builder.init(false);
       Job<M> j = builder.trainModel();
@@ -129,6 +129,8 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
   protected void setCustomParams(P parms) { }
 
   protected void validateParams(P parms) { }
+
+  protected void prepareModel(P parms) {}
 
   protected void cleanup() {
     if (!_parms._keep_base_model_predictions) {
--- a/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
+++ b/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
@@ -7,12 +7,8 @@ import hex.ensemble.StackedEnsembleModel.StackedEnsembleParameters;
 import water.DKV;
 import water.Job;
 import water.Key;
-import water.exceptions.H2OIllegalArgumentException;
 import water.fvec.Frame;
 import water.util.Log;
-
-import java.lang.reflect.Field;
-import java.util.Set;
 
 public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Model<M, P, ?>, P extends Model.Parameters> {
 
@@ -41,7 +37,6 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
   protected P _metalearner_parameters;
   protected boolean _hasMetalearnerParams;
   protected long _metalearnerSeed;
-  protected Set<String> _metalearnerParamsUserOverride;
 
   void init(Frame levelOneTrainingFrame,
             Frame levelOneValidationFrame,
@@ -52,8 +47,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
             Job metalearnerJob,
             StackedEnsembleParameters parms,
             boolean hasMetalearnerParams,
-            long metalearnerSeed,
-            Set<String> metalearnerParamsUserOverride) {
+            long metalearnerSeed) {
 
     _levelOneTrainingFrame = levelOneTrainingFrame;
     _levelOneValidationFrame = levelOneValidationFrame;
@@ -65,7 +59,7 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
     _parms = parms;
     _hasMetalearnerParams = hasMetalearnerParams;
     _metalearnerSeed = metalearnerSeed;
-    _metalearnerParamsUserOverride = metalearnerParamsUserOverride;
+
   }
 
   void compute() {
@@ -79,6 +73,8 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
       setCommonParams(builder._parms);
       setCrossValidationParams(builder._parms);
       setCustomParams(builder._parms);
+
+      validateParams(builder._parms);
 
       builder.init(false);
       Job<M> j = builder.trainModel();
@@ -106,22 +102,6 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
 
   abstract B createBuilder();
 
-  /**
-   * This method sets parms.field=value if it wasn't explicitly set by a user.
-   * @param parms
-   * @param field
-   * @param value
-   * @param <T>
-   */
-  protected <T> void setOverridableParm(P parms, String field, T value) {
-    if (_metalearnerParamsUserOverride.contains(field)) return;
-    try {
-      parms.getClass().getField("_".concat(field)).set(parms, value);
-    } catch (IllegalAccessException | NoSuchFieldException e) {
-      throw new H2OIllegalArgumentException("Unable to set field \"".concat(field).concat("\"."));
-    }
-  }
-
   protected void setCommonParams(P parms) {
     if (parms._seed == -1) { //use _metalearnerSeed only as legacy fallback if not set on metalearner_parameters 
       parms._seed = _metalearnerSeed;
@@ -147,6 +127,8 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
   }
 
   protected void setCustomParams(P parms) { }
+
+  protected void validateParams(P parms) { }
 
   protected void cleanup() {
     if (!_parms._keep_base_model_predictions) {
--- a/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
+++ b/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
@@ -73,8 +73,8 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
       setCommonParams(builder._parms);
       setCrossValidationParams(builder._parms);
       setCustomParams(builder._parms);
+
       validateParams(builder._parms);
-      prepareModel(builder._parms);
 
       builder.init(false);
       Job<M> j = builder.trainModel();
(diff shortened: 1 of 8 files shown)


ledell:
@tomasfryda This looks good so far.  You might want to check in with @sebhrusen about this because he recently suggested that we add `distribution` as an alias to `family` in GLM, so maybe that simplifies things.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -66,6 +67,8 @@ public StackedEnsembleModel(Key selfKey, StackedEnsembleParameters parms, Stacke
     public Metalearner.Algorithm _metalearner_algorithm = Metalearner.Algorithm.AUTO;
     public String _metalearner_params = new String(); //used for clients code-gen only.
     public Model.Parameters _metalearner_parameters;
+    // To keep track of what actually a user set in metalearner_parameters
+    public HashSet<String> _metalearner_parameters_user_override=new HashSet<>();
```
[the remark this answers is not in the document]
sebhrusen: Also, let's try to completely avoid this new field. It really looks like a hacky way to leak nasty logic from schema `StackedEnsembleV99` until the metalearner itself...
From a Java/Scala user (e.g. SparklingWater), this doesn't make sense as those users directly instantiate the parameter object.
So what you need is to embrace the semantic of `Distribution.AUTO` to its full extent.

tomasfryda:
@sebhrusen thank you for the review! In general, I agree with everything you wrote.

About the (2) allow user to override any metalearner param set internally in AUTO mode. The motivation was to allow setting metalearner param for "GLM", the part for AUTO was just extension of that. (But I think we should at least warn user that we will ignore the meta learner params if (s)he sets them for `AUTO`. ) The problem I saw there is that even if a user specified some family for meta learner and set meta learner to GLM it used either binomial, multinomial or Gaussian family ignoring the family and link that the user set...   

In my opinion, the biggest complexity lies in having both  the family and link, and distribution. The problem is that there are families (+ links) that don't have distribution and I didn't find a satisfying solution. That also makes it, I think, impossible to use the thing you proposed - checking if  `_metalearner_parameters._distribution != AUTO`.

Do you have any suggestions about how to deal with the family/link and distribution being set by user? (AFAIK there is no `Family.AUTO`)

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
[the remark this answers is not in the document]
tomasfryda: I got rid of the `_metalearner_parameters_user_override` but to enable the same behaviour I set `_family` to null by default  - meaning it wasn't set by user, can be inferred. When there will be `Family.AUTO` I would use that instead.

h2o-algos/src/main/java/hex/ensemble/Metalearner.java
```diff
@@ -100,6 +106,22 @@ void compute() {
...
+   * This method sets parms.field=value if it wasn't explicitly set by a user.
+   * @param parms
+   * @param field
+   * @param value
+   * @param <T>
+   */
+  protected <T> void setOverridableParm(P parms, String field, T value) {
+    if (_metalearnerParamsUserOverride.contains(field)) return;
+    try {
+      parms.getClass().getField("_".concat(field)).set(parms, value);
```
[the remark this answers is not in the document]
tomasfryda: Thanks for the tip!

tomasfryda:
I modified it in a way that during the `checkAndInheritModelProperties` I set directly the `_metalearner_parameters`. To detect if user set distribution I check whether it's set to AUTO as you suggested. For family parameter, I didn't come up with anything better than a hack to set family to `null` during the parameters initialisation so if user doesn't set them then it goes through the inference logic. In close future, I would use `Family.AUTO` instead.

Remaining problem I have is how to properly detect that a distribution is supported by a given metalearner. For now I have created `MetalearnerWithDistribution` class where I white list distributions that are allowed by that model and during `Metalearner.compute` it checks whether the distribution is OK; otherwise it prints a warning and coerces it. 

The coercion doesn't use model category but `nclasses()` instead - the problem is when all base models have quasi binomial distribution and meta learner doesn't support it, it would fail on Bernoulli  distribution requires categorical response and numeric was given. I don't like that solution but as I see it there are 3 options and I can't decide:

1. fail due to the inference
2. convert response to categorical and use Bernoulli
3. use `nclasses()` to determine whether it's regression, binomial or multinomial classification and use appropriate distribution - which in this case logs an warning and uses Gaussian distribution

What do you think @sebhrusen and @ledell ?

tomasfryda:
> But what is the reason to propagate this if you don't expect user to set it to SE._parms first instead of metalearner_parameters? I find this really really confusing, and I'm afraid it will confuse us for a long time and be the cause of if not bugs, at least unexpected behaviours...
I agree that the fact that SE offers 2 sets of parameters is annoying, but I'd rather modify the StackedEnsemble.init where the validation is done, and raise warnings or exception each time params intended for the metalearner are set at this level instead of metalearner level.

Agreed, previously `_distribution` was used to save the inferred distribution and then during metalearner creation it was used. Since now I use `metalearner_parameters._distribution` instead I don't use `_distribution` at all. Setting `_distribution` from R/Python client fails due to not being in `StackedEnsembleV99.StackedEnsembleParametersV99.fields`. I also added check to throw an exception if `_distribution != AUTO`, to let java/scala users know in case they set it there.

h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
```diff
@@ -153,6 +152,9 @@ public StackedEnsembleParametersV99 fillFromImpl(StackedEnsembleModel.StackedEns
           case glm:
             paramsSchema = new GLMV3.GLMParametersV3();
             params = new GLMModel.GLMParameters();
+            // FIXME: This is here because there is no Family.AUTO. It enables us to know if the user specified family or not.
+            // FIXME: Family.AUTO will be implemented in https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7444
```
sebhrusen, line 156: please ensure you're subscribed to this Jira ticket

h2o-algos/src/main/java/hex/ensemble/Metalearners.java
```diff
@@ -100,23 +103,91 @@ protected SimpleMetalearner(String algo) {
...
+                    distribution = DistributionFamily.multinomial;
+                }
+
+                Log.warn("Distribution \"" + parms._distribution +
+                        "\" is not supported by metalearner algorithm \"" + getAlgo() +
+                        "\". Using \"" + distribution + "\" instead.");
+
+                parms._distribution = distribution;
+            }
+            _model._dist = DistributionFactory.getDistribution(parms);
```
sebhrusen, line 138: is it possible to move this last statement in `compute()`?
I'm not a fan of validating params and modifying the model in the same place.

sebhrusen: you could add another hook to `Metalearner` like you did with `validateParams`

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -83,6 +83,11 @@ public void initMetalearnerParams() {
     public void initMetalearnerParams(Metalearner.Algorithm algo) {
       _metalearner_algorithm = algo;
       _metalearner_parameters = Metalearners.createParameters(algo.name());
+      if (Metalearners.getActualMetalearnerAlgo(algo) == Metalearner.Algorithm.glm){
+        // FIXME: This is here because there is no Family.AUTO. It enables us to know if the user specified family or not.
+        // FIXME: Family.AUTO will be implemented in https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7444
+        ((GLMModel.GLMParameters) _metalearner_parameters)._family = null;
+      }
```
sebhrusen, line 90: I think it's fine workaround for now.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -257,6 +262,18 @@ void doScoreOrCopyMetrics(Job job) {
...
+      return Enum.valueOf(DistributionFamily.class, aFamily.toString());
+    }
+    catch (IllegalArgumentException e) {
+      throw new H2OIllegalArgumentException("Don't know how to find the right DistributionFamily for Family: " + aFamily);
+    }
+  }
+
   private DistributionFamily distributionFamily(Model aModel) {
     // TODO: hack alert: In DRF, _parms._distribution is always set to multinomial.  Yay.
     if (aModel instanceof DRFModel)
```
sebhrusen, line 279: wondering if this logic is not outdated: DRF seems to only exclude `quasi-binomial` and custom distributions currently.
As this is used to identify the SE distribution, this can have a bad impact. wdyt?

tomasfryda: I am not sure. I know there are checks for quasi binomial and custom distribution but in documentation there is no mention of DRF supporting distribution (no mention on DRF's page[1]) and on distribution's page[2] only GBM, DeepLearning and XGBoost are mentioned.

[1] https://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/drf.html?highlight=distribution#
[2] https://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/algo-params/distribution.html?highlight=distribution

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -313,6 +331,113 @@ else if (aModel._output.isClassifier())
...
+        _parms._metalearner_parameters._quantile_alpha = parms._quantile_alpha;
+        break;
+    }
+  }
+
+  /**
+   * Inherit family and its parameters
+   * @param parms
+   */
+  private void inheritFamilyAndParms(Model.Parameters parms) {
```
sebhrusen, line 373: I got confused multiple time with all those params in your `inherit___` methods, especially with calls like:
```
inheritFamilyAndParms(_parms._metalearner_parameters);
...
inheritDistributionAndParms(_parms._metalearner_parameters);
```
this would mean here that `parms` and `metaParams` are the same object, so there's nothing to inherit.

Am I missing sth? Can you please clarify and possibly rename `parms` if it's intended to be only for base model params?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -344,6 +476,32 @@ void checkAndInheritModelProperties() {
...
+      if (cv_required_on_base_model) {
+        // If we don't have a fold_column require:
+        // nfolds > 1
+        // nfolds consistent across base models
+        if (aModel._parms._fold_column == null && aModel._parms._nfolds < 2)
+          throw new H2OIllegalArgumentException("Base model does not use cross-validation: " + aModel._parms._nfolds);
+        if (!aModel._parms._keep_cross_validation_predictions)
+          throw new H2OIllegalArgumentException("Base model does not keep cross-validation predictions: " + aModel._parms._nfolds);
+      }
+
```
sebhrusen, line 504: getting lost here. Is it related with this PR?

sebhrusen: if this is clean-up / improvement, can you please move to different PR?
For a small cleanup, I wouldn't mind, but looks like there was also some logic added.

tomasfryda: Sorry, not completely related to this PR. It's just moved couple lines up from the `if (retrievedFirstModelParams) {` so we check all the base models not just starting with the second one.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -387,26 +533,66 @@ void checkAndInheritModelProperties() {
...
+            if ((aModel instanceof GLMModel) &&
+                    (Metalearners.getActualMetalearnerAlgo(_parms._metalearner_algorithm) == Metalearner.Algorithm.glm) &&
+                    !((GLMModel.GLMParameters) _parms._metalearner_parameters)._link.equals(((GLMModel) aModel)._parms._link)) {
+              if (inferredFromGLM) {
+                sameParams = false;
+              } else {
+                inheritFamilyAndParms(aModel._parms);
+                inferredFromGLM = true;
+              }
+            }
```
sebhrusen, line 574: maybe it would have been more clear to track `firstGLM` instead of `inferredFromGLM`.
This way you can get use similar logic for `sameParams` as for others, sth approx. like:
```
if aModel instanceof GLMModel and meta_algo == glm:
   if firstGLM == null:
      firstGLM = aModel
      inherit...
   else:
      sameParams = _parms._metalearner_parameters._link == aModel._parms._link
```

it has the advantage of being more in line with the code above, and a bit easier to understand imo. wdyt?

h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_distributions.py
```diff
@@ -0,0 +1,521 @@
...
+
+if __name__ == "__main__":
+    pyunit_utils.run_tests([
+        infer_distribution_test,
+        infer_family_test,
+        infer_mixed_family_and_dist_test,
+        metalearner_obeys_metalearner_params_test,
+        infer_uses_defaults_when_base_model_doesnt_support_distributions_test,
+        basic_inference_works_for_DRF_and_NB_test,
+    ])
```
sebhrusen, line 521: I understand it's faster to write Py tests for this kind of logic — I do the same —, and you have great tests with good coverage.
But let's try to have at least basic scenarios tested in Java as well, thx.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
sebhrusen: documentation is more likely to be late than ahead.
@maurever? apparently you're the last one who added validation for DRF distributions, do you know what it supports currently?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
tomasfryda: My bad, this function also did set the `_dist` for the SE so that was the only reason to call it with metaParams. That was moved to different part but I kept it there so it was essentially no-op. I removed the inherit.* calls with metaParams. I also renamed `parms` to `baseModelParms` to make the intention explicit.

h2o-algos/src/main/java/hex/ensemble/Metalearners.java
tomasfryda: I moved it to `compute` but instead of creating it by `DistributionFactory` I use the same `_dist` as the metalearner which I think is more semantically correct. And if the metalearner model such as GLM ignores it, SE should ignore it as well I think and since the `_dist` is property of `Model.Parameters` it should be safe to do it this way and potentially even safer, e.g., if the underlying model coerces the family/distribution somehow during its creation.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
tomasfryda: I agree

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
sebhrusen: to be clear, if you think it's justified to include the changes in this PR, please go on (probably better this than a FIXME).
From a reviewer perspective, I just noticed that some lines were moved, which is totally fine, but I also noticed that some logic was changed without explanation: as it didn't look immediately related with the PR (but maybe it is), I just expect the author to add a small comment in the diff view to inform reviewers, and avoid them scratching their head wondering "why this change?".

tomasfryda: If I would think it's justified I would try to argue a little bit more but this PR is quite complex already. The part with FIXME was a result of experimentation with saving and loading AutoML models that I did when you were on vacation.

I would create a separate jira for that functionality (and tests), if that's ok with you...

h2o-algos/src/main/java/hex/ensemble/Metalearner.java
```diff
@@ -86,6 +88,8 @@ void compute() {
       Log.info("Finished training metalearner model(" + _model._parms._metalearner_algorithm + ").");
 
       _model._output._metalearner = builder.get();
+      _model._dist = _model._output._metalearner._dist;
```
sebhrusen, line 91: Yes, looks better if it is directly copied from metalearner.

sebhrusen: I'm wondering if there are other metalearner properties that we want to propagate to the SE after the training...
(not asking you do this here, just currious)

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
sebhrusen: perfect!

PR #4568, opened 2020-05-01
PUBDEV-6686: Show how concurrent DRF building can lead to a failure

michalkurka:
We need to make sure "_toDelete" is not shared for different instances of model builders. Otherwise completion of one model might trigger deleting data of another model.


michalkurka:
@sebhrusen FIY

PR #4581, opened 2020-05-05
PUBDEV-3292 - Drop duplicate observations

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-3292

### DIFF
--- a/h2o-core/src/main/java/water/rapids/Merge.java
+++ b/h2o-core/src/main/java/water/rapids/Merge.java
@@ -13,6 +13,9 @@ import static java.math.BigInteger.ONE;
 import static water.rapids.SingleThreadRadixOrder.getSortedOXHeaderKey;
 
 public class Merge {
+  
+  public static int ASCENDING = 1;
+  public static int DESCENDING = -1; 
 
   public static Frame sort(final Frame fr, int col) {
     return sort(fr, new int[]{col});
@@ -25,6 +28,7 @@ public class Merge {
 
     return sort(fr, cols, ascending); // default is to sort in ascending order
   }
+  
   // Radix-sort a Frame using the given columns as keys.
   // This is a fully distributed and parallel sort.
   // It is not currently an in-place sort, so the data is doubled and a sorted copy is returned.
--- a/h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
@@ -1,0 +1,88 @@
+package water.rapids.ast.prims.filters.dropduplicates;
+
+import water.fvec.Frame;
+import water.rapids.Env;
+import water.rapids.Val;
+import water.rapids.ast.AstPrimitive;
+import water.rapids.ast.AstRoot;
+import water.rapids.vals.ValFrame;
+import water.util.EnumUtils;
+
+import java.util.Arrays;
+
+/**
+ * Removes duplicated rows, leaving only the first or last observed duplicate in place.
+ */
+public class AstDropDuplicates extends AstPrimitive<AstDropDuplicates> {
+    @Override
+    public int nargs() {
+        return 1 + 3;
+    }
+
+    @Override
+    public String[] args() {
+        return new String[]{"ary", "frame", "cols", "droporder"};
+    }
+
+    @Override
+    public Val apply(Env env, Env.StackHelp stk, AstRoot[] asts) {
+        final Frame deduplicatedFrame = stk.track(asts[1].exec(env)).getFrame();
+        final int[] comparedColumnsIndices = parseComparedColumnIndices(deduplicatedFrame, stk.track(asts[2].exec(env)));
+        final String dropOrderString = asts[3].str();
+        final KeepOrder keepOrder = EnumUtils.valueOfIgnoreCase(KeepOrder.class, dropOrderString)
+                .orElseThrow(() -> new IllegalArgumentException(String.format("Unknown drop order: '%s'. Known types: %s",
+                        dropOrderString, Arrays.toString(KeepOrder.values()))));
+
+        final DropDuplicateRows dropDuplicateRows = new DropDuplicateRows(deduplicatedFrame, comparedColumnsIndices, keepOrder);
+        final Frame outputFrame = dropDuplicateRows.dropDuplicates();
+        return new ValFrame(outputFrame);
+    }
+
+    private int[] parseComparedColumnIndices(final Frame deduplicatedFrame, final Val comparedColumns) {
+        final int[] columnRange;
+
+        if (comparedColumns.isStr()) {
+            final String columnName = comparedColumns.getStr();
+            final int columnIndex = deduplicatedFrame.find(columnName);
+            if (columnIndex == -1) {
+                throw new IllegalArgumentException(String.format("Unknown column name: '%s'", columnName));
+            }
+            columnRange = new int[]{columnIndex};
+        } else if (comparedColumns.isStrs()) {
+            final String[] columnNames = comparedColumns.getStrs();
+            columnRange = new int[columnNames.length];
+
+            for (int i = 0; i < columnNames.length; i++) {
+                final String columnName = columnNames[i];
+                final int columnIndex = deduplicatedFrame.find(columnName);
+                if (columnIndex == -1) {
+                    throw new IllegalArgumentException(String.format("Unknown column name: '%s'", columnName));
+                }
+                columnRange[i] = columnIndex;
+
+            }
+
+        } else if (comparedColumns.isNums()) {
+            final double[] columnRangeDouble = comparedColumns.getNums();
+            columnRange = new int[columnRangeDouble.length];
+            for (int i = 0; i < columnRangeDouble.length; i++) {
+                columnRange[i] = (int) columnRangeDouble[i];
+                if (columnRange[i] < 0 || columnRange[i] > deduplicatedFrame.numCols() - 1) {
+                    throw new IllegalArgumentException(String.format("No such column index: '%d', frame has %d columns," +
+                                    "maximum index is %d.",
+                            columnRange[i], deduplicatedFrame.numCols(), deduplicatedFrame.numCols() - 1));
+                }
+            }
+        } else {
+            throw new IllegalArgumentException(String.format("Column range for deduplication must either be a set of columns, or a " +
+                    "column range. Given type: %s", comparedColumns.type()));
+        }
+
+        return columnRange;
+    }
+
+    @Override
+    public String str() {
+        return "dropdup";
+    }
+}
--- a/h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
@@ -1,6 +1,7 @@
 package water.rapids.ast.prims.filters.dropduplicates;
 
 import water.fvec.Frame;
+import water.fvec.Vec;
 import water.rapids.Env;
 import water.rapids.Val;
 import water.rapids.ast.AstPrimitive;
@@ -27,7 +28,7 @@ public class AstDropDuplicates extends AstPrimitive<AstDropDuplicates> {
   @Override
   public Val apply(Env env, Env.StackHelp stk, AstRoot[] asts) {
     final Frame deduplicatedFrame = stk.track(asts[1].exec(env)).getFrame();
-    final int[] comparedColumnsIndices = parseComparedColumnIndices(deduplicatedFrame, stk.track(asts[2].exec(env)));
+    final int[] comparedColumnsIndices = parseAndCheckComparedColumnIndices(deduplicatedFrame, stk.track(asts[2].exec(env)));
     final String dropOrderString = asts[3].str();
     final KeepOrder keepOrder = EnumUtils.valueOfIgnoreCase(KeepOrder.class, dropOrderString)
             .orElseThrow(() -> new IllegalArgumentException(String.format("Unknown drop order: '%s'. Known types: %s",
@@ -38,7 +39,7 @@ public class AstDropDuplicates extends AstPrimitive<AstDropDuplicates> {
     return new ValFrame(outputFrame);
   }
 
-  private int[] parseComparedColumnIndices(final Frame deduplicatedFrame, final Val comparedColumns) {
+  private int[] parseAndCheckComparedColumnIndices(final Frame deduplicatedFrame, final Val comparedColumns) {
     final int[] columnRange;
 
     if (comparedColumns.isStr()) {
@@ -57,6 +58,9 @@ public class AstDropDuplicates extends AstPrimitive<AstDropDuplicates> {
         final int columnIndex = deduplicatedFrame.find(columnName);
         if (columnIndex == -1) {
           throw new IllegalArgumentException(String.format("Unknown column name: '%s'", columnName));
+        } else if (deduplicatedFrame.types()[columnIndex] == Vec.T_STR || deduplicatedFrame.types()[columnIndex] == Vec.T_BAD
+                || deduplicatedFrame.types()[columnIndex] == Vec.T_UUID) {
+          throw new IllegalArgumentException(String.format("Column '%s' is of unsupported type %s for row de-duplication.", columnName, Vec.TYPE_STR[deduplicatedFrame.types()[columnIndex]]));
         }
         columnRange[i] = columnIndex;
 
--- a/h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
@@ -14,75 +14,75 @@ import java.util.Arrays;
  * Removes duplicated rows, leaving only the first or last observed duplicate in place.
  */
 public class AstDropDuplicates extends AstPrimitive<AstDropDuplicates> {
-    @Override
-    public int nargs() {
-        return 1 + 3;
-    }
+  @Override
+  public int nargs() {
+    return 1 + 3;
+  }
 
-    @Override
-    public String[] args() {
-        return new String[]{"ary", "frame", "cols", "droporder"};
-    }
+  @Override
+  public String[] args() {
+    return new String[]{"ary", "frame", "cols", "droporder"};
+  }
 
-    @Override
-    public Val apply(Env env, Env.StackHelp stk, AstRoot[] asts) {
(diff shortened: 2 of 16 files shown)


h2o-core/src/main/java/water/data/transformations/DropDuplicatesTask.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public DropDuplicatesTask(final int[] comparedColumnIndices) {
+        this.comparedColumnIndices = comparedColumnIndices;
+    }
+
+    @Override
+    public void map(Chunk[] cs) {
+        assert cs.length > 0;
+        int equalRowsStartIndex = -1;
+        for (int row = 0; row < cs[0].len(); row++) {
```
michalkurka, line 18: you probably want to do `row < cs[0].len() - 1`

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
```diff
@@ -0,0 +1,88 @@
...
+    final KeepOrder keepOrder = EnumUtils.valueOfIgnoreCase(KeepOrder.class, dropOrderString)
+            .orElseThrow(() -> new IllegalArgumentException(String.format("Unknown drop order: '%s'. Known types: %s",
+                    dropOrderString, Arrays.toString(KeepOrder.values()))));
+
+    final DropDuplicateRows dropDuplicateRows = new DropDuplicateRows(deduplicatedFrame, comparedColumnsIndices, keepOrder);
+    final Frame outputFrame = dropDuplicateRows.dropDuplicates();
+    return new ValFrame(outputFrame);
+  }
+
+  private int[] parseComparedColumnIndices(final Frame deduplicatedFrame, final Val comparedColumns) {
```
honzasterba, line 41: I would think we have such functionality on Rapids somewhere already, if not I think it deserves its own class and bunch of unit tests for all the execution paths. The AstDropDuplicatesTest does not test this method properly (nor should it)

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/KeepOrder.java
```diff
@@ -0,0 +1,9 @@
+package water.rapids.ast.prims.filters.dropduplicates;
+
+/**
+ * Determines which duplicated row is kept during row de-duplication process.
+ */
+public enum KeepOrder {
+  First, // Retain first, drop last
```
honzasterba, line 7: knit: should be "Retain first, drop rest" and "Retain last, drop rest"
... there used to be a long comment here about sort orders and such things, until I read the rest of the code and found out that its all pretty clear ...

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRows.java
```diff
@@ -0,0 +1,89 @@
...
+      outputFrame = Merge.sort(deDuplicatedFrame, deDuplicatedFrame.numCols() - 1);
+      outputFrame.remove(outputFrame.numCols() - 1);
+      outputFrame._key = sourceFrame._key != null ? Key.make(sourceFrame._key.toString() + "_deduplicated") : Key.make();
+      DKV.put(outputFrame);
+      return outputFrame;
+
+    } finally {
+      final Vec label = sourceFrame.remove(LABEL_COLUMN_NAME);
+      if (label != null) {
+        label.remove();
```
honzasterba, line 59: probably unnecessary since the sortedFrame is tracked

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+      }
+
+    }
+  }
+
+  private boolean compareFirstRowWithPreviousChunk(final Chunk[] chunks, final int row, final int chunkId) {
+    for (int column : comparedColumnIndices) {
+      final double previousValue = chunkBoundaries.vec(column).chunkForChunkIdx(chunkId - 1).atd(0);
+      final double curentValue = chunks[column].atd(row);
```
honzasterba, line 53: does this work for string columns as well? don't see any tests for String

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRows.java
Pscheidl: We label the sourceFrame, it has to be done before sorting. This is to avoid yet another data duplication. H2O is generally single-user, therefore it is safe to assume we can add label vec temporarily to an existing frame.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
Pscheidl: We don't support string columns, unfortunately. For this, we need Merge.Sort to support String columns. And it doesn't. Sorting will fail. For now, I propose adding a check and a nice error message would do. WDYT ?

This brings me to another problem - All chunks have all the methods, and some of them will throw `IllegalArgumentException`  (I know, Java has UnsupportedOperationException for this, don't ask me why this is there). Here, I am using `atd`, thus for chunks of `long`, not all 64-bit signed integers can be represented as double. We have to somehow check the type of the chunk and act accordingly.

honzasterba: yes, if we dont support some column type we should definitely check at the beginning and throw a reasonable error

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRows.java
```diff
@@ -0,0 +1,89 @@
...
+              .doAll(sortedFrame.types(), sortedFrame)
+              .outputFrame());
+
+      final Frame deDuplicatedFrame = Scope.track(new DropDuplicateRowsTask(chunkBoundaries, comparedColumnIndices)
+              .doAll(sortedFrame.types(), sortedFrame)
+              .outputFrame(Key.make(), sourceFrame.names(), sourceFrame.domains())); // Removing duplicates, domains remain the same
+
+      outputFrame = Merge.sort(deDuplicatedFrame, deDuplicatedFrame.numCols() - 1);
+      outputFrame.remove(outputFrame.numCols() - 1);
+      outputFrame._key = sourceFrame._key != null ? Key.make(sourceFrame._key.toString() + "_deduplicated") : Key.make();
```
michalkurka, line 52: imho the output frame doesn't need a key and shouldn't be in DKV - rapids will take of that

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
```diff
@@ -0,0 +1,70 @@
...
+      if (!equal) {
+        for (int columnIndex = 0; columnIndex < chunks.length; columnIndex++) {
+          chunks[columnIndex].extractRows(newChunks[columnIndex], row);
+        }
+      }
+
+    }
+  }
+
+  private boolean compareFirstRowWithPreviousChunk(final Chunk[] chunks, final int row, final int chunkId) {
```
michalkurka, line 50: this is only called for row=0 (and the name of the method also suggests it)

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
```diff
@@ -0,0 +1,70 @@
...
+      final double previousValue = chunkBoundaries.vec(column).chunkForChunkIdx(chunkId - 1).atd(0);
+      final double curentValue = chunks[column].atd(row);
+      if (previousValue != curentValue) return false;
+    }
+
+    return true;
+  }
+
+  private boolean compareRows(final Chunk[] chunks, final int row) {
+    for (final int column : comparedColumnIndices) {
```
michalkurka, line 61: this doesn't work for NAs, NAs are represented as NaN, see
```
double x = Double.NaN;
double y = Double.NaN;
return x == y;
```

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
```diff
@@ -0,0 +1,70 @@
...
+    for (int column : comparedColumnIndices) {
+      final double previousValue = chunkBoundaries.vec(column).chunkForChunkIdx(chunkId - 1).atd(0);
+      final double curentValue = chunks[column].atd(row);
+      if (previousValue != curentValue) return false;
+    }
+
+    return true;
+  }
+
+  private boolean compareRows(final Chunk[] chunks, final int row) {
```
michalkurka, line 60: I suggest changing the signature of this method to:
```
Chunk[] chunksA, int rowA, Chunk[] chunksB, int rowB
```

such a method can be used from `compareFirstRowWithPreviousChunk` and keep the comparison implemented just once (the method will get more complicated once the issue pointed out bellow is fixed)

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
```diff
@@ -0,0 +1,70 @@
...
+    for (final int column : comparedColumnIndices) {
+      final double previousValue = chunks[column].atd(row - 1);
+      final double curentValue = chunks[column].atd(row);
+      if (previousValue != curentValue) return false;
```
michalkurka, line 64: please add tests that have NA values in key-columns

michalkurka:
Very nice @Pscheidl, minor issues found

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRows.java
Pscheidl: No it doesn't. Removed, also removed testing presence of a Key from JUnit tests.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
Pscheidl: Yes, I know. There is a small amount of logic still after the recent changes and I'd personally keep it. Can't really say if JVM will optimize the stackframe creation in this case.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
Pscheidl: Done. Thank you for the tip. Much cleaner now. Also called from the `compareFirstRowWithPreviousChunk` (as suggested).

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
Pscheidl: Added logic to handle NAs beforehand.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
Pscheidl: Added JUnit tests and one test against Pandas using modified "chess" dataset with NA column.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/KeepOrder.java
Pscheidl: Re-worded. Thank you.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
Pscheidl: I've added the check and a test for it. Also added BAD columns (obvious) and UUID columns (can't be sorted as well).

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
Pscheidl: Can't find it. We could use frame.vecs - which we'd still need to iterate for checks. I agree this could be generalized, the closest generalization to this is till `Frame.vecs(String[])` method. We could enhance it with desired type check and check if the column exists.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
Pscheidl: Also, I've added handling for T_CAT and T_TIME separately. In case of T_CAT, we could handle it as double (2^31 of ints fits into IEEE 754 double), but long is closer and a bit cheaper to convert I'd say, therefore used at8 for time..

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/DropDuplicateRowsTask.java
```diff
@@ -0,0 +1,88 @@
...
+  private boolean compareFirstRowWithPreviousChunk(final Chunk[] chunks, final int row, final int chunkId) {
+    final Chunk[] previousRowChunks = new Chunk[chunkBoundaries.numCols()];
+
+    for (int column = 0; column < chunkBoundaries.numCols(); column++) {
+      previousRowChunks[column] = chunkBoundaries.vec(column).chunkForChunkIdx(chunkId - 1);
+    }
+    return compareRows(chunks, row, previousRowChunks, 0);
+  }
+
+  private boolean compareRows(Chunk[] chunksA, int rowA, Chunk[] chunksB, int rowB) {
```
Pscheidl, line 60: Can be final.

Pscheidl: Done.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
```diff
@@ -0,0 +1,92 @@
...
+        columnRange[i] = columnIndex;
+
+      }
+
+    } else if (comparedColumns.isNums()) {
+      final double[] columnRangeDouble = comparedColumns.getNums();
+      columnRange = new int[columnRangeDouble.length];
+      for (int i = 0; i < columnRangeDouble.length; i++) {
+        columnRange[i] = (int) columnRangeDouble[i];
+        if (columnRange[i] < 0 || columnRange[i] > deduplicatedFrame.numCols() - 1) {
```
maurever, line 74: From this check, there could be also list of indices not only a range of indices. Am I correct, if I use an array [1,2,3] but also [2,1,3]? So the second one is not range and still can be used, or not? It is not critical, just quibble. :)

h2o-core/src/main/java/water/rapids/Merge.java
```diff
@@ -13,6 +13,9 @@
 import static water.rapids.SingleThreadRadixOrder.getSortedOXHeaderKey;
 
 public class Merge {
+  
+  public static int ASCENDING = 1;
```
maurever, line 17: Are these constant used somewhere else?

Pscheidl: No, I introduced them to and the code coming with this PR is the first user. I wanted to get rid of magical constants in the code. I myself misunderstood this once and used 0 for descending order. :rabbit:

Pscheidl:
Python test failure due to a missing test library I had locally. Will fix.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
Pscheidl: This part of code only checks if the given column indices are within deduplicated frame's range. It's a basic sanity check not to try to get a vector with index that is not there. If the index is incorrect, a user-friendly errror is thrown with a sort-of descriptive message.

maurever: Yes, I understand this part of the code. I just comment here, because it is the first occurrence. But I thought it global. If a user can use also [2,1, 3] order, it is not range, but indices. :) But it is really just note about naming, not about functionality...

Pscheidl: Oh, I see. Let me fix that m'aam.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/ColumnIndicesParser.java
```diff
@@ -0,0 +1,70 @@
...
+        columnRange[i] = columnIndex;
+      }
+
+    } else if (comparedColumns.isNums()) {
+      final double[] columnRangeDouble = comparedColumns.getNums();
+      columnRange = new int[columnRangeDouble.length];
+      for (int i = 0; i < columnRangeDouble.length; i++) {
+        columnRange[i] = (int) columnRangeDouble[i];
+        if (columnRange[i] < 0 || columnRange[i] > deduplicatedFrame.numCols() - 1) {
+          throw new IllegalArgumentException(String.format("No such column index: '%d', frame has %d columns," +
```
honzasterba, line 51: knit: missing space after comma

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/AstDropDuplicates.java
Pscheidl: Fixed. Thank you.

h2o-core/src/main/java/water/rapids/ast/prims/filters/dropduplicates/ColumnIndicesParser.java
Pscheidl: Will fix.

Pscheidl: Done !

Pscheidl:
Waiting for tests to finish.

PR #4582, opened 2020-05-05
PUBDEV-7504: Make sure user gets a warning when trying to get varimp from Stacked Ensemble

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7504

Fixes R to warn that SE doesn't have var.imp.

In python, it works almost as expected - it **prints** the message but doesn't throw an exception.  I would rather throw an exception but then again h2o-py does quite a lot of things that aren't exactly pythonic so I assume printing the warning is wanted behaviour.  But the user gets a clear message of what's wrong so it's probably not that urgent to change the behaviour.

Not throwing an exception in `varimp`  leads to the following message in `varimp_plot` (the warning is printed and then execution continues and fails on something else).

```

Warning: This model doesn't have variable importances
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-12-0e6c61b8b46e> in <module>
----> 1 _se.varimp_plot()

~/sources/h2o-3/h2o-py/h2o/model/model_base.py in varimp_plot(self, num_of_features, server)
   1356         importances = self.varimp(use_pandas=False)
   1357         # features labels correspond to the first value of each tuple in the importances list
-> 1358         feature_labels = [tup[0] for tup in importances]
   1359         # relative importances correspond to the first value of each tuple in the importances list
   1360         scaled_importances = [tup[2] for tup in importances]

TypeError: 'NoneType' object is not iterable

```

### DIFF
--- a/h2o-r/h2o-package/R/models.R
+++ b/h2o-r/h2o-package/R/models.R
@@ -1944,8 +1944,8 @@ h2o.varimp <- function(object) {
   o <- object
   if( is(o, "H2OModel") ) {
     vi <- o@model$variable_importances
-    if( is.null(vi) ) { # may be glm
-      tvi <- object@model$standardized_coefficient_magnitudes 
+    if( is.null(vi) && !is.null(object@model$standardized_coefficient_magnitudes)) { # may be glm
+      tvi <- object@model$standardized_coefficient_magnitudes
       maxCoeff <- max(tvi$coefficients)
       sumCoeff <- sum(tvi$coefficients)
       scaledCoeff <- tvi$coefficients/maxCoeff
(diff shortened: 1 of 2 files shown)


ledell:
@tomasfryda My only feedback would be to make the test as simple as possible for speed purposes (use 2 models instead of 3, and use 10 trees each), as to not use more resources than necessary in our test harness.  Thanks!

tomasfryda:
Thanks! I modified the test according to your suggestion.

ledell:
This is ready to be merged, did you want to take a look @sebhrusen ?

PR #4584, opened 2020-05-05
PUBDEV-7505: Update default value for min_split_improvement parameter

hannah-tillman:


h2o-docs/src/product/data-science/algo-params/min_split_improvement.rst
```diff
@@ -9,7 +9,7 @@ Description
 
 This option specifies the minimum relative improvement in squared error reduction in order for a split to occur. When properly tuned, this option can help reduce overfitting because the algorithm will stop splitting when all the possible splits lead to worse error measures. In addition, a single tree will stop splitting when there are no more splits that satisfy the ``min_rows`` parameter, if it reaches ``max_depth``, or if there are no splits that satisfy this ``min_split_improvement`` parameter.
 
-This value defaults to 0.00001. Optimal values for this parameter are in the 1e-10...1e-3 range.
+This value defaults to 0.00001 for all models except XGBoost, which defaults to 0. Optimal values for this parameter are in the 1e-10...1e-3 range.
```
angela0xdata, line 12: for all models sounds odd because it's only supported in GBM, DRF, and XGBoost. Maybe change to:
For GBM and DRF models, this value defaults to 0.00001; for XGBoost models, this value defaults to 0.
...
(but if this option gets added to other tree algos, then we might circle back to your original edit.)

PR #4594, opened 2020-05-10
PUBDEV-6938 GroupBy - support for grouping by String columns

jancijen:
This PR adds GroupBy support for` String` values (not for `AstGroup.FCN` functions, but it allows grouping rows by a column which is of type `String`), in order to implement TF-IDF (#4380) by using GroupBy operation (`AstGroup`).

This functionality is not directly accessible from "Ast API", but it can be used via public methods and classes from `AstGroup` as shown in the unit test.

### DIFF
--- a/h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
@@ -1,7 +1,11 @@
 package water.rapids.ast.prims.mungers;
 
 import water.*;
-import water.fvec.*;
+import water.fvec.Chunk;
+import water.fvec.Frame;
+import water.fvec.NewChunk;
+import water.fvec.Vec;
+import water.parser.BufferedString;
 import water.rapids.Env;
 import water.rapids.Merge;
 import water.rapids.Val;
@@ -27,10 +31,8 @@ import java.util.Arrays;
  * returns a row per unique group, with the first columns being the grouping
  * columns, and the last column(s) the reduction result(s).
  * <p/>
- * However, GroupBy operations will not be performed on String columns.  These columns
- * will be skipped. GroupBy for string columns cannot be performed directly using Ast API, but 
- * there is an internal implementation for such GroupBy operations - 
- * {@link AstGroup#performGroupingWithAggregations(Frame, int[], int[], AGG[])}.
+ * Grouping can be performed on both numeric and string columns, but aggregation
+ * functions are limited to the ones listed in {@link AstGroup.FCN}.
  * <p/>
  * The returned column(s).
  */
@@ -288,25 +290,32 @@ public class AstGroup extends AstPrimitive {
 
     return performGroupingWithAggregations(fr, gbCols, aggs);
   }
-  
+
+  /**
+   * Performs grouping and aggregations on given data.
+   *
+   * @param fr          frame on which grouping should be performed
+   * @param gbCols      indices of columns (of given frame) to be used for grouping.
+   * @param aggs        aggregations to be performed on the grouped data.
+   *
+   * @return  grouped data with aggregations results.
+   */
   public ValFrame performGroupingWithAggregations(Frame fr, int[] gbCols, AGG[] aggs) {
-    return performGroupingWithAggregations(fr, gbCols, new int[]{}, aggs);
-  }
-
-  public ValFrame performGroupingWithAggregations(Frame fr, int[] gbColsNum, int[] gbColsStr, AGG[] aggs) {
     final boolean hasMedian = hasMedian(aggs);
-    final IcedHashSet<G> gss = doGroups(fr, gbColsNum, gbColsStr, aggs, hasMedian, _per_node_aggregates);
+    final byte[] gbColsTypes = ArrayUtils.select(fr.types(), gbCols);
+    
+    final IcedHashSet<G> gss = doGroups(fr, gbCols,  gbColsTypes, aggs, hasMedian, _per_node_aggregates);
     final G[] grps = gss.toArray(new G[gss.size()]);
+    
+    applyOrdering(gbCols, gbColsTypes, grps);
 
-    applyOrdering(gbColsNum, gbColsStr, grps);
+    final int medianActionsNeeded = hasMedian ? calculateMediansForGRPS(fr, gbCols, gbColsTypes, aggs, gss, grps) : -1;
 
-    final int medianActionsNeeded = hasMedian ? calculateMediansForGRPS(fr, gbColsNum, gbColsStr, aggs, gss, grps) : -1;
-
-    MRTask mrFill = prepareMRFillTask(grps, aggs, medianActionsNeeded);
+    MRTask mrFill = prepareMRFillTask(grps, aggs, gbColsTypes, medianActionsNeeded);
 
     String[] fcNames = prepareFCNames(fr, aggs);
 
-    Frame f = buildOutput(gbColsNum, gbColsStr, aggs.length, fr, fcNames, grps.length, mrFill);
+    Frame f = buildOutput(gbCols, aggs.length, fr, fcNames, grps.length, mrFill);
     return new ValFrame(f);
   }
 
@@ -316,21 +325,26 @@ public class AstGroup extends AstPrimitive {
         return true;
     return false;
   }
-
-  private MRTask prepareMRFillTask(final G[] grps, final AGG[] aggs, final int medianCount) {
+  private MRTask prepareMRFillTask(final G[] grps, final AGG[] aggs, final byte[] gbColsTypes, final int medianCount) {
     return new MRTask() {
       @Override
       public void map(Chunk[] c, NewChunk[] ncs) {
         int start = (int) c[0].start();
         for (int i = 0; i < c[0]._len; ++i) {
           G g = grps[i + start];  // One Group per row
-          int gbColsCnt = g._gs.length + g._gsStr.length;
+          int gbColsCnt = g.colsCount();
           int j;
-          // TODO: Columns order - currently string columns are first
-          for (j = 0; j < g._gsStr.length; j++) // The string group key, as a row
-            ncs[j].addStr(g._gsStr[j]);
-          for (j = g._gsStr.length; j < gbColsCnt; j++) // The numeric group key, as a row
-            ncs[j].addNum(g._gs[j - g._gsStr.length]);
+
+          int strIdx = 0;
+          int numIdx = 0;
+
+          for (j = 0; j < gbColsCnt; j++) {
+            if (gbColsTypes[j] == Vec.T_STR)
+              ncs[j].addStr(g._gsStr[strIdx++]);
+            else
+              ncs[j].addNum(g._gs[numIdx++]);
+          }
+
           for (int a = 0; a < aggs.length; a++) {
             if ((medianCount >=0) && g.medianR._isMedian[a])
               ncs[j++].addNum(g.medianR._medians[a]);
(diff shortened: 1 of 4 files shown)


h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -753,15 +796,31 @@ public G(int ncols, AGG[] aggs, boolean hasMedian) {
...
 
     public G fill(int row, Chunk chks[]) {
-      for (int c = 0; c < chks.length; c++) // For all selection cols
-        _gs[c] = chks[c].atd(row); // Load into working array
+      for (int c = 0; c < chks.length; c++) { // For all selection cols
+        Vec vec = chks[c].vec();
+        
+        // Load into working array
+        if (vec.isString())
+          _gsStr[c] = chks[c].stringAt(row);
```
michalkurka, line 804: it is better to keep everything in BufferedStrings as much as possible (atStr method), this won't duplicate the memory needed to hold the string, BufferedString just references an existing byte array that holds the actual data

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -383,15 +393,21 @@ private int countNumberOfAggregates(Frame fr, int numberOfColumns, AstRoot asts[
...
         // Compare 2 groups.  Iterate down _gs, stop when _gs[i] > that._gs[i],
         // or _gs[i] < that._gs[i].  Order by various columns specified by
         // gbCols.  NaN is treated as least
         @Override
         public int compare(G g1, G g2) {
-          for (int i = 0; i < gbCols.length; i++) {
+          for (int i = 0; i < gbColsStr.length; i++) {
+            if (g1._gsStr[i] != null && g2._gsStr[i] == null) return -1;
+            if (g1._gsStr[i] == null && g2._gsStr[i] != null) return 1;
+            if (!g1._gsStr[i].equals(g2._gsStr[i])) return g1._gsStr[i].compareTo(g2._gsStr[i]);
```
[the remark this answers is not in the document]
jancijen: I believe that it should not return `0` if we hit values that are are equal - based on the previous ordering implementation, as explained in the comment below:

```
// Compare 2 groups.  Iterate down _gs, stop when _gs[i] > that._gs[i],
// or _gs[i] < that._gs[i].  Order by various columns specified by
// gbCols.  NaN is treated as least
```

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
jancijen: Makes sense. I will adjust it.

jancijen:
@honzasterba I see.

Hm. I don't know whether I get what you mean by:

> ... and maybe two arrays with different value types and one mapping array to find the right value for given column/chunk ...

Could you elaborate a bit more?

The other possiblity that came to my mind was to use single array for column indices by data should be grouped and then add another array of types (containing enum values of certain allowed types - `Numeric` and `String`). Then just add if branching whenever columns are used. This I think, would end up with similar amount of code, but it would at least remove the "ordering issue" - now `String` groupby columns are always first followed by `Numeric` groupby columns.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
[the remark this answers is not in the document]
honzasterba: ah, thats true, than maybe you could do 

`int res = g1._gsStr[i].compareTo(g2._gsStr[i]); if (res != 0) return res;`

honzasterba:
yes, sort order is also an argument to use just one array of cols, I think the string/double dychotomy could be very well hidden inside the Group object.
code like `ncs[j].addStr(g._gsStr[j]);` could become `g.set(ncs[j], j)` or even the whole loops of https://github.com/h2oai/h2o-3/pull/4594/files#diff-4a6a6d53c166511396303e901c6f2338R330 could be moved into Group
also you could make G comparable

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -287,16 +305,47 @@ public ValFrame apply(Env env, Env.StackHelp stk, AstRoot asts[]) {
...
+   * Performs grouping and aggregations on given data.
+   *
+   * @param fr          frame on which grouping should be performed
+   * @param gbCols      indices of columns (of given frame) to be used for grouping.
+   * @param gbColsTypes types of columns which are being used for grouping.
+   * @param aggs        aggregations to be performed on the grouped data.
+   *
+   * @return  grouped data with aggregations results.
+   */
+  public ValFrame performGroupingWithAggregations(Frame fr, int[] gbCols, GroupByColType[] gbColsTypes, AGG[] aggs) {
```
honzasterba, line 335: each vec has type, why do we need to pass the col types? when the method can figure it out?

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -310,17 +359,28 @@ private static boolean hasMedian(AGG[] aggs) {
...
-
-  private MRTask prepareMRFillTask(final G[] grps, final AGG[] aggs, final int medianCount) {
+  private MRTask prepareMRFillTask(final G[] grps, final AGG[] aggs, final GroupByColType[] gbColsTypes, final int medianCount) {
     return new MRTask() {
       @Override
       public void map(Chunk[] c, NewChunk[] ncs) {
         int start = (int) c[0].start();
         for (int i = 0; i < c[0]._len; ++i) {
           G g = grps[i + start];  // One Group per row
+          int gbColsCnt = g._gs.length + g._gsStr.length;
```
honzasterba, line 369: can be replaced by gbColsTypes.length or eve better add g.colCount()

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -257,6 +267,14 @@ public double postPass(double ds[], long n) {
     }
   }
 
+  /**
+   * Allowed column types by which data can be grouped.
+   */
+  public enum GroupByColType {
```
honzasterba, line 273: internally se use byte for column types see Vec.T_STR

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -797,14 +917,19 @@ public String toString() {
   // here.
   private static class BuildGroup extends MRTask<BuildGroup> {
     final int[] _gbCols;
+    final GroupByColType[] _gbColsTypes;
+    final IcedHashMap<GroupByColType, Long> _gbColTypesCounts;
```
honzasterba, line 921: this would imply there is some flexibility to the contents of this hashmap when in fact its only used to store two keys and never iterated over, I think having two fields would end up in less code

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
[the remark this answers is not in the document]
jancijen: Sure, this looks cleaner and more efficient. I changed it.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
jancijen: I changed it to use `BufferedString` internally.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
jancijen: True. Careless mistake. Thank you. I changed it.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
jancijen: Looks good. I changed it.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
jancijen: This makes sense. I changed it throughout the code to use the type constants from `Vec`.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
jancijen: True. I guess I was overthinking this one. It seems that there won't be any other types necessary. I changed it.

jancijen:
So I changed the code to use single array for column indices for grouping the data (based on the previous discussion with @honzasterba). Therefore now it is already supported throught the "Ast API". I added unit test for this case as well.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -444,47 +481,54 @@ public static AstNumList check(long dstX, AstRoot ast) {
   // the selected 'gbCols' columns, and for each group compute aggregrate
   // results using 'aggs'.  Return an array of groups, with the aggregate results.
   public static IcedHashSet<G> doGroups(Frame fr, int[] gbCols, AGG[] aggs) {
-    return doGroups(fr, gbCols, aggs, false, true);
+    final byte[] frameColsTypes = fr.types();
+    final int[] gbColsTypes = Arrays.stream(gbCols).map(gbColIdx -> frameColsTypes[gbColIdx]).toArray();
```
honzasterba, line 485: duplicated code, maybe move to ArrayUtils

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -751,17 +802,46 @@ public G(int ncols, AGG[] aggs, boolean hasMedian) {
...
       for (int c = 0; c < cols.length; c++) // For all selection cols
-        _gs[c] = chks[cols[c]].atd(row); // Load into working array
+        // Load into working array
+        try {
+          if (gbColsTypes[c] == Vec.T_STR)
+            _gsStr[strIdx++] = chks[cols[c]].atStr(new BufferedString(), row);
+          else
+            _gs[numIdx++] = chks[cols[c]].atd(row);
+        }
+      catch (Exception e) {
```
honzasterba, line 842: why are you catching exception here?

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
jancijen: Good point. I implemented `ArrayUtils.select` for `byte[]` and used `byte[]` throughout the code.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
jancijen: Sorry, I missed this. It is a remnant from debugging. I removed it.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -533,14 +576,37 @@ public void atomic_op(double[][] d0ss, long[] n0s, int i, double[] d1s, long n1)
...
+    GBTask(int[] gbCols, byte[] gbColsTypes, AGG[] aggs, boolean hasMedian) {
       _gbCols = gbCols;
+      _gbColsTypes = gbColsTypes;
+      _stringGbColsCnt = ArrayUtils.occurenceCount(_gbColsTypes, Vec.T_STR);
+      _numericGbColsCnt = gbColsTypes.length - _stringGbColsCnt;
       _aggs = aggs;
       _hasMedian = hasMedian;
     }
+
+    protected void map(Chunk[] cs, IcedHashSet<G> groups) {
```
michalkurka, line 595: Very nice, thank you for removing the duplicated code!

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -751,17 +801,42 @@ public G(int ncols, AGG[] aggs, boolean hasMedian) {
...
 
     public G fill(int row, Chunk chks[]) {
-      for (int c = 0; c < chks.length; c++) // For all selection cols
-        _gs[c] = chks[c].atd(row); // Load into working array
+      for (int c = 0; c < chks.length; c++) { // For all selection cols
+        Vec vec = chks[c].vec();
+
+        // Load into working array
+        if (vec.isString())
+          _gsStr[c] = chks[c].atStr(new BufferedString(), row);
```
michalkurka, line 813: The motivation of the design of the G class is to avoid allocating memory per-observation/row

To follow the same pattern we should be re-using (pre-)allocated BufferedStrings when the G class gets created. This is a minor comment.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
```diff
@@ -751,17 +801,42 @@ public G(int ncols, AGG[] aggs, boolean hasMedian) {
...
       }
     }
+    
+    private int colsCount() { return _gs.length + _gsStr.length; }
 
     public G fill(int row, Chunk chks[]) {
-      for (int c = 0; c < chks.length; c++) // For all selection cols
-        _gs[c] = chks[c].atd(row); // Load into working array
+      for (int c = 0; c < chks.length; c++) { // For all selection cols
+        Vec vec = chks[c].vec();
```
michalkurka, line 809: A lot of code in this PR handles switching between the 2 arrays (_gs nad _gsStr). This is due to the fact that we can have an arbitrary order of string/numerical columns on the input. However, in the implementation itself, we can re-order the columns to make sure we will have the exact order we want, eg. numerical first then string ones. Would that be something that would help with making the code less complex?

michalkurka:
@jancijen thank you, it was a pleasure reading this PR.

I suggest thinking about if we can avoid allocation of BufferedStrings in the G#fill method. Also would like to know if re-ordering of the columns would be a feasible way of reducing the complexity of the code (in some parts - for sort you need the original order).

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
jancijen: I see. I changed it to use pre-allocated `BufferedStrings`.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
jancijen: There is nothing that comes to my mind that could solve this problem. Switching between `_gs` and `_gsStr` is done in 2 places - filling the output frame and sorting it. In both cases it is expected to preserve order of the columns. There is a possibility (at least for the "filling" part) of replacing `gbColsTypes`  array with 2 arrays defining order for each of the columns in `_gs` and `_gsStr`.  But I don't know whether it significantly reduces the complexity, if at all.

jancijen:
Thank you both for carefull feedback.

I addressed your comments, but as I mentioned I am not sure about reducing the complexity in switching between `_gs` and `_gsStr`.

h2o-core/src/main/java/water/rapids/ast/prims/mungers/AstGroup.java
maurever: From my point of view, two arrays are ok in this case. One array with both types of values has the same complexity; there should also be `if ` where we will have to decide what to do with string and numeric separately. Maybe some class can solve this, however, I think it is overengineering in this case.

h2o-core/src/test/java/water/rapids/GroupByTest.java
```diff
@@ -329,4 +332,90 @@ private Frame chkTree(String tree, String fname, boolean expectThrow) {
...
+    try {
+      System.out.println("Input frame:");
+      System.out.println(inputFrame.toTwoDimTable().toString());
+
+      Frame resFrame = gbInovcation.run(inputFrame);
+
+      System.out.println("GroupBy result:");
+      System.out.println(resFrame.toTwoDimTable().toString());
+
+      Assert.assertEquals(expectedResFrame.numCols(), resFrame.numCols());
```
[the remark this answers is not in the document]
jancijen: 👍 I added messages to all asserts, and I added unit tests for new methods in `ArrayUtils` - `select` and `occurrenceCount`.

maurever:
So reviews are final, @jancijen, you can merge this PR.

PR #4597, opened 2020-05-11
PUBDEV-6003: Adding Hyperparameter tables to AutoML

angela0xdata:
Adding a table with the hyperparameter and value range.


angela0xdata:
@ledell, I have a couple questions about hard coded models in GBM:

  - `histogram_type` is hard coded to Auto, so I added info about what Auto does rather than just `AUTO`: https://github.com/h2oai/h2o-3/blob/master/h2o-automl/src/main/java/ai/h2o/automl/modeling/GBMStepsProvider.java#L28
  - `learn_rate` is commented out. Should it be removed from the table? I just used the default value, which is 0.1. (Easier to delete than to add): https://github.com/h2oai/h2o-3/blob/master/h2o-automl/src/main/java/ai/h2o/automl/modeling/GBMStepsProvider.java#L28

ledell:
@sebhrusen I'm going to add a few changes, but after that, if you could take a quick look to see if anything looks off, that would be great!

h2o-docs/src/product/automl.rst
```diff
@@ -468,51 +468,103 @@ Information about how to cite the H2O software in general is covered in the `H2O
...
+| ``col_sample_rate``          | ``{0.6, 0.8, 1.0}``                                                                         |
++------------------------------+---------------------------------------------------------------------------------------------+
+| ``col_sample_rate_per_tree`` | ``{0.7, 0.8, 0.9, 1.0}``                                                                    |
++------------------------------+---------------------------------------------------------------------------------------------+
+| ``max_depth``                | ``{5, 10, 15, 20}``                                                                         |          
++------------------------------+---------------------------------------------------------------------------------------------+
+| ``min_rows``                 | ``{0.01, 0.1, 1.0, 3.0, 5.0, 10.0, 15.0, 20.0}``                                            |
++------------------------------+---------------------------------------------------------------------------------------------+
+| ``min_sum_hessian_in_leaf``  | ``{0.01, 0.1, 1.0, 3.0, 5.0, 10.0, 15.0, 20.0}``                                            |
```
sebhrusen, line 504: `min_sum_hessian_in_leaf` has been recently removed (unsupported XGB param) by PUBDEV-7527

h2o-docs/src/product/automl.rst
```diff
@@ -468,51 +468,103 @@ Information about how to cite the H2O software in general is covered in the `H2O
...
+GLM Hyperparameters
+~~~~~~~~~~~~~~~~~~~
 
-**Deep Learning Hyperparameters**
+This table shows the GLM values that are searched over when performing AutoML grid search. Additional information is available `here <https://github.com/h2oai/h2o-3/blob/master/h2o-automl/src/main/java/ai/h2o/automl/modeling/GLMStepsProvider.java>`__.
+
++-----------------------------+---------------------------------------------------------------------------------------------+
+| Parameter                   | Searchable Values                                                                           |
++=============================+=============================================================================================+
+| ``alpha``                   | ``{0.0, 0.2, 0.4, 0.6, 0.8, 1.0}``                                                          |
```
sebhrusen, line 482: technically, we don't run a grid search anymore for GLM, but we build a single model with `lambda_search` enabled, and passing a list of `alpha`s.
GLM algo is supposed to do lambda search for each `alpha` and return only the model with the best `alpha-lambda` combination in that case, whereas a grid would return one model for each `alpha`.

angela0xdata:
@sebhrusen thanks for the review. I'll file a new Jira for your changes since this PR was merged.

PR #4598, opened 2020-05-12
H2ORunner respects doonly & ignored test configuration

Pscheidl:
H2ORunner always runs everything and does not respect the ignored & doonly system properties. This is a problem for smoke tests, as some test suited are annotated with H2ORunner and it runs them all the time.

### DIFF
--- a/h2o-core/src/test/java/water/runner/H2ORunner.java
+++ b/h2o-core/src/test/java/water/runner/H2ORunner.java
@@ -27,8 +27,8 @@ import java.util.Set;
 @Ignore
 public class H2ORunner extends BlockJUnit4ClassRunner {
     private final TestClass testClass;
-    private final HashSet doOnlyTestNames;
-    private final HashSet ignoreTestsNames;
+    private final HashSet<String> doOnlyTestNames;
+    private final HashSet<String> ignoreTestsNames;
 
     /**
      * Creates a BlockJUnit4ClassRunner to run {@code klass}
--- a/h2o-core/src/test/java/water/runner/H2ORunner.java
+++ b/h2o-core/src/test/java/water/runner/H2ORunner.java
@@ -18,6 +18,7 @@ import water.TestUtil;
 import water.Value;
 import water.fvec.Frame;
 import water.fvec.Vec;
+import water.util.IcedHashSet;
 import water.util.Log;
 
 import java.util.HashSet;
@@ -28,6 +29,8 @@ import java.util.Set;
 @Ignore
 public class H2ORunner extends BlockJUnit4ClassRunner {
     private final TestClass testClass;
+    private final HashSet doonlyTestsNames;
+    private final HashSet ignoreTestsNames;
 
     /**
      * Creates a BlockJUnit4ClassRunner to run {@code klass}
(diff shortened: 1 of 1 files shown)


h2o-core/src/test/java/water/runner/H2ORunner.java
```diff
@@ -28,6 +29,8 @@
 @Ignore
 public class H2ORunner extends BlockJUnit4ClassRunner {
     private final TestClass testClass;
+    private final HashSet doonlyTestsNames;
```
Pscheidl, line 32: The tests should be listed in fully qualified names. Therefore, we can use HashSet and do the exact match with `contains` - which is fast. 

TestUtils do `O(n*(i+d))`, where `n` is number of tests (total), `i` is the number of ignored tests listed and `d` is number of do only tests.

h2o-core/src/test/java/water/runner/H2ORunner.java
```diff
@@ -150,14 +156,49 @@ private void printLeakedKeys(final Key[] leakedKeys) {
...
+                    this.ignoreTestsNames.add(ignoredTestName);
+                }
+            }
+        }
+
+        final String doonlyTests = System.getProperty("doonly.tests");
+        if (doonlyTests != null) {
+            final String[] split = doonlyTests.split(",");
+            if (split.length != 1 && !split[0].equals("")) {
+                for (final String ignoredTestName : split) {
```
honzasterba, line 196: confusing variable name

Pscheidl: I'll improve it, thanks.

h2o-core/src/test/java/water/runner/H2ORunner.java
```diff
@@ -150,14 +154,49 @@ private void printLeakedKeys(final Key[] leakedKeys) {
 
     private int fetchCloudSize() {
         final CloudSize annotation = testClass.getAnnotation(CloudSize.class);
-        if (annotation == null) throw new IllegalStateException("@CloudSize annotation is missing for test class: " + testClass.getName());
+        if (annotation == null)
```
Pscheidl, line 157: Formatting.

h2o-core/src/test/java/water/runner/H2ORunner.java
```diff
@@ -150,14 +154,49 @@ private void printLeakedKeys(final Key[] leakedKeys) {
...
+            throw new IllegalStateException("@CloudSize annotation is missing for test class: " + testClass.getName());
 
         final int cloudSize = annotation.value();
 
-        if(cloudSize < 1) throw new IllegalStateException("@CloudSize annotation must specify sizes greater than zero. Given value: " + cloudSize);
+        if (cloudSize < 1)
```
Pscheidl, line 162: Formatting.

h2o-core/src/test/java/water/runner/H2ORunner.java
```diff
@@ -28,6 +27,8 @@
 @Ignore
 public class H2ORunner extends BlockJUnit4ClassRunner {
     private final TestClass testClass;
+    private final HashSet doOnlyTestNames;
```
honzasterba, line 30: why not add <String> generics?

Pscheidl: I also have no idea why I did not put it there. Thanks for noticing it. :man_facepalming: 

Fixed in: https://github.com/h2oai/h2o-3/pull/4598/commits/c74cd201174a6cb0766b1a513c4efed37842218f

PR #4599, opened 2020-05-12
H2ORunner checks cloud size before each test leaf execution

Pscheidl:
When `TestUtil.stall_till_cloud_size` method is called, it checks against `private static boolean _stall_called_before = false;` whether it has already been called or not.

Making sure the cloud size is correct before each leaf is one option - giving us the per-test immediate feedback. Which I'm not sure is needed. But the call is extremely cheap. At this place, it is also made sure this part is not executed if the test is ignored - and we keep the order of method calls defined by original JUnit pipeline.


Pscheidl:
Re-started the tests (remote call failed on multiple tests).

PR #4600, opened 2020-05-12
PUBDEV-6250 Mutliclass PDP

maurever:
JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6250

Implement multinomial PDP. The user can:

1) specify a target class to plot partial dependence of column and target class in one graph
2) specify more than one target class to plot partial dependence of column and target class in one graph

DONE:
- calculation in Java H2O-3 core
- points 1) and 2) for Python and R
- better visualization of categorical values
- fix categorical in R
- Implement PDP plotting also in Flow - in this PR: https://github.com/h2oai/h2o-flow/pull/160


h2o-core/src/main/java/water/api/schemas3/PartialDependenceV3.java
```diff
@@ -58,5 +58,8 @@
   @API(help="Key to store the destination", direction=API.Direction.INPUT)
   public KeyV3.PartialDependenceKeyV3 destination_key;
 
+  @API(help="Target class for multinomial classification", direction=API.Direction.INPUT)
```
michalkurka, line 61: Should this say "target classes"?

h2o-r/h2o-package/R/models.R
```diff
@@ -3927,20 +3928,34 @@ h2o.cross_validation_predictions <- function(object) {
...
+#' # three target classes
+#' h2o.partialPlot(object = iris_gbm, data = iris_hex, cols="Petal.Length", targets=c("setosa", "virginica", "versicolor"))
 #' }
 #' @export
 
 h2o.partialPlot <- function(object, data, cols, destination_key, nbins=20, plot = TRUE, plot_stddev = TRUE,
-                            weight_column=-1, include_na=FALSE, user_splits=NULL, col_pairs_2dpdp=NULL, save_to=NULL,
-row_index=-1) {
+                            weight_column=-1, include_na=FALSE, user_splits=NULL, col_pairs_2dpdp=NULL, save_to=NULL, 
+                            row_index=-1, targets=NULL) {
```
michalkurka, line 3944: was the api validated in #h2o-api-updates?

michalkurka:
LGTM, thank you - very good addition. Please confirm the API change was checked in #h2o-api-updates.

h2o-core/src/main/java/water/api/schemas3/PartialDependenceV3.java
maurever: Yes, it should. Good catch, thank you!

h2o-r/h2o-package/R/models.R
maurever: Nope, I will do.

h2o-core/src/main/java/hex/PartialDependence.java
```diff
@@ -179,30 +218,31 @@ public void compute2() {
...
         double[] col2Vals = workingOn1D ? null : extractColValues(col2, _nbins, fr.vec(col2));
 
         Futures fs = new Futures();
         int responseLength = workingOn1D ? colVals.length : colVals.length * col2Vals.length;
-        final double meanResponse[] = new double[responseLength];
-        final double stddevResponse[] = new double[responseLength];
-        final double stdErrorOfTheMeanResponse[] = new double[responseLength];
+        final double[] meanResponse = new double[responseLength];
+        final double[] stddevResponse = new double[responseLength];
+        final double[] stdErrorOfTheMeanResponse = new double[responseLength];
```
wendycwong, line 236: yes, those are ugly, aren't they?  Thanks for correcting them.

maurever: I just don't like my idea shows me what is bad. Then I need to correct it. :D

PR #4609, opened 2020-05-13
PUBDEV-7481: GLM not running with alpha arrays properly

wendycwong:
This PR aims to fix the problem specified in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7481?filter=-1

It looks like GLM with lambda search enabled only runs with the first alpha value specified in the array.  In addition, each lambda value will start building GLM model with coefficients and other parameters set to the values of the previous GLM model building.

When a user provides an alpha/lambda array, the default setting is to have the system cold start from default initial values for every lambda/alpha value.  However, if a user wants warm start, she can set cold_start=False to force the system to build GLM model starting with the states of the previous GLM model build.

When lambda_search=True, the default is to do warm start unless the user set cold_start=False.

1. Added loop to set alpha value as well as lambda.  Added parametets to keep track of submodels' alpha, submodel index and others.
2. Fixed best model selection for alpha/lambda arrays.
3. Added alpha array to submodel reports.  Finished work for alpha array with empty lambda, alpha and lambda arrays.
4. Add parameter cold_start to allow user to choose starting from initial values if desired.  Here is the default setup: if lambda search is disabled, user can set cold_start=True and every lambda/alpha values will start from initial values.  If user set cold_start=False, next lambda/alpha GLM will start from previous GLM beta values.
5. User can use startval to set GLM initial coefficients if they desired to.

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/ComputationState.java
+++ b/h2o-algos/src/main/java/hex/glm/ComputationState.java
@@ -22,13 +22,14 @@ import java.util.Arrays;
 import java.util.Comparator;
 
 import static hex.glm.GLMUtils.calSmoothNess;
+import static hex.glm.GLMUtils.copyGInfo;
 
 public final class ComputationState {
   final boolean _intercept;
   final int _nclasses;
   private final GLMParameters _parms;
   private BetaConstraint _bc;
-  final double _alpha;
+  double _alpha;
   double[] _ymu;
   double [] _u;
   double [] _z;
@@ -40,6 +41,8 @@ public final class ComputationState {
   private GLMGradientInfo _ginfo; // gradient info excluding l1 penalty
   private double _likelihood;
   private double _gradientErr;
+  private boolean _lambdaNull; // true if lambda was not provided by user
+  private double _gMax; // store max value of original gradient without dividing by math.max(1e-2, _parms._alpha[0])
   private DataInfo _activeData;
   private BetaConstraint _activeBC = null;
   private double[] _beta; // vector of coefficients corresponding to active data
@@ -88,6 +91,7 @@ public final class ComputationState {
     this (job, parms, dinfo, bc, nclasses);
     _penaltyMatrix = penaltyMat;
     _gamBetaIndices = gamColInd;
+    _lambdaNull = (_parms._lambda==null) && !(_parms._lambda_search);
   }
 
 
@@ -146,6 +150,8 @@ public final class ComputationState {
   public double get_tau() {
     return _tau;
   }
+  
+  public boolean getLambdaNull() { return _lambdaNull; }
 
   public void set_tau(double tau) {
     _tau=tau;
@@ -163,9 +169,18 @@ public final class ComputationState {
 
   public GLMGradientSolver gslvr(){return _gslvr;}
   public double lambda(){return _lambda;}
+  public double alpha() {return _alpha;}
   public void setLambdaMax(double lmax) {
     _lambdaMax = lmax;
   }
+  public void setgMax(double gmax) {
+    _gMax = gmax;
+  }
+  public void setAlpha(double alpha) {
+    _alpha=alpha;
+    setLambdaMax(_gMax/Math.max(1e-2,alpha)); // need to set _lmax every time alpha value changes
+  }
+
   public void setLambda(double lambda) {
     adjustToNewLambda(0, _lambda);
     // strong rules are to be applied on the gradient with no l2 penalty
@@ -207,7 +222,9 @@ public final class ComputationState {
   public void dropActiveData(){_activeData = null;}
 
   public String toString() {
-    return "iter=" + _iter + " lmb=" + GLM.lambdaFormatter.format(_lambda) + " obj=" + MathUtils.roundToNDigits(objective(),4) + " imp=" + GLM.lambdaFormatter.format(_relImprovement) + " bdf=" + GLM.lambdaFormatter.format(_betaDiff);
+    return "iter=" + _iter + " lmb=" + GLM.lambdaFormatter.format(_lambda) + " alpha=" + 
+            GLM.lambdaFormatter.format(_alpha)+ " obj=" + MathUtils.roundToNDigits(objective(),4) + " imp=" + 
+            GLM.lambdaFormatter.format(_relImprovement) + " bdf=" + GLM.lambdaFormatter.format(_betaDiff);
   }
 
   private void adjustToNewLambda(double lambdaNew, double lambdaOld) {
@@ -259,7 +276,7 @@ public final class ComputationState {
     int P = _dinfo.fullN();
     _activeBC = _bc;
     _activeData = _activeData != null?_activeData:_dinfo;
-    _allIn = _allIn || _parms._alpha[0]*lambdaNew == 0 || _activeBC.hasBounds();
+    _allIn = _allIn || _alpha*lambdaNew == 0 || _activeBC.hasBounds();
     if (!_allIn) {
       int newlySelected = 0;
       final double rhs = Math.max(0,_alpha * (2 * lambdaNew - lambdaOld));
@@ -270,7 +287,7 @@ public final class ComputationState {
         if(j < oldActiveCols.length && oldActiveCols[j] == i)
           j++;
         else if (_ginfo._gradient[i] > rhs || -_ginfo._gradient[i] > rhs)
-          newCols[newlySelected++] = i;
+          newCols[newlySelected++] = i; // choose active columns here
       }
       if(_parms._max_active_predictors != -1 && (oldActiveCols.length + newlySelected -1) > _parms._max_active_predictors){
         Integer [] bigInts = ArrayUtils.toIntegers(newCols, 0, newlySelected);
@@ -308,8 +325,6 @@ public final class ComputationState {
   public boolean _lsNeeded = false;
 
   private DataInfo [] _activeDataMultinomial;
-//  private int [] _classOffsets = new int[]{0};
-
 
   public DataInfo activeDataMultinomial(int c) {return _activeDataMultinomial != null?_activeDataMultinomial[c]:_dinfo;}
 
@@ -653,6 +668,28 @@ public final class ComputationState {
     _likelihood = ginfo._likelihood;
     return (_relImprovement = (objOld - objective())/Math.abs(objOld));
   }
+  
+  double getBetaDiff() {return _betaDiff;}
+  double getGradientErr() {return _gradientErr;}
+  protected void setBetaDiff(double betaDiff) { _betaDiff = betaDiff; }
+  protected void setGradientErr(double gErr) { _gradientErr = gErr; }
+  protected void setGinfo(GLMGradientInfo ginfo) {
+    _ginfo = copyGInfo(ginfo);
+  }
+  protected void setBeta(double[] beta) {
+    if(_beta == null)_beta = beta.clone();
+    else System.arraycopy(beta,0, _beta, 0, beta.length);
+  }
+  
+  protected void setIter(int iteration) {
+    _iter = iteration;
+  }
+  
+  protected void setLikelihood(double llk) { _likelihood = llk; }
+  protected void setAllIn(boolean val) { _allIn = val; }
+  protected void setGslvrNull() { _gslvr = null; }
+  protected void setActiveDataMultinomialNull() { _activeDataMultinomial = null; }
+  protected void setLambdaSimple(double lambda) { _lambda=lambda; }
 
   protected void setHGLMComputationState(double [] beta, double[] ubeta, double[] psi, double[] phi, 
                                          double hlcorrection, double tau, Frame wpsi, String[] randCoeffNames){
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -102,9 +102,10 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
   @Override public boolean haveMojo() { return true; }
 
   private double _lambdaCVEstimate = Double.NaN; // lambda cross-validation estimate
+  private int _bestCVSubmodel;  // best submodel index found during cv
   private boolean _doInit = true;  // flag setting whether or not to run init
-  private double [] _xval_test_deviances;
-  private double [] _xval_test_sd;
+  private double [] _xval_deviances;
+  private double [] _xval_sd;
 
   /**
    * GLM implementation of N-fold cross-validation.
@@ -124,50 +125,44 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
 
   /**
    * If run with lambda search, we need to take extra action performed after cross-val models are built.
-   * Each of the folds have been computed with ots own private validation datasetd and it performed early stopping based on it.
+   * Each of the folds have been computed with ots own private validation dataset and it performed early stopping based on it.
    * => We need to:
    *   1. compute cross-validated lambda estimate
-   *   2. set the lambda estimate too all n-folds models (might require extra model fitting if the particular model stopped too early!)
+   *   2. set the lambda estimate to all n-folds models (might require extra model fitting if the particular model
+   *   stopped too early!)
    *   3. compute cross-validated scoring history (cross-validated deviance standard error per lambda)
    *   4. unlock the n-folds models (they are changed here, so the unlocking happens here)
    */
   @Override
   public void cv_computeAndSetOptimalParameters(ModelBuilder[] cvModelBuilders) {
     if(_parms._max_runtime_secs != 0) _parms._max_runtime_secs = 0;
-    if(_parms._lambda_search) {
-      _xval_test_deviances = new double[_parms._lambda.length];
-      _xval_test_sd = new double [_parms._lambda.length];
+      _xval_deviances = new double[_parms._lambda.length*_parms._alpha.length];
+      _xval_sd = new double [_parms._lambda.length*_parms._alpha.length];
       double bestTestDev = Double.POSITIVE_INFINITY;
       int lmin_max = 0;
-      for (int i = 0; i < cvModelBuilders.length; ++i) {
+      for (int i = 0; i < cvModelBuilders.length; ++i) {  // find the highest best_submodel_idx we need to go through
         GLM g = (GLM) cvModelBuilders[i];
-        lmin_max = Math.max(lmin_max,g._model._output._best_lambda_idx);
+        lmin_max = Math.max(lmin_max,g._model._output._selected_submodel_idx);
       }
-      int lidx = 0;
-      int bestId = 0;
+      int lidx = 0; // index into submodel
+      int bestId = 0;   // submodel indedx with best Deviance from xval
       int cnt = 0;
-      for (; lidx < lmin_max; ++lidx) {
+      for (; lidx < lmin_max; ++lidx) { // search through submodel with same lambda and alpha values
         double testDev = 0;
-        for (int i = 0; i < cvModelBuilders.length; ++i) {
+        double testDevSq = 0;
+        for (int i = 0; i < cvModelBuilders.length; ++i) {  // run cv for each lambda value
           GLM g = (GLM) cvModelBuilders[i];
-          double x = _parms._lambda[lidx];
-          if (g._model._output.getSubmodel(x) == null)
-            g._driver.computeSubmodel(lidx, x, Double.NaN, Double.NaN);
-          testDev += g._model._output.getSubmodel(x).devianceTest;
+          if (g._model._output._submodels[lidx] != null) {
+            double lambda = g._model._output._submodels[lidx].lambda_value;
+            g._driver.computeSubmodel(lidx, lambda, Double.NaN, Double.NaN);
+            testDev += g._model._output._submodels[lidx].devianceValid;
+            testDevSq += g._model._output._submodels[lidx].devianceValid * g._model._output._submodels[lidx].devianceValid;
+          }
         }
-        double testDevAvg = testDev / cvModelBuilders.length;
-        double testDevSE = 0;
-        // compute deviance standard error
-        for (int i = 0; i < cvModelBuilders.length; ++i) {
-          GLM g = (GLM) cvModelBuilders[i];
-          double x = _parms._lambda[lidx];
-          if (g._model._output.getSubmodel(x) == null)
-            g._driver.computeSubmodel(lidx, x, Double.NaN, Double.NaN); // stopped too early, need to FIT extra model
-          double diff = testDevAvg - (g._model._output.getSubmodel(x).devianceTest);
-          testDevSE += diff*diff;
-        }
-        _xval_test_sd[lidx] = Math.sqrt(testDevSE/((cvModelBuilders.length-1)*cvModelBuilders.length));
-        _xval_test_deviances[lidx] = testDevAvg;
+        double testDevAvg = testDev / cvModelBuilders.length; // average testDevAvg for fixed submodel index
+        double testDevSE = testDevSq - testDevAvg*testDev;
+        _xval_sd[lidx] = Math.sqrt(testDevSE/((cvModelBuilders.length-1)*cvModelBuilders.length));
+        _xval_deviances[lidx] = testDevAvg;
         if(testDevAvg < bestTestDev) {
           bestTestDev = testDevAvg;
           bestId = lidx;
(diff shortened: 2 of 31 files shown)


h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -31,9 +31,16 @@
...
-    @API(help="Lambda minimizing the objective value, only applicable with lambd search")
+    @API(help="Lambda minimizing the objective value, only applicable with lambda search or when arrays of alpha and " +
+            "lambdas are provided")
     double lambda_best;
 
+    @API(help="Alpha minimizing the objective value, only applicable when arrays of alphas are given ")
+    double alpha_best;
+
+    @API(help="submodel index minimizing the objective value, only applicable for arrays of alphas/lambda ")
+    int best_submodel_index; // denote the submodel index that yields the best result
```
sebhrusen, line 42: this is exposed only for tests right? Otherwise I don't see how this could be any use for the user.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1075,12 +1083,20 @@ public void setZValues(double [] zValues, double dispersion, boolean dispersionE
...
+    public int _selected_alpha_idx;     // alpha index which is being worked on
+    public int _selected_submodel_idx;  // submodel index is being worked
+    public double lambda_best(){return _submodels.length == 0 ? -1 : _submodels[_best_submodel_idx].lambda_value;}
     public double dispersion(){ return _dispersion;}
+    public double alpha_best() { return _submodels.length == 0 ? -1 : _submodels[_best_submodel_idx].alpha_value;}
     public double lambda_1se(){return _lambda_1se == -1 || _lambda_1se >= _submodels.length?-1:_submodels.length == 0 ? -1 : _submodels[_lambda_1se].lambda_value;}
-    public double lambda_selected(){return _submodels[_selected_lambda_idx].lambda_value;}
+    public int bestSubmodelIndex() { return _best_submodel_idx; }
+    public double lambda_selected(){
+      return _submodels[_best_submodel_idx].lambda_value;
```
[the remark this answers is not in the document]
wendycwong: @sebhrusen Yes, the _best_submodel_idx will  be available to you so that you know which submodel index gives the best results.

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
wendycwong: This index is useful when you do a glm.getGLMRegularizationPath(mL), this index will be the one that gives the best performance.

h2o-algos/src/main/java/hex/glm/GLMUtils.java
```diff
@@ -30,6 +30,13 @@
     }
     return gamColIndices;
   }
+  
+  public static GLM.GLMGradientInfo copyginfo(GLM.GLMGradientInfo ginfo) {
+    double[] gradient = new double[ginfo._gradient.length];
+    System.arraycopy(ginfo._gradient, 0, gradient, 0, gradient.length);
```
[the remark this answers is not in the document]
wendycwong: I changed it anyway.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2143,39 +2186,74 @@ public void computeImpl() {
...
+        _betaDiffStart = _state.getBetaDiff();
+      }
+      // alpha, lambda search loop
+      int submodelCount = 0;
+      for (int alphaInd = 0; alphaInd < _parms._alpha.length; alphaInd++) {
+        _state.setAlpha(_parms._alpha[alphaInd]);   // loop through the alphas
+        if ((!_parms._HGLM) && (alphaInd > 0)) // no need for cold start during the first iteration
+          coldStart(devHistoryTrain, devHistoryTest);  // reset beta, lambda, currGram
+        _model._output._selected_alpha_idx = alphaInd;
+       // if (_state.getLambdaNull()) _parms._lambda[0] = _gmax/Math.max(1e-2, _parms._alpha[alphaInd]);
```
honzasterba, line 2201: commented code

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1263,18 +1279,40 @@ public int nclasses() {
...
-      int bestId = 0;
-      Submodel best = _submodels[0];
-      for(int i = 1; i < _submodels.length; ++i) {
-        Submodel sm = _submodels[i];
-        if(!(sm.devianceTest > best.devianceTest) && sm.devianceTrain < best.devianceTrain){
-          bestId = i;
-          best = sm;
+    public double selectBestSubmodel(double trainDev, double testDev, double[] xval_test_deviances, int alphaIndex,
+                                     int lambdaIndex, int submodelIdx, double bestDevValue, boolean setAlphaLambdaIdx) {
+      // if testDev exists, use it above everything else;
```
sebhrusen, line 1284: by testDev, you mean validationDev?
test data should not be used in the selection process.
But if you mean validation, then usually xval takes precedence over validation, at least that's what we do with metrics.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1263,18 +1279,40 @@ public int nclasses() {
...
+      // else if xval exists, use it above trainDev
+      // use trainDev if that is all that is available
+      boolean foundBetterDev = false;
+      double bestValue = bestDevValue;
+      if (!Double.isNaN(testDev) && (testDev < bestDevValue)) { // use validation metrics
```
sebhrusen, line 1289: doesn't look right: if we have a `testDev` but it's larger that `bestDevValue`, then we fall back to xval or `trainDev`, which itself may be lower that `bestDevValue`: so we provide sth with high precedence, but the lower precedence deviance will be used at the end...
Looks like the logic should be split into 2 steps:
1. select the deviance that takes precedence: `testDev`, `trainDev`, `xvalDev`.
2. compare its value with `bestDev`.

h2o-algos/src/main/java/hex/glm/GLMUtils.java
```diff
@@ -30,6 +30,12 @@
     }
     return gamColIndices;
   }
+  
+  public static GLM.GLMGradientInfo copyginfo(GLM.GLMGradientInfo ginfo) {
```
sebhrusen, line 34: camel case won't hurt readability here: `copyGInfo`

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -4283,3 +4283,78 @@ def saveModelMojo(model):
     os.makedirs(tmpdir)
     model.download_mojo(path=tmpdir)    # save mojo
     return tmpdir
+
```
sebhrusen, line 4286: don't know if it's a good idea to put those functions in bloated `utilsPy`.
This file is already full of very specific functions that are used by only a few tests.
Wouldn't it be nicer to have a utility module in the `testdir_algos/glm` module?

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7481_glm_startvals.py
```diff
@@ -0,0 +1,48 @@
...
+# test startval to set GLM coefficients
+def set_glm_startvals():
+    # read in the dataset and construct training set (and validation set)
+    d = h2o.import_file(path=pyunit_utils.locate("smalldata/logreg/prostate.csv"))
+    mL = glm(family='binomial')
+    mL.train(training_frame=d,x=[2,3,4,5,6,7,8],y=1)
+    mLcoeff = mL.coef()
+    r = glm.getGLMRegularizationPath(mL)
+    rcoeff = r["coefficients"][0]
+    startval1 = [0,0,0,0,0,0,0,-0.3945120960889672]
```
sebhrusen, line 16: `-0.3945120960889672`? where does this come from?

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7481_glm_startvals.py
```diff
@@ -0,0 +1,48 @@
...
+    ml1Coeff = ml1.coef()
+    pyunit_utils.assertEqualCoeffDicts(mLcoeff, ml1Coeff , tol = 1e-6) # coeffs should be the same
+
+    ml2 = glm(family="binomial", startval = startval2) # different starting condition from GLM
+    ml2.train(training_frame=d,x=[2,3,4,5,6,7,8],y=1)
+    ml2Coeff = ml2.coef()   
+    
+    try:
+        pyunit_utils.assertEqualCoeffDicts(mLcoeff, ml2Coeff , tol = 1e-6)
+        sys.exit("Should have thrown an error as coefficients are different!")
```
sebhrusen, line 32: no `sys.exit` please, there may be more tests to run...
this is the pattern:
```
try:
    # statement expected to raise
    assert False, "Should have thrown an error as coefficients are different!"
except Exception as expected:
    # verify that the exception that has been thrown is the one you expect
    assert "expected message" in str(expected)
```

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7481_glm_startvals.py
```diff
@@ -0,0 +1,48 @@
...
+    except Exception as ex:
+        print(ex)
+    
+    try:
+        mlbad =  glm(family="binomial", startval = startvalBad)
+        mlbad.train(training_frame=d,x=[2,3,4,5,6,7,8],y=1)
+        sys.exit("Should have thrown an error with bad GLM initial values!")
+    except Exception as ex:
+        print(ex)
+        print("Test completed!  Success!")
```
sebhrusen, line 42: same as above

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7481_lambda_alpha_array_multinomial_coldStart.py
```diff
@@ -0,0 +1,63 @@
...
+    coefClassSetNorm = ['std_coefs_class_0','std_coefs_class_1', 'std_coefs_class_2', 'std_coefs_class_3', 
+                        'std_coefs_class_4', 'std_coefs_class_5', 'std_coefs_class_6', 'std_coefs_class_7']
+    startVal = [0]*371
+    startVal[52] = -1.9166426513747452
+    startVal[105] = -1.200312843494449
+    startVal[158] = -2.2256240518579173
+    startVal[211] = -2.2256240518579173
+    startVal[264] = -2.121931593300788
+    startVal[317] = -2.2256240518579173
+    startVal[370] = -2.2256240518579173
```
sebhrusen, line 40: is it possible to write tests without those magic numbers?

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -653,6 +666,28 @@ protected double updateState(double [] beta,GLMGradientInfo ginfo){
...
+  double getBetaDiff() {return _betaDiff;}
+  double getGradientErr() {return _gradientErr;}
+  protected void setBetaDiff(double betaDiff) { _betaDiff = betaDiff; }
+  protected void setGradientErr(double gErr) { _gradientErr = gErr; }
+  protected void setGinfo(GLMGradientInfo ginfo) {
+    _ginfo = copyginfo(ginfo);
+  }
+  protected void setBeta(double[] beta) {
+    if(_beta == null)_beta = beta.clone();
+    else System.arraycopy(beta,0,_beta,0,beta.length);
```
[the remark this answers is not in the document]
wendycwong: @sebhrusen No, I am not angry with spaces.  I am angry with my life.  I have not had a vacation since X'mas!

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: @sebhrusen Here is the deal, during xval, the training frame is split into training folds and validation folds.  The testDev will refer to the results from the validation.  During the final main model training, if the lambdaCVestimate is not null, the xval decision will be used to determine the best model.  So, your wish has come true in this case.

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: @sebhrusen That is bug.  I need to fix it.

h2o-algos/src/main/java/hex/glm/GLMUtils.java
wendycwong: yes, I will change it.

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7481_glm_startvals.py
wendycwong: @sebhrusen I cheated here and look into glm.java computeImpl method and check what the initial coefficient values are and put it there.

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7481_glm_startvals.py
wendycwong: I will use assert False then.

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7481_glm_startvals.py
wendycwong: got it

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: Yeah, it is good to refractor it.

h2o-algos/src/main/java/hex/glm/GLMModel.java
sebhrusen: ok, can you please rename to `validationDev` then to avoid confusion?

h2o-algos/src/main/java/hex/glm/ComputationState.java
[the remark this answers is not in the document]
wendycwong: I added the space.  Thanks.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -375,10 +370,14 @@ public synchronized TwoDimTable to2dTable() {
   private transient double _iceptAdjust = 0;
 
   private double _lmax;
+  private double _gmax; // gradient max without dividing by math.max(1e-2, _parms._alpha[0])
   private transient long _nobs;
   private transient GLMModel _model;
   private boolean _earlyStop = false;  // set by earlyStopping
-
+  GLMGradientInfo _ginfoStart;
```
honzasterba, line 377: why not private?

h2o-algos/src/main/java/hex/schemas/GLMV3.java
```diff
@@ -133,6 +134,9 @@
     @API(help = "Standardize numeric columns to have zero mean and unit variance", level = Level.critical)
     public boolean standardize;
 
+    @API(help = "If true will start GLM model from scratch, otherwise, will build from previous results", level = Level.critical)
```
honzasterba, line 137: the doc needs improvements, every algo starts from scratch at the begging (unless checpoints are used), I guess this is related to the lambda search loop?

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -207,7 +221,8 @@ public DataInfo activeData(){
   public void dropActiveData(){_activeData = null;}
 
   public String toString() {
-    return "iter=" + _iter + " lmb=" + GLM.lambdaFormatter.format(_lambda) + " obj=" + MathUtils.roundToNDigits(objective(),4) + " imp=" + GLM.lambdaFormatter.format(_relImprovement) + " bdf=" + GLM.lambdaFormatter.format(_betaDiff);
+    return "iter=" + _iter + " lmb=" + GLM.lambdaFormatter.format(_lambda) + " alpha=" +
```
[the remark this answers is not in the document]
wendycwong: Good idea.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -126,41 +126,41 @@ public void computeCrossValidation() {
...
    *   1. compute cross-validated lambda estimate
-   *   2. set the lambda estimate too all n-folds models (might require extra model fitting if the particular model stopped too early!)
+   *   2. set the lambda estimate to all n-folds models (might require extra model fitting if the particular model
+   *   stopped too early!)
    *   3. compute cross-validated scoring history (cross-validated deviance standard error per lambda)
    *   4. unlock the n-folds models (they are changed here, so the unlocking happens here)
    */
   @Override
-  public void cv_computeAndSetOptimalParameters(ModelBuilder[] cvModelBuilders) {
+  public void cv_computeAndSetOptimalParameters(ModelBuilder[] cvModelBuilders) { // fix this so that it will work with both alpha and lambda array
```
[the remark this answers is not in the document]
wendycwong: A to do.  I have finished it though.

h2o-algos/src/main/java/hex/schemas/GLMV3.java
wendycwong: That is correct.  Lambda search starts from old computationState values from last lambda value

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: I removed the function but did try to rename the other test deviance to valid deviance.

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: I removed the whole bestValue thing and just use what was there before.

h2o-algos/src/main/java/hex/glm/GLMUtils.java
wendycwong: Done

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7481_glm_startvals.py
wendycwong: Done

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -558,6 +557,8 @@ public void init(boolean expensive) {
       if (_valid != null)
         _validDinfo = _dinfo.validDinfo(_valid);
       _state = new ComputationState(_job, _parms, _dinfo, null, nclasses(), _penaltyMatrix, _gamColIndices);
+      _state.setLambdaNull((_parms._lambda==null) && !(_parms._lambda_search));
```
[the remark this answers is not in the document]
wendycwong: You are right

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: They are private now.

h2o-algos/src/main/java/hex/schemas/GLMV3.java
wendycwong: I changed the description.  Please take a look now.

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7481_glm_startvals.py
wendycwong: I actually have the test generate the initial intercept values nowl

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7481_lambda_alpha_array_multinomial_coldStart.py
wendycwong: Challenge accepted and completed.

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: I removed the function that I used to glmUtils.py.  I did not put it in testdir_algos/glm because the run.py may not like it.

honzasterba:
rebase to rel-zeno

h2o-r/h2o-package/R/glm.R
```diff
@@ -83,12 +83,15 @@
...
-#' @param startval double array to initialize fixed and random coefficients for HGLM.
+#' @param startval double array to initialize fixed and random coefficients for HGLM, coefficients for GLM.
 #' @param calc_like \code{Logical}. if true, will return likelihood function value for HGLM. Defaults to FALSE.
 #' @param HGLM \code{Logical}. If set to true, will return HGLM model.  Otherwise, normal GLM model will be returned Defaults
 #'        to FALSE.
 #' @param prior Prior probability for y==1. To be used only for logistic regression iff the data has been sampled and the mean
 #'        of response does not reflect reality. Defaults to -1.
+#' @param cold_start \code{Logical}. Only applicable to multiple alpha/lambda values.  If false, build the next model for next set
+#'        of alpha/lambda values starting from the values provided by current model.  If true will start GLM model from
+#'        scratch.  Default to false. Defaults to FALSE.
```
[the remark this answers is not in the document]
wendycwong: Hi @ledell 

In the original implementation, when lambda_search is enabled, for each lambda value, we actually do not start at the beginning.  Rather we start at the state we were in with computationState and continue onwards.  There is a reason for that.  Lambda search will start with big lambda value and then gradually decrease the lambda value.  As the lambda value drops, we are hoping to see more coefficients not being forced to zero and so on and so forth.  However, if we are just trying out a bunch of alpha and lambda combinations in no particular order, the assumption for lambda_search will be false.  Hence, I have added the cold_start variable and it is initialized to FALSE to ensure the default behavior is the same.  This should not affect your stack ensemble assumptions.  I will check the R implementation to make sure it is also defaulted to false.  Thank you for checking it.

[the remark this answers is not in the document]
ledell: Hi @wendycwong 

- With my first comment, I just mean that the code contains a typo -- it says it defaults to false, twice in that line: `"Default to false. Defaults to FALSE"`.  
- Did you understand my second question about where the changes should be made?  In the gen file and not the glm.R file directly?  I couldn't see any changes to `gen_glm.py` (which generates `glm.R`).

wendycwong:
@michalk:  I finished the benchmark on my branch with master.  In terms of performance (AUC, Logloss, ..), no change.  In terms of training time, no changes either.  Here is a summary of the table of comparison:

image

h2o-r/h2o-package/R/glm.R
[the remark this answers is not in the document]
wendycwong: @erin:  I speak without reading your comments somehow.  That is bad on my part.  Yes, I finally understand what your concern is.  The Default to false is repeated somehow.  Let me look and I need to fix it.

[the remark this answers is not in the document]
wendycwong: @erin:  It looks like R just grab the comments in GLMV3.java and then it tags on Default to FALSE from the default value.  I have added in the comments of GLMV3.java that is is default to false.  Hence, you have this two default to false statement there.  I removed the comment in GLMV3.java and this should fix the problem.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2152,45 +2175,70 @@ public void computeImpl() {
...
+        _state.setAlpha(_parms._alpha[alphaInd]);   // loop through the alphas
+        if ((!_parms._HGLM) && (alphaInd > 0)) // no need for cold start during the first iteration
+          coldStart(devHistoryTrain, devHistoryTest);  // reset beta, lambda, currGram
+        for (int i = 0; i < _parms._lambda.length; ++i) {  // for lambda search, can quit before it is done
+          if (_job.stop_requested() || (timeout() && _model._output._submodels.length > 0))
+            break;  //need at least one submodel on timeout to avoid issues.
+          if (_parms._max_iterations != -1 && _state._iter >= _parms._max_iterations) 
+            break;// iterations accumulate across all lambda/alpha values
+          if ((!_parms._HGLM && (_parms._cold_start || (!_parms._lambda_search && _parms._cold_start))) && (i > 0)) // default: cold_start for non lambda_search
+            coldStart(devHistoryTrain, devHistoryTest);
```
michalkurka, line 2201: coldStart will reset _state._iter to 0, is that correct? how are we counting the number of iterations if it is reset here?

wendycwong: The meaning of cold start is to treat each parameter combination of alpha and lambda as a new model that has nothing to do with the previous model being built.  Hence, apart from reseting the coefficients, gram matrix and others, we also need to reset the model.iteration number.  Otherwise, we are counting the iteration number as a cumulation of all the previous model's iteration counts which I think is incorrect.  However, during lambda_search or when cold_start is not set, the model iteration number will accumulate across all models built.

wendycwong: @michalk:  When you give me an array of alpha and lambda values and cold_start = True, I assume you want me to return a number of GLM models that are built independent of the other GLM models.  You can still control the max_iterations such that each model being built will not exceed that max_iterations.  As far as I am concerned, when you set cold-start = true, you are looking for independent models built and returned to you.  Does this make sense to you?

ledell:
Hey @wendycwong, I saw that this is merged, but I think you missed some of my proposed changes?

- I don't see the `cold_start` param in `gen_glm.py`
- The `cold_start` param does not appear in the user guide.

PR #4610, opened 2020-05-13
PUBDEV-7139 Feature Importance for Black-box Models

ardKelmendi:
Implementation Feature Importance for Black Box models. 

https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7139


h2o-core/src/test/java/water/rapids/ast/prims/FeatureImportance4BBMTest.java
```diff
@@ -211,6 +210,116 @@ public void Classification() throws Exception {
...
+
+            // Done building model; produce a score column with predictions
+            Frame scored = gbm.score(inputFrame);
+            Scope.track(scored);
+            
+            Scope.track(inputFrame);
+//            Scope.track(expectedOutputFrame);
+
+            
+            Val resVal = Rapids.exec("(Perm_Feature_importance " + frameName + " gbm" + ")", sess); // FIXME: find how to pass the model
```
ardKelmendi, line 314: Uncertain how to pass the model to AstFeatureImportance4BBM.java class

h2o-py/tests/testdir_jira/pyunit_pubdev_6504.py
```diff
@@ -0,0 +1,85 @@
...
+    # scikit
+    auc_sci = roc_auc_score(testDataResponse, gbm_sci.predict_proba(testDataFeatures)[:,1])
+
+    # h2o
+    gbm_perf = gbm_h2o.model_performance(prostate_test)
+    auc_h2o = gbm_perf.auc()
+    
+    gbm_h2o.model_performance()
+    
+    model_reliance(prostate_test, gbm_sci)
```
maurever, line 73: Are you sure to pass gbm_sci here and not gbm_h2o?

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+    public Frame _ouput;
+    private Frame _train_fr; // original dataset used to train model
+    private Frame _validate_dataset; // original dataset used to train model
+    private Frame _predcited_frame;
+
+    private boolean _data_in_holdOut_form; // do I have the data split in train and validation, if not spliting will happen
+
+    private String _responce_col;
+
+    private AUC2 og_auc;
```
maurever, line 28: Not every model as the AUC metric.

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private double original_mse;
+    private double sq_err;
+            
+    public FeatureImportance4BBM(Model model) {_model = model;}
```
maurever, line 33: I suggest changing the name to not contain a number. If you are implementing Permutation Feature importance, you can use this naming..

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public FeatureImportance4BBM(Model model, Frame data_set, Frame prediction_frame) {
+        _model = model;
+        _train_fr = data_set;
+        _data_in_holdOut_form = false;
+        _validate_dataset = null;
```
maurever, line 39: If you want to use the validation dataset, you also should think about cross-validation datasets... Maybe for the beginning, only training dataset is enough.

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+        _predcited_frame = prediction_frame;
+        _responce_col = _train_fr.name(0); // first column as default
+    }
+
+    public FeatureImportance4BBM(Model model, Frame data_set, Frame prediction_frame, String response_col) {
```
maurever, line 44: The response column is not necessary to put. You can find it via model output.

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+        _data_in_holdOut_form = false;
+        _validate_dataset = null;
+        _predcited_frame = prediction_frame;
+        _responce_col = response_col;
+    }
+    
+    
+    //MSE set as default loss function
+    public void getFeatureImportance(){
+        getFeatureImportance("MSE");
```
maurever, line 56: MSE works only for regression. For classification, you need to have default loss for example log loss.

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+        if (!_data_in_holdOut_form) {
+            //TODO split the data if needed
+        }
+
+        String[] features = _train_fr.names();
+        HashMap<String, Double> FI = new HashMap<String, Double>(_train_fr.vecs().length);
+
+        // original error calc using MSE or AUC as a loss function
+        if (metric.equals("MSE")){
+            sq_err = new MathUtils.SquareError().doAll(_train_fr.vecs()[1], _predcited_frame.vecs()[0])._sum;
```
maurever, line 69: You cannot compute the metric, in model output, there are all available metrics computed during training.

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+            original_mse = sq_err / _predcited_frame.numRows();
+        }   else    {
+            og_auc = new AUC2(_train_fr.vec(_responce_col), _predcited_frame.vecs()[0]); //claisifiication metric
+        }
+        for (int f = 1; f < _train_fr.numCols(); f++) //f = 0 is predicted col
```
maurever, line 74: You expect here the response column is always at zero index, however, it always cannot be. Users can pass the dataset with a response column at a random index. The response column name is saved in model outputs.

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+            Vec shuffled_feature = ShuffleVec.ShuffleTask.shuffle(_train_fr.vec(features[f]));
+            Vec og_feature = _train_fr.replace(f, shuffled_feature);
+            DKV.put(_train_fr); // "Caller must perform global update (DKV.put) on this updated frame"
+
+            // score the model again and compute diff
+            Frame new_score = _model.score(_train_fr);
+
+            //calucalate error from new score prediction
+            if (metric.equals("MSE")){ // regression
+                sq_err = new MathUtils.SquareError().doAll(_train_fr.vec(_responce_col), new_score.vecs()[0])._sum;
```
maurever, line 86: Typo 
```suggestion
                sq_err = new MathUtils.SquareError().doAll(_train_fr.vec(_response_col), new_score.vecs()[0])._sum;
```

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+            original_mse = sq_err / _predcited_frame.numRows();
+        }   else    {
+            og_auc = new AUC2(_train_fr.vec(_responce_col), _predcited_frame.vecs()[0]); //claisifiication metric
+        }
+        for (int f = 1; f < _train_fr.numCols(); f++) //f = 0 is predicted col
+        {
+            //permuate vector
+            Vec shuffled_feature = ShuffleVec.ShuffleTask.shuffle(_train_fr.vec(features[f]));
+            Vec og_feature = _train_fr.replace(f, shuffled_feature);
+            DKV.put(_train_fr); // "Caller must perform global update (DKV.put) on this updated frame"
```
maurever, line 79: You should make a copy of the frame here.

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+                FI.put(features[f], og_auc._auc - auc._auc);
+            }
+            
+            //return the original data
+            _train_fr.replace(f, og_feature);
+            DKV.put(_train_fr); // "Caller must perform global update (DKV.put) on this updated frame"
+
+            new_score.remove(); // clearing leaks
+        }
+        /*return */HashMap<String, Double> sorted_FI_mse = sortHashMapByValues(FI);
```
maurever, line 99: Why are you use HashMap here? There should be similar implementation on how to sort feature importance output for example in tree algorithms. Try to find it.

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    // Sort HashMap in descenting order
+    private LinkedHashMap<String, Double> sortHashMapByValues(
```
maurever, line 103: This looks very expensive...

h2o-core/src/main/java/water/rapids/ShuffleVec.java
```diff
@@ -0,0 +1,76 @@
...
+    String _col_name;
+    
+    
+    
+    public ShuffleVec(Frame fr) { 
+        _fr = fr;    
+    }
+    
+    /* create a frame with shuffled columns */
+    public /*Frame*/ void ShuffleFrameVecs(int features_no, Frame out){
```
maurever, line 26: How `Frame out` is planned to use here?

h2o-core/src/main/java/water/rapids/ast/prims/models/AstFeatureImportance4BBM.java
```diff
@@ -0,0 +1,54 @@
...
+    public String str() {return "Perm_Feature_importance";} // or call it model_reliance 
+    
+    @Override
+    public ValFrame apply(Env env, Env.StackHelp stk, AstRoot asts[]){
+        Frame fr = stk.track(asts[1].exec(env)).getFrame();
+        Model model = stk.track(asts[2].exec(env)).getModel();
+
+        Frame fr2 = model.score(fr); // prediction on original dataset
+        
+//      model._train.get(); TODO find a way to avoid passing the frame and get it from the model instead
```
maurever, line 38: It would be nice to add the possibility of using the training dataset from the model. There is its id in model output. However, sometimes it could be useful to use different dataset so that I would keep both possibilities. For example, if the input dataset is null, use the training dataset from the model.

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -838,4 +838,66 @@ public void map(Chunk c, NewChunk nc) {
...
 
+  /**
+   * shuffle values of a Vec {@link Vec}.
+   * 
+   * The 'new' values of the vector will be the already exisitng one's however randomly shuffeled
+   * 
+   * 
+   */
+  
+  public static class shuffleVec extends MRTask<shuffleVec>{
```
maurever, line 849: Why are you have two implementations of ShuffleVec? VecUtils is probably a better place to place the code.

maurever, line 849:
```suggestion
  public static class ShuffleVec extends MRTask<ShuffleVec>{
```

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -838,4 +838,66 @@ public void map(Chunk c, NewChunk nc) {
...
+    public Vec _vec;
+        
+    public shuffleVec(Vec vec) {this._vec = vec;}
+        
+        @Override
+        public void map(Chunk ch/*, NewChunk nc*/) {
```
maurever, line 855: Yes, change the original chunk is probably better. Otherwise, you should deal with the rows which did not be affected by random shuffling.

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -838,4 +838,66 @@ public void map(Chunk c, NewChunk nc) {
...
+        }
+            
+        @Override 
+        public void reduce(shuffleVec mrt) {
+            if( _vec != mrt._vec ) _vec = mrt._vec;
+        }
+
+    }
+
+    public static class shuffleVecVol2 extends MRTask<shuffleVecVol2>{
```
maurever, line 874: This MRTask won't work. The _result is limited by double size.

h2o-py/h2o/model/model_reliance.py
```diff
@@ -0,0 +1,25 @@
...
+
+:copyright: (c) 2020 H2O.ai
+:license:   Apache License Version 2.0 (see LICENSE for details)
+"""
+
+from h2o.frame import H2OFrame
+from h2o.expr import ExprNode
+
+
+def model_reliance(frame, model):
```
maurever, line 13: Why model_reliance?

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
ardKelmendi: Yes, good point. This was the initial name I put, will change it when refactoring

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
ardKelmendi: It has been resolved but not pushed yet.

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
ardKelmendi: Resolved

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
ardKelmendi: It is. That's why I was talking about, and working on, creating a `TwoDimTable` instead.
Just wanted to have a nice way to see (feature) -> (importance) mapping

h2o-core/src/main/java/water/rapids/ast/prims/models/AstFeatureImportance4BBM.java
ardKelmendi: hmm, it certainly would, good point.

h2o-core/src/main/java/water/util/VecUtils.java
ardKelmendi: I would call it brainstorming :)

Yes upon refactoring that's where I will put it. I had an Idea to shuffle the whole training data Frame as "pre-processing" therefore I wanted to store the Shuffled Frame on a class and then only use swapping methods between two frames. However, not pursuing that idea.

h2o-core/src/main/java/water/util/VecUtils.java
ardKelmendi: This was my first attempt, it's horrible.

h2o-py/h2o/model/model_reliance.py
ardKelmendi: Got motivated by literature I was reading. I will rename it something along Permutation Feature Importance

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -838,4 +838,66 @@ public void map(Chunk c, NewChunk nc) {
...
+   * 
+   * The 'new' values of the vector will be the already exisitng one's however randomly shuffeled
+   * 
+   * 
+   */
+  
+  public static class shuffleVec extends MRTask<shuffleVec>{
+    public Vec _vec;
+        
+    public shuffleVec(Vec vec) {this._vec = vec;}
```
maurever, line 852: 1. You are calling this MR task with the input vector `vec` which you are saving in every node here. For example, if you run your code with 4 nodes, every node gets its own copy of `vec`. For reading it is no problem, however for writing it is not a good idea (for writing to a vec it is better to use input like `NewChunk nc`).

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -838,4 +838,66 @@ public void map(Chunk c, NewChunk nc) {
...
+        
+        @Override
+        public void map(Chunk ch/*, NewChunk nc*/) {
+//          nc.alloc_doubles(ch._len);
+                
+            for (int i = 0; i < ch._len; ++i) {
```
maurever, line 858: 2. Here you are going through the rows of chunk and modify only `ch._len` numbers of rows in the` _vec`. But, `ch._len` is usually smaller than `_vec.len`. So you change here only **the first** `ch._len` number of rows in the `_vec`.

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -838,4 +838,66 @@ public void map(Chunk c, NewChunk nc) {
...
+                double vecVal = ch.atd(i);
+                double nextvecVal = ch.atd(randIndex);
+                _vec.set(i, nextvecVal);      // this
+                _vec.set(randIndex, vecVal);  // and this lines must be expensive! *For now leaving it like this
+            }
+        }
+            
+        @Override 
+        public void reduce(shuffleVec mrt) {
+            if( _vec != mrt._vec ) _vec = mrt._vec;
```
maurever, line 869: 3. In the reduce phase, you always replace the whole changed copy of `_vec` from one node to another changed copy of `_vec` from another node. When you try to run this code on big data, it will return you a dataset, where the first `ch._len` of rows will be shuffled, but other rows will be in the same order.

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -838,4 +838,66 @@ public void map(Chunk c, NewChunk nc) {
...
+        }
+
+    }
```
maurever, line 872: So depends on 1., 2. and 3. you need to rewrite your MRTask to shuffle only values in the input chunk. You can shuffle the original chunk, or you have to use `NewChunk nc` to save the output, but then you should deal with rows that can be unchanged. In the reduce phase all chunks are connected back to the final frame.  If you dont want random shuffle some values across chunks, you cannot implement `public void reduce(shuffleVec mrt)` method.

maurever: A simple example to see what is happening insight MRTask: 

You have input vector [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12] and three nodes. The chunk size is 4. 

- node 1 gets chunk with [1, 2, 3, 4] data
- node 2 gets chunk with [5, 6, 7, 8] data
- node 3 gets chunk with [9, 10, 11, 12] data.

Every node randomly shuffles its data. So, for example:

- node 1 has now shuffled chunk with  [3, 4, 1, 2] data
- node 2 has now shuffled chunk with [7, 5, 8, 6] data
- node 3has now shuffled chunk with [12, 11, 9, 10] data.

In the reduce phase,  all chunks are connected to one result frame: [3, 4, 1, 2, 7, 5, 8, 6, 12, 11, 9, 10], or a few items can be shuffled across all chunks. 

This is example of just a naive implementation to demonstrate how it works. I hope there are a plenty of algorithms how to distributed shuffle a vector.

h2o-core/src/test/java/water/rapids/ast/prims/PermutationFeatureImportanceTest.java
```diff
@@ -0,0 +1,328 @@
...
+        }
+        return res;
+    }
+    
+    @Test
+    public void shuffleVecBiggerData(){
+        Frame fr = null;
+        try {
+            Scope.enter(); 
+            fr = parse_test_file("./smalldata/airlines/AirlinesTrainWgt.csv"); // has 24422; should be enough to test speed
```
maurever, line 97: This file is not so big to test performance let alone MRTask is working. The number of rows is so small to create more than one chunk.

h2o-core/src/test/java/water/rapids/ast/prims/PermutationFeatureImportanceTest.java
```diff
@@ -0,0 +1,328 @@
...
+        } finally {
+            Scope.exit();
+        }
+
+    }
+   
+    
+    @Test
+    public void Regression() {
+        GBMModel gbm = null;
```
maurever, line 119: Maybe it is better to test results with the GLM model, which is more straightforward because it is base on linear regression.

h2o-py/h2o/model/PermutationFeatureImportance.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def permutation_featue_importance(frame, model):
+    
+    if type(frame) is H2OFrame:
+        m_frame = H2OFrame._expr(ExprNode("Perm_Feature_importance", frame, model.model_id))
+    else:
+        raise ValueError("Frame is not H2OFrame")
+    
+    # TODO
```
maurever, line 20: If I expect correct, this return H2OFrame because you are returning it in your Ast. You can inspire how to deal with this in https://github.com/h2oai/h2o-3/blob/9c86f1f64bf25027490d4781f5e16c8ba987e96e/h2o-py/h2o/model/model_base.py#L476. You can use frame here, you cannot use TwoDimTable here. Try to debug how the structure from Ast looks; it is always some JSON which is formatted to predefined Python or R structures.

ledell:
I know this is a work-in-progress, but I wanted to make a few comments about the main function, `permutation_featue_importance`:

- There's a typo -- it should be "feature" not "featue"
- We already use the term, "variable importance" in H2O, instead of "feature importance", so I think it would make sense to keep that convention.
- Our current function is called `varimp()`.  So maybe a new feature could be called, `permutation_varimp`, or `permutation_importance` (that's what [sklearn](https://scikit-learn.org/stable/modules/permutation_importance.html) calls it).
- Lastly, since the data frame used to generate the importance is a "validation frame" of sorts, maybe we could use the term, `validation_frame` instead of `frame`?

h2o-core/src/main/java/water/rapids/FeatureImportance4BBM.java
ardKelmendi: Now I check the following way based on model:
```_model._output.isBinomialClassifier()``` : AUC
```_model._output.isMultinomialClassifier()``` : logloss
```_model._output.isSupervised()``` : MSE

File: h2o-core/src/main/java/water/rapids/PermutationFeatureImportance.java [hunk 27]
```diff
@@ -0,0 +1,244 @@
...
+        // all values will be doubles
+        String [] metric_values = new String[_train_fr.numCols() - 1]; // minus response col
+        Arrays.fill(metric_values, "double");
+
+        // printed as "%lf"
+        String [] metric_values_f = new String[_train_fr.numCols() - 1]; // minus response col
+        Arrays.fill(metric_values_f, "%lf");
+        
+        // `_training_metrics` table has these metrics
+        String [] Rows = new String[]{"RMSE", "LogLoss", "AUC", "pr_auc", "Lift", "Classification Error"};
```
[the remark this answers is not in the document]
ardKelmendi: For table creation I now use ```ModelMetrics.calcVarImp(...)```

h2o-core/src/main/java/water/rapids/ShuffleVec.java
ardKelmendi: Initial Idea was a bit different. It is removed

h2o-core/src/main/java/water/util/VecUtils.java
ardKelmendi: So instead of passing ```Chunk ch``` I believe I should pass ```Chunk [] chs```  and go through 
```for (int col = 0; col < chs.length; ++col) {
            Chunk i_col = cs[col];
            for (int row = 0; row < i_col._len; ++row) { ... }}```
          

or is that only for Frames?

File: h2o-core/src/test/java/water/rapids/ast/prims/PermutationFeatureImportanceTest.java [hunk 28]
```diff
@@ -0,0 +1,444 @@
...
+            if( fr2 != null ) fr2.remove();
+            if( gbm != null ) gbm.remove();
+            Scope.exit();
+        }        
+        
+    }
+
+    //testing if model is getting the GBMmodel on my class (PermutationFeatureImportanceM.java)
+    @Test
+    public void Classification() throws Exception {
```
maurever, line 277:
```suggestion
    public void testClassification() throws Exception {
```

h2o-core/src/test/java/water/rapids/ast/prims/PermutationFeatureImportanceTest.java
```diff
@@ -0,0 +1,444 @@
...
+import org.junit.BeforeClass;
+
+import java.util.Arrays;
+
+import static hex.genmodel.utils.DistributionFamily.gaussian;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+
+public class PermutationFeatureImportanceTest extends TestUtil {
```
maurever, line 30: The tests here are usually not completed. Printing the feature importance at the end of the test is not correct test case. More work is needed here.

I miss a test where the feature importance from original model (for example, GLM model) is compared with permutation feature importance.

maurever (changes requested):
@ardKelmendi, good progress.

But there is still a lot of work to improve and finish.

h2o-core/src/main/java/water/rapids/ast/prims/models/AstPermutationFeatureImportanceM.java
```diff
@@ -0,0 +1,53 @@
...
+import water.rapids.Val;
+import water.rapids.ast.AstPrimitive;
+import water.rapids.ast.AstRoot;
+import water.rapids.vals.ValFrame;
+import water.util.TwoDimTable;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class AstPermutationFeatureImportanceM extends AstPrimitive {
```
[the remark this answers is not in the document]
ardKelmendi: I added it for when only passing the Model, I realise now that I can get the Frame from the model itself without needing to create a seperate Ast class

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -838,4 +837,49 @@ public void map(Chunk c, NewChunk nc) {
...
+  public static long seed(int cidx) {
+    return (0xe031e74f321f7e29L + ((long) cidx << 32L));
+  }
+
+
+  /**
+   * Randomly shuffle randomly a vector from a Frame
+   * */
+  
+  public static class ShuffleVecTask extends MRTask<ShuffleVecTask> {
```
maurever, line 849: Which ShuffleTask is used? There is still more than one implementation. It is confusing. Please keep only one implementation.

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -838,4 +837,49 @@ public void map(Chunk c, NewChunk nc) {
...
+    @Override public void map(Chunk ic, Chunk nc) {
+      Random rng = getRNG(seed(ic.cidx()));
+      for (int i = 1; i < ic._len ; i++) {
+        int j = rng.nextInt(i); // inclusive upper bound <0,i>
```
maurever, line 853: Why you generate the new index in rage <0, i>? Why dont you use the whole range <0, ic._len) ?

h2o-core/src/main/java/water/rapids/ShuffleVec.java
```diff
@@ -0,0 +1,83 @@
...
+import water.fvec.*;
+import water.parser.BufferedString;
+import water.util.FrameUtils;
+import water.util.RandomUtils;
+
+import java.util.Random;
+
+import static water.util.RandomUtils.getRNG;
+
+public class ShuffleVec {
```
maurever, line 14: Are you using this class somewhere? It is confusing; there is also another ShuffleTask... Please, keep only the final version.

h2o-core/src/main/java/water/rapids/ast/prims/models/AstPermutationFeatureImportanceM.java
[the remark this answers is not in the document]
maurever: So you can use the same Ast but the input frame can be null.

h2o-core/src/main/java/water/rapids/ast/prims/models/PermutationFeatureImportanceAstHelper.java
```diff
@@ -0,0 +1,33 @@
+package water.rapids.ast.prims.models;
+
+import water.fvec.Vec;
+import water.util.TwoDimTable;
+
+public class PermutationFeatureImportanceAstHelper {
```
maurever, line 6: Dont you Ast in the name if it is not Ast class, please. 

```suggestion
public class PermutationFeatureImportanceHelper {
```

h2o-core/src/main/java/water/rapids/ast/prims/models/PermutationFeatureImportanceAstHelper.java
```diff
@@ -0,0 +1,33 @@
...
+
+    Vec [] _vecs;
+    String [] _all_names;
+            
+    Vec[] do_Vecs(TwoDimTable varImp_t){
```
maurever, line 11:
```suggestion
    Vec[] doVecs(TwoDimTable varImpTable){
```

h2o-core/src/main/java/water/rapids/ast/prims/models/PermutationFeatureImportanceAstHelper.java
```diff
@@ -0,0 +1,33 @@
...
+            for (int j = 0 ; j < varImp_t.getColDim() ; j++){
+                tmp_row[j] = (double) varImp_t.get(i,j);
+            }
+            _vecs[i] = Vec.makeVec(tmp_row, Vec.newKey());
+        }
+        _vecs[varImp_t.getRowDim()] = Vec.makeVec(varImp_t.getColHeaders(), Vec.newKey());
+        return _vecs;
+    }
+
+    String [] do_Names(String [] table_names){
```
maurever, line 24:
```suggestion
    String [] doNames(String [] tableNames){
```

h2o-core/src/main/java/water/rapids/ast/prims/models/PermutationFeatureImportanceAstHelper.java
maurever: This is Java, use the correct naming style, please.

h2o-core/src/test/java/water/fvec/VecShuffleTest.java
```diff
@@ -0,0 +1,227 @@
...
+        } finally {
+            v.remove();
+        }
+        assert fr != null;
+        return fr;
+    }
+
+    // shuffle a dummy frame and check if all elements are there.
+    @Test
+    public void test_shuffle_vec(){
```
maurever, line 49: Java naming style, please.

h2o-core/src/test/java/water/fvec/VecShuffleTest.java
```diff
@@ -0,0 +1,227 @@
+package water.fvec;
+
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import water.*;
+import water.util.VecUtils;
+
+
+public class VecShuffleTest extends TestUtil{
```
maurever, line 10: A test for the string column is missing.

h2o-core/src/test/java/water/fvec/VecShuffleTest.java
```diff
@@ -0,0 +1,227 @@
...
+
+
+    @BeforeClass
+    public static void stall() {
+        stall_till_cloudsize(1);
+    } 
+
+
+    /** Create a frame with different Vec types to test MRTask for each type */
```
maurever, line 19: Please, use the correct docstring for every method here.

h2o-core/src/main/java/water/rapids/PermutationVarImp.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+    public String [] get_features_wo_res_test(){return _featuresWithNoResCol;}
+    
+    /**
+     * Creates a new array of Strings without the response column
+     * */
+    public void removeResCol(){
+        // remove response col from array of string to be added to TwoDimTable
+        _featuresWithNoResCol = _trainFr.names();
+        if (_featuresWithNoResCol[0].equals(_responseCol)) {
```
maurever, line 81: The response column is not always at index 0.

h2o-core/src/test/java/water/rapids/PermutationVarImpTest.java
```diff
@@ -0,0 +1,659 @@
...
+            System.out.println("COEF is  = " + rnd_coe);
+
+//            Assert.assertNotEquals(0, rnd_coe);
+        } finally {
+            Scope.exit();
+        }
+    }
+    
+    @Test
+    public void shuffleVecBiggerData(){
```
maurever, line 238: These test cases should be in ShuffleTaskTest not here.

h2o-core/src/test/java/water/rapids/ast/prims/PermutationFeatureImportanceTest.java
maurever: What is the difference between PermutationFeatureImportanceTest.java and PermutationVarImpTest.java file? Keep only one or explain their purpose, please.

h2o-py/h2o/model/PermutationFeatureImportance.py
```diff
@@ -0,0 +1,31 @@
...
+
+from h2o.frame import H2OFrame
+from h2o.expr import ExprNode
+
+
+def permutation_varimp(validation_frame, model, use_pandas=True):
+
+    m_frame = H2OFrame._expr(ExprNode("Perm_Feature_importance", validation_frame, model))
+
+    if type(m_frame) is not H2OFrame:
```
maurever, line 17: You should test input frame not your output.
```suggestion
    if type(validation_frame) is not H2OFrame:
```

h2o-py/h2o/model/PermutationFeatureImportance.py
```diff
@@ -0,0 +1,31 @@
...
+
+:copyright: (c) 2020 H2O.ai
+:license:   Apache License Version 2.0 (see LICENSE for details)
+"""
+
+from h2o.frame import H2OFrame
+from h2o.expr import ExprNode
+
+
+def permutation_varimp(validation_frame, model, use_pandas=True):
```
maurever, line 13: Why validation frame? I suggest pass model as first parameter. If the frame will be None, you can use training frame from the model. 

```suggestion
def permutation_varimp(model, frame=None, use_pandas=True):
```

h2o-py/h2o/model/PermutationFeatureImportance.py
```diff
@@ -0,0 +1,31 @@
+# -*- encoding: utf-8 -*-
+"""
+H2O Permutation Feature Importance.
```
maurever, line 3: Which file is used? PermutationFeatureImportance.py  or permutation_varimp.py? Keep only one, please.

h2o-py/h2o/model/permutation_varimp.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+
+from h2o.frame import H2OFrame
+from h2o.expr import ExprNode
+from h2o.utils.shared_utils import can_use_pandas
+
+import h2o
+
+
+def H2OFrame_to_pandas(m_frame):
```
maurever, line 16:
```suggestion
def varimp_to_pandas(m_frame):
```

h2o-r/h2o-package/R/permutation_varimp.R
```diff
@@ -0,0 +1,23 @@
+#'
+#' Calculate Permutation Feature Importance 
+#'
+
+h2o.permutation_varimp <- function(model, validation_frame){
```
maurever, line 5:
```suggestion
h2o.permutation_varimp <- function(model, frame){
```

maurever (changes requested):

h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
```diff
@@ -340,6 +340,7 @@ int prep(Frame fr) {
       // Done building model; produce a score column with predictions
       fr2 = gbm.score(fr);
 
+      new PermutationVarImp(gbm, fr).oat();
```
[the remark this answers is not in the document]
ardKelmendi: My bad, didn't inted to commit these files

h2o-core/src/main/java/water/rapids/PermutationVarImp.java
ardKelmendi: I left it for the cases which is. I have a different aproach now and am also removing the ignored variables from the params

h2o-core/src/test/java/water/rapids/ast/prims/PermutationFeatureImportanceTest.java
ardKelmendi: Will keep only PermutationVarImpTest.java

h2o-core/src/main/java/hex/Model.java
```diff
@@ -654,6 +658,19 @@ public Frame removeInteractionOnlyColumns(Frame f) {
 
   }
 
+  /**
+   * Should be called after Model is scored, Otherwise no one can tell you what will happen
```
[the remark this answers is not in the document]
ardKelmendi: Good point. Which h2o error would be best-suited for this scenario?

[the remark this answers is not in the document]
ardKelmendi: `throw new IllegalArgumentException("Model " + this._key + "must be scored! ");`
as of now

h2o-core/src/main/java/water/util/VecUtils.java
ardKelmendi: I removed the others. This is the one I use

h2o-core/src/main/java/water/util/VecUtils.java
ardKelmendi: Thats a great question. I was following the Fisher-Yates pseudocode 
for i from 1 to n do
      j ← random integer such that 0 ≤ j ≤ i

Using the entire range makes sense, I will check the difference

h2o-py/h2o/model/permutation_varimp.py
```diff
@@ -26,6 +26,12 @@ def permutation_varimp(model, frame, use_pandas=True, metric="mse"):
     if type(frame) is not H2OFrame:
         raise ValueError("Frame is not H2OFrame")
 
+    # Check if metric exists for the model
```
ardKelmendi, line 29: We talked about getting the available metrics from json. For GBM it works, however, for standalone_test(perm_var_imp_plot_glm) on pyunit_pubdev_7139_permutation_var_imp.py. The json format doesn't show all the metrics. Not even MSE nor logloss which this model's family is Binomial

maurever (changes requested):
Today I finished the review of PermutationVarImp.java class only. I will be continuing tomorrow.

h2o-core/src/main/java/water/rapids/MetricNotFoundExeption.java
```diff
@@ -0,0 +1,10 @@
+/**
+* Exception is thrown when the metric specified by the user could not be retrived.
+*/
+package water.rapids;
+
+public class MetricNotFoundExeption extends Exception {
```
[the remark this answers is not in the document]
ardKelmendi: replacing it with IllegalArgumentException

h2o-core/src/main/java/water/rapids/PermutationVarImp.java
```diff
@@ -0,0 +1,321 @@
...
+    public TwoDimTable PermutationVarImportance(){
+        removeResCol(); 
+        setOgMetric(); // get the metric value from the model
+        _varImpMap = new HashMap<>(_varsToShuffle.length);
+
+        int id = 0;
+        for (int f = 0; f < _trainFr.numCols(); f++)
+        {
+            // skip for response column
+            if (_var[f].equals(_responseCol) || isIgnored(_var[f]))
```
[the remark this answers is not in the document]
ardKelmendi: That's what I tried, however, ```_inputFrame.replace(f, shuffledFeature);``` takes an integer as a parameter, If I was to use the _varsToShuffe I would be losing the # of the column on the input frame.

h2o-core/src/test/java/water/rapids/PermutationVarImpTest.java
```diff
@@ -0,0 +1,372 @@
...
+    public void testPermVarImpGLM(){
+        Scope.enter();
+        Key parsed = Key.make("cars_parsed");
+        Frame fr = null;
+        GLMModel model = null;
+        try {
+            fr = parse_test_file(parsed, "smalldata/junit/cars.csv");
+            GLMModel.GLMParameters params = new GLMModel.GLMParameters(GLMModel.GLMParameters.Family.poisson, GLMModel.GLMParameters.Family.poisson.defaultLink, new double[]{0}, new double[]{0},0,0);
+            params._response_column = "power (hp)";
+            // params._response = fr.find(params._response_column);
```
maurever, line 287: Remove unused comments, please.

h2o-core/src/test/java/water/rapids/PermutationVarImpTest.java
```diff
@@ -0,0 +1,372 @@
...
+                             pvi[i].equals(coeff[i - 1]) ||
+                             pvi[i].equals(coeff[i - 2]) ||
+                             pvi[i].equals(coeff[i + 1]));
+
+                } else if (i == size - 1) {
+                     assertTrue(pvi[i].equals(coeff[i]) ||
+                             pvi[i].equals(coeff[i - 1]) ||
+                             pvi[i].equals(coeff[i - 2]));
+                } else{
+                    System.out.println("You messed it up!");
```
maurever, line 363: This print is not necessary here.

h2o-docs/src/product/data-science/permutationVarImp.rst
```diff
@@ -0,0 +1,70 @@
+Permutation Variable Importance
+-----------------------------------
+
+Introduction
+~~~~~~~~~~~~~~~~~
+
+Permutation feature importance (PFI) measures the increase in the prediction error of the model after we permuted the feature’s values, which breaks the relationship between the feature and the true outcome.
```
maurever, line 7:
```suggestion
Permutation feature (variable) importance (PFI) measures the increase in the prediction error of the model after we permuted the feature’s values, which breaks the relationship between the feature and the true outcome.
```

h2o-docs/src/product/data-science/permutationVarImp.rst
```diff
@@ -0,0 +1,70 @@
...
+    prostate_train["CAPSULE"] = prostate_train["CAPSULE"].asfactor()
+
+	# train model
+    gbm_h2o = H2OGradientBoostingEstimator(distribution="bernoulli")
+    gbm_h2o.train(x=list(range(1, prostate_train.ncol)), y="CAPSULE", training_frame=prostate_train)
+
+	#get PFI H2OFrame
+	pm_h2o_df = permutation_varimp(model, fr, use_pandas=True, metric=metric="auc)
+	
+	..code-tab:: r R
```
maurever, line 54:
```suggestion
	.. code-tab:: r R
```

R Tab should be the first to be consistent.

h2o-py/h2o/model/permutation_varimp.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+
+import h2o
+from h2o.frame import H2OFrame
+from h2o.expr import ExprNode
+from h2o.model.model_base import _get_matplotlib_pyplot
+from h2o.utils.shared_utils import can_use_pandas
+
+
+def permutation_varimp_oat(model, frame):
```
maurever, line 16: This method should be also documented with examples of usage in R and Python.

h2o-py/h2o/model/permutation_varimp.py
maurever: I will take a look.

maurever (changes requested):
Not finished yet. Will be continuing.

h2o-core/src/test/java/water/fvec/VecShuffleTest.java
```diff
@@ -0,0 +1,243 @@
...
+     */
+    @Test
+    public void shuffleBigVec(){
+        Key parsed = Key.make("shuffle_test_2_parsed");
+        Frame fr = null;
+        Vec shuffledFeature = null ;
+        try {
+            Scope.enter();
+           // contains 10,000,000 unique doubles
+            fr = parse_test_file(parsed, "smalldata/jira/shuffle_test_2.csv");
```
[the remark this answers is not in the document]
ardKelmendi: I see, I will use the bigdata file (loans) for this test and remove the files I have added to the smalldata folder.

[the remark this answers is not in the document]
ardKelmendi: I see, I will use the bigdata file (loans) for this test and remove the files I have added to the smalldata folder. Was using this one because every row has a unique value (to test the performance of the shuffling algorithm).

h2o-py/h2o/model/permutation_varimp.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    
+    if type(frame) is not H2OFrame:
+        raise ValueError("Frame is not H2OFrame")
+
+    # Check if metric exists for the model
+    existring_metrics = model._model_json['output']['scoring_history'].col_header
+    selected_metric = "training_" + metric.lower()
+    if selected_metric not in existring_metrics:
+        raise H2OValueError("Metric doesn't exist for this model.")
```
maurever, line 71: You have to import H2OValueError from the h2o.exception module...

h2o-py/h2o/model/permutation_varimp.py
```diff
@@ -0,0 +1,124 @@
...
+    influential (non-interacting), less influential, non-influential. This method plots the mean and standard
+    deviation of the Permutation Variable Importance.
+    :param model: trained model
+    :param frame: training frame
+    """
+
+    if type(frame) is not H2OFrame:
+        raise ValueError("Frame is not H2OFrame")
+
+    m_frame = H2OFrame._expr(ExprNode("PermutationVarImpOat", model, frame))
```
maurever, line 28: Not able to run this because your Ast classes are not registered in Env.java.

h2o-py/h2o/model/permutation_varimp.py
maurever, line 71:
```suggestion
        raise H2OValueError("Metric "+metric+" doesn't exist for this model.")
```

h2o-py/h2o/model/permutation_varimp.py
maurever: Try look into training metrics instead of into scoring history.  For example this command `model._model_json['output']['training_metrics']._metric_json` returns list of available metrics.

maurever: But be careful, the metrics there have concrete names for example MSE instead of mse.

h2o-r/h2o-package/R/permutation_varimp.R
```diff
@@ -0,0 +1,23 @@
+#'
+#' Calculate Permutation Feature Importance 
+#'
+
+h2o.permutation_varimp <- function(model, frame){
+    if (is.H2OFrame(validation_frame))
+         tryCatch(permutation_varim_table <- .newExpr('Perm_Feature_importance', model, validation_frame),
```
maurever, line 7: Cant work. Be sure, every test locally passed before saying this is done.

```suggestion
         tryCatch(permutation_varim_table <- .newExpr('PermutationVarImp', model, validation_frame), 
```

maurever (changes requested):
Last review part. I have a lot of problems building the project and run at least one test. Your Python and R code can't work. You should also rebase your branch against the master to make it easier for reviewers. Please. fix all my suggestions and be sure everything works locally!

h2o-core/src/main/java/water/rapids/PermutationVarImp.java
```diff
@@ -0,0 +1,310 @@
...
+        _inputFrame = fr;
+        _responseCol = _model._parms._response_column;
+        _variables = _inputFrame.names();
+        _metrics = new String[]{"r2", "mse", "rmse", "gini coefficient", "f1", "logloss", "auc"};
+    }
+    
+    /**
+     * Creates a new array of Strings without the response column and ignored columns
+     */
+    public void removeResCol(){
```
maurever, line 68: You should rename this function if you are removing not only the response column.

h2o-docs/src/product/data-science/permutationVarImp.rst
```diff
@@ -0,0 +1,70 @@
+Permutation Variable Importance
```
maurever, line 1: You should also write something about OAT and add an example of usage. Just because It is already implemented and it will be available in H2O by this PR.

h2o-py/h2o/model/permutation_varimp.py
```diff
@@ -0,0 +1,126 @@
...
+    :param use_pandas: select true to return pandas data frame
+    :param metric: (str) loss function metrics to be used 
+    :return: H2OFrame or Pandas data frame
+    """
+
+    if type(frame) is not H2OFrame:
+        raise H2OValueError("Frame is not H2OFrame")
+
+    # Check if metric exists for the model
+    # existring_metrics = model._model_json['output']['scoring_history'].col_header
```
maurever, line 71: Delete commented code, please.

h2o-py/tests/testdir_jira/pyunit_pubdev_7139_permutation_var_imp.py~HEAD
```diff
@@ -0,0 +1,119 @@
+import sys, os
```
maurever, line 1: This file is here maybe accidentally.

h2o-r/h2o-package/R/permutation_varimp.R
```diff
@@ -0,0 +1,18 @@
+#'
+#' Calculate Permutation Feature Importance 
+#'
+#' @param model     Trained model which's score is going to be used.
+#' @param frame     Training frame of the model which is going to be permuted
+#' @param metric    Loss Function metric (defalt is MSE)
+#' @return          Resulting frame with Relative, Scaled and Percentage scaled importances
```
maurever, line 7:
```suggestion
#' @return          Frame with Relative, Scaled and Percentage scaled importances
```

h2o-r/scripts/h2o-r-test-setup.R
```diff
@@ -183,10 +182,15 @@ function() {
     } else if (IS.RUNIT) {
         # source h2o-r/h2o-package/R. overrides h2o package load
         to_src <- c("aggregator.R", "classes.R", "connection.R","config.R", "constants.R", "logging.R", "communication.R",
-                    "kvstore.R", "frame.R", "targetencoder.R", "astfun.R","automl.R", "import.R", "parse.R", "export.R", "models.R", "edicts.R",
+                    "kvstore.R", "frame.R", "targetencoder.R", "targetencoder_deprecated.R", "astfun.R","automl.R", "import.R", "parse.R", "export.R", "models.R", "edicts.R",
```
maurever, line 185: Some unresolved conflict still here...

smalldata/jira/full_testset.csv
```diff
@@ -0,0 +1,44 @@
+full,meat (gr), price (eur), restaurant, city, gender
```
maurever, line 1: This file should not be used. We add files to our files differently, not by a commit.

maurever: Same for shuffle_test_2.csv file.

maurever (changes requested):
Just a quick review before I tried to run the code. There are some unresolved conflicts.

maurever:
The variable importance for the black-box model will be implemented in this new PR: #5495.

PR #4615, opened 2020-05-13
Pavel/gradle6.3 revert

Pscheidl:
## Los Revertos Grande

Gradle 6.3 reverted due to mailfunctioning build.

Win

Reverted one by one for better readability during PR state. Can be squashed of course (recommended).

The following screenshot might help a bit:

Screenshot from 2020-05-13 20-20-52


Pscheidl:
+ we need to revert today's commits. At the time this PR was built, I did not know the definitive state.

PR #4620, opened 2020-05-14
PUBDEV-7454: Make get_leaderboard work for re-attached H2OAutoML object

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7454

In both, python and R, get_leaderboard was failing on reattached AutoML object:

In python, it used `key` property that had the same value as `project_name` in the reattached object instead of `project_name@@response_column`.

In R, it used `id` attribute that wasn't set, so I set it in `h2o.get_automl`. 

For both I use value retrieved from `automl_job$automl_id$name`.

### DIFF
--- a/h2o-py/tests/testdir_algos/automl/pyunit_automl_get_automl.py
+++ b/h2o-py/tests/testdir_algos/automl/pyunit_automl_get_automl.py
@@ -3,7 +3,7 @@ import sys, os
 sys.path.insert(1, os.path.join("..","..",".."))
 import h2o
 from tests import pyunit_utils
-from h2o.automl import H2OAutoML
+from h2o.automl import H2OAutoML, get_leaderboard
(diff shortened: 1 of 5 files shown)


h2o-py/tests/testdir_algos/automl/pyunit_automl_get_automl.py
```diff
@@ -37,5 +37,7 @@ def test_get_automl():
     predictions_from_output = get_aml.predict(train)
     assert (predictions == predictions_from_output).all()
 
+    # Test get_leaderboard PUBDEV-7454
+    assert (get_leaderboard(aml) == get_leaderboard(get_aml)).all()
```
sebhrusen, line 41: can you please also double check that it works with the extra columns? It should, but better verify:
```get_leaderboard(get_aml, 'ALL')```

PR #4627, opened 2020-05-19
PUBDEV-7547: Create an additional StackedEnsemble only from monotonically constrained models

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7547

If user specified `monotone_constraints` in AutoML and during the run we create at least 2 monotone models, create a SE with just the monotone models.

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
@@ -12,6 +12,7 @@ import hex.glm.GLMModel;
 import water.DKV;
 import water.Job;
 import water.Key;
+import water.util.PojoUtils;
 
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -144,11 +145,11 @@ public class StackedEnsembleStepsProvider
                     boolean hasMonotoneConstrains(Key<Model> modelKey) {
                         Model model = DKV.getGet(modelKey);
                         try {
-                            KeyValue[] mc = (KeyValue[]) model._parms.getClass()
-                                    .getField("_monotone_constraints")
-                                    .get(model._parms);
+                            KeyValue[] mc = (KeyValue[]) PojoUtils.getFieldValue(
+                                    model._parms, "_monotone_constraints",
+                                    PojoUtils.FieldNaming.CONSISTENT);
                             return mc != null && mc.length > 0;
-                        } catch (NoSuchFieldException | IllegalAccessException e) {
+                        } catch (IllegalArgumentException e) {
                             return false;
                         }
                     }
@@ -190,10 +191,8 @@ public class StackedEnsembleStepsProvider
                     protected StackedEnsembleParameters getStackedEnsembleParameters(Key<Model>[] baseModels, boolean isLast) {
                         StackedEnsembleParameters params = super.getStackedEnsembleParameters(baseModels, isLast);
                         GLMModel.GLMParameters metalearnerParams = new GLMModel.GLMParameters();
-                        // Ensure that we have non negative constraint to make it easier to reason about the whole ensemble
-                        // Hard wired in case Metalearner.Algorithm.AUTO changes its implementation
-                        metalearnerParams._non_negative = true;
-                        params._metalearner_algorithm = Metalearner.Algorithm.glm;
+                        // AUTO has non negative constraint to make it easier to reason about the whole ensemble
+                        params._metalearner_algorithm = Metalearner.Algorithm.AUTO;
                         params._metalearner_parameters = metalearnerParams;
                         return params;
                     }
--- a/h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
@@ -188,16 +188,6 @@ public class StackedEnsembleStepsProvider
                     }
 
                     @Override
-                    protected StackedEnsembleParameters getStackedEnsembleParameters(Key<Model>[] baseModels, boolean isLast) {
-                        StackedEnsembleParameters params = super.getStackedEnsembleParameters(baseModels, isLast);
-                        GLMModel.GLMParameters metalearnerParams = new GLMModel.GLMParameters();
-                        // AUTO has non negative constraint to make it easier to reason about the whole ensemble
-                        params._metalearner_algorithm = Metalearner.Algorithm.AUTO;
-                        params._metalearner_parameters = metalearnerParams;
-                        return params;
-                    }
-
-                    @Override
                     protected Job<StackedEnsembleModel> startJob() {
                         return stack(_algo + "_MonotonicallyConstrainedModels", getBaseModels(), true);
                     }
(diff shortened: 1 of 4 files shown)


h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
```diff
@@ -130,6 +138,71 @@ protected boolean isStackedEnsemble(Key<Model> key) {
...
+                    boolean hasMonotoneConstrains(Key<Model> modelKey) {
+                        Model model = DKV.getGet(modelKey);
+                        try {
+                            KeyValue[] mc = (KeyValue[]) model._parms.getClass()
+                                    .getField("_monotone_constraints")
+                                    .get(model._parms);
+                            return mc != null && mc.length > 0;
+                        } catch (NoSuchFieldException | IllegalAccessException e) {
+                            return false;
+                        }
```
sebhrusen, line 153: please try to use `PojoUtils` whenever you can for reflection.
`PojoUtils.getFieldValue()` should do the job here.

h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
```diff
@@ -130,6 +138,71 @@ protected boolean isStackedEnsemble(Key<Model> key) {
...
+                                .filter(k -> !isStackedEnsemble(k) && hasMonotoneConstrains(k))
+                                .toArray(Key[]::new);
+                    }
+
+                    @Override
+                    protected StackedEnsembleParameters getStackedEnsembleParameters(Key<Model>[] baseModels, boolean isLast) {
+                        StackedEnsembleParameters params = super.getStackedEnsembleParameters(baseModels, isLast);
+                        GLMModel.GLMParameters metalearnerParams = new GLMModel.GLMParameters();
+                        // Ensure that we have non negative constraint to make it easier to reason about the whole ensemble
+                        // Hard wired in case Metalearner.Algorithm.AUTO changes its implementation
```
sebhrusen, line 194: I understand your concern, but I would still simply use AUTO for now to avoid multiplication of configurations.
We're maintaining both SE and AutoML, so I can't imagine us doing such important changes in SE impl. without checking those used in AutoML.
We should probably keep a comment saying the `non_negative` is a requirement for this particular SE.

h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
sebhrusen: or `ReflectionUtils`... whatever

tomasfryda: Thanks for the suggestion!

h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
```diff
@@ -130,6 +139,69 @@ protected boolean isStackedEnsemble(Key<Model> key) {
...
+                    }
+
+                    @Override
+                    protected StackedEnsembleParameters getStackedEnsembleParameters(Key<Model>[] baseModels, boolean isLast) {
+                        StackedEnsembleParameters params = super.getStackedEnsembleParameters(baseModels, isLast);
+                        GLMModel.GLMParameters metalearnerParams = new GLMModel.GLMParameters();
+                        // AUTO has non negative constraint to make it easier to reason about the whole ensemble
+                        params._metalearner_algorithm = Metalearner.Algorithm.AUTO;
+                        params._metalearner_parameters = metalearnerParams;
+                        return params;
```
sebhrusen, line 197: can still be improved.

in your `getStackedEnsembleParameters`, there's a call to `stackedEnsembleParameters.initMetalearnerParams();`.
This last method is overloaded so that you can directly pass the metalearner algo to initialize the parameters.
Note that with your current logic, the metalearner params are created twice: 
1. once in `super.getStackedEnsembleParameters` where some additional generic params about CV are also set.
2. once here, completely erasing the previous ones, including the k`keep_cv.__` params.
It's probably not your objective.

If you just want to enforce metalearner algo, (btw, it think it's doable only using the non-official `algo_parameters` functionality), you should be able to simply call sth like:
```
return super.getStackedEnsembleParameters(Metalearner.Algorithm.AUTO, baseModels, isLast);
```
Also note that if one day we need to enfore GLM, you can still do
```
StackedEnsembleParameters params = super.getStackedEnsembleParameters(Metalearner.Algorithm.GLM, baseModels, isLast);
((GLMParameters)(params._metalearner_parameters))._non_negative = true
```
still without erasing some of the params set in `getStackedEnsembleParameters`.

tomasfryda: Thanks @sebhrusen! 
Do I understand correctly, that I don't need to implement `getStackedEnsembleParameters` at all? 
It will work well unless we change the implementation of `AUTO` metalearner or if we allow to specify default metalearner in AutoML or if the user uses the non-official `algo_parameters` etc...

sebhrusen: I'm not fully sure we should remove the `getStackedEnsembleParameters`.
Intuitively, I would have kept it and just enforce the `AUTO` metalearner, but after checking, I realize that the `algo_parameters` are applied after this anyway (in `trainModel`) so, it would give actually the false impression that it's enforcing something...

I just want to keep code as simple as possible, without leaving any blind spot, so I just created those tickets:
https://0xdata.atlassian.net/browse/PUBDEV-7570  with a note so that it get fixed during next improvement of `algo_parameters`.
https://0xdata.atlassian.net/browse/PUBDEV-7571 so that we keep in mind the specificity of the monotonic SE.

tomasfryda:
Thanks Seb. Just so you know, I noticed a typo so I amended the last commit and I will wait for the tests and if everything's OK I will merge it.

h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
```diff
@@ -130,6 +139,59 @@ protected boolean isStackedEnsemble(Key<Model> key) {
...
+                    @SuppressWarnings("unchecked")
+                    protected Key<Model>[] getBaseModels() {
+                        return Stream.of(getTrainedModelsKeys())
+                                .filter(k -> !isStackedEnsemble(k) && hasMonotoneConstrains(k))
+                                .toArray(Key[]::new);
+                    }
+
+                    @Override
+                    protected Job<StackedEnsembleModel> startJob() {
+                        return stack(_algo + "_MonotonicallyConstrainedModels", getBaseModels(), true);
```
ledell, line 192: I know this has been merged but I didn't get a chance to review... @tomasfryda @sebhrusen Do you think it might be better to use a shorter name? e.g. `"_Monotonic"` or something?  Our Model ID names, especially SE, is already quite long and making the first column in the leaderboard very wide.

tomasfryda: I agree, "Monotonic" is better.

tomasfryda: I created a new PR since I failed to include it into this one: https://github.com/h2oai/h2o-3/pull/4656

PR #4628, opened 2020-05-20
PUBDEV-7554: R docs: Add back in "destination_frame" parameter

hannah-tillman:


angela0xdata:
@hannah-tillman I'm not sure why R tests are failing after this fix. 
@michalkurka are the failures unrelated?

h2o-r/h2o-package/R/frame.R
```diff
@@ -2170,6 +2170,7 @@ h2o.nlevels <- function(x) {
 #' that are referencing this column). If you want to make a copy of the column instead, use parameter in.place = FALSE.
 #'
 #' @param x A single categorical column.
+#' @param destination_frame A string indicating the destination key. If empty, this will be auto-generated by H2O.
```
angela0xdata, line 2173: ??
`destination_frame` is not a parameter in setLevels. It's the inconsistency use of `new_levels`/`new.levels` in the example that is causing the error.

angela0xdata:
💥 R3.5 CMD checks successful!

PR #4629, opened 2020-05-20
Convenience version of h2o.get_model that works with H2OFrame

michalkurka:
Useful with Segment Models and elsewhere where users have an H2OFrame
of model ids in a `model` column and filter the set to just a single row

### DIFF
--- a/h2o-py/h2o/h2o.py
+++ b/h2o-py/h2o/h2o.py
@@ -989,7 +989,13 @@ def get_model(model_id):
     ...             training_frame=airlines)
     >>> model2 = h2o.get_model(model.model_id)
     """
-    assert_is_type(model_id, str)
+    assert_is_type(model_id, str, H2OFrame)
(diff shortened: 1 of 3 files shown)


sebhrusen:
@michalkurka why not rather implement something like `get_models` if you want to support `H2OFrame` or possibly list as a parameter?
Given your use case for segment models, you would be able to retrieve them all at once, without the single row limitation.
To avoid accidental abuse, we could also have the following signature:
```
def get_models(model_ids, limit=10)
```
wdyt?

grigory93:
> Given your use case for segment models, you would be able to retrieve them all at once, without the single row limitation.

in case of segment models retrieving models without segment value(s) makes little sense. The method in class `h2o.model.segment_models.H2OSegmentModels.get_segment_model(**segment)` or something like this would address this problem, where `**segment` is a dictionary of segment column/value pairs.

PR #4630, opened 2020-05-20
[PUBDEV-7527] remove unsupported xgboost options

honzasterba:
min_data_in_leaf and min_sum_hessian_in_leaf were actually never supported by xgboost


h2o-r/h2o-package/R/xgboost.R
```diff
@@ -63,6 +63,7 @@
 #' @param colsample_bylevel (same as col_sample_rate) Column sample rate (from 0.0 to 1.0) Defaults to 1.
 #' @param col_sample_rate_per_tree (same as colsample_bytree) Column sample rate per tree (from 0.0 to 1.0) Defaults to 1.
 #' @param colsample_bytree (same as col_sample_rate_per_tree) Column sample rate per tree (from 0.0 to 1.0) Defaults to 1.
+#' @param colsample_bynode Column sample rate per tree node (from 0.0 to 1.0) Defaults to 1.
```
angela0xdata, line 66: Does this have an alias? It's not documented in the https://github.com//h2oai/h2o-3/blob/master/h2o-docs/src/product/data-science/xgboost.rst file.

honzasterba: good catch, will add it

honzasterba: fixed

PR #4631, opened 2020-05-20
Changed enum columns to string columns in frame compare to avoid different factor levels

wendycwong:
In runit_PUBDEV_7362_merge_duplicate_others.R, I merged two frames and then compare the merged frame with a result frame that contains the right answers.  However, h2o can be confused by the different factor levels in the merged frame and the result frame as shown here:

 Error: Test failed: 'PUBDEV-7362: check merge'
* level sets of factors are different

To avoid this problem, I first change the enum columns to string columns before doing the comparison.

### DIFF
--- a/h2o-r/tests/runitUtils/utilsR.R
+++ b/h2o-r/tests/runitUtils/utilsR.R
@@ -931,27 +931,33 @@ random_NN <- function(actFunc, max_layers, max_node_number) {
 # Parameters:  frame1, frame2: H2O frames to be compared.
 #              tolerance: tolerance of comparison
 #----------------------------------------------------------------------
-compareFrames <- function(frame1, frame2, prob=0.5, tolerance=1e-6) {
+compareFrames <- function(frame1, frame2, prob=0.5, tolerance=1e-6, enum2String=FALSE) {
   expect_true(nrow(frame1) == nrow(frame2) && ncol(frame1) == ncol(frame2), info="frame1 and frame2 are different in size.")
+  rframe1 <- as.data.frame(frame1)
+  rframe2 <- as.data.frame(frame2)
   for (colInd in c(1:ncol(frame1))) {
-
     notNumericCols = !(h2o.isnumeric(frame1[,colInd]) && h2o.isnumeric(frame2[,colInd]))
     if (notNumericCols) {
-      temp1 = frame1[,colInd]
-      temp2 = frame2[,colInd]
+      if (enum2String) {
+        temp1 <- as.character(rframe1[,colInd])
+        temp2 <- as.character(rframe2[,colInd])
+      } else {
+        temp1 <- as.factor(rframe1[,colInd])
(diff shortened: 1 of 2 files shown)


h2o-r/tests/runitUtils/utilsR.R
```diff
@@ -931,17 +931,22 @@ random_NN <- function(actFunc, max_layers, max_node_number) {
...
+        temp1 = as.data.frame(h2o.ascharacter(frame1[,colInd]))
+        temp2 = as.data.frame(h2o.ascharacter(frame2[,colInd]))       
+      } else {
+        temp1 = as.data.frame(frame1[,colInd])
+        temp2 = as.data.frame(frame2[,colInd])
+      }
     } else { 
-      temp1=as.numeric(frame1[,colInd])
-      temp2=as.numeric(frame2[,colInd])
+      temp1=as.data.frame(frame1[,colInd])
```
honzasterba, line 948: why do you need the as.data.frame everywhere?

wendycwong: I want to grab all data in one shot from H2O to R.  By using as.data.frame, R will grab one column of H2O Data.  If I just do frame1[, colInd] and then later on when I iterate over each temp1[ind], H2O will give R one element at a time and this can be slow especially if the dataset is large.

wendycwong: I change the code so that as.data.frame is called only once.

honzasterba: in python tests we have two different functions for that compare_frames and compare_frames_local

https://github.com/h2oai/h2o-3/blob/b9e241e19f64e876f57ecdfa9fd6cd6f865dfebd/h2o-py/tests/pyunit_utils/utilsPY.py#L3468

does it make sense to have the same dual approach available in R tests? I am not sure why you would ever use the non-local version since its so much slower, but perhaps there is one?

h2o-r/tests/runitUtils/utilsR.R
```diff
@@ -931,27 +931,33 @@ random_NN <- function(actFunc, max_layers, max_node_number) {
 # Parameters:  frame1, frame2: H2O frames to be compared.
 #              tolerance: tolerance of comparison
 #----------------------------------------------------------------------
-compareFrames <- function(frame1, frame2, prob=0.5, tolerance=1e-6) {
+compareFrames <- function(frame1, frame2, prob=0.5, tolerance=1e-6, enum2String=FALSE) {
```
michalkurka, line 934: `enum2Strings` - is the purpose of the flag to ignore the order of the factors? can we find a better fitting name if that is the case? eg. ignore_factor_order?

h2o-r/tests/runitUtils/utilsR.R
wendycwong: Nope, the nonlocal version was written first.  Then I added the local version.

PR #4633, opened 2020-05-21
[nps] refactor duplicate code

honzasterba:


h2o-core/src/main/java/water/server/ServletUtils.java
```diff
@@ -173,6 +175,21 @@ public static String getDecodedUri(HttpServletRequest request) {
       throw new RuntimeException(e);
     }
   }
+  
+  public static String[] parseUriParams(String uri, HttpServletResponse response, Pattern p, int numParams) throws IOException {
+    Matcher m = p.matcher(uri);
+    if (!m.matches()) {
```
Pscheidl, line 181: I'd suggesting verifying the number of matches stored inside the matcher equals the expected number of params. Or derive the number of expected params from the number of matches if that does not need to be checked.

h2o-core/src/main/java/water/api/NpsBinServlet.java
```diff
@@ -18,22 +18,20 @@
 /**
  */
 public class NpsBinServlet extends HttpServlet {
+  
+  private static final Pattern URL_PATTERN = Pattern.compile(".*/NodePersistentStorage.bin/([^/]+)/([^/]+)");
```
Pscheidl, line 22: Good approach, the pattern is expensive to create.

h2o-core/src/main/java/water/server/ServletUtils.java
```diff
@@ -173,6 +175,21 @@ public static String getDecodedUri(HttpServletRequest request) {
       throw new RuntimeException(e);
     }
   }
+  
+  public static String[] parseUriParams(String uri, HttpServletResponse response, Pattern p, int numParams) throws IOException {
```
Pscheidl, line 179: Minor: There is only one pattern really right now, the method is static anyway and can internally refer to the `Pattern p` instead of passing it as an additional parameter. A matter of preference really.

h2o-core/src/main/java/water/api/NpsBinServlet.java
```diff
@@ -18,22 +18,20 @@
...
     String uri = ServletUtils.getDecodedUri(request);
     try {
-      Pattern p = Pattern.compile(".*/NodePersistentStorage.bin/([^/]+)/([^/]+)");
-      Matcher m = p.matcher(uri);
-      boolean b = m.matches();
-      if (!b) {
-        ServletUtils.setResponseStatus(response, HttpServletResponse.SC_BAD_REQUEST);
-        response.getWriter().write("Improperly formatted URI");
+      String[] params = ServletUtils.parseUriParams(uri, response, URL_PATTERN, 2);
+      if (params == null) {
```
Pscheidl, line 29: Does using `Optional` sound good ? It would force the null/empty check.

PR #4634, opened 2020-05-21
PUBDEV-7549: Suppress data.table warning messages inside h2o.automl() in R

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7549

Tested locally, but I am not sure if it's worth it to make an automated test for this change since I would probably need to install/uninstall packages during the test and the test itself would be more complex than the actual change.

### DIFF
--- a/h2o-r/h2o-package/R/automl.R
+++ b/h2o-r/h2o-package/R/automl.R
@@ -412,7 +412,7 @@ h2o.predict.H2OAutoML <- function(object, newdata, ...) {
   is_progress <- isTRUE(as.logical(.h2o.is_progress()))
   if (show_progress) h2o.show_progress() else h2o.no_progress()
   frame <- tryCatch(
-    as.h2o(table, destination_frame=destination_frame),
+    suppressWarnings(as.h2o(table, destination_frame=destination_frame)),
     error = identity,
     finally = if (is_progress) h2o.show_progress() else h2o.no_progress()
   )
--- a/h2o-r/h2o-package/R/automl.R
+++ b/h2o-r/h2o-package/R/automl.R
@@ -412,7 +412,7 @@ h2o.predict.H2OAutoML <- function(object, newdata, ...) {
   is_progress <- isTRUE(as.logical(.h2o.is_progress()))
   if (show_progress) h2o.show_progress() else h2o.no_progress()
   frame <- tryCatch(
-    suppressWarnings(as.h2o(table, destination_frame=destination_frame)),
+    as.h2o(table, destination_frame=destination_frame, use_datatable=FALSE),
     error = identity,
     finally = if (is_progress) h2o.show_progress() else h2o.no_progress()
   )
--- a/h2o-r/h2o-package/R/frame.R
+++ b/h2o-r/h2o-package/R/frame.R
@@ -4034,8 +4034,8 @@ use.package <- function(package,
             is.character(version), length(version)==1L,
             is.logical(use), length(use)==1L)
 
-   if (package == "data.table" && use && length(find.package("data.table", quiet = TRUE)) > 0) { # not sure if this is needed.  Keeping it for now.
-     if ((length(find.package("bit64", quiet = TRUE)) == 0) || (packageVersion("bit64") < as.package_version("0.9.7"))) {
+   if (package == "data.table" && use && suppressWarnings(require("data.table", quietly = TRUE))) { # not sure if this is needed.  Keeping it for now.
+     if ((!suppressWarnings(require("bit64", quietly = TRUE))) || (packageVersion("bit64") < as.package_version("0.9.7"))) {
         # print out warning to install bit64 in order to use data.table
        warning("data.table cannot be used without R package bit64 version 0.9.7 or higher.  Please upgrade to take advangage of data.table speedups.")
        return(FALSE)
--- a/h2o-r/h2o-package/R/frame.R
+++ b/h2o-r/h2o-package/R/frame.R
@@ -4034,8 +4034,8 @@ use.package <- function(package,
             is.character(version), length(version)==1L,
             is.logical(use), length(use)==1L)
 
-   if (package=="data.table" && use) { # not sure if this is needed.  Keeping it for now.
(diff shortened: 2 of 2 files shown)


h2o-r/h2o-package/R/frame.R
```diff
@@ -4034,7 +4034,7 @@ use.package <- function(package,
             is.character(version), length(version)==1L,
             is.logical(use), length(use)==1L)
 
-   if (package=="data.table" && use) { # not sure if this is needed.  Keeping it for now.
+   if (package=="data.table" && use && "data.table" %in% rownames(installed.packages())) { # not sure if this is needed.  Keeping it for now.
```
sebhrusen, line 4037: looks like `find.package` is much more efficient that listing `installed.packages`:
```
use && length(find.package("data.table", quiet=TRUE)) > 0
```

sebhrusen: can probably improvedfor the "but64" test below as well

sebhrusen: btw, should we not at least emit a warning if user is explicitely calling `use.package("data.table")` but the package is not installed? or is it only used internally?

sebhrusen: oh I see, it just returns a boolean

tomasfryda: Thanks for the tip!

sebhrusen:
sorry for that:
```
find.package is not usually the right tool to find out if a package is available for use: the only way to do that is to use require to try to load it. It need not be installed for the correct platform, it might have a version requirement not met by the running version of R, there might be dependencies which are not available, ....
```

tomasfryda:
Yeah, I was thinking about using the `require` but it also attaches the package, so I am wondering if that is what we should use since we might not want to attach the package, e.g., when we test for `bit64` and its version...

sebhrusen:
- for `data.tables` package will be attached anyway if `use` is TRUE and package is available,  right?
- for `bit64`, wouldn't `data.table` attach it anyway if it's here?

I'm no R expert, but `require` looks safer in any case. @ledell ?

tomasfryda:
@sebhrusen  I think you are right. As far as I can tell it wouldn't cause any problems for the current usage. I just don't like predicates that change the outer scope in general. 

But unfortunately, it seems there is no better way than using `require` so I will go with that. (Especially because we are querying for packages that aren't written only in R and thus there can be issues like package is compiled for other R version etc.)

michalkurka:
@wendycwong can you please take a look?

h2o-r/h2o-package/R/automl.R
```diff
@@ -412,7 +412,7 @@ h2o.predict.H2OAutoML <- function(object, newdata, ...) {
   is_progress <- isTRUE(as.logical(.h2o.is_progress()))
   if (show_progress) h2o.show_progress() else h2o.no_progress()
   frame <- tryCatch(
-    as.h2o(table, destination_frame=destination_frame),
+    suppressWarnings(as.h2o(table, destination_frame=destination_frame)),
```
michalkurka, line 415: Are the warning suppressed because of data.table?

IMHO we should address it better:
- in this context the tables will be slow, it doesn't even make sense to attempt to use data.table
- as.h2o could have an option that would disable d.t and or/ d.t would be triggered only for frames of certain sizes

honzasterba: this suppresses the warning of data.table being installed without bit64, which means we do not use it

michalkurka: yes, that is what I thought - my point is we should give users a mechanism to disable use of data.table (in as.h2o)

we know the data is small enough and DT won't provide any benefit

another point of view is that we don't even need to attempt to use DT when the data is this small

honzasterba:
@sebhrusen I think we do not watn to globally expose data.table, I think requireNamespace is the prefered way of accessing dependencies in packages

sebhrusen:
@honzasterba yes, looks exactly what we need, thx!

h2o-r/h2o-package/R/automl.R
tomasfryda: Yes, it's because of this message:
```
warning("data.table cannot be used without R package bit64 version 0.9.7 or higher.  Please upgrade to take advangage of data.table speedups.")
```

honzasterba: we do

```use=getOption("h2o.use.data.table", TRUE)[package=="data.table"]```

michalkurka: > in as.h2o - not just using a global flag

not in as.h2o

michalkurka: The point I am trying to make is that we are suppressing a warning from something we don't need at all in this case.

And possibly other (unrelated) warnings. Instead of that we could just disable data.table in a specific call to `as.h2o`. Does that make sense?

honzasterba: data.table is used in as.h2o and as.data.frame.H2OFrame only, are you suggesting separate property for each? I dont see any benefit in that, also I dont think we want to disable the use of data.table in as.h2o for automl, we want to use it if available

michalkurka: Not a property, an option passed to `as.h2o` function. I think that would be beneficial to be able to bypass data.table on case-by-case basis.

>  I dont think we want to disable the use of data.table in as.h2o for automl, we want to use it if available

DT doesn't add any value in this particular case. The only reason why DT was added to H2O as an option is that it is faster on larger data. On small data (that we handle in this method) it doesn't matter.

tomasfryda: I added `use_datatable` option to `as.h2o` and removed the `suppressWarnings` from the automl's call so we don't suppress other potentially useful warnings.

tomasfryda: And I manually tested it again with these 3 scenarios:

- no data.table installed
- data.table but no bit64 installed
- data.table and bit64 installed

PR #4640, opened 2020-05-22
[PUBDEV-746] add suport for external cluster with basic auth

honzasterba:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/task/XGBoostUploadMatrixTask.java
```diff
@@ -15,17 +15,23 @@
...
         this.remoteNodes = remoteNodes;
         this.https = https;
+        this.userName = userName;
+        this.password = password;
     }
 
     @Override
     protected void execute() {
-        XGBoostHttpClient client = new XGBoostHttpClient(remoteNodes[H2O.SELF.index()], https);
+        XGBoostHttpClient client = new XGBoostHttpClient(remoteNodes[H2O.SELF.index()], https, userName, password);
```
Pscheidl, line 34: New XGBoostHttpClient is created every single time. By looking at it's constructor, it looks kind-of expensive to construct.

- Pool it (with destruction timeout ? Ideal scenario would be trigger deletion after model training completion) - preferred option.
- Make it cheaper to instantiate by lazily instantiating it once.

honzasterba: yes, but this is done only once per node, only on leader node this is done twice per training

Pscheidl: That's good then.

PR #4646, opened 2020-05-26
Fix test-r-cmd-check-as-cran

tomasfryda:
Failed due to adding `use_datatable` to `as.data.frame.H2OFrame`, which was unnecessary in https://github.com/h2oai/h2o-3/pull/4634.

So I removed it from `as.data.frame.H2OFrame` but kept it in `as.h2o` where it was required. The problem is with having a conflict of keyword arguments among the methods that implement `as.data.frame` generic function. 

To get around using an explicit argument I could do something like (getting the information from `...`):
```
use_datatable <- as.list(match.call())$use_datatable
use_datatable <- is.null(use_datatable) || use_datatable
```

But I am not sure if we need it at all there.

Second part of the problem was missing documentation string for the `use_datatable` argument.

### DIFF
--- a/h2o-r/h2o-package/R/frame.R
+++ b/h2o-r/h2o-package/R/frame.R
@@ -4053,6 +4053,7 @@ use.package <- function(package,
 #'
 #' @param x An \code{R} object.
 #' @param destination_frame A string with the desired name for the H2OFrame
+#' @param use_datatable allow usage of data.table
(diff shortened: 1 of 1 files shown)


wendycwong:
Regardless of your questions, what you do actually fixed the r_cmd_check_as_cran problem.  I will fix my merge bug.  I think I know what is wrong with it.

PR #4648, opened 2020-05-26
PUBDEV-7565: Enable the use of cacert when used from config

michalkurka:


h2o-py/tests/testdir_munging/pyunit_cacert_conf.py
```diff
@@ -0,0 +1,33 @@
...
+        "https": True
+    }
+    try:
+        h2o.connect(config=cfg)
+        assert False
+    except H2OConnectionError as e:
+        assert "CERTIFICATE_VERIFY_FAILED" in e.message
+
+    # not ideal, but at least we test the parameter does get used
+    cfg["cacert"] = "/invalid/cacert/path"
```
michalkurka, line 22: @honzasterba if you have a better idea of testing this, please suggest - I would ideally want to provide a valid bundle: https://h2oai.slack.com/archives/C1Q2M34JG/p1590523056395200

h2o-py/tests/testdir_munging/pyunit_cacert_conf.py
```diff
@@ -0,0 +1,33 @@
+import sys
```
honzasterba, line 1: this test is in the wrong directory

h2o-py/tests/testdir_munging/pyunit_cacert_conf.py
honzasterba: I think (mostly just guessing) this about how the original trust-store is generated, I think we need first generate the CAcert and then use that to generate the .jks, then we can use that here

h2o-py/tests/testdir_munging/pyunit_cacert_conf.py
michalkurka: oh. munging instead of miscellaneous :)

h2o-py/tests/testdir_munging/pyunit_cacert_conf.py
michalkurka: yeah, it could be done that way, eg.:

```
            keytool -genkey -dname "cn=localhost, ou=H2O-3, o=H2O.ai, c=US" -alias h2o -keystore mykeystore.jks -storepass h2oh2o -keypass h2oh2o -keyalg RSA -keysize 2048
            keytool -exportcert -alias h2o -keystore mykeystore.jks -rfc -file serverpub.pem -storepass h2oh2o -keypass h2oh2o
```

I don't think I want to change the testing infra for this and I don't see a way how to do it in a self-contained test. I think I will keep the test as is, it should be a sufficient check. WDYT?

honzasterba: I am a little un-easy about not having any tests for the happy path. I think it's not that much work to get it done. we could use the hadoop tests to test the happy path, there we generate the keys in the test script

https://github.com/h2oai/h2o-3/blob/d5e503b754c1ff7d43e027b229b741fe7aab2287/scripts/jenkins/groovy/hadoopStage.groovy#L28

michalkurka: yep, that is the place where I would add the 2 lines posted in the previous comment

you just gave me an idea of how to do it though

I can still start the cluster the old way with HTTPS but keep the connection insecure (no cert verification) and keep this test to use h2o.connect on the existing cluster. that is imho doable

PR #4649, opened 2020-05-26
[build] allow hadoop target distros to be set from gradle.properties

honzasterba:
also cleanup some weird gradle formating

### DIFF
--- a/gradle/components/xgboost.gradle
+++ b/gradle/components/xgboost.gradle
@@ -1,4 +1,3 @@
 dependencies {
-    compile project(":h2o-genmodel-ext-xgboost")
     compile project(":h2o-ext-xgboost")
 }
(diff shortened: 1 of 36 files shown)


gradle/components/xgboost.gradle
```diff
@@ -1,4 +1,3 @@
 dependencies {
-    compile project(":h2o-genmodel-ext-xgboost")
```
honzasterba, line 2: unecessary, added as transitive

Pscheidl:
Build error: `Could not get unknown property 'hadoopTargets' for settings 'h2o-3' of type org.gradle.initialization.DefaultSettings.`

http://mr-0xc1:8080/blue/organizations/jenkins/h2o-3-pipeline/detail/PR-4649/2/pipeline

PR #4651, opened 2020-05-27
PUBDEV-7567: Enable using imported models as base models in Stacked Ensembles

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7567

This is a part of code that I "extracted" from the PR containing quantile distribution support.

This will be potentially useful for saving and loading AutoML. (I experimented with that some time ago, I ended up with some R code[1] to do that) 

[1] https://gist.github.com/tomasfryda/b46a0e910e92da9f051f60a0c0140ecc

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
@@ -487,10 +487,12 @@ public class StackedEnsembleModel extends Model<StackedEnsembleModel,StackedEnse
 
         if (require_consistent_training_frames) {
           if (trainingFrameRows < 0) trainingFrameRows = _parms.train().numRows();
-          Frame aTrainingFrame = aModel._parms.train(); // FIXME: can be null when aModel was imported
-          if (trainingFrameRows != aTrainingFrame.numRows())
+          long numOfRowsUsedToTrain = aModel._parms.train() == null ?
+                  aModel._output._cross_validation_holdout_predictions_frame_id.get().numRows() :
+                  aModel._parms.train().numRows();
+          if (trainingFrameRows != numOfRowsUsedToTrain)
             throw new H2OIllegalArgumentException("Base models are inconsistent: they use different size (number of rows) training frames."
-                    +" Found: "+trainingFrameRows+" (StackedEnsemble) and "+aTrainingFrame.numRows()+" (model "+k+").");
+                    +" Found: "+trainingFrameRows+" (StackedEnsemble) and "+numOfRowsUsedToTrain+" (model "+k+").");
         }
 
         if (cv_required_on_base_model) {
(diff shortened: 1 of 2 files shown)


h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -487,7 +487,9 @@ void checkAndInheritModelProperties() {
 
         if (require_consistent_training_frames) {
           if (trainingFrameRows < 0) trainingFrameRows = _parms.train().numRows();
-          Frame aTrainingFrame = aModel._parms.train(); // FIXME: can be null when aModel was imported
+          Frame aTrainingFrame = aModel._parms.train();
+          if (aTrainingFrame == null)
+            aTrainingFrame = aModel._output._cross_validation_holdout_predictions_frame_id.get();
```
sebhrusen, line 492: to avoid confusion, and as we're just concerned about how many rows were used to train the base models, can you please try to manipulate only rowNums instead of frames? 
Just to avoid this:
```
aTrainingFrame = aModel._output._cross_validation_holdout_predictions_frame_id.get();
```
as this is a purely ad-hoc statement that is not actually true.

tomasfryda: Good point! When I was writing it I was thinking about holdout_predictions as a part of training frame for metalearner so it didn't struck as odd.  😅

h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_base_models.py
```diff
@@ -195,9 +195,69 @@ def test():
...
+        drf_holdout_id = drf.cross_validation_holdout_predictions().frame_id
+        h2o.export_file(gbm.cross_validation_holdout_predictions(), TMP_DIR + "/gbm.holdout")
+        h2o.export_file(drf.cross_validation_holdout_predictions(), TMP_DIR + "/drf.holdout")
+
+        h2o.remove_all()
+
+        h2o.import_file(TMP_DIR + "/gbm.holdout", gbm_holdout_id)
+        h2o.import_file(TMP_DIR + "/drf.holdout", drf_holdout_id)
+
+        gbm = h2o.upload_model(glob.glob(TMP_DIR + "/gbm.model/*")[0])
```
ledell, line 238: @tomasfryda Is there a reason you're using `h2o.upload_model()` instead of `h2o.load_model()`?  I think because the client/H2O cluster machine are the same in Jenkins, either works.  But historically, `h2o.load_model()` is the default for some reason.

tomasfryda: No reason at all. I think `h2o.upload_model` felt safer to me. I just tested `h2o.load_model` it works locally so I can modify it, if you think I should. I guess the default is a bit more efficient communication-wise but I doubt there would be any noticeable effect on models with this size...

PR #4655, opened 2020-05-28
Ignore NBHM demo/test (doesn't work reliably on Jenkins)

michalkurka:


h2o-core/src/test/java/water/util/NonBlockingHashMapTest.java
```diff
@@ -38,11 +38,13 @@ public void testPutIfAbsent_PubDev6319() throws Exception {
...
     }
 
-    assertNotEquals(0, map.size());       // sanity check - something was inserted
-    assertEquals(map.size(), totalInserts);          // size of the map should be the same as the aggregated number of inserts
+    assertNotEquals(0, map.size());         // sanity check - something was inserted
+    assertEquals(map.size(), totalInserts); // size of the map should be the same as the aggregated number of inserts
     assertEquals(0, totalForeignValues);    // values should be inserted only once and shouldn't change over time
-    assertNotEquals(0, totalForeignKeys); // we have no guarantee whose key will be used - it doesn't have to
-                                                     // be the thread that actually made the insert - this causes PUBDEV-6319
+    assertTrue(totalForeignKeys >= 0);      // we have no guarantee whose key will be used - it doesn't have to
```
honzasterba, line 44: in this case we can just remove this assert and totalForeignKeys with it

just an idea - what about using something more deterministic, like generating an array of random "work" for each putter instead of using Nano time, that way we could have exact expectations for these results

michalkurka: I would say the time is better for your idea to work you need some kind of synchronization between the threads, the idea of using time is that each thread does take the same value at the same time already 

it works very well locally with milliseconds and nanoseconds - on Jenkins the thread scheduling seems messed up to me (maybe the machines are overloaded)

honzasterba: no we would not, we would not care about the order only about the number of conficts and that does not depend on order of operations

michalkurka: how come? if you pre-generate an array of stuff to insert then there is no guarantee there will be any conflict

michalkurka: (I agree the test is crippled this way - not a good idea to do it that way, maybe I can ignore it on Jenkins instead since it is not a true test - it is a demonstration of behavior)

michalkurka: Ignored instead :)

PR #4660, opened 2020-05-28
PUVDEV-6424 add multinomial makeGLMModel

wendycwong:
This PR completes the task in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-6424?filter=-1

I have added the capability to make a GLM multinomial model to makeGLMModel and a test to go with it.

### DIFF
--- a/h2o-algos/src/main/java/hex/api/MakeGLMModelHandler.java
+++ b/h2o-algos/src/main/java/hex/api/MakeGLMModelHandler.java
@@ -26,27 +26,31 @@ public class MakeGLMModelHandler extends Handler {
     GLMModel model = DKV.getGet(args.model.key());
     if(model == null)
       throw new IllegalArgumentException("missing source model " + args.model);
-    String [] names = model._output.coefficientNames();
+    boolean multiClass = model._output._multinomial || model._output._ordinal;
+    String [] names = multiClass?model._output.multiClassCoeffNames():model._output.coefficientNames(); // coefficient names in order and with Intercept
     Map<String,Double> coefs = model.coefficients();
     if (args.beta.length != names.length) {
       throw new IllegalArgumentException("model coefficient length " + names.length + " is different from coefficient" +
-              " provided by user " + args.beta.length+".\n model coefficients needed are:\n"+String.join("\n", names));
+              " provided by user " + args.beta.length + ".\n model coefficients needed are:\n" + String.join("\n", names));
     }
     for(int i = 0; i < args.names.length; ++i)
       coefs.put(args.names[i],args.beta[i]);
     double [] beta = model.beta().clone();
     for(int i = 0; i < beta.length; ++i)
       beta[i] = coefs.get(names[i]);
-    GLMModel m = new GLMModel(args.dest != null?args.dest.key():Key.make(),model._parms,null, model._ymu, Double.NaN, Double.NaN, -1);
+    GLMModel m = new GLMModel(args.dest != null?args.dest.key():Key.make(),model._parms,null, model._ymu,
+            Double.NaN, Double.NaN, -1);
     DataInfo dinfo = model.dinfo();
     dinfo.setPredictorTransform(TransformType.NONE);
-    m._output = new GLMOutput(model.dinfo(),model._output._names,model._output._column_types,  model._output._domains, model._output.coefficientNames(), model._output._binomial, beta);
+    m._output = new GLMOutput(model.dinfo(), model._output._names, model._output._column_types, model._output._domains,
+            model._output.coefficientNames(), beta, model._output._binomial, model._output._multinomial, 
+            model._output._ordinal);
     DKV.put(m._key, m);
     GLMModelV3 res = new GLMModelV3();
     res.fillFromImpl(m);
     return res;
   }
-
+  
   public GLMRegularizationPathV3 extractRegularizationPath(int v, GLMRegularizationPathV3 args) {
     GLMModel model = DKV.getGet(args.model.key());
     if(model == null)
@@ -56,7 +60,6 @@ public class MakeGLMModelHandler extends Handler {
   // instead of adding a new endpoint, just put this stupid test functionality here
  /** Get the expanded (interactions + offsets) dataset. Careful printing! Test only
   */
-
   public DataInfoFrameV3 getDataInfoFrame(int version, DataInfoFrameV3 args) {
     Frame fr = DKV.getGet(args.frame.key());
     if( null==fr ) throw new IllegalArgumentException("no frame found");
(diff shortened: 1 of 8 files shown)


h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_6424_multinomial_makeGLMModel.py
```diff
@@ -0,0 +1,31 @@
...
+    f1 = mL.predict(d)
+    f2 = m2.predict(d)
+    pyunit_utils.compare_frames_local(f1, f2, prob=1)
+    
+    coefs = r['coefficients'][0]
+    coefs["wendy_dreams"] = 0.123 # add extra coefficients to model coefficient
+    
+    try:
+        m3 = glm.makeGLMModel(model=mL, coefs=coefs)
+        sys.exit("Should have thrown an exception!")
```
honzasterba, line 23: use assert False

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_6424_multinomial_makeGLMModel.py
```diff
@@ -0,0 +1,31 @@
...
+    except Exception as ex:
+        print(ex)
```
honzasterba, line 25: check that the exception message is as expected

h2o-algos/src/main/java/hex/api/MakeGLMModelHandler.java
```diff
@@ -26,23 +27,69 @@ public GLMModelV3 make_model(int version, MakeGLMModelV3 args){
...
+      
+    }
+      return newCoefs;
+  }
+  
+  // calling model._output.coefficientNames() will only return you names like C1, C2, ..., Intercept.  However
+  // for multinomials, there are N classes of coefficients which are writtenb as C1_1, C2_1, to denote the
+  // coefficients for each class.  This method is to take the coefficient names of one class and extend it to
+  // coefficient names for all N classes.
+  public String[] multinomialCoefNames(String[] names, GLMModel model) {
```
maurever, line 79: Why it returns bad values from the output? Would it be better to make these changes in the model output instead of here? The same also for the `modifyCoeffs` method.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1166,6 +1166,20 @@ public GLMOutput(DataInfo dinfo, String[] column_names, String[] column_types, S
...
+    public GLMOutput(DataInfo dinfo, String[] column_names, String[] column_types, String[][] domains, 
+                     String[] coefficient_names, double[] beta) {
+      super(dinfo._weights, dinfo._offset, dinfo._fold);
+      _dinfo = dinfo.clone();
+      setNames(column_names, column_types);
+      _domains = domains;
+      _coefficient_names = coefficient_names;
+      _multinomial = true;
+      _nclasses = beta.length/coefficient_names.length;
+      assert !ArrayUtils.hasNaNsOrInfs(beta);
```
maurever, line 1178: Please, add an assert message.

h2o-core/src/main/java/water/util/ArrayUtils.java
```diff
@@ -1485,10 +1485,10 @@ public int compare(Integer x, Integer y) {
   }
 
   public static double [][] convertTo2DMatrix(double [] x, int N) {
-    assert x.length % N == 0;
+    assert x.length % N == 0; // N is number of coefficients per class
```
maurever, line 1488: Again assert message please. We will immediately know what is wrong if the code starts failing.

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: Done.

h2o-algos/src/main/java/hex/api/MakeGLMModelHandler.java
wendycwong: Hi @maurever:

Yes, I move the multinomialCoefNames to GLMModel and renamed it multiClassCoeffNames().  I rewrite the model.coefficients() so that for multi-class, it will provide coefficients as coefficient_names+"_"+target domain.

maurever: Cool, thank you @wendycwong!

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1161,14 +1178,31 @@ public GLMOutput(DataInfo dinfo, String[] column_names, String[] column_types, S
 
     public GLMOutput(DataInfo dinfo, String[] column_names, String[] column_types, String[][] domains, String[] coefficient_names, boolean binomial, double[] beta) {
       this(dinfo,column_names,column_types, domains,coefficient_names,binomial);
-      assert !ArrayUtils.hasNaNsOrInfs(beta);
+      assert !ArrayUtils.hasNaNsOrInfs(beta): "Coefficients contain NA or Infs.";
       _global_beta=beta;
       _submodels = new Submodel[]{new Submodel(0,beta,-1,Double.NaN,Double.NaN)};
     }
 
+    public GLMOutput(DataInfo dinfo, String[] column_names, String[] column_types, String[][] domains,
```
honzasterba, line 1186: this is poor API design having one constructor like `... double[] beta, boolean isMultinomial` and another `boolean binomial, double[] beta` is very confusing, could we merge those two, maybe have two boolean params or better an Enum since both cannot be true at once I would think?

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1161,14 +1178,31 @@ public GLMOutput(DataInfo dinfo, String[] column_names, String[] column_types, S
...
       this(dinfo,column_names,column_types, domains,coefficient_names,binomial);
-      assert !ArrayUtils.hasNaNsOrInfs(beta);
+      assert !ArrayUtils.hasNaNsOrInfs(beta): "Coefficients contain NA or Infs.";
       _global_beta=beta;
       _submodels = new Submodel[]{new Submodel(0,beta,-1,Double.NaN,Double.NaN)};
     }
 
+    public GLMOutput(DataInfo dinfo, String[] column_names, String[] column_types, String[][] domains, 
+                     String[] coefficient_names, double[] beta, boolean isMultinomial) {
+      super(dinfo._weights, dinfo._offset, dinfo._fold);
```
honzasterba, line 1188: do we need to copy past this code? I think there is a way how to extend the current
`    public GLMOutput(DataInfo dinfo, String[] column_names, String[] column_types, String[][] domains, String[] coefficient_names, boolean binomial) {
`
contructor and just call this() here

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: I just combine all of them into one constructor.

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: Deleted.

PR #4661, opened 2020-05-29
PUBDEV-7576: Doc updates to PDP topics

angela0xdata:
- In the Flow chapter, noted that PDPs can be returned for binary or multiclass classification
- In the Performance and Prediction chapter, noted that PDPs can be returned for multiclass classification and added a link to the pdp for multiclass ipynb demo. Also added an example showing how to retrieve 2dim pdps. 
Driveby fix: formatting fix in Prediction section.

### DIFF
--- a/h2o-docs/src/product/flow.rst
+++ b/h2o-docs/src/product/flow.rst
@@ -1329,6 +1329,11 @@ Partial Dependence Plots
 
 For models that include only numerical values, you can view a Partial Dependence Plot (PDP) for that model. This provides a graphical representation of the marginal effect of a variable on the class probability (binary and multiclass classification) or response (regression). 
 
+**Notes**:
+
+- For multiclass problems, users must specify a target class
+- In Flow (unlike Python or R), only one target class can be specified for multiclass problems. 
+
 Viewing Partial Dependence Plots
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/h2o-docs/src/product/flow.rst
+++ b/h2o-docs/src/product/flow.rst
@@ -1327,7 +1327,7 @@ The lower-left side of the graph represents less tolerance for false positives w
 Partial Dependence Plots
 ------------------------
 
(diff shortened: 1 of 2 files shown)


h2o-docs/src/product/performance-and-prediction.rst
```diff
@@ -1564,9 +1564,10 @@ Thus, the one-dimensional partial dependence of function :math:`g` on :math:`X_j
 
 **Notes**:
 
-- The partial dependence of a given feature is :math:`Xj` (where :math:`j` is the column index)
-- You can also change the equation to sum from 1 to N instead of 0 to N-1
-- Use the ``col_pairs_2dpdp`` option along with a list containing pairs of column names to generate 2D partial dependence plots
+- The partial dependence of a given feature is :math:`Xj` (where :math:`j` is the column index).
+- You can also change the equation to sum from 1 to N instead of 0 to N-1.
+- Use the ``col_pairs_2dpdp`` option along with a list containing pairs of column names to generate 2D partial dependence plots.
+- A `Python demo <https://github.com/h2oai/h2o-3/blob/master/h2o-py/demos/pdp_multiclass.ipynb>`__ is available showing how to retrieve PDPs for multiclass problems.
```
maurever, line 1570: I suggest to add something like: 

```suggestion
- For multiclass problems, use the ``targets`` parameter. A `Python demo <https://github.com/h2oai/h2o-3/blob/master/h2o-py/demos/pdp_multiclass.ipynb>`__ is available showing how to retrieve PDPs for multiclass problems.
```

h2o-docs/src/product/flow.rst
```diff
@@ -1327,7 +1327,7 @@ The lower-left side of the graph represents less tolerance for false positives w
 Partial Dependence Plots
 ------------------------
 
-For models that include only numerical values, you can view a Partial Dependence Plot (PDP) for that model. This provides a graphical representation of the marginal effect of a variable on the class probability (classification) or response (regression). 
+For models that include only numerical values, you can view a Partial Dependence Plot (PDP) for that model. This provides a graphical representation of the marginal effect of a variable on the class probability (binary and multiclass classification) or response (regression).
```
maurever, line 1330: Could we also add a note that for multiclass classification a user has to define one target class?

PR #4663, opened 2020-05-29
PUBDEV-7144: Add examples for saving and loading grids and for parallelism to User Guide

hannah-tillman:


h2o-docs/src/product/grid-search.rst
```diff
@@ -42,6 +42,8 @@ Grid Search in R and Python
           parameters (e.g., ``list(ntrees = c(1,100), learn_rate = c(0.1, 0.001))``)
        -  ``search_criteria`` optional attribute for specifying a more
           advanced search strategy  
+       -  ``parallelism`` (number of models built in parallel) allows
```
angela0xdata, line 45: Add more info for both R and Python

``parallelism`` The number of models to build in parallel. Parallelism allows the leader node to search the hyperspace and build models in a parallel way, which ultimately speeds up grid search on small data. A value of 1 (default) specifies sequential building. Specify 0 for adaptive parallelism, which is decided by H2O. Any number > 1 sets the exact number of models built in parallel.

h2o-docs/src/product/grid-search.rst
```diff
@@ -85,6 +89,169 @@ Grid Search in R and Python
         {'strategy': "RandomDiscrete", 'stopping_tolerance': 0.001, 'stopping_rounds': 10}
         {'strategy': "RandomDiscrete", 'stopping_metric': "misclassification", 'stopping_tolerance': 0.0005, 'stopping_rounds': 5}
 
+Saving and Loading a Grid Search
```
angela0xdata, line 92: These examples should go after the Grid Search and Random Grid Search examples. (Basically, you'd want to describe how to build grid search models before describing how to save them.) Also, maybe these can be a conclusion of the Grid Search example rather than a completely new one.

h2o-docs/src/product/grid-search.rst
```diff
@@ -85,6 +89,169 @@ Grid Search in R and Python
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+H2O supports saving and loading grids even after a cluster wipe or complete cluster restart. The ``save_grid`` function will export a grid and its models into a given folder while the ``load_grid`` function loads a previously saved grid and all its models from the given folder.
+
+There are two modes to save a grid (in both R and Python):
+
+- Use auto-checkpointing and supply the ``export_checkpoitns_dir`` parameter
```
angela0xdata, line 99: ``export_checkpoints_dir`` (minor typo)

h2o-docs/src/product/grid-search.rst
```diff
@@ -85,6 +89,169 @@ Grid Search in R and Python
...
+- Call the new function ``h2o.save_grid`` for manual export
```
angela0xdata, line 100: Avoid "new" because who knows when we'll finally take that out.

- Call the ``h2o.save_grid`` function for manual export.

h2o-docs/src/product/grid-search.rst
hannah-tillman: The examples now use the Grid Search example--please let me know if how I integrated it makes sense!

h2o-docs/src/product/grid-search.rst
```diff
@@ -294,6 +297,138 @@ Random Grid Search Examples
...
+    h2o.removeAll()
+
+    # Retrieve the saved grid
+    grid <- h2o.loadGrid(paste0(tempdir(), "/", grid_id))
+    grid
+
+
+  .. code-tab:: python
+
+    # Add the save location
```
angela0xdata, line 343: you need to import tempfile before you can run this command.

h2o-docs/src/product/grid-search.rst
```diff
@@ -294,6 +297,138 @@ Random Grid Search Examples
...
+    # Identify the grid_id and model_ids
+    grid_id = gbm_grid1.grid_id
+    old_grid_model_count = len(gbm_grid1.model_ids)
+
+    # Save the grid
+    saved_path = h2o.save_grid(checkpoints_dir, grid_id)
+
+    # Wipe the cloud to simulate cluster restart 
+    #(the models will no longer be available)
+    h2o.removeAll()
```
angela0xdata, line 425: For Python, this should be `h2o.remove_all()`

h2o-docs/src/product/grid-search.rst
```diff
@@ -294,6 +297,138 @@ Random Grid Search Examples
...
+
+    # Retrieve the saved grid
+    grid <- h2o.loadGrid(saved_path)
+    grid
+
+
+  .. code-tab:: python
+
+    # Add the save location
```
angela0xdata, line 404: Same comment here. Add `import tempfile` because users will likely run one example or the other.

PR #4665, opened 2020-06-01
PUBDEV-7586 - Create official Docker hub image

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7586

### DIFF
--- a/docker/public/Dockerfile-h2o-release
+++ b/docker/public/Dockerfile-h2o-release
@@ -1,6 +1,6 @@
 FROM openjdk:11
 
 RUN mkdir -p /opt/h2oai/h2o-3/
-COPY h2o.jar /opt/h2oai/h2o-3/
+COPY build/git sh2o.jar /opt/h2oai/h2o-3/
 
 CMD java -XX:+UseContainerSupport -jar /opt/h2oai/h2o-3/h2o.jar
--- a/docker/public/Dockerfile-h2o-release
+++ b/docker/public/Dockerfile-h2o-release
@@ -1,0 +1,6 @@
+FROM openjdk:11
+
+RUN mkdir -p /opt/h2oai/h2o-3/
+COPY h2o.jar /opt/h2oai/h2o-3/
+
+CMD java -XX:+UseContainerSupport -jar /opt/h2oai/h2o-3/h2o.jar
--- a/docker/public/Dockerfile-h2o-release
+++ b/docker/public/Dockerfile-h2o-release
@@ -1,6 +1,0 @@
-FROM openjdk:11
-
-RUN mkdir -p /opt/h2oai/h2o-3/
-COPY h2o.jar /opt/h2oai/h2o-3/
-
-CMD java -XX:+UseContainerSupport -jar /opt/h2oai/h2o-3/h2o.jar
--- a/docker/public/Dockerfile-h2o-release
+++ b/docker/public/Dockerfile-h2o-release
@@ -1,6 +1,6 @@
 FROM openjdk:11
 
 RUN mkdir -p /opt/h2oai/h2o-3/
-COPY build/git sh2o.jar /opt/h2oai/h2o-3/
+COPY h2o.jar /opt/h2oai/h2o-3/
 
 CMD java -XX:+UseContainerSupport -jar /opt/h2oai/h2o-3/h2o.jar
--- a/scripts/jenkins/Makefile.jenkins
+++ b/scripts/jenkins/Makefile.jenkins
@@ -487,9 +487,9 @@ check-leaks:
 	fi
 
 h2o-docker-build:
-    docker build -t $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG) -f docker/public/Dockerfile-h2o-release .
+	docker build -t $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG) -f docker/public/Dockerfile-h2o-release .
 
 h2o-docker-push:
 	docker login -u $(DOCKERHUB_USERNAME) -p $(DOCKERHUB_PASSWORD)
-    docker push $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)
+	docker push $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)
 
--- a/scripts/jenkins/Makefile.jenkins
+++ b/scripts/jenkins/Makefile.jenkins
@@ -486,17 +486,17 @@ check-leaks:
 		exit 1; \
 	fi
 
-h2o-docker-build:
-	ifeq ($(DOCKER_IMAGE_LATEST_TAG),)
-		SECOND_TAG=-t $(DOCKER_IMAGE_NAME):${DOCKER_IMAGE_LATEST_TAG}
-	endif
-	docker build -t $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG) $(SECOND_TAG) -f docker/public/Dockerfile-h2o-release build/
+h2o-k8s-docker-build:
+	 docker build -t $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG) -f docker/public/Dockerfile-h2o-release build/
 
-h2o-docker-push:
+h2o-k8s-docker-push:
 	docker login -u $(DOCKERHUB_USERNAME) -p $(DOCKERHUB_PASSWORD)
 	docker push $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)
-	ifeq ($(DOCKER_IMAGE_LATEST_TAG),)
-		docker push $(DOCKER_IMAGE_NAME):${DOCKER_IMAGE_LATEST_TAG}
-	endif
 
+h2o-k8s-docker-build-latest:
+	docker build -t $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG) -t $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_LATEST_TAG) -f docker/public/Dockerfile-h2o-release build/
 
+h2o-k8s-docker-push-latest:
+	docker login -u $(DOCKERHUB_USERNAME) -p $(DOCKERHUB_PASSWORD)
+	docker push $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)
+	docker push $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_LATEST_TAG)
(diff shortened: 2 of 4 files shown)


docker/public/Dockerfile-h2o-release
```diff
@@ -0,0 +1,6 @@
+FROM openjdk:11
+
+RUN mkdir -p /opt/h2oai/h2o-3/
+COPY h2o.jar /opt/h2oai/h2o-3/
+
+CMD java -XX:+UseContainerSupport -jar /opt/h2oai/h2o-3/h2o.jar
```
Pscheidl, line 6: Default, overridable. `-XX:+UseContainerSupport` should be enabled by default according to https://hub.docker.com/_/openjdk , yet I added it just to be sure.

scripts/jenkins/Makefile.jenkins
```diff
@@ -485,3 +485,11 @@ check-leaks:
 		echo "Credentials leaked! Please check $(PWD)/leak-check.out"; \
 		exit 1; \
 	fi
+
+h2o-docker-build:
+    docker build -t h2oai/h2o-open-source:$(PROJECT_VERSION) -f docker/public/Dockerfile-h2o-release .
```
Pscheidl, line 490: `PROJECT_VERSION` is not available in nightly - we need to export that in nightly pipeline.

scripts/jenkins/Makefile.jenkins
```diff
@@ -485,3 +485,11 @@ check-leaks:
 		echo "Credentials leaked! Please check $(PWD)/leak-check.out"; \
 		exit 1; \
 	fi
+
+h2o-docker-build:
+    docker build -t h2oai/h2o-open-source:$(DOCKER_IMAGE_TAG) -f docker/public/Dockerfile-h2o-release .
```
michalkurka, line 490: I think it might be a good idea to denote that this image is meant for k8s deployments. WDYT?

Pscheidl: Agreed. Doing it now.

Pscheidl: Done.

Pscheidl:
Sparkling water waits for this (@jakubhava @mn-mikke).

scripts/jenkins/Makefile.jenkins
```diff
@@ -485,3 +485,11 @@ check-leaks:
 		echo "Credentials leaked! Please check $(PWD)/leak-check.out"; \
 		exit 1; \
 	fi
+
+h2o-docker-build:
+    docker build -t $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG) -f docker/public/Dockerfile-h2o-release .
+
+h2o-docker-push:
+	docker login -u $(DOCKERHUB_USERNAME) -p $(DOCKERHUB_PASSWORD)
+    docker push $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)
```
michal-raska, line 494: nit: indentation

scripts/jenkins/Makefile.jenkins
```diff
@@ -485,3 +485,11 @@ check-leaks:
 		echo "Credentials leaked! Please check $(PWD)/leak-check.out"; \
 		exit 1; \
 	fi
+
+h2o-docker-build:
+    docker build -t $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG) -f docker/public/Dockerfile-h2o-release .
```
michal-raska, line 490: nit: indentation, seems like spaces are used instead of tabs

scripts/jenkins/jenkinsfiles/Jenkinsfile-Release
```diff
@@ -216,6 +217,34 @@ try {
...
+                def BUILD_H2O_DOCKER_STAGE_NAME = 'Build H2O Docker container'
+                if (params.BUILD_H2O_DOCKER) {
+                    stage(BUILD_H2O_DOCKER_STAGE_NAME) {
+                        try {
+                            pipelineContext.getBuildSummary().addStageSummary(this, BUILD_H2O_DOCKER_STAGE_NAME, env.BUILD_NUMBER_DIR)
+                            pipelineContext.getBuildSummary().setStageDetails(this, BUILD_H2O_DOCKER_STAGE_NAME, env.NODE_NAME, env.WORKSPACE)
+                            withCredentials([usernamePassword(credentialsId: 'dockerhub', passwordVariable: 'DOCKERHUB_PASSWORD', usernameVariable: 'DOCKERHUB_USERNAME')]) {
+                                environment {
+                                    if (params.TEST_RELEASE) {
+                                        DOCKER_IMAGE_NAME = "h2oai/h2o-open-source-k8s-test"
```
michal-raska, line 229: q: this means, that if two branches will be built, they'll produce the test image with the same tag (so it'll be overwritten). I'm not sure that's an issue. Maybe I would like to have the branch name part of the image for branches other than release. Or we could push to harbor for test releases and docker hub for "real" release. WDYT?

scripts/jenkins/Makefile.jenkins
Pscheidl: Replaced with tabs. Fixed. Thanks.

scripts/jenkins/Makefile.jenkins
Pscheidl: Replaced with tabs everywhere. Fixed. Thanks.

scripts/jenkins/jenkinsfiles/Jenkinsfile-Release
Pscheidl: tag will be different. This is name. The tag is the version and that should be different per each release really. If the tag is the same, it will be overridden.

michal-raska: fair point. I'm not sure that the tags are different though. I think it picks whatever is in `gradle.properties` and appends build number to it. I don't think branch name is used in any way.

Pscheidl: Version is just a number in terms of nightly. When doing proper release, the version should be unique. Or did I miss something ?

scripts/jenkins/Makefile.jenkins
```diff
@@ -485,3 +485,11 @@ check-leaks:
 		echo "Credentials leaked! Please check $(PWD)/leak-check.out"; \
 		exit 1; \
 	fi
+
+h2o-docker-build:
+	docker build -t $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG) -f docker/public/Dockerfile-h2o-release .
+
+h2o-docker-push:
+	docker login -u $(DOCKERHUB_USERNAME) -p $(DOCKERHUB_PASSWORD)
+	docker push $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)
```
honzasterba, line 494: should we also update latest?

Pscheidl: I'm not sure what should be latest. I'd say latest is not just anything that came out or nightly. From my POV, it should be latest stable release. WDYT ?

scripts/jenkins/jenkinsfiles/Jenkinsfile-Release
Pscheidl: Michal Raška is correct, the nightly tag looks like this: `docker build -t h2oai/h2o-open-source-k8s-test:3.30.0.9 -f docker/public/Dockerfile-h2o-release .`
I'll correct it.

scripts/jenkins/Makefile.jenkins
honzasterba: yes, latest should be stable, we can have also latest-nightly tag

michalkurka: latest should correspond with what is on the download page

Pscheidl: That is any latest release, doesn't matter if from master or from rel- branch - or ?

honzasterba: I think nightly is not really a release. So I would stick to stable, or we can switch the tags to latest and latest-stable, WDYT @michalkurka ?

michalkurka: I wouldn't worry about the nightly builds at all. latest should correspond point to latest stable. nightly builds are sometimes misunderstood, having them tagged as "latest" could add to the confusion

Pscheidl:
A couple of new changes:

- After discussion with @mmalohlava , the official image will go to `h2oai/` namespace, not into `opsh2oai` namespace. The test release package will go to `opsh2oai`.
- The `latest` tag is only used for proper releases, not nightlies. There is no `latest-nightly`, as according to @michalkurka , this might be confusing. The latest release, be it major or fix release, will always bear the `latest` tag.
- Tag of proper releases (not nighlies) will be `env["DOCKER_IMAGE_TAG"] = env["PROJECT_VERSION"]` - e.g. `:3.30.0.4` or `3.31.0.1`.
- Tag of nightly releases will have `env["BRANCH_NAME"] + "-" + currentBuild.number` as a tag. E.g. `rel-zahradnik-382` the last number is build number. This way, there is a proper distinction among nightlies. Also, this gives the user idea about which branch the nightly build is. Our S3 repository works in the same way (branch-name/build-number/).

Please have your say everyone @michalkurka @honzasterba @jakubhava @michal-raska whether you like this or not. Asking everyone to re-review, as this changed.

honzasterba:
just an idea: do we really need a separate tag for each nightly? thats a lot of images generated and stored for no good reason, what about tagging all nightlies for one version the same? 3.31.0.NIGHTLY

Pscheidl:
> just an idea: do we really need a separate tag for each nightly? thats a lot of images generated and stored for no good reason, what about tagging all nightlies for one version the same? 3.31.0.NIGHTLY

Interesting idea. The only con I can think of is somebody relying on yesterday's image. Which is not recommended anyway. I agree less images are actually more. Therefore, I support this idea. What do others think ?

Pscheidl:
I know Sparkling water is waiting for this, would such regime of nightly releases suit you @jakubhava and @mn-mikke ? Thanks.

michal-raska:
> The latest tag is only used for proper releases, not nightlies. There is no latest-nightly, as according to @michalkurka , this might be confusing. The latest release, be it major or fix release, will always bear the latest tag.

I think, we need to make sure, that if we have 3.22.0.1 as latest and we release 3.20.0.8 for whatever reason, it'll not overwrite the latest it it'll still point to 3.22.0.1.

Btw I'm not sure how this is handled for the downloads page...

michal-raska:
...maybe we can use the same variable `params.UPDATE_LATEST` to make the decision, whether "this" docker image should be promoted as latest.

Pscheidl:
Good idea @michal-raska, I'll make it so !

Pscheidl:
Works: https://hub.docker.com/repository/docker/opsh2oai/h2o-open-source-k8s-test

PR #4666, opened 2020-06-01
PUBDEV-7217: Add information about print_mojo to Productionizing chapter; add print_mojo example to Python docs

hannah-tillman:


h2o-docs/src/product/mojo-quickstart.rst
```diff
@@ -693,6 +693,54 @@ If you cannot install Graphviz on your environment, another option is to produce
   :alt: Example Output Picture
   :scale: 50%
 
+GBM MOJO to ONNX Converter
```
honzasterba, line 696: this is not accurate, the json output is just a intermediate format which is then used by ONNXMLTools to generate the final ONNX model.

https://github.com/onnx/onnxmltools#h2o-to-onnx-conversion

angela0xdata: @hannah-tillman just change the heading to "Printing GBM MOJOs"

h2o-docs/src/product/mojo-quickstart.rst
```diff
@@ -693,6 +693,54 @@ If you cannot install Graphviz on your environment, another option is to produce
   :alt: Example Output Picture
   :scale: 50%
 
+Printing GBM MOJOs
+~~~~~~~~~~~~~~~~~~
+
+The ``print_mojo`` function allows a GBM model to be coverted to `ONNX <https://onnx.ai/about.html>`__ representation, which means the model will be rendered in an "open format built to represent" ML models. This allows you to use the model with a variety of frameworks, tools, runtimes, and compilers.
```
honzasterba, line 699: this is still innacurate, the output has nothing to do with ONNX, also this works for all tree-like models

I would reduce this to

"The ``print_mojo`` function allows a model to be converted to machine readable representation - either to JSON for further processing, dot for rendering images.

@koniecsveta maybe we should update print_mojofunction to also allow PNG output?

koniecsveta: yesyes, I think it will be an easy thing after adding it to flow.

PR #4667, opened 2020-06-02
PUBDEV-7560 - example flows should not rely on filesystem data

honzasterba:


honzasterba:
thanks, yes, tested all locally and all work

PR #4670, opened 2020-06-02
PUBDEV-7444 add AUTO option for the family parameter of GLM

koniecsveta:
https://0xdata.atlassian.net/browse/PUBDEV-7444

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -381,7 +381,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
   private transient GLMModel _model;
   @Override
   public int nclasses() {
-    if (_parms._family == Family.multinomial || _parms._family == Family.ordinal)
+    if (_parms._family == Family.multinomial || _parms._family == Family.ordinal || _parms._family == Family.AUTO)
       return _nclass;
     if (Family.binomial == _parms._family || Family.quasibinomial == _parms._family
             || Family.fractionalbinomial == _parms._family)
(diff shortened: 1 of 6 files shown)


h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -438,6 +440,16 @@ public void init(boolean expensive) {
         error("_solver", "Solvers GRADIENT_DESCENT_LH and GRADIENT_DESCENT_SQERR are only " +
                 "supported for ordinal regression.  Do not choose them unless you specify your family to be ordinal");
       switch (_parms._family) {
```
wendycwong, line 442: When a user specify Family.AUTO, what it mean is we will try to assign a family to the dataset, right?  So, instead of carrying the Family.AUTO, are we going to assign it a family during init()?  For example, if nclass() ==1, we assign Family.Gaussian, nclass() == 2, we assign Family.Binomial, nclass() > 2, we assign multinomial?

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -519,6 +531,15 @@ public void init(boolean expensive) {
...
+      if (_parms._family == Family.AUTO) {
+        if (_nclass == 1) {
+          _parms._family = Family.gaussian;
+        } else if (_nclass == 2) {
+          _parms._family = Family.binomial;
+        } else {
+          _parms._family = Family.multinomial;
+        }
+      }
       if (_parms._link == Link.family_default)
```
wendycwong, line 543: If you move line 534 to 542 before line 442, you don't need to worry about any options for Family.AUTO, right?

koniecsveta: Yesyes - I just thought that when you input AUTO and incompatible link for given response and get error that AUTO requires different link for this response, it could be more clear than getting error that your link is incomatible for some specific family which you haven't input - but in the end it's the same thing, and will tell you which link is compatible so I can move it.

PR #4674, opened 2020-06-03
PUBDEV-7593 h2o version fix when running from whl

koniecsveta:
https://0xdata.atlassian.net/browse/PUBDEV-7593

### DIFF
--- a/h2o-py/h2o/__init__.py
+++ b/h2o-py/h2o/__init__.py
@@ -26,21 +26,36 @@ from h2o.h2o import (connect, init, api, connection,
 from h2o.frame import H2OFrame  # NOQA
 from h2o.utils.shared_utils import mojo_predict_csv, mojo_predict_pandas
 
+import zipfile
 import os
 from codecs import open
 here = os.path.abspath(os.path.dirname(__file__))
 
-try:
+def readTxtFromWhl(name, fallback):
+    if here.endswith('.whl/h2o'):
+        with zipfile.ZipFile(here[:-4]) as whl:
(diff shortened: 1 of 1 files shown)


h2o-py/h2o/__init__.py
```diff
@@ -26,19 +26,30 @@
...
 
+import zipfile
 import os
 from codecs import open
 here = os.path.abspath(os.path.dirname(__file__))
 
 try:
-    with open(os.path.join(here, 'buildinfo.txt'), encoding='utf-8') as f:
-        __buildinfo__ = f.read()
+    if here.endswith('.whl/h2o'):
```
honzasterba, line 35: there is no guaratee that here.endswith('.whl/h2o') means .whl is a zipfile, I see two options
1) make this a fallback when the original open() fails 
2) add check that the parent is whl and it is in fact a regular file not a dir
i think 1) is better

h2o-py/h2o/__init__.py
```diff
@@ -26,19 +26,30 @@
...
+    else:            
+        with open(os.path.join(here, 'buildinfo.txt'), encoding='utf-8') as f:
+            __buildinfo__ = f.read()
 except:
     __buildinfo__ = "unknown"
 
 try:
-    with open(os.path.join(here, 'version.txt'), encoding='utf-8') as f:
-        __version__ = f.read()
+    if here.endswith('.whl/h2o'):
```
honzasterba, line 46: also since this is duplicated i would move it to a function

PR #4678, opened 2020-06-03
PUBDEV-7594 add explicit dependency on webserver module

satai:
See https://0xdata.atlassian.net/browse/TN-13 why is this module needed.
To specifiy dependency via test runtime classpath is not enough - `testMultiNode`
can be run before other dependencies for `test` target are evaluated.

### DIFF
--- a/gradle/multiNodeTesting.gradle
+++ b/gradle/multiNodeTesting.gradle
@@ -1,6 +1,6 @@
 // Actually run the tests by building a cluster of free-running JVMs first.
 task testSingleNode(type: Exec) {
-    dependsOn jar, testJar, ":${defaultWebserverModule}:jar"
+    dependsOn jar, testJar
 
     environment "ROOT_DIR", rootDir
     environment "PROJECT_NAME", project.name
@@ -16,7 +16,7 @@ task testSingleNode(type: Exec) {
     commandLine args
 }
 task testSingleNodeOneProc(type: Exec) {
-    dependsOn jar, testJar, ":${defaultWebserverModule}:jar"
+    dependsOn jar, testJar
 
     environment "ROOT_DIR", rootDir
     environment "PROJECT_NAME", project.name
@@ -32,7 +32,7 @@ task testSingleNodeOneProc(type: Exec) {
     commandLine args
 }
 task testMultiNode(type: Exec) {
-    dependsOn jar, testJar, ":${defaultWebserverModule}:jar"
+    dependsOn jar, testJar
 
     environment "ROOT_DIR", rootDir
     environment "PROJECT_NAME", project.name
--- a/gradle/multiNodeTesting.gradle
+++ b/gradle/multiNodeTesting.gradle
@@ -1,6 +1,6 @@
 // Actually run the tests by building a cluster of free-running JVMs first.
 task testSingleNode(type: Exec) {
-    dependsOn jar, testJar
+    dependsOn jar, testJar, ":${defaultWebserverModule}:jar"
 
     environment "ROOT_DIR", rootDir
     environment "PROJECT_NAME", project.name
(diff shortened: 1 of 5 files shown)


gradle/multiNodeTesting.gradle
```diff
@@ -1,6 +1,6 @@
 // Actually run the tests by building a cluster of free-running JVMs first.
 task testSingleNode(type: Exec) {
-    dependsOn jar, testJar
+    dependsOn jar, testJar, ":${defaultWebserverModule}:jar"
```
Pscheidl, line 3: It's interesting that for our Gradle `jar` doesn't mean defaultWebserverModule jar.

Pscheidl: Investigating.

Pscheidl: testJar is defined as:

```
task testJar(type: Jar, dependsOn: testClasses, group: "build") {
    classifier = 'test'
    from sourceSets.test.output
}
```

honzasterba: agree with @Pscheidl that `jar` should already depend on  ":${defaultWebserverModule}:jar" but thats probably not true for all modules

Pscheidl: That task won't gather any testRuntimeOnly dependencies. testCompile works. I'm not even sure whether this task gets executed.

Pscheidl: Ok, two observations. This code is a copy&paste of some old stuff usable for old gradles. Obviously, in the resulting classpath, the runtime-only dependencies are not present.

This will fix it:

```
task testJar(type: Jar, dependsOn: testClasses, group: "build") {
    classifier = 'test'
    from sourceSets.test.output + sourceSets.test.runtimeClasspath
}
```

Just add the runtime classpath to the jar and we're done.

Pscheidl:
We should also replace every instance of `runtime project(":${defaultWebserverModule}")` with `runtimeOnly project(":${defaultWebserverModule}")` - @satai feel free to make it part of this PR if you agree. `runtimeOnly` is the official scope and `runtime` probably won't be supported in Gradle 7.

Pscheidl:
Thank you @satai for opening this. After today's discussion, @satai decided to review my changes, as I already had them locally to save him some time.

Now please review it, modify it and merge when you're ready. We should definitely wait for tests to finish.

Pscheidl:
After several builds failures on timeout while doing `testJar` goal on `h2o-core`, i investigated a bit. And indeed, this task seems to run indefinitely (who am I to solved the halting problem).

The rabbit hole seems to go deeper and there seems to be a classpath problem either introduced by my "fix" or something we overlooked.

Pscheidl:
Now I found the problem - as we're now making Gradle to actually compile runtime dependencies, it'll make the `h2o-test-support` module compile it's `h2o-core` dependency. However, `h2o-core` has runtimeOnly dependency on `testSupport` - and just like that, the circular dependency is created.

Solution is easy - have `h2o:core` as compileOnly, as at runtime, h2o:core is always provided.

In future, we should re-visit our dependency graph (Gradle has so means to visualize it) and maybe do some optimizations.

Pscheidl:
Such a solution solves the circular dependency issue, but core artifacts are no longer available in h2o-algos. I'd have to be added as a standalone dependency not just for `h2o-algos`, but also to every module previously counting on `test-support` module. 

This is not ideal. Converting test-support to `test` scope does not help either, as our testJar classpath would be OK, but the normal classpath when executing `gradle test` would be broken.

Therefore, reverting my latest change and starting over.

Pscheidl:
I think this leads us to the final solution - define the testJar artifact in a proper way.

satai:
LGTM, probably should be squashed in the end

Pscheidl:
This solution will make tests time out. It's not solved.

PR #4680, opened 2020-06-03
Log DEBUG information when S3 cache is being refreshed.

Pscheidl:
In the logs, there is no information whether S3 cache update has been triggered. This may help debug cache build-ups for large buckets. From experience, I know this might take quite some time.

We should definitely look into the cache collection, maybe update the client or increase the payload size...generally investigate how to help speed this up.

### DIFF
--- a/h2o-persist-s3/src/main/java/water/persist/PersistS3.java
+++ b/h2o-persist-s3/src/main/java/water/persist/PersistS3.java
@@ -439,6 +439,7 @@ public final class PersistS3 extends Persist {
 
     public boolean containsKey(String k) { return Arrays.binarySearch(_cache,k) >= 0;}
     protected String [] update(){
+      Log.debug("Renewing S3 bucket cache.");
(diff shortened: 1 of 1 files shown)


h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -439,6 +439,7 @@ public Key uriToKey(URI uri) throws IOException {
 
     public boolean containsKey(String k) { return Arrays.binarySearch(_cache,k) >= 0;}
     protected String [] update(){
+      Log.debug("Renewing S3 bucket cache.");
```
honzasterba, line 442: maybe next time use Log4j

Pscheidl: https://github.com/h2oai/h2o-3/pull/4682 Fixing it now. Thanks.

PR #4681, opened 2020-06-04
PUBDEV-7588: fix problem when Lambda=0 with interaction

wendycwong:
This PR fixes the problem in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7588.

When there is interaction and when Lambda=0, use_all_factor_level = false.  As a result, the number of levels for interaction column is reduced.  Hence, it will run into problem with the assertion: assert _valid : "Categorical value out of bounds, got " + val + ", next cat starts at " + fullCatOffsets()[cid + 1] in DataInfo.java method getCategoricalId().  The reason for the failure is the number of enum values are reduced.

I added a check to make sure the assert is only called under the following conditions: 
- useAllFactorLevels=True or
-  if useAllFactorLevels=False, isInteraction = False.

### DIFF
--- a/h2o-algos/src/main/java/hex/DataInfo.java
+++ b/h2o-algos/src/main/java/hex/DataInfo.java
@@ -470,7 +470,7 @@ public class DataInfo extends Keyed<DataInfo> {
     if(v.isCategorical()) {
       if (useAllFactorLevels) return v.mode();
       long[] bins = v.bins();
-      return ArrayUtils.maxIndex(bins,1);
+      return ArrayUtils.maxIndex(bins,0);
     }
     return (int)Math.round(v.mean());
   }
@@ -1081,7 +1081,10 @@ public class DataInfo extends Keyed<DataInfo> {
   public final int getCategoricalId(int cid, int val) {
     boolean isIWV = isInteractionVec(cid);
     if(val == -1) { // NA
-      val = _catNAFill[cid];
+      if (isIWV && !_useAllFactorLevels) 
+        val = _catNAFill[cid]-1; // need to -1 here because no -1 in 6 lines later for isIWV vector
+      else
+        val = _catNAFill[cid];
     }
 
     if (!_useAllFactorLevels && !isIWV) {  // categorical interaction vecs drop reference level in a special way
@@ -1091,7 +1094,7 @@ public class DataInfo extends Keyed<DataInfo> {
     int [] offs = fullCatOffsets();
     int expandedVal = val + offs[cid];
     if(expandedVal >= offs[cid+1]) {  // previously unseen level
-      assert _valid : "Categorical value out of bounds, got " + val + ", next cat starts at " + fullCatOffsets()[cid + 1];
+      assert (isIWV && !_useAllFactorLevels) || _valid : "Categorical value out of bounds, got " + val + ", next cat starts at " + fullCatOffsets()[cid + 1];
       if(_skipMissing)
         return -1;
       val = _catNAFill[cid];
--- a/h2o-core/src/main/java/water/fvec/InteractionWrappedVec.java
+++ b/h2o-core/src/main/java/water/fvec/InteractionWrappedVec.java
@@ -231,7 +231,7 @@ public class InteractionWrappedVec extends WrappedVec {
         if( (!((leftIsNA=left.isNA(i)) | (riteIsNA=rite.isNA(i)))) ) {
           lval = (int)left.at8(i);
           rval = (int)rite.at8(i);
-          if( !_useAllLvls && ( 0==lval || 0==rval )) {
+          if( !_useAllLvls && ( 0==lval || 0==rval )) { // deal with all illegal levels here when _useAllLvls==False
             _perChkMapMissing.putIfAbsent(_left[lval] + "_" + _rite[rval],"");
             continue;
           }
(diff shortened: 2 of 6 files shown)


h2o-algos/src/main/java/hex/DataInfo.java
```diff
@@ -1091,7 +1091,8 @@ public final int getCategoricalId(int cid, int val) {
     int [] offs = fullCatOffsets();
     int expandedVal = val + offs[cid];
     if(expandedVal >= offs[cid+1]) {  // previously unseen level
-      assert _valid : "Categorical value out of bounds, got " + val + ", next cat starts at " + fullCatOffsets()[cid + 1];
+      if (!isIWV || _useAllFactorLevels) // assertion only needed for non-interaction columns or _useAllFactorLevels=true
```
honzasterba, line 1094: why not merge with the assert? i.e.: assert (isIWV && !_useAllFactorLevels) || _isValid

honzasterba: or move into the if branch bewow

michalkurka:
@wendycwong please don't merge yet, I want to understand why is this correct

michalkurka:
we have use_all_factor_level = False

Consider a simple categorical-numeric interaction of 2 columns: X (domain: a, b, c), Y (numeric)

with `use_all_factor_level = False` we should see columns Xa.Y, Xb.Y (+ missing) in the expanded 1-hot encoded frame; there shouldn't be a column Xc.Y because that one is skipped

now the scenario of categorical-categorical interaction, X (domain: a, b), Y (domain: r, s)

if use_all_factor_level=True then we would have columns Xa.Yr, Xa.Yr, Xa.Ys, Xb.Ys, Xb.Ys, Xb.Ys, Xb.Yr, Xb.Yr (+ missing). When we use_all_factor_level we should IMHO only drop the last column Xb.Yr.

This makes me think we should never reach outside of the domain in getCategoricalId, except really for a validation frame with truly unseen values.

Please explain, it seems to me the bug is elsewhere.

wendycwong:
@michalk:

@michalk: Sure, I will wait.  But let me explain anyway.

When lambda=0, use_all_factor_level = False.  Hence, if the original factor level is 4, it will not become 3.  Hence, the check with assert.

However, when you have interaction and lambda=9, use_all_factor_level=False, the combined column categorical level is going to drop for more than 1!  Consider gender (2) and race (2), the final combined gender_race interaction column is of 1 level only!  This is because we try to combine gender (1) and race (1) due to use_all_factor_level=False.  Hence, it will fail the assert.

Concrete example: gender=<Female, Male>, race=<Red, Green>.  The complete interaction will be Female_Red, Female_Green, Male_Red, Male_Green.  If use_all_factor_level=False, then, the interaction domain will have Male_Green in this case.

michalkurka:
> Concrete example: gender=<Female, Male>, race=<Red, Green>. The complete interaction will be Female_Red, Female_Green, Male_Red, Male_Green. If use_all_factor_level=False, then, the interaction domain will have Male_Green in this case.

My point is that it shouldn't have that domain value, there should never be Male_Green. The code that is modified is calling `getCategoricalId` for Male_Green - but it should never be doing that because use_all_factor_level=False. It should never need to find out a categorical id for something that is not there.

What am I missing?

wendycwong:
@michalkurka 

For enum x enum interaction with 2 enum levels per column, then, there is only one categorical domain which is Male_Green.  The column will be 1 if gender = Male and race=Green.  However, its value is 0 if a. gender=Female, b. gender=Male but race = Red.  That is why we still need to check the interaction column enum level even though there is only one value.  Does this make sense?

michalkurka:
@wendycwong I am not sure what is the right way but it seems to me there is an inconsistency in the H2O code that is not addressed by this PR.

I debugged the code and found out that the modified method returns categorical id = 3 for the "unseen" categorical. This however doesn't correspond to any beta coefficient for predictor variables, it corresponds to the coefficient of the intercept. See screenshot

Screen Shot 2020-06-08 at 11 33 30 AM

The code is actually doing this:
1. Missing bucket is disabled (it will not create a column for missing)
2. The cat-cat interaction has just a single value, rest of it is na.
3. The missing categorical value is imputed with a value that is _not_ in the domain (if it was imputed correctly it would make the column constant...).

To me the issues in in method imputeCat which does seem to correctly work for InteractionWrappedVecs, please compare the result of `bins` to `getBins` as on the screenshot:

Screen Shot 2020-06-08 at 11 52 16 AM

`imputeCat` is clearly returning a value that doesn't exist in the domain of the InteractionWrappedVec. InteractionWrappedVecs are dropping the extra level in a special way and thus need a special handling in this method.

wendycwong:
@michalkurka 

Let's consider the gender (FEMALE) and income(<=50K, >50K).  With use_all_factor_levels=True, the interaction between the two will be:
FEMALE <= 50K
FEMALE >50K
MALE<=50K
MALE > 50K. 

GLM will have a coefficient for each of the four categorical domain.

Now, consider when use_all_factor_levels=False, then gender = MALE, income (>50K), the interaction domain will only be MALE_>50K.  I actually understand how they derive this domain value but I don't agree with it.  Hence, in this case, the value for the interaction column can be 1 when the gender is MALE and income >50K and 0 otherwise.

However, let me look into the code, something is strange there though.  Will get back to you after my investigations.

wendycwong:
@michalkurka  There are two issues here.  

1. When use_all_factor_levels = true, here is what the interaction vector domain looks like:
image

We are agree that this is correct.

wendycwong:
@michalk  Here is the confusing part.  When use_all_factor_levels = false.  Here is what the interaction vector looks like:

image

The interaction vector is generated from gender (MALE), income (>50K), hence, the domain is just MALE_>50K.  Also, the number of rows that have gender=MALE and income=>50K is 9918.  So, here by default,  it is going to impute with MALE_>50K since it recognize only one domain value.   I somehow don't think this is correct because the number of rows with the other gender and income combinations = 14423+22732+1769 > 9918.  I agree with you this is not right.

wendycwong:
@michalk:  To me, this is incorrect but consistent.

wendycwong:
Michalk:  I am all out of ideas...  Dead.

wendycwong:
@michalk:  Coming back at it.  There is something wrong in the code....

h2o-algos/src/main/java/hex/DataInfo.java
```diff
@@ -1091,7 +1094,7 @@ public final int getCategoricalId(int cid, int val) {
     int [] offs = fullCatOffsets();
     int expandedVal = val + offs[cid];
     if(expandedVal >= offs[cid+1]) {  // previously unseen level
-      assert _valid : "Categorical value out of bounds, got " + val + ", next cat starts at " + fullCatOffsets()[cid + 1];
+      assert (isIWV && !_useAllFactorLevels) || _valid : "Categorical value out of bounds, got " + val + ", next cat starts at " + fullCatOffsets()[cid + 1];
```
michalkurka, line 1097: is this assert still needed?

wendycwong: @michalk:  Yes, we still need to skip the original assert.  Because the use_all_factor_levels=False does not reduce the categorical levels like normal use.  For normal enum column, when use_all_factor_levels=False, the enum level just drop by one.  However, with interaction columns, the enum level is determined by number of enum level column1 (now is one less than the old level) times number of enum level column2(now is one less than the old level).  So, the assert still does not apply to interaction columns. However, the assert is still correct for regular enum columns and I would like to keep it there.

h2o-core/src/main/java/water/fvec/InteractionWrappedVec.java
```diff
@@ -351,9 +351,19 @@ else if( _perChkMapMissing!=null && t._perChkMapMissing!=null ) {
     @Override public double atd_impl(int idx) {
       if( _isCat )
         if( isNA_impl(idx) ) return Double.NaN;
-      return _isCat ? Arrays.binarySearch(_vec.domain(), getKey(idx)) : ( _c1IsCat?1: (_c[0].atd(idx))) * ( _c2IsCat?1: (_c[1].atd(idx)) );
+       if (_isCat) {
+         int val = Arrays.binarySearch(_vec.domain(), getKey(idx));
+         return ((val < 0)?-1:val);
+       } else
```
wendycwong, line 357: note that val = Arrays.binarySearch(_vec.domain(), getKey(idx)) can return bad values like -3.  Hence, I added the check to make sure we only return -1 for illegal values.

PR #4682, opened 2020-06-04
Change logger in PersistS3 class to Log4J

Pscheidl:


h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -31,10 +32,10 @@
 
 /** Persistence backend for S3 */
 public final class PersistS3 extends Persist {
-  private static final String HELP = "You can specify a credentials properties file with the -aws_credentials command line switch.";
 
   private static final String KEY_PREFIX = "s3://";
   private static final int KEY_PREFIX_LEN = KEY_PREFIX.length();
+  private static final Logger LOGGER = Logger.getLogger("water.default");
```
Pscheidl, line 38: @honzasterba Do we want to use this convention everywhere ? I'd put proper class in there.

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -31,7 +31,7 @@
 
 /** Persistence backend for S3 */
 public final class PersistS3 extends Persist {
-  private static final String HELP = "You can specify a credentials properties file with the -aws_credentials command line switch.";
```
Pscheidl, line 34: Unused.

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -259,8 +255,8 @@ public void importFiles(String path, String pattern, ArrayList<String> files, Ar
     }
   }
 
-  private static void ignoreAndWait(final Exception e, boolean printException) {
-    Log.ignore(e, "Hit the S3 reset problem, waiting and retrying...", printException);
+  private static void ignoreAndWait(final Exception e) {
+    LOGGER.debug("Hit the S3 reset problem, waiting and retrying...", e);
```
Pscheidl, line 259: I always want to print it. Former ignore was `debug` anyway. Why not print the exception all the time in debug ?

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -31,7 +31,7 @@
 
 /** Persistence backend for S3 */
 public final class PersistS3 extends Persist {
-  private static final String HELP = "You can specify a credentials properties file with the -aws_credentials command line switch.";
+  private static final Logger LOGGER = Logger.getLogger(PersistS3.class);
```
honzasterba, line 34: I would rename this to LOG, same I used in xgboost

honzasterba:
also this should go to master where we use Log4j, it will not work corectly in rel-zahrandik

h2o-persist-s3/src/main/java/water/persist/PersistS3.java
Pscheidl: Done.

Pscheidl:
@honzasterba Not it aims at master. I'll squash it before merging into one commit.

PR #4686, opened 2020-06-04
Fix pyunit_aggregator_mapping_frame

tomasfryda:
S3 contains now the `smalldata/wine/winequality-redwhite-no-BOM.csv`.


michalkurka:
Thank you, I owe you one!

PR #4689, opened 2020-06-05
Use 50% of available cgroup memory limit for JVM HEAP in official H2O Docker containers.

Pscheidl:


docker/public/Dockerfile-h2o-release
```diff
@@ -3,4 +3,4 @@ FROM openjdk:11
 RUN mkdir -p /opt/h2oai/h2o-3/
 COPY h2o.jar /opt/h2oai/h2o-3/
 
-CMD java -XX:+UseContainerSupport -jar /opt/h2oai/h2o-3/h2o.jar
+CMD java -XX:+UseContainerSupport -XX:MaxRAMPercentage=90 -jar /opt/h2oai/h2o-3/h2o.jar
```
michalkurka, line 6: With this setting we should disable XGBoost, @honzasterba what do you think?

Pscheidl:
@honzasterba @michalkurka or set it at least to 50 percent by default. And document the behavior for users to override the default CMD with higher number, in case they don't use XGBOOST. 

For Kubernetes itself, the planned extern XGBoost cluster makes perfect sense - independent scaling. 

This is really about setting the right default. 1/4 of available memory by default wastes resources.

michalkurka:
It is hard to provide a sensible default because the decision is between wasting memory (XGBoost enabled by default) or not giving users the option to run XGBoost by default.

On Hadoop, we ignore this issue... we just document it but there is no actual check. With a bit more controlled environment as we have thanks to docker IMHO we can do better.

@Pscheidl externalized XGB is the answer yes, but that is not available as of now.

I am leaning towards the option of using just 50% of the memory instead of disabling XGB. WDYT?

Pscheidl:
50 percent and documenting it everywhere will do. Give me some time to adjust the PR, I'm switched to other context. Few hours.

Pscheidl:
People need to know it's configurable and overridable.

docker/public/Dockerfile-h2o-release
honzasterba: yes, 10% is not enough

honzasterba:
I agree with 50% configurable default

Pscheidl:
Updating descriptions on Docker hub and in our K8S tutorials accordingly.

Pscheidl:
I'll do that separately, let's get this merged so it reaches tonight's nightly.

PR #4693, opened 2020-06-08
WIP PUBDEV-7608 - make tests compatible with pandas 1.0.x

satai:


honzasterba:
is this backwards compatible with 0.22? if not we will need to also update our docker images and test-requirements.txt to use pandas 1.0

honzasterba:
jenkins has the answer: `AttributeError: 'DataFrame' object has no attribute 'to_numpy'`

satai:
@honzasterba The test requirements should be already updated in this PR, docker images will follow (I hope).

sebhrusen:
@honzasterba , @satai 
> is this backwards compatible with 0.22?

pandas 1.0 requires Py3.6+, and H2O-3 still support Py2.7 and Py3.5+ so the test requirements should take this into consideration and we should still be able to use `pandas 0.2x` with Py versions not supported by `pandas 1.0`.
Not everyone switched to latest pandas already, especially with breaking changes, and we should ensure that H2O-3 client integration with pandas works with both old and new versions.

honzasterba:
if "pandas 1.0 requires Py3.6" than I am affraid we cannot merge this and have to stick with 0.22 until we can drop py2 support :(

sebhrusen:
@honzasterba if the direct usages of pandas API are not too big, you can create a small `pandas_compat.py` module for translation of older API: it should always be possible with our legacy code.

Pscheidl:
What is the latest version of Pandas compatible with H2O-3 ? Also, when are we going to drop support for Python 2.7 ? Do we know ?

michalkurka:
@satai should this be closed?

satai:
@michalkurka Yes, I think this can be closed

PR #4694, opened 2020-06-08
PUBDEV-7596: Make leaf node assignment work for tree-depths >= 31

michalkurka:
But still preserve the limit of 64 maximum path lenght


h2o-algos/src/test/java/hex/tree/drf/DRFTest.java
```diff
@@ -1834,31 +1834,52 @@ private void isFeatureUsedInPredictHelper(boolean ignoreConstCols, boolean multi
   }
 
   @Test
-  public void testDeepLeafNodeAssignmentConsistency() {
+  public void checkDeepLeafNodeAssignmentConsistency() {
```
Pscheidl, line 1837: I like this test - clears scope after every try.

PR #4697, opened 2020-06-08
PUBDEV-7609: use loading_name to pass x frame name in Java backend.

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7609

@jakubhava noticed that loading_frame parameter is never used.
It is supposed to allow user to choose a name to save the X frame input if they like.  I fix this problem by actually using it to set the name of the X frame.

However, in the model._output, _represntation_name is used to denote the frame name used to save the X frame.  I did not change this part.  

This will not break any user code however, we are now using two names to represent the same thing.  

@ledell What are your thoughts on this?  It is cleaner to just use one name but it may break user code if they use loading_frame in their code even though the loading_frame is not doing anything useful until my fix here.

### DIFF
--- a/h2o-r/tests/testdir_algos/glrm/runit_PUBDEV_7609_glrm_representation_loading_name.R
+++ b/h2o-r/tests/testdir_algos/glrm/runit_PUBDEV_7609_glrm_representation_loading_name.R
@@ -1,0 +1,42 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../../scripts/h2o-r-test-setup.R")
+
+test.glrm.arrests <- function() {
+  Log.info("Importing arrests.csv data...") 
+  arrestsR <- read.csv(locate("smalldata/pca_test/USArrests.csv"), header = TRUE)
+  arrestsH2O <- h2o.uploadFile(locate("smalldata/pca_test/USArrests.csv"), destination_frame = "arrestsH2O")
+  initCent <- scale(arrestsR, center = TRUE, scale = FALSE)[1:4,]
+  loading_name <- "are_you_confused"
+  representation_name <- "are_we_confused"
+  #use representation_name
+  fitH2O <- h2o.glrm(arrestsH2O, k = 4, init = "User", user_y = initCent, transform = "DEMEAN", 
+                     loss = "Quadratic", regularization_x = "None", regularization_y = "None", 
+                     recover_svd = TRUE, representation_name=representation_name)
+  print("fitH2O@model$representation_name is ")
+  print(fitH2O@model$representation_name)
+  print("representation name is ")
+  print(representation_name)
+  expect_true(fitH2O@model$representation_name==representation_name)
+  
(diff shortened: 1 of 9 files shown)


h2o-py/tests/testdir_algos/glrm/pyunit_PUBDEV_7609_GLRM_loading_name.py
```diff
@@ -0,0 +1,37 @@
...
+               [1.012,  19.24, -15.54, -1.732]]
+  initial_y_h2o = h2o.H2OFrame(list(zip(*initial_y)))
+  initial_y_h2o.show()
+
+  print("H2O GLRM on de-meaned data with quadratic loss:\n")
+  glrm_h2o = H2OGeneralizedLowRankEstimator(k=4, transform="DEMEAN", loss="Quadratic", gamma_x=0, gamma_y=0, 
+                                            init="User", user_y=initial_y_h2o, recover_svd=True,
+                                            loading_name=loading_name)
+  glrm_h2o.train(x=arrestsH2O.names, training_frame=arrestsH2O)
+  assert loading_name == str(glrm_h2o._model_json["output"]['representation_name']), \
```
michalkurka, line 28: I think this line demonstrates the issue of this approach, there is an inconsistency between the input/output API

since loading_frame was never used, I would suggest this:
1. expose representation_name
2. map loading_name to representation name in the V3 class
3. deprecate loading_name and modify the documentation to use representation_name
4. temporarily keep both loading_name and representation_name
5. delete loading_name from the API

wendycwong: I like this approach.  Let me put it in.

h2o-r/tests/testdir_algos/glrm/runit_PUBDEV_7609_glrm_representation_loading_name.R
```diff
@@ -0,0 +1,30 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../../scripts/h2o-r-test-setup.R")
+
+test.glrm.arrests <- function() {
+  Log.info("Importing arrests.csv data...") 
+  arrestsR <- read.csv(locate("smalldata/pca_test/USArrests.csv"), header = TRUE)
+  arrestsH2O <- h2o.uploadFile(locate("smalldata/pca_test/USArrests.csv"), destination_frame = "arrestsH2O")
+  initCent <- scale(arrestsR, center = TRUE, scale = FALSE)[1:4,]
+  loading_name = "are_you_confused"
```
michalkurka, line 9: it is a confusing world!

however, I would still suggest to use

    loading_name <- "are_you_confused"

to make it less confusing ;)

wendycwong: Right.  Let me change it!

PR #4698, opened 2020-06-09
PUBDEV-7444_make_AUTO_GLMs_default_family

koniecsveta:
https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7444

### DIFF
--- a/h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7181_check_model_scoring.py
+++ b/h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7181_check_model_scoring.py
@@ -16,7 +16,9 @@ def test_gam_model_predict():
     h2o_data["C2"] = h2o_data["C2"].asfactor()
     myY = "C21"
     model_test_data = h2o.import_file(pyunit_utils.locate("smalldata/gam_test/predictGaussianGAM1.csv"))
-    buildModelCheckPredict(h2o_data, h2o_data,  model_test_data, myY, ["C11", "C12", "C13"], 'gaussian')
+    pred_gauss = buildModelCheckPredict(h2o_data, h2o_data,  model_test_data, myY, ["C11", "C12", "C13"], 'gaussian', 'gaussian')
+    pred_auto_gauss = buildModelCheckPredict(h2o_data, h2o_data,  model_test_data, myY, ["C11", "C12", "C13"], 'AUTO', 'gaussian')
+    pyunit_utils.compare_frames_local(pred_gauss, pred_auto_gauss, prob=1)
 
     print("Checking model scoring for multinomial")
     h2o_data = h2o.import_file(pyunit_utils.locate("smalldata/glm_test/multinomial_10_classes_10_cols_10000_Rows_train.csv"))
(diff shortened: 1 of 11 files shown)


h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
```diff
@@ -86,7 +86,7 @@ public static void copy2DArray(double[][] src_array, double[][] dest_array) {
   }
 
   public static GLMParameters copyGAMParams2GLMParams(GAMParameters parms, Frame trainData, Frame valid) {
-    GLMParameters glmParam = new GLMParameters();
+    GLMParameters glmParam = new GLMParameters(GLMParameters.Family.gaussian, GLMParameters.Link.family_default);
```
Pscheidl, line 89: Is this necessary ? GLM receive whatever is set to GAM, be it GAM's default or custom value. We should ask @wendycwong if this should be the default for GAM as well.

wendycwong: @koniecsveta @Pscheidl : I looked at GLMParameters constructor without any arguments:

    public GLMParameters() {
      this(Family.gaussian, Link.family_default);
      assert _link == Link.family_default;
      _stopping_rounds = 3;
      _stopping_metric = ScoreKeeper.StoppingMetric.deviance;
      _stopping_tolerance = 1e-4;
    }

It already sets to Gaussian family.  However, now I am wondering if this should be changed to family AUTO?

I agree with @Pscheidl that there is no need to add arguments to new GLMParameters.  I also agree with @koniecsveta changing the constructor to GLMParameters without parameters to what she did below.

koniecsveta:
> @koniecsveta From a user’s perspective, what does `auto` mean?
> 
> For example, if a user specifies `family="auto"` and `link="tweedie"`, will `auto` default to the `tweedie` link?
> 
> Or if link isn’t specified or if it is specified as `Family_Default`, then does `auto` default to `gaussian` for real and int response?

@angela0xdata If you use family = AUTO and don't specify link, link = default_family is used (default for the family to which AUTO falls). If you use family = AUTO and do specify link and if specified link is compatible with the family to which AUTO falls, then this link is used, otherwise you get error message saying that AUTO for underlying data requires different link + list of possible compatible links.

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7181_check_model_scoring.py
```diff
@@ -16,7 +16,9 @@ def test_gam_model_predict():
     h2o_data["C2"] = h2o_data["C2"].asfactor()
     myY = "C21"
     model_test_data = h2o.import_file(pyunit_utils.locate("smalldata/gam_test/predictGaussianGAM1.csv"))
-    buildModelCheckPredict(h2o_data, h2o_data,  model_test_data, myY, ["C11", "C12", "C13"], 'gaussian')
+    pred_gauss = buildModelCheckPredict(h2o_data, h2o_data,  model_test_data, myY, ["C11", "C12", "C13"], 'gaussian', 'gaussian')
```
Pscheidl, line 19: Nobody will look for tests verifying PUBDEV-7444 under test for PUBDEV-7181.

tomasfryda:
I'm thinking whether we should update the documentation. What do you think? 

I'm mainly concerned about these 3 files:
- `h2o-docs/src/product/data-science/gam.rst`
- `h2o-docs/src/product/data-science/glm.rst`
- `h2o-docs/src/product/data-science/algo-params/family.rst`

koniecsveta:
> I'm thinking whether we should update the documentation. What do you think?
> 
> I'm mainly concerned about these 3 files:
> 
> * `h2o-docs/src/product/data-science/gam.rst`
> * `h2o-docs/src/product/data-science/glm.rst`
> * `h2o-docs/src/product/data-science/algo-params/family.rst`

content to be added prepared here: https://0xdata.atlassian.net/browse/PUBDEV-7648 @angela0xdata could you please review whether it can be added in this form?

PR #4699, opened 2020-06-09
PUBDEV-7610: Update front page for h2o-3 docs

angela0xdata:
This index.html file will be published to S3 and then included in the zip with all of the docs-website docs. Included only those docs that are also included in the docs-website. (Note that tutorials are not included.)

The files to review in this branch are:

- h2o-docs/src/front/index.html
- make-dist.sh (re-added back this index.html file in the build)

The rest of the files are assets related to the docs page stylesheet/javascripts/fonts/etc


h2o-docs/src/front/index.html
```diff
@@ -1,970 +1,207 @@
...
+          <h4 class="underline-yellow wow wefadeInUp">H2O-3</h4>
+          <h6 class="wow wefadeInUp">The H2O open source platform works with R, Python, an Scala on Hadoop/Yarn, Spark, or your laptop</h6>
+        </div>
+
+        <div class="col-two right">
+
+          <div class="bottom-wrapper wow wefadeInUp" style="padding-top: 30px;">
+              <h6>End User Documentation</h6>
+              <div class="col-one" style="padding-top: 15px;">
+                <p><a href="/h2o/latest-stable/h2o-docs/index.html">H2O User Guide</a></p>
```
bilcus, line 109: I don't think the relative links are correct. I would imagine you have to drop the `/h2o/latest-stable` prefix on all of them since the index.html would in this kind of directory structure:
https://s3.console.aws.amazon.com/s3/buckets/h2o-release/h2o/rel-yule/1/docs-website/?region=us-east-1&tab=overview#

angela0xdata: Oh, you're correct. I updated these so that they're relative to the docs-website folder.

angela0xdata:
@bilcus 
> Is there any way I can generate an example release?

Not that I know of. I guess maybe creating a test build after this is merged?

angela0xdata:
@bilcus I just tested this by pasting the index.html file and the assets folder into the unzipped file that Michal created.
(s3cmd get s3://test.0xdata.com/test-release/h2o/rel-zahradnik-rc5-mk1/4/h2o-3.30.0.4-cdh6.3.zip)

Everything looks good now that I've removed the beginning "/" for the doc paths.

bilcus:
Ok, I guess let's merge and see how the nightly turns out.

angela0xdata:
Should I wait for Michal to merge his change to make-dist.sh first, and then pull in his changes?
https://github.com/h2oai/h2o-3/pull/4701/files#diff-c73cd4ea21296507f9e621c0a41c8d65

bilcus:
Sure, let's do that.

I almost forgot, could you please add a version somewhere in the index.html (if it is not there already)

PR #4703, opened 2020-06-10
PUBDEV-7622 concordance

satai:


michalkurka:
@satai the test doesn't work because you are not populating training metrics; the metrics are only seen as "regular" metrics and thus not returned as training/valid/cv metrics.

I modified the code this way:
```
    perf <- h2o.performance(hexModel, data=heart.hex)
    print(perf@metrics$concordance)
    ## [1] 0.1234543
    expect_equal(0.1234543, perf@metrics$concordance, tolerance = 1e-3)
    print("YAY!")
```

this shows the value is propagated to the client

The question here is whether we do want to include concordance in training metrics - I would say no at this point - especially since the calculation is slow.

grigory93:
> The question here is whether we do want to include concordance in training metrics - I would say no at this point - especially since the calculation is slow.

we should give user an option to compute concordance in training with the model - apparently not by default.

michalkurka:
@grigory93 thanks for the input, do you agree it could be added later in order to speed-up the delivery? is exposing in just `h2o.performance` good enough to start with?

satai:
I am not sure about h2o API compatibility customs... but it' generally much safer to add stuff later then remove or fix stuff later. My two cents.

grigory93:
> @grigory93 thanks for the input, do you agree it could be added later in order to speed-up the delivery? is exposing in just `h2o.performance` good enough to start with?

Yes, agree.

h2o-genmodel/src/main/java/hex/genmodel/attributes/metrics/MojoModelMetricsRegressionCoxPH.java
```diff
@@ -0,0 +1,8 @@
+package hex.genmodel.attributes.metrics;
+
+import hex.genmodel.attributes.SerializedName;
+
+public class MojoModelMetricsRegressionCoxPH extends MojoModelMetricsRegression {
+  @SerializedName("concordance")
+  public long concordance;
```
michalkurka, line 7: Why is this long?

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
```diff
@@ -0,0 +1,231 @@
...
+              strataVecs.stream().map(it -> it.new Reader()).collect(toList()),
+              estimateVec.new Reader(), length);
+
+      return stats;
+    }
+
+    private static Stats concordanceStats(Vec.Reader startVec, Vec.Reader stopVec, Vec.Reader eventVec, List<Vec.Reader> strataVecs, Vec.Reader estimateVec, long length) {
+      Stream<Integer> allIndexes = IntStream.range(0, (int) length).boxed();
+
+      Map<List<Long>, List<Integer>> byStrata = allIndexes.collect(groupingBy(
```
michalkurka, line 169: My understanding this code is not final and performance should be improved - in that case, please link a JIRA for the performance improvement.

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
```diff
@@ -0,0 +1,231 @@
...
+      final Stats stats = concordanceStats(null == startVec ? null : startVec.new Reader(), 
+              stopVec.new Reader(), eventVec.new Reader(),
+              strataVecs.stream().map(it -> it.new Reader()).collect(toList()),
+              estimateVec.new Reader(), length);
+
+      return stats;
+    }
+
+    private static Stats concordanceStats(Vec.Reader startVec, Vec.Reader stopVec, Vec.Reader eventVec, List<Vec.Reader> strataVecs, Vec.Reader estimateVec, long length) {
+      Stream<Integer> allIndexes = IntStream.range(0, (int) length).boxed();
```
michalkurka, line 167: At least an assert is appropriate when casting long to integer.

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
```diff
@@ -0,0 +1,231 @@
...
+
+          final double t1 = stopVec.at(i) - ((startVec != null) ? startVec.at(i) : 0d);
+          final double t2 = stopVec.at(j) - ((startVec != null) ? startVec.at(j) : 0d);
+
+          final long event1 = eventVec.at8(i);
+          final long event2 = eventVec.at8(j);
+          final double estimate1 = estimateVec.at(i);
+          final double estimate2 = estimateVec.at(j);
+
+          strataVecs.stream().forEach(
```
michalkurka, line 198: creating streams in a double for loop seems not necessary since this is only evaluated when assertions are enabled

h2o-core/src/main/java/water/api/schemas3/ModelMetricsRegressionCoxPHV3.java
```diff
@@ -0,0 +1,30 @@
+package water.api.schemas3;
+
+import hex.ModelMetricsRegressionCoxPH;
+import water.api.API;
+import water.api.API.Direction;
+
+public class ModelMetricsRegressionCoxPHV3 extends ModelMetricsRegressionV3<ModelMetricsRegressionCoxPH, ModelMetricsRegressionCoxPHV3> {
+  @API(help="concordance",direction = Direction.OUTPUT)
```
michalkurka, line 8: Better description should be provided? At least for stuff like "tiedY"

h2o-genmodel/src/main/java/hex/genmodel/attributes/metrics/MojoModelMetricsRegressionCoxPH.java
michalkurka: Is this class used?

michalkurka:
@satai there are no Python changes, is that intended?

Otherwise, the implementation is okay for a POC, performance will need to get optimized.

satai:
Python stuff was forgotten, It's pushed now.

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
satai: done

h2o-core/src/main/java/water/api/schemas3/ModelMetricsRegressionCoxPHV3.java
satai: done

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
satai: done

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
satai: https://0xdata.atlassian.net/browse/PUBDEV-7736

h2o-genmodel/src/main/java/hex/genmodel/attributes/metrics/MojoModelMetricsRegressionCoxPH.java
satai: in `determineModelMetricsType()` in `ModelAttributes`

PR #4705, opened 2020-06-11
[PUBDEV-7615] Fix error Cannot find Builder for algo url name quantiles

jakubhava:


h2o-core/src/main/java/hex/Model.java
```diff
@@ -60,7 +61,7 @@
     final Key[] modelKeys = KeySnapshot.globalSnapshot().filter(new KeySnapshot.KVFilter() {
       @Override
       public boolean filter(KeySnapshot.KeyInfo k) {
-        return Value.isSubclassOf(k._type, Model.class);
+        return Value.isSubclassOf(k._type, Model.class) && !Value.isSubclassOf(k._type, QuantileModel.class);
```
Pscheidl, line 64: Or we could hide models that are supposed to be invisible to our users under user-hidden keys (`Key.makeUserHidden`) and only filter out those here.

It'd require a slight modification to ModelBuilder.

jakubhava: Yup, that is a good idea. I started with the least intrusive change as the `list` method would fail anyway when the quantiles model exists and I wasn't sure if any clients are accessing the quantiles model via some other paths.

Pscheidl: I'm ok with it.

jakubhava:
@michalkurka Thank you for the PR with the proper fix!

PR #4708, opened 2020-06-11
PUBDEV-7616: Propagate warnings from ModelBuilder in Python

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7616

Implements similar functionality as is present in R. When model builder emits warnings in R we show them, in Python we ignored them until this PR.

This is motivated by https://github.com/h2oai/h2o-3/pull/4559.

### DIFF
--- a/h2o-py/h2o/estimators/estimator_base.py
+++ b/h2o-py/h2o/estimators/estimator_base.py
@@ -194,6 +194,11 @@ class H2OEstimator(ModelBase):
         model_builder_json = h2o.api("POST /%d/ModelBuilders/%s" % (rest_ver, self.algo), data=parms)
         job = H2OJob(model_builder_json, job_type=(self.algo + " Model Build"))
 
+        if model_builder_json["messages"] is not None:
+            for mesg in model_builder_json["messages"]:
+                if mesg["message_type"] == "WARN":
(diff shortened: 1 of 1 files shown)


h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -194,6 +194,10 @@ def _train(self, parms, verbose=False):
         model_builder_json = h2o.api("POST /%d/ModelBuilders/%s" % (rest_ver, self.algo), data=parms)
         job = H2OJob(model_builder_json, job_type=(self.algo + " Model Build"))
 
+        for mesg in model_builder_json["messages"]:
+            if mesg["message_type"] == "WARN":
+                warnings.warn(mesg["message"], RuntimeWarning)
+
```
sebhrusen, line 200: according to `ModelBuilderSchema`, looks like there is a possibility for `messages` to be null/None.
Maybe it doesn't occur in practice though... doesn't the R client do any check?

tomasfryda: Thank you, that's a good point. R checks length... I ran locally all python stacked ensemble tests and it didn't fail on anything - possibly due to always having some warning... But better safe than sorry...

PR #4710, opened 2020-06-11
Python Doc update: Adding persist module to the docs

angela0xdata:
- Added h2o.persist.persist to the data.rst file so that the `set_s3_credentials` function shows up in the docs
- In the persist.py file, used "param" notation so that the parameters list shows up correctly in the Python docs.


angela0xdata:
Screenshot of updated docs:

Screen Shot 2020-06-11 at 12 42 24 PM

PR #4713, opened 2020-06-12
PUBDEV-7319: Integrate TE to H2O models as another encoder

sebhrusen:
As a pre-condition to AutoML integration, to be able to productionize AutoML models when TE is enabled, this exposes TE models as another encoder in H2O Models.
Note that this exposition is only for internal (Java) usage: this doesn't offer the possibility for users to specify the TE encoder for a given model. Currently, the user should still produce 2 models, one for TE, and one for the "main" model.

This is an up-to-date version of https://github.com/h2oai/h2o-3/pull/4362
No major changes.

### DIFF
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -257,6 +257,9 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
       final boolean _needResponse;
       boolean needsResponse() { return _needResponse; }
     }
+
+    public Key<Model> _te_model;
+
     public long _seed = -1;
     public long getOrMakeRealSeed(){
       while (_seed==-1) {
@@ -715,7 +718,10 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
     public Key _cross_validation_predictions[];
     public Key<Frame> _cross_validation_holdout_predictions_frame_id;
     public Key<Frame> _cross_validation_fold_assignment_frame_id;
-    
+
+    // Key to a TE model
+    public Key<Model> _te_model_key;
+
     // Model-specific start/end/run times
     // Each individual model's start/end/run time is reported here, not the total time to build N+1 cross-validation models, or all grid models
     public long _start_time;
@@ -760,6 +766,7 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
       _distribution = b._distribution;
       _priorClassDist = b._priorClassDist;
       _reproducibility_information_table = createReproducibilityInformationTable(b);
+      _te_model_key = b.internal_getTEModelKey();
       assert(_job==null);  // only set after job completion
       _defaultThreshold = -1;
     }
@@ -810,7 +817,7 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
      * and checksums for each frame used on the input of the algorithm
      */
     public TwoDimTable[] _reproducibility_information_table;
-    
+
 
     /**
      * User-facing model scoring history - 2D table with modeling accuracy as a function of time/trees/epochs/iterations, etc.
@@ -994,7 +1001,7 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
 
       return table;
     }
-    
+
     public int getInformationTableNumRows() {
       return 2; // 1 row per each input frame (training frame, validation frame)
     }
@@ -1559,7 +1566,8 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
     return score(fr, destination_key, j, computeMetrics, CFuncRef.NOP);
   }
   public Frame score(Frame fr, String destination_key, Job j, boolean computeMetrics, CFuncRef customMetricFunc) throws IllegalArgumentException {
-    Frame adaptFr = new Frame(fr);
+    Frame newFr = new Frame(fr);
+    Frame adaptFr = applyTEIfModelAvailable(newFr);
     computeMetrics = computeMetrics && (!isSupervised() || (adaptFr.vec(_output.responseName()) != null && !adaptFr.vec(_output.responseName()).isBad()));
     String[] msg = adaptTestForTrain(adaptFr,true, computeMetrics);   // Adapt
     // clean up the previous score warning messages
@@ -1603,6 +1611,18 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
     }
     Frame.deleteTempFrameAndItsNonSharedVecs(adaptFr, fr);
     return output;
+  }
+
+  private Frame applyTEIfModelAvailable(Frame newFr) {
+    if(_output._te_model_key != null ) {
+      Frame encodedFrame = FrameUtils.internal_applyTargetEncoder(DKV.getGet(_output._te_model_key), newFr, false);
+      Arrays.stream(encodedFrame.keys()).forEach(key -> {
+        _toDelete.put(key, "track encoded by TE frame");
+      });
+      return encodedFrame;
+    } else {
+      return newFr;
+    }
   }
 
   /**
@@ -1948,6 +1968,7 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
       deleteCrossValidationFoldAssignment();
       deleteCrossValidationPreds();
       deleteCrossValidationModels();
+      internal_deleteTEModel();
     }
     cleanUp(_toDelete);
     return super.remove_impl(fs, cascade);
@@ -2608,6 +2629,17 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
       Log.info("Cleaning up CV Models for " + _key);
       int count = deleteAll(_output._cross_validation_models);
       Log.info(count+" CV models were removed");
+    }
+  }
+
+  /**
+   * delete from the output TE model stored in DKV.
+   */
+  public void internal_deleteTEModel() {
+    if (_output._te_model_key != null) {
+      Log.info("Cleaning up TE Model for " + _key);
+      int count = deleteAll(new Key[]{_output._te_model_key});
+      Log.info(count+" TE model was removed");
     }
   }
 
--- a/h2o-core/src/main/java/hex/ModelBuilder.java
+++ b/h2o-core/src/main/java/hex/ModelBuilder.java
@@ -128,6 +128,23 @@ abstract public class ModelBuilder<M extends Model<M,P,O>, P extends Model.Param
   /** gbm -> "hex.schemas." ; custAlgo -> "org.myOrg.schemas." */
   public static String schemaDirectory(String urlName) { return SCHEMAS[ArrayUtils.find(ALGOBASES,urlName)]; }
 
+  /**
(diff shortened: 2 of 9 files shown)


h2o-core/src/main/java/hex/Model.java
```diff
@@ -257,6 +257,9 @@ public void setTrain(Key<Frame> train) {
       final boolean _needResponse;
       boolean needsResponse() { return _needResponse; }
     }
+
+    public Key<Model> _te_model;
```
Pscheidl, line 261: In the middle of nowhere. Let's move it up a little bit or group it with other TE methods.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -715,7 +718,10 @@ public void setNames(String[] names, String[] columntypes) {
     public Key _cross_validation_predictions[];
     public Key<Frame> _cross_validation_holdout_predictions_frame_id;
     public Key<Frame> _cross_validation_fold_assignment_frame_id;
-    
+
+    // Key to a TE model
+    public Key<Model> _te_model_key;
```
Pscheidl, line 723: In the middle of nowhere. Let's move it up a little bit or group it with other TE methods and attributes.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -128,6 +128,23 @@ protected ModelBuilder(P parms, boolean startup_once, String externalSchemaDirec
...
+   */
+  public void addTEModelKey(Key<Model> model) {
+    _parms._te_model = model;
+  }
+
+  /**
+   *
+   * @return TE model's key if one was assigned to a model builder, null otherwise
+   */
+  public Key<Model> internal_getTEModelKey() {
```
Pscheidl, line 144: If something is internal, let's not have it public and use packages to moderate the access. Also, Java is not snake_case.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -128,6 +128,23 @@ protected ModelBuilder(P parms, boolean startup_once, String externalSchemaDirec
   /** gbm -> "hex.schemas." ; custAlgo -> "org.myOrg.schemas." */
   public static String schemaDirectory(String urlName) { return SCHEMAS[ArrayUtils.find(ALGOBASES,urlName)]; }
 
+  /**
+   * Helper method that enables preprocessing with TE model. It's user's responsibility to make sure TE is trained in an appropriate way to avoid data leakage.
+   * E.g. TE without KFold strategy being applied for cv main model leads to a data leakage.
+   * @param model key to the trained TE model
+   */
+  public void addTEModelKey(Key<Model> model) {
```
Pscheidl, line 136: This is a setter.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -2611,6 +2632,17 @@ public void deleteCrossValidationModels() {
     }
   }
 
+  /**
+   * delete from the output TE model stored in DKV.
+   */
+  public void internal_deleteTEModel() {
```
Pscheidl, line 2638: Snake case prefix again - not Java.

h2o-core/src/main/java/water/util/FrameUtils.java
```diff
@@ -111,6 +111,37 @@ public static Frame categoricalEncoder(Frame dataset, String[] skipCols, Model.P
...
   }
 
+  /**
+   * Helper method to apply TargetEncoderModel
+   * @param teModel trained TargetEncoderModel
+   * @param inputFrame frame to apply target encoding to
+   * @param scopeTrack `true` if Scope is tracking keys
+   * @return a Frame with encodings
+   */
+  public static Frame internal_applyTargetEncoder(Model teModel, Frame inputFrame, boolean scopeTrack) {
```
Pscheidl, line 121: This belongs somewhere close to Target Encoder package, not to generate frame utils. This way, anything could be utils.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -128,6 +128,23 @@ protected ModelBuilder(P parms, boolean startup_once, String externalSchemaDirec
   /** gbm -> "hex.schemas." ; custAlgo -> "org.myOrg.schemas." */
   public static String schemaDirectory(String urlName) { return SCHEMAS[ArrayUtils.find(ALGOBASES,urlName)]; }
 
+  /**
+   * Helper method that enables preprocessing with TE model. It's user's responsibility to make sure TE is trained in an appropriate way to avoid data leakage.
```
Pscheidl, line 132: It's definitely true right now it's up to the user to provide a correct TE model, but we can do a lot of validations - we might reject a target encoder provided by a user if there are obvious mismatches, e.g. target variable type.

Such validations can definitely be done.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -2611,6 +2632,17 @@ public void deleteCrossValidationModels() {
...
+    if (_output._te_model_key != null) {
+      Log.info("Cleaning up TE Model for " + _key);
```
Pscheidl, line 2640: Let's use our new LOG4J in new code.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1605,6 +1613,18 @@ public Frame score(Frame fr, String destination_key, Job j, boolean computeMetri
     return output;
   }
 
+  private Frame applyTEIfModelAvailable(Frame newFr) {
+    if(_output._te_model_key != null ) {
+      Frame encodedFrame = FrameUtils.internal_applyTargetEncoder(DKV.getGet(_output._te_model_key), newFr, false);
+      Arrays.stream(encodedFrame.keys()).forEach(key -> {
```
Pscheidl, line 1619: Converting an Array to Stream just to iterate over it introduces unnecessary overhead. A simple `fori` would do.

h2o-core/src/main/java/water/util/FrameUtils.java
```diff
@@ -111,6 +111,37 @@ public static Frame categoricalEncoder(Frame dataset, String[] skipCols, Model.P
...
+    Frame transformed = teModel.score(inputFrame);
+    if (scopeTrack) Scope.untrack(transformed.keys());
+
+    // following manipulations with columns is a responsibility of a future pipeline concept. Support for input/output columns' names for the models would be helpful as well.
+    Frame train = teModel._parms.train();
+    List<String> catColumnNames = new ArrayList<>();
+    for(String name : train.names() ) {
+      if(train.vec(name).isCategorical()) catColumnNames.add(name);
+    }
+    String[] predictorColumns = ArrayUtils.difference(catColumnNames.toArray(new String[0]), teModel._parms._ignored_columns);
```
Pscheidl, line 131: We know the length of catColumnNames array beforehand. Passing properly-sized array is easy. There is no need to let the code underneath do it for us again and waste one allocation.

h2o-core/src/main/java/water/util/FrameUtils.java
```diff
@@ -111,6 +111,37 @@ public static Frame categoricalEncoder(Frame dataset, String[] skipCols, Model.P
...
+    final String[] encodedColumns = ArrayUtils.remove(ArrayUtils.remove(predictorColumns, teModel._parms._fold_column), teModel._parms._response_column);
+
+    // We need to rearrange columns ( as some logic relies on response being a last column etc)
+    final String TE_ENCODED_COLUMN_POSTFIX = "_te";
+    Arrays.stream(encodedColumns).forEach(ec -> {
```
Pscheidl, line 136: Abuse of streams - creating a Stream from and Array just to iterate over the array is worthless and makes the performance worse for no reason.

h2o-core/src/main/java/water/util/FrameUtils.java
```diff
@@ -111,6 +111,37 @@ public static Frame categoricalEncoder(Frame dataset, String[] skipCols, Model.P
...
+      transformed
+        .swap(transformed.find(ec), transformed.find(ec + TE_ENCODED_COLUMN_POSTFIX));
```
Pscheidl, line 138: Just a gentle warning, the `find` method is `O(n)`. I'd say the expected number of encoded column is not that large, yet the actual number of columns to search within could be.

h2o-core/src/main/java/water/util/FrameUtils.java
```diff
@@ -111,6 +111,37 @@ public static Frame categoricalEncoder(Frame dataset, String[] skipCols, Model.P
...
+   * @param inputFrame frame to apply target encoding to
+   * @param scopeTrack `true` if Scope is tracking keys
+   * @return a Frame with encodings
+   */
+  public static Frame internal_applyTargetEncoder(Model teModel, Frame inputFrame, boolean scopeTrack) {
+    Frame transformed = teModel.score(inputFrame);
+    if (scopeTrack) Scope.untrack(transformed.keys());
+
+    // following manipulations with columns is a responsibility of a future pipeline concept. Support for input/output columns' names for the models would be helpful as well.
+    Frame train = teModel._parms.train();
```
Pscheidl, line 126: Finals are used at some places, at other places they're not ....

h2o-extensions/target-encoder/src/test/java/ai/h2o/targetencoding/ModelBuilderWithTETest.java
```diff
@@ -0,0 +1,288 @@
...
+    builder._parms._seed = builderSeed;
+
+    builder._parms._train = fr._key;
+    builder._parms._response_column = responseColumnName;
+    builder._parms._keep_cross_validation_models = true;
+    builder._parms._keep_cross_validation_predictions = true;
+    return builder;
+  }
+
+  public static class ModelBuilderWithTENonParametrizedTest extends TestUtil {
```
Pscheidl, line 46: H2ORunner is available :wink:

h2o-extensions/target-encoder/src/test/java/water/H2OTestNodeStarter.java
```diff
@@ -0,0 +1,15 @@
...
+ * This class is intended to be ran during distributed
+ * testing from Idea.
+ */
+@Ignore("Support for tests, but no actual tests here")
+public class H2OTestNodeStarter extends H2OStarter {
+
+  public static void main(String[] args) {
+    start(args, System.getProperty("user.dir"));
+  }
+}
```
Pscheidl, line 15: Missing one empty line at the end.

h2o-core/src/main/java/water/util/FrameUtils.java
sebhrusen: Agree that this should go somewhere else.
Can't go to TE module though as it would create a dependency from core to targetencoder.
Let me think about some abstraction, even if temporary, that will allow to group all this logic to the TE module.

h2o-core/src/main/java/hex/Model.java
sebhrusen: agree, mainly don't like `forEach` for another reason: it allows side-effects, and I recommend streams only when  purely functional... will replace those.

h2o-core/src/main/java/hex/Model.java
sebhrusen: god! we've got 2 loggers now?

h2o-core/src/main/java/hex/ModelBuilder.java
sebhrusen: verified that it was not needed for AutoML integration, so, visibility should definitely be changed.

PR #4714, opened 2020-06-15
[PUBDEV-7467] websocket endpoint support, minimal impl for Steam

honzasterba:
since the websocket api is different in jetty8 and 9 we have to use a wrapper class and custom WebsocketServlet interface


h2o-extensions/steam/src/main/java/hex/steam/SteamMessageSender.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static hex.steam.SteamMessenger.ID;
+import static hex.steam.SteamMessenger.TYPE;
+
+public interface SteamMessageSender {
+
+    Logger LOG = Logger.getLogger(SteamHelloMessenger.class);
```
michalkurka, line 14: the logger class seems incorrect

honzasterba: thanks, fixed

h2o-jetty-9/src/main/java/water/webserver/jetty9/Jetty9WebsocketServlet.java
```diff
@@ -0,0 +1,76 @@
...
+    static class Jetty9WebsocketConnection implements WebsocketConnection {
+
+        private final Session sess;
+
+        Jetty9WebsocketConnection(Session sess) {
+            this.sess = sess;
+        }
+
+        @Override
+        public void sendMessage(String message) throws IOException {
```
michalkurka, line 31: Would it make sense to primarily use the asynchronous API that enables you to receive a response in a straightforward way?

https://www.eclipse.org/jetty/javadoc/9.4.30.v20200611/org/eclipse/jetty/websocket/api/RemoteEndpoint.html#sendStringByFuture(java.lang.String)

michalkurka: Do we need the ACK method?

honzasterba: ad async API: there is no request-response in websockets, but I agree we could make this API easier to work with in the case we require every message to be confirmed with a ACK response, which I am not sure I want to do

the ACK helper is steam specific, this part of the code is completely independent, for now the hello-ack messages are an easy way to test the comunication and I think something like this could be useful in the future, but I dont think I will want to keep ACK the way it is now

Pscheidl: I like how this is enforced with the interface introduced. And it's steam specific, it's part of the contract. This is completely fine to me. 

Req/resp has to be implemented on top of WS exactly like that. There are implementations and standarda for that, e.g. JAX-RS and RestEasy/Jersey. Those are however quite heavy.

Pscheidl: Not that much heavy, but I like the minimalist approach in H2O more.

Pscheidl:
Only one comment:  Let's have JavaDoc on top of all interfaces.

I find the interface-driven way of enforcing ACKs for Steam only spot on. It should be documented, the pattern might not be obvious for everyone.

PR #4716, opened 2020-06-15
Fix pyunit_model_id_confict: correct the test assumptions

michalkurka:


Pscheidl:
Tests were green, merging. Let's get the fix in ASAP.

PR #4723, opened 2020-06-16
PUBDEV-7589: Doc updates for AutoML Random Grid Search parameters section

angela0xdata:
- Removed `min_sum_hessian_in_leaf` from XGBoost table because it is no longer a supported parameter
- Removed GLM table and replaced with info describing how GLM behaves in AutoML
- Fixed some typos

### DIFF
--- a/h2o-docs/src/product/automl.rst
+++ b/h2o-docs/src/product/automl.rst
@@ -411,11 +411,11 @@ FAQ
 
 -  **Which models are trained in the AutoML process?**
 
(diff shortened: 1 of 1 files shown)


ledell:
@angela0xdata The GLM "grid search" is not a normal grid search in the sense that it uses the H2O Grid interface... however, we do in fact do a grid search, it's just returning a single model (with best alpha) vs all the models (each alpha).  GLM supports it's own internal grid search and that's the one we use here.  So, I think it makes sense to keep the GLM search listed in the table, but you can edit the note to reflect this more nuanced description of what's actually happening.

angela0xdata:
@ledell thanks for the clarification.

ledell:
Thanks!

PR #4726, opened 2020-06-17
Make it clear we don't actually use a (wrong) threshold to make preds

michalkurka:


h2o-algos/src/main/java/hex/tree/Score.java
```diff
@@ -95,7 +94,14 @@ else if (_sii != null)
...
           tmp[i] = chks[i].atd(row);
 
-      if( nclass > 1 ) cdists[0] = GenModel.getPrediction(cdists, m._output._priorClassDist, tmp, m.defaultThreshold()); // Fill in prediction
+      if (nclass > 2) { // Fill in prediction for multinomial
+        cdists[0] = GenModel.getPredictionMultinomial(cdists, m._output._priorClassDist, tmp);
+      } else if (nclass == 2) {
+        // for binomial the predicted class is not needed
+        // and it even cannot be returned because the threshold is calculated based on model metrics that are not known yet
+        // (we are just building the metrics)
+        cdists[0] = -1;
```
maurever, line 103: Is it necessary to assign -1? There could be something like:

```suggestion
      if( nclass > 2 ) cdists[0] = GenModel.getPrediction(cdists, m._output._priorClassDist, tmp, m.defaultThreshold());
```

And the comment of course...

h2o-genmodel/src/main/java/hex/genmodel/GenModel.java
```diff
@@ -332,8 +332,17 @@ public boolean calibrateClassProbabilities(double preds[]) {
    */
   public static int getPrediction(double[] preds, double[] priorClassDist, double[] data, double threshold) {
     if (preds.length == 3) {
-      return (preds[2] >= threshold) ? 1 : 0; //no tie-breaking
+      return getPredictionBinomial(preds, threshold);
```
maurever, line 335: These changes are definitely more understandable. :+1:

h2o-algos/src/main/java/hex/tree/Score.java
michalkurka: It is not necessary but it is better that way in my opinion, assigning an invalid value can help discover issues if the assumptions change = ModelMetricsBinomial starts using the class.

I want to make sure there is a clearly invalid value in all cases, even for an empty model (when the class actually is assigned /using the wrong threshold again/ by score0) method. Invalidating the value helps to make it clear it can never get used.

WDYT?

maurever: If you are sure, the `cdist` is never used then for the binomial case, I am ok with this change. The tests are green.

michalkurka: If it was actually used it would be wrong because to assign the class you need a threshold and you don't have the correct threshold until you finish scoring all the observations.

Fortunately, it is not used and all is good.

PR #4730, opened 2020-06-18
PUBDEV-7404 - Kolmogorov-Smirnov metric

Pscheidl:
[PUBDEV-7404](https://0xdata.atlassian.net/browse/PUBDEV-7404)

Requirements taken from the JIRA:

1. have a function that calculates KS :heavy_check_mark: 
2. allow the user the ability to specify the granularity of the bin (Set via new `gainslift_bins` option ) **R**: `model <- h2o.gbm(x = c("Origin", "Distance"), y = "IsDepDelayed", training_frame = data, ntrees = 1, gainslift_bins = 500)`. **Python**: `model = H2OGradientBoostingEstimator(ntrees=1, gainslift_bins=20)`
3. returns not just the maximum KS but also the distribution of positive/negative counts by prediction bin _(can be seen in the Gains/Lift table ?)_

R client :heavy_check_mark: 
Python client: :heavy_check_mark: 
Flow: :heavy_check_mark: (works naturally thanks to TwoDimTable).

Screenshot from 2020-06-18 14-59-57

## Basic usage

```python
    airlines = h2o.import_file(path=pyunit_utils.locate("smalldata/testng/airlines_train.csv"))
    model = H2OGradientBoostingEstimator(ntrees=1, gainslift_bins=20)
    model.train(x=["Origin", "Distance"], y="IsDepDelayed", training_frame=airlines)

    print(model.gains_lift())
    print(model.kolmogorov_smirnov())
```

<details>
<summary>OUTPUT</summary>

```
Gains/Lift Table: Avg response rate: 54.69 %, avg score: 54.68 %
group    cumulative_data_fraction    lower_threshold    lift      cumulative_lift    response_rate    score     cumulative_response_rate    cumulative_score    capture_rate    cumulative_capture_rate    gain      cumulative_gain    kolmogorov_smirnov
-------  --------------------------  -----------------  --------  -----------------  ---------------  --------  --------------------------  ------------------  --------------  -------------------------  --------  -----------------  --------------------
1        0.0382458                   0.576503           1.56234   1.56234            0.85439          0.5774    0.85439                     0.5774              0.0597529       0.0597529                  56.234    56.234             0.047463
2        0.0409893                   0.575572           1.52839   1.56007            0.835821         0.575572  0.853147                    0.577277            0.00419319      0.0639461                  52.8385   56.0067            0.0506622
3        0.0917243                   0.568189           1.39175   1.46696            0.761098         0.568189  0.802232                    0.57225             0.0706103       0.134556                   39.1746   46.6965            0.0945238
4        0.121248                    0.566824           1.36701   1.44263            0.747573         0.566849  0.788923                    0.570935            0.0403594       0.174916                   36.7014   44.2627            0.118436
5        0.206707                    0.558832           1.21965   1.35044            0.666986         0.558848  0.73851                     0.565938            0.104231        0.279146                   21.9653   35.0443            0.159862
6        0.423652                    0.556035           1.16799   1.25701            0.638732         0.556035  0.687415                    0.560867            0.253388        0.532535                   16.7987   25.701             0.240288
7        0.503173                    0.549092           1.0772    1.22859            0.589083         0.551083  0.671875                    0.559321            0.0856608       0.618195                   7.72      22.8593            0.253836
8        0.702264                    0.542676           0.927089  1.14312            0.506993         0.542875  0.625131                    0.554658            0.184575        0.80277                    -7.29108  14.3117            0.221802
9        0.850006                    0.534575           0.791145  1.08194            0.43265          0.535421  0.591675                    0.551315            0.116885        0.919656                   -20.8855  8.19399            0.153706
10       0.95164                     0.522945           0.587187  1.0291             0.321112         0.524208  0.56278                     0.54842             0.059678        0.979334                   -41.2813  2.91009            0.0611156
11       1                           0.491833           0.427345  1                  0.2337           0.515403  0.546865                    0.546823            0.0206664       1                          -57.2655  0                  0

0.2538361336693768
```

^^^ The last row with a number is the K-S metric printed to console.
</details>

### DIFF
--- a/h2o-core/src/main/java/hex/GainsLift.java
+++ b/h2o-core/src/main/java/hex/GainsLift.java
@@ -147,10 +147,10 @@ public class GainsLift extends Iced {
             "Gains/Lift Table",
             "Avg response rate: " + PrettyPrint.formatPct(avg_response_rate) + ", avg score: " + PrettyPrint.formatPct(avg_score),
             new String[events.length],
-            new String[]{"Group", "Cumulative Data Fraction", "Lower Threshold", "Lift", "Cumulative Lift", "Response Rate", "Score", "Cumulative Response Rate", "Cumulative Score", "Capture Rate", "Cumulative Capture Rate", "Gain", "Cumulative Gain"},
-            new String[]{"int", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double"},
-            new String[]{"%d", "%.8f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f"},
-            "");
+            new String[]{"Group", "Cumulative Data Fraction", "Lower Threshold", "Lift", "Cumulative Lift", "Response Rate", "Score", "Cumulative Response Rate", "Cumulative Score", "Capture Rate", "Cumulative Capture Rate", "Gain", "Cumulative Gain", "Kolmogorov Smirnov"},
+            new String[]{"int", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double"},
+            new String[]{"%d", "%.8f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f","%5f"},
+            null);
     long sum_e_i = 0;
     long sum_n_i = 0;
     double sum_s_i = 0;
@@ -167,6 +167,8 @@ public class GainsLift extends Iced {
       sum_s_i += n_i * s_i;
       double lift=p_i/P; //can be NaN if P==0
       double sum_lift=(double)sum_e_i/sum_n_i/P; //can be NaN if P==0
+      double cum_event = sum_e_i/(double)E;
+      double cum_non_event = (sum_n_i - sum_e_i) / (double) (N - E);
       table.set(i,0,i+1); //group
       table.set(i,1,(double)sum_n_i/N); //cumulative_data_fraction
       table.set(i,2,_quantiles[i]); //lower_threshold
@@ -180,6 +182,7 @@ public class GainsLift extends Iced {
       table.set(i,10,(double)sum_e_i/E); //cumulative_capture_rate
       table.set(i,11,100*(lift-1)); //gain
       table.set(i,12,100*(sum_lift-1)); //cumulative gain
+      table.set(i,13,cum_event - cum_non_event); //Kolmogorov-Smirnov metric
       if (i== events.length-1) {
         assert(sum_n_i == N) : "Cumulative data fraction must be 1.0, but is " + (double)sum_n_i/N;
         assert(sum_e_i == E) : "Cumulative capture rate must be 1.0, but is " + (double)sum_e_i/E;
--- a/h2o-core/src/main/java/hex/GainsLift.java
+++ b/h2o-core/src/main/java/hex/GainsLift.java
@@ -167,8 +167,10 @@ public class GainsLift extends Iced {
       sum_s_i += n_i * s_i;
       double lift=p_i/P; //can be NaN if P==0
       double sum_lift=(double)sum_e_i/sum_n_i/P; //can be NaN if P==0
-      double cum_event = sum_e_i/(double)E;
-      double cum_non_event = (sum_n_i - sum_e_i) / (double) (N - E);
+      final double cum_event = sum_e_i / (double)E;
+      final double total_non_event = (double) (N - E);
+      // If response rate is 1, there are non non-events and the cumulative count will always be zero
+      final double cum_non_event = total_non_event == 0 ? 0 : (sum_n_i - sum_e_i) / total_non_event;
       table.set(i,0,i+1); //group
       table.set(i,1,(double)sum_n_i/N); //cumulative_data_fraction
       table.set(i,2,_quantiles[i]); //lower_threshold
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -358,7 +358,7 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
     /**
      * Bins for Gains/Lift table, if applicable. Ignored if G/L are not calculated.
      */
-    public int _gainslift_bins;
+    public int _gainslift_bins = -1;
 
     // Public no-arg constructor for reflective creation
     public Parameters() { _ignore_const_cols = defaultDropConsCols(); }
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -355,6 +355,11 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
      */
     public String _export_checkpoints_dir;
 
+    /**
+     * Bins for Gains/Lift table, if applicable. Ignored if G/L are not calculated.
+     */
+    public int _gainslift_bins;
+
     // Public no-arg constructor for reflective creation
     public Parameters() { _ignore_const_cols = defaultDropConsCols(); }
 
--- a/h2o-core/src/main/java/hex/ModelMetricsBinomial.java
+++ b/h2o-core/src/main/java/hex/ModelMetricsBinomial.java
@@ -270,13 +270,13 @@ public class ModelMetricsBinomial extends ModelMetricsSupervised {
 
     private GainsLift calculateGainsLift(Model m, Frame preds, Vec resp, Vec weights) {
       GainsLift gl = null;
-        gl = new GainsLift(preds.lastVec(), resp, weights);
-        if (m != null && m._parms._gainslift_bins != 0) {
-          if (m._parms._gainslift_bins < 1)
-            throw new IllegalArgumentException("Number of G/L bins must be greater than zero.");
-          gl._groups = m._parms._gainslift_bins;
-        }
-        gl.exec(m != null ? m._output._job : null);
+      gl = new GainsLift(preds.lastVec(), resp, weights);
+      if (m != null && m._parms._gainslift_bins != 0) {
+        if (m._parms._gainslift_bins < 1)
+          throw new IllegalArgumentException("Number of G/L bins must be greater than zero.");
+        gl._groups = m._parms._gainslift_bins;
+      }
+      gl.exec(m != null ? m._output._job : null);
       return gl;
     }
 
--- a/h2o-core/src/main/java/hex/ModelMetricsBinomial.java
+++ b/h2o-core/src/main/java/hex/ModelMetricsBinomial.java
@@ -282,7 +282,7 @@ public class ModelMetricsBinomial extends ModelMetricsSupervised {
      */
     private Optional<GainsLift> calculateGainsLift(Model m, Frame preds, Vec resp, Vec weights) {
       final GainsLift gl = new GainsLift(preds.lastVec(), resp, weights);
-      if (m._parms._gainslift_bins < -1) {
+      if (m != null && m._parms._gainslift_bins < -1) {
         throw new IllegalArgumentException("Number of G/L bins must be greater or equal than -1.");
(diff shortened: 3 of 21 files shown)


Pscheidl:
I'm not sure about the API design. There is a new function `kolmogorov_smirnov` which returns just the single scalar representing the metric and the details are available in G/L table. The reason to put the K-S metrics per-group there is well is simple - other "important" metrics like capture_rate are there as well.

WDYT @michalkurka @meganjkurka ?

h2o-core/src/main/java/hex/GainsLift.java
```diff
@@ -147,10 +147,10 @@ public TwoDimTable createTwoDimTable() {
...
             new String[events.length],
-            new String[]{"Group", "Cumulative Data Fraction", "Lower Threshold", "Lift", "Cumulative Lift", "Response Rate", "Score", "Cumulative Response Rate", "Cumulative Score", "Capture Rate", "Cumulative Capture Rate", "Gain", "Cumulative Gain"},
-            new String[]{"int", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double"},
-            new String[]{"%d", "%.8f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f"},
-            "");
+            new String[]{"Group", "Cumulative Data Fraction", "Lower Threshold", "Lift", "Cumulative Lift", "Response Rate", "Score", "Cumulative Response Rate", "Cumulative Score", "Capture Rate", "Cumulative Capture Rate", "Gain", "Cumulative Gain", "Kolmogorov Smirnov"},
+            new String[]{"int", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double", "double"},
+            new String[]{"%d", "%.8f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f", "%5f","%5f"},
+            null);
```
Pscheidl, line 153: This removes the unnecessary empty column with no header at the beginning of the table.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -355,6 +355,11 @@ public long getOrMakeRealSeed(){
      */
     public String _export_checkpoints_dir;
 
+    /**
+     * Bins for Gains/Lift table, if applicable. Ignored if G/L are not calculated.
+     */
+    public int _gainslift_bins;
```
Pscheidl, line 361: Exposing it as a parameter set before the actual model training has the best value/time invested ratio. G/L are calculated as a part of model driver's logic. The G/L calculations uses vecs not available after the training is done.

h2o-core/src/main/java/hex/ModelMetricsBinomial.java
```diff
@@ -272,8 +272,12 @@ private GainsLift calculateGainsLift(Model m, Frame preds, Vec resp, Vec weights
       GainsLift gl = null;
       try {
         gl = new GainsLift(preds.lastVec(), resp, weights);
+        if (m._parms._gainslift_bins !=0 ) {
+          if (m._parms._gainslift_bins < 1) throw new IllegalArgumentException("Number of G/L bins must be greater than zero.");
+          gl._groups = m._parms._gainslift_bins;
+        }
         gl.exec(m != null ? m._output._job : null);
-      } catch(Throwable t) { // TODO: Why do we need to catch Throwable here?
+      } catch (Throwable t) { // TODO: Why do we need to catch Throwable here?
```
honzasterba, line 280: yes, we really should not

h2o-core/src/main/java/hex/ModelMetricsBinomial.java
```diff
@@ -272,8 +272,12 @@ private GainsLift calculateGainsLift(Model m, Frame preds, Vec resp, Vec weights
       GainsLift gl = null;
       try {
         gl = new GainsLift(preds.lastVec(), resp, weights);
+        if (m._parms._gainslift_bins !=0 ) {
+          if (m._parms._gainslift_bins < 1) throw new IllegalArgumentException("Number of G/L bins must be greater than zero.");
+          gl._groups = m._parms._gainslift_bins;
```
honzasterba, line 277: interesting that this was there but was actually never used :)

h2o-core/src/main/java/hex/ModelMetricsBinomial.java
Pscheidl: I can make this right in this PR I guess.

Pscheidl: Completely removed.

h2o-core/src/main/java/hex/ModelMetricsBinomial.java
Pscheidl: Yes, truly amazing.

h2o-core/src/main/java/hex/ModelMetricsBinomial.java
honzasterba: 👍

h2o-core/src/main/java/hex/ModelMetricsBinomial.java
```diff
@@ -270,12 +270,12 @@ private ModelMetrics makeModelMetrics(Model m, Frame f, GainsLift gl) {
 
     private GainsLift calculateGainsLift(Model m, Frame preds, Vec resp, Vec weights) {
       GainsLift gl = null;
-      try {
         gl = new GainsLift(preds.lastVec(), resp, weights);
+        if (m._parms._gainslift_bins !=0 ) {
```
koniecsveta, line 274: There are multiple tests failing on NPE on this line (e.g. hex.tree.xgboost.XGBoostTest#WeatherBinary...) Also you can fix the condition formatting here if you want.

Pscheidl: Thank you. There definitely is a scenario where the model is null - some `make` methods of `ModelMetricsBinomial` allow it. Fixed by a null check.

h2o-core/src/main/java/hex/ModelMetricsBinomial.java
```diff
@@ -270,12 +270,13 @@ private ModelMetrics makeModelMetrics(Model m, Frame f, GainsLift gl) {
 
     private GainsLift calculateGainsLift(Model m, Frame preds, Vec resp, Vec weights) {
       GainsLift gl = null;
-      try {
         gl = new GainsLift(preds.lastVec(), resp, weights);
```
honzasterba, line 273: knit: fix indent

Pscheidl: Resolved. Thanks.

h2o-py/h2o/estimators/naive_bayes.py
```diff
@@ -788,3 +788,18 @@ def export_checkpoints_dir(self, export_checkpoints_dir):
         self._parms["export_checkpoints_dir"] = export_checkpoints_dir
 
 
+    @property
+    def gainslift_bins(self):
+        """
+        Gains/Lift table bins
+
+        Type: ``int``  (default: ``0``).
```
michalkurka, line 796: default is 0?

h2o-core/src/main/java/hex/GainsLift.java
```diff
@@ -167,6 +167,8 @@ public TwoDimTable createTwoDimTable() {
       sum_s_i += n_i * s_i;
       double lift=p_i/P; //can be NaN if P==0
       double sum_lift=(double)sum_e_i/sum_n_i/P; //can be NaN if P==0
+      double cum_event = sum_e_i/(double)E;
+      double cum_non_event = (sum_n_i - sum_e_i) / (double) (N - E);
```
michalkurka, line 171: avg_response_rate can be 1 and thus N can be equal to E, leading to a division by zero

h2o-core/src/main/java/hex/ModelMetricsBinomial.java
```diff
@@ -270,12 +270,13 @@ private ModelMetrics makeModelMetrics(Model m, Frame f, GainsLift gl) {
 
     private GainsLift calculateGainsLift(Model m, Frame preds, Vec resp, Vec weights) {
       GainsLift gl = null;
-      try {
-        gl = new GainsLift(preds.lastVec(), resp, weights);
-        gl.exec(m != null ? m._output._job : null);
-      } catch(Throwable t) { // TODO: Why do we need to catch Throwable here?
-        Log.debug("Calculating Gains-Lift failed", t);
+      gl = new GainsLift(preds.lastVec(), resp, weights);
+      if (m != null && m._parms._gainslift_bins != 0) {
```
michalkurka, line 274: it looks like there is no default (0), does this mean GL are not calculated anymore?

Pscheidl: The default is 0, as the parameter is int and that is by default 0 - if unset, it defaults to 0. The internal logic inside `GainsLift.java` is already wrapped around 0 as default.

h2o-py/h2o/estimators/naive_bayes.py
Pscheidl: Yes. https://github.com/h2oai/h2o-3/pull/4730/files/cbd7de9c2bcb6685d4f40888ab1999ea7c3ba1db#r444702721

Pscheidl: The original designed revolved around 0 as default, so I only kept it. If unspecified, H2O will "do it's own thing" automatically.

h2o-core/src/main/java/hex/GainsLift.java
Pscheidl: I don't see how this is possible. The `init` method has:

```
    if (_labels.cardinality() != -1 && _labels.cardinality() != 2)
      throw new IllegalArgumentException("Actual column must contain binary class labels, but found cardinality " + _labels.cardinality() + "!");
```

Which makes sure there is at least one non-response (or vice versa) in the actual labels. This hould result in 0.9999 rate but not `1`.

Pscheidl: I've added a test to demonstrate this behavior into `GainsLiftTest.java`.

Pscheidl: I can see some rounding errors could result in the resp. rate to be 1. I've put a guard against that into the code:

```
      final double cum_event = sum_e_i / (double)E;
      final double total_non_event = (double) (N - E);
      // If response rate is 1, there are non non-events and the cumulative count will always be zero
      final double cum_non_event = total_non_event == 0 ? 0 : (sum_n_i - sum_e_i) / total_non_event;
```

h2o-core/src/main/java/hex/ModelMetricsBinomial.java
michalkurka: @Pscheidl sorry, that was my misunderstanding, I didn't notice `gl.exec` is outside of the if-statement. Thanks for explaining.

Pscheidl: No problem at all !

h2o-py/h2o/estimators/naive_bayes.py
michalkurka: My point is that default is not actually 0, by default we have 16 bins that are partitioned using a non-linear scale. Since we are exposing this parameter it should say what the default does in the method documentation. Instead of using 0, I would maybe use -1 (use default 16) and 0 to disable (no bins). What do you think?

h2o-core/src/main/java/hex/GainsLift.java
michalkurka: Yes, but there are various edge cases that this check doesn't cover. You can eg. have zero weights for one of the predicted classes and the check will not catch that. See here: https://github.com/h2oai/h2o-3/blob/master/h2o-core/src/main/java/hex/GainsLift.java#L238

Pscheidl: A new test `testAverageResponseRate` has been added to test exactly that case. Resolved.

PR #4734, opened 2020-06-18
Quick (& dirty) POC of saving memory with GBM trees

michalkurka:
This PR introduces the idea of saving for Keyed objects that are essentially wrappers around a byte-array.

The perfect example of such an object is CompressedTree which represents both the scoring and auxiliary tree. By using a LightKeyed object we able to re-use the bytes stored in Value object and thus reducing the needed memory to a half.


h2o-core/src/main/java/water/MemKeyed.java
```diff
@@ -0,0 +1,31 @@
+package water;
+
+public class MemKeyed<T extends MemKeyed<T>> extends Keyed<T> {
```
mmalohlava, line 3: So this is something like `Chunk` if i understand well?

h2o-core/src/main/java/water/MemKeyed.java
```diff
@@ -0,0 +1,31 @@
...
+
+  public MemKeyed(Key<T> key, byte[] mem) {
+    super(key);
+    _mem = mem;
+  }
+
+  @SuppressWarnings("unchecked")
+  public final T reloadFromBytes(Key<T> k, byte[] ary) {
+    _key = k;
+    _mem = ary;
```
mmalohlava, line 15: Just to make sure i understand correctly, this is the key trick - simply do not materialize a new object from `_mem` but simply just keep reference (which is shared from `Value`).

h2o-core/src/main/java/water/MemKeyed.java
michalkurka: this is the same pattern as chunk except for it also is Keyed which Chunk itself isn't

h2o-core/src/main/java/water/MemKeyed.java
michalkurka: yes, precisely - that is the trick

PR #4735, opened 2020-06-19
PUBDEV-7627: k-LIME easy predict wrapper should use Regression or KLime as a model category instead of just KLime

navdeep-G:
* Resolves https://0xdata.atlassian.net/browse/PUBDEV-7627

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/algos/klime/KLimeMojoModel.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/algos/klime/KLimeMojoModel.java
@@ -1,7 +1,10 @@
 package hex.genmodel.algos.klime;
 
+import hex.ModelCategory;
 import hex.genmodel.MojoModel;
 import hex.genmodel.algos.glm.GlmMojoModel;
+
+import java.util.EnumSet;
 
 public class KLimeMojoModel extends MojoModel {
 
@@ -10,6 +13,11 @@ public class KLimeMojoModel extends MojoModel {
   MojoModel[] _clusterRegressionModels;
   int[] _rowSubsetMap;
 
+  private static EnumSet<ModelCategory> CATEGORIES = EnumSet.of(ModelCategory.Regression, ModelCategory.KLime);
+  @Override public EnumSet<ModelCategory> getModelCategories() {
+    return CATEGORIES;
+  }
+  
   KLimeMojoModel(String[] columns, String[][] domains, String responseColumn) {
     super(columns, domains, responseColumn);
   }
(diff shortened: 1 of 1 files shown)


navdeep-G:
@mmalohlava Thanks! Who else can look at this?

h2o-genmodel/src/main/java/hex/genmodel/algos/klime/KLimeMojoModel.java
```diff
@@ -1,15 +1,23 @@
...
 public class KLimeMojoModel extends MojoModel {
 
   MojoModel _clusteringModel;
   MojoModel _globalRegressionModel;
   MojoModel[] _clusterRegressionModels;
   int[] _rowSubsetMap;
 
+  private static EnumSet<ModelCategory> CATEGORIES = EnumSet.of(ModelCategory.Regression, ModelCategory.KLime);
+  @Override public EnumSet<ModelCategory> getModelCategories() {
+    return CATEGORIES;
```
Pscheidl, line 18: Knit: It's rather hard to make the EnumSet unmodifiable, as the contract of the method requires `EnumSet`. However, the value currently returned is a pointer to the CATEGORIES section and can be potentially manipulated by the client, which is not wanted.

I'd therefore return a new copy per each method call. The following patch should do:

```patch
Index: h2o-genmodel/src/main/java/hex/genmodel/algos/klime/KLimeMojoModel.java
IDEA additional info:
<+>UTF-8
--- h2o-genmodel/src/main/java/hex/genmodel/algos/klime/KLimeMojoModel.java	(revision 1529d2b1d87d215c3dc146cbe79cd6b696ce47c5)
+++ h2o-genmodel/src/main/java/hex/genmodel/algos/klime/KLimeMojoModel.java	(date 1594320134075)
@@ -13,9 +13,8 @@
   MojoModel[] _clusterRegressionModels;
   int[] _rowSubsetMap;
 
-  private static EnumSet<ModelCategory> CATEGORIES = EnumSet.of(ModelCategory.Regression, ModelCategory.KLime);
   @Override public EnumSet<ModelCategory> getModelCategories() {
-    return CATEGORIES;
+    return EnumSet.of(ModelCategory.Regression, ModelCategory.KLime);
   }
   
   KLimeMojoModel(String[] columns, String[][] domains, String responseColumn) {
```

However I'm fine with the PR anyway.

navdeep-G: I agree and have addressed that here: https://github.com/h2oai/h2o-3/pull/4735/commits/e99693a9df6d18cf6e779f8edb8ecd7b47f04251

mmalohlava: good catch.

2 comments:
 - i think that original pattern is already used somewhere in genmodel code - worth checking and fixing
 - what about using Collections.unmodifiableSet?

Pscheidl: We can't use it here, as it returns a more generic structure that's not EnumSet. These functions return EnumSet and it's hard to change the APP without breaking changes.

PR #4736, opened 2020-06-19
PUBDEV-7611: make GLM support early-stop

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7611?filter=-1

GAM will make use of the early-stopping capability in GLM.  However, GLM did not support early-stopping.  Hence, I have done the following:
1. add early stopping parameters to GAM and GLM;
2. enable early-stop in GLM.  This means disabling some of the early stopping mechanism in GLM.
3. early-stop is disabled when lambda_search=True.
4. R and Python unit tests are written to make sure early stopping is working.

### DIFF
--- a/h2o-algos/src/main/java/hex/gam/GAM.java
+++ b/h2o-algos/src/main/java/hex/gam/GAM.java
@@ -359,13 +359,16 @@ public class GAM extends ModelBuilder<GAMModel, GAMModel.GAMParameters, GAMModel
         _valid = rebalance(cleanUpInputFrame(_parms.valid().clone(), _parms, _gamColNamesCenter, _binvD, _zTranspose, _knots,
                 _numKnots), false, _result+".temporary.valid");
       }
+      Frame newValidFrame = _valid==null?null:new Frame(_valid);
       
       DKV.put(newTFrame); // This one will cause deleted vectors if add to Scope.track
+      if (newValidFrame != null)
+        DKV.put(newValidFrame);
       _job.update(0, "Initializing model training");
-      buildModel(newTFrame); // build gam model 
+      buildModel(newTFrame, newValidFrame); // build gam model 
     }
 
-    public final void buildModel(Frame newTFrame) {
+    public final void buildModel(Frame newTFrame, Frame newValidFrame) {
       GAMModel model = null;
       DataInfo dinfo = null;
       try {
@@ -383,7 +386,7 @@ public class GAM extends ModelBuilder<GAMModel, GAMModel.GAMParameters, GAMModel
           model._output._gamTransformedTrainCenter = newTFrame._key;
         }
         _job.update(1, "calling GLM to build GAM model...");
-        GLMModel glmModel = buildGLMModel(_parms, newTFrame); // obtained GLM model
+        GLMModel glmModel = buildGLMModel(_parms, newTFrame, newValidFrame); // obtained GLM model
         Scope.track_generic(glmModel);
         _job.update(0, "Building out GAM model...");
         fillOutGAMModel(glmModel, model, dinfo); // build up GAM model
@@ -407,6 +410,9 @@ public class GAM extends ModelBuilder<GAMModel, GAMModel.GAMParameters, GAMModel
         }
         if (dinfo!=null)
           dinfo.remove();
+        if (newValidFrame != null) {
+          DKV.remove(newValidFrame._key);
+        }
       }
     }
     
@@ -435,8 +441,8 @@ public class GAM extends ModelBuilder<GAMModel, GAMModel.GAMParameters, GAMModel
       }
     }
 
-    GLMModel buildGLMModel(GAMParameters parms, Frame trainData) {
-      GLMParameters glmParam = GamUtils.copyGAMParams2GLMParams(parms, trainData, valid());  // copy parameter from GAM to GLM
+    GLMModel buildGLMModel(GAMParameters parms, Frame trainData, Frame validFrame) {
+      GLMParameters glmParam = GamUtils.copyGAMParams2GLMParams(parms, trainData, validFrame);  // copy parameter from GAM to GLM
       int numGamCols = _parms._gam_columns.length;
       for (int find = 0; find < numGamCols; find++) {
         if ((_parms._scale != null) && (_parms._scale[find] != 1.0))
@@ -473,8 +479,8 @@ public class GAM extends ModelBuilder<GAMModel, GAMModel.GAMParameters, GAMModel
       model._output._glm_training_metrics = glmModel._output._training_metrics;
       if (valid() != null)
         model._output._glm_validation_metrics = glmModel._output._validation_metrics;
-      model._output._glm_scoring_history = model.copyTwoDimTable(glmModel._output._scoring_history);
       model._output._glm_model_summary = model.copyTwoDimTable(glmModel._output._model_summary);
+      model._output._glm_scoring_history = model.copyTwoDimTable(glmModel._output._scoring_history);
       if (_parms._family == multinomial || _parms._family == ordinal) {
         model._output._coefficients_table = model.genCoefficientTableMultinomial(new String[]{"Coefficients",
                         "Standardized Coefficients"}, model._output._model_beta_multinomial,
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -38,13 +38,9 @@ import water.util.*;
 
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Random;
+import java.util.*;
 
-import static hex.glm.GLMUtils.calSmoothNess;
-import static hex.glm.GLMUtils.extractAdaptedFrameIndices;
+import static hex.glm.GLMUtils.*;
 
 /**
  * Created by tomasnykodym on 8/27/14.
@@ -64,6 +60,8 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
   public String[][] _gamColnames = null;
   public int[][] _gamColIndices = null; // corresponding column indices in dataInfo
   public static int _totalBetaLen;
+  private boolean _earlyStopEnabled = false;
+
   public GLM(boolean startup_once){super(new GLMParameters(),startup_once);}
   public GLM(GLMModel.GLMParameters parms) {
     super(parms);
@@ -154,7 +152,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
           GLM g = (GLM) cvModelBuilders[i];
           double x = _parms._lambda[lidx];
           if (g._model._output.getSubmodel(x) == null)
-            g._driver.computeSubmodel(lidx, x);
+            g._driver.computeSubmodel(lidx, x, Double.NaN, Double.NaN);
           testDev += g._model._output.getSubmodel(x).devianceTest;
         }
         double testDevAvg = testDev / cvModelBuilders.length;
@@ -164,7 +162,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
           GLM g = (GLM) cvModelBuilders[i];
           double x = _parms._lambda[lidx];
           if (g._model._output.getSubmodel(x) == null)
-            g._driver.computeSubmodel(lidx, x); // stopped too early, need to FIT extra model
+            g._driver.computeSubmodel(lidx, x, Double.NaN, Double.NaN); // stopped too early, need to FIT extra model
           double diff = testDevAvg - (g._model._output.getSubmodel(x).devianceTest);
           testDevSE += diff*diff;
         }
@@ -379,6 +377,8 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
   private double _lmax;
   private transient long _nobs;
   private transient GLMModel _model;
+  private boolean _earlyStop = false;  // set by earlyStopping
+
   @Override
   public int nclasses() {
     if (_parms._family == Family.multinomial || _parms._family == Family.ordinal)
@@ -508,6 +508,12 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
     }
     if (expensive) {
       if (error_count() > 0) return;
+      if (_parms._lambda_search && (_parms._stopping_rounds > 0)) {
+        error("early stop", " cannot run when lambda_search=True.  Lambda_search has its own " +
+                "early-stopping mechanism");
+      }
+      if (!_parms._lambda_search && (_parms._stopping_rounds > 0))  // early stop is on!
+        _earlyStopEnabled = true;
       if (_parms._alpha == null)
         _parms._alpha = new double[]{_parms._solver == Solver.L_BFGS ? 0 : .5};
       if (_parms._lambda_search  &&_parms._nlambdas == -1)
@@ -848,6 +854,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
   public final class GLMDriver extends Driver implements ProgressMonitor {
     private long _workPerIteration;
     private transient double[][] _vcov;
+    List<Integer> _scoreIterationList = new ArrayList<Integer>(); // keep track of iteration where scoring occurs
 
 
     private void doCleanup() {
@@ -1766,7 +1773,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
         _state.set_likelihoodInfo(mmHGLMBd._hlik, mmHGLMBd._pvh, mmHGLMBd._pbvh, mmHGLMBd._caic);
       }
       mb.makeModelMetrics(_model, train, _dinfo._adaptedFrame, null);  // add generated metric to DKV
-      scoreEnd(train, t1);
+      scorePostProcessing(train, t1);
     }
     
     private void calHlikStuff(ModelMetricsHGLM.MetricBuilderHGLM mmHGLMBd, Frame glmmmeReturns, Frame augXZ) {
@@ -1978,14 +1985,24 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       long t1 = System.currentTimeMillis();
       Frame train = DKV.<Frame>getGet(_parms._train); // need to keep this frame to get scoring metrics back
       _model.score(train).delete();
-      scoreEnd(train, t1);
+      scorePostProcessing(train, t1);
     }
     
-    private void scoreEnd(Frame train, long t1) {
+    private void scorePostProcessing(Frame train, long t1) {
+      _scoreIterationList.add(_state._iter);
       ModelMetrics mtrain = ModelMetrics.getFromDKV(_model, train); // updated by model.scoreAndUpdateModel
       long t2 = System.currentTimeMillis();
+      if (_parms._lambda_search)
+        _model._output._scoring_history = _lsc.to2dTable();
+      else if (_parms._HGLM)
+        _model._output._scoring_history = _sc.to2dTableHGLM();
+      else
+        _model._output._scoring_history = _sc.to2dTable();
       if (!(mtrain == null)) {
         _model._output._training_metrics = mtrain;
+        _model._output._training_time_ms = t2-_model._output._start_time; // remember training time
+        ScoreKeeper trainScore = new ScoreKeeper(Double.NaN);
+        trainScore.fillFrom(mtrain);
         Log.info(LogMsg(mtrain.toString()));
       } else {
         Log.info(LogMsg("ModelMetrics mtrain is null"));
@@ -1995,8 +2012,11 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
         Frame valid = DKV.<Frame>getGet(_parms._valid);
         _model.score(valid).delete();
         _model._output._validation_metrics = ModelMetrics.getFromDKV(_model, valid); //updated by model.scoreAndUpdateModel
+        ScoreKeeper validScore = new ScoreKeeper(Double.NaN);
+        validScore.fillFrom(_model._output._validation_metrics);
       }
-      _model._output._scoring_history = _parms._lambda_search?_lsc.to2dTable():(_parms._HGLM?_sc.to2dTableHGLM():_sc.to2dTable());
+      
+      _model.addScoringInfo(_parms, nclasses(), t2);  // add to scoringInfo for early stopping
       _model.update(_job._key);
       if (_parms._HGLM)
         _model.generateSummaryHGLM(_parms._train,_state._iter);
@@ -2007,10 +2027,10 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       _scoringInterval = Math.max(_scoringInterval,20*scoringTime); // at most 5% overhead for scoring
     }
 
-    protected Submodel computeSubmodel(int i,double lambda) {
+    protected Submodel computeSubmodel(int i,double lambda, double nullDevTrain, double nullDevTest) {
       Submodel sm;
       if(lambda >= _lmax && _state.l1pen() > 0)
-        _model.addSubmodel(sm = new Submodel(lambda,getNullBeta(),_state._iter,_nullDevTrain,_nullDevTest));
+        _model.addSubmodel(sm = new Submodel(lambda,getNullBeta(),_state._iter, nullDevTrain, nullDevTest));
       else {  // this is also the path for HGLM model
         sm = new Submodel(lambda, _state.beta(),_state._iter,-1,-1);
         if (_parms._HGLM) // add random coefficients for random effects/columns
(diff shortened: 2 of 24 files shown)


h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -465,6 +465,8 @@ private void copyGLMtoGAMModel(GAMModel model, GLMModel glmModel) {
         model._output._glm_validation_metrics = glmModel._output._validation_metrics;
       model._output._glm_scoring_history = model.copyTwoDimTable(glmModel._output._scoring_history);
       model._output._glm_model_summary = model.copyTwoDimTable(glmModel._output._model_summary);
+      model._output._glm_scoring_history_early_stop = glmModel._output._scoring_history_early_stop==null?null:
+              model.copyTwoDimTable(glmModel._output._scoring_history_early_stop);
       if (_parms._family == multinomial || _parms._family == ordinal) {
```
wendycwong, line 470: I need access to scoring_history.  However, GLM already has its own, so I created one just for early_stop

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2154,7 +2149,7 @@ public void computeImpl() {
         oldDevTest = testDev;
         devHistoryTrain[i % devHistoryTrain.length] = (oldDevTrain - trainDev)/oldDevTrain;
         oldDevTrain = trainDev;
-        if(_parms._lambda[i] < _lmax && Double.isNaN(_lambdaCVEstimate) /** if we have cv lambda estimate we should use it, can not stop before reaching it */) {
+        if(_parms._lambda[i] < _lmax && Double.isNaN(_lambdaCVEstimate) && !_calScoreKeeper /** if we have cv lambda estimate we should use it, can not stop before reaching it */) {
           if (_parms._early_stopping && _state._iter >= devHistoryTrain.length) {
```
wendycwong, line 2153: disable the early stopping mechanism in GLM if early-stopping is enabled.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2269,34 +2269,34 @@ public void onCompletion(CountedCompleter caller) {
           _state.updateState(beta, gginfo);
         if (!_parms._lambda_search)
           updateProgress(true);
-        boolean converged = _state.converged();
+        boolean converged = _calScoreKeeper?false:_state.converged(); // GLM specific early stop.  Disable is early stop is enabled
         if (converged) Log.info(LogMsg(_state.convergenceMsg));
```
wendycwong, line 2273: disable the GLM stopping mechanism if early-stopping is enabled

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1087,6 +1089,18 @@ public void setZValues(double [] zValues, double dispersion, boolean dispersionE
     double [][] _vcov;
     private double _dispersion;
     private boolean _dispersionEstimated;
+    public ArrayList<ScoreKeeper> _scored_train;
+    public ArrayList<ScoreKeeper> _scored_valid;
+    public ArrayList<ScoreKeeper> _scored_xval;
+    public TwoDimTable _scoring_history_early_stop;
+    public ScoreKeeper[] scoreKeepers() {
```
wendycwong, line 1096: I use ArrayList for scored_train/valid/xval.  GLM iteration is fast and hence if we are to copy and extend the array every time, it can be slow???  Not sure.  I understand ArrayList has more overhead then arrays but could be worth it.  Please let me know guys.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2305,10 +2305,14 @@ protected void updateProgress(GLMModel fixedModel, GLMModel[] randModels, Frame
                                   double[] VC1, double[][] VC2, double sumDiff2, double convergence, boolean canScore,
                                   double[][] cholR, Frame augXZ) {
       _sc.addIterationScore(_state._iter, _state._sumEtaSquareConvergence);
-      if(canScore && (_parms._score_each_iteration || timeSinceLastScoring() > _scoringInterval)) {
+      if(canScore && (_parms._score_each_iteration || timeSinceLastScoring() > _scoringInterval || 
+              ((_parms._score_iteration_interval > 0) && ((_state._iter % _parms._score_iteration_interval) == 0)))) {
         _model.update(_state.expandBeta(_state.beta()), _state.ubeta(),-1, -1, _state._iter);
         scoreAndUpdateModelHGLM(fixedModel, randModels, glmmmeReturns, hvDataOnly, VC1, VC2, sumDiff2, convergence, 
                 cholR, augXZ, false);
+        _earlyStop = !_earlyStop?ScoreKeeper.stopEarly(_model._output.scoreKeepers(),
```
honzasterba, line 2313: can be simplified with = earlyStop ||

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2083,6 +2076,8 @@ protected Submodel computeSubmodel(int i,double lambda) {
 
     private transient double _nullDevTrain = Double.NaN;
     private transient double _nullDevTest = Double.NaN;
+    private boolean _earlyStop = false;  // set by earlyStopping
```
honzasterba, line 2079: this confused the hell out of me.
1) _nullDevTrain do not need to be instance variables with global scope, these can be just regular vars inside computeImpl
2) either move earlyStop to the top with the rest of the instance vars or (preferably) make it a regular var inside computeImpl

h2o-r/h2o-package/R/models.R
```diff
@@ -2029,6 +2029,21 @@ h2o.scoreHistory <- function(object) {
   }
 }
 
+#'
+#' Retrieve GLM Model Score History buried in GAM model
+#' @export
+h2o.scoreHistoryGAM <- function(object) {
+    o <- object
+    if( is(o, "H2OModel") ) {
+        sh <- o@model$glm_scoring_history
```
honzasterba, line 2038: why not just return( o@model$glm_scoring_history )? I dont understand the purpose of the if statement

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1408,7 +1425,111 @@ public Submodel getSubmodel(double lambdaCVEstimate) {
     }
     return res;
   }
-  
+
+
+  // create scoring history table using scoreKeeper arrays scored_train, scored_xval, scored_valid
+  public TwoDimTable createScoringHistoryTable(boolean hasValidation,
+                                                      boolean hasCrossValidation) {
```
honzasterba, line 1432: formatting

h2o-algos/src/main/java/hex/gam/GAM.java
honzasterba: I dont undestant the comment, scoring history is there for the reason of early stopping, so why do you need another one? how are they different?

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -64,6 +64,8 @@
   public String[][] _gamColnames = null;
   public int[][] _gamColIndices = null; // corresponding column indices in dataInfo
   public static int _totalBetaLen;
+  private boolean _calScoreKeeper = false; // true to update scored_train/scored_valid/scored_xval for early stop
```
honzasterba, line 67: should this be just _earlyStoppingEnabled? the commend does not make what the name means

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1979,38 +1987,19 @@ private void scoreAndUpdateModel(){
...
-      _lastScore = System.currentTimeMillis();
-      long scoringTime = System.currentTimeMillis() - t1;
-      _scoringInterval = Math.max(_scoringInterval,20*scoringTime); // at most 5% overhead for scoring
-*/
     }
     
     private void scoreEnd(Frame train, long t1) {
       ModelMetrics mtrain = ModelMetrics.getFromDKV(_model, train); // updated by model.scoreAndUpdateModel
       long t2 = System.currentTimeMillis();
+      _model._output._scoring_history = _parms._lambda_search?_lsc.to2dTable():(_parms._HGLM?_sc.to2dTableHGLM():_sc.to2dTable());
```
honzasterba, line 1995: I have no idea whats happening here, please refactor into a method and if statements, also, are you sure this are correct? I dont see any tests covering all these branches.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1408,7 +1425,111 @@ public Submodel getSubmodel(double lambdaCVEstimate) {
...
+      return table;
+    int iteration = 0;
+    for (ScoreKeeper si : _output._scored_train) { // fill out the table
+      int col = 0;
+      assert (row < table.getRowDim());
+      assert (col < table.getColDim());
+      DateTimeFormatter fmt = DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss");
+      long training_time_ms = _output._training_time_ms.get(row);
+      table.set(row, col++, fmt.print(training_time_ms));
+      table.set(row, col++, PrettyPrint.msecs(training_time_ms-_output._start_time, true));
```
honzasterba, line 1468: I am pretty sure this is supposed to be a duration of this stop not its time from beggining - there are missing tests for this code

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1408,7 +1425,111 @@ public Submodel getSubmodel(double lambdaCVEstimate) {
...
+      if (hasValidation && (_output._scored_valid.size() > 0))
+        col = setTableEntry(table, _output._scored_valid.get(row), row, col, modelCategory, isClassifier);
+      if (hasCrossValidation  && (_output._scored_xval.size() > 0))
+        setTableEntry(table, _output._scored_xval.get(row), row, col, modelCategory, isClassifier);
+      row++;
+    }
+    return table;
+  }
+
+  public void setColHeader(List<String> colHeaders, List<String> colTypes, List<String> colFormat,
```
honzasterba, line 1480: I find it very unlikely that all this new code is somehow specific to GLM, dont other algos create the same outputs?

h2o-algos/src/main/java/hex/gam/GAM.java
wendycwong: @honzasterba The GLM scoring history is totally different from the scoring history that is used for early stopping.  Since it has been there for a long time, I don't want to just remove it.  I added a new one for early stopping.  There are many differences between GLM and the other Algos.  It is before my time at H2O.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -508,6 +510,12 @@ public void init(boolean expensive) {
     }
     if (expensive) {
       if (error_count() > 0) return;
+      if (_parms._lambda_search && (_parms._stopping_rounds > 0)) {
+        warn("early stop", "disabled when lambda_search=true.");
+        _parms._stopping_rounds = 0;  // set stopping_round = 0 to disable early stopping
```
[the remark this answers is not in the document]
wendycwong: I was struggling with what to do here.  Your suggestion is a sound one.  I will throw an error then.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Moved nullDevTrain/nullDevTest to inside computeimpl.  Moved earlyStop to the top.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Clever!

h2o-r/h2o-package/R/models.R
wendycwong: Yes, that should work.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: thanks for the name

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: I changed the scoring history of gam to copy the glm scoring history early stop if that is enabled.  Otherwise, it will just be the glm scoring history.

I break down the model._output._scoring_history into if statements to make it clear.

h2o-algos/src/main/java/hex/gam/GAM.java
honzasterba: whats the original scoring history used for? can we somehow unify? it seems wasteful to have two, I thought that scoring is done exactly for the reason of early stopping

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
```diff
@@ -38,7 +38,8 @@
     return array3D;
   }
 
-  public static enum AllocateType {firstOneLess, sameOrig, bothOneLess, firstTwoLess} // special functions are performed depending on GLMType.  Internal use
+  public static enum
```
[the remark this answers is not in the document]
honzasterba: marked as resolved yet the change is still here

h2o-algos/src/main/java/hex/gam/GAM.java
wendycwong: Nope.  GLM is written very differently from the others.  The glm scoring history is very light weight and very simple.  Since it is there for a very long time, I don't want to change it to break some customer code.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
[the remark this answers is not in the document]
wendycwong: I have not pushed my changes yet.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Actually, I just have gam copy the correct scoring_history and do not have to deal with glm scoring history.

h2o-algos/src/main/java/hex/gam/GAM.java
wendycwong: I have combined the two scoring_history into one with the original GLM scoring history coming first.  Hope this minimizes breaking anyone's code

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2319,6 +2325,9 @@ protected void updateProgress(boolean canScore){
       if(canScore && (_parms._score_each_iteration || timeSinceLastScoring() > _scoringInterval)) {
         _model.update(_state.expandBeta(_state.beta()), -1, -1, _state._iter);
         scoreAndUpdateModel();
+        _earlyStop = !_earlyStop?ScoreKeeper.stopEarly(_model.scoreKeepers(),
```
koniecsveta, line 2328: readability of this could be improved - probably extract the expression to method since the ternary expression seems to me too complicated already

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -40,6 +38,45 @@ public GLMModel(Key selfKey, GLMParameters parms, GLM job, double [] ymu, double
...
+    scoringInfo[0].is_classification = nclasses > 1;
+    scoringInfo[0].validation = parms.valid() != null;
+    scoringInfo[0].time_stamp_ms = _output._start_time;
+  }
+
+  public ScoreKeeper[] scoreKeepers() {
+    int size = scoringInfo==null?0:scoringInfo.length;
+    ScoreKeeper[] sk = new ScoreKeeper[size];
+    for (int i=0;i<size;++i) {
+      sk[i] = scoringInfo[i].cross_validation ? scoringInfo[i].scored_xval
```
koniecsveta, line 53: I would recommend using parentheses when combining multiple ternary operators, overall, the formatting of the code is not ideal and its hard to read it

h2o-algos/src/main/java/hex/glm/GLMUtils.java
```diff
@@ -31,6 +33,52 @@
     return gamColIndices;
   }
 
+  public static TwoDimTable combineScoringHistory(TwoDimTable glmSc1, TwoDimTable earlyStopSc2) {
+    String[] esColTypes = earlyStopSc2.getColTypes();
+    String[] esColHeaders = earlyStopSc2.getRowHeaders();
+    String[] esColFormats = earlyStopSc2.getColFormats();
+    ArrayList<String> finalRowHeaders = new ArrayList<>(Arrays.asList(glmSc1.getRowHeaders()));
```
koniecsveta, line 40: I think that interface List should be used instead of the ArrayList for variable type definition on all places (I did not notice usage of anything ArrayList specific in the code) - then, you don't have to use the ArrayList constructor and remove redundant creation of another list and just assign the return value of Arrays.asList to variable

wendycwong: @koniecsveta:  I am trying to combine two lists into one final list.  Since I do not know how long the final list is going to be, I use ArrayList which will automatically expand to accommodate any final length I want.  In addition, I also need to preserve the order of the combination of the two lists as well.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: I simplified it.

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: I just break it out into if elseif else statements to make it clear.

wendycwong:
@honzasterba I just combine the two scoring_history into one preserving the order of the first one and adding the second one to the end of the first one without repeating fields.  Now, there is only one scoring_history at the end.  This should not break any codes I hope.

h2o-algos/src/main/java/hex/glm/GLMUtils.java
koniecsveta: I agree with the ArrayList usage but the comment was suggesting this change (for all five list initializations):

`List<String> finalRowHeaders = Arrays.asList(glmSc1.getRowHeaders());`

since the interface should be used for variable type definition and Arrays.asList already creates new instance of ArrayList, therefore, putting this as constructor parameter is unnecessary

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1978,39 +1987,23 @@ private void scoreAndUpdateModel(){
       long t1 = System.currentTimeMillis();
       Frame train = DKV.<Frame>getGet(_parms._train); // need to keep this frame to get scoring metrics back
       _model.score(train).delete();
-      scoreEnd(train, t1);
-/*
```
honzasterba, line 1982: remove commented out code

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1081,6 +1120,7 @@ public void setZValues(double [] zValues, double dispersion, boolean dispersionE
     public String[] _coefficient_names;
     String[] _random_coefficient_names; // for HGLM
     String[] _random_column_names;
+    public ArrayList<Long> _training_time_ms = new ArrayList<>();
```
honzasterba, line 1123: use List for variable types

h2o-algos/src/main/java/hex/glm/GLMUtils.java
honzasterba: what zuzana is saying is that this should be `List<String> finalRowHeaders = ...`

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -3169,10 +3168,13 @@ def extract_from_twoDimTable(metricOfInterest, fieldOfInterest, takeFirst=False)
     """
 
     allFields = metricOfInterest._col_header
+    return  extract_field_from_twoDimTable(allFields, metricOfInterest.cell_values, fieldOfInterest, takeFirst=False)
+
+def extract_field_from_twoDimTable(allFields, cell_values, fieldOfInterest, takeFirst=False):
```
honzasterba, line 3173: indent seems off

h2o-r/h2o-package/R/models.R
```diff
@@ -2029,6 +2029,22 @@ h2o.scoreHistory <- function(object) {
   }
 }
 
+#'
+#' Retrieve GLM Model Score History buried in GAM model
+#' @export
+h2o.scoreHistoryGAM <- function(object) {
+    return(object@model$glm_scoring_history)
+    #o <- object
```
honzasterba, line 2037: remove commentd out code

h2o-algos/src/main/java/hex/glm/GLMUtils.java
wendycwong: Actually, I need to do List<String> finalRowHeaders =new ArrayList<>( Arrays.asList(glmSc1.getRowHeaders()));

in order to be able to use finalRowHeaders.add(colName);

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: Yes, there is an extra space between return and extract_field..

h2o-r/h2o-package/R/models.R
wendycwong: Removed.

h2o-algos/src/main/java/hex/glm/GLMUtils.java
koniecsveta: java.util.Arrays.asList() looks like this:

`public static <T> List<T> asList(T... a) {
        return new ArrayList<>(a);
    }`

so by doing
` new ArrayList<>( Arrays.asList(glmSc1.getRowHeaders()));`
you actually do 
`new ArrayList<>(new ArrayList<>(glmSc1.getRowHeaders()))`

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2305,10 +2307,14 @@ protected void updateProgress(GLMModel fixedModel, GLMModel[] randModels, Frame
                                   double[] VC1, double[][] VC2, double sumDiff2, double convergence, boolean canScore,
                                   double[][] cholR, Frame augXZ) {
       _sc.addIterationScore(_state._iter, _state._sumEtaSquareConvergence);
-      if(canScore && (_parms._score_each_iteration || timeSinceLastScoring() > _scoringInterval)) {
+      if(canScore && (_parms._score_each_iteration || timeSinceLastScoring() > _scoringInterval || 
+              ((_parms._score_iteration_interval > 0) && ((_state._iter % _parms._score_iteration_interval) == 0)))) {
         _model.update(_state.expandBeta(_state.beta()), _state.ubeta(),-1, -1, _state._iter);
         scoreAndUpdateModelHGLM(fixedModel, randModels, glmmmeReturns, hvDataOnly, VC1, VC2, sumDiff2, convergence, 
                 cholR, augXZ, false);
+        _earlyStop = ScoreKeeper.stopEarly(_model.scoreKeepers(),
```
koniecsveta, line 2315: I would suggest to change it to `_earlyStop = _earlyStop || ...;` to avoid unnecessary method evaluation if the `_earlyStop` is already `true`

h2o-algos/src/main/java/hex/glm/GLMUtils.java
wendycwong: Got it.  I think I got them all.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Nice optimization!

h2o-algos/src/main/java/hex/glm/GLM.java
honzasterba: unresolving this comment since the code was not removed

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2269,34 +2269,34 @@ public void onCompletion(CountedCompleter caller) {
           _state.updateState(beta, gginfo);
         if (!_parms._lambda_search)
           updateProgress(true);
-        boolean converged = _state.converged();
+        boolean converged = _earlyStopEnabled ?false:_state.converged(); // GLM specific early stop.  Disable is early stop is enabled
```
honzasterba, line 2272: could be simplified to `_earlyStopEnabled && _state.converged()`
or made more explicit by `_earlyStopEnabled ? _earlyStop : _state.converged()` and removing `&& ! _earlyStop` from L#2274

also typo in the comment `Disable is early...`should be `Disable if early...`

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2269,34 +2269,34 @@ public void onCompletion(CountedCompleter caller) {
...
-        return !stop_requested() && !converged && _state._iter < _parms._max_iterations;
+        return !stop_requested() && !converged && _state._iter < _parms._max_iterations && !_earlyStop;
       }
     }
 
     public boolean progressHGLMGLMMME(double sumDiff2, double sumeta2, int iteration, boolean atGLMMME, GLMModel 
             fixedModel, GLMModel[] randModels, Frame glmmmeReturns, Frame hvDataOnly, double[] VC1, double[][] VC2,
                                       double[][] cholR, Frame augZ) {
-      boolean converged = (sumDiff2 < (_parms._objective_epsilon*sumeta2));
+      boolean converged = _earlyStopEnabled ?false:(sumDiff2 < (_parms._objective_epsilon*sumeta2));
```
honzasterba, line 2281: could be simplified same as above

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2269,34 +2269,34 @@ public void onCompletion(CountedCompleter caller) {
...
+      return !stop_requested() && !converged && (iteration < _parms._max_iterations) && !_earlyStop;
     }
     
     public boolean progress(double [] beta, double likelihood) {
       _state._iter++;
       _state.updateState(beta,likelihood);
       if(!_parms._lambda_search)
         updateProgress(true);
-      boolean converged = _state.converged();
+      boolean converged = _earlyStopEnabled ? false:_state.converged();
```
honzasterba, line 2297: I see a pattern here, this code is repeated twice, should be put into a converged() method

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2319,6 +2323,10 @@ protected void updateProgress(boolean canScore){
       if(canScore && (_parms._score_each_iteration || timeSinceLastScoring() > _scoringInterval)) {
         _model.update(_state.expandBeta(_state.beta()), -1, -1, _state._iter);
         scoreAndUpdateModel();
+        if (!_earlyStop)
```
honzasterba, line 2326: simplify same as above _earlyStop = _earlyStop || ScoreKeeper...
also, this is the exact same code as above, so even better move to  a `updateEarlyStop()` method

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: It is done.

wendycwong: It is removed as I am checking now.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Fixed the type.  Good catch on the simplification

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Fixed.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Yeah, good idea.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Done.

PR #4737, opened 2020-06-19
Clarification in docs for balance_classes in automl and for supported POJOs

angela0xdata:
- Clarified oversampling/undersampling for balance_classes in AutoML
- Added Word2Vec to list of unsupported algos for POJOs


Pscheidl:
Thank you.

PR #4742, opened 2020-06-24
PUBDEV-7367: Modify Random Discrete Hyper Space Walker for GAM specific params

km2018:
JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7367

PR adds code to the random discrete hyper space walker such that for a grid search with GAM specific parameters, only legal combinations are allowed, specifically it forces the lengths of the arrays for gam_columns, scale, bs, and num_knots to all be the same.

This branch is being tested here: https://github.com/h2oai/h2o-3/pull/4744.  However, please put your review here.


h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -499,18 +502,48 @@ public void onModelFailure(Model failedModel, Consumer<Object[]> withFailedModel
...
+         * @param param
+         * @param param_values
+         * @return
+         */
+        private int nextGAMParamIndex(int gam_columns_length, String param, Object[] param_values) {
+          int[] filtered_param_indices = IntStream.range(0, _hyperParams.get(param).length)
+                  .filter(i -> ((ArrayList<Integer>)(param_values[i])).size() == gam_columns_length).toArray();
+          return filtered_param_indices[_random.nextInt(filtered_param_indices.length)];
+        }
+
```
wendycwong, line 546: This is good.

wendycwong:
karthik: 

Good job with the Java backend.  However, need to make sure the tests are written to test what you have implemented.  The current tests would like invoke randomDiscrete Gridsearch.  I left comments in the code on how to fix that.  Let's discuss this more in our meeting.

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7367_random_gridsearch_specific.py
```diff
@@ -0,0 +1,60 @@
...
+        for model in self.h2o_model.models:
+            print(model.actual_params['scale'], model.actual_params['gam_columns'], model.actual_params['num_knots'])
+        print("done")
+
+def test_gridsearch():
+    test_gam_grid = test_gam_gridsearch()
+    test_gam_grid.train_models()
+
+if __name__ == "__main__":
+    h2o.init(ip='192.168.1.2', port=54321, strict_version_check=False)
```
wendycwong, line 56: Karthik:  Remember to remove this before you check the files in.  It only works for you on your computer;.

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7367_random_gridsearch_specific.py
```diff
@@ -0,0 +1,60 @@
...
+    pyunit_utils.standalone_test(test_gridsearch)
+else:
+    h2o.init(ip='192.168.1.2', port=54321, strict_version_check=False)
+    test_gridsearch()
```
wendycwong, line 60: Removed the one here too.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -226,10 +228,11 @@ public MP getParams() {
     protected MP getModelParams(MP params, Object[] hyperParams) {
       ModelParametersBuilderFactory.ModelParametersBuilder<MP>
               paramsBuilder = _paramsBuilderFactory.get(params);
+      
       for (int i = 0; i < _hyperParamNames.length; i++) {
         String paramName = _hyperParamNames[i];
         Object paramValue = hyperParams[i];
-
+        
         if (paramName.equals("valid")) {  // change paramValue to key<Frame> for validation_frame
```
wendycwong, line 236: Karthik:  Remove this text formatting.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -226,10 +228,11 @@ public MP getParams() {
     protected MP getModelParams(MP params, Object[] hyperParams) {
       ModelParametersBuilderFactory.ModelParametersBuilder<MP>
               paramsBuilder = _paramsBuilderFactory.get(params);
+
```
wendycwong, line 231: Remove this text formatting.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -499,18 +502,49 @@ public void onModelFailure(Model failedModel, Consumer<Object[]> withFailedModel
...
-              hyperparamIndices[i] = _random.nextInt(_hyperParams.get(_hyperParamNames[i]).length);
+              if(_hyperParamNames[i].equals("gam_columns")) {
+                hyperparamIndices[i] = gam_columns_index;
+              } else if(gam_columns_index != -1 && gam_columns_length != 0 && gam_specific_params.contains(_hyperParamNames[i])) {
+                hyperparamIndices[i] = nextGAMParamIndex(gam_columns_length, _hyperParamNames[i], _hyperParams.get(_hyperParamNames[i]));
+              } else {
+                hyperparamIndices[i] = _random.nextInt(_hyperParams.get(_hyperParamNames[i]).length);  
+              }
             }
+            // Log.info(“scale_” + scale + ”, num_knots:”+numknots);
```
wendycwong, line 524: Remove this commented out code.

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7367_random_gridsearch_specific.py
```diff
@@ -0,0 +1,83 @@
...
+# If the coefficients do not match or an incorrect number of models is generated, the test throws an assertion error.
+class test_random_gam_gridsearch_specific:
+
+    h2o_data = []
+    myX = []
+    myY = []
+    h2o_model = []
+    search_criteria = {'strategy': 'RandomDiscrete', "max_models": 8, "seed": 1}
+    hyper_parameters = {'scale': [[1, 1], [2, 2], [0.01], [0.05]], 'gam_columns': [["C6", "C7"], ["C6"]], 'num_knots': [[5, 5], [6, 6], [5], [6]]}
+    manual_gam_models = []
```
wendycwong, line 24: Karthik:  Can you add a gam-column set of three columns?  and add two choices for it in scale and num_knots?  Just want to check when things get bigger, it still works.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -11,8 +11,10 @@
 import water.exceptions.H2OIllegalArgumentException;
 import water.util.PojoUtils;
 
+import java.lang.reflect.Array;
```
honzasterba, line 14: unused

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -499,18 +501,48 @@ public void onModelFailure(Model failedModel, Consumer<Object[]> withFailedModel
          */
         private int[] nextModelIndices() {
           int[] hyperparamIndices =  new int[_hyperParamNames.length];
-
+          Set<String> gam_specific_params = new HashSet<>(Arrays.asList("bs", "scale", "num_knots"));
```
honzasterba, line 504: should be a private static final constant with unmodifiableSet

wendycwong: @honzasterba:

Basically, this is what is going on with GAM:

1. a user can choose which sets of gam columns to be included to transform into gam columns.  Some times you may want to run gam with say [col0, col1] and sometimes, you may want to try to run gam with [col1, col2, col3].  We can make the gridsearch more restrictive for the users to force them to only choose a subset of 2 gam columns or N-gam columns.  Doing it this way, all the scale/num_knots/bs will have the same length.

2. Our current implementation is to allow users to choose any gam column sizes.  Then, in scale/num_knots/bs, users can also add different sizes.  However, our code will automatically choose the correct size scale/num_knots/bs according to the gam_column that is randomly chosen.

If we go with 1., we need to make sure the documentation reflect that the user can gridsearch over gam columns of a fixed size.  If they want to gridsearch over gam columns of different sizes, they will need to do it manually.  Let me know what you think.  Thanks, Wendy

honzasterba: so we could also fix this by mandating that all scale/... arrays be same or bigger than smallest GAM-columns array

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7367_random_gridsearch_grouped_params.py
```diff
@@ -0,0 +1,87 @@
...
+def test_gridsearch():
+    test_gam_grid = test_random_gam_gridsearch_specific()
+    test_gam_grid.train_models()
+    test_gam_grid.match_models()
+
+if __name__ == "__main__":
+    h2o.init(ip='192.168.1.5', port=54321, strict_version_check=False)
+    pyunit_utils.standalone_test(test_gridsearch)
+else:
+    h2o.init(ip='192.168.1.5', port=54321, strict_version_check=False)
```
wendycwong, line 86: Remove h2o.init

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7367_gridsearch_grouped_params.py
```diff
@@ -0,0 +1,84 @@
...
+
+        assert self.num_grid_models == self.num_expected_models, "Grid search model parameters incorrect or incorrect number of models generated"
+
+def test_gridsearch_specific():
+    test_gam_grid = test_gam_gridsearch_specific()
+    test_gam_grid.train_models()
+    test_gam_grid.match_models()
+
+if __name__ == "__main__":
+    h2o.init(ip='192.168.1.5', port=54321, strict_version_check=False)
```
wendycwong, line 80: remove h2o.init

wendycwong:
Karthik:  Add the following test in which I failed to give a valid choice for scale.  Hence, scale will not have a correct array size when gam_column size == 1.  This currently run into the following error:
	at hex.grid.HyperSpaceWalker$BaseWalker.hypers(HyperSpaceWalker.java:298)
	at hex.grid.HyperSpaceWalker$RandomDiscreteValueWalker$1.onModelFailure(HyperSpaceWalker.java:562)
	at hex.grid.GridSearch.gridSearch(GridSearch.java:392)
	at hex.grid.GridSearch.access$000(GridSearch.java:69)
	at hex.grid.GridSearch$1.compute2(GridSearch.java:140)
	at water.H2O$H2OCountedCompleter.compute(H2O.java:1557)

Please fix it. 
Here is the test:
from __future__ import division
from __future__ import print_function
from past.utils import old_div
import sys
sys.path.insert(1, "../../../")
import h2o
from tests import pyunit_utils
from h2o.estimators.gam import H2OGeneralizedAdditiveEstimator

# in this test, I am going to not include a correct length in Scale and see if the algo will work.
class test_random_gam_gridsearch_specific:

    h2o_data = []
    myX = []
    myY = []
    h2o_model = []
    search_criteria = {'strategy': 'RandomDiscrete', "max_models": 12, "seed": 1, "grouped_parameters": ['scale', 'gam_columns', 'num_knots']}
    hyper_parameters = {'scale': [[1, 1, 1], [2, 2, 2], [1, 1], [2, 2]],
                        'gam_columns': [["C6", "C7", "C8"], ["C6", "C7"], ["C6"]],
                        'num_knots': [[5, 5, 5], [6, 6, 6], [5, 5], [6, 6], [5], [6]]}
    manual_gam_models = []
    num_grid_models = 0
    num_expected_models = 12

    def __init__(self):
        self.setup_data()

    def setup_data(self):
        """
        This function performs all initializations necessary:
        load the data sets and set the training set indices and response column index
        """
        self.h2o_data = h2o.import_file(
            path = pyunit_utils.locate("smalldata/glm_test/multinomial_10_classes_10_cols_10000_Rows_train.csv"))
        self.h2o_data["C1"] = self.h2o_data["C1"].asfactor()
        self.h2o_data["C2"] = self.h2o_data["C2"].asfactor()
        self.myX = ["C1", "C2"]
        self.myY = "C11"
        self.h2o_data["C11"] = self.h2o_data["C11"].asfactor()
        for scale in self.hyper_parameters['scale']:
            for gam_columns in self.hyper_parameters['gam_columns']:
                for num_knots in self.hyper_parameters['num_knots']:
                    if len(scale) != len(gam_columns) or len(num_knots) != len(gam_columns):
                        continue
                    self.manual_gam_models.append(H2OGeneralizedAdditiveEstimator(family="multinomial", gam_columns=gam_columns,
                                                                                  keep_gam_cols=True, scale=scale, num_knots=num_knots,
                                                                                  ))

    def train_models(self):
        self.h2o_model = H2OGridSearch(H2OGeneralizedAdditiveEstimator(family="multinomial",
                                                                       keep_gam_cols=True), hyper_params=self.hyper_parameters, search_criteria=self.search_criteria)
        self.h2o_model.train(x = self.myX, y = self.myY, training_frame = self.h2o_data)
        for model in self.manual_gam_models:
            model.train(x = self.myX, y = self.myY, training_frame = self.h2o_data)
        print("done")

    def match_models(self):
        for model in self.manual_gam_models:
            scale = model.actual_params['scale']
            gam_columns = model.actual_params['gam_columns']
            num_knots = model.actual_params['num_knots']
            for grid_search_model in self.h2o_model.models:
                if grid_search_model.actual_params['gam_columns'] == gam_columns \
                    and grid_search_model.actual_params['scale'] == scale \
                    and grid_search_model.actual_params['num_knots'] == num_knots:
                    self.num_grid_models += 1
                    assert grid_search_model.coef() == model.coef(), "coefficients should be equal"
                    break

        assert self.num_grid_models == self.num_expected_models, "Grid search model parameters incorrect or incorrect number of models generated"

def test_gridsearch():
    test_gam_grid = test_random_gam_gridsearch_specific()
    test_gam_grid.train_models()
    test_gam_grid.match_models()

if __name__ == "__main__":
    pyunit_utils.standalone_test(test_gridsearch)
else:
    test_gridsearch()

wendycwong:
Karthik:

Please also add the following test and make sure it runs:

from __future__ import division
from __future__ import print_function
from past.utils import old_div
import sys
sys.path.insert(1, "../../../")
import h2o
from tests import pyunit_utils
from h2o.estimators.gam import H2OGeneralizedAdditiveEstimator

# In this test, we check to make sure that a grid search on a GAM works with GAM-specific parameters.
# The test searches over 4 parameters, alpha, lambda, scale, and num_knots.
# The test then compares the results of the grid search models with the models we created
# by manually searching over the hyperspace.
# If the coefficients do not match or an incorrect number of models is generated, the test throws an assertion error.
class test_gam_gridsearch_specific:

    h2o_data = []
    myX = []
    myY = []
    search_criteria = {'strategy': 'Cartesian', "grouped_parameters": ['scale', 'gam_columns', 'num_knots']}
    hyper_parameters = {'scale': [[1], [2], [1, 1, 1], [2, 2, 2]], 
                        'num_knots': [[5, 5], [6, 6], [5, 5, 5], [5, 6, 7]],
                        'gam_columns': [["C11", "C12", "C13"], ["C11", "C12"], ["C12", "C13"]]}
    manual_gam_models = []
    h2o_model = []
    num_grid_models = 0
    num_expected_models = 4

    def __init__(self):
        self.setup_data()

    def setup_data(self):
        """
        This function performs all initializations necessary:
        load the data sets and set the training set indices and response column index
        """
        self.h2o_data = h2o.import_file(
            path = pyunit_utils.locate("smalldata/glm_test/gaussian_20cols_10000Rows.csv"))
        self.h2o_data["C1"] = self.h2o_data["C1"].asfactor()
        self.h2o_data["C2"] = self.h2o_data["C2"].asfactor()
        self.myX = ["C1", "C2"]
        self.myY = "C21"
        for gam_columns in self.hyper_parameters['gam_columns']:
            for scale in self.hyper_parameters['scale']:
                for num_knots in self.hyper_parameters['num_knots']:
                    if len(scale) != len(gam_columns) or len(scale) != len(num_knots):
                        continue
                    self.manual_gam_models.append(H2OGeneralizedAdditiveEstimator(family = "gaussian", gam_columns=gam_columns,
                                                                                  keep_gam_cols = True, scale = scale, num_knots = num_knots))

    def train_models(self):
        self.h2o_model = H2OGridSearch(H2OGeneralizedAdditiveEstimator(
            family = "gaussian", keep_gam_cols = True), self.hyper_parameters, search_criteria=self.search_criteria)
        self.h2o_model.train(x = self.myX, y = self.myY, training_frame = self.h2o_data)
        for model in self.manual_gam_models:
            model.train(x = self.myX, y = self.myY, training_frame = self.h2o_data)

    def match_models(self):
        for model in self.manual_gam_models:
            gam_columns = model.actual_params['gam_columns']
            scale = model.actual_params['scale']
            num_knots = model.actual_params['num_knots']
            for grid_search_model in self.h2o_model.models:
                if grid_search_model.actual_params['gam_columns'] == gam_columns \
                    and grid_search_model.actual_params['scale'] == scale \
                    and grid_search_model.actual_params['num_knots'] == num_knots:
                    self.num_grid_models += 1
                    assert grid_search_model.coef() == model.coef(), "coefficients should be equal"
                    break

        assert self.num_grid_models == self.num_expected_models, "Grid search model parameters incorrect or incorrect number of models generated"

def test_gridsearch_specific():
    test_gam_grid = test_gam_gridsearch_specific()
    test_gam_grid.train_models()
    test_gam_grid.match_models()

if __name__ == "__main__":
    pyunit_utils.standalone_test(test_gridsearch_specific)
else:
    test_gridsearch_specific()

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -499,18 +501,48 @@ public void onModelFailure(Model failedModel, Consumer<Object[]> withFailedModel
...
+          int gam_columns_index = -1;
```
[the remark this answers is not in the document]
wendycwong: @honzasterba  Yes, I think this is a great idea.  Karthik has come up with some way to do it without being algo-specific

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
wendycwong: @honzasterba a non-algo specific is used.

wendycwong:
@km2018 The following tests are failing on Jenkins:
Test Result (14 failures / ±0)
Py2.7 Smoke / r_suite.pyunit_gbm_grid.py
Py2.7 Smoke / r_suite.pyunit_grid_carsGLM.py
Py2.7 Smoke / r_suite.pyunit_grid_carsNB.py
Py2.7 Smoke / r_suite.pyunit_iris_grid_deeplearning.py
Py2.7 Smoke / r_suite.pyunit_kmeans_grid_iris.py
R3.5 Smoke / r_suite.runit_GLM_grid_cars.R
R3.5 Smoke / r_suite.runit_kmeans_grid_iris.R
R3.5 Smoke / r_suite.runit_kmeans_grid_iris_negative.R
R3.5 Smoke / r_suite.runit_naivebayes_grid_cars_negative.R
Java 8 Smoke / hex.kmeans.KMeansGridTest.testIrisGrid
Java 8 Smoke / hex.kmeans.KMeansGridTest.testDuplicatesCarsGrid
Java 8 Smoke / hex.kmeans.KMeansGridTest.testUserPointsCarsGrid
Java 8 Smoke / hex.tree.drf.DRFGridTest.testDuplicatesCarsGrid
Java 8 Smoke / hex.tree.gbm.GBMGridTest.testDuplicatesCarsGrid

Please run them by hand and make sure they pass.  Thanks.

wendycwong:
Karthik:  Much better, only these are failing now:

R3.5 Medium-large / r_suite.runit_GBMRandomGrid_airlines_large.R
Py3.5 Small / r_suite.pyunit_benign_glm_grid.py
Py3.5 Small / r_suite.pyunit_gbm_random_grid.py
Py3.5 Small / r_suite.pyunit_h2oget_grid.py

sebhrusen:
Hey @km2018 , I have some concerns regarding the generality, or rather specifity of your solution.
If I understand correctly, the idea is to use only search parameter values that are compatible among params listed in `grouped_parameters`: this is the general objective.
The implementation/solution however looks overly specific to me, meaning it may be good enough for GAM, but also limited in terms of what kind of constraints we can apply: basically we consider among grouped parameters, we can only match "array values" (first limitation) of same length (second limitation).

As this choice also determines how the user defines those constraints, I'm afraid it could make it more difficult to create more generic constraints in the future.
For example, taking your example here
```python
search_criteria = {'strategy': 'Cartesian', "grouped_parameters": ['scale', 'gam_columns', 'num_knots']}
hyper_parameters = {'scale': [[1], [2], [1, 1, 1], [2, 2, 2]], 
                    'num_knots': [[5, 5], [6, 6], [5, 5, 5], [5, 6, 7]],
                    'gam_columns': [["C11", "C12", "C13"], ["C11", "C12"], ["C12", "C13"]]}
...
num_expected_models = 4
```
First, we let the user specify values that won't be used without apparent notice, and second it's very hard for him/her to see the exact constraints, especially as the `hyper_parameters` logic is now partly declared in the `search_criteria`.

From a user perspective, I expect something more explicit, for example:
```python
hyper_parameters = dict(
    max_iterations=[100, 200, 500],
    constraints=[
        dict(
            gam_columns=[["C11", "C12", "C13"]],
            num_knots=[[5, 5, 5], [5, 6, 7]],
            scale=[[1, 1, 1], [2, 2, 2]],
        ),
        dict(
            gam_columns=[["C11", "C12"], ["C12", "C13"]],
            num_knots=[[5, 5], [6, 6]],
            scale=[[1, 1], [2, 2], [1, 2]],
        ),
    ]
)
```
Note that this allows me to declare a much larger variety of constraints, for example:
```python
hyper_parameters = dict(
    constraints=[
        dict(lambda_search=True, 
               nlambdas=[10, 30, 50, 100]),
        dict(lambda_search=False,
               lambda=[1, .5, .2, .1, .05, .01])
    ]
)
```
Then remains the problem of multiplying those constraints, which is an "implementation detail". 
First, we could do some validation: a parameter should not appear both as a "free param" (see `max_iterations` in my first example), and a constrained one (any param like `gam_columns`, `num_knots`, `scale` in this same example appearing in a constraint). Thanks to this, we are able to multiply "free params" will all constraints.
Second, we could decide to only multiply constraints that don't intersect (no param in common). In this first example, it means we can multiply `max_iterations` with all combinations in the first constraint or in the second, but we can't multiply combinations from those constraints. 
However, if we add the constraints from the second example to the first example, then we can multiply the constraints on lambdas with the ones on smoothing.

Does it make sense?
Again, my point here is to provide an API that looks clear for the user, and generic enough to allow constraints that are not only based on array length...

wendycwong:
@sebhrusen You can actually choose not to specify anything in grouped_parameters.  In this case like this: search_criteria = {'strategy': 'Cartesian'}
hyper_parameters = {'scale': [[1], [2], [1, 1, 1], [2, 2, 2]], 
                    'num_knots': [[5, 5], [6, 6], [5, 5, 5], [5, 6, 7]],
                    'gam_columns': [["C11", "C12", "C13"], ["C11", "C12"], ["C12", "C13"]]
                    'alpha': [0.1,0.5,0.9]}

There could be 144 models but only 12 models will be returned since there are only 12 legal model parameter combinations.

We felt that it is rather wasteful and hence would like to not try all possible parameter combinations but rather legal parameter combinations only.  Hence Karthik comes up with this idea to minimize code change and yet enable users to specify what parameters have size limitations.  This does not restrict the free parameter combination generation.  It only constraints the grouped_parameter generation to ensure only legal combs are considered.

Your solution is definitely very elegant.  I like it at first sight already.  The implementation will be interesting to Karthik.  He is always looking for challenge.

honzasterba:
rebase to master

wendycwong:
@km2018

I fixed the issue with TwoDimTable and fixed some other minor bugs.  Your code should run to completion now.  Please cherry-pick my branch here: km2018-PUBDEV-7367-GAM-gridsearch-karthik-2Dtable

Also, you will need to rebase to rel-zeno.  I pushed some new changes that will affect your tests.

h2o-core/src/main/java/hex/grid/HyperSpaceSearchCriteria.java
```diff
@@ -58,7 +58,7 @@ public double getStoppingTolerance() {
 
   public final Strategy _strategy;
   public final Strategy strategy() { return _strategy; }
-
+  
   public StoppingCriteria stoppingCriteria() { return null; }
```
wendycwong, line 62: Karthik:

Please remove this formatting as it serves no purpose.

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7367_gridsearch_constraints.py
```diff
@@ -0,0 +1,106 @@
...
+    search_criteria = {'strategy': 'Cartesian'}
+    hyper_parameters = {'alpha': [0.9, 0.01],
+                        'constraints': [
+                            {'lambda_search': True, 'nlambdas': [5, 10, 15], 'gam_columns': [["C11", "C12", "C13"]]},
+                            {'scale': [[1, 1, 1], [2, 2, 2]],
+                             'num_knots': [[5, 5, 5], [5, 6, 7]],
+                             'gam_columns': [["C11", "C12", "C13"]]},
+                            {'scale': [[1, 1], [2, 2]],
+                             'num_knots': [[5, 5], [6, 6]],
+                             'gam_columns': [["C11", "C12"], ["C12", "C13"]]}]}
```
wendycwong, line 30: Karthik:  Please add one more constraints:
{'lambda_search':True, 'nlambdas':[5,10,15]}

You will need to change the maximum number of models as well.

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7367_random_gridsearch_constraints.py
```diff
@@ -0,0 +1,93 @@
...
+    search_criteria = {'strategy': 'RandomDiscrete', "max_models": 24, "seed": 1}
+    hyper_parameters = {'lambda': [0, 0.01],
+                        'constraints': [{'scale': [[1, 1, 1], [2, 2, 2]],
+                                         'num_knots': [[5, 5, 5], [5, 6, 7]],
+                                         'gam_columns': [["C6", "C7", "C8"]]},
+                                        {'scale': [[1, 1], [2, 2]],
+                                         'num_knots': [[5, 5], [6, 6]],
+                                         'gam_columns': [["C6", "C7"], ["C7", "C8"]]}]}
+    manual_gam_models = []
+    num_grid_models = 0
```
wendycwong, line 30: Karhtik:  Please add one more constraints:
{'lambda_search':False, 'alpha':[0.9,0.1]}

Remember to fix the maximum number of models expected.

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7367_random_gridsearch_invalid.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import division
```
wendycwong, line 1: Karthik:  Please add one more invalid test with you adding gam_columns to both the hyper-parameters and the hyper_parameters['constraints'] to make sure it will trigger your new error conditions.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -240,28 +256,69 @@ protected MP getModelParams(MP params, Object[] hyperParams) {
...
       return hypers;
     }
 
-    protected int integerHash(int[] ar) {
-      Integer[] hashMe = new Integer[ar.length];
+    protected int integerHash(Map<String, Object[]> hyperParams, String[] hyperParamNames, int[] ar, int constraintNum) {
+      Integer[] hashMe = new Integer[ar.length + 1];
       for (int i = 0; i < ar.length; i++)
-        hashMe[i] = ar[i] * _hyperParams.get(_hyperParamNames[i]).length;
+        hashMe[i] = ar[i] * hyperParams.get(hyperParamNames[i]).length;
```
wendycwong, line 320: Karthik:

In order to test this hash function implementation, can you add a test with constraints containing 5 arrays but with each array containing exactly the same model parameters?  Something like this:

{'scale': [[1, 1], [2, 2]], 'num_knots': [[5, 5], [6, 6]], 'gam_columns': [["C11", "C12"], ["C12", "C13"]]}

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -191,9 +191,22 @@ public BaseWalker(MP params,
...
+        for (int cNum = 0; cNum < constraintsNumber; cNum++) {
+          Map<String, Object[]> constraints = (Map<String, Object[]>)_hyperParams.get("constraints")[cNum];
+          for (Map.Entry<String, Object[]> p : constraints.entrySet()) {
+            constraintHyperParams.add(p.getKey());
+          }
+        }
+        _hyperParamNamesConstraint = new String[constraintHyperParams.size()];
+        constraintHyperParams.toArray(_hyperParamNamesConstraint);
+      } else 
+        _hyperParamNamesConstraint = new String[0];
```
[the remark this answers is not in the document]
sebhrusen: Also, I would suggest that you remove the groups/constraints from `_hyperParams` immediately here in the constructor. This would avoid you to have to handle this reserved key multiple times in the code later.

We can and should separate the representation at client API level which tries to be convenient for the end user (so that he can see all params at once), from the internal representation that must focus on robustness.

h2o-core/src/main/java/water/api/GridSearchHandler.java
```diff
@@ -140,11 +140,11 @@ protected void validateHyperParams(P params, Map<String, Object[]> hyperParams)
           break;
         }
       }
-      if (fieldMetadata == null) {
+      if (fieldMetadata == null && !hparam.getKey().equals("constraints")) {
```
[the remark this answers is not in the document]
sebhrusen: also, given the type issues with the Maps in `HyperspaceWalker`: if by any chance you get arrays converted into lists, this is probably happening in this class...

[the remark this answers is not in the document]
wendycwong: @sebhrusen The name constraints are descriptive of what is going on.  Basically, we are dividing the model parameters into two sets which can iterate among themselves freely but not with the other set.

h2o-core/src/main/java/water/api/GridSearchHandler.java
```diff
@@ -140,11 +142,11 @@ protected void validateHyperParams(P params, Map<String, Object[]> hyperParams)
           break;
         }
       }
-      if (fieldMetadata == null) {
+      if (fieldMetadata == null && !hparam.getKey().equals(CONSTRAINTS)) {
         throw new H2OIllegalArgumentException(hparam.getKey(), "grid",
                                               "Unknown hyper parameter for grid search!");
       }
-      if (!fieldMetadata.is_gridable) {
+      if (!hparam.getKey().equals(CONSTRAINTS) && !fieldMetadata.is_gridable) {
```
sebhrusen, line 149: for simplicity and also to add proper validation of the "constraints", I'd suggest sth like the following:
```
Set<String> allKeys = hyperParams.entrySet();
allKeys.remove(CONSTRAINTS);
for (Map.Entry<String, Object[]> hparam : allKeys) {
    ... // no need to test for CONSTRAINTS here anymore
}
if (hyperParams.contains(CONSTRAINTS)) {
    // now you get nested validation for free
    validateHyperParams(params, hyperParams.get(CONSTRAINTS));
}
```

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -147,7 +148,10 @@ public boolean stopEarly(Model model, ScoringInfo[] sk) {
      * Cached names of used hyper parameters.
      */
     final protected String[] _hyperParamNames;
+    
+    final protected String[] _hyperParamNamesConstraint;  // model parameters specified in constraints of hyper parameters
 
+    protected Map<String, Object>[] _hyperParamConstraints;
```
sebhrusen, line 154: if these are always lists (probably due to parser), let's at least force the type to `List` instead of `Object`, your code below will also become simpler.

But let's ask more: we want to have `Map<String, Object[]>[] _hyperParamConstraints;` here, same type as for `hyperParams`.
This would allow us to use the same logic everywhere we can use `hyperParams`: a typical example would be to support nested grouping later.
To enforce `Map<String, Object[]>[] _hyperParamConstraints`, this should be done in `GridSearchSchema#fillFromParms`: 
```
        // Convert lists and singletons into arrays
        for (Map.Entry<String, Object> e : m.entrySet()) {
          Object o = e.getValue();
          Object[] o2 = o instanceof List ? ((List) o).toArray() : new Object[]{o};

          hyper_parameters.put(e.getKey(),o2);
        }
```

This code above could be adapted to handle `CONSTRAINTS` key differently:
```
private static Map<String, Object[]> paramValuesToArray(Map<String, Object> params) {
        Map<String, Object[]> result = new HashMap<>();
        for (Map.Entry<String, Object> e : m.entrySet()) {
          String k = e.getKey();
          Object v = e.getValue();
          Object[] arr = CONSTRAINTS.equals(k) ? paramValuesToArray(v)
                           : v instanceof List ? ((List) v).toArray() 
                           : new Object[]{v};
          result.put(k, v);
        }
        return result;
```
Now we have arrays everywhere, and besides it handles nested constraints.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -276,33 +331,55 @@ private void validateParams() {
...
+          if (_hyperParams.get(key) != null) {
+            throw new H2OIllegalArgumentException("Grid search model parameter '" + key + "' is set in " +
+                    "both the constraints and in the hyperparameters map.  This is ambiguous; set it in one place" +
+                    " or the other, not both.");
+          }
+          
+          validateParamVals(key);
+        }
+      }
+    }
```
sebhrusen, line 355: if I'm not mistaken, once you have `_hyperParamConstraints` with the correct type, you can then reuse the existing logic validating `hyperParams` to validate the constraints: 
```
private void validateParams() {
   validateParams(_hyperParams);
   validateParams(_hyperParamsConstraints);
}
```

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
km2018: Thank you for this suggestion - reduced code in the hyperspace walker and improved the design. Very elegant solution.

sebhrusen:
Hi @km2018 , before I do a last review, can you please do a quick rename?
We agreed internally that we would prefer to have the special hyperparameter key being renamed from `constraints` to `subspaces`. Can you please do that quickly? And ideally, please also rename the variables on Java side to avoid confusion: the `___Constraint` variables.
Thanks!

km2018:
> Hi @km2018 , before I do a last review, can you please do a quick renaming?
> We agreed internally that we would prefer to have the special hyperparameter key being renamed from `constraints` to `subspaces`. Can you please do that quickly? And ideally, please also rename the variables on Java side to avoid confusion: the `___Constraint` variables.
> Thanks!

Just took care of this - let me know if you find anything else. Thanks.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -240,69 +266,99 @@ protected MP getModelParams(MP params, Object[] hyperParams) {
...
+      for (String key : params.keySet()) {
         // Throw if the user passed an empty value list:
-        Object[] values = _hyperParams.get(key);
+        Object[] values = params.get(key);
         if (0 == values.length)
           throw new H2OIllegalArgumentException("Grid search hyperparameter value list is empty for hyperparameter: " + key);
 
         if ("seed".equals(key) || "_seed".equals(key)) continue;  // initialized to the wall clock
 
+        if (isSubspace && _hyperParams.get(key) != null) {
```
sebhrusen, line 325: except if you allow null values in your maps and give it a specific meaning, it's usually preferable to use `map.containsKey(key)` to `map.get(key) != null`.
Not extremely important though...

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -240,69 +266,99 @@ protected MP getModelParams(MP params, Object[] hyperParams) {
...
+      for (Object[] p : _hyperParams.values()) {
+        free_param_combos *= p.length;
       }
+
+      work = work == 0 ? free_param_combos : free_param_combos * work;
+
       return work;
     }
 
+    protected Map<String, Object[]> mergeHashMaps(Map<String, Object[]> hyperparams, Map<String, Object[]> subspaces) {
```
sebhrusen, line 289: we defined:
```java
protected Map<String, Object[]>[] _hyperParamSubspaces;
```
so, please try to keep plural for array, and singular for elements of the array: here 
```java
Map<String, Object[]> subspace
```
small details, but when you read code on and on, it saves some brain processing :)

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -222,14 +234,28 @@ public MP getParams() {
...
     }
-
-    protected MP getModelParams(MP params, Object[] hyperParams) {
+    
+    private Map<String, Object[]>[] extractSubspaces() {
+      if(_hyperParams.get(SUBSPACES) == null) { return new Map[0]; }
+      Map<String, Object[]>[] subspaces = new Map[_hyperParams.get(SUBSPACES).length];
+      for (int i = 0; i < _hyperParams.get(SUBSPACES).length; i++) {
+        subspaces[i] = (Map<String, Object[]>)_hyperParams.get(SUBSPACES)[i];
+      }
```
sebhrusen, line 243: not sure you still need to do that now that you ensured that subspaces value is of the correct type.
I think
```java
return (Map<String, Object[]>[])_hyperParams.get(SUBSPACES);
```
may work just fine.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -191,22 +195,30 @@ public BaseWalker(MP params,
...
     } // BaseWalker()
 
     @Override
     public String[] getHyperParamNames() {
       return _hyperParamNames;
     }
+    
+    public String[] getHyperParamNamesSubspace() {
+      return _hyperParamNamesSubspace;
+    }
```
sebhrusen, line 221: for clarity, I would rename this method and attribute to 
`getAllHyperParamNamesInSubspaces` / `_allHyperParamNamesInSubspaces`

sebhrusen (commented):
made several suggestions, mainly for polishing, otherwise it looks great: thx for this effort!

Please don't forget to rename the `constraints` key in your tests: they should be failing currently.

Also we probably want to have at least a basic test for the R client, just to stay on the safe side.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
km2018: Casting directly seemed to throw an error, so I used
```
return Stream.of(_hyperParams.get(SUBSPACES)).toArray(Map[]::new);
```
If you think there's a better way to do this, let me know.

PR #4743, opened 2020-06-24
[PUBDEV-7467] external XGBoost Steam integration

honzasterba:


h2o-extensions/xgboost/testMultiNode.sh
```diff
@@ -100,9 +100,9 @@ echo $DOONLY > $OUTDIR/tests.doonly.txt
 # Launch 3 helper JVMs.  All output redir'd at the OS level to sandbox files.
 CLUSTER_NAME=junit_cluster_$$
 CLUSTER_BASEPORT=44000
-$JVM water.H2O -name $CLUSTER_NAME -ip $H2O_NODE_IP -baseport $CLUSTER_BASEPORT -ga_opt_out $SSL 1> $OUTDIR/out.1 2>&1 & PID_1=$!
-$JVM water.H2O -name $CLUSTER_NAME -ip $H2O_NODE_IP -baseport $CLUSTER_BASEPORT -ga_opt_out $SSL 1> $OUTDIR/out.2 2>&1 & PID_2=$!
-$JVM water.H2O -name $CLUSTER_NAME -ip $H2O_NODE_IP -baseport $CLUSTER_BASEPORT -ga_opt_out $SSL 1> $OUTDIR/out.3 2>&1 & PID_3=$!
+$JVM water.H2OApp -name $CLUSTER_NAME -ip $H2O_NODE_IP -baseport $CLUSTER_BASEPORT -ga_opt_out $SSL 1> $OUTDIR/out.1 2>&1 & PID_1=$!
```
honzasterba, line 103: Since REST API is required for multinode XGBoost I had to switch this to H2OApp to enable it, this is non standard @Pscheidl @michalkurka do you see a better solution?

Pscheidl: I don't see any problem in what you did whatsoever.

h2o-extensions/xgboost/build.gradle
```diff
@@ -27,6 +28,7 @@ dependencies {
     compile "ai.h2o:xgboost4j-linux-minimal:${xgboost4jVersion}"
     compile "ai.h2o:xgboost4j-osx-minimal:${xgboost4jVersion}"
 
+    testCompile project(":h2o-app")
```
michalkurka, line 31: This gives a cycle in dependencies (if we ignore that it is a dependency for test).

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -345,14 +347,18 @@ final void buildModel() {
...
+        return SteamExecutorStarter.getInstance().getRemoteExecutor(model, _train);
       } else {
-        String userName = H2O.getSysProperty("xgboost.external.user", null);
-        String password = H2O.getSysProperty("xgboost.external.password", null);
-        return new RemoteXGBoostExecutor(propVal, model, _train, userName, password);
+        String remoteUriFromProp = H2O.getSysProperty("xgboost.external.address", null);
+        if (remoteUriFromProp == null) {
+          return new LocalXGBoostExecutor(model, _train);
+        } else {
+          String userName = H2O.getSysProperty("xgboost.external.user", null);
```
michalkurka, line 358: Why is password in system properties? Can this be avoided? Is it passed using `-D` and visible eg. in process list?

h2o-extensions/xgboost/testMultiNode.sh
michalkurka: @honzasterba I don't like the dependency on h2o-app; imho it would be better to avoid that because that brings in other dependencies

IMHO it would be better to use `H2OTestNodeStarter` instead of H2OApp, it will do the same thing

h2o-test-support/src/main/java/water/test/WebsocketClient.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public Map<String, String> waitToReceiveMessage(String message) {
+        return waitToReceiveMessage(message, 10_000);
+    }
+    
+    public Map<String, String> waitToReceiveMessage(String message, int timeout) {
+        return waitToReceiveMessage(message, timeout, true);
+    }
+
+    public synchronized Map<String, String> waitToReceiveMessage(String message, int timeout, boolean failOnNone) {
```
Pscheidl, line 57: If one client can be reused at  multiple places, won't there be a race condition with potential loss of data ? When two threads send a message using the same client (why would the `waitToReceiveMessage` method be synchronized otherwise) at once and then wait for the response, only one response is reacted to.

I bent one of the tests to demonstrate:

```java
    @Test
    public void testSendTwoMessages() throws Exception {
        final WebsocketClient steam = new WebsocketClient();
        try {
            Map<String, String> helloMsg = new HashMap<>();
            helloMsg.put("_type", "hello");
            helloMsg.put("_id", "123456");

            Map<String, String> byeMsg = new HashMap<>();
            byeMsg.put("_type", "bye");
            byeMsg.put("_id", "12345678");

            steam.sendMessage(byeMsg);
            steam.sendMessage(helloMsg);
            System.out.println("Hello message sent.");
            Map<String, String> response = steam.waitToReceiveMessage("hello response");
            Map<String, String> response1 = steam.waitToReceiveMessage("bye response");
            assertEquals("hello_response", response.get("_type"));
            assertEquals("123456_response", response.get("_id"));
            assertEquals(H2O.ABV.projectVersion(), response.get("version"));
            assertEquals(H2O.ABV.branchName(), response.get("branch"));
            assertEquals(H2O.ABV.lastCommitHash(), response.get("hash"));
            assertEquals(String.valueOf(H2O.CLOUD.size()), response.get("cloud_size"));
        } finally {
            steam.close();
        }
    }
```

I'd personally like to have this scenario tested as well (ability of H2O to respond to multiple messages at time).

h2o-test-support/src/main/java/water/test/WebsocketClient.java
```diff
@@ -0,0 +1,72 @@
...
+        ClientManager client = ClientManager.createClient();
+        ClientEndpointConfig cec = ClientEndpointConfig.Builder.create().build();
+        sess = client.connectToServer(this, cec, new URI(destUri));
+    }
+    
+    public void close() throws IOException {
+        sess.close(new CloseReason(CloseReason.CloseCodes.NORMAL_CLOSURE, "Test Done"));
+    }
+
+    @OnOpen
```
Pscheidl, line 34: This annotation is inappropriate here. There is nothing to process it. I'd suggest putting `@Override` there instead.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/remote/SteamExecutorStarter.java
```diff
@@ -0,0 +1,169 @@
...
+        } else {
+            throw new IllegalStateException("Failed to start external cluster: " + response.get("reason"));
+        }
+    }
+
+    private RemoteXGBoostExecutor makeExecutor(XGBoostModel model, Frame train) {
+        return new RemoteXGBoostExecutor(model, train, cluster.uri, cluster.userName, cluster.password);
+    }
+    
+    private Map<String, String> waitForResponse(Map<String, String> startRequest) {
```
koniecsveta, line 85: is it possible that steam won't respond on start message and thread will be hanging there forever?

h2o-extensions/steam/src/main/java/hex/steam/SteamHelloMessenger.java
```diff
@@ -23,12 +23,12 @@ public void onMessage(Map<String, String> message) {
         if (message.get(TYPE).equals("hello")) {
             assert sender != null : "Received message but sender is null";
             Map<String, String> response = new HashMap<>();
-            response.put(TYPE, "cloud_info");
+            response.put(TYPE, "hello_response");
```
koniecsveta, line 26: event type hello_response might be stored as a constant, because it is used multiple times

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
honzasterba: This is a way to use external cluster used mainly for testing, system properties can also be set via Rapids.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/remote/SteamExecutorStarter.java
honzasterba: yes, although unlikely, will make this escapable by canceling the xgboost job

h2o-test-support/src/main/java/water/test/WebsocketClient.java
honzasterba: the annotation is processed by Tyrus, this method does not override anything, its called based on the annotation

h2o-test-support/src/main/java/water/test/WebsocketClient.java
```diff
@@ -0,0 +1,72 @@
...
+            }
+        });
+    }
+    
+    public void sendMessage(Object msg) throws IOException {
+        sess.getBasicRemote().sendText(gson.toJson(msg));
+    }
+
+    public Map<String, String> waitToReceiveMessage(String message) {
+        return waitToReceiveMessage(message, 10_000);
```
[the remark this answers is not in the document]
honzasterba: you can call waitToReceiveMessage variant which also takes the timeout, this is the default

h2o-test-support/src/main/java/water/test/WebsocketClient.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+    
+    public Map<String, String> waitToReceiveMessage(String message, int timeout) {
```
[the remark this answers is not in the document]
honzasterba: it is used in SteamExecutorStarterTest

h2o-test-support/src/main/java/water/test/WebsocketClient.java
```diff
@@ -0,0 +1,72 @@
...
+            Thread.currentThread().interrupt();
+        }
+        if (receivedMessage != null) {
+            Map<String, String> res = receivedMessage;
+            receivedMessage = null;
+            return res;
+        } else if (failOnNone) {
+            fail("Expected " + message + ", but no message received from H2O.");
+        }
+        return null;
```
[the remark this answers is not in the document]
honzasterba: fixed

h2o-extensions/steam/src/main/java/hex/steam/SteamHelloMessenger.java
honzasterba: in code its used only here, its better in tests to use the same string literal

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/remote/SteamExecutorStarter.java
```diff
@@ -0,0 +1,169 @@
...
+        } else {
+            throw new IllegalStateException("Failed to start external cluster: " + response.get("reason"));
+        }
+    }
+
+    private RemoteXGBoostExecutor makeExecutor(XGBoostModel model, Frame train) {
+        return new RemoteXGBoostExecutor(model, train, cluster.uri, cluster.userName, cluster.password);
+    }
+    
+    private Map<String, String> waitForResponse(Map<String, String> startRequest, Job<XGBoostModel> job) {
```
koniecsveta, line 85: is it possible that waitForResponse() will be called before queueResponseMessage?

honzasterba: yes, in fact that always happens, that's why I have the while loop waiting for queueResponseMessage to put the response into received messages

koniecsveta: I was thinking whether there could be possible deadlock because both methods have synchronized block on the same object but it was only my confusion how wait and notify works

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/remote/SteamExecutorStarter.java
```diff
@@ -0,0 +1,169 @@
...
+                sender.sendMessage(message);
+            } else {
+                throw new IOException("Steam communication chanel is not open.");
+            }
+        }
+    }
+
+    @Override
+    public void onMessage(Map<String, String> message) {
+        if (message.get(TYPE).equals("stopXGBoostClusterNotification")) {
```
koniecsveta, line 118: this might be like "stopXGBoostClusterNotification".equals(message.get(TYPE)) to prevent NPE

honzasterba: you're right, thats the better way to do it even though type is guaranteed to be non-null by onMessage contract

PR #4744, opened 2020-06-24
 PUBDEV-7367 gam random gridsearch karthik

wendycwong:
This is only used to perform Jenkins test for Karthik PR: https://github.com/h2oai/h2o-3/pull/4742


wendycwong:
Merged in another PR.

PR #4746, opened 2020-06-25
PUBDEV-7641: Allow to specify labeled validation frame in Isolation Forest

michalkurka:


michalkurka:
@Pscheidl this PR needs more work but early feedback is appreciated!

h2o-algos/src/main/java/hex/tree/Score.java
```diff
@@ -51,23 +51,31 @@ private Score(SharedTree bldr, boolean is_train, ScoreIncInfo sii, boolean oob,
 
   @Override public void map(Chunk allchks[]) {
     final Chunk[] chks = getScoringChunks(allchks);
-    Chunk ys = _bldr.isSupervised() ? _bldr.chk_resp(chks) : new C0DChunk(0, chks[0]._len);  // Response
+    Chunk ys; // Response
```
Pscheidl, line 54:
```suggestion
    final Chunk ys; // Response
```

h2o-algos/src/main/java/hex/tree/isofor/IsolationForest.java
```diff
@@ -71,7 +74,27 @@
...
+      }
+    }
+    if (expensive) {
+      if (vresponse() != null) {
+        if (!vresponse().isBinary() || vresponse().domain()==null) {
+          error("_response_column", "The response column of the validation frame needs to have a binary categorical domain (not anomaly/anomaly).");
+        }
+      }
+      if (response() != null) {
+        error("_training_frame", "Training frame should not have a response column");
```
Pscheidl, line 96: This leads me back to my comment in `IsolationForestV3` about the parameter naming. Naming it differently, e.g. `validation_response_column` would send out a clear message the parameter has nothing to do with training phase.

Also, if the training data contains the column, shouldn't we just ignore it ? Just a question we should have answered before commiting to this. It's probably better to fail and let the user know, so I vote for keeping the error in place.

h2o-algos/src/main/java/hex/tree/isofor/IsolationForest.java
```diff
@@ -71,7 +74,27 @@
...
+      } else if (expensive && vresponse() == null) {
+        error("_response_column", "Validation frame is missing response column `" + _parms._response_column + "`.");
+      }
+      if (_parms._contamination > 0) {
+        error("_contamination", "Contamination parameter cannot be used together with a validation frame.");
+      }
+    }
+    if (expensive) {
+      if (vresponse() != null) {
+        if (!vresponse().isBinary() || vresponse().domain()==null) {
```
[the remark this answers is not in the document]
michalkurka: that works, thank you @Pscheidl !

PR #4748, opened 2020-06-26
PUBDEV-6919 use weights in coxph

satai:


h2o-r/tests/testdir_algos/coxph/runit_coxph_predict.R
```diff
@@ -41,6 +41,58 @@ test.CoxPH.predict <- function() {
     hex.lp.3 <- pred(hex.fit.3, tstdata.hex.3)
 
     expect_equal(fit.3$linear.predictors, hex.lp.3, scale = 1, tolerance = 1e-3)
+    
+    # Case 4: only numeric, same weights
```
Pscheidl, line 45: This could be a simple JUnit test as well.

PR #4749, opened 2020-06-26
PUBDEV-6968 add python test that covers stratification

satai:
WIP, cc @michalkurka @honzasterba @Pscheidl 

It seems there is an problem with strata in R too :-/ (case 8)

The case #8 should give the same results, shouldn't it?

### DIFF
--- a/h2o-py/tests/testdir_algos/coxph/pyunit_coxph_smoke_stratify.py
+++ b/h2o-py/tests/testdir_algos/coxph/pyunit_coxph_smoke_stratify.py
@@ -1,0 +1,27 @@
+from __future__ import print_function
+import sys
+sys.path.insert(1,"../../../")
+import h2o
+from tests import pyunit_utils
+from h2o.estimators.coxph import H2OCoxProportionalHazardsEstimator
+
+
+def coxph_smoke_strata():
+    heart = h2o.import_file(pyunit_utils.locate("smalldata/coxph_test/heart.csv"))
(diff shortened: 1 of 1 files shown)


h2o-py/tests/testdir_algos/coxph/pyunit_coxph_smoke_stratify.py
```diff
@@ -0,0 +1,27 @@
...
+    heart['transplant'] = heart['transplant'].asfactor()
+    coxph = H2OCoxProportionalHazardsEstimator(stratify_by = ["transplant"], start_column="start", stop_column="stop")
+    coxph.train(x=["age", "surgery", "transplant"], y="event", training_frame=heart)
+
+    assert coxph.model_id != ""
+    assert coxph.formula() == "Surv(start, stop, event) ~ age + surgery + strata(transplant)", \
+        "Expected formula to be 'Surv(start, stop, event) ~ age + surgery + strata(transplant)' but it was " + coxph.formula()
+
+    pred = coxph.predict(test_data=heart)
+    assert len(pred) == len(heart)
```
honzasterba, line 20: maybe to test this a bit more I would suggest to train two models one with stratify_by one without and compare the preds are different

satai: Stratification results are part of an other PR, comparing with R as reference.

The original claim in PUBDEV-6968 was crash when used in python.

PR #4750, opened 2020-06-26
PUBDEV-7645: Added warning if user enable lambda_search and specify lambda-value

wendycwong:
This PR finish the work specified in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7645

I added the warning in init() of GLM and a pyunit test to make sure the warning is there.

### DIFF
--- a/h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7645_lambda_search_warning.py
+++ b/h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7645_lambda_search_warning.py
@@ -1,0 +1,35 @@
+from __future__ import print_function
+
+import warnings
+from builtins import range
+import sys, os
+sys.path.insert(1,"../../../")
+import h2o
+from h2o.estimators.glm import H2OGeneralizedLinearEstimator
+from tests import pyunit_utils
+
+# This method test to make sure when a user set lambda_search=True and choose a lambda value,  a warning message 
+# should be generated and passed to the user.
+def test_lambda_warning():
+    training_data = h2o.import_file(pyunit_utils.locate("smalldata/gbm_test/BostonHousing.csv"))
+    Y = 13
+    X = list(range(13))
+    model = H2OGeneralizedLinearEstimator(family="Gaussian", lambda_search=True, Lambda=[0.01])
+    model.train(x=X, y=Y, training_frame=training_data)
+
+    for v in sys.modules.values():
+        if getattr(v, '__warningregistry__', None):
+            v.__warningregistry__ = {}
+    with warnings.catch_warnings(record=True) as ws:
+        # Get all RuntimeWarnings, copy from TomasF, thanks!
+        warnings.simplefilter("always", RuntimeWarning)
+        model = H2OGeneralizedLinearEstimator(family="Gaussian", lambda_search=True, Lambda=[0.01])
+        model.train(x=X, y=Y, training_frame=training_data)
+
+        assert any((issubclass(w.category, RuntimeWarning) and 'disabled when user specified any lambda value(s)' in str(w.message) 
+                    for w in ws))
+
+if __name__ == "__main__":
+    pyunit_utils.standalone_test(test_lambda_warning)
+else:
+    test_lambda_warning()
--- a/h2o-algos/src/main/java/hex/glm/GLMModel.java
+++ b/h2o-algos/src/main/java/hex/glm/GLMModel.java
@@ -250,6 +250,8 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
     public void validate(GLM glm) {
       if (_solver.equals(Solver.COORDINATE_DESCENT_NAIVE) && _family.equals(Family.multinomial))
         throw H2O.unimpl("Naive coordinate descent is not supported for multinomial.");
+      if ((_lambda != null) && _lambda_search)
+        glm.warn("lambda_search", "disabled when user specified any lambda value(s).");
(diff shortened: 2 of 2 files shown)


wendycwong:
@michalk:  However, the warning somehow was not passed over to the python client.  Do you know why off the top of your head?  If not I can debug it.

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7645_lambda_search_warning.py
```diff
@@ -0,0 +1,53 @@
...
+    check_warnings("lambda_search:  will be disabled if you specify lambda values.", X, Y, 1, training_data)   # expect no warning here
+
+def check_warnings(model_id, X, Y, warnNumber, training_data):
+    buffer = StringIO() # redirect output
+    sys.stderr=buffer
+    model = H2OGeneralizedLinearEstimator(family="Gaussian", lambda_search=True, Lambda=[0.01])
+    model.train(x=X, y=Y, training_frame=training_data)
+    warn_phrase = "UserWarning"
+    warn_string_of_interest = "slash (/) found"
+    sys.stderr=sys.__stderr__   # redirect printout back to normal path
```
honzasterba, line 29: should be in finally

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7645_lambda_search_warning.py
```diff
@@ -0,0 +1,53 @@
...
+from tests import pyunit_utils
+
+try:    # redirect python output
+    from StringIO import StringIO  # for python 3
+except ImportError:
+    from io import StringIO  # for python 2
+
+# This method test to make sure when a user set lambda_search=True and choose a lambda value,  a warning message 
+# should be generated and passed to the user.
+def test_benign():
```
honzasterba, line 16: is this a type or does `benign` has some meaning?

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7645_lambda_search_warning.py
```diff
@@ -0,0 +1,53 @@
...
+    warn_string_of_interest = "slash (/) found"
+    sys.stderr=sys.__stderr__   # redirect printout back to normal path
+
+    try:        # for python 2.7
+        assert len(buffer.buflist)==warnNumber
+        if len(buffer.buflist) > 0:  # check to make sure we have the right number of warning
+            for index in range(len(buffer.buflist)):
+                print("*** captured warning message: {0}".format(buffer.buflist[index]))
+                assert (warn_phrase in buffer.buflist[index]) and (warn_string_of_interest in buffer.buflist[index])
+    except:     # for python 3.
```
honzasterba, line 37: this will catch an assertion error as well, am I right?

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7645_lambda_search_warning.py
wendycwong: @honza:

This code is trying to catch the warning phrase.  Therefore, it is not in finally.

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7645_lambda_search_warning.py
wendycwong: The try and except is to process warning message from python 2 or 3.  The try part will capture the warning message for python 2 and the except will capture the warning message for Python 3.

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7645_lambda_search_warning.py
wendycwong: tumor type not cancerous.

wendycwong: benign (good) or malignant (bad)

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_7645_lambda_search_warning.py
tomasfryda: I think @honzasterba is right about the `try` `except` catching the assert. Instead, I would use `except Warning:`.

Although, now I struggle to understand the logic of the `check_warnings` function. It looks to me that it tries to get everything that goes to stderr, then it goes through all the buffered strings in `try` `except` which I don't see a reason to do.

`try` `except` will catch all exceptions (even stuff like `sys.exit()`) but I don't see anything except the assertions that should throw an exception (the warnings are already just buffered strings). 

I use the following snippet and it seems to work well in both Py27 and Py3:

```
    for v in sys.modules.values():
         if getattr(v, '__warningregistry__', None):
             v.__warningregistry__ = {}
     with warnings.catch_warnings(record=True) as ws:
         # Get all UserWarnings
         warnings.simplefilter("always", RuntimeWarning)
         # Do something that might throw some warnings
         se.train(x=x, y=y, training_frame=train)
         assert any((
             issubclass(w.category, RuntimeWarning) and
             'DESIRED WARNING' in str(w.message)
             for w in ws
         ))
```

wendycwong: @tomasfryda Thanks Tomas.  Let me try it out.

PR #4754, opened 2020-06-30
[R] remove links to sparkling-water managed demo files

honzasterba:


h2o-py/demos/word2vec_craigslistjobtitles.ipynb
```diff
@@ -196,4 +196,4 @@
  },
  "nbformat": 4,
  "nbformat_minor": 0
-}
+}
```
jakubhava, line 199: NIT: missing newline. 

Just an idea: In Sparkling Water we are using https://github.com/diffplug/spotless/tree/main/plugin-gradle. We are using it for full formatting of Java/Scala/Gradle code, but we are also using it to automatically add new lines to the files with specified extensions.

It might be overkill for H2O, but just mentioning it. It helped is in SW as we no longer need to spend time on formatting during reviews

PR #4756, opened 2020-06-30
PUBDEV-7185: added MOJO to GAM

wendycwong:
This PR completes the task in JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7185?filter=myopenissues&orderby=priority%20DESC

In particular, I want to bring you to the following:
The coefficients generated from GLM are centered and hence to perform scoring, the following process is needed:
raw dataset -> add gam columns -> center the gam columns (multiply gam columns with a z matrix) -> use the coefficients from GLM to perform scoring.

However, I feel that the following will be faster for mojo scoring:
- transform the centered coefficients from GLM to non-centered coefficients by multiplying it with a z matrix and this is done only once in the mojo writer.

Then, during mojo scoring, we only need to do the following:
- raw dataset -> add gam columns -> use non-centered coefficients to perform scoring.

I have opened a JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7644?filter=myopenissues&orderby=priority%20DESC to improve scoring speed by using the non-centered coefficients.  By I am not sure it will make it to this release.

1. Java backend to support GAM Mojo;
2. Java unit test.
3.  Python unit tests
4. R unit tests.

### DIFF
--- a/h2o-algos/src/main/java/hex/gam/GAMModel.java
+++ b/h2o-algos/src/main/java/hex/gam/GAMModel.java
@@ -6,7 +6,6 @@ import hex.gam.MatrixFrameUtils.AddGamColumns;
 import hex.glm.GLM;
 import hex.glm.GLMModel;
 import hex.glm.GLMModel.GLMParameters.Family;
-import hex.glm.GLMModel.GLMParameters.GLMType;
 import hex.glm.GLMModel.GLMParameters.Link;
 import hex.glm.GLMModel.GLMParameters.Solver;
 import water.*;
@@ -15,10 +14,7 @@ import water.fvec.Frame;
 import water.fvec.NewChunk;
 import water.fvec.Vec;
 import water.udf.CFuncRef;
-import water.util.ArrayUtils;
-import water.util.FrameUtils;
-import water.util.Log;
-import water.util.TwoDimTable;
+import water.util.*;
 
 import java.io.Serializable;
 import java.util.Arrays;
@@ -28,9 +24,11 @@ import static hex.gam.MatrixFrameUtils.GamUtils.sortCoeffMags;
 import static hex.glm.GLMModel.GLMParameters.MissingValuesHandling;
 
 public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.GAMModelOutput> {
+  private static final String[] BINOMIAL_CLASS_NAMES = new String[]{"0", "1"};
   public String[][] _gamColNamesNoCentering; // store column names only for GAM columns
   public String[][] _gamColNames; // store column names only for GAM columns after decentering
   public Key<Frame>[] _gamFrameKeysCenter;
+  public double[] _gamColMeans;
   public int _nclass; // 2 for binomial, > 2 for multinomial and ordinal
   public double[] _ymu;
   public long _nobs;
@@ -46,8 +44,12 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
   
   @Override public ModelMetrics.MetricBuilder makeMetricBuilder(String[] domain) {
     if (domain==null && (_parms._family==Family.binomial || _parms._family==Family.quasibinomial || 
-            _parms._family==Family.negativebinomial || _parms._family==Family.fractionalbinomial))
-      domain = new String[]{"0","1"};
+            _parms._family==Family.negativebinomial || _parms._family==Family.fractionalbinomial)) {
+      if (_parms._family == Family.fractionalbinomial)
+        domain = BINOMIAL_CLASS_NAMES;
+      else
+        domain = _output._responseDomains;
+    }
     GLMModel.GLMWeightsFun glmf = new GLMModel.GLMWeightsFun(_parms._family, _parms._link, _parms._tweedie_variance_power,
             _parms._tweedie_link_power, _parms._theta);
     return new MetricBuilderGAM(domain, _ymu, glmf, _rank, true, _parms._intercept, _nclass);
@@ -220,29 +222,19 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
     public double _tweedie_variance_power;
     public double _tweedie_link_power;
     public double _theta; // 1/k and is used by negative binomial distribution only
-    public double _invTheta;
     public double [] _alpha;
     public double [] _lambda;
     public Serializable _missing_values_handling = MissingValuesHandling.MeanImputation;
-    public double _prior = -1;
     public boolean _lambda_search = false;
-    public int _nlambdas = -1;
-    public boolean _non_negative = false;
-    public boolean _exactLambdas = false;
-    public double _lambda_min_ratio = -1; // special
     public boolean _use_all_factor_levels = false;
     public int _max_iterations = -1;
     public boolean _intercept = true;
     public double _beta_epsilon = 1e-4;
     public double _objective_epsilon = -1;
-    public double _gradient_epsilon = -1;
     public double _obj_reg = -1;
     public boolean _compute_p_values = false;
-    public boolean _remove_collinear_columns = false;
     public String[] _interactions=null;
     public StringPair[] _interaction_pairs=null;
-    public boolean _early_stopping = true;
-    public Key<Frame> _beta_constraints = null;
     public Key<Frame> _plug_values = null;
     // internal parameter, handle with care. GLM will stop when there is more than this number of active predictors (after strong rule screening)
     public int _max_active_predictors = -1;
@@ -250,19 +242,15 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
 
     // the following parameters are for GAM
     public int[] _num_knots; // array storing number of knots per basis function
-    public double[][] _knots;// store knots for each gam column specified in _gam_X
     public String[] _knot_ids;  // store frame keys that contain knots location for each gam column in gam_X;
     public String[] _gam_columns; // array storing which predictor columns are needed
-   // public BSType _bs; // choose spline function for gam column
     public int[] _bs; // choose spline function for gam column, 0 = cr
     public double[] _scale;  // array storing scaling values to control wriggliness of fit
-    public GLMType _glmType = GLMType.gam; // internal parameter
     public boolean _saveZMatrix = false;  // if asserted will save Z matrix
     public boolean _keep_gam_cols = false;  // if true will save the keys to gam Columns only
     public boolean _savePenaltyMat = false; // if true will save penalty matrices as tripple array
-
     public String algoName() { return "GAM"; }
-    public String fullName() { return "General Additive Model"; }
+    public String fullName() { return "Generalized Additive Model"; }
     public String javaName() { return GAMModel.class.getName(); }
 
     @Override
@@ -307,7 +295,6 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
 
     public final static double linkInv(double x, Link link, double tweedie_link_power) {
       switch(link) {
-//        case multinomial: // should not be used
         case identity:
           return x;
         case ologlog:
@@ -328,6 +315,21 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
           throw new RuntimeException("unexpected link function  " + link.toString());
       }
     }
+  }
+
+  @Override
+  protected String[][] scoringDomains(){
+    String [][] domains = _output._domains;
+    if ((_parms._family == Family.binomial || _parms._family == Family.quasibinomial ||
+            _parms._family == Family.fractionalbinomial)
+            && _output._domains[_output._dinfo.responseChunkId(0)] == null) {
+      domains = domains.clone();
+      if (_parms._family == Family.fractionalbinomial)
+        domains[_output._dinfo.responseChunkId(0)] = BINOMIAL_CLASS_NAMES;
+      else
+        domains[_output._dinfo.responseChunkId(0)] = _output._responseDomains;
+    }
+    return domains;
   }
 
   public static class GAMModelOutput extends Model.Output {
@@ -361,6 +363,7 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
     private double[] _zvalues;
     private double _dispersion;
     private boolean _dispersionEstimated;
+    public String[][] _gamColNames; // store gam column names after transformation and decentering
     public double[][][] _zTranspose; // Z matrix for de-centralization, can be null
     public double[][][] _penaltyMatrices_center; // stores t(Z)*t(D)*Binv*D*Z and can be null
     public double[][][] _penaltyMatrices;          // store t(D)*Binv*D and can be null
@@ -388,9 +391,11 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
     /** Names of levels for a categorical response column. */
     @Override
     public String[] classNames() {
-      if (_family == Family.fractionalbinomial) {
-        return new String[]{"0", "1"};
-      } else 
+      if (_family == Family.quasibinomial || _family == Family.binomial)
+        return _responseDomains;
+      else if (_family == Family.fractionalbinomial)
+        return BINOMIAL_CLASS_NAMES;
+      else
         return super.classNames();
     }
 
@@ -398,12 +403,17 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
       super(b, adaptr);
       _dinfo = dinfo;
       _domains = dinfo._adaptedFrame.domains(); // get domain of dataset predictors
-      _responseDomains = dinfo._adaptedFrame.lastVec().domain();
       _family = b._parms._family;
+      if (_family.equals(Family.quasibinomial)) {
+        _responseDomains = new VecUtils.CollectDoubleDomain(null, 2).doAll(dinfo._adaptedFrame.vec(b._parms._response_column)).stringDomain(dinfo._adaptedFrame.vec(b._parms._response_column).isInt());
+      } else
+        _responseDomains = dinfo._adaptedFrame.lastVec().domain();
     }
 
     @Override public ModelCategory getModelCategory() {
       switch (_family) {
+        case quasibinomial:
+        case fractionalbinomial:
         case binomial: return ModelCategory.Binomial;
         case multinomial: return ModelCategory.Multinomial;
         case ordinal: return ModelCategory.Ordinal;
@@ -470,12 +480,7 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
         }
       }
     }
-    int numCols = adptedF.numCols();  // remove constant or bad frames.
-    for (int vInd=0; vInd<numCols; vInd++) {
-      Vec v = adptedF.vec(vInd);
-      if ((parms._ignore_const_cols &&  v.isConst()) || v.isBad())
-        adptedF.remove(vInd);
-    }
+
     Vec respV = null;
     if (ArrayUtils.contains(testNames, parms._response_column))
       respV = adptedF.remove(parms._response_column);
@@ -506,10 +511,10 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
       adaptFrm = new Frame(adaptFrm.names(),adaptFrm.vecs());
       adaptFrm.remove(responseId);
     }
-// Build up the names & domains.
     final boolean detectedComputeMetrics = computeMetrics && (adaptFrm.vec(_output.responseName()) != null && !adaptFrm.vec(_output.responseName()).isBad());
     String [] domain = _output.nclasses()<=1 ? null : (!detectedComputeMetrics ? _output._domains[_output._domains.length-1] : adaptFrm.lastVec().domain());
-// Score the dataset, building the class distribution & predictions
+    if (_parms._family.equals(Family.quasibinomial))
+      domain = _output._responseDomains;
     return new GAMScore(j, this, _output._dinfo.scoringInfo(_output._names,adaptFrm),domain,detectedComputeMetrics);
   }
 
@@ -530,6 +535,7 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
     final boolean _sparse;
     private transient double[][] _vcov;
     private transient double[] _tmp;
+    private boolean _classifier2class;
 
 
     private GAMScore(Job j, GAMModel m, DataInfo dinfo, String[] domain, boolean computeMetrics) {
@@ -540,6 +546,8 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
       _predDomains = domain;
       _m._parms = m._parms;
       _nclass = m._output.nclasses();
+      _classifier2class = _m._parms._family == GLMModel.GLMParameters.Family.binomial || 
+              _m._parms._family == Family.quasibinomial || _m._parms._family == Family.fractionalbinomial;
       if(_m._parms._family == GLMModel.GLMParameters.Family.multinomial ||
               _m._parms._family == GLMModel.GLMParameters.Family.ordinal){
         _coeffs = null;
@@ -620,10 +628,8 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
     public double[] scoreRow(DataInfo.Row r, double offset, double[] preds) {
       double mu = _m._parms.linkInv(r.innerProduct(_coeffs) + offset, _m._parms._link,
               _m._parms._tweedie_link_power);
-      if (_m._parms._family == GLMModel.GLMParameters.Family.binomial ||
-              _m._parms._family == GLMModel.GLMParameters.Family.quasibinomial ||
-      _m._parms._family == Family.negativebinomial || _m._parms._family == Family.fractionalbinomial) { // threshold for prediction
-        preds[0] = mu >= _defaultThreshold?1:0;
+      if (_classifier2class) { // threshold for prediction
+        preds[0] = mu >= _defaultThreshold ? 1 : 0;
         preds[1] = 1.0 - mu; // class 0
         preds[2] = mu; // class 1
       } else
@@ -692,6 +698,11 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
   @Override
   public double[] score0(double[] data, double[] preds) {
     throw new UnsupportedOperationException("GAMModel.score0 should never be called");
+  }
+
+  @Override
+  public GAMMojoWriter getMojo() {
+    return new GAMMojoWriter(this);
   }
 
   @Override
--- a/h2o-algos/src/main/java/hex/gam/GAMMojoWriter.java
+++ b/h2o-algos/src/main/java/hex/gam/GAMMojoWriter.java
@@ -1,0 +1,98 @@
+package hex.gam;
+
+
+import hex.ModelMojoWriter;
+import hex.glm.GLMModel;
+
+import java.io.IOException;
+
+import static hex.glm.GLMModel.GLMParameters.Family.*;
+
+public class GAMMojoWriter extends ModelMojoWriter<GAMModel, GAMModel.GAMParameters, GAMModel.GAMModelOutput> {
+  @Override
+  public String mojoVersion() {
+    return "1.00";
+  }
+  
+  public GAMMojoWriter(GAMModel model) {
+    super(model);
+  }
+
+  @Override
+  protected void writeModelData() throws IOException {
+    int numGamCols = model._parms._gam_columns.length;
+    writekv("use_all_factor_levels", model._parms._use_all_factor_levels);
+    writekv("cats", model._output._dinfo._cats);
+    writekv("cat_offsets", model._output._dinfo._catOffsets);
+    writekv("numsCenter", model._output._dinfo._nums);
+    writekv("num", model._output._dinfo._nums+numGamCols);
+
+    boolean imputeMeans = model._parms.missingValuesHandling().equals(GLMModel.GLMParameters.MissingValuesHandling.MeanImputation);
+    writekv("mean_imputation", imputeMeans);
+    if (imputeMeans) {
+      writekv("numNAFillsCenter", model._output._dinfo.numNAFill());
+      double[] numNAFills = new double[model._output._dinfo.numNAFill().length+numGamCols];
+      System.arraycopy(model._output._dinfo.numNAFill(),0, numNAFills, 0, 
+              model._output._dinfo.numNAFill().length);
+      int startind = numNAFills.length-model._gamColMeans.length;
+      System.arraycopy(model._gamColMeans, 0, numNAFills, startind, model._gamColMeans.length);
+      writekv("numNAFills", numNAFills);
+      writekv("catNAFills", model._output._dinfo.catNAFill());
+    }
+    if (model._parms._family.equals(binomial))
+      writekv("family", "bernoulli");
+    else
+      writekv("family", model._parms._family);
+    writekv("link", model._parms._link);
+    if (model._parms._family.equals(GLMModel.GLMParameters.Family.tweedie))
+      writekv("tweedie_link_power", model._parms._tweedie_link_power);
+    // add GAM specific parameters
+    writekv("num_knots", model._parms._num_knots); // an array
+    writeStringArrays(model._parms._gam_columns, "gam_columns"); // gam_columns specified by users
+    int numGamLength = 0;
+    int numGamCLength = 0;
+    for (int cInd=0; cInd < numGamCols; cInd++)  { // only contains expanded gam column names not centered
+      writeStringArrays(model._gamColNames[cInd], "gamColNamesCenter_"+model._parms._gam_columns[cInd]);
+      writeStringArrays(model._gamColNamesNoCentering[cInd], "gamColNames_"+model._parms._gam_columns[cInd]);
+      numGamLength += model._gamColNamesNoCentering[cInd].length;
+      numGamCLength += model._gamColNames[cInd].length;
+    }
+    String[] trainColGamColNoCenter = genTrainColGamCols(numGamLength, numGamCLength);
+    writekv("num_expanded_gam_columns", numGamLength);
+    writeStringArrays(trainColGamColNoCenter, "_names_no_centering"); // column names without centering
+    writekv("total feature size", trainColGamColNoCenter.length);
+    if (model._parms._family==multinomial || model._parms._family==ordinal) {
+      writeDoubleArray(model._output._model_beta_multinomial_no_centering, "beta_multinomial");
+      writekv("beta length per class", model._output._model_beta_multinomial_no_centering[0].length);
+      writeDoubleArray(model._output._model_beta_multinomial, "beta_multinomial_centering");
+      writekv("beta center length per class", model._output._model_beta_multinomial[0].length);
+    } else {
+      writekv("beta", model._output._model_beta_no_centering); // beta without centering
+      writekv("beta length per class", model._output._model_beta_no_centering.length);
+      writekv("beta_center", model._output._model_beta);
+      writekv("beta center length per class", model._output._model_beta.length);
+    }
+    writekv("bs", model._parms._bs);  // an array of choice of spline functions
+    writeDoubleArray(model._output._knots, "knots");
+    int countGamCols = 0;
+    for (String gamCol : model._parms._gam_columns) {
+      writeDoubleArray(model._output._zTranspose[countGamCols], gamCol+"_zTranspose");      // write zTranspose
+      writeDoubleArray(model._output._binvD[countGamCols++], gamCol+"_binvD");      // write binvD
+    }
+    // store variable importance information
+  }
+  
+  public String[] genTrainColGamCols(int gamColLength, int gamCColLength) {
+    int colLength = model._output._names.length-gamCColLength+gamColLength-1;// to exclude response
+    int normalColLength = model._output._names.length-gamCColLength-1;
+    String[] trainNamesNGamNames = new String[colLength];
+    System.arraycopy(model._output._names, 0, trainNamesNGamNames, 0, normalColLength);
+    int startInd = normalColLength;
+    for (int gind = 0; gind < model._gamColNamesNoCentering.length; gind++) {
+      int copyLen = model._gamColNamesNoCentering[gind].length;
+      System.arraycopy(model._gamColNamesNoCentering[gind], 0, trainNamesNGamNames, startInd, copyLen);
+      startInd += copyLen;
+    }
+    return trainNamesNGamNames;
+  }
+}
--- a/h2o-algos/src/main/java/hex/gam/GAMMojoWriter.java
+++ b/h2o-algos/src/main/java/hex/gam/GAMMojoWriter.java
@@ -14,6 +14,9 @@ public class GAMMojoWriter extends ModelMojoWriter<GAMModel, GAMModel.GAMParamet
     return "1.00";
   }
   
+  @SuppressWarnings("unused")
+  public GAMMojoWriter(){}
+  
   public GAMMojoWriter(GAMModel model) {
     super(model);
   }
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -1979,31 +1979,6 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       Frame train = DKV.<Frame>getGet(_parms._train); // need to keep this frame to get scoring metrics back
       _model.score(train).delete();
       scoreEnd(train, t1);
-/*
-        ModelMetrics mtrain = ModelMetrics.getFromDKV(_model, train); // updated by model.scoreAndUpdateModel
-        long t2 = System.currentTimeMillis();
-        if (!(mtrain == null)) {
-          _model._output._training_metrics = mtrain;
-          Log.info(LogMsg(mtrain.toString()));
-        } else {
-          Log.info(LogMsg("ModelMetrics mtrain is null"));
-        }
-        Log.info(LogMsg("Training metrics computed in " + (t2 - t1) + "ms"));
-        if (_valid != null) {
-          Frame valid = DKV.<Frame>getGet(_parms._valid);
-          _model.score(valid).delete();
-          _model._output._validation_metrics = ModelMetrics.getFromDKV(_model, valid); //updated by model.scoreAndUpdateModel
-        }
-      _model._output._scoring_history = _parms._lambda_search?_lsc.to2dTable():(_parms._HGLM?_sc.to2dTableHGLM():_sc.to2dTable());
-      _model.update(_job._key);
-      if (_parms._HGLM)
-        _model.generateSummary(_state._iter, _parms._train);
-      else
-        _model.generateSummary(_parms._train,_state._iter);
-      _lastScore = System.currentTimeMillis();
-      long scoringTime = System.currentTimeMillis() - t1;
-      _scoringInterval = Math.max(_scoringInterval,20*scoringTime); // at most 5% overhead for scoring
-*/
     }
     
     private void scoreEnd(Frame train, long t1) {
--- a/h2o-algos/src/test/java/hex/gam/GamTestPiping.java
+++ b/h2o-algos/src/test/java/hex/gam/GamTestPiping.java
@@ -19,7 +19,7 @@ import water.runner.H2ORunner;
 import water.util.ArrayUtils;
 
 import static hex.gam.MatrixFrameUtils.GamUtils.equalColNames;
-import static hex.gam.MatrixFrameUtils.GamUtils.locateBin;
+import static hex.genmodel.algos.gam.GamUtilsCubicRegression.locateBin;
 import static hex.glm.GLMModel.GLMParameters.Family.*;
 import static hex.glm.GLMModel.GLMParameters.GLMType.gam;
 import static hex.glm.GLMModel.GLMParameters.GLMType.glm;
@@ -32,7 +32,6 @@ import static org.junit.Assert.assertEquals;
 @RunWith(H2ORunner.class)
 @CloudSize(1)
 public class GamTestPiping extends TestUtil {
-
   /**
    * This test is to make sure that we carried out the expansion of ONE gam column to basis functions
    * correctly with and without centering.  I will compare the following with R runs:
@@ -54,7 +53,8 @@ public class GamTestPiping extends TestUtil {
       final GAMModel model = getModel(gaussian,
               Scope.track(parse_test_file("smalldata/glm_test/multinomial_10_classes_10_cols_10000_Rows_train.csv")),
               "C11", gamCols, ignoredCols, new int[]{5}, new int[]{0}, false,
-              true, new double[]{1}, new double[]{0}, new double[]{0}, false, null,null);
+              true, new double[]{1}, new double[]{0}, new double[]{0}, false, null,
+              null, true);
       Scope.track_generic(model);
       final double[][] rBinvD = new double[][]{{1.5605080,
               -3.5620961, 2.5465468, -0.6524143, 0.1074557}, {-0.4210098, 2.5559955, -4.3258597, 2.6228736,
@@ -75,7 +75,7 @@ public class GamTestPiping extends TestUtil {
       Frame rTransformedData = parse_test_file("smalldata/gam_test/multinomial_10_classes_10_cols_10000_Rows_train_C6Gam.csv");
       Scope.track(rTransformedData);
 
-      Frame transformedDataC = ((Frame) DKV.getGet(model._output._gamTransformedTrainCenter));  // compare model matrix with centering
+      Frame transformedDataC = ((Frame) DKV.getGet(model._output._gamTransformedTrainCenter));// compare model matrix with centering
       Scope.track(transformedDataC);
       Scope.track(transformedDataC.remove("C11"));
       Frame rTransformedDataC = parse_test_file("smalldata/gam_test/multinomial_10_classes_10_cols_10000_Rows_train_C6Gam_center.csv");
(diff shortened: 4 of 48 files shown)


h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -3643,6 +3643,20 @@ def build_save_model_GLM(params, x, train, respName):
     model.download_mojo(path=TMPDIR)    # save mojo
     return model
 
+def build_save_model_GAM(params, x, train, respName):
```
Pscheidl, line 3646: I'd personally create a common file in the pyunit folder with GAM tests and not put it into test common utils, but it's not really a big issue.

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -3643,6 +3643,20 @@ def build_save_model_GLM(params, x, train, respName):
...
+    # build a model
+    model = H2OGeneralizedAdditiveEstimator(**params)
+    model.train(x=x, y=respName, training_frame=train)
+    # save model
+    regex = re.compile("[+\\-* !@#$%^&()={}\\[\\]|;:'\"<>,.?/]")
+    MOJONAME = regex.sub("_", model._id)
+
+    print("Downloading Java prediction model code from H2O")
+    TMPDIR = os.path.normpath(os.path.join(os.path.dirname(os.path.realpath('__file__')), "..", "results", MOJONAME))
```
Pscheidl, line 3655: You can just `import tempfile` and then do `tempfile.mkdtemp()` - Please look at `pyunit_generic_model_mojo_glm.py` to see how I did it. It's also guaranteed the file will be removed afterwards, now it's not guaranteed (at least I can't see it). It's a simplification, so I suggest doing it this way.

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -3643,6 +3643,20 @@ def build_save_model_GLM(params, x, train, respName):
...
+    os.makedirs(TMPDIR)
+    model.download_mojo(path=TMPDIR)    # save mojo
+    return model
```
Pscheidl, line 3658:
```suggestion
    return model.download_mojo(path=TMPDIR)    # save mojo
```
Shouldn't we return the path returned by the `download.mojo` method ?

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoMultinomialModel.java
```diff
@@ -0,0 +1,68 @@
+package hex.genmodel.algos.gam;
+
+import java.util.Arrays;
+
+public class GamMojoMultinomialModel extends GamMojoModelBase {
+    private boolean _multinomial; // true multinomial, false ordinal
```
Pscheidl, line 6: Let's choose a better name in order not to require reading the commentary. Something like `trueMultinomial` would do.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoMultinomialModel.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+    
+    @Override
+    double[] gamScore0(double[] row, double[] preds) {
+        if (row.length == nfeatures())
+            _beta_multinomial = _beta_multinomial_center;
+        else
+            _beta_multinomial = _beta_multinomial_no_center;
+        
+        Arrays.fill(preds, 0.0);
```
Pscheidl, line 24: Do we need to fill the preds with zeros ? The preds should be pre-initialized with 0s already. I can see this is copied from GLM. Just that I noticed this.

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: I actually write a build_save_model_generic that can be shared among different algos

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: yes, I switched to tempfile

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: In this case, I actually need to model back in order to use it to do scoring using the model.  Then, I use mojo to do the scoring and compare the two to make sure they are the same.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoMultinomialModel.java
wendycwong: I use _trueMultinomial

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoMultinomialModel.java
wendycwong: I deleted it.  Thanks.

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -470,12 +488,12 @@ public static Frame cleanUpInputFrame(Frame adptedF, GAMParameters parms, String
         }
       }
     }
-    int numCols = adptedF.numCols();  // remove constant or bad frames.
+ /*   int numCols = adptedF.numCols();  // remove constant or bad frames.
```
honzasterba: remove commented out code

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -622,7 +640,7 @@ else if (r.weight == 0)
               _m._parms._tweedie_link_power);
       if (_m._parms._family == GLMModel.GLMParameters.Family.binomial ||
               _m._parms._family == GLMModel.GLMParameters.Family.quasibinomial ||
-      _m._parms._family == Family.negativebinomial || _m._parms._family == Family.fractionalbinomial) { // threshold for prediction
+      _m._parms._family == Family.fractionalbinomial) { // threshold for prediction
```
honzasterba, line 638: indent

h2o-algos/src/main/java/hex/gam/GAMMojoWriter.java
```diff
@@ -0,0 +1,117 @@
...
+    int startInd = normalColLength;
+    for (int gind = 0; gind < model._gamColNamesNoCentering.length; gind++) {
+      int copyLen = model._gamColNamesNoCentering[gind].length;
+      System.arraycopy(model._gamColNamesNoCentering[gind], 0, trainNamesNGamNames, startInd, copyLen);
+      startInd += copyLen;
+    }
+    return trainNamesNGamNames;
+  }
+  
+  public void writeStringArrays(String[] sArrays, String title) throws IOException {
```
honzasterba, line 98: move to ModelMojoWriter, same below

h2o-algos/src/test/java/hex/gam/GamMojoTesting.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.util.ArrayList;
+
+import static hex.gam.GamTestPiping.getModel;
+import static hex.gam.GamTestPiping.massageFrame;
+import static hex.glm.GLMModel.GLMParameters.Family.*;
+
+@RunWith(H2ORunner.class)
+@CloudSize(1)
+public class GamMojoTesting extends TestUtil {
```
honzasterba, line 23: no need to extend TestUtil with H2ORunner
class name does not follow convention should be something like GamMojoModelTest

h2o-genmodel/src/main/java/hex/genmodel/tools/PredictCsv.java
```diff
@@ -238,8 +240,18 @@ public void run() throws Exception {
       String[] splitLine;
       //Reader in the column names here.
       if ((splitLine = reader.readNext()) != null) {
-        inputColumnNames = splitLine;
+        inputColumnNames = splitLine; // contains the column names of input data
         checkMissingColumns(inputColumnNames);
+        if (model.m instanceof GamMojoModelBase) {
```
honzasterba, line 245: I feel like this could be handled by GamMojoModelBase.predict

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -62,8 +63,14 @@
   private final boolean enableStagedProbabilities; // if set true, staged probabilities from tree agos are returned
   private final boolean enableContributions; // if set to true, will return prediction contributions (SHAP values) - for GBM & XGBoost
   private final int glrmIterNumber; // allow user to set GLRM mojo iteration number in constructing x.
-
   private final PredictContributions predictContributions;
+  private int predictColumnNum; // Count number of columns that are found in input data row and gam predictor columns
+
+  public void setPredictColumnNum(int predNum) throws IOException {
```
honzasterba, line 69: this exposes a public API for the sake of one algo, no user will ever call this, either we need to make this private or at least package private
also I think this functionality should be hidden inside GamMojoModel

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -0,0 +1,144 @@
...
+  int _totFeatureSize; // number of predictors plus gam columns no centered
+  int _betaSizePerClass;
+  int _betaCenterSizePerClass;
+  double _tweedieLinkPower;
+  double[][] _basisVals; // store basis values array for each gam column
+  double[][] _hj;   // difference between knot values
+  int _numExpandedGamCols; // number of expanded gam columns
+  int _lastClass;
+  
+  public int get_totFeatureSize() { return _totFeatureSize;}
```
honzasterba, line 45: please respect java naming convention when naming methods and vars, same below

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -46,7 +47,7 @@
   
   @Override public ModelMetrics.MetricBuilder makeMetricBuilder(String[] domain) {
     if (domain==null && (_parms._family==Family.binomial || _parms._family==Family.quasibinomial || 
-            _parms._family==Family.negativebinomial || _parms._family==Family.fractionalbinomial))
```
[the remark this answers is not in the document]
wendycwong: Yes, in the Gam_Mojo test, I have a test for negativebinomial and fractionalbinomial distribution where I build a model and compare the model predict and mojo predict.

h2o-algos/src/test/java/hex/gam/GamMojoTesting.java
wendycwong: Done.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: Sure.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -2380,6 +2381,9 @@ public boolean testJavaScoring(Frame data, Frame model_predictions, EasyPredictM
                   .setEnableStagedProbabilities(genmodel instanceof SharedTreeMojoModel)
                   .setUseExternalEncoding(true); // input Frame is already adapted!
           epmw = new EasyPredictModelWrapper(config);
+          if (genmodel instanceof GamMojoModelBase) {
+            ((GamMojoModelBase) genmodel).setPredictColumnName(genmodel._names.length); // column names found in input and original model
```
honzasterba, line 2385: this looks like poor design, basically what you are doing is genmodel.setPredictColumnName(genmodel._names.length)
1) telling genmodel object about something it already knows about
2) set*Name(.length) is confusing naming

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModel.java
```diff
@@ -0,0 +1,37 @@
...
+  double _tweedieLinkPower;
+  private boolean _binomial;
+  
+  GamMojoModel(String[] columns, String[][] domains, String responseColumn) {
+    super(columns, domains, responseColumn);
+  }
+  
+  void init() {
+    super.init();
+    _binomial = _family.equals("binomial") || _family.equals("fractionalbinomial") || _family.equals("quasibinomial");
```
honzasterba, line 14: using String comparison here sounds dangerous and error prone, ohter Mojos use DistributionFamily internally as well, I suggest chaning the reader to do DistributionFamily.valueOf and use DistributionFamily internally, same for _link

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoMultinomialModel.java
```diff
@@ -0,0 +1,65 @@
...
+public class GamMojoMultinomialModel extends GamMojoModelBase {
+    private boolean _trueMultinomial;
+
+    GamMojoMultinomialModel(String[] columns, String[][] domains, String responseColumn) {
+        super(columns, domains, responseColumn);
+    }
+
+    void init() {
+        super.init();
+        _trueMultinomial = _family.equals("multinomial");
```
honzasterba, line 12: use enum

h2o-core/src/main/java/hex/Model.java
wendycwong: This is needed because I need to know what the input row column names are.  In this case, it is use the genmodel._names.  However, in other case, it can be something very different.  I did change a bit to make it clearer.

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -28,9 +28,11 @@
 import static hex.glm.GLMModel.GLMParameters.MissingValuesHandling;
 
 public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.GAMModelOutput> {
+  final private static String[] BINOMIALCLASSNAMES = new String[]{"0", "1"};
```
honzasterba, line 31: should be `private static final BINOMIAL_CLASS_NAMES `

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -0,0 +1,175 @@
...
+  public void setPredictColumNum(String[] inputColumnNames) {
+    _predictColumnNum = 0;
+    String[] colNames = _names; // contains predictor columns for Gam including expanded gam columns
+    for (String cName : colNames) {
+      if (Arrays.asList(inputColumnNames).contains(cName))
+        _predictColumnNum++; // count column names that are present both in input data row and gam predictor names
+    }
+  }
+  
+ // public void setPredictColumnName(int colNum) { _predictColumnNum = colNum;}
```
honzasterba, line 73: remove commented out code

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -0,0 +1,175 @@
...
+  
+  GamMojoModelBase(String[] columns, String[][] domains, String responseColumn) {
+    super(columns, domains, responseColumn);
+  }
+
+  public void setPredictColumNum(String[] inputColumnNames) {
+    _predictColumnNum = 0;
+    String[] colNames = _names; // contains predictor columns for Gam including expanded gam columns
+    for (String cName : colNames) {
+      if (Arrays.asList(inputColumnNames).contains(cName))
```
honzasterba, line 68: can it happen that `_predictColumnNum != inputColumnNames.lenth - _names.length`? if not than this is unnecessary and can be done in predict

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -0,0 +1,175 @@
...
+  public String[] getNames() { return _names; }  // get column names centered
+  public int getPredictColumnNum() { return _predictColumnNum;}
+  
+  GamMojoModelBase(String[] columns, String[][] domains, String responseColumn) {
+    super(columns, domains, responseColumn);
+  }
+
+  public void setPredictColumNum(String[] inputColumnNames) {
+    _predictColumnNum = 0;
+    String[] colNames = _names; // contains predictor columns for Gam including expanded gam columns
```
honzasterba, line 66: unnecessary variable

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -0,0 +1,175 @@
...
+    int noff = _catOffsets[_cats] - _cats;
+    for (int i = _cats; i < beta.length - 1 - noff; ++i)
+      eta += beta[noff + i] * data[i];
+    eta += beta[beta.length - 1]; // add intercept
+    return eta;
+  }
+  
+  // this method will add to each data row the expanded gam columns
+  public void addExpandGamCols(double[] rawData, RowData rowData) { // add all expanded gam columns here
+    // add expanded gam columns to rowData
```
honzasterba, line 151: mentions rowData but the method modifies rawData, I am confused

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: @honza:  Yes, this can happen.  With GAM, gam_columns will be added to the training frame.  When you send me a training dataset, I will add the expanded gam_columns to the training frame.  If you want to do scoring with test dataset, it will not have the expanded gam_columns and hence inputColumnNames will be less than the _names that includes the predictors and expanded gam columns.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: @honzasterba for GAM, we need to take a predictor column, expanded it into basis functions and obtained the coefficients.  The rowData contains the value of the predictor column.  The newly expanded gam columns and the necessary predictor values are put into rawData which will later be used to do the scoring.  If you want to understand how the gam columns are constructed, I will slack you do doc I wrote.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
honzasterba: this is MOJO we are talking about, so there is no train/test dataset, there is only the production dataset, that is the same structure as train, isnt that right?

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
honzasterba: let me try to rephrase, the comment says "add expanded gam columns to rowData" yet I see no modifications done to rowData in the method, is the comment wrong or is the method wrong?
also the comment above the method mentions something about "each data row" yet this method operates strictly on one row of data
hence my confusion

h2o-genmodel/src/main/java/hex/genmodel/utils/DistributionFamily.java
```diff
@@ -8,6 +8,7 @@
 public enum DistributionFamily {
   AUTO,  // model-specific behavior
   bernoulli,
+  binomial,
```
honzasterba, line 11: why were these new distribution types not needed for the algo impl, but are needed for mojo?

h2o-genmodel/src/main/java/hex/genmodel/utils/LinkFunctionType.java
```diff
@@ -12,5 +12,6 @@
     ologit,
     ologlog,
     oprobit,
-    inverse
+    inverse,
+    tweedie
```
honzasterba, line 16: same as the distribution, why it is needed now?

h2o-genmodel/src/main/java/hex/genmodel/utils/DistributionFamily.java
wendycwong: @honzasterba:  For GLM, the families are defined as Gaussian, Binomial, Multinomial,....

For some reason, the other H2O toolboxes use Bernoulli for Binomial.  That is why I added the binomial to the DistributionFamily to accommodate both ways.

wendycwong: Also, thanks for reminding me about this.  I left out one family that are binomial as well, namely, fractionalbinomial.  Thanks.

h2o-genmodel/src/main/java/hex/genmodel/utils/LinkFunctionType.java
wendycwong: For GLM, the link function for tweedie is also named Tweedie.  Please don't ask me why.  Tomas did it.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: @honzasterba :  That is correct.  The user will give us a dataset that have the same predictor columns as in the train dataset.  Note that the gam_columns are built off the train dataset predictors that the user specified in gam_columns.  Here is the data flow:
1. input: train dataset,
2. add gam_columns: attach gam_columns that are derived from predictor columns in train dataset.

When we have scoring, we will do the same thing:
1. input: production/test dataset
2. add gam_columns: attach gam_columns that are derived from predictor columns in test dataset.

Is this clear?  Please let me know.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: @honza:  I got what you mean now.

Okay, rowData contains say c0, c1, c2.  The user specify gam_columns = ["c0"].

So, the next step is Gam will take the value of c0 in rowData and generate a new set of gam_columns and put these new columns into rawData.  So, we need the information from rowData and modifies only rawData.

I can make RowData final though since it is not modified.  Thanks.

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -3468,8 +3469,8 @@ def compare_frames_local(f1, f2, prob=0.5, tol=1e-6, returnResult=False):
     :param returnResult:
     :return:
     '''
-    assert (f1.nrow==f2.nrow) and (f1.ncol==f2.ncol), "Frame 1 row {0}, col {1}.  Frame 2 row {2}, col {3}.  " \
-                                                      "They are different.".format(f1.nrow, f1.ncol, f2.nrow, f2.ncol)
+    assert (f1.nrow==f2.nrow) and (f1.ncol==f2.ncol), "Frame 1 row {0}, col {1}.  Frame 2 row {2}.  They are " \
```
honzasterba, line 3472: did you remove {3} on purspose?

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -261,7 +262,7 @@ def mojo_predict(model, tmpdir, mojoname, glrmReconstruct=False, get_leaf_node_a
     mojoZip = os.path.join(tmpdir, mojoname) + ".zip"
     if not(zipFilePath==None):
         mojoZip = zipFilePath
-    genJarDir = str.split(str(tmpdir),'/')
+    genJarDir = str.split(os.path.realpath("__file__"),'/')
```
honzasterba, line 265: what is the reason for this?

h2o-r/tests/testdir_javapredict/runit_PUBDEV_7185_binomial_gam_mojo_large.R
```diff
@@ -0,0 +1,56 @@
...
+      modelAndDir<-buildModelSaveMojoGAM(params_prob_data$params) # build the model and save mojo
+      filename = sprintf("%s/in.csv", modelAndDir$dirName) # save the test dataset into a in.csv file.
+      h2o.downloadCSV(params_prob_data$tDataset[,params_prob_data$params$x], filename)
+      twoFrames<-mojoH2Opredict(modelAndDir$model, modelAndDir$dirName, filename) # perform H2O and mojo prediction and return frames
+      h2o.downloadCSV(twoFrames$h2oPredict, sprintf("%s/h2oPred.csv", modelAndDir$dirName))
+      h2o.downloadCSV(twoFrames$mojoPredict, sprintf("%s/mojoOut.csv", modelAndDir$dirname))
+      compareFrames(twoFrames$h2oPredict,twoFrames$mojoPredict, prob=1, tolerance = 1e-4)
+    }, error = function(x) x)
+    if (!is.null(e)&& (!all(sapply("wget", grepl, e[[1]]))))
+      FAIL(e)   # throw error unless it is the stupid wget error.
```
honzasterba, line 23: what wget error? if this sometimes fails on network it should fail for real or try again, this is just hiding test failures

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamRowToRawDataConverter.java
```diff
@@ -0,0 +1,33 @@
...
+  public double[] convert(RowData data, double[] rawData) throws PredictException {
+    if (_m.getPredictColumnNum() == 0) 
+      _m.setPredictColumNum(data.keySet()); // set predictColColumnNum if it has not been set.
+    if (rawData.length != _m._predictColumnNum) {
+      rawData = nanArray(_m.getTotFeatureSize());
+      rawData = super.convert(data, rawData);
+      _m.addExpandGamCols(rawData, data);
+      return rawData;
+    } else 
+      return super.convert(data, rawData);
```
honzasterba, line 31: in all your tests-scenarios, does this ever really happen? I don't think it does.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -0,0 +1,171 @@
...
+  public int getPredictColumnNum() { return _predictColumnNum;}
+  
+  GamMojoModelBase(String[] columns, String[][] domains, String responseColumn) {
+    super(columns, domains, responseColumn);
+  }
+
+  public void setPredictColumNum(Set<String> inputColumnNames) {
+     for (String cName : features()) {
+      if (inputColumnNames.contains(cName))
+        _predictColumnNum++; // count column names that are present both in input data row and gam predictor names
```
honzasterba, line 67: I still feel like this logic is unnecessary, in addExpandGamCols you expect the exact size of the row to be features+gam_cols, in now way has _predictColumnNum impact on that method

wendycwong: @honza:  When you send in the training dataset, all you have is predictors, columns chosen to be transformed to the gam_columns,  ignored columns and response column.  Then, gam will add expanded gam_cols to the training dataset and the features will include just the predictors and the expanded gam_cols and the response column.

Then, you are given the test dataset which should include the predictors, columns chosen to be transformed and other columns that we don't care about like the ignored columns and others.  The setPredictColumNum is just trying to find out how many predictor columns the test dataset contain when compared to the features that the model desired.  

I thought about getting rid of it too cannot find a way around it.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamRowToRawDataConverter.java
wendycwong: It actually does in testJavaScoring when I try to write the Junit test for GamMojo.

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: I changed the directory to write the mojo.zip into.  Originally, it is written off h2o-3/h2o-py/tests/results directory.  This means tmpdir = /absolutepath/h2o-3/h2o-py/tests/results.  However, with @Pscheidl suggestion, I use tempfile and the directory is no longer part of h2o-3 path. 

I need to find the directory of the genmodel.jar and it is off the h2o-3/h2o-genmodel/...   Hence, I changed genJarDir from str.split(str(tmpdir), '/') to str.split(os.path.realpath("__file__"), '/') which will find the directory of the file being executed.  The current file that is being executed is located in /absolute-path/h2o-3/h2o-py/tests/testdir_algos/gam/.  Hence, we can use this to find where the genmodel.jar is located.

h2o-r/tests/testdir_javapredict/runit_PUBDEV_7185_binomial_gam_mojo_large.R
wendycwong: I removed it.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -0,0 +1,171 @@
...
+  
+  public int getTotFeatureSize() { return _totFeatureSize;}
+  public String[] getNames() { return _names; }  // get column names centered
+  public int getFeautesPresentInData() { return _featuresPresentInDataCount;}
+  
+  GamMojoModelBase(String[] columns, String[][] domains, String responseColumn) {
+    super(columns, domains, responseColumn);
+  }
+
+  public void countFeautesPresentInData(Set<String> inputColumnNames) {
```
honzasterba, line 64: The problem here is that this makes the MOJO stateful. Meaning that two different orders of calls will produce different results.
We definitely need to get rid of _featuresPresentInDataCount.
There are two ways to do this:
1) (which I dont like) is to do countFeautesPresentInData() on every row we want to score, this will make scoring very slow and ultimatelly is not necessary
2) countFeautesPresentInData is present only because testJavaPreditct() is calling this mojo with weird synthetic data, which in turn is caused by the mojo._names containing gam columns, which are synthetic computational columns that should not be in _names in the first place.
WDYT?

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -330,6 +332,18 @@ public final static double linkInv(double x, Link link, double tweedie_link_powe
...
   }
 
+  @Override
+  protected String[][] scoringDomains(){
+    String [][] domains = _output._domains;
+    if ((_parms._family == Family.binomial || _parms._family == Family.quasibinomial ||
+            _parms._family == Family.fractionalbinomial)
+            && _output._domains[_output._dinfo.responseChunkId(0)] == null) {
+      domains = domains.clone();
+      domains[_output._dinfo.responseChunkId(0)] = BINOMIAL_CLASS_NAMES;
```
maurever, line 342: The quasibinomial case can have a completely different domain, for example, {-1, 3.42}. it is not correct to assign {0, 1} domain by default. A few months ago I implemented MRTask `CollectDoubleDomain` to calculate the quasibinomial domain correctly. Link to the class [here]( https://github.com/h2oai/h2o-3/blob/423c907d0c8f1d31e649f8a58e1f47f76e148788/h2o-core/src/main/java/water/util/VecUtils.java#L412).

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -388,7 +403,7 @@ public int nclasses() {
     /** Names of levels for a categorical response column. */
     @Override
     public String[] classNames() {
-      if (_family == Family.fractionalbinomial) {
+      if (_family == Family.fractionalbinomial || _family == Family.quasibinomial) {
```
maurever, line 406: Same here, quasibinomial could have a different domain than {0,1}.

h2o-algos/src/test/java/hex/gam/GamTestPiping.java
```diff
@@ -497,19 +499,22 @@ public void testModelScoring() {
...
       Scope.track_generic(binomialModel);
       binomialModel._output._training_metrics = null; // force prediction threshold of 0.5
       Frame predictBinomial = Scope.track(binomialModel.score(trainBinomial));
       Frame predictGLMBinomial = Scope.track(parse_test_file("smalldata/gam_test/predictBinomialGAM1.csv"));
       TestUtil.assertIdenticalUpToRelTolerance(predictBinomial, predictGLMBinomial, 1e-6);
       
+      // test for Gaussian
       Frame trainGaussian = Scope.track(massageFrame(parse_test_file("smalldata/glm_test/gaussian_20cols_10000Rows.csv"), gaussian));
       DKV.put(trainGaussian);
       GAMModel gaussianmodel = getModel(gaussian,
```
maurever, line 513: Just to be consistent... :)
```suggestion
      GAMModel gaussianModel = getModel(gaussian,
```

h2o-algos/src/test/java/hex/gam/GamTestPiping.java
```diff
@@ -573,15 +579,15 @@ public void testModelMetricsGeneration() {
       GAMModel binomialModel = getModel(binomial,  Scope.track(parse_test_file("smalldata/glm_test/binomial_20_cols_10KRows.csv"))
               , "C21", gamCols, ignoredCols, new int[]{5, 5, 5}, new int[]{0, 0, 0},
               false, true, new double[]{1, 1, 1}, new double[]{0, 0, 0},
-              new double[]{0, 0, 0}, true, null,null);
+              new double[]{0, 0, 0}, true, null,null, true);
       Scope.track_generic(binomialModel);
       verifyModelMetrics(binomialModel,1e-6);
 
       GAMModel gaussianmodel = getModel(gaussian,
```
maurever, line 586:
```suggestion
      GAMModel gaussianModel = getModel(gaussian,
```

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoFactory.java
```diff
@@ -2,17 +2,18 @@
 
 import hex.genmodel.algos.deeplearning.DeeplearningMojoReader;
 import hex.genmodel.algos.drf.DrfMojoReader;
+import hex.genmodel.algos.ensemble.StackedEnsembleMojoReader;
```
maurever, line 5: Are the  `hex.genmodel.algos.ensemble.StackedEnsembleMojoReader` and `hex.genmodel.algos.pca.PCAMojoReader` imports relevant to this PR?

h2o-genmodel/src/main/java/hex/genmodel/utils/DistributionFamily.java
maurever: @wendycwong, are you sure, you are using correct `DistributionFamily` enum and  `LinkFunctionType` enum? Because the GLM uses its own implementation of these enums. Other implementations use these enums. That is why binomial is missing here. The GLM really needs to refactor these parts of code to be consistent with other algos. So if it possible for GAM to use also Bernoulli, it is better than add the same distribution with the different name into this enum.

h2o-genmodel/src/main/java/hex/genmodel/utils/LinkFunctionType.java
maurever: @wendy, do you really use this enum in GLM? GLM has its own implementation of Distribution and LinkFunction.

h2o-r/tests/testdir_javapredict/runit_PUBDEV_7185_quasibinomial_gam_MOJO.R
```diff
@@ -0,0 +1,48 @@
+setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
+source("../../scripts/h2o-r-test-setup.R")
+
+test.GAM.quasi_binomial <- function() {
```
maurever, line 4:
```suggestion
test.GAM.quasibinomial <- function() {
```

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: @honzasterba I think I have a solution for that without using _featuresPresentInDataCount

h2o-algos/src/main/java/hex/gam/GAMModel.java
wendycwong: Clever!  I used it.

h2o-algos/src/main/java/hex/gam/GAMModel.java
wendycwong: Changed that too.

h2o-algos/src/test/java/hex/gam/GamTestPiping.java
wendycwong: Done

h2o-algos/src/test/java/hex/gam/GamTestPiping.java
wendycwong: Done

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoFactory.java
wendycwong: I am not sure what is going on, IntelliJ did auto import on my behave.  Super weird.

h2o-algos/src/main/java/hex/gam/GAMModel.java
wendycwong: I changed it.  Thanks.

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -28,9 +25,12 @@
 import static hex.glm.GLMModel.GLMParameters.MissingValuesHandling;
 
 public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.GAMModelOutput> {
+  private static final String[] BINOMIAL_CLASS_NAMES = new String[]{"0", "1"};
+  private static String[] _quasiBinomialDomain;
```
honzasterba, line 29: so if there are many mojo instances in the JVM they are all supposed to share _quasiBinomialDomain?

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -209,6 +212,10 @@ private void fillUpCoeffs(double[] coeffValues, double[] coeffValuesStand, TwoDi
       arrCounter++;
     }
   }
+  
+  public static void set_quasiBinomialDomain(String[] domains) {
```
maurever, line 216:
```suggestion
  public static void setQuasiBinomialDomain(String[] domains) {
```

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoFactory.java
maurever: So please, could you remove these imports if there are not related to your changes? Sometimes IntelliJ adds automatic import when I select wrongly autocompleted class, so that is maybe why these imports appeared. :)

h2o-genmodel/src/main/java/hex/genmodel/utils/DistributionFamily.java
maurever: @wendycwong, I am sorry, but I can't see this issue was resolved. There are actually two possibilities:
1. use GLM Distribution enum
OR
2. use this enum but with Bernoulli distribution.

Add the same distribution whit a different name is confusing here.

h2o-algos/src/main/java/hex/gam/GAMModel.java
maurever: Definitely should not be static.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamRowToRawDataConverter.java
```diff
@@ -0,0 +1,32 @@
...
+public class GamRowToRawDataConverter extends RowToRawDataConverter {
+  GamMojoModelBase _m;
+  public GamRowToRawDataConverter(GenModel m, Map<String, Integer> modelColumnNameToIndexMap, Map<Integer, CategoricalEncoder> domainMap, EasyPredictModelWrapper.ErrorConsumer errorConsumer, EasyPredictModelWrapper.Config config) {
+    super(m, modelColumnNameToIndexMap, domainMap, errorConsumer, config);
+    _m = (GamMojoModelBase) m;
+  }
+  
+  @Override
+  public double[] convert(RowData data, double[] rawData) throws PredictException {
+    if (data.containsKey(_m._gamColNamesCenter[0][0])) {
```
Pscheidl, line 23: This part should be dropped, as data in production do NOT contain those columns. The only meaningful way the "gamified" columns can reach this place is our tests. 

We can always remove the columns after GAM model training is finished by simply calling `trainBinomial.remove(gamCols);` - and test it on real-world-like data.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamRowToRawDataConverter.java
```diff
@@ -0,0 +1,32 @@
...
+      return super.convert(data, rawData);  // case where the input already contains the expanded gam columns
+    } else {
+      rawData = nanArray(_m.getTotFeatureSize());
+      rawData = super.convert(data, rawData);
```
Pscheidl, line 27: The `rawData = super.convert(data, rawData);` part fills all the columns but the gamified ones, and the `_m.addExpandGamCols(rawData, data);` should only fill the GAM coefs. Insteada, the latter somehow erases the original features and puts the gam coefficients at the beginning of the array while leaving everything else as NaN.

Screenshot from 2020-07-16 13-46-47

Screenshot from 2020-07-16 13-47-00


The screenshots above demonstrate this behavior.

h2o-genmodel/src/main/java/hex/genmodel/utils/DistributionFamily.java
wendycwong: @maurever:  Yes, I am sure that the DistributionFamily and LinkFunctionType are correct for GLM.  GLM does not have its own implementation but it does have its own names.  In addition, GAM defers to GLM to perform link function evaluation:

  double evalLink(double val) {
    switch (_link_function) {
      case identity: return GenModel.GLM_identityInv(val);
      case logit: return GenModel.GLM_logitInv(val);
      case log: return GenModel.GLM_logInv(val);
      case inverse: return GenModel.GLM_inverseInv(val);
      case tweedie: return GenModel.GLM_tweedieInv(val, _tweedieLinkPower);
      default: throw new UnsupportedOperationException("Unexpected link function "+_link_function);
    }
  }

I am not going to change GLM for now.  Plus we cannot change GLM family now anyway in order to avoid breaking use code.  I can make GAM mojo recognize Bernoulli though here.

h2o-algos/src/main/java/hex/gam/GAMModel.java
wendycwong: Removed static.  Thanks!

h2o-algos/src/main/java/hex/gam/GAMModel.java
wendycwong: I removed the function.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamRowToRawDataConverter.java
wendycwong: Fixed.  I hope.

Pscheidl:
@wendycwong GAM leaks a key when the gamified columns are not kept (the centered version). This makes several tests fail. Please apply the following patch to resolve the issue:

```patch
Index: h2o-algos/src/main/java/hex/gam/GAM.java
IDEA additional info:
<+>UTF-8
--- h2o-algos/src/main/java/hex/gam/GAM.java	(revision 4f2628539107b0d3456042c3aafe6919f426035e)
+++ h2o-algos/src/main/java/hex/gam/GAM.java	(date 1594975632892)
@@ -399,6 +399,8 @@
         if (model != null) {
           if (_parms._keep_gam_cols) {
             addFrameKeys2Keep(keep, newTFrame._key);
+          } else {
+            newTFrame.delete();
           }
           model.unlock(_job);
           Scope.untrack(keep);  // leave the vectors alone.
Index: h2o-algos/src/test/java/hex/gam/GamTestPiping.java
IDEA additional info:
<+>UTF-8
--- h2o-algos/src/test/java/hex/gam/GamTestPiping.java	(revision 4f2628539107b0d3456042c3aafe6919f426035e)
+++ h2o-algos/src/test/java/hex/gam/GamTestPiping.java	(date 1594975536044)
@@ -36,11 +36,9 @@
   // test and make sure that quasibinomial works.
   @Test public void testQuasibinomial(){
     Scope.enter();
-    GAMModel model = null;
-    Frame fr = parse_test_file("smalldata/glm_test/prostate_cat_replaced.csv");
-    Scope.track(fr);
     try {
-      GAMModel.GAMParameters params = new GAMModel.GAMParameters();
+      final Frame fr = Scope.track(parse_test_file("smalldata/glm_test/prostate_cat_replaced.csv"));
+      final GAMModel.GAMParameters params = new GAMModel.GAMParameters();
       params._response_column = "CAPSULE";
       params._family = quasibinomial;
       params._ignored_columns = new String[]{"ID"};
@@ -50,15 +48,9 @@
       params._gam_columns = new String[]{"PSA"};
       params._num_knots = new int[]{5};
       params._link = GLMModel.GLMParameters.Link.logit;
-      model = new GAM(params).trainModel().get();
-      System.out.println("wow");
+      final GAMModel model = new GAM(params).trainModel().get();
       Scope.track_generic(model);
     } finally {
-      fr.delete();
-      if(model != null){
-        model.deleteCrossValidationModels();
-        model.delete();
-      }
       Scope.exit();
     }
   }
```

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -0,0 +1,170 @@
...
+  int _betaSizePerClass;
+  int _betaCenterSizePerClass;
+  double _tweedieLinkPower;
+  double[][] _basisVals; // store basis values array for each gam column
+  double[][] _hj;   // difference between knot values
+  int _numExpandedGamCols; // number of expanded gam columns
+  int _lastClass;
+  
+  public int getTotFeatureSize() { return _totFeatureSize;}
+  public String[] getNames() { return _names; }  // get column names centered
```
honzasterba, line 56: can be removed, overrides already existing public method

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -0,0 +1,170 @@
...
+  int _totFeatureSize; // Gam Algo predictors numbers that include: predictors, expanded gam columns no centered
+  int _betaSizePerClass;
+  int _betaCenterSizePerClass;
+  double _tweedieLinkPower;
+  double[][] _basisVals; // store basis values array for each gam column
+  double[][] _hj;   // difference between knot values
+  int _numExpandedGamCols; // number of expanded gam columns
+  int _lastClass;
+  
+  public int getTotFeatureSize() { return _totFeatureSize;}
```
honzasterba, line 55: unnecessary public API, this is only used internally in this class we do not need to expose this to the user

wendycwong: Done

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: Done

Pscheidl:
@wendycwong I pushed a fix for the GBM JUnit test.

There are tests failing - `pyunit_check_model_scoring.py` and `GamTestPiping.testModelScoring` in Java - it seems to me one duplicates the other and vice versa.

The test is done against and outdated dataset with static predictions. This is strange thing to do, yet we should investigate why the predictions changed so much.

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -307,7 +295,6 @@ public MissingValuesHandling missingValuesHandling() {
 
     public final static double linkInv(double x, Link link, double tweedie_link_power) {
       switch(link) {
-//        case multinomial: // should not be used
```
honzasterba, line 310: remove commented out code

h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
```diff
@@ -3050,7 +3050,9 @@ public void testHuberNoise() {
   @Test
   public void testDeviances() {
     for (DistributionFamily dist : DistributionFamily.values()) {
-      if (dist == modified_huber || dist == quasibinomial || dist == ordinal || dist == custom) continue;
+      if (dist == modified_huber || dist == quasibinomial || dist == ordinal || dist == custom ||
```
honzasterba, line 3053: formating

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1979,31 +1979,6 @@ private void scoreAndUpdateModel(){
       Frame train = DKV.<Frame>getGet(_parms._train); // need to keep this frame to get scoring metrics back
       _model.score(train).delete();
       scoreEnd(train, t1);
-/*
```
honzasterba, line 1982: why is this commented out? was it replaced by some other code somewhere else?

h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
wendycwong: Done

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: The metrics computation is in scoreEnd now.

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoFactory.java
maurever: @wendycwong, WDYT about the removing these imports?

wendycwong:
@maurever :  Those imports are automatically added by intelliJ.  I tried to remove them and they come right back.  Please let me know how to get around IntelliJ and I can do it.  Thanks.

Pscheidl:
@wendycwong I synced the datasets, please check the tests once again (restarted). Especially have a look if there are any tests that generate output to tmp, as those will not be available on all nodes using `import` :wink:  - I have a suspicion I saw such tests there. Just asking.

wendycwong:
Pavel:  

You are right, it did happen once in a while.  It happened after I started using tempfile.  What would you propose to fix it?


Wendy

PR #4761, opened 2020-07-02
PUBDEV-6946 - Test R API of Isolation Forest Grid Search integration

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6946

- Add missing R test for already present functionality
- Rename the Python test to be more precise.


h2o-r/tests/testdir_algos/grid/runit_grid_isofor.R
```diff
@@ -0,0 +1,60 @@
...
+    h2o.grid(
+      "isolationforest",
+      grid_id = "isofor_grid_test_validation",
+      x = 1:4,
+      y = 5,
+      training_frame = train,
+      validation_frame = test,
+      hyper_params = hyper_parameters,
+      validation_response_column = "label"
+    )
```
satai, line 57: Maybe I miss something here but shouldn't be some kind assertion in the end too? Or does the second part of the test just check the code doesn't crash?

Pscheidl: Exactly. Not crashing it is fine :) Asserting the model is actually trained would be redundant really.

satai: OK. If ti doesnt'crash put it in production!

PR #4765, opened 2020-07-02
PUBDEV-7657: link proper base R manuals in R h2o manual

jangorecki:
https://0xdata.atlassian.net/browse/PUBDEV-7657

Can be tested on R-devel image using `_R_CHECK_XREFS_MIND_SUSPECT_ANCHORS_=true` env var.
Below example assumes `repos` generated by this PR. PR does not produce R repo one needs to put package sources manually in current directory.
```sh
docker pull registry.gitlab.com/jangorecki/dockerfiles/r-devel
docker run -it registry.gitlab.com/jangorecki/dockerfiles/r-devel /bin/bash
R -q -e 'install.packages(c("RCurl","jsonlite"))'
R -q -e 'download.packages("h2o", ".", repos="https://h2o-release.s3.amazonaws.com/h2o/PUBDEV-7657/5118/R")'
#no package ‘h2o’ at the repositories
export _R_CHECK_XREFS_MIND_SUSPECT_ANCHORS_=true
export _R_CHECK_FORCE_SUGGESTS_=false
R CMD check --ignore-vignettes --as-cran h2o_3.31.0.5118.tar.gz
```
And look for `checking Rd cross-references` line.

----

Moreover I think there is space for improvement, for example on merging manual pages. It is not really reasonable to have `log`, `log10`, `log1p`, `log2` each having own manual page. IMO it would be easier to write nice documentation when not trying to fit into roxygen templates.

### DIFF
--- a/h2o-r/h2o-package/R/frame.R
+++ b/h2o-r/h2o-package/R/frame.R
@@ -3354,7 +3354,7 @@ h2o.ischaracter <- function(x) {
 #'
 #' @name h2o.asfactor
 #' @param x An H2OFrame object.
-#' @seealso \code{\link[base]{numeric}} for the base R implementation, \code{as.factor()}.
+#' @seealso \code{\link[base]{factor}} for the base R implementation, \code{as.factor()}.
 #' @examples 
 #' \dontrun{
 #' library(h2o)
--- a/h2o-r/h2o-package/R/frame.R
+++ b/h2o-r/h2o-package/R/frame.R
@@ -2710,7 +2710,7 @@ median.H2OFrame <- h2o.median
 #'                             is ignored.
 #' @param return_frame \code{logical}. Indicate whether to return an H2O frame or a list. Default is FALSE (returns a list).
 #' @param ... Further arguments to be passed from or to other methods.
-#' @seealso \code{\link[base]{Round}} \code{\link[base]{colSums}} for the base R implementation, \code{colMeans()}.
+#' @seealso \code{\link[base]{Round}} for base R implementation, \code{mean()} and \code{\link[base]{colSums}} for the base R implementation, \code{colMeans()}.
 #' @return Returns a list containing the mean for each column (NaN for non-numeric columns) if return_frame is set to FALSE.
 #'         If return_frame is set to TRUE, then it will return an H2O frame with means per column or row (depends on axis argument).
 #' @examples
--- a/h2o-r/h2o-package/R/frame.R
+++ b/h2o-r/h2o-package/R/frame.R
@@ -3374,7 +3374,7 @@ h2o.asfactor <- function(x) {
 #'
 #' @name h2o.asnumeric
 #' @param x An H2OFrame object.
-#' @seealso \code{\link[base]{factor}} for the base R implementation, \code{as.numeric()}.
+#' @seealso \code{\link[base]{numeric}} for the base R implementation, \code{as.numeric()}.
 #' @examples 
 #' \dontrun{
 #' library(h2o)
--- a/h2o-r/h2o-package/R/frame.R
+++ b/h2o-r/h2o-package/R/frame.R
@@ -1934,7 +1934,7 @@ h2o.which <- function(x) {
 #' @param na.rm \code{logical}. Indicate whether missing values should be removed.
 #' @param axis \code{integer}. Indicate whether to calculate the mean down a column (0) or across a row (1).
 #' @return Returns an H2OFrame object.
-#' @seealso \code{\link[base]{which.min}} for the base R method.
+#' @seealso \code{\link[base]{which.min}} for the base R method, \code{which.max()}.
 #' @examples 
 #' \dontrun{
 #' library(h2o)
@@ -2710,7 +2710,7 @@ median.H2OFrame <- h2o.median
 #'                             is ignored.
 #' @param return_frame \code{logical}. Indicate whether to return an H2O frame or a list. Default is FALSE (returns a list).
 #' @param ... Further arguments to be passed from or to other methods.
-#' @seealso \code{\link[base]{mean}}, \code{\link[base]{colSums}} for the base R implementation
+#' @seealso \code{\link[base]{Round}} \code{\link[base]{colSums}} for the base R implementation, \code{colMeans()}.
 #' @return Returns a list containing the mean for each column (NaN for non-numeric columns) if return_frame is set to FALSE.
 #'         If return_frame is set to TRUE, then it will return an H2O frame with means per column or row (depends on axis argument).
 #' @examples
@@ -2815,7 +2815,7 @@ kurtosis.H2OFrame <- h2o.kurtosis
 #'   "everything"            - outputs NaNs whenever one of its contributing observations is missing
 #'   "all.obs"               - presence of missing observations will throw an error
 #'   "complete.obs"          - discards missing values along with all observations in their rows so that only complete observations are used
-#' @seealso \code{\link[stats]{cor}} for the base R implementation. \code{\link{h2o.sd}} for standard deviation.
+#' @seealso \code{\link[stats]{cor}} for the base R implementation, \code{var()}. \code{\link{h2o.sd}} for standard deviation.
 #' @examples
 #' \dontrun{
 #' library(h2o)
@@ -3000,7 +3000,7 @@ sd <- function(x, na.rm=FALSE) {
 #' @name h2o.signif
 #' @param x An H2OFrame object.
 #' @param digits Number of significant digits to round doubles/floats.
-#' @seealso \code{\link[base]{Round}} for the base R implementation.
+#' @seealso \code{\link[base]{Round}} for the base R implementation, \code{signif()}.
 #' @examples 
 #' \dontrun{
 #' library(h2o)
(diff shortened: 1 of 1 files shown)


ledell:
Hi @jangorecki Can you change this to merge into rel-zahradnik instead?  We need to get this fixed on the current release.  I think it shouldn't have any conflicts...

h2o-r/h2o-package/R/frame.R
```diff
@@ -1934,7 +1934,7 @@ h2o.which <- function(x) {
 #' @param na.rm \code{logical}. Indicate whether missing values should be removed.
 #' @param axis \code{integer}. Indicate whether to calculate the mean down a column (0) or across a row (1).
 #' @return Returns an H2OFrame object.
-#' @seealso \code{\link[base]{which.max}} for the base R method.
+#' @seealso \code{\link[base]{which.min}} for the base R method.
```
ledell, line 1937:
```suggestion
#' @seealso \code{\link[base]{which.max}} for the base R method.
```

ledell: @jangorecki Do you know why this was changed from max to min?

h2o-r/h2o-package/R/frame.R
```diff
@@ -2710,7 +2710,7 @@ median.H2OFrame <- h2o.median
 #'                             is ignored.
 #' @param return_frame \code{logical}. Indicate whether to return an H2O frame or a list. Default is FALSE (returns a list).
 #' @param ... Further arguments to be passed from or to other methods.
-#' @seealso \code{\link[base]{mean}} , \code{\link[base]{rowMeans}}, or \code{\link[base]{colMeans}} for the base R implementation
+#' @seealso \code{\link[base]{mean}} , \code{\link[base]{colSums}}, or \code{\link[base]{colMeans}} for the base R implementation
```
ledell, line 2713:
```suggestion
#' @seealso \code{\link[base]{mean}} , \code{\link[base]{rowMeans}}, or \code{\link[base]{colMeans}} for the base R implementation
```

ledell: @jangorecki Do you know why this was changed from rowMeans to colSums?

h2o-r/h2o-package/R/frame.R
```diff
@@ -2815,7 +2815,7 @@ kurtosis.H2OFrame <- h2o.kurtosis
 #'   "everything"            - outputs NaNs whenever one of its contributing observations is missing
 #'   "all.obs"               - presence of missing observations will throw an error
 #'   "complete.obs"          - discards missing values along with all observations in their rows so that only complete observations are used
-#' @seealso \code{\link[stats]{var}} for the base R implementation. \code{\link{h2o.sd}} for standard deviation.
+#' @seealso \code{\link[stats]{cor}} for the base R implementation. \code{\link{h2o.sd}} for standard deviation.
```
ledell, line 2818:
```suggestion
#' @seealso \code{\link[stats]{var}} for the base R implementation. \code{\link{h2o.sd}} for standard deviation.
```

ledell: @jangorecki Do you know why this was changed from var to cor?

h2o-r/h2o-package/R/frame.R
jangorecki: Manual page of `?which.max` is located under `which.min`. Documentation links should have always link manual pages rather than topics (`which.max` is topic), but it was never tested before. Now it is.

jangorecki:
I followed the advice on the mailing list about addressing this change. Unfortunately I don't have an environment on hand to build h2o package from this branch, so cannot fully test that.

jangorecki:
Testing revealed one more missing entry to fix.
Tested on the 5118 master after applying fix from this branch:
```sh
docker pull registry.gitlab.com/jangorecki/dockerfiles/r-devel
docker run -it registry.gitlab.com/jangorecki/dockerfiles/r-devel /bin/bash
apt-get -qq update
apt-get install -y libxml2-dev
R -q -e 'install.packages(c("RCurl","jsonlite","roxygen2"))'
R -q -e 'download.packages("h2o", ".", repos="https://h2o-release.s3.amazonaws.com/h2o/master/5118/R")'
tar xvf h2o_3.31.0.5118.tar.gz
mv h2o_3.31.0.5118.tar.gz h2o_3.31.0.5118.tar.gz_old
cd h2o
git init
wget https://gist.githubusercontent.com/jangorecki/9884a919188c06b320bc3231990f3cf4/raw/0adeec6b8f184db0a0005241a053d016104b0f60/h2o-patch.diff
sed -i 's+h2o-r/h2o-package/R+R+g' h2o-patch.diff
git apply h2o-patch.diff
cd ..
R -q -e 'roxygen2::roxygenize("h2o")'
R CMD build --no-build-vignettes h2o
export _R_CHECK_XREFS_MIND_SUSPECT_ANCHORS_=true
export _R_CHECK_FORCE_SUGGESTS_=false
R CMD check --ignore-vignettes --as-cran h2o_3.31.0.5118.tar.gz
```

h2o-r/h2o-package/R/frame.R
ledell: @jangorecki Ok yes, you're right.  Sorry I missed that.  I asked an R friend about this topic and this is what she said:

> This has been discussed a lot recently on the r-devel mailing list. Gabor Csardi is, I believe, preparing a patch to R itself that might solve the root problem and allow what many of us want to do (what you’re doing) to continue.
However, because it’s unclear how that will turn out and on what timescale, a workaround has already been put into roxygen2 v7.1.1, which was recently released:
https://roxygen2.r-lib.org/news/index.html#roxygen2-7-1-1-2020-06-27
If you make your docs with roxygen2, you can keep linking to the desired topic (or alias) and roxygen2 will look up the correct page or file when it builds the .Rd files.

So with that in mind, perhaps the least invasive solution (where we can keep the references to the exact functions that we want (even if they are only "topics"), would be to build the documentation with roxygen2 v7.1.1.  What do you think?

ledell:
@jangorecki Please see my comments [here](https://github.com/h2oai/h2o-3/pull/4765#discussion_r449988804) and let me know what you think.

You said you don't have an environment where you can test this branch... would you be able to build the h2o.tar.qz file from rel-zahradnik with the updated Roxygen2 installed (locally), and then test it if passes the CRAN check on r-devel just by using [R-hub](https://r-hub.github.io/rhub/reference/check_for_cran.html)?  

In order for me to get the latest version up on CRAN, this has to be fixed on rel-zahradnik.  Once everything looks good, then I was planning on manually patching the R files inside the existing h2o_3.30.0.6.tar.gz release because it will be a while until 3.30.0.7 comes out and this needs to be up on CRAN before Friday (UseR 2020).

h2o-r/h2o-package/R/frame.R
jangorecki: Building documentation with 7.1.1 didn't solve the problem. Not sure but turning links from Rd format to Markdown could help here, haven't try because it would be better to replace all links then rather than some of them.

jangorecki:
Environment for testing this change is in my previous post. Only replace h2o package there by changing master to required branch.
As for merging into other branch, you can checkout this branch into exactly same branch but new name, push it to GH and open PR against another branch. Should be easier than cherry picking. I don't have workstation on hand already.

h2o-r/h2o-package/R/frame.R
```diff
@@ -2710,7 +2710,7 @@ median.H2OFrame <- h2o.median
 #'                             is ignored.
 #' @param return_frame \code{logical}. Indicate whether to return an H2O frame or a list. Default is FALSE (returns a list).
 #' @param ... Further arguments to be passed from or to other methods.
-#' @seealso \code{\link[base]{mean}} , \code{\link[base]{rowMeans}}, or \code{\link[base]{colMeans}} for the base R implementation
+#' @seealso \code{\link[base]{Round}} \code{\link[base]{colSums}} for the base R implementation, \code{colMeans()}.
```
ledell, line 2713:
```suggestion
#' @seealso \code{\link[base]{Round}} for base R implementation, \code{mean()} and \code{\link[base]{colSums}} for the base R implementation, \code{colMeans()}.
```

h2o-r/h2o-package/R/frame.R
```diff
@@ -3374,7 +3374,7 @@ h2o.asfactor <- function(x) {
 #'
 #' @name h2o.asnumeric
 #' @param x An H2OFrame object.
-#' @seealso \code{\link[base]{as.factor}} for the base R implementation.
+#' @seealso \code{\link[base]{factor}} for the base R implementation, \code{as.numeric()}.
```
ledell, line 3377:
```suggestion
#' @seealso \code{\link[base]{numeric}} for the base R implementation, \code{as.numeric()}.
```

h2o-r/h2o-package/R/frame.R
```diff
@@ -3354,7 +3354,7 @@ h2o.ischaracter <- function(x) {
 #'
 #' @name h2o.asfactor
 #' @param x An H2OFrame object.
-#' @seealso \code{\link[base]{as.numeric}} for the base R implementation.
+#' @seealso \code{\link[base]{numeric}} for the base R implementation, \code{as.factor()}.
```
ledell, line 3357:
```suggestion
#' @seealso \code{\link[base]{factor}} for the base R implementation, \code{as.factor()}.
```

ledell (approved):
@jangorecki I added some extra info (I put the corresponding R base function name in `\code{}` to not lose context).  This is ready to go!

PR #4767, opened 2020-07-03
PUBDEV-6119  add tests for predictions with stratification

satai:
`stratify_by` seems to produce a different result than in R.

### DIFF
--- a/h2o-r/tests/testdir_algos/coxph/runit_coxph_predict.R
+++ b/h2o-r/tests/testdir_algos/coxph/runit_coxph_predict.R
@@ -2,97 +2,153 @@ setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
 source("../../../scripts/h2o-r-test-setup.R")
 
 
+pred.r <- function(fit.8, tstdata.8) {
+    fit.8.pred <- predict(fit.8, tstdata.8, type = "lp")
+    names(fit.8.pred) <- NULL
+    return(fit.8.pred)
+}
+
+pred.h2o <- function(model, data) {
+    hex.lp <- h2o.predict(model, data)
+    as.data.frame(hex.lp)$lp
+}
+
+compare.results <- function(fit.8, hex.fit.8, tstdata.8, tstdata.hex.8) {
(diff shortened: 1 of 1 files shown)


h2o-r/tests/testdir_algos/coxph/runit_coxph_predict.R
```diff
@@ -2,99 +2,153 @@ setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
...
-        lp[!is.na(lp)]
-    }
-
-    # Case 1: categoricals, numeric + interactions
-    tstdata.1 <- cancer
-    tstdata.1$sex <- as.factor(tstdata.1$sex)
+   
+
+    ## Case 1: categoricals, numeric + interactions
+    tstdata.1 <- cancer.with.sex()
```
honzasterba, line 32: I dont think you need tstdata.N, I think you should be fine with one frame being modified for each test if necessary

h2o-r/tests/testdir_algos/coxph/runit_coxph_predict.R
```diff
@@ -2,99 +2,153 @@ setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
 source("../../../scripts/h2o-r-test-setup.R")
 
 
+pred.r <- function(fit.8, tstdata.8) {
```
honzasterba, line 5: param names indicate this has something to do with cast #8, but I dont think thats the case, samoe for compare.results

satai: You are definitely right, fixing this.

h2o-r/tests/testdir_algos/coxph/runit_coxph_predict.R
satai: Technicaly, there is no need for it.

I would like to keep it this way anyway. Now is each test case structured in given-when-then feshion without sideeffect on data (and data is shared only when needed for comparisons - such as in the case that checks that equal weights don't afect the result).

PR #4768, opened 2020-07-04
Update models.R

Weekend-Warrior:
For partials of 2 variables, h2o.partialPlot is getting hung up when the pp tables' structure changes to more than one variable for partial dependence plotting. The original indexes are set so that the second column is the mean response value but this is not the case in a partial on more than 1 variable. To update this, we'll take the column dimension and index backwards from there. 

In the future this section can probably just be removed because I don't think it directly affects any of the functions that follow.

### DIFF
--- a/h2o-r/h2o-package/R/models.R
+++ b/h2o-r/h2o-package/R/models.R
@@ -4141,18 +4141,20 @@ h2o.partialPlot <- function(object, data, cols, destination_key, nbins=20, plot 
   ## Change feature names to the original supplied, the following is okay because order is preserved
       
   pps <- res$partial_dependence_data
-  min_y <- min(pps[[1]][,2])
-  max_y <- max(pps[[1]][,2])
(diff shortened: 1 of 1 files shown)


ledell:
@navdeep-G Adding you as a reviewer, as you may be most familiar with this code.  If you're not, no worries.

ledell:
@Weekend-Warrior Did you create a ticket for this on [JIRA](https://0xdata.atlassian.net/projects/PUBDEV)?  If not, no worries, we can create one (just wanted to check).

Weekend-Warrior:
Sorry, no I haven’t.

navdeep-G:
@ledell Thx. I think @wendycwong worked directly with the 2D PDP. So, I will add her as a reviewer.

maurever:
Hi @Weekend-Warrior. Thank you for your contribution. We currently don't have the capacity to assist you in completing this task. We decided that this change is not critical for improving our codebase, so we have closed this PR.

PR #4770, opened 2020-07-07
PUBDEV-7664 - Expose nbins_cat parameter for Isolation Forest

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7664

The `nbins_cat` parameter was not exposed. Exposing it now.

- Python tests
- R tests

Flow works:

Screenshot from 2020-07-07 08-36-19


Pscheidl:
Do not merge until @michalkurka approves !

honzasterba:
Pavel, you do not want this in the next release?

Pscheidl:
There was a consultation with @michalkurka and @meganjkurka and even though it is not written  anywhere, it was not exposed **on purpose** - I had no way to know that. So let's close the PR for now in order not to mess up with things and it can be re-opened by Michal Kurka once he comes back from vacation.

PR #4771, opened 2020-07-07
PUBDEV-4007 decision tree language rules printing

koniecsveta:
https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-4007


h2o-r/tests/testdir_jira/runit_pubdev_6488.R
```diff
@@ -58,6 +61,8 @@ test.tree.fetch <- function() {
   for (i in seq(1,ntrees)) {
     tree <- h2o.getModelTree(model, i)
     expect_false(is.null(tree))
+      expect_false(is.null(tree@language_tree_representation))
```
honzasterba, line 64: fix indent, also in rest of file

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -62,10 +63,107 @@ else if (!(possibleModel instanceof SharedTreeModel) && !(possibleModel instance
...
+            SharedTreeNode rightChild = node.getRightChild();
+            String stringToParseInclusiveLevelsFrom = rightChild.getInclusiveLevels().toString();
+            int inclusiveLevelsLength = rightChild.getInclusiveLevels().toString().length();
+            if (inclusiveLevelsLength  > 2) {
+                // get rid of curly braces:
+                stringToParseInclusiveLevelsFrom = stringToParseInclusiveLevelsFrom.substring(1, inclusiveLevelsLength - 1);
+                String[] inclusiveLevels = stringToParseInclusiveLevelsFrom.split(",");
+                Arrays.stream(inclusiveLevels)
+                        .map(String::trim)
+                        .map(Integer::parseInt)
```
Pscheidl, line 157:
```suggestion
                        .map(level -> Integer.parseInt(level.trim()))
```

One pass is better than two.

h2o-algos/src/main/java/hex/schemas/TreeV3.java
```diff
@@ -42,4 +42,9 @@
     @API(direction = API.Direction.OUTPUT, help = "Prediction values on terminal nodes")
     public float[] predictions;
 
+    @API(direction = API.Direction.OUTPUT, help = "Plain language rules representation of a trained decision tree")
+    public String language_tree_representation;
```
Pscheidl, line 46: Should the former `description` be removed and replaced by this ? Or `description` kept and the content change to this ? What's the point of previous `description` at the moment ? I'd get rid of it in favor of this.

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -175,6 +280,116 @@ private static void append(final int[] rightChildren, final int[] leftChildren,
                 discoveredNodes, pointer, true);
     }
 
+    private static ArrayList<Integer> extractInternalIds(TreeProperties properties) {
```
Pscheidl, line 283: Name of the method is misleading. This method also replaces the internal node IDs (which is not a sequence with equal distances) with a continuous equidistant sequence. Which is from user's POV more usable. I suggest to add some javadoc at least.

One question: Do we really want to replace those ? What about other tree visualization tools ? Will those node ids/numbers be comparable ?

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -175,6 +280,116 @@ private static void append(final int[] rightChildren, final int[] leftChildren,
...
+        if (!valuePrinted) {
+            // print prediction value
+            nodePathr += "Predicted value: " + properties._predictions[currentNodeIndex] + "\n";
+            valuePrinted = true;
+            nodePathr += fillNodePath(nodeId, nodeIds, valuePrinted, properties, domainValues);
+        } else {
+            // print conditions leading to prediction value
+            int[] leftChildren = properties._leftChildren;
+            int[] rightChildren = properties._rightChildren;
+            if (IntStream.of(leftChildren).anyMatch(i -> i == currentNodeIndex)) {
```
Pscheidl, line 326: How efficient is `anyMatch` on IntStream ? Looks like it translates to `ForEachOp`. At least binary search would be nice - the array `lefChildren` should be sorted in ascending order, there should NOT be smaller numbers at lower index than there are at higher index. We have it implemented as `Array.binarySearch`.

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -175,6 +280,116 @@ private static void append(final int[] rightChildren, final int[] leftChildren,
...
+                // parent from right
+                parentIndex = IntStream.range(0, leftChildren.length).filter(i -> leftChildren[i] == currentNodeIndex).findAny().getAsInt();
+                parentId = nodeIds.get(parentIndex);
+                condition = getConditionByIndex(parentIndex, "R", properties, domainValues);
+            }
+            if (IntStream.of(rightChildren).anyMatch(i -> i == currentNodeIndex)) {
```
Pscheidl, line 332: Same as above, I suggest more effective search.

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -175,6 +280,116 @@ private static void append(final int[] rightChildren, final int[] leftChildren,
                 discoveredNodes, pointer, true);
     }
 
+    private static ArrayList<Integer> extractInternalIds(TreeProperties properties) {
+        int nodeId = 0;
+        ArrayList<Integer> nodeIds = new ArrayList<>();
```
Pscheidl, line 285: For large trees/forests, having an array list of `Integer` might actually be much more memory intensive than just having `int[]` - the (typically 64bit) pointers are omitted. The price to pay for that in the current state seems to be passing through the children twice. The antidote to that seems to be making this part of the original recursion.

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -119,12 +220,16 @@ static TreeProperties convertSharedTreeSubgraph(final SharedTreeSubgraph sharedT
...
+        treeprops._languageTreeRepresentation = getLanguageRepresentation(sharedTreeSubgraph);
+        treeprops._languagePathsRepresentations[0] = "Predicted value: " + sharedTreeSubgraph.rootNode.getPredValue();
+
 
         List<SharedTreeNode> nodesToTraverse = new ArrayList<>();
         nodesToTraverse.add(sharedTreeSubgraph.rootNode);
         append(treeprops._rightChildren, treeprops._leftChildren,
                 treeprops._descriptions, treeprops._thresholds, treeprops._features, treeprops._nas,
                 treeprops.levels, treeprops._predictions, nodesToTraverse, -1, false);
+        fillLanguagePathRepresentation(treeprops, getDomainValues(sharedTreeSubgraph));
```
Pscheidl, line 232: As far as I can see, why aren't the new descriptions filled as a part of the original recursion ? It should have everything required, only one new parameter with domain values extracted is required - the `getDomainValues(sharedTreeSubgraph)` can be done before `append` is called for the first time.

Am I missing something ?

h2o-py/tests/testdir_algos/gbm/pyunit_4007_decision_tree_language_rules_printing.py
```diff
@@ -0,0 +1,67 @@
...
+    print(" ----- categorical case: -----")
+    airlines_data = h2o.import_file(path=pyunit_utils.locate("smalldata/airlines/allyears2k_headers.zip"))
+    model = H2OGradientBoostingEstimator(ntrees = 3, max_depth = 2)
+    model.train(x=["Origin", "Distance"], y="IsDepDelayed", training_frame=airlines_data)
+    tree = H2OTree(model = model, tree_number = 0, tree_class = "NO")
+
+    print(" -- Tree predictions: -- ")
+    print(tree.predictions)
+    print(" -- Language tree representation: -- ")
+    assert tree.language_tree_representation is not None
```
Pscheidl, line 55: We might iterate over the language representation and check for presence of certain words, e.g. prediction for terminal nodes and split description for others etc ? Just a substring matches to make sure the crucial value is there in the description.

h2o-algos/src/main/java/hex/tree/TreeHandler.java
honzasterba: I think zuzana's code is more reasable and the efficiency is the same

Pscheidl: After some hidden optimizations maybe, I'm not sure. The code shouts to map twice to me.

h2o-py/tests/testdir_algos/gbm/pyunit_4007_decision_tree_language_rules_printing.py
```diff
@@ -0,0 +1,84 @@
...
+                 validation_frame=train)
+    first_tree = H2OTree(model = my_gbm, tree_number = 0, tree_class = None)
+
+    print(" -- Tree predictions: -- ")
+    print(first_tree.predictions)
+    
+    print(" -- Language tree representation: -- ")
+    assert first_tree.language_tree_representation is not None
+    print(first_tree.language_tree_representation)
+    text_file = open("pyunit_4007_language_tree_representation_numerical_case.txt", "r")
```
honzasterba, line 35: This expects the working dir to be in the test dir, but I think it is not, pyunit_utils.locate should resolve that

h2o-py/tests/testdir_algos/gbm/pyunit_4007_decision_tree_language_rules_printing.py
```diff
@@ -0,0 +1,84 @@
...
+    assert tree.language_path_representations[tree.root_node.id] is not None
+    print(tree.language_path_representations[tree.root_node.id])
+    
+    print(" -- Language path representation - node ", tree.predictions.index(tree.predictions[3]), " (with pv = ", tree.predictions[3], "): -- ")
+    assert tree.language_path_representations[tree.predictions.index(tree.predictions[3])] is not None
+    print(tree.language_path_representations[tree.predictions.index(tree.predictions[3])])
+    text_file = open("pyunit_4007_language_path_representation_categorical_case.txt", "r")
+    expected_tree_representation = text_file.read()
+    text_file.close()
+    assert expected_tree_representation == tree.language_path_representations[tree.predictions.index(tree.predictions[3])]
```
honzasterba, line 79: repeats a bunch of times what about assert read_fixture("file_name.txt") == tree.language_path...

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -62,10 +63,106 @@ else if (!(possibleModel instanceof SharedTreeModel) && !(possibleModel instance
...
 
         return args;
     }
 
+
+    private static String getLanguageRepresentation(SharedTreeSubgraph sharedTreeSubgraph) {
+        return getNodeRepresentation(sharedTreeSubgraph.rootNode, "", 0);
+    }
+
+    private static String getNodeRepresentation(SharedTreeNode node, String languageRepresentation, int padding) {
```
honzasterba, line 77: language representation should be StringBuilder

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -62,10 +63,106 @@ else if (!(possibleModel instanceof SharedTreeModel) && !(possibleModel instance
...
+        if (node.getLeftChild() != null || node.getRightChild() != null) {
+            if (padding != 0)
+                languageRepresentation += getNewPaddedLine(padding);
```
honzasterba, line 80: this is padding for the condition? I think it would make this more readable if getConditionLine and getNodeRepresentation already returned with padding

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -62,10 +63,106 @@ else if (!(possibleModel instanceof SharedTreeModel) && !(possibleModel instance
...
+            languageRepresentation += getConditionLine(node);
+            if (node.getRightChild() != null) {
+                languageRepresentation += getNewPaddedLine(padding);
+                languageRepresentation = getNodeRepresentation(node.getRightChild(), languageRepresentation, padding + 1);
+            }
+            languageRepresentation += getNewPaddedLine(padding);
+            languageRepresentation += getElseLine(node);
```
honzasterba, line 87: I thnk getLeftChild is never null if it were, the else branch would be empty and the "code" invalid

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -62,10 +63,106 @@ else if (!(possibleModel instanceof SharedTreeModel) && !(possibleModel instance
...
+                String[] inclusiveLevels = stringToParseInclusiveLevelsFrom.split(",");
+                for (String index : inclusiveLevels) {
+                    elseLine.append(node.getDomainValues()[Integer.parseInt(index.trim())] + " ");
+                }
+            } else {
+                elseLine.append("Missing set of levels for underlying node");
+            }
+            elseLine.append("]) {");
+        }
+        return elseLine.toString();
```
honzasterba, line 133: why not return the builder and append it to parrent builder?

h2o-r/tests/testdir_jira/runit_pubdev_6488.R
koniecsveta: fixed

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: Changed so that we don't replace but have private "normalized" versions on TreeProperties which I can use in language representation calculation

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: fixed

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: fixed

h2o-py/tests/testdir_algos/gbm/pyunit_4007_decision_tree_language_rules_printing.py
koniecsveta: fixed

h2o-py/tests/testdir_algos/gbm/pyunit_4007_decision_tree_language_rules_printing.py
koniecsveta: fixed

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: fixed

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: fixed

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: fixed

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: fixed

h2o-py/tests/testdir_algos/gbm/pyunit_4007_decision_tree_language_rules_printing.py
koniecsveta: fixed / did test against a text file

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: fixed so that it returns int[]. I would leave making this part of original recursion for future optimization.

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: would leave this as is

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: binary search can't be used - there are cases when properties._leftChildren is like ... 62 64 66 -1 68 70 72 ... 86 88 90 -1 -1 -1 ... -> properties._leftChildrenNormalized are created like ...31 33 35 -1 37 39 ... 55 57 59 -1 -1 -1 ... hence even when I cut -1's from right it is possible that leftChildren/_leftChildrenNormalized  is not sorted and I can't sort it because it would invalidate information... (this is the case from pyunit_tree.py which is failing after this change)

Pscheidl: I see. Let's have it this way then and optimize later.

h2o-algos/src/main/java/hex/tree/TreeHandler.java
Pscheidl: The real-world experience would be the same, it's highly unlikely for anyone to notice even on larger trees.

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: That was my original intention but in getConditionByIndex() I use not only current node index when accessing information in tree properties, but parentIndex and targetIndex. in append() there is a "pointer" index by which it is possible to set up tree infos of current node. I think there might be a way how to do this, but not entirely straightforward - would leave for next PR or put this PR on hold.

h2o-algos/src/main/java/hex/schemas/TreeV3.java
koniecsveta: descriptions deprecated

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -172,14 +288,138 @@ private static void append(final int[] rightChildren, final int[] leftChildren,
...
+
+    private static void fillLanguagePathRepresentation(TreeProperties properties) {
+        ArrayList<Integer> nodeIds = extractInternalIds(properties);
+        nodeIds.forEach((list_path_id) -> {
+                    int index = nodeIds.indexOf(list_path_id);
+                    properties._decisionPaths[index] = fillNodePath(list_path_id, nodeIds, false, properties);
+                });
+    }
+
+    private static String fillNodePath(int nodeId, ArrayList<Integer> nodeIds, boolean valuePrinted, TreeProperties properties) {
```
honzasterba, line 326: use List in methods

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -62,10 +63,112 @@ else if (!(possibleModel instanceof SharedTreeModel) && !(possibleModel instance
...
         args.levels = treeProperties.levels;
         args.predictions = treeProperties._predictions;
-
+        args.tree_decision_path = treeProperties._treeDecisionPath;
+        args.decision_paths = treeProperties._decisionPaths;
         return args;
     }
 
+
+    private static String getLanguageRepresentation(SharedTreeSubgraph sharedTreeSubgraph) {
```
honzasterba, line 72: I have just noticed that all this code is in Handler class, this is not ideal, we want Handlers to be skinny as possible whith actual logic residing in something thats easier to unit test

h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -172,14 +288,138 @@ private static void append(final int[] rightChildren, final int[] leftChildren,
...
+            } else {
+                properties._rightChildrenNormalized[i] = -1;
+            }
+        }
+        return nodeIds;
+    }
+
+    private static void fillLanguagePathRepresentation(TreeProperties properties) {
+        ArrayList<Integer> nodeIds = extractInternalIds(properties);
+        nodeIds.forEach((list_path_id) -> {
```
honzasterba, line 320: `list_path_id` please use Java naming convention

koniecsveta: fixed

h2o-algos/src/main/java/hex/tree/TreeHandler.java
koniecsveta: fixed

h2o-algos/src/main/java/hex/tree/TreeHandler.java
Pscheidl: This is what I did. Not her fault.  It'd a dedicated handler, moving it to other class would result in ..one more class created. Could be cleaner if named accordingly. :wink:

koniecsveta: as discussed, refactor will be done in a separate PR

PR #4773, opened 2020-07-07
[PUBDEV-7283] save frame to hive - intial impl

honzasterba:
- things to improve - save each node its local chunks and import from a directory


h2o-hive/src/main/java/water/hive/HiveFrameSaverImpl.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private String makeCreateTableStatement(String table, Frame frame) {
+        StringBuilder sb = new StringBuilder();
+        sb.append("CREATE TABLE ").append(table).append(" (");
+        for (int i = 0; i < frame.numCols(); i++) {
+            if (i > 0) sb.append(",\n");
+            sb.append(frame.name(i)).append(" ").append(sqlType(frame.vec(i).get_type()));
+        }
+        sb.append(") ROW FORMAT SERDE 'org.apache.hadoop.hive.serde2.OpenCSVSerde'\n" +
```
Pscheidl, line 138: This is ok, because `https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java` - I'll take this information from SO as a fact.

PR #4778, opened 2020-07-09
[PUBDEV-7283] add correct escaping to hive CSV

honzasterba:


h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1638,6 +1640,18 @@ public CSVStream(Chunk[] chks, String[] names, int nChunks, CSVStreamParams parm
       _escapedCategoricalVecDomains = escapeCategoricalVecDomains(_curChks);
     }
 
+    private static Chunk[] firstChunks(Frame fr) {
```
honzasterba, line 1643: just moved after constructors

PR #4780, opened 2020-07-09
PUBDEV-7680: added `modified_huber` & `ordinal` to distribution.rst

hannah-tillman:


h2o-docs/src/product/data-science/algo-params/distribution.rst
```diff
@@ -22,7 +22,9 @@ Certain cases can exist, however, in which the median starting value for this lo
 - If the distribution is ``laplace``, the response column must be numeric.
 - If the distribution is ``quantile``, the response column must be numeric.
 - If the distribution is ``huber``, the response column must be numeric.
+- If the distribution is ``modified_huber``, the response column must be 2-class categorical.
 - If the distribution is ``tweedie``, the response column must be numeric.
+- If the distribution is ``ordinal``, the response column must be 3-class categorical.
```
mn-mikke, line 27: Aren't more classes supported as well?

honzasterba: must be at least 3-class categorical

https://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/algo-params/family.html?highlight=ordinal

wendycwong: I think you meant to say this:
If the distribution is ``ordinal``, the response column must be categorical with at least 3 levels.

PR #4781, opened 2020-07-09
PUBDEV-7679: Document 'negativebinomial' Value on the Family Page

hannah-tillman:


mn-mikke:
cc @ledell @wendycwong @maurever @honzasterba @Pscheidl

h2o-docs/src/product/data-science/algo-params/family.rst
```diff
@@ -26,6 +26,7 @@ You can specify one of the following ``family`` options based on the response co
 - ``poisson``: The data must be numeric and non-negative (Int).
 - ``gamma``: The data must be numeric and continuous and positive (Real or Int).
 - ``tweedie``: The data must be numeric and continuous (Real) and non-negative.
+- ``negativebinomial``: The data must be numeric and non-negative (Int).
```
honzasterba, line 29: I would think that this would mean binomial with negative values,
@wendycwong please clarify

hannah-tillman: If it is, I will have to change it in http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/glm.html#families as well 👍

wendycwong: The response data must be numeric and non-negative integers like 0,1,2,3,...  So the description is actually correct.

PR #4789, opened 2020-07-14
[PUBDEV-7467] improvements in Steam XGBoost integration

honzasterba:
- fix deadlock when steam requests stop while a cluster is starting
- improve response handling, ignore messages not related to xgboost

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/remote/SteamExecutorStarter.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/remote/SteamExecutorStarter.java
@@ -67,14 +67,18 @@ public class SteamExecutorStarter implements SteamMessenger {
         Map<String, String> startRequest = makeStartRequest();
         sendMessage(startRequest);
         Map<String, String> response = waitForResponse(startRequest, job);
-        if (response != null && "started".equals(response.get("status"))) {
-            String remoteUri = response.get("uri");
-            String userName = response.get("user");
-            String password = response.get("password");
-            cluster = new ClusterInfo(remoteUri, userName, password);
-            LOG.info("External cluster started at " + remoteUri + ".");
+        if (response != null) {
+            if ("started".equals(response.get("status"))) {
+                String remoteUri = response.get("uri");
+                String userName = response.get("user");
+                String password = response.get("password");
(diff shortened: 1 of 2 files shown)


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/remote/SteamExecutorStarter.java
```diff
@@ -67,14 +67,18 @@ private void startCluster(Job<XGBoostModel> job) throws IOException {
...
         sendMessage(startRequest);
         Map<String, String> response = waitForResponse(startRequest, job);
-        if (response != null && "started".equals(response.get("status"))) {
-            String remoteUri = response.get("uri");
-            String userName = response.get("user");
-            String password = response.get("password");
-            cluster = new ClusterInfo(remoteUri, userName, password);
-            LOG.info("External cluster started at " + remoteUri + ".");
+        if (response != null) {
+            if ("started".equals(response.get("status"))) {
```
Pscheidl, line 71: Could be final.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/remote/SteamExecutorStarter.java
```diff
@@ -112,40 +117,51 @@ private synchronized void sendMessage(Map<String, String> message) throws IOExce
...
+    private void queueMessageIfResponse(Map<String, String> message) {
         synchronized (receivedMessages) {
-            receivedMessages.put(message.get(ID), message);
-            receivedMessages.notifyAll();
+            if (receivedMessages.containsKey(message.get(ID))) {
+                LOG.info("Received message response " + message.get(ID));
+                receivedMessages.put(message.get(ID), message);
+                receivedMessages.notifyAll();
+            } else {
+                LOG.debug("Ignoring message " + message.get(ID));
```
Pscheidl, line 137: Should we log the ignored ones at info-level as well ?

honzasterba: No since the steam servlet will distribute all messages to all listeners, we do not need to log messages that an listerner ignores

PR #4794, opened 2020-07-17
PUBDEV-7489 - Impossible to set Max_depth to unlimited on DRF classifer

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7489

The JIRA title is a little bit misleading. The maximum possible depth at the moment is limited by Java int, whish is `2^31`. And by providing 0, H2O should initially set `max_depth` to the maximum possible value.

Doesn't affect only DRF, affected algos are:

- XGBoost
- GBM
- IF
- DRF

Kept the default check in SharedTreeModel to `> 0` - all algos supporting this behavior must override.

Another **option** is to change the documentation and not support this behavior. The JIRA is marked by me to go to `3.30.1.1`, yet it's not absolutely necessary for it to make it there. It's potentially easy, but if there is any controversy about our API style, then we should postpone it.


h2o-algos/src/main/java/hex/tree/SharedTree.java
```diff
@@ -165,6 +165,11 @@ protected void ignoreInvalidColumns(int npredictors, boolean expensive) {
 
     PlattScalingHelper.initCalibration(this, _parms, expensive);
   }
+  
+  
+  protected void validateMaxDepth(){
+    if (_parms._max_depth <= 0) error ("_max_depth", "_max_depth must be > 0.");
```
sebhrusen, line 171: why chosing this default implementation for `validateMaxDepth` here if you override it with the same logic in all subclasses?

h2o-algos/src/main/java/hex/tree/gbm/GBM.java
```diff
@@ -184,6 +184,12 @@
     }
   }
 
+  @Override
+  protected void validateMaxDepth() {
+    if (_parms._max_depth < 0) error ("_max_depth", "_max_depth must be >= 0.");
+    if(_parms._max_depth == 0) _parms._max_depth = Integer.MAX_VALUE;
```
sebhrusen, line 190: I would set to unlimited as soon as `max_depth <= 0`, raising on negative looks unnecessary.

h2o-algos/src/main/java/hex/tree/isofor/IsolationForest.java
```diff
@@ -103,6 +103,12 @@
     }
   }
 
+  @Override
+  protected void validateMaxDepth() {
+    if (_parms._max_depth < 0) error ("_max_depth", "_max_depth must be >= 0.");
+    if(_parms._max_depth == 0) _parms._max_depth = Integer.MAX_VALUE;
```
koniecsveta, line 109: just a nitpick: missing space (also in GBM.java, DRF.java, XGBoost.java)

h2o-algos/src/main/java/hex/tree/SharedTree.java
honzasterba: I am with seb on this. also the override is the same in all subclasses, so ti does not make much sense.
also  agree that forcing 0 to mean unlimited is uneccessary I think <=0 is better for the user

Pscheidl: Yes, the override is in all classes and can be replaced.

Regarding the convention - we kind-of agreed with @sebhrusen (also what @michalkurka says) that 0 means disabled and -1 is automatic behavior. In this case, disabling max_depth restriction implies making the trees as deep as technically possible. Therefore 0. Still valid @sebhrusen ? See api-updates channel :)

sebhrusen: Yes, I'm ok with the distinction `disabled` (=0) vs `auto` (<0).
My original argument was that for some use-cases, the distinction is not always clear, although here, we can imagine a case were "auto" would mean "use an estimated value". TBH, it's all theoretical because such a feature doesn't exist, but it shows that we may want to use negative values for a different semantic in the future.

PR #4796, opened 2020-07-18
R 4.0 tests fixes

honzasterba:


h2o-r/tests/testdir_algos/gbm/runit_GBM_quantile.R
```diff
@@ -64,7 +64,8 @@ gbm.quantile.test = function(){
   #a test to look at h2o.quantile() - init_F is equal to POJO splits for one node.
   left_q = h2o.quantile(data.hex.test.left[,resp],Q) 
   left_q_test = left_q - m@model$init_f
-  print(paste("Quantile GBM left split", left_q_test, LEFT_POJO))
+  left_q_test
+  LEFT_POJO
```
sebhrusen: this does nothing, does it?

honzasterba: true, eventually I ended up not touching this test

h2o-r/tests/runitUtils/utilsR.R
```diff
@@ -409,6 +410,18 @@ function(..., envir=parent.frame()) {
     list(tests=tests, envir=envir)
 }
 
+setSeedPriorR36<-
+function(seed) {
+    if (as.numeric(R.Version()$major) >= 4 || 
+        (as.numeric(R.Version()$major) == 3 && as.numeric(substr(R.Version()$minor, 1, 1)) >= 6)
```
sebhrusen, line 416: strictly speaking, to recover the minor version (imagine 3.11.5), you should use sth like
```
as.numeric(unlist(strsplit(R.Version()$minor, '.', fixed=TRUE))[[1]])
```

h2o-r/tests/testdir_misc/runit_quantile.R
```diff
@@ -1,46 +1,44 @@
...
-aa$h_wts = as.h2o(wts)
-#r_q = wtd.quantile(x, wts, probs = seq(0,1,.05))
-r_q=c(3.895489e-06,4.863379e-02,9.789691e-02,1.470487e-01,1.977443e-01,2.473365e-01,2.975013e-01,3.482667e-01,3.980460e-01,4.483631e-01,4.990024e-01,5.489128e-01,5.986945e-01,6.486255e-01,6.991498e-01,7.500031e-01,8.001472e-01,8.504057e-01,8.996923e-01,9.498159e-01,9.999471e-01)
-h_q = h2o.quantile(aa,probs = seq(0,1,.05),weights_column = "h_wts")
-expect_true(max(abs((r_q-h_q)/r_q)) < 1e-5)
+#This tests quantile and weighted quantile on synthetic data by comparing with R
+test.quantile <- function() {
 
+    # set random seed to generate random dataset
+    set.seed(1234)
```
sebhrusen, line 8: so there are cases for which `set.seed` is reproducible accross all versions, and some for which we should rely on `setSeedPriorR36`?
How do we know? Shouldn't we make `setSeedPriorR36` our default?
```
setSeed <- set.seed
set.seed <- setSeedPriorR36  # use setSeed internally
```

honzasterba: good idea, will add

h2o-r/tests/runitUtils/utilsR.R
honzasterba: fixed

PR #4798, opened 2020-07-20
PUBDEV-7578 fix grid model seeds pyunit test

maurever:
Set `score_tree_interval` parameter, which has to be set to reproduce GBM on singlenode and multinode too.

JIRA: https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7578

### DIFF
--- a/h2o-py/tests/testdir_algos/gbm/pyunit_PUBDEV_7578_gbm_reproducibility.py
+++ b/h2o-py/tests/testdir_algos/gbm/pyunit_PUBDEV_7578_gbm_reproducibility.py
@@ -1,6 +1,4 @@
 from __future__ import print_function
-from builtins import range
-import math
 import sys
 sys.path.insert(1,"../../../")
 import h2o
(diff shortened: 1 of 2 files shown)


h2o-py/tests/testdir_algos/gbm/pyunit_PUBDEV_7578_gbm_reproducibility.py
```diff
@@ -0,0 +1,240 @@
...
+    model_seed_list_string = ','.join(str(x) for x in model_seed_list[0:model_num])
+    print(model_seed_list_string)
+    model_seed_list.sort()
+    print("model seed list after sort: ")
+    model_seed_list_string = ','.join(str(x) for x in model_seed_list[0:model_num])
+    print(model_seed_list_string)
+    return model_seed_list
+
+
+def standalone_test(test):
```
honzasterba, line 222: duplicates pyunit_utils, why?

h2o-py/tests/testdir_algos/gbm/pyunit_PUBDEV_7578_gbm_reproducibility.py
```diff
@@ -0,0 +1,240 @@
...
+    rmse1 = extract_from_twoDimTable(gbm1._model_json["output"]["scoring_history"], "training_rmse", False)
+    rmse2 = extract_from_twoDimTable(gbm2._model_json["output"]["scoring_history"], "training_rmse", False)
+    print(rmse1)
+    print(rmse2)
+
+    assert equal_two_arrays(rmse1, rmse2, 1e-5, 1e-6, False), \
+        "Training_rmse are different between the two grid search models.  Tests are supposed to be repeatable in " \
+        "this case.  Make sure model seeds are actually set correctly in the Java backend."
+
+def equal_two_arrays(array1, array2, eps, tolerance, throwError=True):
```
honzasterba, line 123: shoud we move this to utils?

h2o-py/tests/testdir_algos/gbm/pyunit_PUBDEV_7578_gbm_reproducibility.py
maurever: Sorry, I just use this attached test from the JIRA ticket without paying attention to the whole code... Thank you for your suggestion. I will change it.

h2o-py/tests/testdir_algos/gbm/pyunit_PUBDEV_7578_gbm_reproducibility.py
maurever: It is copied from pyunit_utils. I did not catch this when I copied this test from the JIRA ticket attachments... Fixed.

maurever:
> Is this solving the same issue as was described on the ticket? Because the one which is described there happens on 1 node - it's visible there from the copy of output in ticket description.

I think yes. From the scoring history, we can see, there is problem with early stopping. The `score_tree_interval ` has to be set also for single node to reproduce same result for GBM. More information in our doc: http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/gbm-faq/reproducibility.html

h2o-py/tests/testdir_algos/gbm/pyunit_PUBDEV_7578_gbm_reproducibility.py
```diff
@@ -0,0 +1,113 @@
+from __future__ import print_function
+import sys
+sys.path.insert(1,"../../../")
+import h2o
+from tests import pyunit_utils
+from h2o.estimators.gbm import H2OGradientBoostingEstimator
+
+
+def random_grid_model_seeds_failing_case():
```
sebhrusen, line 9: can you please explain?
did you obtain the params below from a grid search and this specific model was not reproduceable, even when specifying the same seed?

sebhrusen: is this test here to narrow down the reproducibility issue (to better analyze next time it fails) or was there a fix associated to it?

maurever: Yes, exactly. This case found @koniecsveta and reproduce the error more often than with a random grid search. However, this case still is not always reproducible. Sometimes GBM is not reproducible if `score_tree_interval` is not set. 

This test is here to be sure, we fix the issue. But I can remove it if you think it is not useful.

sebhrusen: by failing test, I just mean:
```
try:
    should_fail()
except Exception as e:
    assert "expected message" in str(e)

maurever: I think catching the failing case like ^^^ is not correct. The first problem (still the same), it will sometimes fail because not every time it should fail even the early stopping parameters are not set correctly (reproducibly). The second problem I don't see useful to keep this unpredictable failing case because I don't have idea how to keep it. 

I spent more time to investigate why `score_tree_interval` parameter setting is important for reproducibility. If a user doesn't set this parameter, a time-based duty-cycle heuristic is used to calculate the moment when to score. And then, the scoring result for training two identical models can differ. If the user set  `score_tree_interval = 1` or  `score_tree_interval = 2`, the heuristic is not used, and the model score even when one respectively two trees are built.

See: https://github.com/h2oai/h2o-3/blob/master/h2o-algos/src/main/java/hex/tree/SharedTree.java#L715-L720

As the documentation [says](http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/gbm-faq/reproducibility.html): No early stopping performed or early stopping with `score_tree_interval` is explicitly specified and the same validation data. And this holds for single node and multinode too.

PR #4803, opened 2020-07-21
[PUBDEV-7649] make hadoop docker images S3 independent

honzasterba:


honzasterba:
adding dummy numbers to enable SORT and MERGE benchmarks to actually execute and store

PR #4809, opened 2020-07-24
Fix Gam Mojo bug when NaN appears in gam_columns

wendycwong:
@pavel has discovered a bug in Gam Mojo when there is NaN in gam columns.

We discussed this and noticed the following:
1. for RowData rowData, it will drop any column that contains NaN.  If you have columns C1, C2, C3 and C2 is NaN, then RowData will only return you C1, C3.
2. Wendy relies on the predictors present in RowData to determine if gamification of gam_columns are needed.  She did not take into account of the absence of a column due to NaN.  Hence, the wrong decision is made in addExpandGamCols.

I have added a fix to not use RowData information to make the decision and I hope this will fix the problem.

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
@@ -62,10 +62,10 @@ public abstract class GamMojoModelBase extends MojoModel implements ConverterFac
     super(columns, domains, responseColumn);
   }
 
-  public void extractInputDataNames(String[] cNames) {
-    final List<String> colHeaders = new ArrayList<>(Arrays.asList(cNames));
-    for (String gCol : _gam_columns) {
-      if (!colHeaders.contains(gCol)) {
+  void extractInputDataNames(String[] columnNames) {
+    final List<String> names = Arrays.asList(columnNames);
+    for (String gamColumnName : _gam_columns) {
+      if (!names.contains(gamColumnName)) {
         _inputContainGamCols = false;
         return;
       }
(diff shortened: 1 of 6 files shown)


Pscheidl:
@wendycwong Please wait for tests before merging, locally they ALL went fine (including JUnit and Python).

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -133,28 +148,29 @@ int readCatVal(double data, int dataIndex) {
   }
   
   // this method will add to each data row the expanded gam columns
-  public double[] addExpandGamCols(double[] rawData, final RowData rowData) { // add all expanded gam columns here
+  double[] addExpandGamCols(double[] rawData, final RowData rowData) { // add all expanded gam columns here
+    if (!_inputContainGamCols) {
```
honzasterba, line 152: This `if` confuses me, something is off here `if not _inputContainGamCols` do nothing, `else` add gam cols. Is the name of the variable wrong or am I just not understanding what is happening here?

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -51,13 +54,25 @@
...
   int _lastClass;
+  boolean _inputContainGamCols; // true if input dataset contains gam column names
   
   private int getTotFeatureSize() { return _totFeatureSize;}
   
   GamMojoModelBase(String[] columns, String[][] domains, String responseColumn) {
     super(columns, domains, responseColumn);
   }
 
+  void extractInputDataNames(String[] columnNames) {
```
honzasterba, line 65: could use a better name, since the current name has nothing to do what this does, maybe `checkNamesContainsGamCols`?

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: It is a bug.

wendycwong: Basically, if the input data have already been gamified, you may not find the gam columns in the input data.  Hence, if that happens, there is no need to add gamified columns to the frame and everything we need to do scoring is already there.

wendycwong:
@pavel: Your fix will not fix the problem and it is because of the following issue:
input dataset to mojo has the following column names
[‘response’, ‘C1’, ‘C2’, ‘C3’, ‘C4’]
However, the model._output._names contains the names of dinfo._adaptedFrame.
model._output._names:
0 = “C1”
1 = “C2"
2 = “C3”
3 = “C1_0_center__0"
4 = “C1_0_center__1”
5 = “C1_0_center__2"
6 = “C1_0_center__3”
7 = “C1_0_center__4"
8 = “C1_0_center__5”
9 = “C1_0_center__6"
10 = “C1_0_center__7”
11 = “C1_0_center__8"
12 = “C2_0_center__0”
13 = “C2_0_center__1"
14 = “C2_0_center__2”
15 = “C2_0_center__3"
16 = “C2_0_center__4”
17 = “C2_0_center__5"
18 = “C2_0_center__6”
19 = “C2_0_center__7"
20 = “C2_0_center__8”
21 = “C3_0_center__0"
22 = “C3_0_center__1”
23 = “C3_0_center__2"
24 = “C3_0_center__3”
25 = “C3_0_center__4"
26 = “C3_0_center__5”
27 = “C3_0_center__6"
28 = “C3_0_center__7”
29 = “C3_0_center__8"
30 = “response”
I actually need the input dataset column names in order to make the right decision. If you can find a good solution to what I already done, that will be great.
Wendy

wendycwong:
@pavel:  My latest push will resolve this issue.  However, it is not an elegant solution but get the job done.  If you can think of something better, please feel free to do so.  Thanks, Wendy

Pscheidl:
As far as I understand the situation, the user can theoretically provide their own GAM coefficients. However, the question is, why would anyone do that ?

wendycwong:
@pavel:  I don't think it is reasonable for users to provide gamified column names because if the user set k = 100, then, she has to type in 100 column names, that is too much to ask for.  Also, since the gamification process is deterministic once the user specified the gam_columns, types, and num_knots.

wendycwong:
@honzasterba The error in this code has nothing to do with fixing unseen categorical levels.  The problem is the code need to know what the input dataset column names are.  That is why I added the method checkNamesContainsGamCols to compare the gam_columns and the dataset column names.

wendycwong:
okay @honzasterba , I come up with a way to fix the problem of not knowing input dataset columns without using the extra methods.  Basically, if a dataset is given to us without gamification, the gamified columns will all contain NaN.  If they don't all contains NaN, that means gamification has been added and we don't need to do it again.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: I removed the function.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -132,29 +132,38 @@ int readCatVal(double data, int dataIndex) {
...
+        return false;
+    return true;
+  }
+  
   // this method will add to each data row the expanded gam columns
-  public double[] addExpandGamCols(double[] rawData, final RowData rowData) { // add all expanded gam columns here
+  double[] addExpandGamCols(double[] rawData, final RowData rowData) { // add all expanded gam columns here
+    int dataIndEnd = _totFeatureSize - _numExpandedGamCols; // starting index to fill out the rawData
+    if (!gamificationNeeded(rawData, dataIndEnd)) {
+       return rawData;     // already contain gamified columns.  Nothing needs to be done.
```
honzasterba, line 148: This is definitelly not the fix.
We should throw an exception here because gamified colums should never be provided from the outside, user will never to that. This is just a work-around for testJavaScoring method which based on gam cols being in mojo.names generates extra columns.

wendycwong: @honza:
I did not use _names in this fix.  I am just checking if the gamified columns are there.

wendycwong:
@honza, @pavel, here is some of my thoughts:

Here are the assumptions to Gam Mojo:
1. Users can provide us with raw dataset or with predictiors+gamififed columns.
2. Gam will add gamification if given raw dataset.
3. The crust of the matter is to determine if the user has provided us with a raw dataset or a predictor+gamified columns.
4. This can be easily done by looking at the rawData that is obtained from rawData = super.convert(data, rawData).  Check the gamified column values, if they are all NaN, that means they are not there.  Gamification is needed in this case.  Otherwise, gamification is not needed.  To speed up this process, I added a variable to keep track of the check and will 

The fix to do gamification regardless will run into the following problem:
If gam_columns are not included as predictors, then, your data will contain predictors+gamified columns.  In the addExpandGamCols(rawData, data) code, you will not the gam_column to gamified and instead fill in with NaN instead which is incorrect.

The fix to add feature names to Gam Mojo will not fix anything either because it contains the needed predictors and gam_columns but that does not tell us anything about either the input dataset is gamified or not since we do not have access to the input column names.  The RowData will only contain column names of non NaN values.  If you are unlucky and have lots of NaN, you may get the wrong input data column names.

pavel:
@wendycwong Thanks for notifying me, but it does not seem that I'm involved in this project in any way.

https://blog.github.com/2011-03-23-mention-somebody-they-re-notified/

PR #4812, opened 2020-07-27
GAM predictions fixes

Pscheidl:
A spin-off from https://github.com/h2oai/h2o-3/pull/4809

**Not done yet.**

The NaN problem was truly caused by iterating over non-existing columns (thus the array has been filled with NaNs - the NaN index was hit randomly and depended on the generated value). That is now fixed. **No column names** are relied upon. The GAM coefficients are ALWAYS added, as in reality there is no way the test dataset could contain them (why and how ?). Only slightly bent the code written by @wendycwong .

However, the predictions are slightly off - I wonder what the root cause for that is. @wendycwong can you please help me compare `GamModel` predictions with `GAMMojo` predictions and spot the difference ? I have a suspicion it's somewhere in the ETA function - do we use the correct indices all the time ?

### DIFF
--- a/h2o-algos/src/main/java/hex/gam/GAMModel.java
+++ b/h2o-algos/src/main/java/hex/gam/GAMModel.java
@@ -19,6 +19,8 @@ import water.util.*;
 
 import java.io.Serializable;
 import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 
 import static hex.gam.MatrixFrameUtils.GamUtils.equalColNames;
 import static hex.gam.MatrixFrameUtils.GamUtils.sortCoeffMags;
@@ -380,13 +382,9 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
     public String[] _responseDomains;
     public String _gam_transformed_center_key;
     final Family _family;
+    String[] _featureNames;
     
     public double dispersion(){ return _dispersion;}
-
-    @Override
-    public int nfeatures() {
-      return _names.length;
-    }
 
     @Override
     public int nclasses() {
@@ -412,6 +410,28 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
     public GAMModelOutput(GAM b, Frame adaptr, DataInfo dinfo) {
       super(b, adaptr);
       _dinfo = dinfo;
+      final int numColumns = b._parms._gam_columns.length + (b._parms.train().numCols() - b._parms._ignored_columns.length - 1);
+
+      final Set<String> names = new HashSet(b._parms.train().names().length);
+      for (String name: b._parms.train().names()){
+        names.add(name);
+      }
+      for(String ignored : b._parms._ignored_columns){
+        names.remove(ignored);
+      }
+      names.remove(b._parms._response_column);
+
+      _featureNames = new String[numColumns];
+      
+      int i = 0;
+      for (String name : names){
+        _featureNames[i++] = name;
+      }
+      
+      for (int j = 0; j < b._parms._gam_columns.length; j++) {
+        _featureNames[i++] = b._parms._gam_columns[j];
+      }
+      
       _domains = dinfo._adaptedFrame.domains(); // get domain of dataset predictors
       _family = b._parms._family;
       if (_family.equals(Family.quasibinomial)) {
--- a/h2o-algos/src/main/java/hex/gam/GAMModel.java
+++ b/h2o-algos/src/main/java/hex/gam/GAMModel.java
@@ -384,6 +384,11 @@ public class GAMModel extends Model<GAMModel, GAMModel.GAMParameters, GAMModel.G
     public double dispersion(){ return _dispersion;}
 
     @Override
+    public int nfeatures() {
+      return _names.length;
+    }
+
+    @Override
     public int nclasses() {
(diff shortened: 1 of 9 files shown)


h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -282,6 +282,7 @@ def mojo_predict(model, tmpdir, mojoname, glrmReconstruct=False, get_leaf_node_a
 
     p = subprocess.Popen(java_cmd, stdout=PIPE, stderr=STDOUT)
     o, e = p.communicate()
+    print(o)
```
Pscheidl, line 285: If the subprocess fail, the reason is hidden in `o` - thus printing it.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoReader.java
```diff
@@ -22,7 +22,6 @@ protected void readModelData() throws IOException {
     _model._family = DistributionFamily.valueOf((String)readkv("family"));
     _model._cats = readkv("cats", -1);
     _model._nums = readkv("num");
-    _model._numsCenter = readkv("numsCenter");
```
honzasterba, line 25: should we also remove from writer since no woriies abou bacwards compatibility (this is the first release of gam mojo)

Pscheidl: Oh sure !

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -60,8 +60,7 @@
 
   @Override
   public double[] score0(double[] row, double[] preds) {
-    if (_meanImputation)
-      imputeMissingWithMeans(row);  // perform imputation for each row
+    imputeMissingWithMeans(row);  // perform imputation for each row
```
honzasterba, line 63: should also remove _meanImputation since it is unused now

Pscheidl: Sure.

h2o-py/tests/testdir_javapredict/pyunit_PUBDEV_7185_GAM_mojo_gaussian.py
```diff
@@ -27,7 +27,9 @@ def gam_gaussian_mojo():
     
     train = df[NTESTROWS:, :]
     test = df[:NTESTROWS, :]
-    x = list(set(df.names) - {"response"})
+    exclude = {"response"}
+    exclude.add(params["gam_columns"][0])
+    x = list(set(df.names) - exclude)
 
     TMPDIR = tempfile.mkdtemp()
     gamGaussianModel = pyunit_utils.build_save_model_generic(params, x, train, "response", "gam", TMPDIR) # build and save mojo model
```
wendycwong, line 35: @pavel:  Need to change test[x] to test in the save in.csv part of the code.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -134,29 +133,26 @@ int readCatVal(double data, int dataIndex) {
...
-        double gam_col_data = Double.NaN;
-        if (dataObject == null) {  // gamified columns are already included
-          return rawData;
-        } else
-          gam_col_data = (dataObject instanceof String)?Double.parseDouble((String)dataObject):(double) dataObject;
-        GamUtilsCubicRegression.expandOneGamCol(gam_col_data, _binvD[cind], _basisVals[cind], _hj[cind], _knots[cind]);
-      } else {
-        throw new IllegalArgumentException("spline type not implemented!");
+      // add expanded gam columns to rowData
+      int dataIndStart = _totFeatureSize - _numExpandedGamCols; // starting index to fill out the rawData
```
wendycwong, line 137: @pavel:  You will run into problem here with testJavaScoring.  However, this should work when you are doing normal scoring for python unit.  I am puzzled by why it is failing at both places.

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -631,6 +636,12 @@ else if (r.weight == 0)
     }
 
     public double[] scoreRow(DataInfo.Row r, double offset, double[] preds) {
+      for (int i = 0; i < r.numVals.length; i++) {
+        if(Double.isNaN(r.numVals[i])){
+          throw new IllegalArgumentException("NaN found.");
+        }
+      }
+
```
wendycwong, line 644: If meanImputation is enabled, there will be no NaN.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: I was able to run the python test to completion successfully.  However, as I mentioned before, it is not going to work with testJavaScore because there will be a gam_column missing and in this case, it is just going to leave NaN in the gamified columns.  Later during scoring, it will then replace the NaN with column  means.  That is why the model prediction and mojo predictions are different.

PR #4813, opened 2020-07-27
Fix pyunit_pubdev_4702_algo_max_runtime_secs_large.py

wendycwong:
pyunit_pubdev_4702_algo_max_runtime_secs_large.py failed with the following error message:

UnboundLocalError: local variable 'system_overloaded' referenced before assignment
Closing connection _sid_b0e4 at exit

system_overloaded is actually a global variable.  It needs to be declared before it is used.


wendycwong:
already in

PR #4816, opened 2020-07-28
PUBDEV-7702: Target encoding refactoring + cleanup.

sebhrusen:
See description in https://0xdata.atlassian.net/browse/PUBDEV-7702

No new feature here outside the fixes listed below + the support for regression as the change of numerator/denominator types were enough to get `regression` working after having fixed some bugs in the logic.
Goal is to get a "definitive" maintainable implementation with a standard quality level API.

- Ensured that I could change the client APIs, **deleted the deprecated ones** (based on Rapids expression), and do some backend renaming/simplifications and keep the tests passing.
- Redesigned and fixed the `transform` API: it was just plain buggy.
- `TargetEncoder` class is now the builder, this way implementation is much simpler and follow the usual model structure.
- Rewrote the Java tests to be more understandable...
- Fixed some issues like keys and exception handling.
- Added proper standard validation in the builder, this should also benefit Flow users.
- Moved part of data leakage handling to the training phase for optimization: this avoids repeating the same manipulation on the encodings frame (the frame containing the information required to encode a test frame) each time we apply TE with KFold strategy.
- Rename some utility methods + improved Javadoc each time I found it could help future maintainer.

### API Changes:
on all clients (Py, R, Rest, Flow):
- `k` -> `inflection_point`
- `f` -> `smoothing`
- `noise_level` -> `noise`
Those renamings are backwards compatible on `Py` and `R`: will work to get them on SW as well.

on Py, R, Java:
- `transform` method now accepts only `noise`, and `blending` related params: user can't change the data leakage handling mode, or the seed defined during training, like for any other model.
- by default, `transform` is expected to behave like `score`: the latter only always use the default noise and blending.
- to transform training data, `transform` accepts a new parameter `as_training=True` : this informs user and ensures that data leakage handling is applied to training data. For backwards compatibility, user that used to call `transform` with (deprecated) data_leakage_handling != none, will get this flag automatically set to True.
- in Java, there are 2 methods for clarity as it will be used by AutoML: `transform` to encode new data, `transformTraining` to encode training data.

To sum up:
old way:
```python
noise = 0.2
data_leakage_handling = "kfold"
seed = 42
use_blending = True
inflection_point = 5
smoothing = 10
te = H2OTargetEncoderEstimator(noise_level=noise, 
                          data_leakage_handling=data_leakage_handling,
                          blending=use_blending,
                          k=inflection_point,
                          f=smoothing,
                          seed=seed)
te.train(y="target", training_frame=train, fold_column='foldc')

# couldn't set blending on transform, so it was always set to false, ignoring the fact that blending was set on model...
# also used `noise` parameter here, but `noise_level` in constructor and noise+seed had to be repeated to be applied.
train_enc = te.transform(train, 
                         data_leakage_handling=data_leakage_handling, 
                         noise=noise, 
                         seed=seed)
test_enc = te.transform(test, data_leakage_handling="none", noise=0) 
```

new way:
```python
te = H2OTargetEncoderEstimator(noise=0.2, 
                        data_leakage_handling="kfold",
                        blending=True,
                        inflection_point=5,
                        smoothing=10,
                        seed=42)
te.train(y="target", training_frame=train, fold_column='foldc')

# transform now automatically inherits from properties set on constructor.
# params like `noise`, `blending`, `inflection_point`, `smoothing` can though be overridden if necessary.
train_enc = te.transform(train, as_training=True)
test_enc = te.transform(test, noise=0) 
```

### Major Java changes (logic)
- `TargetEncoder` is now the main TE algo class, in charge of the model training, `TargetEncoderModel` handles mainly the data transformation logic, and useful MRTasks are mainly grouped in the `TargetEncoderHelper` utility class.
- `TargetEncoder` now validates parameters in the `init` method, using the model validation API for consistent error reporting on Flow especially.
- `TargetEncoder#applyLeakageStrategyToEncodings` optimizes the application of the `data_leakage_strategy` by applying some logic during the training phase instead of having to do it each time user needs to encode data.
- `TargetEncoderHandler`: full rewrite of the handling of the `transform` method so that any non-provided param systematically defaults to its value on the model instance.
- improved encapsulation in some methods: for example, `TargetEncoderBroadcastJoin` is doesn't require to have the joined columns already created on the left frame.
- ensure that the encodings `numerator` is ALWAYS represented as a double to allow support for regression.
- lowered visibility for most methods (almost everything used to be `public`).
- minor changes, mainly on the utiliy functions, aimed at improving reasoning on the code: for example, methods modiying input return void, whereas methods returning new frame return Frame... and other similar changes.

### TODO
- there are still test placeholder files for R clients: this means more tests to come.
- review/update documentation.
- review/update tutorials.


Pscheidl:
The API changes look promising - a step needed. Fully agree.

h2o-bindings/bin/custom/R/gen_targetencoder.py
```diff
@@ -5,7 +5,24 @@
 parms$ignored_columns <- args$x_ignore
```
sebhrusen, line 5: adding backwards compatibility support for renamed params `k`, `f`, `noise_level`.

jakubhava:
@sebhrusen > Those renamings are backwards compatible on Py and R: will work to get them on SW as well. 
We use rest api for target encoder as well, so should be fine

Pscheidl:
Just started to review, noticed related test failures @sebhrusen .

sebhrusen:
@Pscheidl thanks for the pointer, R failures are expected, I didn't fix those...
Will fix the java ones asap: didn't realize there ware TE tests outside the TE module...

@jakubhava @mn-mikke 
> We use rest api for target encoder as well, so should be fine
No, you won't be fine :)
The REST API is not backwards compatible, the backwards-compatibility has been enforced on client-side.
Let's have a sync with Marek so that I can explain the changes in detail and show how to make it BC in your side.

sebhrusen:
@michalkurka @Pscheidl 
see https://github.com/h2oai/h2o-3/pull/4816/commits/c0aa606f30bd6b278e94d806e33f03eff844582e fixing the previously failing `GBMEncoding` and `StackedEnsembleEncoding` after my changes in default domain ordering in `TesrFrameBuilder`.
It's interesting that some encoders don't mind about the domain order when others do...

mn-mikke:
Really like that `k`, `f` parameter was replaced with something more descriptive!

h2o-core/src/main/java/water/Scope.java
```diff
@@ -99,20 +104,31 @@ static private void track_impl(Scope scope, Key key) {
     if (scope._keys.size() > 0 && !scope._keys.peek().contains(key))
       scope._keys.peek().add(key);            // Track key
   }
-
-  static public void untrack(Key<Vec>... keys) {
+
```
Pscheidl, line 107:
```suggestion

```

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TEBroadcastJoin.java
```diff
@@ -0,0 +1,218 @@
+package ai.h2o.targetencoding;
+
+import water.MRTask;
+import water.MemoryManager;
+import water.fvec.CategoricalWrappedVec;
+import water.fvec.Chunk;
+import water.fvec.Frame;
+
+class TEBroadcastJoin {
```
Pscheidl, line 9: Knit: Let's prefix everything related to target encoder with the same prefix, e.g. `TargetEncoder` - not really that important though :wink:  Just an idea.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TEBroadcastJoin.java
```diff
@@ -0,0 +1,218 @@
...
+   * @param maxFoldValue the highest fold value (e.g. 4 if 5 folds).
+   * @return the `leftFrame` with joined numerators and denominators.
+   */
+  static Frame join(Frame leftFrame, int[] leftCatColumnsIdxs, int leftFoldColumnIdx,
+                    Frame rightFrame, int[] rightCatColumnsIdxs, int rightFoldColumnIdx,
+                    int maxFoldValue) {
+    int rightNumeratorIdx = rightFrame.find(TargetEncoderHelper.NUMERATOR_COL);
+    int rightDenominatorIdx = rightNumeratorIdx + 1; // enforced by 
+    
+    // currently supporting only one categorical column
```
Pscheidl, line 29: Why do we have this limitation ? Can we get rid of it ? Not necessary in this PR - just a question.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TEBroadcastJoin.java
```diff
@@ -0,0 +1,218 @@
...
+    // BroadcastJoiner is currently modifying the frame in-place, so we need to provide the numerator and denominator columns.
+    Frame resultFrame = new Frame(leftFrame);
+    resultFrame.add(TargetEncoderHelper.NUMERATOR_COL, resultFrame.anyVec().makeCon(0));
+    resultFrame.add(TargetEncoderHelper.DENOMINATOR_COL, resultFrame.anyVec().makeCon(0));
+    new BroadcastJoiner(leftCatColumnsIdxs, leftFoldColumnIdx, encodingData, levelMappings, rightCatCardinality-1)
+            .doAll(resultFrame);
+    return resultFrame;
+  }
+  
+  static double[][] encodingsToArray(Frame encodingsFrame, int categoricalColIdx, int foldColIdx,
```
Pscheidl, line 61: TBH I don't understand the point of the original design (I fully realize this PR can't resolve evertything) moving everything back and forth between frame representation and arrays. I'd personally try to avoid these steps. Just a personal note.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TEBroadcastJoin.java
sebhrusen: I guess it's just because TE doesn't need to join on more than one column for now.
It would make sense to join on multiple columns only if the encodings frame was created for 2 categorical columns at once, and I don't think anyone does this.
Another possibility is that the original author had something in mind for multiclass maybe, but I can't find any good alternative to the paper's proposal for such problems, so this is still a mystery...

h2o-extensions/target-encoder/src/main/java/hex/api/targetencoding/TargetEncoderHandler.java
```diff
@@ -0,0 +1,53 @@
...
+import water.fvec.Frame;
+
+public class TargetEncoderHandler extends Handler {
+  @SuppressWarnings("unused")
+  public KeyV3.FrameKeyV3 transform(final int version, final TargetEncoderTransformParametersV3 parametersV3) {
+    final TargetEncoderTransformParameters parameters = new TargetEncoderTransformParameters();
+    parametersV3.fillImpl(parameters);
+    final TargetEncoderModel model = parameters._model.get();
+
+//    final boolean useBlending = parameters._blending == null ? model._parms._blending : parameters._blending;
```
Pscheidl, line 19: Those are better to be deleted ?

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TEBroadcastJoin.java
sebhrusen: Will see in the future if I should just change the signature (avoid arrays if they're not used) or add support for join on multiple columns

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TEBroadcastJoin.java
sebhrusen: It confused me as well, don't want to change this here though: it works. I resisted in not changing a few things to avoid getting sunk into this :)

h2o-extensions/target-encoder/src/main/java/hex/api/targetencoding/TargetEncoderHandler.java
sebhrusen: True, I can remove them now, and I will

sebhrusen:
> I have only one big question/issue with the parameter deprecation. The old k and f arguments were bad design, yet I believe we may keep them and explicitly state in our documentation those are the same.
Is there something preventing us from mapping the old parameters to the new ones internally and keep them, while marking them as deprecated @sebhrusen @michalkurka ?

Do you mean on Java side? I tried to keep `k`/`f` backwards compatible on client side, but on Java, it adds complexity that I found unnecessary and will add confusion on the Flow UI where both `k` and `inflection_point` will appear.
I'm open to any suggestion though.

Pscheidl:
> > I have only one big question/issue with the parameter deprecation. The old k and f arguments were bad design, yet I believe we may keep them and explicitly state in our documentation those are the same.
> > Is there something preventing us from mapping the old parameters to the new ones internally and keep them, while marking them as deprecated @sebhrusen @michalkurka ?
> 
> @Pscheidl Do you mean on Java side? I tried to keep `k`/`f` backwards compatible on client side, but on Java, it adds complexity that I found unnecessary and will add confusion on the Flow UI where both `k` and `inflection_point` will appear.
> I'm open to any suggestion though.

Could we simply make Flow ignore some parameters marked to ignore ?

Pscheidl:
I'd personally make Java compatible as well, but we can omit that. But the clients should remain compatible from my POV. That's the only point. Is this doable @sebhrusen ?

h2o-docs/src/product/api-changes.rst
```diff
@@ -6,7 +6,38 @@ H2O-3 does its best to keep backwards compatibility between major versions, but
...
+- ``k`` -> ``inflection_point``
+- ``f`` -> ``smoothing``
+- ``noise_level`` -> ``noise``
+- ``use_blending`` (R only) -> ``blending``
+
+Legacy client code using the deprecated parameters should expect a deprecation warning when using them, they are strongly encouraged to update their code to use the new naming.
+
+In an objective of performance optimization on the backend, and of simplification of the API, the ``transform`` method used to apply target encoding was modified as follow.
+- The R ``h2o.transform`` function (accepting a target encoder model as the first argument) and the Python ``H2OTargetEncoderEstimator.transform`` methods are now fully compatible: they accept the same parameters and work consistently.
+- The parameters ``data_leakage_handling``, ``seed`` are now ignored on those methods: ``transform`` will use by default the corresponding values defined when building the TargetEncoder model.
```
michalkurka, line 29: @sebhrusen what is the reasoning behind this, this one of Andrey's design goals to be able to change these parameters without re-training. I think dropping seed shouldn't hurt but what about the leakage handling?

Does this also affect SW API? If so how? cc: @mn-mikke (adding as a reviewer)

h2o-core/src/main/java/water/api/SchemaMetadata.java
```diff
@@ -222,9 +222,27 @@ public static String consType(Schema schema, Class clz, String field_name, API a
...
+      if (clz.equals(Boolean.TYPE) 
+              || clz.equals(Byte.TYPE) 
+              || clz.equals(Short.TYPE) 
+              || clz.equals(Integer.TYPE) 
+              || clz.equals(Long.TYPE) 
+              || clz.equals(Float.TYPE)
+              || clz.equals(Double.TYPE))
         return clz.toString();
 
+      // use the same representation for wrapper classes as for their primitive.
```
michalkurka, line 234: Where are we starting to use objects instead of primitives in the schemas?

h2o-extensions/target-encoder/src/main/java/hex/api/targetencoding/TargetEncoderHandler.java
```diff
@@ -0,0 +1,49 @@
...
+public class TargetEncoderHandler extends Handler {
+  @SuppressWarnings("unused")
+  public KeyV3.FrameKeyV3 transform(final int version, final TargetEncoderTransformParametersV3 parametersV3) {
+    final TargetEncoderTransformParameters parameters = new TargetEncoderTransformParameters();
+    parametersV3.fillImpl(parameters);
+    final TargetEncoderModel model = parameters._model.get();
+
+    final boolean asTraining = parameters._as_training;
+    final boolean useBlending = parameters._blending;
+    final double noise = parameters._noise < -1  ? model._parms._noise : parameters._noise;
```
michalkurka, line 21: Do you know why `< -1` and not `< 0`?

h2o-extensions/target-encoder/src/main/java/hex/api/targetencoding/TargetEncoderHandler.java
```diff
@@ -0,0 +1,49 @@
...
+    final double inflectionPoint = parameters._inflection_point < 0 ? model._parms._inflection_point : parameters._inflection_point;
```
michalkurka, line 22: I wouldn't parse parameters that are not needed. I would hide this in `if useBlending` - makes the flow clearer.

michalkurka:
Good stuff, thanks for this effort. Can you please summarize what changes in the actual implementation were made in the Java code? It is hard to review this line-by-line. If some stuff is mostly untouched, we could skip this.

h2o-core/src/main/java/water/api/SchemaMetadata.java
sebhrusen: I can revert this now, it was a former attempt to use objects for `TargetEncoderHandler`: the problem is that the legacy implementation was using the primitive `blending` there, and in Python client for example, there was no way to specify the blending value for the `transform` method, which means that `blending` was always `false`, even if the model was trained with `blending=true`...
There were several inconsistencies like that that were making the `transform` API especially very hard to understand.

I solved this default boolean value issue (how to identify if a boolean parameter was provided by user or not?), by setting the default value on client side, except for Flow, that doesn't allow to use the `transform` method: actually I'm even wondering if TE is fully usable on Flow, given that an encoder is supposed be applied differently on a train frame and a test or validation frame...

h2o-extensions/target-encoder/src/main/java/hex/api/targetencoding/TargetEncoderHandler.java
sebhrusen: I see, you mean having a big `if (parameters._blending)` block instead? Sure, can do that.

h2o-docs/src/product/api-changes.rst
sebhrusen: From my understanding, the idea of providing `data_leakage_handling` in `transform` method was not to avoid re-training (we're still using the same encodings), but to allow a different encoding of training data vs. validation/test data.
Typically, when user wants to use a data leakage strategy, this is the for the training data:
- kfold requires a fold column, that you're unlikely to find in validation/test data, and that you surely don't have in production data.
- leave-one-out requires the target, that you may not have in test data, and that you surely don't have in production data.
So, I think the original approach was to expose those low-level parameters and let the user decide when to apply the `data_leakage_handling`, meaning that the user will have to deeply understand how it works to avoid mistakes.
For example if you trained your TE model with `kfold` strategy, then it doesn't really make sense to encode your training data with `leave-one-out` strategy, but it was still possible.
One the other way, if you trained your TE model with `leave-one-out` strategy, then you wouldn't be able to encode your data with `kfold`... same if you trained it with `none` strategy.

Due to these inconsistencies, I prefer to train the TE with one strategy, and then use the new param `as_training` in the `transform` methods, to use the strategy for training data, and disable it (as_training=false by default) for other data, focusing on user intent instead of technical parameters that require user to understand in detail what's happening.

Also note that by preventing use of `data_leakage_handling` in `transform`, it allows me to do some optimizations in the Java implementation: some groupings can be done during training instead of doing them each time we call `transform`...

sebhrusen:
@michalkurka I agree that it's hard to review because I renamed classes to align with other models, and I also renamed many methods to help me reasoning about this implementation (for example some methods could be called only if the consumer had prepared one param in a specific state..., making things really hard to reason about).
I added a section in the PR description with changes in the Java logic, hope this helps.

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -599,18 +599,26 @@ public BinomialModelPrediction predictBinomial(RowData data, double offset) thro
...
+  }
+
   /**
    * Perform target encoding based on TargetEncoderMojoModel
    * @param data RowData structure with data for which we want to produce transformations
    * @return TargetEncoderPrediction with transformations ordered in accordance with corresponding categorical columns' indices in training data
    * @throws PredictException
    */
-  public TargetEncoderPrediction transformWithTargetEncoding(RowData data) throws PredictException{
+  public TargetEncoderPrediction predictTargetEncoding(RowData data) throws PredictException{
```
michalkurka, line 616: well... this is what I suggested originally but it had to be transform back then...

sebhrusen: Yeah, I think that at API level, consistency is more important than perfect naming... and word2vec has a similar semantic as TE but it is still exposed as `predictWord2Vec`, so I don't see any reason to keep (outside backwards compatibility) this `transform__`.

sebhrusen:
> I think that since we are now restarting the development with fresh minds and objectives, it would also be good to add a TE benchmark to the benchmark suite.

@michalkurka do you mean microbenchmarks? Is it ok if I just create a ticket for those now and write them only when I'm done with the integration?

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
ledell: Internal conventions aside, I do think that "transform" is the more sensible verb to describe what we're doing here (it's not really predicting anything in the ML sense of the word "predict"; it's just computing the values).  It sounds like the preference to use the "predict" verb comes with the desire to be treating TE as an "estimator", or is there something more than that?  Are there other methods like "predictSomething" in the Python API?

ledell:
Hi @sebhrusen Here's my first pass -- I will do a final review tomorrow to see if I missed anything and also look more closely at the tests.

It looks great!  Most of my comments are on existing conventions (not directly related to the code changes in this PR), but since we are doing some refactoring / API changing, it seems relevant to bring up now (vs later/never):

Is there a reason that we are using different names for the `data_leakage_strategy` parameters in R vs Python?  We typically use the same argument string values in R/Py in H2O and it makes it a lot easier to have consistency between the two (and there's really no reason not to).  Is there a way that we can get these in sync right now?  Or do you see issues with doing that due to backwards compatibility, etc.  To me, it would make more sense to use the Python convention (since Python has stronger conventions around argument value names) -- I assume we can always keep supporting the "old" R names through some simple gsub code on the client side if need be?

- `none` (Python)/`None` (R): Do not holdout anything. Using whole frame for training
- `k_fold` (Python)/`KFold` (R): Encodings for a fold are generated based on out-of-fold data.
- `leave_one_out` (Python)/`LeaveOneOut` (R): The current row’s response value is subtracted from the pre-calculated per-level frequencies.

I'm a bit confused by the use of the term `response_column` in Python instead of the usual `y` to define this column?  All of our demos/code/etc use `y` at the user-level -- the term itself is never use in the R API (anywhere) and `response_column()` in Python is a method to grab the name of the column from the model object, it's not used to *define* it (`y` is always used instead).   Another issue is that the current TE [docs](http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-munging/target-encoding.html?highlight=response_column) only talk about `response_column` don't use the term `y` at all (which is the only name in R).  I see `y` used in Python (and R) tests, so that's good.

Another API question that's not related to current PR -- At some point long ago, we had talked about using a name other than `x` to represent the columns to be included (since `x` represents "all" predictor columns, and here we are only defining a subset) -- I assume using `x` here is required because of the fact that TE is an estimator, and thus requires it?  I'm not suggesting we change it, I am just curious if the name `x` is required because it's an estimator.

Some of the argument definitions in the user guide are missing default values (e.g. `blending`).  We are pushing a new trend of [always including defaults](https://0xdata.atlassian.net/browse/PUBDEV-7745) in the user guide parameter definitions now.

Are all the R tests complete?  It still says "there are still test placeholder files for R clients: this means more tests to come."

sebhrusen:
@erin, thanks for your feedback.

> I'm a bit confused by the use of the term `response_column`

Yeah, this is one oddity with the Python API, nothing specific to TE here, all algos expose those `response_column` and `ignored_columns` which imo should be kept (hidden) only for internal logic.
They should definitely be removed from the doc though, it can only confuse the user.

> we had talked about using a name other than x to represent the columns to be included.

I see your point, the main benefit would be for `sklearn` pipelines where the distinction between `X` and `columns_to_encode` is clear. Let me create a ticket for this.
Although for now, outside the `sklearn`/pipeline scenario, the current use of `x` doesn't limit what you can do with TE, it's just that by default, if `x` is not specified for specified or if includes all predictors, TE will encode all categorical columns included in `x`.

> Is there a reason that we are using different names for the data_leakage_strategy parameters in R vs Python?

Outside the fact that it's legacy and changing those values in R is annoying, no, there's no good reason for having different names.
Backend understands both `k_fold` and `KFold` for example, this is part of the conversion logic from the client parameter as a string to the internal representation as an enum.
So I guess that the easiest way to make them compatible is to support both cases on both clients... it should be trivial.

> Some of the argument definitions in the user guide are missing default values.

Ok, need to go through the docs again.

> Are all the R tests complete? 

correct, my bad! I wanted to add tests for the backward compatibility logic after the renamings. Will do....

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
sebhrusen: I think `predict` is currently better simply by convention: I agree that semantically it's not the perfect choice for TargetEncoder, Word2Vec, AutoEncoder, DimReduction...

For the future, maybe we could keep this `predict` prefix everywhere for consistency, and for clarity, add a `transform` or `apply` alias for those who actually don't produce predictions.

ledell (approved):
Looks good!  Since this needs to be merged (as to not block/slow down other TE updates), we can address the smaller changes in a follow-up PR.

sebhrusen:
synced with @mn-mikke and agreed to merge this now even if it breaks SW dev build, given the small amount of changes needed on his side: cf. `targetencoder.py` for an example of required changes to adapt to the new  REST API.

PR #4819, opened 2020-07-30
PUBDEV-7119 - CSVStream does not escape quotes.

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7119

The parser works fine. Problem is our CSVStream's parameters, where for some reason the escaping parameter is not set. For the parser inside Python, quotes should always be escaped (by double quotes).

R has this scenario covered already: 

```R
urlSuffix <- paste0('DownloadDataset',
                      '?frame_id=', URLencode(h2o.getId(x)),
                      '&hex_string=', ifelse(useHexString, "true", "false"),
                      '&escape_quotes=', ifelse(useDataTable, "false", "true"))
```

The data only appear as broken (in Python) when importing back into Pandas. In H2O itself, the format is correct.

### DIFF
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -2183,7 +2183,7 @@ class H2OFrame(Keyed):
         >>> iris = h2o.import_file("http://h2o-public-test-data.s3.amazonaws.com/smalldata/iris/iris_wheader.csv")
         >>> iris.get_frame_data()
         """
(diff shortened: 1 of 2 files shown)


Pscheidl:
Everything is green, except for `Py3.7 Small / pyunit_PUBDEV_7185_GAM_mojo_ordinal.py – r_suite` - unrelated failure. merging.

PR #4820, opened 2020-07-30
PUBDEV-7708 Add GAM parameters in GAM.V3 to GAMParameters

wendycwong:
This PR completed the job in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7708

Simple PR to add back parameters defined in GAM.V3 back to GAMParameters.


mn-mikke:
Thanks a lot @wendycwong!

mn-mikke:
There is one more item called `lambda_min_ratio` which is in the V3 class and missing in `GAMParameters`. cc @wendycwong @Pscheidl

PR #4821, opened 2020-07-31
[GAM] gam_columns should not be ignored

honzasterba:


h2o-algos/src/main/java/hex/schemas/GAMV3.java
```diff
@@ -169,11 +168,6 @@
     @API(help = "Beta constraints", direction = API.Direction.INPUT /* Not required, to allow initial params validation: , required=true */)
     public FrameKeyV3 beta_constraints;
 
-    @API(help="Maximum number of active predictors during computation. Use as a stopping criterion" +
-            " to prevent expensive model building with many predictors." + " Default indicates: If the IRLSM solver is used," +
-            " the value of max_active_predictors is set to 5000 otherwise it is set to 100000000.", direction = Direction.INPUT, level = Level.expert)
-    public int max_active_predictors = -1;
-
     @API(help="A list of predictor column indices to interact. All pairwise combinations will be computed for the list.", direction=Direction.INPUT, level=Level.expert)
```
wendycwong, line 171: Need to be kept.

h2o-py/h2o/estimators/gam.py
```diff
@@ -677,23 +677,6 @@ def beta_constraints(self, beta_constraints):
...
-
-        Type: ``int``  (default: ``-1``).
-        """
-        return self._parms.get("max_active_predictors")
-
-    @max_active_predictors.setter
-    def max_active_predictors(self, max_active_predictors):
-        assert_is_type(max_active_predictors, None, int)
-        self._parms["max_active_predictors"] = max_active_predictors
-
```
wendycwong, line 695: Please keep these.

h2o-py/tests/testdir_javapredict/pyunit_PUBDEV_7185_GAM_mojo_gaussian.py
```diff
@@ -9,8 +9,8 @@ def gam_gaussian_mojo():
     h2o.remove_all()
     NTESTROWS = 200    # number of test dataset rows
     PROBLEM="gaussian"
-    params = set_params()   # set deeplearning model parameters
-    df = pyunit_utils.random_dataset(PROBLEM, missing_fraction=0.001)   # generate random dataset
+    params = set_params() 
+    df = pyunit_utils.random_dataset(PROBLEM, seed=1, missing_fraction=0.5)
```
wendycwong, line 13: Remove the seed.  I found that you can set the seed and it will make the test pass without fixing anything.
Add in the following change:

    exclude_list = {"response"}
    exclude_list.add(params["gam_columns"][0])
    x = list(set(df.names) - exclude_list)

replace the save in.csv with this line:
    h2o.download_csv(test, os.path.join(TMPDIR, 'in.csv'))  # save test file, h2o predict/mojo use same file

h2o-bindings/bin/custom/python/gen_gam.py
```diff
@@ -15,6 +15,12 @@ def Lambda(self):
     def Lambda(self, value):
         self._parms["lambda"] = value
 
+    def _additional_used_columns(self, parms):
+        """
+        :return: Gam columns if specified.
+        """
+        return parms["gam_columns"]
+
 extensions = dict(
```
wendycwong, line 24: I don't understand the use of addtional_used_columns.  What is the term used for?

honzasterba: if not used the client (python) will automatically add any cols not in X,Y,weight etc. to ignored_columns

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -241,9 +241,6 @@ private void fillUpCoeffs(double[] coeffValues, double[] coeffValuesStand, TwoDi
     public String[] _interactions=null;
     public StringPair[] _interaction_pairs=null;
     public Key<Frame> _plug_values = null;
-    // internal parameter, handle with care. GLM will stop when there is more than this number of active predictors (after strong rule screening)
-    public int _max_active_predictors = -1;
-    public boolean _stdOverride; // standardization override by beta constraints
```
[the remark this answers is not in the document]
honzasterba: ah, ok, good catch, but this is true for _max_active_predictors
_stdOverride is not exposed or used, why does it need to be here?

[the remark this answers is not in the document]
wendycwong: Actually, the user can set it if they want to.  It is for users to decide.

h2o-bindings/bin/custom/python/gen_gam.py
wendycwong: It is fine for the gam_columns to be included in the ignored_columns if they are not specified as a predictor.  The reason is because in this case, the gamified columns derived from the gam_column s are used as predictors.

h2o-py/tests/testdir_javapredict/pyunit_PUBDEV_7185_GAM_mojo_gaussian.py
honzasterba: done

h2o-algos/src/main/java/hex/gam/GAMModel.java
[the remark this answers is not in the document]
honzasterba: its not exposed via rest API so user cannot change it

h2o-algos/src/main/java/hex/schemas/GAMV3.java
```diff
@@ -55,7 +55,6 @@
             "prior",
             "lambda_min_ratio",
             "beta_constraints",
-            "max_active_predictors",
             "interactions",
```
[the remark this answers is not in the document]
honzasterba: reverted

h2o-algos/src/main/java/hex/schemas/GAMV3.java
honzasterba: reverted

h2o-py/h2o/estimators/gam.py
honzasterba: reverted

h2o-py/tests/testdir_javapredict/pyunit_PUBDEV_7185_GAM_mojo_gaussian.py
honzasterba: I kept the seed, random tests are not good, tests should be deterministic, even with the seed fails non-deterministically, need to investigate more

h2o-bindings/bin/custom/python/gen_gam.py
honzasterba: I added this more for consistency sake, the client should not mark gam_columns as ingored

h2o-algos/src/main/java/hex/gam/GAMModel.java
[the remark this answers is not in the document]
wendycwong: Actually, I was asked to change it recently.  Here is the PR: https://github.com/h2oai/h2o-3/pull/4820/files

h2o-bindings/bin/custom/python/gen_gam.py
wendycwong: I actually have to manually add gam_columns that are not used as predictors into ignored_columns in Java.  See testModelScoring in GamPipingTest.

h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -298,7 +298,19 @@ Frame adaptTrain() {
...
+    public Frame buildGamFrame() {
+      Vec responseVec = _train.remove(_parms._response_column);
+      for (int colIdx = 0; colIdx < _parms._gam_columns.length; colIdx++) {  // append the augmented columns to _train
+        Frame gamFrame = Scope.track(_gamFrameKeysCenter[colIdx].get());
+        _train.add(gamFrame.names(), gamFrame.removeAll());
+        _train.remove(_parms._gam_columns[colIdx]);
+      }
+      if (responseVec != null)
+        _train.add(_parms._response_column, responseVec);
+      return _train;
```
wendycwong, line 313: nice move.

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -285,7 +285,11 @@ def mojo_predict(model, tmpdir, mojoname, glrmReconstruct=False, get_leaf_node_a
     files = os.listdir(tmpdir)
     print("listing files {1} in directory {0}".format(tmpdir, files))
     outfile = os.path.join(tmpdir, 'out_mojo.csv')
-    print("***** importing file {0}".format(outfile))
+    if not os.path.exists(outfile) or os.stat(outfile).st_size == 0:
+        print("MOJO SCORING FAILED:")
+        print("--------------------")
+        print(o.decode("utf-8"))
+    print("***** importing file {0}".format(outfile))
```
wendycwong, line 292: Thanks, this is great.  Usually the test just failed with no warning.

h2o-py/tests/testdir_javapredict/pyunit_PUBDEV_7185_GAM_mojo_gaussian.py
```diff
@@ -5,13 +5,15 @@
...
 
+
 def gam_gaussian_mojo():
     h2o.remove_all()
     NTESTROWS = 200    # number of test dataset rows
     PROBLEM="gaussian"
-    params = set_params()   # set deeplearning model parameters
-    df = pyunit_utils.random_dataset(PROBLEM, missing_fraction=0.001)   # generate random dataset
+    params = set_params() 
+    df = pyunit_utils.random_dataset(PROBLEM, seed=2, missing_fraction=0.5)
```
wendycwong, line 14: Please remove seed to get a more robust test.

honzasterba: What do you mean by robust? I think in software robust means predictable, deterministic. This test is anything but when it fails only once in a while, setting the seed is my attempt at making it more robust, now the test still fails but fails more often, there is more work to be done in making this test more robust.

h2o-bindings/bin/custom/python/gen_gam.py
honzasterba: that is probably unnecessary and could be removed

h2o-algos/src/main/java/hex/gam/GAMModel.java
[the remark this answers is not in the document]
honzasterba: _stdOverride is not part of the mentioned PR, nor it is part of GLM params, nor is it used anywhere in the code, thats why I think its ok to remove it

PR #4823, opened 2020-07-31
PUBDEV-7681 get_hyperparams_dict should return correct params

koniecsveta:
https://0xdata.atlassian.net/browse/PUBDEV-7681
2 changes in get_hyperparams_dict():

1. fold_assignment for the fold models used during the cross validation is set to AUTO here: ModelBuilder.java L639.
I think this should not overwrite fold_assignment of overall model in get_hyperparams_dict(). Hence I am extracting only max_runtime_secs from the first fold model (what seems to be intention of underlying change in PUBDEV-1843, but it was done in such a way, that currently, all params of overall model are overwritten by params of the first fold model in get_hyperparams_dict()).

2. if some hyperparam value is of type list, only its first element is saved into resulting hyperparams dict, what is causing wrong content of sample_rate_per_class. So this was also fixed.

### DIFF
--- a/h2o-py/h2o/grid/grid_search.py
+++ b/h2o-py/h2o/grid/grid_search.py
@@ -1420,14 +1420,14 @@ class H2OGridSearch(h2o_meta(Keyed)):
 
         model_params = dict()
 
-        # if cross-validation is turned on, parameters in one of the fold model actual contains the max_runtime_secs
-        # parameter and not the main model that is returned.
-        if model._is_xvalidated:
-            model = h2o.get_model(model._xval_keys[0])
-
         for param_name in self.hyper_names:
-            model_params[param_name] = model.params[param_name]['actual'][0] if \
-                isinstance(model.params[param_name]['actual'], list) else model.params[param_name]['actual']
+            # if cross-validation is turned on, parameters in one of the fold model actual contains the max_runtime_secs
+            # parameter and not the main model that is returned.
+            if 'max_runtime_secs' == param_name and model._is_xvalidated:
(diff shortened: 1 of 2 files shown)


h2o-py/h2o/grid/grid_search.py
```diff
@@ -1420,14 +1420,14 @@ def get_hyperparams_dict(self, id, display=True):
...
-            model = h2o.get_model(model._xval_keys[0])
-
         for param_name in self.hyper_names:
-            model_params[param_name] = model.params[param_name]['actual'][0] if \
-                isinstance(model.params[param_name]['actual'], list) else model.params[param_name]['actual']
+            # if cross-validation is turned on, parameters in one of the fold model actual contains the max_runtime_secs
+            # parameter and not the main model that is returned.
+            if 'max_runtime_secs' == param_name and model._is_xvalidated:
+                xvalidated_model = h2o.get_model(model._xval_keys[0])
+                model_params[param_name] = xvalidated_model.params['max_runtime_secs']['actual']
```
honzasterba, line 1428: i would either use `[param_name]` or `['max_runtime_secs']` on both sides of the assignment

honzasterba:
since this is a bug could probably go to rel-zeno

Pscheidl:
Only "well-known" tests failed:

```
Py3.7 Small / pyunit_PUBDEV_7185_GAM_mojo_ordinal.py – r_suite4s
Py3.7 Medium-large / pyunit_pubdev_4702_algo_max_runtime_secs_large.py – r_suite
```

LGTM.

PR #4826, opened 2020-08-03
PUBDEV-7709: fix gam column error message

wendycwong:
This PR completes the work in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7709.

The same error message is generated for both the following error conditions:
1. When there not enough values in a chosen gam_columns, the knots will all have the same values;
2. When knots are provided by users and they are not in strictly-increasing order.

I decided that this should not get the same error message because it will not help the user to determine what to do.  Hence, I separated out the two error cases to help the user fixes their code.


h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -107,16 +107,24 @@ public GAM(GAMModel.GAMParameters parms, Key<GAMModel> key) {
...
-        error("knots formation", String.format("knots not sorted in ascending order. Knots at index %d: %f." +
-                "  Knots at index %d: %f", index-1, knots[index-1], index, knots[index]));
+      if (index > 0 && knots[index - 1] > knots[index]) {
+        error("knots formation", String.format("knots not sorted in ascending order for gam_column %s. " +
+                        "Knots at index %d: %f.  Knots at index %d: %f",_parms._gam_columns[gam_column_index], index-1, 
+                knots[index-1], index, knots[index]));
+        return;
+      }
+      if (index > 0 && knots[index - 1] == knots[index]) {
+        error("knots formation", String.format("chosen gam_column %s do have not enough values to generate " +
```
michalkurka, line 124: is there a typo? `do have not`? (does not have)?

h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -107,16 +107,24 @@ public GAM(GAMModel.GAMParameters parms, Key<GAMModel> key) {
...
       if (Double.isNaN(knots[index])) {
         error("knots formation", "default knots generation contain NaN.   Please specify it" +
                 " manually");
         return;
       }
-      if (index > 0 && knots[index - 1] >= knots[index]) { // knots must be increasing in order
-        error("knots formation", String.format("knots not sorted in ascending order. Knots at index %d: %f." +
-                "  Knots at index %d: %f", index-1, knots[index-1], index, knots[index]));
+      if (index > 0 && knots[index - 1] > knots[index]) {
+        error("knots formation", String.format("knots not sorted in ascending order for gam_column %s. " +
```
michalkurka, line 118: the first argument should be the name of the field the error pertains to, "knots formation" doesn't seem to be a name of a field

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7709_gam_error_messages.py
```diff
@@ -0,0 +1,45 @@
...
+from h2o.estimators.gam import H2OGeneralizedAdditiveEstimator
+
+# In this test, we want to make sure the correct error message is returned when a gam column has two
+# few values to be divided into bins and when the knot locations are not increasing in value.
+def test_gam_knots_key():
+    # bad gam_column with not enough values is chosen to be gam_column
+    h2o_data = h2o.import_file(path=pyunit_utils.locate("smalldata/prostate/prostate_complete.csv.zip"))
+    h2o_data.head()
+    myY = "CAPSULE"
+    myX = ["ID","AGE","RACE","GLEASON","DCAPS","PSA","VOL","DPROS"]
```
mn-mikke, line 17: can gen be a gam column ("GLEASON") also one of the feature columns?

wendycwong: yes.

h2o-algos/src/main/java/hex/gam/GAM.java
wendycwong: In this case, I am trying to specify what action is causing the problem.  In this case, it is in the knots formation.  I did specify which gam_column this is for in the later message.  What do you think?

h2o-algos/src/main/java/hex/gam/GAM.java
wendycwong: Yes, you are right.  Grammar issue

h2o-algos/src/main/java/hex/gam/GAM.java
Pscheidl: If this is a convention in H2O (I know we do use this across algo validations, not sure if everywhere, don't have precise stats), I'd be all for keeping the convention, as users will be instantly familiar with what is going on and which field to adjust.

h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -107,16 +107,25 @@ public GAM(GAMModel.GAMParameters parms, Key<GAMModel> key) {
   // this function will check and make sure the knots location specified in knots are valid in the following sense:
   // 1. They do not contain NaN
   // 2. They are sorted in ascending order.
-  public void failVerifyKnots(double[] knots) {
+  public void failVerifyKnots(double[] knots, int gam_column_index) {
     for (int index = 0; index < knots.length; index++) {
       if (Double.isNaN(knots[index])) {
         error("knots formation", "default knots generation contain NaN.   Please specify it" +
```
Pscheidl, line 113:
```suggestion
        error("knots formation", "Knots generated by default ended up containing a NaN value.   Please specify it" +
```

PR #4829, opened 2020-08-04
Fix Stacked Ensemble's retrieval of seed parameter value

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7667

Seed parameter was set and used but retrieved value didn't correspond to what it was set to. 
(`ensemble@parameters$seed` reported incorrect value and `ensemble@model$metalearner_model@parameters$seed` reported the correct value.)

This was caused by shadowing the `_seed` in StackedEnsembleParameters - `fillFromImpl` used `_seed` from Model.Parameters hence it contained different than the set value.


ledell (approved):
Nice catch!  Thank you @tomasfryda.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -74,7 +74,6 @@ public void initActualParamValues() {
     public Metalearner.Algorithm _metalearner_algorithm = Metalearner.Algorithm.AUTO;
     public String _metalearner_params = new String(); //used for clients code-gen only.
     public Model.Parameters _metalearner_parameters;
-    public long _seed;
```
sebhrusen, line 77: omg, we did that?

sebhrusen (approved):
thx Tomaš

PR #4831, opened 2020-08-05
Add GAM to ml-benchmark.

wendycwong:
Add gam to algo section so that ml-benchmark will run with GAM.


wendycwong:
I have collected a bunch of runs to figure out what the max and min thresholds I should set:

image

PR #4834, opened 2020-08-06
HEXDEV-746 - Add left/right split levels for categoricals to H2O Tree API

Pscheidl:
https://0xdata.atlassian.net/browse/HEXDEV-746

When walking the tree - it is useful to know which categorical levels (if the split is categorical) go to the left child and which go to the right child. Just a small checklist

- We already have NA directions in the `nas` field.
- The feature the split is made by/on is there as well in the `features` field.
- Our on-the-client assembled tree structure already has this feature. Only the scikit-like array format did not have it.

### DIFF
--- a/h2o-algos/src/main/java/hex/tree/TreeHandler.java
+++ b/h2o-algos/src/main/java/hex/tree/TreeHandler.java
@@ -65,6 +65,8 @@ public class TreeHandler extends Handler {
         args.predictions = treeProperties._predictions;
         args.tree_decision_path = treeProperties._treeDecisionPath;
         args.decision_paths = treeProperties._decisionPaths;
+        args.left_cat_split = treeProperties.left_cat_split;
+        args.right_cat_split = treeProperties.right_cat_split;
         return args;
     }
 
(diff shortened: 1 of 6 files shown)


h2o-algos/src/main/java/hex/tree/TreeHandler.java
```diff
@@ -504,11 +528,23 @@ private static void fillNodeSplitTowardsChildren(final StringBuilder nodeDescrip
...
+        return names;
+    }
+
+    private static void fillNodeCategoricalSplitDescription(final StringBuilder nodeDescriptionBuilder, final SharedTreeNode node,
+                                                            String[][] leftCatSplits, final String[][] rightCatSplits, final int pointer) {
         final SharedTreeNode leftChild = node.getLeftChild();
         final SharedTreeNode rightChild = node.getRightChild();
         final int[] leftChildLevels = extractNodeLevels(leftChild);
         final int[] rightChildLevels = extractNodeLevels(rightChild);
+        leftCatSplits[pointer] = namesForNodelevels(leftChildLevels, node.getDomainValues());
```
Pscheidl, line 546: Not much complexity added, this is done in-place of the original description builder. Only some more args are passed in to allow these to be calculated as well.

h2o-r/h2o-package/R/models.R
```diff
@@ -4747,6 +4751,14 @@ h2o.getModelTree <- function(model, tree_number, tree_class = NA) {
     res$predictions <- as.numeric(res$predictions)
   }
   
+  if(is.logical(res$left_cat_split)){
```
Pscheidl, line 4754: This must be here in case there are no categorical splits (we do have such tests) - R would assign "logical" type to such lists and it fails on type checks later on.

Pscheidl:
Change: Let's do it the slightly harder way without changing backend.

PR #4835, opened 2020-08-06
PUBDEV-7717: MOJO should expose names of score0 output values

michalkurka:


h2o-core/src/main/java/hex/Model.java
```diff
@@ -2432,8 +2432,21 @@ public boolean testJavaScoring(Frame data, Frame model_predictions, EasyPredictM
             boolean deleted = new File(filename).delete();
             if (!deleted) Log.warn("Failed to delete the file");
           }
-        }
 
+          if (! Arrays.equals(model_predictions.names(), genmodel.getOutputNames())) {
+            if (_parms._distribution == DistributionFamily.quasibinomial) {
+              Log.warn("Quasibinomial doesn't correctly return output names in MOJO"); 
+            } else if (genmodel.getModelCategory() == ModelCategory.Clustering && Arrays.equals(genmodel.getOutputNames(), new String[]{"cluster"})) {
```
wendycwong, line 2439: @michalk:  I am not sure this is correct.  @maurever went through all the trouble to make sure I use the correct names and here you said that is not correct?

h2o-py/tests/testdir_misc/pyunit_mojo_predict.py
```diff
@@ -155,8 +155,8 @@ def mojo_predict_csv_test(target_dir):
     prediction_result = h2o.mojo_predict_csv(input_csv_path=input_csv, mojo_zip_path=mojo_zip_path,
                                              output_csv_path=output_csv)
 
-    mojo_prediction_0 = float(prediction_result[0]['0'])
-    mojo_prediction_1 = float(prediction_result[0]['1'])
+    mojo_prediction_0 = float(prediction_result[0]['p0'])
+    mojo_prediction_1 = float(prediction_result[0]['p1'])
     print("Binomial prediction: p0: " + str(mojo_prediction_0))
```
wendycwong, line 160: @michalk:  Will this break customer code?

michalkurka: yes, this is a breaking change - people will have to adapt to it - should be noted in the migration document

h2o-core/src/main/java/hex/Model.java
michalkurka: @wendycwong Definitely not consistent, PredictCsv returns "cluster" and H2O uses "predict".

PR #4838, opened 2020-08-06
PUBDEV-7720 Rulefit POC

koniecsveta:
https://0xdata.atlassian.net/browse/PUBDEV-7720


h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    @Override
+    protected double[] score0(double data[], double preds[]) {
+        throw new UnsupportedOperationException("RuleFitModel doesn't support scoring on raw data. Use score() instead.");
+    }
+
+    @Override
+    public Frame score(Frame fr, String destination_key, Job j, boolean computeMetrics, CFuncRef customMetricFunc) throws IllegalArgumentException {
```
michalkurka, line 108: Should the model itself be represented by rules instead of the original models?

This will be harder to manage - load/save model will not work out of the box and MOJO will be less straightforward. 

this package https://github.com/christophM/rulefit/blob/master/rulefit/rulefit.py does represent the model as rules

cc: @meganjkurka

PR #4839, opened 2020-08-07
PUBDEV-7713 - Add implications of using Jetty 9 to migration documentation

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7713

Just a small note, the original version according to documentation was ai.h2o.org.eclipse.jetty.plus.jaas.spi.LdapLoginModule required

https://docs.h2o.ai/h2o/latest-stable/h2o-docs/security.html

This has been there since January 2018, as Jetty 8 had been upgraded. The class `ai.h2o.org.eclipse.jetty.plus.jaas.spi.LdapLoginModuleis still present, therefore this should not break anything per se. Some users might use the old eclipse version and for those, there is the note that this is no longer supported. Yet from official H2O documentation’s POV, **this is not a breaking change.**

Also updated technote: https://0xdata.atlassian.net/browse/PUBDEV-7713

### DIFF
--- a/h2o-docs/src/product/security.rst
+++ b/h2o-docs/src/product/security.rst
@@ -251,8 +251,8 @@ manipulated on the command line with the
 `keytool <http://docs.oracle.com/javase/6/docs/technotes/tools/solaris/keytool.html>`_
 command.
 
-The underlying HTTPS implementation is provided by Jetty 8 and the Java
-runtime. (**Note**: Jetty 8 was chosen to retain Java 6 compatibility.)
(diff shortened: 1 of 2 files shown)


h2o-docs/src/product/security.rst
```diff
@@ -965,9 +962,9 @@ tool:
 
     java -cp h2o.jar org.eclipse.jetty.util.security.Password username password
 
-See the `Jetty 8 HashLoginService
+See the `Jetty 9 HashLoginService
```
Pscheidl, line 965: Links are still valid.,

PR #4840, opened 2020-08-07
PUBDEV-7660 add categorical encoding validation

koniecsveta:
https://0xdata.atlassian.net/browse/PUBDEV-7660

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
@@ -190,6 +190,10 @@ public class XGBoost extends ModelBuilder<XGBoostModel,XGBoostModel.XGBoostParam
     if (_parms._distribution == DistributionFamily.quasibinomial)
       error("_distribution", "Quasibinomial is not supported for XGBoost in current H2O.");
 
+    if (_parms._categorical_encoding == Model.Parameters.CategoricalEncodingScheme.Enum) {
+      error("_categorical_encoding", "Enum encoding is not supported for XGBoost in current H2O.");
+    }
+    
     switch( _parms._distribution) {
       case bernoulli:
         if( _nclass != 2 /*&& !couldBeBool(_response)*/)


Pscheidl:
```
Stacktrace
water.exceptions.H2OModelBuilderIllegalArgumentException: Illegal argument(s) for KMeans model: KMeans_model_1596801606635_6625.  Details: ERRR on field: _categorical_encoding: EnumLimited is not supported for K-Means.
```

This looks related.

h2o-core/src/main/java/hex/CategoricalEncodingSupport.java
```diff
@@ -0,0 +1,5 @@
+package hex;
+
+public interface CategoricalEncodingSupport {
+    Model.Parameters.CategoricalEncodingScheme[] supportedCategoricalEncodingSchemes();
```
Pscheidl, line 4: 1. I'd strongly advise adding Javadoc. Such a crucial interface in our core modules only benefits from having it. Part of the contract described should be whether the returned array is modifiable or not.
2. If Set was returned (ImmutableSet ideally), searching would be faster afterwards. Not sure if this is possible due to our custom serialization, just a general question.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -722,4 +722,17 @@ private static synchronized boolean hasGPU_impl(int gpu_id) {
     warn("_max_runtime_secs", "Disabling maximum allowed runtime for cross-validation main model.");
   }
 
+  @Override
+  public Model.Parameters.CategoricalEncodingScheme[] supportedCategoricalEncodingSchemes() {
+    return new Model.Parameters.CategoricalEncodingScheme[] {
```
Pscheidl, line 727: Fresh copy is returned each time, which is nice if we want to stick with arrays. Great.

h2o-algos/src/main/java/hex/tree/gbm/GBM.java
```diff
@@ -182,6 +179,9 @@
     if ((_train != null) && (_parms._monotone_constraints != null)) {
       TreeUtils.checkMonotoneConstraints(this, _train, _parms._monotone_constraints);
     }
+    if (!Arrays.stream(supportedCategoricalEncodingSchemes()).anyMatch(_parms._categorical_encoding::equals)) {
+      error("_categorical_encoding", _parms._categorical_encoding + " is not supported for GBM.");
```
Pscheidl, line 183: 1. Do we want to compare ignoring case ? Users can always make a mistake in lower/upper case. I believe we're benevolent to this kind of user-inpuit mistakes in the API cc @michalkurka .
2. The `contains; method of a set would be faster and would not require allocating additional memory by calling `Arrays.stream`, but this is just something to think about. These arrays are small and the change won't be noticeable. But it's another piece of evidence to use Set, if possible due to serialization reasons.
3. Would be nice if we printed the list of available encodings right away.

h2o-algos/src/main/java/hex/aggregator/Aggregator.java
```diff
@@ -481,4 +484,15 @@ public void map(Chunk c) {
 
     }
   }
+
+  public Model.Parameters.CategoricalEncodingScheme[] supportedCategoricalEncodingSchemes() {
```
michalkurka, line 488: I think instead of managing a white-list we should black-list the encodings that are not supported (and perhaps add a reason why).

koniecsveta:
> ```
> Stacktrace
> water.exceptions.H2OModelBuilderIllegalArgumentException: Illegal argument(s) for KMeans model: KMeans_model_1596801606635_6625.  Details: ERRR on field: _categorical_encoding: EnumLimited is not supported for K-Means.
> ```
> 
> 
> This looks like a related test failure.

yesyes it is. I did this based on documentation but there is a discrepancy between what is documented and what is tested. EnumLimited is not documented as supported by kmeans but then there exists a test which is using it in our code (and now failing). If I am supposed to revisit docs information dealing with this matter it will take longer time because I was counting on it here.

h2o-algos/src/main/java/hex/aggregator/Aggregator.java
Pscheidl: Do we want to maintain a blacklist for the possiblity to actually **display** the reason which given encoding is not allowed?

From developer's POV, it makes more sense to me to actually define a set of allowed encodings, rather than subtracting from a global set. Such an approach has one big vulnerability - whenever a new encoding is added, it has to be blacklisted in every algorithm it's not supported at and could be potentially more error prone then explicitly enabling it. The developer usually known the algorithm the encoding is built for, but needn't realize all the places to exclude it from.

From engineering POV, whitelist makes more sense to me.

koniecsveta:
I've removed previous interface and just added missing case to be validated in XGBoost.java#init(). Other cases are already being validated or there is needed a fix in documentation in https://0xdata.atlassian.net/browse/PUBDEV-7731

PR #4847, opened 2020-08-10
PUBDEV-7722 - Kubernetes basic installation on download page

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7722

Give users the basic YAML to deploy to kubernetes. With placeholders. This one is the one we use for our CLI tool, it already contains parallel pod management and other fancy stuff, yet is kind-of short and simplistic :wink:

### DIFF
--- a/h2o-dist/index.html
+++ b/h2o-dist/index.html
@@ -376,7 +376,7 @@
 
       replaceDockerImageTag = function () {
           const versionRegexp = /(\d+\.\d+\.\d+\.\d{4,})/g; // Nightly builds have at least 4 digits at the end
-          var dockerTag = "SUBST_PROJECT_VERSION";
+          var dockerTag = "3.30.1.1";
           if (dockerTag.match(versionRegexp) != null) {
               dockerTag = "nightly" // If the last number is not 4 or more digits, substitute the version for "nightly"
           }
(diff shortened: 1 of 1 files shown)


h2o-dist/index.html
```diff
@@ -636,6 +665,89 @@ <h2>Gradle-style specification for Maven artifacts</h2>
...
+            <div id="quickstart-kubernetes" style="display:none">
+                <h2>Get started with H<sub>2</sub>O on Kubernetes</h2>
+                <p>1. Set-up kubernetes cluster and kubectl.</p>
+                <p>2. Replace the placehorders in the YAML file above with desired values</p>
+    
+                <button class="btn copy_button" id="btnCopy7" data-copied-hint="Copied!" data-clipboard-target='to_copy7'>
+                    <span class="octicon octicon-clippy"></span>
+                </button>
+                <p class="terminal" id="to_copy7">
+apiVersion: apps/v1<br/>
```
michalkurka, line 677: This is great, the other how examples are however copy&paste for the most trivial 1 node case.

Question: if we assumed a single node and hard-coded other defaults (like memory, cpu), could we provide a simple copy&paste example? Would the DNS config still be needed?

Just a question at this point ;)

Pscheidl: Yes, this is probably the best solution. This can be done. Definitely. Will adjust.

Pscheidl:
I've verified the file works on local K3S.

h2o-dist/index.html
```diff
@@ -636,6 +665,91 @@ <h2>Gradle-style specification for Maven artifacts</h2>
...
+    matchLabels:<br/>
+      app: h2o-cluster<br/>
+  template:<br/>
+    metadata:<br/>
+      labels:<br/>
+        app: h2o-cluster<br/>
+    spec:<br/>
+      containers:<br/>
+        - name: h2o-cluster<br/>
+          image: 'h2oai/h2o-open-source-k8s:latest'<br/>
```
michalkurka, line 696: I think this is technically not correct because the download page should only refer to a specific version. Eg. when you go on the download page for 3.30.0.1 you should only get version 3.30.0.1.

How about nightly builds? Will this be available for nightlies as well?

Pscheidl: Sure. the version is available in the `SUBST_PROJECT_VERSION` placeholder- I'll fix it. If the format is the `nightly` format, then it'll just substitute it with `nightly` - sounds like a plan ?

michalkurka: Sounds good, thank you!

h2o-dist/index.html
Pscheidl: Done.

h2o-dist/index.html
Pscheidl: Done. Please review. This is hard to automatically test ATM, therefore I did a manual test with manually filled version. 

I utilize the fact that nightlies have at least 4 digits at the end (correct me if I'm wrong). If it's a nightly, then the docker tag is changed to  `nightly` from the original version.

Screenshot from 2020-08-13 10-27-49
Screenshot from 2020-08-13 10-28-05

Pscheidl:
@michalkurka Thanks ! Mucho bettero now ! 
Screenshot from 2020-08-17 09-15-50

PR #4848, opened 2020-08-11
HEXDEV-762 Unique operation gives malformed output

Pscheidl:
https://0xdata.atlassian.net/browse/HEXDEV-762

The original JIRA mentions three problems:

1. NAs not counted as a categorical level by the `unique` operation - fixed in this PR by adding NA at the end of the vec if the original vec has naCnt > 0.
1. Temporary frames contain vecs with domain of the original vec from the original frame (the domain might change) **NOT** fixed by this PR.
1. `unique` operation shows "breaks" on high number of rows (10 million) and above. This is to be investigated and is **NOT** addressed by this PR.

### DIFF
--- a/h2o-core/src/main/java/water/rapids/ast/prims/assign/AstTmpAssign.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/assign/AstTmpAssign.java
@@ -57,7 +57,6 @@ public class AstTmpAssign extends AstPrimitive {
         throw new IllegalArgumentException("Temp ID " + id + " already exists");
     }
     Frame dst = new Frame(id, srcFrame._names, srcFrame.vecs());
-    FrameUtils.shrinkDomainsToObservedSubset(dst);
     return new ValFrame(env._ses.track_tmp(dst)); // Track new session-wide ID
   }
 }
--- a/h2o-core/src/main/java/water/rapids/ast/prims/assign/AstTmpAssign.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/assign/AstTmpAssign.java
@@ -9,6 +9,7 @@ import water.rapids.Val;
 import water.rapids.ast.AstPrimitive;
 import water.rapids.ast.AstRoot;
 import water.rapids.vals.ValFrame;
+import water.util.FrameUtils;
 
 /**
  * Assign a temp.  All such assignments are final (cannot change), but the temp can be deleted.  Temp is returned for
@@ -56,6 +57,7 @@ public class AstTmpAssign extends AstPrimitive {
         throw new IllegalArgumentException("Temp ID " + id + " already exists");
     }
     Frame dst = new Frame(id, srcFrame._names, srcFrame.vecs());
+    FrameUtils.shrinkDomainsToObservedSubset(dst);
     return new ValFrame(env._ses.track_tmp(dst)); // Track new session-wide ID
   }
 }
--- a/h2o-core/src/main/java/water/rapids/ast/prims/assign/AstTmpAssign.java
+++ b/h2o-core/src/main/java/water/rapids/ast/prims/assign/AstTmpAssign.java
@@ -9,7 +9,6 @@ import water.rapids.Val;
 import water.rapids.ast.AstPrimitive;
 import water.rapids.ast.AstRoot;
 import water.rapids.vals.ValFrame;
-import water.util.FrameUtils;
 
 /**
  * Assign a temp.  All such assignments are final (cannot change), but the temp can be deleted.  Temp is returned for
@@ -57,7 +56,6 @@ public class AstTmpAssign extends AstPrimitive {
         throw new IllegalArgumentException("Temp ID " + id + " already exists");
     }
     Frame dst = new Frame(id, srcFrame._names, srcFrame.vecs());
-    FrameUtils.shrinkDomainsToObservedSubset(dst);
     return new ValFrame(env._ses.track_tmp(dst)); // Track new session-wide ID
   }
 }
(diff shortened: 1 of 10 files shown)


h2o-core/src/main/java/water/rapids/ast/prims/assign/AstTmpAssign.java
```diff
@@ -56,6 +57,7 @@ public ValFrame apply(Env env, Env.StackHelp stk, AstRoot[] asts) {
         throw new IllegalArgumentException("Temp ID " + id + " already exists");
     }
     Frame dst = new Frame(id, srcFrame._names, srcFrame.vecs());
+    FrameUtils.shrinkDomainsToObservedSubset(dst);
```
michalkurka, line 60: Why is this needed?

Pscheidl: I just realized I made a mistake. This is required, but not at `AstTmpAssign` but at `AstRowSlice`. This solves the point number 2 in the description - vectors in new frames created by row slicing had domains of the old vectors - unique did not work vell on categorical columns then, but it was just a symptom.

h2o-core/src/main/java/water/rapids/ast/prims/assign/AstTmpAssign.java
```diff
@@ -9,6 +9,7 @@
 import water.rapids.ast.AstPrimitive;
 import water.rapids.ast.AstRoot;
 import water.rapids.vals.ValFrame;
+import water.util.FrameUtils;
```
satai, line 12: Is it necessary?

Pscheidl: Why not ?

Pscheidl: Oh I see. No it's not ! Thanks.

h2o-core/src/main/java/water/rapids/ast/prims/assign/AstTmpAssign.java
michalkurka: This would be a significant change of the behavior - I think this deserves a separate jira and a PR.

Pscheidl: Ok let's split it.

Pscheidl: Reverted everything except for the original issue with NAs.

michalkurka: H2O is consistent with R, so it is not that clear the second proposed change can be actually implemented without consequences, eg.:

> X <- data.frame(c1=c(1, 2), f1=as.factor(c("a", "b")))
> x1 = X[1, ]
> x1
  c1 f1
1  1  a
> x1$f1
[1] a
Levels: a b
> levels(x1$f1)
[1] "a" "b"

Pscheidl: This would force us to change the way how we calculate uniques for categoricals - which can be done. There is supposedly a performance problem anyway (asked for more details in the JIRA) and this can be done - shrink the domain on a copy of the original vector which is immediately dropped is one way to achieve correctness. What bothers me are the yet unknown consequences of this behavior.

Would it be too politically incorrect to consider aligning it with Python instead of R, as the majority of people uses that anyway ? It's just a question spoken out loud. Then the baseline would be fixed for every operation in H2O.

Pscheidl: Is there any reason for a behavior like that ?

michalkurka: The question of `unique` is IMHO orthogonal to the rest of the discussion. The current implementation is incorrect because it is not returning the unique values actually found in the categorical column.

The second question is what should data munging do to the categorical values. It goes beyond what R or Python does and in this PR I am not presenting my opinion on what should we do - more saying that we cannot simply implement your proposed solution as this would break current behavior for some users.

We can have different behavior in Python and R and make sure peoples' expectations are met. But this is a discussion we should have in #h2o-3-api-updates and similar channels. Also, such a breaking change cannot be IMHO done in a fix release, users don't expect we would break stuff in fix releases.

Pscheidl: The `unique` function now works just fine, as it assumes the domain on a vec returns actual VALID domain, not domain of some of it's predecessors. 

Unless the frame has been created by AstRowSlice, it will work just fine with this fix.

Problem is in the layer below - provided the contract of `getDomain()` method is that it returns **actual** domain, then the fix to the second question is to make it so. But this will be another PR.

Is this PR worth merging for you, as it actually fixes a bug on one layer of our code ?

michalkurka: The PR implementation of the unique function doesn't seem correct to me.

- it returns values that might not actually be in the column,
- it automatically includes NAs, this should be only optional as it would break backward compatibility.

> The unique function now works just fine, as it assumes the domain on a vec returns actual VALID domain, not domain of some of it's predecessors.

Having a domain that is larger than the actual values in the column is 100% correct use case and essential to work for model building. In model building, we make a union of the training domain and validation domain - both training and validation frame need to have the same domain for categorical encodings to work.

Pscheidl:
- Actual domain is now collected just in the the unique values are calculated for categorical, as H2O might provide a larger domain than present in the vector.
- NAs are included only when explicitly enabled.
- Tested for Python and R (both enabled and disabled states).

cc @michalkurka

PR #4849, opened 2020-08-11
README - recommend to install `R.utils`

satai:


README.md
```diff
@@ -277,7 +277,7 @@ Install the following R packages:
 To install these packages from within an R session:
 
 ```r
-pkgs <- c("RCurl", "jsonlite", "statmod", "devtools", "roxygen2", "testthat")
+pkgs <- c("RCurl", "jsonlite", "statmod", "devtools", "roxygen2", "testthat", "R.utils")
```
michalkurka, line 280: what is it needed for? we should know ;)

satai: When I start with a totally clean computer and install only the documented packages, I get

```> setwd(normalizePath(dirname(R.utils::commandArgs(asValues=TRUE)$"f")))
Error in loadNamespace(name) : there is no package called ‘R.utils’
```

when runnning tests (`R -f runit_coxph_concordance_heart.r`)

michalkurka: The list of packages to run tests is longer, IMHO maybe the guide should only list the packages needed to build without tests. For running tests, there is a utility function that installs all required packages.

satai: OK, I will make it clear in README...

PR #4851, opened 2020-08-11
PUBDEV-7704: Document supported encodings for csv

hannah-tillman:
I added the supported encoding for csv to the notes section right beneath the file formats. Let me know if that's okay or if it should be its own small section.

### DIFF
--- a/h2o-docs/src/product/getting-data-into-h2o.rst
+++ b/h2o-docs/src/product/getting-data-into-h2o.rst
@@ -10,7 +10,7 @@ Supported File Formats
 
 H2O currently supports the following file types:
 
-- CSV (delimited) files (including GZipped CSV)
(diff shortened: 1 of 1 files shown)


h2o-docs/src/product/getting-data-into-h2o.rst
```diff
@@ -21,7 +21,7 @@ H2O currently supports the following file types:
 
 **Notes**: 
  
- - H2O does not support UTF-16 encoding for csv files. Please convert to UTF-8 encoding.
+ - H2O supports UTF-8 encodings for CSV files. Please convert UTF-16 encodings to UTF-8 encoding before parsing CSV files into H2O
```
g-eoj, line 24: Probably should add a period for consistency.

PR #4852, opened 2020-08-12
Do not suppress IOException when loading DKV value

michalkurka:


h2o-core/src/main/java/water/persist/PersistNFS.java
```diff
@@ -51,24 +51,15 @@ public static InputStream openStream(Key k) throws IOException {
...
-      Log.debug("[h2o] Problem ignored: "+e.toString());
-      return null;
+    Key<?> k = v._key;
+    final long skip = k.isChunkKey() ? water.fvec.NFSFileVec.chunkOffset(k) : 0;
+    try (FileInputStream s = new FileInputStream(getFileForKey(k));
+         FileChannel fc = s.getChannel()) {
+      fc.position(skip);
+      AutoBuffer ab = new AutoBuffer(fc, true, Value.NFS);
+      byte[] b = ab.getA1(v._max);
+      ab.close();
```
Pscheidl, line 61: Should we make sure this is always closed ? Unlikely to crash I'd say.

michalkurka: for reading `close` only attempts to close the underlying `FileChannel` if it fails there an exception is thrown with a stacktrace where it came from exactly - which might be helpful for debugging

for resource management, it doesn't really have any effect because resources are properly closed by the try-block

Pscheidl: Good. Thanks.

PR #4857, opened 2020-08-14
Remove redundant hyperparamter combos of min_rows from grid search in AutoML

michalkurka:
Small improvement in it's simplest form, can be generalized


sebhrusen:
> can be generalized

Yes, I'll need to come up with better abstractions to define hyperparams ranges depending on various use-cases: this is something we will surely do more and more, first by replacing our monolithic grids with more specific ones.
But for now, this change makes sense.

PR #4859, opened 2020-08-17
Fix test: make early stopping reproducible

michalkurka:


h2o-r/tests/testdir_algos/randomforest/runit_RF_early_stopping.R
```diff
@@ -11,7 +11,8 @@ rf_early_stopping_suite <- function() {
     ref_model <- h2o.randomForest(x=x, y=y, training_frame=train, seed=1)
     ref_trees <- ref_model@model$model_summary$number_of_trees
     stop_model <- h2o.randomForest(x=x, y=y, training_frame=train, seed=1,
-                                   stopping_metric="AUC", stopping_rounds=3, stopping_tolerance=0.1)
+                                   stopping_metric="AUC", stopping_rounds=3, stopping_tolerance=0.1,
+                                   score_each_iteration=TRUE)
```
sebhrusen, line 15: why is it better to use `score_each_iteration` than `score_tree_interval` in this case? For RF, it's equivalent to `score_tree_interval=1` right? (mainly curious as we never use this param in AutoML, as it looks time-consuming).

michalkurka: It is the same thing. In this case, we are comparing stopping against a model trained without early stopping. I chose score_tree_interval=1 to make sure we definitely stop early.

PR #4861, opened 2020-08-18
Check for at sign before trying to run K8S test cluster

Pscheidl:
This is for the convenience of people reading the error log - as the message about volume binding failed is hidden somewhere in the middle of a long log.

This way, it will fail right away and not wait for H2O to start. Also, the error message will be clear to anyone.

Change proposed by @michalkurka .


Pscheidl:
Requires tests to befinished before merging:http://mr-0xc1:8080/view/H2O-3/job/h2o-3-nightly-pipeline/job/pavel%252Fk8s-tests-path-check/1/

PR #4864, opened 2020-08-19
PUBDEV-7367: Verify grid search functionality for GAM, Extend grid search to allow subspaces in hyper space

km2018:
JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7367

PR adds unit tests to verify grid search functionality for GAM. Also, allows users to create more complex hyper spaces for grid search by adding "subspaces" key and functionality to grid search backend.


h2o-core/src/main/java/hex/schemas/GridSearchSchema.java
```diff
@@ -62,19 +63,26 @@
 
   private static final int SEQUENTIAL_GRID_SEARCH = 1; // 1 model built at a time = sequential :)
 
+  private static Map<String, Object[]> paramValuesToArray(Map<String, Object> params) {
+    Map<String, Object[]> result = new HashMap<>();
+    for (Map.Entry<String, Object> e : params.entrySet()) {
+      String k = e.getKey();
+      Object v = e.getValue();
+      Object[] arr = SUBSPACES.equals(k) ? ((List) v).stream().map(x -> paramValuesToArray((Map<String, Object>) x)).toArray()
```
sebhrusen, line 71:
```suggestion
      Object[] arr = SUBSPACES.equals(k) ? ((List) v).stream().map(x -> paramValuesToArray((Map<String, Object>) x)).toArray(Map[]::new)
```

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -222,14 +234,24 @@ public MP getParams() {
     public ModelParametersBuilderFactory<MP> getParametersBuilderFactory() {
       return _paramsBuilderFactory;
     }
-
-    protected MP getModelParams(MP params, Object[] hyperParams) {
+    
+    private Map<String, Object[]>[] extractSubspaces() {
+      if(!_hyperParams.containsKey(SUBSPACES)) { return new Map[0]; }
+      return Stream.of(_hyperParams.get(SUBSPACES)).toArray(Map[]::new);
```
sebhrusen, line 240: ok, I think I know why you need this strange conversion: see my suggestion in h2o-core/src/main/java/hex/schemas/GridSearchSchema.java
The original array was created as an `Object[]` instead of a`Map[]`, so the Java interpreter fails casting it to `Map[]` later. You have to create it as a `Map[]`, then it is casted as `Object[]`: casting is mainly useful for the compiler, not the interpreter, so that you'll then be able to read it directly and cast it here:
```suggestion
      return (Map<>[])_hyperParams.get(SUBSPACES));
```

sebhrusen (commented):
last suggestion (please try it, I didn't), then if it works, I'll approve.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
km2018: Had to cast it to ```Map<String, Object[]>[]```, but this seemed to work. Thanks for the suggestion.

sebhrusen (approved):

PR #4865, opened 2020-08-20
Make AutoML agrs test that depends on real-world time more stable

michalkurka:
We first measure how long it takes to complete 5 models. Then we use
this number as max_runtime_secs for the actually tested scenario.
We shoulnd't be able to built much more than 5 models (20) and
thus the max_runtime_secs should kick-in.

### DIFF
--- a/h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
+++ b/h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
@@ -409,14 +409,21 @@ def test_max_runtime_secs_can_be_set_in_combination_with_max_models_and_max_mode
 
 def test_max_runtime_secs_can_be_set_in_combination_with_max_models_and_max_runtime_wins():
     ds = import_dataset()
-    aml = H2OAutoML(project_name="py_all_stopping_constraints", seed=1, max_models=20, max_runtime_secs=12)
+    aml_probe = H2OAutoML(project_name="py_all_stopping_constraints_probe",
+                          max_models=5,
+                          seed=1, exclude_algos=["XGBoost", "StackedEnsemble"])
+    aml_probe.train(y=ds['target'], training_frame=ds['train'])
+    probe_duration = int(aml_probe.training_info['duration_secs'])
+
+    aml = H2OAutoML(project_name="py_all_stopping_constraints",
+                    max_models=20, max_runtime_secs=probe_duration,
+                    seed=1, exclude_algos=["XGBoost", "StackedEnsemble"])
     aml.train(y=ds['target'], training_frame=ds['train'])
-    max_runtime = aml._build_resp['build_control']['stopping_criteria']['max_runtime_secs']
-    max_models = aml._build_resp['build_control']['stopping_criteria']['max_models']
-    assert max_runtime == 12
-    assert max_models == 20
+
+    assert aml._build_resp['build_control']['stopping_criteria']['max_runtime_secs'] == probe_duration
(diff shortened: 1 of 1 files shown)


h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
```diff
@@ -409,14 +409,21 @@ def test_max_runtime_secs_can_be_set_in_combination_with_max_models_and_max_mode
 
 def test_max_runtime_secs_can_be_set_in_combination_with_max_models_and_max_runtime_wins():
     ds = import_dataset()
-    aml = H2OAutoML(project_name="py_all_stopping_constraints", seed=1, max_models=20, max_runtime_secs=12)
+    aml_probe = H2OAutoML(project_name="py_all_stopping_constraints_probe",
+                          max_models=5,
+                          seed=1, exclude_algos=["XGBoost", "StackedEnsemble"])
```
sebhrusen, line 414: why is your probe using less models?
I would understand if `max_models` was the same in both cases, but here I don't see the connection between `t_probe` and `t_tested`...

Also, excluding `StackedEnsemble` definitely makes sense given that SEs are allowed to use extra-time after `max_runtime_secs`, but why do you want to exclude `XGBoost`?

h2o-py/tests/testdir_algos/automl/pyunit_automl_args.py
```diff
@@ -409,14 +409,21 @@ def test_max_runtime_secs_can_be_set_in_combination_with_max_models_and_max_mode
...
-    max_runtime = aml._build_resp['build_control']['stopping_criteria']['max_runtime_secs']
-    max_models = aml._build_resp['build_control']['stopping_criteria']['max_models']
-    assert max_runtime == 12
-    assert max_models == 20
+
+    assert aml._build_resp['build_control']['stopping_criteria']['max_runtime_secs'] == probe_duration
+    assert aml._build_resp['build_control']['stopping_criteria']['max_models'] == 20
     assert aml.leaderboard.nrows < 20
-    assert int(aml.training_info['duration_secs']) < 2*max_runtime  # being generous to avoid errors on slow Jenkins
+    assert int(aml.training_info['duration_secs']) < 2*probe_duration  # being generous to avoid errors on slow Jenkins
```
sebhrusen, line 426: the assertion on `nrows` was wrong by contract, as there's nothing preventing us to have 20 models on a very fast machine, it should be:
```
    assert aml.leaderboard.nrows <= 20
```

As we have now a probe, maybe we can be less generous.
This would be my suggestion:
aml_probe: max_models=5   ->  probe_duration
aml_tested: max_models=5, max_runtime_secs=probe_duration/2  -> tested_duration
assert aml_tested.leaderboard.nrows <= 5
assert int(aml_tested.training_info['duration_secs']) < probe_duration

As we use the same `max_models` in both, the 2 assertions should be valid by contract.

sebhrusen (commented):
@michalkurka I like the probe idea, thanks.
Added some suggestions if you want to finish this, otherwise I can try it.

sebhrusen:
> We shoulnd't be able to built much more than 5 models (20) and thus the max_runtime_secs should kick-in.

This is not true actually, there is a major difference when using only `max_models` and when enabled `max_runtime_secs`.
In the first case, all models are trained until convergence.
In the second case, they have a time budget allocated by AutoML (sharing the total budget between default models, and then the leftovers between grids).
So you can theoretically still have 20 models trained in the time it took to train only 5 models (with `max_models`).

I need to verify the doc, this may not be completely clear there: this behaviour was changed in 3.30.

PR #4867, opened 2020-08-21
PUBDEV-7672: TargetEncoder support for multiclass problems.

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-7672

- full support for multiclass
- fixed rounding issue with regression that was causing a small difference between Mojo predictions and in-memory model predictions, discovered when adding mojo integration tests.
- Also includes the fix for https://0xdata.atlassian.net/browse/PUBDEV-7714 as this is required for some multiclass tests to pass.


h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderModelOutputV3.java
```diff
@@ -1,14 +1,7 @@
 package hex.schemas;
 
 import ai.h2o.targetencoding.TargetEncoderModel;
-import water.api.API;
 import water.api.schemas3.ModelOutputSchemaV3;
 
-public class TargetEncoderModelOutputV3 extends ModelOutputSchemaV3<TargetEncoderModel.TargetEncoderOutput, TargetEncoderModelOutputV3> {
-  @API(help = "Prior mean value", direction = API.Direction.OUTPUT)
-  public double prior_mean;
```
sebhrusen, line 9: a single priorMean doesn't make any sense in case of multiclass.
As it seemed to have been mainly exposed only for testing purposes on the client side, I prefer to completely remove it.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
```diff
@@ -70,20 +76,20 @@ protected boolean defaultDropConsCols() {
...
+
+    public final TargetEncoderParameters _parms;
+    public final int _nclasses;
+    public final IcedHashMap<String, Frame> _target_encoding_map;
+    public final IcedHashMap<String, Boolean> _te_column_to_hasNAs;
     
-    public IcedHashMap<String, Frame> _target_encoding_map;
-    public TargetEncoderParameters _parms;
-    public IcedHashMap<String, Boolean> _te_column_to_hasNAs;
-    public double _prior_mean;
```
sebhrusen, line 77: Completely removed `prior_mean` from output: this is unnecessary as the priorMean can easily be computed from the encodings frames, and besides, it is problematic for multiple reasons:
- meaningless in case of multiclass (there is a prior for each target value).
- the global prior should be used wisely as it leaks data, so we should not encourage anyone using it directly.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderBroadcastJoin.java
```diff
@@ -108,7 +108,7 @@ public void map(Chunk[] cs) {
         // We are allowed to do casting to `int` as we have validation before submitting this MRTask
         int foldValue = _foldColumnIdx != -1 ? (int) cs[_foldColumnIdx].at8(i) : 0;
         double[] numDenArray = _encodingDataPerNode[foldValue];
-        numDenArray[2*level] = numeratorChunk.at8(i);
+        numDenArray[2*level] = numeratorChunk.atd(i);
```
sebhrusen, line 111: the rounding fix that caused issues with regression.

ledell:
Hey @sebhrusen, this looks great!

Are you planning on adding R tests for multiclass (maybe in a follow-up PR)?  I didn't see any.

sebhrusen:
@ledell adding basic multiclass tests now

sebhrusen:
@michalkurka @ledell @Pscheidl don't be to afraid of reviewing this PR, among all the changed files:
- 7 are from the main `targetencoder` package.
- 4 are from the `genmodel` package for mojo support.
- all the rest are tests.
Thanks!

ledell (approved):
Code looks good @sebhrusen!  

Here are my comments:

The pyunits have hardcoded values to check against, but the runits just check the structure of the returned objects (dimension of the encoded data & that correct columns are there).  Is it a redundant effort to double-check the golden values in R since we check them already in Python?  It's pulling the same computed TE values, I guess, so maybe it's ok to skip in R...  

You're using different datasets in runit and pyunit. Is that intentional, and if so, why?

- Python: `smalldata/titanic/titanic_expanded.csv`
- R: `smalldata/gbm_test/titanic.csv`

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java
```diff
@@ -51,14 +59,26 @@ public static int addKFoldColumn(Frame frame, String name, int nfolds, long seed
...
-    return numeratorVec.mean() / denominatorVec.mean();
+  static double computePriorMean(Frame encodings) {
+    assert encodings.find(TARGETCLASS_COL) < 0;
+    return computePriorMean(encodings, NO_TARGET_CLASS);
+  }
+  
+  static double computePriorMean(Frame encodings, int targetClass) {
+    int tcIdx = encodings.find(TARGETCLASS_COL);
+    assert (targetClass == NO_TARGET_CLASS) == (tcIdx < 0);
+    Frame fr = null;
```
Pscheidl, line 70: Note to self: scope would work as well

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java
```diff
@@ -51,14 +59,26 @@ public static int addKFoldColumn(Frame frame, String name, int nfolds, long seed
...
+    try {
+      fr = tcIdx < 0 ? encodings:filterByValue(encodings, tcIdx, targetClass);
+      Vec numeratorVec = fr.vec(NUMERATOR_COL);
+      Vec denominatorVec = fr.vec(DENOMINATOR_COL);
+      assert numeratorVec!=null;
+      assert denominatorVec!=null;
+      return numeratorVec.mean() / denominatorVec.mean();
```
Pscheidl, line 77: Thumbs up for using rollups.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java
```diff
@@ -69,24 +89,81 @@ static double calculatePriorMean(Frame encodings) {
...
+        renameColumn(result, "nrow", DENOMINATOR_COL);
+//        printFrame(result);
+      }
 
-    Frame result = new AstGroup().performGroupingWithAggregations(fr, groupBy, aggs).getFrame();
-    //change the default column names assigned by the aggregation task
-    renameColumn(result, "sum_" + fr.name(targetIdx), NUMERATOR_COL);
-    renameColumn(result, "nrow", DENOMINATOR_COL);
-    return register(result);
+      Scope.untrack(result);
```
Pscheidl, line 160: Nice failsafe mechanics. Merci.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderMojoWriter.java
```diff
@@ -45,18 +44,17 @@ protected void writeExtraInfo() {
...
+    TargetEncoderOutput output = model._output;
+    TargetEncoderParameters teParams = output._parms;
     writekv("with_blending", teParams._blending);
     if (teParams._blending) {
       writekv("inflection_point", teParams._inflection_point);
       writekv("smoothing", teParams._smoothing);
     }
-    writekv("priorMean", output._prior_mean);
     
     Map<String, Boolean> col2HasNAs = output._te_column_to_hasNAs;
```
Pscheidl, line 55: This could be simplified to a set ? If the column is in the set, it has NAs. We're talking about the original code here, so let's leave it. Just making some notes.

Pscheidl: Vecs have NA counts in rollups available, not sure if this is even required.

Pscheidl (approved):

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderMojoWriter.java
sebhrusen: good tip, I agree that a set would be better here: although it needs to be changed carefully as afaik the mojo stores this map (also not optimally) as a list of kv pairs.
Will add a comment and try to chnage it in a future PR.

Pscheidl: Thank you so much. That's as much as I was hoping for, nothing more. I owe you one :croissant: .

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java
sebhrusen: agree, there's only one frame to manage though, so the direct call to delete may be simpler in this case.

sebhrusen:
latest build looked good: 3 tests failing totally unrelated with TE -> merging.

PR #4832, opened 2020-08-05
[GAM] NaN MOJO fix + tests + cleanup

honzasterba:

PR #4869, opened 2020-08-22
PUBDEV-7366: GAM cross-validation

wendycwong:
This PR completes the work required in JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7366

There are only 7 files to review.  The other four are tests.

Users can now use cross-validation to find the best alpha/lambda values when building a GAM model.  In the near future, we will be adding more hyper-parameters search capability.  Stay tuned...

I have added to Java backend to ensure that cross-validation works for GAM.  I added tests in R, Java and Python to make sure everything checks out.

### DIFF
--- a/h2o-algos/src/main/java/hex/gam/GAM.java
+++ b/h2o-algos/src/main/java/hex/gam/GAM.java
@@ -1,6 +1,9 @@
 package hex.gam;
 
-import hex.*;
+import hex.DataInfo;
+import hex.ModelBuilder;
+import hex.ModelCategory;
+import hex.ModelMetrics;
 import hex.gam.GAMModel.GAMParameters;
 import hex.gam.MatrixFrameUtils.GamUtils;
 import hex.gam.MatrixFrameUtils.GenerateGamMatrixOneColumn;
@@ -16,6 +19,7 @@ import water.exceptions.H2OModelBuilderIllegalArgumentException;
 import water.fvec.Frame;
 import water.fvec.Vec;
 import water.util.ArrayUtils;
+import water.util.IcedHashSet;
 import water.util.Log;
 
 import java.util.ArrayList;
@@ -31,7 +35,15 @@ import static hex.glm.GLMModel.GLMParameters.Family.ordinal;
 
 
 public class GAM extends ModelBuilder<GAMModel, GAMModel.GAMParameters, GAMModel.GAMModelOutput> {
-
+  double[][] _knots;
+  boolean _cvOn = false; // set to true if cross-validation is enabled
+  boolean _doInit = true; // perform init inside computeImpl if true and false to skip during main model of cv on
+  double[] _cv_alpha = null;  // best alpha value found from cross-validation
+  double[] _cv_lambda = null; // bset lambda value found from cross-validation
+  // add future hyper-parameters that we want to find during cross-validation here
+  
+  IcedHashSet<Key<Frame>> _validKeys = null;  // store validation frame keys from various folds
+  
   @Override
   public ModelCategory[] can_build() {
     return new ModelCategory[]{ModelCategory.Regression};
@@ -70,15 +82,46 @@ public class GAM extends ModelBuilder<GAMModel, GAMModel.GAMParameters, GAMModel
     super(parms, key);
     init(false);
   }
-  
-  double[][] _knots;
 
-  /***
-   * This method will look at the keys of knots stored in _parms._knot_ids and copy them over to double[][]
-   * array.
-   *
-   * @return double[][] array containing the knots specified by users
-   */
+  // cross validation can be used to choose the best alpha/lambda values among a whole collection of alpha
+  // and lambda values.  Future hyperparameters can be added for cross-validation to choose as well.
+  @Override
+  public void computeCrossValidation() {
+    _cvOn = true;  // set cross-validation on to be true
+    _validKeys = new IcedHashSet<>();
+    if (error_count() > 0)
+      throw H2OModelBuilderIllegalArgumentException.makeFromBuilder(GAM.this);
+    super.computeCrossValidation();
+  }
+
+  // find the best alpha/lambda values used to build the main model moving forward by looking at the devianceValid
+  @Override
+  public void cv_computeAndSetOptimalParameters(ModelBuilder[] cvModelBuilders) {
+    double deviance_valid = Double.POSITIVE_INFINITY;
+    double best_alpha = 0;
+    double best_lambda = 0;
+    for (int i = 0; i < cvModelBuilders.length; ++i) {  // run cv for each lambda value
+      GAMModel g = (GAMModel) cvModelBuilders[i].dest().get();
+      if (g._output._devianceValid < deviance_valid) {
+        best_alpha= g._output._best_alpha;
+        best_lambda = g._output._best_lambda;
+      }
+      
+      g.write_lock(_job);
+      g.update(_job);
+      g.unlock(_job);
+    }
+    _cv_alpha = new double[]{best_alpha};
+    _cv_lambda = new double[]{best_lambda};
+    _doInit = false;  // for main model, disable the init()
+    _cvOn = false;   // building main model now.  Set CV off.
+  }
+    /***
+     * This method will look at the keys of knots stored in _parms._knot_ids and copy them over to double[][]
+     * array.
+     *
+     * @return double[][] array containing the knots specified by users
+     */
   public double[][] generateKnotsFromKeys() {
     int numGamCols = _parms._gam_columns.length;
     double[][] knots = new double[numGamCols][];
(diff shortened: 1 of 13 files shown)


h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
```diff
@@ -93,6 +92,12 @@ public static GLMParameters copyGAMParams2GLMParams(GAMParameters parms, Frame t
     setParamField(parms, glmParam, true, field2);
     glmParam._train = trainData._key;
     glmParam._valid = valid==null?null:valid._key;
+    glmParam._nfolds = 0; // always set nfolds to 0 to disable cv in GLM.  It is done in GAM
```
Pscheidl, line 95: Can be removed. This is always zero by default and the class is instantiated by using the default constructor.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
```diff
@@ -93,6 +92,12 @@ public static GLMParameters copyGAMParams2GLMParams(GAMParameters parms, Frame t
...
+    glmParam._fold_assignment = Model.Parameters.FoldAssignmentScheme.AUTO;
```
Pscheidl, line 96: Can be removed. Default is the same.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
```diff
@@ -93,6 +92,12 @@ public static GLMParameters copyGAMParams2GLMParams(GAMParameters parms, Frame t
...
+    glmParam._keep_cross_validation_fold_assignment = false;
```
Pscheidl, line 97: Can be removed. Default is the same.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
```diff
@@ -93,6 +92,12 @@ public static GLMParameters copyGAMParams2GLMParams(GAMParameters parms, Frame t
...
+    glmParam._keep_cross_validation_models = false;
```
Pscheidl, line 98: This is true by default and I believe we might omit this line as well and leave it to true - CV is disabled anyway.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
```diff
@@ -93,6 +92,12 @@ public static GLMParameters copyGAMParams2GLMParams(GAMParameters parms, Frame t
...
+    glmParam._keep_cross_validation_predictions = false;
+    glmParam._is_cv_model = false; // disable cv in GLM.
```
Pscheidl, line 100: False by default again, even if not explicitly stated. No need to do it explicitly.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java
```diff
@@ -48,19 +48,29 @@ public void map(Chunk[] chk, NewChunk[] newGamCols) {
...
-      int binIndex = locateBin(xval,_knots); // location to update
-      // update from F matrix F matrix = [0;invB*D;0] and c functions
-      GamUtilsCubicRegression.updateFMatrixCFunc(basisVals, xval, binIndex, _knots, crSplines._hj, _bInvD);
-      // update from a+ and a- functions
-      GamUtilsCubicRegression.updateAFunc(basisVals, xval, binIndex, _knots, crSplines._hj);
-      // copy updates to the newChunk row
-      for (int colIndex = 0; colIndex < _numKnots; colIndex++) {
-        newGamCols[colIndex].addNum(basisVals[colIndex]);
-        gamRowSum += Math.abs(basisVals[colIndex]);
+      if (chk[1].atd(rowIndex) != 0) {  // consider weight column value during gamification.  If 0, insert rows of NaN.
```
Pscheidl, line 51: The comment says row this row of the matrix will be inserted with `NaN`, yet in the `else` block `0`s are inserted. I believe zeros are correct, as when substituted to the final linear formula , the result will be zero, as the result of the polynomial function(s) approximating the space between knots will be zero.

Am I right @wendycwong ?

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java
```diff
@@ -48,19 +48,29 @@ public void map(Chunk[] chk, NewChunk[] newGamCols) {
...
+        double xval = chk[0].atd(rowIndex);
+        if (Double.isNaN(xval)) { // fill with NaN
+          for (int colIndex = 0; colIndex < _numKnots; colIndex++)
```
Pscheidl, line 54: Here, the NaNs make sense to me - if the predictor is none, there is no other choice really.

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
```diff
@@ -84,7 +84,7 @@ private void imputeMissingWithMeans(double[] data) {
 
     if (data.length == nfeatures()) { // using centered gam cols, nfeatures denotes centered gam columns
       for (int ind = 0; ind < _numsCenter; ind++)
-        if (Double.isNaN(data[ind + _cats])) data[ind] = _numNAFillsCenter[ind];
+        if (Double.isNaN(data[ind + _cats])) data[ind + _cats] = _numNAFillsCenter[ind];
```
Pscheidl, line 87: Please note this part has wrong indexing and is being fixed: https://github.com/h2oai/h2o-3/pull/4832 (but not merged yet at the moment).

We might copy & paste the code from there if we want to.

h2o-algos/src/test/java/hex/gam/GamCVTest.java
```diff
@@ -0,0 +1,93 @@
...
+// 1. fold_assignment = random
+// 2. keep_cross_validation_model
+// 3. keep_cross_validation_predictions
+// 4. keep_cross_validation_fold_assignment
+//
+// If we keep the cross-validation models and the fold assignment, then the prediction using the folds and
+// the predictions kept from cross-validation should yield the same result!  
+  @Test
+  public void testCVBinomial() {
+    Frame train;
```
Pscheidl, line 32: Knitpick/codestyle: No need to define al lthe variables like at if scopes are used.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
wendycwong: @pavel:  In this case, I am copying GAM parameters to GLM parameters.  If I want cross-validation, I will set nfolds = 3, and therefore after the copying of parameters, GLM parameter nfolds will be 3.  Hence, I need to set it to 0.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
wendycwong: @pavel: In this case, I am copying GAM parameters to GLM parameters. If I want cross-validation, I will set nfolds = 3, and therefore after the copying of parameters, GLM parameter nfolds will be 3. Hence, I need to set it to 0.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
wendycwong: @pavel: In this case, I am copying GAM parameters to GLM parameters. If I want cross-validation and set _keep_cross_validation_fold_assignment, and therefore after the copying of parameters, GLM parameter _keep_cross_validation_fold_assignment will also be set.  Hence, I need to manually set them.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
wendycwong: @pavel: In this case, I am copying GAM parameters to GLM parameters. If I want cross-validation and set _keep_cross_validation_models, and therefore after the copying of parameters, GLM parameter _keep_cross_validation_models will also be set. Hence, I need to manually set them.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GamUtils.java
wendycwong: @pavel: In this case, I am copying GAM parameters to GLM parameters. If I want cross-validation and set nfolds > 1, _is_cv_model will be true and therefore after the copying of parameters, GLM parameter _is_cv_model will also be true. Hence, I need to manually set them.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java
wendycwong: Agreed

h2o-genmodel/src/main/java/hex/genmodel/algos/gam/GamMojoModelBase.java
wendycwong: I would like to keep this here because it is a bug and it makes me sad to ignore it.

h2o-algos/src/test/java/hex/gam/GamCVTest.java
wendycwong: I care about your knitpicking.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java
Pscheidl: Did we address this @wendycwong ? The comment still says "If 0, insert rows of NaN" - but that's not what is happening in the `else` block. See comment above.

wendycwong: @pavel:  I am such a dumb ass.  I fixed it in my branch without pushing it.  I just pushed and it looks correct now.  Thanks for catching this.

Pscheidl: Oh my god, no you're not.

Thank you @wendycwong .

Pscheidl (approved):

h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -31,7 +35,15 @@
 
 
 public class GAM extends ModelBuilder<GAMModel, GAMModel.GAMParameters, GAMModel.GAMModelOutput> {
-
+  double[][] _knots;
+  boolean _cv_on = false; // set to true if cross-validation is enabled
```
maurever, line 39: Do we have any other internal rule for using an underscore in Java code than at the beginning of a parameter? Because the standard is to use CamelCase. It is not clear to have two different styles in our Java code. 

```suggestion
  boolean _cvOn = false; // set to true if cross-validation is enabled
```

h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -70,15 +82,46 @@ public GAM(GAMModel.GAMParameters parms, Key<GAMModel> key) {
...
+    _cv_on = true;  // set cross-validation on to be true
+    _validKeys = new IcedHashSet<>();
+    if (error_count() > 0)
+      throw H2OModelBuilderIllegalArgumentException.makeFromBuilder(GAM.this);
+    super.computeCrossValidation();
+  }
+
+  // find the best alpha/lambda values used to build the main model moving forward by looking at the devianceValid
+  @Override
+  public void cv_computeAndSetOptimalParameters(ModelBuilder[] cvModelBuilders) {
```
maurever, line 99: Same problem with CamelCase here. It is hard to read underscore in a Java method. :( But if there is a rule use cross-validation methods like this, keep it as is.

maurever (approved):
@wendycwong, thank you for this PR. It looks good to me. I have just a note about using CamelCase style as much as possible to keep our Java code clear and readable.

h2o-algos/src/main/java/hex/gam/GAM.java
wendycwong: I think _cvOn is a good idea.  I changed it.

h2o-algos/src/main/java/hex/gam/GAM.java
wendycwong: Yeah, I am overriding an existing method.  I agree with you about the name issues though.

Pscheidl:
I'll be finishing this PR as Wendy is on vacation.

Pscheidl:
GamMojoModelTest fixed. One failure was resolved by rebasing (the NaN problem that's already fixed on master). I'll go through the code one more time tomorrow and then merge or ask for further reviews eventually.

Pscheidl:
Added two more tests for multinomial and regression, did some minor code polishing (was not really required) and did one micro-optimization - validation keys were copied to the model for no reason. Just moved the pointer there :wink: .

Waiting for tests. Feel free to review @michalkurka  and others. No big changes since the recent reviews.

h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -70,15 +81,47 @@ public GAM(GAMModel.GAMParameters parms, Key<GAMModel> key) {
...
+    double best_alpha = 0;
+    double best_lambda = 0;
+    for (int i = 0; i < cvModelBuilders.length; ++i) {  // run cv for each lambda value
+      GAMModel g = (GAMModel) cvModelBuilders[i].dest().get();
+      if (g._output._devianceValid < deviance_valid) {
+        best_alpha= g._output._best_alpha;
+        best_lambda = g._output._best_lambda;
+      }
+      
+      g.write_lock(_job);
```
michalkurka, line 110: this looks like a copy&paste from GLM; it doesn't seem needed here because the GAM models are not modified here

h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -70,15 +81,47 @@ public GAM(GAMModel.GAMParameters parms, Key<GAMModel> key) {
...
+      g.update(_job);
+      g.unlock(_job);
+    }
+    _cv_alpha = new double[]{best_alpha};
+    _cv_lambda = new double[]{best_lambda};
+    _doInit = false;  // for main model, disable the init()
```
michalkurka, line 116: this sounds suspicious and/or redundant - the main model already knows it is the main model

h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -70,15 +81,47 @@ public GAM(GAMModel.GAMParameters parms, Key<GAMModel> key) {
...
+    _cvOn = false;   // building main model now.  Set CV off.
```
michalkurka, line 117: also suspicious

h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -384,10 +439,11 @@ void verifyGamTransformedFrame(Frame gamTransformed) {
     
     @Override
     public void computeImpl() {
-      init(true);     //this can change the seed if it was set to -1
+      if (_doInit)                  // disable when in CV and building the main model
+        init(true);       //this can change the seed if it was set to -1
+      validateGamParameters();
```
michalkurka, line 444: I wonder why do we need this, however, if we keep `_doInit` then we should do

if (_doInit) {
 init(true)
} else {
 validateGamParameters();
}

however, skipping init is suspicious

Pscheidl: Thank you for catching this. I though this is there because of the way GAM is instantiated, but it seems to be false theory. I'll look into it.

Pscheidl: The only things that seems to be really effectively required from init is the `validateGamParameters` anyway.

Pscheidl:
@michalkurka 

I reviewed this part of code in-depth and gave my best to optimize the complexity of the logic. The result is saved in the https://github.com/h2oai/h2o-3/pull/4869/commits/c40a93bb070fb9c4d55c95858ec80d13e956f3f8 commit.

- Init is always done, as `validateGamParameters` is required. It doesn't only do "validation", but also sets knots. Therefore, it has to be done together with whatever logic is in init(). As mentioned, the main model should already know it is the main model.
- Removed cv-on boolean switch. Can't find any real use for that. The logic it was used in was redundant anyway (see the code removed in `buildModel` method).
- Added a couple of additional asserts to JUnits to specifically test presence of basic xval output structures in model output.
-` _keep_gam_cols` is now enabled in one test to try out removal of such keys and rule out key leaks - couldn't find this being tested before.

Please give this another go, thank you.

Pscheidl:
I most likely found the issue - the `buildGamFrame` method modified the `_train` directly.

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java
```diff
@@ -33,34 +34,51 @@ public GenerateGamMatrixOneColumn(int splineType, int numKnots, double[] knots,
...
-    _gamX = gamx;
+    _maxAbsRowSum = new double[gamx.vec(0).nChunks()];
     _knots = knots;
   }
 
   @Override
   public void map(Chunk[] chk, NewChunk[] newGamCols) {
-    _maxAbsRowSum = new double[_gamX.vec(0).nChunks()];
+    for (int i = 0; i < chk.length; i++) {
+      if (chk[i].len() != chk[0].len()) throw new IllegalArgumentException("Chunks are not the same len");
```
Pscheidl, line 44: Ok, this check actually fails. Shouldn't the chunk layout be all the same in a frame @michalkurka @honzasterba and possibly others ? What am I missing ?

h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java
```diff
@@ -48,19 +48,29 @@ public void map(Chunk[] chk, NewChunk[] newGamCols) {
     for (int rowIndex = 0; rowIndex < chunkRows; rowIndex++) {
```
Pscheidl, line 48: The chunks might have different length !

honzasterba: do you mean `chk[i].len() != chk[j].len()` for some i,j? that should not be possible

Pscheidl: Exactly.

```
    for (int i = 0; i < chk.length; i++) {
      if(chk[0].len() != chk[i].len()){
        throw new IllegalArgumentException(String.format("Chunks have different length. %s vs %s", chk[0].len(), chk[i].len()));
      }
    }
```

is actually triggered (even locally).

Pscheidl: Here is a full patch to easily apply:

```patch
Index: h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java
IDEA additional info:
<+>UTF-8
--- h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java	(revision c40a93bb070fb9c4d55c95858ec80d13e956f3f8)
+++ h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java	(date 1600675633440)
@@ -39,6 +39,11 @@
 
   @Override
   public void map(Chunk[] chk, NewChunk[] newGamCols) {
+    for (int i = 0; i < chk.length; i++) {
+      if(chk[0].len() != chk[i].len()){
+        throw new IllegalArgumentException(String.format("Chunks have different length. %s vs %s", chk[0].len(), chk[i].len()));
+      }
+    }
     _maxAbsRowSum = new double[_gamX.vec(0).nChunks()];
     int cIndex = chk[0].cidx();
     _maxAbsRowSum[cIndex] = Double.NEGATIVE_INFINITY;
```

Pscheidl: Next step is obvious, look at the input and see how the chunks are created and whether the layout is the same. At first sight, it should NOT be possible as those are added to a new Frame and checks are NOT disabled, but that `Frame` thing has like a million of constructors, so maybe I overlooked it.

Let the digging continue.

honzasterba: I dont think its related to anyhing but I find it interesting that every map() call overwrites the array `_maxAbsRowSum = new double[_gamX.vec(0).nChunks()];`

Pscheidl: That is just a bad design I believe. But thank you, every clue, no matter how small is important. I investigated this one as well. The Frame itself can be omitted from serialization and this value can just be initialized properly where it should be (inside the constructor). This has no effect on the error (see diff below to quickly apply if you feel like it)

```patch
Index: h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java
IDEA additional info:
<+>UTF-8
--- h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java	(revision c40a93bb070fb9c4d55c95858ec80d13e956f3f8)
+++ h2o-algos/src/main/java/hex/gam/MatrixFrameUtils/GenerateGamMatrixOneColumn.java	(date 1600676523134)
@@ -19,7 +19,6 @@
   int _splineType;
   public int _numKnots;      // number of knots
   public double[][] _bInvD;  // store inv(B)*D
-  Frame _gamX;
   double[] _u; // store transpose(X)*1, sum across rows per column
   public double[][] _ZTransp;  // store Z matrix transpose
   public double[][] _penaltyMat;  // store penalty matrix
@@ -33,13 +32,12 @@
      CubicRegressionSplines crSplines = new CubicRegressionSplines(numKnots, knots);
     _bInvD = crSplines.gen_BIndvD(crSplines._hj);
     _penaltyMat = crSplines.gen_penalty_matrix(crSplines._hj, _bInvD);
-    _gamX = gamx;
+    _maxAbsRowSum = new double[gamx.vec(0).nChunks()];
     _knots = knots;
   }
 
   @Override
   public void map(Chunk[] chk, NewChunk[] newGamCols) {
-    _maxAbsRowSum = new double[_gamX.vec(0).nChunks()];
     int cIndex = chk[0].cidx();
     _maxAbsRowSum[cIndex] = Double.NEGATIVE_INFINITY;
     int chunkRows = chk[0].len(); // number of rows in chunk

```

Pscheidl: Both vectors in the `predictVec` Frame (that naming !) report the same length and the same `espc`:

```java
            GenerateGamMatrixOneColumn genOneGamCol = new GenerateGamMatrixOneColumn(splineType, numKnots, 
                    _knots[frameIndex], predictVec).doAll(numKnots, Vec.T_NUM, predictVec);
```

On this level, everything seems to be just fine.

Pscheidl: Also, one more observation. Notice the whole MRTask is done in parallel in a batch of `RecursiveAction` - not sure why we need recursive action for that. Yet I've tried to make this sequential to quickly rule out a problem with concurrent modification on that side, to no avail. Problem still remains.

Pscheidl: Oh wow,

```java
        final Frame predictVec = new Frame();
        predictVec.add(_parms._gam_columns[index], _parms._train.get().vec(_parms._gam_columns[index]));
        predictVec.add("weights_column", weights_column);
```

This seems to produce even chunks and proper layout. Wow. Let's try that.

Pscheidl: Hooray, Fixed ! http://mr-0xc1:8080/blue/organizations/jenkins/h2o-3-pipeline/detail/PR-4869/44/tests/

honzasterba (approved):

michalkurka (approved):
Thank you @Pscheidl

Pscheidl:
There are 4 other key management issues, solved 3 of them, working on the last one. Those were shadowed by the former problems.

Pscheidl:
Resolved. One validation frame lifecycle was to blame - it was not properly deleted, as the IcedHashSet was not initialized if validation frame was declared by the user. It was only initialized when cross validation was done. This is no longer true.

Pscheidl:
This one dataset is reused and it's deleted in `remove_impl` of `GamModel`. That's the original design. I just made sure it's deleted in all two cases (user-provided validation data) and not just xval.

Pscheidl:
Oh, now I see what you're asking about. Let me fix that really quickly.

PR #4873, opened 2020-08-26
Fixing broken R and Python examples in perf User Guide page

ledell:
Hey @hannah-tillman I found some broken examples in the [Performance and Prediction](http://docs.h2o.ai/h2o/latest-stable/h2o-docs/performance-and-prediction.html) user guide page.  I also changed the dataset name to be "pros" everywhere instead of "pros" and "prostate".  Can you review this and run the examples in R and Python to make sure it's all working properly?  Thank you!


ledell:
Thanks for catching those, @hannah-tillman!

PR #4874, opened 2020-08-26
Improved reporting of leaked keys in H2ORunner

michalkurka:


Pscheidl:
This is great @michalkurka ! :clap: 

Your way of storing leakage info is **superior** to my older one. More information, properly encapsulated. Therefore, I allowed myself to take your idea and make it the new default for `H2ORunner`. It's all in https://github.com/h2oai/h2o-3/pull/4874/commits/1e33f6958e4b043d9865014a5b139ec7b3a1d392 commit. If you don't like it, just remove that one commit and you're done. I felt if you'd like it, having the changes commited would **save some of your time**.

## Changes summary

1. Every key leaked is now represented with `LeakInfo`. This means I made LeakInfo generic and not just suit `RollupStats`. Minor change.
1. Keys are still grouped together by Frame -> Vec -> Chunk -> RollupStats, if possible. Leak info is of course used here.
1. Uncategorized keys that don't belong to any frame are still printed after the grouped keys are printed. `LeakInfo` is used here as well.
1. Introduced `@ExpectedLeaks` annotations to make `H2ORunner` more testable. Number inserted into that annotation represents a precise number of leaks expected. If the number is the same, the test passes. 
1. As we have `@ExpectedLeaks` annotation, we're able to actually run such tests on daily basis, even in Jenkins. No need for exceptions. Therefore, I changed the name of `H2ORunnerPlayground` to `H2ORunnerTest`, as it's now able to run regularly. Leaked keys after the test should **not be a problem**, as `CleanAllKeysTask` clears all leaked keys after each class in `H2ORunnerAfters` class. Therefore, any memory leaks created in `H2ORunnerTest` will **not** affect other tests.

Pscheidl (approved):

Pscheidl (approved):

PR #4877, opened 2020-08-27
PUBDEV-7738: Fix broken URLs in R docs causing CRAN failures

ledell:
https://0xdata.atlassian.net/browse/PUBDEV-7738


ledell:
There was an extra NOTE on the CRAN check, causing it to fail, but I looked into it and it's caused by a time-check API that has gone down, so it's unrelated.

```
[2020-08-28T07:29:09.369Z] * checking for future file timestamps ... NOTE
[2020-08-28T07:29:09.369Z] unable to verify current time
```

PR #4883, opened 2020-08-31
PUBDEV-7743: new TargetEncoder param allowing to remove original features automatically

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-7743

By default, the original features are kept: it's not a nice default imo as encoders usually replace the column to encode, but it keeps the API backwards compatible...

Param is currently named `keep_original_features`.

### Please note
The legacy model had another inconsistency:
- trained model would not remove original features by default.
- Mojo model would only return encoded columns.

Therefore it is impossible to keep both backwards compatibility and consistency between trained model and Mojo model.
Chosing for now to leave Mojo as is, meaning ignoring the new parameter, which somehow makes sense for production.

### DIFF
--- a/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
+++ b/h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
@@ -43,6 +43,7 @@ public class TargetEncoderModel extends Model<TargetEncoderModel, TargetEncoderM
     public double _smoothing =DEFAULT_BLENDING_PARAMS.getSmoothing();
     public DataLeakageHandlingStrategy _data_leakage_handling = DataLeakageHandlingStrategy.None;
     public double _noise = 0.01;
+    public boolean _keep_original_features = true; // not a good default, but backwards compatible.
     
     @Override
     public String algoName() {
@@ -333,6 +334,9 @@ public class TargetEncoderModel extends Model<TargetEncoderModel, TargetEncoderM
             tmpKey = workingFrame._key;
           }
         } // end for each target 
+        
+        if (!_parms._keep_original_features)
+          workingFrame.remove(colIdx);
       } // end for each columnToEncode
 
       DKV.remove(tmpKey);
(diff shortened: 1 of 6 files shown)


Pscheidl:
FYI I'd personally like to see https://github.com/h2oai/h2o-3/pull/4867 merged first before merging other changes. We could create a dependency hell among PRs easily. This is only about the merge order, there is nothing preventing me to review this.

sebhrusen:
@Pscheidl no worry, the plan is to merge them all in the order they were created.
As soon as the refactoring, then multiclass PRs are not approved and merged, this one won't go in.
But I still need to work in between ;)

ledell:
@sebhrusen This seems useful, thanks for adding!  I am wondering -- would it make sense to have an argument that let's you list the precise columns that you want to keep the original features for, rather than a single on/off switch?  
e.g. `keep_encoded_categorical_columns = "NONE"` or `"ALL"` or `["column1", "column2"]`

sebhrusen:
@ledell are you also suggesting to change the property name? `keep_original_categorical_columns` instead of `keep_original_features`?
I think your proposal is actually better, because only categorical columns that were encoded get to be removed.

Regarding the suggestion: "NONE" or "ALL" or ["column1", "column2"]
I'm not sure I like it:
- on backend it would have to be represented as an array, so it there wouldn't be any naming conflict, we would just have the following on client side:
  - "NONE" or false -> []
  - "ALL" or true -> [list all columns from training frame]  # will have to be filled dynamically when calling `train` in Python client, requires a hook.
- true/false make it just a convenience, covers the great majority of use-cases, and it's simple.
- allowing the user to select which column should be removed or not is useful for a small minority of use-cases, and for those, user can still remove the columns on client side.

I'm not against it, just don't think it's must have currently, will rather extend the semantic later in a separate PR if needed.

sebhrusen:
Also, I need to add the new property to the doc.

ledell (approved):
@sebhrusen Ok, I am on board with keeping the arg a binary on/off switch.  Can we change the param name to `keep_original_categorical_columns` ?

PR #4884, opened 2020-08-31
Fix outside dataset R

wendycwong:
Same problem as in python for runit_PUBDEV_7588_GLM_lambda0_interactions.R


wendycwong:
Redundant

PR #4886, opened 2020-08-31
PUBDEV-7730: generate precision recall plots

wendycwong:
This PR competes this JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7730 for python clients.  Will need to add the same thing for R clients.

I have derived precision/recall from h2o-metrics. 
Next, using matplotlib, pr plots are made for training/validation/cross-validation.

Please check the API for pr_plot for me as well.


wendycwong:
Here is an example of the plot:
image

neemah2o (approved):
LGTM, but is there a reason a new function was created (`pr_plot()`)? Couldn't you use `plot()`? Like for AUC-ROC: `.plot(type = "roc")`

maurever:
The plot by @wendy looks good. However, I think,  we should keep the same code as the current AUC curve plot implementation. So I used the Wendy's code and implemented the Precision-Recall curve plot the same way as the AUC curve plot. 

The usage now is like @neemah2o mentioned:

```
perf = model.performance()
perf.plot(type="roc")
perf.plot(type="pr")
```

Here are the results: 
auc_test
pr_test

The disadvantage is, the user cannot print train, valid, and xval curves in one plot. But he/she cannot do it with ROC as well.

WDYT @michalkurka , @neemah2o , @wendycwong ?

neemah2o (approved):

maurever:
@ledell and @michalkurka WDYT? :)

maurever:
I added the PR curve plot to R too. The result plot is:

r_pr

With the same style as ROC curve:

r_roc

The design could be better, but it could be done in different JIRA.

maurever:
I unified the plots a little bit. Results:

**Python:** 
Python_roc_new
Python_pr_new

**R:**
R_roc_new
R_pr_new

WDYT @michalkurka and @ledell ?

michalkurka (approved):

ledell (approved):
This looks good!  Yes, good change to make it an extension of the existing ROC plot.  I like the ability to add multiple plots (train, valid, CV) on a single plot, but consistency with existing plotting is better.  We can consider adding some more plotting options later with a custom function.

PR #4888, opened 2020-09-01
PUBDEV-7319: TE integration to ModelBuilder

sebhrusen:

### Java API (and only Java)

- For seamless integration, introduced a `ModelPreprocessor` abstraction. Making clear that this abstraction is probably only temporary and should not be implemented or used by external users.
- For CV support, added the possibility for TE to encode a frame (usually the out-of-fold training frame) using the encodings generated for a specific fold, thanks to the new `transformTraining(Frame fr, int fold)` method.

### Logical changes
- Introduced strict columns ordering of columns after applying TE: this is necessary as the same ordering needs to be replicated on Mojo side when TE Mojo is integrated to the main model Mojo.

### Mojo integration
- as TE is defined during training as a model `preprocessor`, a parallel is implemented on Mojo side:
  - model preprocessors are bundled/zipped inside the main model as a standalone submodel.
  - when loading the bundled mojo, the TE submodel is loaded as a `MojoPreprocessor`.
  - when building the `RowToRawDataConverter`, we loop through the mojo preprocessors and the final categorical encoding, each providing a partial `RowToRawDataConverter` (for TE, see `TargetEncoderAsModelProcessor.TargetEncoderRowToRawDataConverter` to produce a `CompositeRowToRawDataConverter` that will be used for predictions.

### Notes
- TE could not be implemented as a simple `CategoricalEncoding` for the mojo integration. This because TE doesn't systematically encode all categorical columns, so in many cases, the "default" `CategoricalEncoding` still needs to be applied after TE.
- The current interface defining `CategoricalEncoding` doesn't allow to create sth like a `CompositeCategoricalEncoding`: this is because the 2 methods `createColumnMapping` and `createCategoricalEncoders` are strongly coupled, and don't offer space for the ad-hoc logic necessary to apply TE correctly. Hence the approach through the `RowToRawDataConverter` abstraction.
- I strongly suspect serious bugs with legacy Mojos when the model was trained with a fold/weights/offset column and with a non-default categorical encoding like binary or OHE (actually it just fails...): this is because the training phase ensures that with those encoding, the non-predictors columns listed above are moved to the "end", whereas on Mojo side, they are considered as normal numerical columns, leading to column mismatch. I solved this in the particular case for TE, but I think the same needs to be done for those categorical encodings requiring a strict column ordering too.
- You can ignore the Py tests at the end: I had to regenerate the golden TE transformations after enforcing the column ordering.


h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
```diff
@@ -6,6 +6,7 @@
 import static hex.genmodel.algos.tree.SharedTreeMojoModel.__INTERNAL_MAX_TREE_DEPTH;
 
 import hex.genmodel.CategoricalEncoding;
+import hex.genmodel.CategoricalEncodings;
```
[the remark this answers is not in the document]
michalkurka: imho that is fine, no issue that I can foresee - these are not really public classes

h2o-core/src/main/java/hex/Model.java
```diff
@@ -396,6 +400,12 @@ public long getOrMakeRealSeed(){
     /** @return the validation frame instance, or null
      *  if a validation frame was not specified */
     public final Frame valid() { return _valid==null ? null : _valid.get(); }
+    
+    public String[] getNonPredictors() {
+        return Arrays.stream(new String[]{_weights_column, _offset_column, _fold_column, _response_column})
+                .filter(Objects::nonNull)
+                .toArray(String[]::new);
+    }
```
sebhrusen, line 408: this list is used in several places: better to have a single access point

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoder.java
```diff
@@ -64,9 +64,9 @@ public void init(boolean expensive) {
     }
   }
 
-  private void disableIgnoreConstColsFeature() {
+  private void disableIgnoreConstColsFeature(boolean expensive) {
     _parms._ignore_const_cols = false;
-    if (logger.isInfoEnabled())
+    if (expensive && logger.isInfoEnabled())
```
sebhrusen, line 69: non-related: disabled the warning at startup when building the builder prototypes

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
```diff
@@ -147,38 +150,47 @@ public TargetEncoderModel(Key<TargetEncoderModel> selfKey, TargetEncoderParamete
   }
 
   public Frame transformTraining(Frame fr) {
-    return transform(fr, _parms.getBlendingParameters(), _parms._noise, true);
+    return transformTraining(fr, NO_FOLD);
+  }
+  
+  public Frame transformTraining(Frame fr, int outOfFold) {
+    assert outOfFold == NO_FOLD || _parms._data_leakage_handling == DataLeakageHandlingStrategy.KFold;
+    return transform(fr, true, outOfFold, _parms.getBlendingParameters(), _parms._noise);
```
sebhrusen, line 158: this method is used when applying TE during CV.
This ensures that there's no leakage from out of fold data into the cv training data.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
```diff
@@ -463,30 +513,45 @@ protected void removeNumeratorAndDenominatorColumns(Frame fr) {
...
+        Scope.track(tmpFoldCol);
+        workingFrame.add(new String[] {_foldColumn+TMP_COLUMN_POSTFIX}, new Vec[]{tmpFoldCol});
+        foldColIdx = workingFrame.numCols()-1;
+      }
       int encodingsFoldColIdx = encodings.find(_foldColumn);
       int encodingsTEColIdx = encodings.find(columnToEncode);
       long[] foldValues = getUniqueColumnValues(encodings, encodingsFoldColIdx);
       int maxFoldValue = (int) ArrayUtils.maxValue(foldValues);
-      double priorMean = computePriorMean(encodings);
+      double priorMean = computePriorMean(encodings); //FIXME: we want prior for the outOfFold encodings
```
sebhrusen, line 543: not fixing this now because the whole TE implementation is currently using a global prior with KFold strategy.
It probably creates a small data leakage, but as it is a legacy issue, will fix in a future improvement.

h2o-extensions/target-encoder/src/test/java/ai/h2o/targetencoding/TargetEncoderPreprocessorTest.java
```diff
@@ -0,0 +1,297 @@
...
+            assertEquals(predictions.vec(1).at(0), mojoPredictions.classProbabilities[0], 1e-8);
+            assertEquals(predictions.vec(2).at(0), mojoPredictions.classProbabilities[1], 1e-8);
+        } finally {
+            Scope.exit();
+        }
+    }
+    
+    
+    @Test
+    public void test_model_mojo_predictions_consistency_with_TE_encoding_all_categoricals() {
```
sebhrusen, line 166: adding those now

h2o-genmodel-extensions/deepwater/src/main/java/hex/genmodel/algos/deepwater/DWImageConverter.java
```diff
@@ -2,8 +2,8 @@
 
 import hex.genmodel.GenModel;
 import hex.genmodel.easy.CategoricalEncoder;
+import hex.genmodel.easy.DefaultRowToRawDataConverter;
```
sebhrusen, line 5: the changes in non-TE-related genmodel classes are due to renamings:
- I mentioned `CategoricalEncoding` enum -> `CategoricalEncodings`.
- here, `RowToRawDataConverter` has been renamed to `DefaultRowToRawDataConverter` and the extracted interface keeps the oiginal name.

h2o-genmodel/src/main/java/hex/genmodel/AbstractMojoWriter.java
```diff
@@ -45,7 +45,7 @@ public AbstractMojoWriter(ModelDescriptor model) {
   /**
    * Override in subclasses to write the actual model data.
    */
-  protected abstract void writeModelData() throws IOException;
+  protected abstract void writeModelKV() throws IOException;
```
sebhrusen, line 48: renamed this as I needed to keep `writeModelData` for specific model implementations.
In `ModelMojoWriter`, `writeModelKV` is responsible to write both preprocessor related data + model specific data (= `writeModelData`).

h2o-genmodel/src/main/java/hex/genmodel/CategoricalEncodingAsModelProcessor.java
```diff
@@ -0,0 +1,33 @@
+package hex.genmodel;
+
+import hex.genmodel.easy.CategoricalEncoder;
+import hex.genmodel.easy.EasyPredictModelWrapper;
+import hex.genmodel.easy.RowToRawDataConverter;
+
+import java.util.Map;
+
+public class CategoricalEncodingAsModelProcessor implements MojoPreprocessor.ModelProcessor {
```
sebhrusen, line 9: allows to turn `CategoricalEncoding` into a preprocessor to create a single `CompositeRowToRawDataConverter` at the end.

h2o-genmodel/src/main/java/hex/genmodel/CategoricalEncodings.java
```diff
@@ -0,0 +1,74 @@
+package hex.genmodel;
```
sebhrusen, line 1: nothing new except the renaming + the `implements`.

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
```diff
@@ -196,7 +196,7 @@ private void readAll(final boolean readModelMetadata) throws IOException {
     _model._offsetColumn = readkv("offset_column");
     _model._mojo_version = ((Number) readkv("mojo_version")).doubleValue();
     checkMaxSupportedMojoVersion();
-    readModelData();
+    readModelKV();
```
sebhrusen, line 199: logic here is symmetric with changes in `ModelMojoWriter`

h2o-genmodel/src/main/java/hex/genmodel/MojoModel.java
```diff
@@ -71,5 +76,39 @@ public static MojoModel load(MojoReaderBackend mojoReader) throws IOException {
...
+    Map<Integer, CategoricalEncoder> offsetToEncoder = categoricalEncoding.createCategoricalEncoders(this, columnToOffsetIdx);
+    return makeDefaultRowConverter(columnToOffsetIdx, offsetToEncoder, errorConsumer, config);
+  }
   
+  protected RowToRawDataConverter makeDefaultRowConverter(Map<String, Integer> columnToOffsetIdx,
+                                                          Map<Integer, CategoricalEncoder> offsetToEncoder,
+                                                          ErrorConsumer errorConsumer,
+                                                          Config config) {
+    return new DefaultRowToRawDataConverter<>(columnToOffsetIdx, offsetToEncoder, errorConsumer, config);
+  }
```
sebhrusen, line 113: subclasses that need a specific converter now override this method.

h2o-genmodel/src/main/java/hex/genmodel/MultiModelMojoReader.java
```diff
@@ -34,28 +33,4 @@ protected MojoModel getModel(String key) {
 
   protected abstract void readParentModelData() throws IOException;
 
-  private class NestedMojoReaderBackend implements MojoReaderBackend {
```
sebhrusen, line 37: this class extraction is not needed anymore, will revert

sebhrusen, line 37: extracted this reader as I'm reusing it to read the bundled TE model.

h2o-genmodel/src/main/java/hex/genmodel/algos/targetencoder/TargetEncoderAsModelProcessor.java
```diff
@@ -0,0 +1,237 @@
...
+        public double[] convert(RowData data, double[] rawData) throws PredictException {
+            double[] converted = super.convert(data, rawData);
+            if (!_teModel._keepOriginalCategoricalColumns) {
+                for (String teColumn : _teModel._encodingsByCol.keySet()) data.remove(teColumn);
+            }
+            return converted;
+        }
+    }
+
+    public static class VirtualTargetEncodedModel extends GenModel {
```
sebhrusen, line 192: will add some doc and change visibility: this virtual model is used to "trick" further processors and especially `Categoricalencoding` to provide him with a model/frame description that looks as if TE was already applied.

h2o-genmodel/src/main/java/hex/genmodel/easy/DefaultRowToRawDataConverter.java
```diff
@@ -0,0 +1,111 @@
+package hex.genmodel.easy;
```
sebhrusen, line 1: no major changes here, just renamed from `RowToRawDataConverter`

Pscheidl (changes requested):
This should have been discussed before implementation. There is a slight change we want to do things in other way. And the way has a name: pipelines. Used "everywhere" else. Making models dependent on preprocessor, (postprocessors are next to come ?) and other processors makes it too constrained by these transformations. The upside of having a generic concept of a pipeline represented by our data structures makes it easy to make the encapsulating algorithm as generic as possible. The impact of MOJO is simple: a MOJO just embeds and represents the pipeline.

Such a pipeline is configurable, certains steps may be skipped or even replaced on demand. It's nicely replayable as well. It may be checkpointed - resuming/replaying only from certain step. It's also configurable. A ML model is therefore only one part of the whole recipe.

This approach seems adequate in it's genericity and future proofness. This has been discussed by myself several times quite some time ago. If the urge is to have this API, then I'm not completely against it, yet I'd like to have a plan of what happens to models with preprocessors after (and if) pipelines are introduced. Would we be able to deprecate them grafully while keeping backwards compatibility ? How would this change impact Generic model ? Currently, MOJO only knows about a preprocessor, not about it's details. Which details will we be able to show to our users once the MOJO is imported ? 

Pipelines have on inherent advantage - order of operations/transformations is known. In this PR, it is assumed order of preprocessors, which may be perceived as transformations placed before the model inside a pipeline do not have an order. Is this always the case ? I'm not sure at all. Would we be able to add this in future ?

sebhrusen:
@Pscheidl thanks for this high-level feedback.
I think everyone here agrees that pipelines are a much better approach: this has also been discussed with @michalkurka, me and Andrey long ago and we're all aware of its long term benefits. However we now need a solution to allow the bundling of TE into model Mojo so that when we enable TE in AutoML, user can use the generated Mojo.

That's why the ModelPreprocessor interface is documented as temporary and there's NO plan to expose those preprocessors to the clients for the reason that we want to be able to replace them with pipelines once someone comes with a REAL implementation plan.
Many of us suggested this concept/feature in the past but no progress has been done so far, so we can't wait for it to be able to integrate TE today (this has also been discussed with @michalkurka and Andrey), or it means that we knew from the beginning that TE in AutoML could not be delivered without pipelines and that core team should have dedicated some effort to it long ago.

Let's look at the current implementation as a temporary solution and ensure that it doesn't prevent us from replacing it at any time in the future:
- it is used only internally: client API doesn't rely on preprocessors.
- it is not invasive: it is abstracted enough so that we can clearly see the impact of preprocessors in the codebase.
- same concerns apply on the Mojo integration.

cc: @michalkurka

Pscheidl:
Pipelines will NOT be done in this PR !

h2o-core/src/main/java/hex/Model.java
```diff
@@ -268,6 +267,9 @@ public void setTrain(Key<Frame> train) {
       final boolean _needResponse;
       boolean needsResponse() { return _needResponse; }
     }
+    
+    public Key<ModelPreprocessor>[] _preprocessors;
```
Pscheidl, line 271: Do we want the preprocessors to have an explicit order ? Or do we want to rely on serialization/deserialization order and iterations ?

Pscheidl (commented):
Generic model has to be adjusted as well to show info about preprocessors used. Doesn't need to be done in this PR. Just a note.

Pscheidl (approved):
Saw some minor formatting issues (mainly the first lines of some for cycles etc). Otherwise the content LGTM.

h2o-core/src/main/java/hex/Model.java
sebhrusen: @Pscheidl if we apply multiple TE models, for example te1 apply to some col1 with params1, te2 applies to col2 with params2, and te3 applies to grouped columns te3+te4... then these TE applications are not fully commutative as they generate new columns, and we must guarantee that the order of those columns is the same during training and in the mojo.
The easiest way to guarantee this order currently is to apply the various TE models in the same order in the Mojo.

So yeah, explicit order is important.

sebhrusen: Note that the serialization order that you mention is the same as the order of preprocessors applied during training.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1625,9 +1644,30 @@ public Frame score(Frame fr, String destination_key, Job j, boolean computeMetri
...
-    Frame.deleteTempFrameAndItsNonSharedVecs(adaptFr, fr);
+    for (Frame tmp : tmpFrames) Frame.deleteTempFrameAndItsNonSharedVecs(tmp, fr);
     return output;
   }
+  
+  private void applyPreprocessors(Frame fr, List<Frame> tmpFrames) {
+    if (_parms._preprocessors == null) return;
+    
+    for (Key<ModelPreprocessor> key : _parms._preprocessors) {
+      DKV.prefetch(key);
```
honzasterba, line 1655: what is the benefit of this call?

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1583,6 +1596,12 @@ public Frame score(Frame fr, String destination_key, Job j, boolean computeMetri
   }
   public Frame score(Frame fr, String destination_key, Job j, boolean computeMetrics, CFuncRef customMetricFunc) throws IllegalArgumentException {
     Frame adaptFr = new Frame(fr);
+    List<Frame> tmpFrames = new ArrayList<>();
+    tmpFrames.add(adaptFr);
+    applyPreprocessors(adaptFr, tmpFrames);
+    //PUBDEV-7775: we need to apply encoding independently from adaptTestForTrain, otherwise previously encoded columns are removed during adaptation
+    //if enabling this, call to adaptTesttoTrain below should use catEncoded=true, however it breaks several tests currently: necessary only for TE in combination with encoders like OHE
+//    encodeCategoricals(adaptFr, tmpFrames);
```
honzasterba, line 1604: commented out code
formatting
generally comments like this have little value in the long term, if I understand this correctly `applyPreprocessors(adaptFr, tmpFrames);` now does `encodeCategoricals`, if thats the case I would be fine with
`applyPreprocessors(adaptFr, tmpFrames);  // also does categorical encoding`

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1625,9 +1644,30 @@ public Frame score(Frame fr, String destination_key, Job j, boolean computeMetri
...
+    Frame result = fr;
+    for (Key<ModelPreprocessor> key : _parms._preprocessors) {
+      ModelPreprocessor preprocessor = key.get();
+      result = preprocessor.processScoring(result, this);
+      tmpFrames.add(result);
+    }
+    fr.restructure(result.names(), result.vecs()); //inplace
+  }
+  
+  private void encodeCategoricals(Frame fr, List<Frame> tmpFrames) {
```
honzasterba, line 1666: seems unused

h2o-core/src/main/java/hex/ModelMojoWriter.java
```diff
@@ -80,12 +82,39 @@ public ModelMojoWriter(M model) {
...
+    super.writeExtraInfo();
     writeModelDetails();
     writeModelDetailsReadme();
+    writeModelPreprocessors();
+  }
+  
+  protected void writeModelPreprocessors() throws IOException {
+    if (model._parms._preprocessors == null) return;
+    for (int i=0; i < model._parms._preprocessors.length; i++) {
+      startWritingTextFile("experimental/preprocessor_"+i);
```
honzasterba, line 110: no use using experimental dir name, this will aways forever stay the same, can be just `preprocessors/preprocessor_i`

h2o-core/src/main/java/water/DKV.java
```diff
@@ -1,12 +1,5 @@
 package water;
 
-import hex.Model;
```
honzasterba, line 3: 👍

h2o-genmodel/src/main/java/hex/genmodel/CategoricalEncoding.java
```diff
@@ -4,76 +4,12 @@
...
-    public Map<Integer, CategoricalEncoder> createCategoricalEncoders(GenModel m, Map<String, Integer> columnMapping) {
-      return new EigenEncoderDomainMapConstructor(m, columnMapping).create();
-    }
-  },
-  LabelEncoder {
-    @Override
-    public Map<String, Integer> createColumnMapping(GenModel m) {
-      return new EnumEncoderColumnMapper(m).create();
-    }
+public interface CategoricalEncoding {
```
honzasterba, line 7: so this refactor to interface and move of the enums is entirely unnecessary for now? if yes I would lean towards reverting this part

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
```diff
@@ -223,6 +223,23 @@ protected ModelAttributes readModelSpecificAttributes() {
     }
   }
 
+  protected void readModelKV() throws IOException {
+    readCommonModelData();
+    readModelData();
+  }
+
+  protected void readCommonModelData() throws IOException {
```
honzasterba, line 231: do we need to override this somewhere? should the call just be moved to readModelKV?

h2o-genmodel/src/main/java/hex/genmodel/ModelMojoReader.java
```diff
@@ -223,6 +223,23 @@ protected ModelAttributes readModelSpecificAttributes() {
     }
   }
 
+  protected void readModelKV() throws IOException {
```
honzasterba, line 226: should be final?

h2o-genmodel/src/main/java/hex/genmodel/MojoModel.java
```diff
@@ -71,5 +76,39 @@ public static MojoModel load(MojoReaderBackend mojoReader) throws IOException {
...
+      int i = 0;
+      GenModel preprocessedModel = this;
+      for (MojoPreprocessor preprocessor : _preprocessors) {
+        MojoPreprocessor.ModelProcessor processor = preprocessor.makeProcessor(preprocessedModel);
+        converters[i] = processor.makeRowConverter(errorConsumer, config);
+        preprocessedModel = processor.getProcessedModel();
+        i++;
+      }
+      converters[i] = new CategoricalEncodingAsModelProcessor(preprocessedModel, this, categoricalEncoding).makeRowConverter(errorConsumer, config);
+//      converters[i] = makeDefaultRowConverter(columnToOffsetIdx, offsetToEncoder, errorConsumer, config);
```
honzasterba, line 97: commented out code

honzasterba (commented):
very quick rough review, this still seems like WIP

h2o-core/src/main/java/hex/ModelMojoWriter.java
sebhrusen: good point, was actually wondering about this as changing the path later is a source for future backwards incompatibility, will change.

h2o-core/src/main/java/hex/Model.java
sebhrusen: I'll remove this, I have a separate branch now where I started to experiment with this PUBDEV-7775 issue.
Some encodings like binary and ohe are actually broken in some conditions (especially the Mojo), and I was trying to get them working with TE on top, which is just too challening for this PR, and I'm 99% sure that `adaptTesttoTrain` with its crazy cyclomatic complexity and without any clear separation of concerns is responsible for this and needs to be rethought...

h2o-core/src/main/java/hex/Model.java
sebhrusen: It's a technic I learnt from Michal when you need to load multiple keys:
- one loop to prefetch
- one loop to use the fetched objects: this way when you're using the first one, the others can be deserialized in the background, so this loop is faster then.

h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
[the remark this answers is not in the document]
sebhrusen: personally I think it's a good practice to have an interface for non-trivial enums like those categorical encodings.
Otherwise each time you need or simply try to do something like combining those enums or create a plug something else (like I was trying at some point with TE), you end up having to rename everything.

So, I would keep the interface.
I'm open to suggestions if you prefer to change the name for the enum though: `DefaultCategoricalEncodings`?

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1369,7 +1374,7 @@ else if (_parms._weights_column != null && _weights != null && !_weights.isBinar
       if (isResponseOptional() && _parms._response_column != null && _response == null) {
         _vresponse = va.vec(_parms._response_column);
       }
-      _valid = adaptFrameToTrain(va, "Validation Frame", "_validation_frame", expensive);
+      _valid = adaptFrameToTrain(va, "Validation Frame", "_validation_frame", expensive, false); //FIXME: this encodes validation frame BEFORE TE is applied
```
sebhrusen, line 1377: creating a ticket for this one: this is closely related  to PUBDEV-7775 but not exactly equivalent.
1. we can wonder why in some circumstances only known to `adaptFrameToTrain` and not made explicit at the time this method is called, this method will decide to encode the categorical columns... which makes it impossible to work in combination with an additional encoder.
2. I have an experiment in local branch that modifies `adaptFrameToTrain` so that the encoder is injected as a param instead of being guessed. This allows to inject an encoder that does both TE + categorical encoding.
3. This is an issue only when model `categorical_encoding != AUTO`
4. AutoML always uses `categorical_encoding == AUTO` for now.
5. AutoML is the only "client" that can use this TE integration.
6. Therefore, I suggest to fix this concern in a separate PR.

h2o-genmodel/src/main/java/hex/genmodel/MojoModel.java
```diff
@@ -71,5 +76,38 @@ public static MojoModel load(MojoReaderBackend mojoReader) throws IOException {
...
+        MojoPreprocessor.ModelProcessor processor = preprocessor.makeProcessor(preprocessedModel);
+        converters[i] = processor.makeRowConverter(errorConsumer, config);
+        preprocessedModel = processor.getProcessedModel();
+        i++;
+      }
+      converters[i] = new CategoricalEncodingAsModelProcessor(preprocessedModel, this, categoricalEncoding).makeRowConverter(errorConsumer, config);
+      return new CompositeRowToRawDataConverter<>(converters);
+    }
+    
+    //FIXME: currently some categoricalEncoding that create new columns (binary, onehot...) don't put the non-predictors columns (e.g. fold) to the end (wrong offset).
```
sebhrusen, line 100: this is actually PUBDEV-7775, won't fix in this PR as it's not TE-related and doesn't impact the goal of this  integration for AutoML (AutoML uses only AUTO categorical encoding).

sebhrusen: removing this FIXME: using its content in PUBDEV-7775 description.

h2o-core/src/main/java/hex/ModelBuilder.java
sebhrusen: created follow-up tickt PUBDEV-7785 instead.

sebhrusen:
@erin, @honzasterba , @Pscheidl build is green ()outside r4.0 cran + one rulefit test): ready to merge: time to complain if you disagree :)

Pscheidl:
Punch it Mister Sulu, punch it !

https://i.gifer.com/RL8T.gif

h2o-genmodel/src/main/java/hex/genmodel/RowConverterFactory.java
```diff
@@ -0,0 +1,15 @@
...
+import hex.genmodel.easy.EasyPredictModelWrapper;
+import hex.genmodel.easy.RowToRawDataConverter;
+
+public interface RowConverterFactory {
+
+    /**
+     * Prefixed by _ as it is an internal API (skipped for validation by Mojoland).
+     * @return A new instance of {@link RowToRawDataConverter}
+     */
+    RowToRawDataConverter _makeRowConverter(CategoricalEncoding categoricalEncoding,
```
honzasterba, line 12: as MojoModel is the only class implementing this interface, it seems unnecessary and adds an ugly _public method to our API, why do we need this ifc?

sebhrusen: it's because it's called by `GenModel`

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1369,7 +1374,7 @@ else if (_parms._weights_column != null && _weights != null && !_weights.isBinar
       if (isResponseOptional() && _parms._response_column != null && _response == null) {
         _vresponse = va.vec(_parms._response_column);
       }
-      _valid = adaptFrameToTrain(va, "Validation Frame", "_validation_frame", expensive);
+      _valid = adaptFrameToTrain(va, "Validation Frame", "_validation_frame", expensive, false);  //cf. PUBDEV-7785
```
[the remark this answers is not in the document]
sebhrusen: cf. = confer = see also : https://en.wiktionary.org/wiki/cf.

h2o-core/src/main/java/hex/ModelMojoWriter.java
```diff
@@ -80,11 +82,34 @@ public ModelMojoWriter(M model) {
...
+    super.writeExtraInfo();
     writeModelDetails();
     writeModelDetailsReadme();
+    writeModelPreprocessors();
   }
+  
+  protected void writeModelPreprocessors() throws IOException {
+    if (model._parms._preprocessors == null) return;
+    for (int i=0; i < model._parms._preprocessors.length; i++) {
+      startWritingTextFile("preprocessing/preprocessor_"+i);
```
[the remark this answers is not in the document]
sebhrusen: good catch, probably leftover from a previous approach

h2o-extensions/target-encoder/src/test/java/ai/h2o/targetencoding/TargetEncoderPreprocessorTest.java
```diff
@@ -0,0 +1,672 @@
...
+            Model model = buildModel(train, null, tePreproc, CategoricalEncodingScheme.AUTO);
+            Scope.track_generic(model);
+            
+            int expectedCVModels = 3;  //3 folds -> 3 cv models
+            assertEquals(expectedCVModels, model._output._cross_validation_models.length);
+            assertTrue(ArrayUtils.contains(model._output._names, ENCODED));
+            assertFalse(ArrayUtils.contains(model._output._names, TO_ENCODE));
+            assertTrue(ArrayUtils.contains(model._output._names, NOT_ENCODED));
+            
+//            verify(teModel, times(expectedCVModels)).transformTraining(any(), anyInt());
```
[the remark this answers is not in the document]
sebhrusen: no, tried to use Mockito, but like almost everywhere in h2o-3, spies are unusable (final methods...), deleting this

h2o-genmodel/src/main/java/hex/genmodel/RowConverterFactory.java
sebhrusen: Btw this is public but this is not API.

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -376,9 +368,23 @@ public AbstractPrediction predict(RowData data, ModelCategory mc) throws Predict
   public AbstractPrediction predict(RowData data) throws PredictException {
     return predict(data, m.getModelCategory());
   }
+  
+  RowToRawDataConverter makeRowDataConverter(Config config, ErrorConsumer errorConsumer) {
+    CategoricalEncoding categoricalEncoding = config.getUseExternalEncoding() ? CategoricalEncodings.AUTO : m.getCategoricalEncoding();
+
+    RowToRawDataConverter rowDataConverter;
+    if (m instanceof RowConverterFactory) {
```
honzasterba, line 376: I think we can get rid of RowConverterFactory as it adds no value, you can do just `instanceof MojoModel` here and achieve the same result, also you can remove the underscore from the method name, the original method was public as well and did not need underscore

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1427,7 +1437,7 @@ else if (_parms._weights_column != null && _weights != null && !_weights.isBinar
         _train.restructure(_train.names(), vecs);
     }
     boolean names_may_differ = _parms._categorical_encoding == Model.Parameters.CategoricalEncodingScheme.Binary;
-    boolean names_differ = _valid !=null && !Arrays.equals(_train._names, _valid._names);
+    boolean names_differ = _valid !=null && ArrayUtils.difference(_train._names, _valid._names).length != 0;;
```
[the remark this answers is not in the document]
sebhrusen: no, order of columns in `_valid` should not matter, and before the now infamous PUBDEV-7785 is fixed, `_valid` can have the same columns as `_train` but in  a different order.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1390,7 +1397,10 @@ else if (_parms._weights_column != null && _weights != null && !_weights.isBinar
         _origTrain = null;
       }
       if (_valid != null) {
-        _valid = encodeFrameCategoricals(_valid, ! _parms._is_cv_model /* for CV, need to score one more time in outer loop */);
+        Frame newvalid = applyPreprocessors(_valid, false, scopeTrack);
+        newvalid = encodeFrameCategoricals(newvalid, scopeTrack /* for CV, need to score one more time in outer loop */);
+        newvalid = adaptFrameToTrain(newvalid, "Validation Frame", "_validation_frame", expensive, true);
```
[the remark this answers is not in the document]
sebhrusen: correct, this is not currently necessary: will only be needed with a proper fix for PUBDEV-7785

h2o-algos/src/main/java/hex/tree/SharedTreeModel.java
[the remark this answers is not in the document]
sebhrusen: @honzasterba please suggest a name, I'll rename..

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
sebhrusen: ok, I'll do that. I thought people would have more concerns about an explicit casting than about a underscore...

sebhrusen: well, now realizing that `MojoModel` is abstract, so I agree that interface can be removed

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1390,7 +1397,10 @@ else if (_parms._weights_column != null && _weights != null && !_weights.isBinar
         _origTrain = null;
       }
       if (_valid != null) {
-        _valid = encodeFrameCategoricals(_valid, ! _parms._is_cv_model /* for CV, need to score one more time in outer loop */);
+        Frame newvalid = applyPreprocessors(_valid, false, scopeTrack);
+        newvalid = encodeFrameCategoricals(newvalid, scopeTrack /* for CV, need to score one more time in outer loop */);
```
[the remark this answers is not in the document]
sebhrusen: it's applied only twice, 2 statements each time, and with different params, so I'm not convinced that grouping them in one method will improve anything.

honzasterba (commented):
my final tought: I think this solution is only half done, I think is too big of a change not be done properly. i think should decide one of two:
- make TE into a CategoricalEncoding
- make CategoricalEncoding into a ModelPreprocessor
The current solution does neither and bring additional complexity to already complex code.

h2o-core/src/main/java/hex/ModelBuilder.java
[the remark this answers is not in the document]
sebhrusen: we don't need symmetry, we just need to verify that all columns from train are in valid.

sebhrusen:
@honzasterba Let's repeat the goal of this integration again: this will be used only by AutoML.

therefore, goal is to have a small footprint change.

> my final tought: I think this solution is only half done, I think is too big of a change not be done properly. i think should decide one of two:
    make TE into a CategoricalEncoding
    make CategoricalEncoding into a ModelPreprocessor
    The current solution does neither and bring additional complexity to already complex code.

- making TE a CategoricalEncoding won't work, because TE is not just a categorical encoder. In most cases it needs to be followed by a categorical encoder because TE doesn't always encode all categorical columns.
- also as you can see with TargetEncoderPreprocessor, contrary to CategoricalEncoding, TE doesn't apply the same way on all frames, that's why need something more powerful than current CategoricalEncoding.
- making CategoricalEncoding as a ModelPreprocessor is a possibility (I mentioned in code comments), but it can be done as a follow-up ticket. As stated above, my goal here for AutoML is to avoid making structural changes on the existing code to avoid breaking things if not necessary, this is what we previously agreed with @michalkurka. Turning CategoricalEncoding into a ModelPreprocessor would probably be bigger than this PR, but for now, there's no good reason to put TE and CategoricalEncoding on the same level:
- Cat Encoders encode all remaining cat columns and they apply systematically. 
- TE encoder applies on selected columns, and there are complex logic defining how it should be applied to training vs validation vs test data...

michalkurka:
> making TE a CategoricalEncoding won't work, because TE is not just a categorical encoder. In most cases it needs to be followed by a categorical encoder because TE doesn't always encode all categorical columns.

but isn't this only due to the limitation of the current API that lets you specify a single encoding for the whole frame? if the API allowed to specify encoding per column then would the issue still exist?

in the context of AutoML we can ignore the issue completely: AutoML is providing an encoder that just handles the frame - no one needs to know this encoder is internally made of a composition of TE + default (or other AutoML selected) encoding

for algo users (eventually), it would be nice if they could just select TE as the desired encoding and H2O would "smartly" apply a reasonable default TE on columns with >= min_categories and use default for the rest. 

To me it makes little sense to fallback to anything else than the default to be honest... I cannot forsee a reasonable use case when someone will do TE with fallback to eg. eigenencoding. That is just not going to happen. And if someone has these desires then they could eventually build a pipeline. But the behavior out of the box should be straightforward and optimized for "real" world use cases.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1524,18 +1546,43 @@ private Frame adaptFrameToTrain(Frame fr, String frDesc, String field, boolean e
...
-      assert encoded._key != null;
-      if (scopeTrack)
-        Scope.track(encoded);
-      else
-        _workspace.getToDelete(true).put(encoded._key, Arrays.toString(Thread.currentThread().getStackTrace()));
+  private Frame applyPreprocessors(Frame fr, boolean isTraining, boolean scopeTrack) {
+    if (_parms._preprocessors == null) return fr;
+
+    for (Key<ModelPreprocessor> key : _parms._preprocessors) {
+      DKV.prefetch(key);
```
michalkurka, line 1553: Thumbs up!

sebhrusen:
See version without MOJO/genmodel changes here: https://github.com/h2oai/h2o-3/pull/4945

PR #4889, opened 2020-09-01
word2vec: remove misleading examples using pre_trained parameter

michalkurka:


jakubhava (approved):
Thank you, makes more sense now :)

PR #4891, opened 2020-09-02
PUBDEV-7746 add int array deserialization

koniecsveta:
https://0xdata.atlassian.net/browse/PUBDEV-7746


mn-mikke (approved):
LGTM and also built locally with SW and can confirm that this change my problem with the `hidden` parameter. Thanks a lot @koniecsveta !

PR #4894, opened 2020-09-04
PUBDEV-7747 fix testDiabetesWithWeights() and testBestPracticeExample…

koniecsveta:
…WithLinearVariables() failing on multinode

https://0xdata.atlassian.net/browse/PUBDEV-7747

### DIFF
--- a/h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
+++ b/h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
@@ -6,31 +6,29 @@ import hex.genmodel.utils.DistributionFamily;
 import hex.glm.GLM;
 import hex.glm.GLMModel;
 import org.junit.Assert;
-import org.junit.BeforeClass;
 import org.junit.Test;
 
+import org.junit.runner.RunWith;
 import water.DKV;
-import water.Key;
 import water.Scope;
 import water.TestUtil;
 import water.fvec.Frame;
 import water.fvec.Vec;
+import water.runner.CloudSize;
+import water.runner.H2ORunner;
 import water.test.util.ConfusionMatrixUtils;
 import static org.junit.Assert.*;
 
+@RunWith(H2ORunner.class)
+@CloudSize(1)
 public class RuleFitTest extends TestUtil {
-
-    @BeforeClass public static void setup() { stall_till_cloudsize(1); }
 
     @Test
     public void testBestPracticeExample() {
         // https://github.com/h2oai/h2o-tutorials/blob/8df6b492afa172095e2595922f0b67f8d715d1e0/best-practices/explainable-models/rulefit_analysis.ipynb
-        RuleFitModel rfModel = null;
-        GLMModel glmModel = null;
-        Frame fr = null, fr2 = null, fr3 = null;
         try {
             Scope.enter();
-            fr = parse_test_file("./smalldata/gbm_test/titanic.csv");
+            final Frame fr = Scope.track(parse_test_file("./smalldata/gbm_test/titanic.csv"));
             Scope.track(fr);
 
             String responseColumnName = "survived";
@@ -52,12 +50,13 @@ public class RuleFitTest extends TestUtil {
             params._model_type = RuleFitModel.ModelType.RULES;
             params._distribution = DistributionFamily.bernoulli;
 
-            rfModel = new RuleFit(params).trainModel().get();
+            final RuleFitModel rfModel = new RuleFit(params).trainModel().get();
+            Scope.track_generic(rfModel);
 
             System.out.println("Intercept: \n" + rfModel._output._intercept[0]);
             System.out.println(rfModel._output._rule_importance);
 
-            fr2 = rfModel.score(fr);
+            final Frame fr2 = Scope.track(rfModel.score(fr));
 
             Assert.assertTrue(rfModel.testJavaScoring(fr,fr2,1e-4));
 
@@ -70,9 +69,10 @@ public class RuleFitTest extends TestUtil {
 
             GLMModel.GLMParameters glmParameters = rfModel.glmModel._parms;
             glmParameters._train = fr._key;
-            glmModel = new GLM(glmParameters).trainModel().get();
+            final GLMModel glmModel = new GLM(glmParameters).trainModel().get();
+            Scope.track_generic(glmModel);
 
-            fr3 = glmModel.score(fr);
+            final Frame fr3 = Scope.track(glmModel.score(fr));
             predictions = fr3.vec("predict");
 
             ConfusionMatrix glmConfusionMatrix = ConfusionMatrixUtils.buildCM(data, predictions);
@@ -103,19 +103,6 @@ public class RuleFitTest extends TestUtil {
             System.out.println("GLM r2: " + GLMScoringInfo.scored_train._r2);
             
         } finally {
-            if (fr != null) fr.remove();
-            if (fr2 != null) fr2.remove();
-            if (fr3 != null) fr3.remove();
-            if (rfModel != null) {
-                if (rfModel.glmModel != null) rfModel.glmModel.remove();
-                if (rfModel.treeModels != null) {
-                    for (int i = 0; i < rfModel.treeModels.length; i++) {
-                        rfModel.treeModels[i].remove();
-                    }
-                }
-                rfModel.remove();
-            }
-            if (glmModel != null) glmModel.remove();
             Scope.exit();
         }
     }
@@ -123,13 +110,9 @@ public class RuleFitTest extends TestUtil {
     @Test
     public void testBestPracticeExampleWithLinearVariables() {
         // the same as above but uses rules + linear terms
-        RuleFitModel rfModel = null;
-        GLMModel glmModel = null;
-        Frame fr = null, fr2 = null, fr3 = null;
         try {
             Scope.enter();
-            fr = parse_test_file("./smalldata/gbm_test/titanic.csv");
-
+            final Frame fr = Scope.track(parse_test_file("/Users/zuzanaolajcova/IdeaProjects/h2o-3/smalldata/gbm_test/titanic.csv"));
             String responseColumnName = "survived";
             asFactor(fr, responseColumnName);
             asFactor(fr, "pclass");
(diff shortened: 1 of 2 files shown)


tomasfryda (approved):
It fixes the issue but I am not so comfortable saying this is correct fix (not enough experience with `Scope`/`DKV` on my side).

From what I gather using `@RunWith(H2ORunner.class)` is preferred to detect key leakages (I'm not 100% sure but Pavel recommended it to me in one case; @maurever might know...).
E.g.:
```java
@RunWith(H2ORunner.class)
@CloudSize(1)
public class ...
```

koniecsveta:
> It fixes the issue but I am not so comfortable saying this is correct fix (not enough experience with `Scope`/`DKV` on my side).
> 
> From what I gather using `@RunWith(H2ORunner.class)` is preferred to detect key leakages (I'm not 100% sure but Pavel recommended it to me in one case; @maurever might know...).
> E.g.:
> 
> ```java
> @RunWith(H2ORunner.class)
> @CloudSize(1)
> public class ...
> ```

@Pscheidl what do you think?

h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
```diff
@@ -128,7 +129,6 @@ public void testBestPracticeExampleWithLinearVariables() {
         try {
             Scope.enter();
             fr = parse_test_file("./smalldata/gbm_test/titanic.csv");
-            Scope.track(fr);
```
maurever, line 131: I think you should tract the frame after all changes and then it should work correctly. 

Do you run this test on a multinode locally?

Pscheidl:
Using `H2ORunner` has some advantages - it is the preferred way to do it. Besides it being a proper MR task, collecting leaked keys from each node separately etc., which is a minor technical detail (the logic is actually the same as `TestUtil`), it has some user-facing benefits

- Key leak detection occurs after each test, and not after the test class. Therefore, the leak is properly marked. TestUtil only marks the last test as the failure, yet the origin might be elsewhere.
- Better output - Groups leaked keys. Also better output for rollupstats, including node of origin etc.
- Test class-scoped resources (introduced in BeforeClass) are not reported as leaks :wink: 

I'd highly recommend using it, as this is the one development focuses on.

h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
Pscheidl: Veronika is probably right, there are untracked vecs added. Haven't seen the failure. I'll try to come up with a sample fix.

Pscheidl:
This is just thinking out loud at the moment, but in the finally block, the removal code looks like this:

```java
                if (rfModel.treeModels != null) {
                    for (int i = 0; i < rfModel.treeModels.length; i++) {
                        rfModel.treeModels[i].remove();
                    }
                }
 ```

At first sight, this looks suspicious. When a model is removed (scope ended), these trees should be deleted. Looks like implementation of remove could be improved. @koniecsveta can you confirm the lifetime of these trees is exactly the same as the lifetime of RF model ? Or in other words, if a RF model is removed, should those trees be removed as well ? I keen to say yes.

**Comment by Pscheidl** [comment]
As mentioned above, the cleanup of a RuleFit model represented by `remove_impl` method was not sufficient - the internals models were not removed, resulting in a leakage.

Below is a patch to be found that fixes it from my perspective. I've modified the test you marked as failing and only used scopes in a "clean" way (up to you to judge). I've also implemented the `remove_impl` method, which, when `cascade = true` removes the internal trees and models. This gives me nice clean DKV afterwards.

If you find that solution acceptable @koniecsveta and others, it would be nice to modify the other tests to only use `Scope` as well.

```patch
Index: h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
IDEA additional info:
<+>UTF-8
--- h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java	(revision 7ac103414a3731342abc573590a30b06b35834b5)
+++ h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java	(date 1599559293397)
@@ -136,4 +136,17 @@
         return destination;
     }
 
+    @Override
+    protected Futures remove_impl(Futures fs, boolean cascade) {
+        super.remove_impl(fs, cascade);
+        if(cascade) {
+            for (SharedTreeModel treeModel : treeModels) {
+                treeModel.remove(fs);
+            }
+
+            glmModel.remove(fs);
+        }
+        
+        return fs;
+    }
 }
Index: h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
IDEA additional info:
<+>UTF-8
--- h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java	(revision 7ac103414a3731342abc573590a30b06b35834b5)
+++ h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java	(date 1599558800961)
@@ -9,19 +9,21 @@
 import org.junit.BeforeClass;
 import org.junit.Test;
 
-import water.DKV;
+import org.junit.runner.RunWith;
 import water.Key;
 import water.Scope;
 import water.TestUtil;
 import water.fvec.Frame;
 import water.fvec.Vec;
+import water.runner.CloudSize;
+import water.runner.H2ORunner;
 import water.test.util.ConfusionMatrixUtils;
 
+@RunWith(H2ORunner.class)
+@CloudSize(1)
 public class RuleFitTest extends TestUtil {
 
-    @BeforeClass public static void setup() { stall_till_cloudsize(1); }
-
     @Test
     public void testBestPracticeExample() {
         // https://github.com/h2oai/h2o-tutorials/blob/8df6b492afa172095e2595922f0b67f8d715d1e0/best-practices/explainable-models/rulefit_analysis.ipynb
@@ -123,12 +125,9 @@
     @Test
     public void testBestPracticeExampleWithLinearVariables() {
         // the same as above but uses rules + linear terms
-        RuleFitModel rfModel = null;
-        GLMModel glmModel = null;
-        Frame fr = null, fr2 = null, fr3 = null;
         try {
             Scope.enter();
-            fr = parse_test_file("./smalldata/gbm_test/titanic.csv");
+            final Frame fr = parse_test_file("./smalldata/gbm_test/titanic.csv");
 
             String responseColumnName = "survived";
             asFactor(fr, responseColumnName);
@@ -144,7 +143,7 @@
             final Vec weightsVector = Vec.makeOne(fr.numRows());
             final String weightsColumnName = "weights";
             fr.add(weightsColumnName, weightsVector);
-            DKV.put(fr);
+            Scope.track(fr);
 
             RuleFitModel.RuleFitParameters params = new RuleFitModel.RuleFitParameters();
             params._seed = 1234;
@@ -154,12 +153,13 @@
             params._model_type = RuleFitModel.ModelType.RULES_AND_LINEAR;
             params._weights_column = "weights";
 
-            rfModel = new RuleFit(params).trainModel().get();
+            final RuleFitModel rfModel = new RuleFit(params).trainModel().get();
+            Scope.track_generic(rfModel);
 
             System.out.println("Intercept: \n" + rfModel._output._intercept[0]);
             System.out.println(rfModel._output._rule_importance);
 
-            fr2 = rfModel.score(fr);
+            final Frame fr2 = Scope.track(rfModel.score(fr));
 
             Assert.assertTrue(rfModel.testJavaScoring(fr,fr2,1e-4));
 
@@ -172,9 +172,10 @@
 
             GLMModel.GLMParameters glmParameters = rfModel.glmModel._parms;
             glmParameters._train = fr._key;
-            glmModel = new GLM(glmParameters).trainModel().get();
+            final GLMModel glmModel = new GLM(glmParameters).trainModel().get();
+            Scope.track_generic(glmModel);
 
-            fr3 = glmModel.score(fr);
+            final Frame fr3 = Scope.track(glmModel.score(fr));
             predictions = fr3.vec("predict");
 
             ConfusionMatrix glmConfusionMatrix = ConfusionMatrixUtils.buildCM(data, predictions);
@@ -205,19 +206,6 @@
             System.out.println("GLM r2: " + GLMScoringInfo.scored_train._r2);
 
         } finally {
-            if (fr != null) fr.remove();
-            if (fr2 != null) fr2.remove();
-            if (fr3 != null) fr3.remove();
-            if (rfModel != null) {
-                if (rfModel.glmModel != null) rfModel.glmModel.remove();
-                if (rfModel.treeModels != null) {
-                    for (int i = 0; i < rfModel.treeModels.length; i++) {
-                        rfModel.treeModels[i].remove();
-                    }
-                }
-                rfModel.remove();
-            }
-            if (glmModel != null) glmModel.remove();
             Scope.exit();
         }
     }
@@ -513,11 +501,11 @@
         try {
             Scope.enter();
             fr = parse_test_file("./smalldata/diabetes/diabetes_text_train.csv");
+            Scope.track(fr);
 
             final Vec weightsVector = createRandomBinaryWeightsVec(fr.numRows(), 10);
             final String weightsColumnName = "weights";
             fr.add(weightsColumnName, weightsVector);
-            DKV.put(fr);
             
             RuleFitModel.RuleFitParameters params = new RuleFitModel.RuleFitParameters();
             params._seed = 12345;
```

h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
koniecsveta: yes when I do only Scope.track even after changes it still fails on multinode (4 nodes)

koniecsveta:
@Pscheidl Thanks a lot! I will incorporate this as a general clean up of the RF, but unfortunately, this is not solving the failing cases.

Pscheidl:
Sure, let's fix that as well. One after another.

Pscheidl:
@koniecsveta 

Yes, the solution is simple. There were columns added to the frame, yet DKV doesn't know about those vecs. Adding `DKV.put(fr);`

after 

```java
            final Vec weightsVector = Vec.makeOne(fr.numRows());
            final String weightsColumnName = "weights";
            fr.add(weightsColumnName, weightsVector);
```

solves the problem.

koniecsveta:
> @koniecsveta
> 
> Yes, the solution is simple. There were columns added to the frame, yet local KV store on other nodes doesn't know about those vecs. Adding `DKV.put(fr);`
> 
> after
> 
> ```java
>             final Vec weightsVector = Vec.makeOne(fr.numRows());
>             final String weightsColumnName = "weights";
>             fr.add(weightsColumnName, weightsVector);
> ```
> 
> solves the problem.

perfect! this is what was used in 7ac1034

h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
```diff
@@ -69,9 +69,10 @@ public void testBestPracticeExample() {
 
             GLMModel.GLMParameters glmParameters = rfModel.glmModel._parms;
             glmParameters._train = fr._key;
-            glmModel = new GLM(glmParameters).trainModel().get();
+            final GLMModel glmModel = new GLM(glmParameters).trainModel().get();
+            Scope.track_generic(glmModel);
 
-            fr3 = glmModel.score(fr);
+            final Frame fr3 = Scope.track(glmModel.score(fr));
```
Pscheidl, line 75: Unused variable.

Pscheidl (approved):
One unused variable to fix, otherwise we're all good to merge. Thank you @koniecsveta .

h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
koniecsveta: fixed

PR #4895, opened 2020-09-04
PUBDEV-7736 use btree for coxph computation instead the direct approach

satai:


h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
```diff
@@ -166,64 +174,343 @@ static Stats concordance(final Vec startVec, final Vec stopVec, final Vec eventV
...
+                 .boxed()
+                 .collect(groupingBy(
+                      i -> strataVecs.stream()
+                              .map(v -> v.at8(i))
+                              .collect(toList())
+                 )).values();
+
+      return byStrata.stream()
+        .map(
+          indexes -> statsForAStrata(startVec, stopVec, eventVec, estimateVec, indexes)
```
michalkurka, line 189: Why is this using the Vec API at all? If you know you have only a very limited number of indices - can it be collected to non-distributed structures and avoid all overhead with using the Vec API to access a single value?

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
```diff
@@ -166,64 +174,343 @@ static Stats concordance(final Vec startVec, final Vec stopVec, final Vec eventV
     private static Stats concordanceStats(Vec.Reader startVec, Vec.Reader stopVec, Vec.Reader eventVec, List<Vec.Reader> strataVecs, Vec.Reader estimateVec, long length) {
       assert 0 <= length && length <= Integer.MAX_VALUE;
       
-      Stream<Integer> allIndexes = IntStream.range(0, (int) length).boxed();
+      Collection<List<Integer>> byStrata =
```
michalkurka, line 177: this should be rewritten to MRTask

satai: As you wish

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
```diff
@@ -147,83 +137,470 @@ long discordant() {
...
+        fr = prepareFrameForConcordanceComputation(eventVec, strataVecs, estimateVec, durations);
+        return concordanceStats(fr);
+      } finally {
+        Scope.exit();
+        fr.delete(false);
+      }
+    }
+
+    private static Frame prepareFrameForConcordanceComputation(Vec eventVec, List<Vec> strataVecs, Vec estimateVec, Vec durations) {
+      final Frame fr = new Frame(Key.make());
```
michalkurka, line 170: does it need to have a key?

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
```diff
@@ -147,83 +137,470 @@ long discordant() {
...
+      fr.add("duration", durations);
+      fr.add("event", eventVec);
+      fr.add("estimate", estimateVec);
+      for (int i = 0; i < strataVecs.size(); i++) {
+        fr.add("strata_" + i, strataVecs.get(i));
+      }
+      DKV.put(fr);
```
michalkurka, line 177: probably no need to install into DKV

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
```diff
@@ -147,83 +137,470 @@ long discordant() {
...
+        @Override
+        public void map(Chunk c0, Chunk c1, NewChunk nc) {
+          for (int i = 0; i < c0._len; i++)
+            nc.addNum(c1.atd(i) - c0.atd(i));
+        }
+      }.doAll(1, Vec.T_NUM, fr1)
+              .outputFrame(Key.make("durations_" + fr1._key), new String[]{"durations"}, null));
+      Scope.track(frame);
+      
+      fr1.delete(false);
```
michalkurka, line 196: why false?

michalkurka:
I think there is a lot of unnecessary DKV#puts, you really only need to put a Frame in DKV when it is a user-facing result. Most (if not all) your frames have a temporary nature - for regular MRTasks they don't need to have a key, don't need to be in DKV.

Working with Scope and explicit deletes is suspicious and unexpected.

satai:
@michalkurka you were right, that DKV and Scope ops were not necessary. Pushed a cleanup commit

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
```diff
@@ -147,83 +134,455 @@ long discordant() {
...
+      fr.add("event", eventVec);
+      fr.add("estimate", estimateVec);
+      for (int i = 0; i < strataVecs.size(); i++) {
+        fr.add("strata_" + i, strataVecs.get(i));
+      }
+      return fr;
+    }
+
+    private static Vec durations(Vec startVec, Vec stopVec) {
+      Vec vec = null == startVec ? stopVec.makeZero() : startVec;
```
michalkurka, line 170: I don't see the temporary frame disposed of anywhere.

On another note - if startVec is not given then durations = startVec. You can skip the MRTask altogether.

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
```diff
@@ -147,83 +134,455 @@ long discordant() {
...
+
+    private static Frame removeNAs(Frame fr) {
+      final int[] iDontWantNAsInThisCols = new int[]{0, 2};
+
+      final Frame withoutNas = new Merge.RemoveNAsTask(iDontWantNAsInThisCols)
+                                        .doAll(fr.types(), fr)
+                                        .outputFrame(fr.names(), fr.domains());
 
-      return stats;
+      withoutNas.replace(1, withoutNas.vec("event").toNumericVec());
```
michalkurka, line 244: removeNAs is also converting event column to numeric? 

I would expect such transformations to be done in top-level (public) methods instead of utility functions.

h2o-py/tests/testdir_algos/coxph/pyunit_coxph_concordance.py
```diff
@@ -20,8 +20,6 @@ def coxph_smoke():
 
     metrics = coxph.model_performance(heart)
     assert 0.581 > metrics.concordance() and 0.580 < metrics.concordance()
-    assert 3696 == metrics.concordant()
```
michalkurka, line 23: intentional?

h2o-r/tests/testdir_algos/coxph/runit_coxph_concordance.R
```diff
@@ -24,7 +24,7 @@ test.CoxPH.concordance <- function() {
     rConcordance <- unname(summary(rModel)$concordance)[1]
     hexConcordance <- h2o.performance(hexModel, data=tstdataHex)@metrics$concordance
     
-    expect_equal(rConcordance, hexConcordance)
+    expect_equal(rConcordance, hexConcordance, tolerance = 1e-3, scale = 1)
```
michalkurka, line 27: Isn't the tolerance quite high? This is not complex math

h2o-py/tests/testdir_algos/coxph/pyunit_coxph_concordance.py
satai: Rebased, the test was redesigned in an other PR to use lifelines instead of hardcoded results.

h2o-r/tests/testdir_algos/coxph/runit_coxph_concordance.R
satai: It was used in all the coxph tests so far I have seen.

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
satai: was removed

h2o-core/src/main/java/hex/ModelMetricsRegressionCoxPH.java
satai: updated

michalkurka (approved):

PR #4903, opened 2020-09-09
Let user define logger config for h2o (library mode)

sebhrusen:
https://0xdata.atlassian.net/browse/PUBDEV-7755


michalkurka (approved):
Thank you for the quick fix

PR #4906, opened 2020-09-10
PUBDEV-7760: Make sure we don't use lazy-loading for sequential zip parse

michalkurka:


h2o-persist-http/src/main/java/water/persist/PersistHTTP.java
```diff
@@ -93,35 +100,49 @@ private static long parseContentRangeLength(Header contentRange) {
...
+      return checkRangeSupport(uri, response);
+    }
+  }
+
+  static boolean isCompressed(HttpResponse response) {
+    Header contentTypeHeader = response.getFirstHeader(HttpHeaders.CONTENT_TYPE);
+    if (contentTypeHeader == null)
+      return false; // assume not compressed
+    String contentType = contentTypeHeader.getValue();
+    return COMPRESSED_CONTENT_TYPES.contains(contentType);
```
satai, line 125: I am afraid content type can be case insensitive so contentType.toLowerCase() to play it safe?

h2o-persist-http/src/main/java/water/persist/PersistHTTP.java
```diff
@@ -93,35 +100,51 @@ private static long parseContentRangeLength(Header contentRange) {
...
+    try (CloseableHttpClient client = HttpClientBuilder.create().build();
+         CloseableHttpResponse response = client.execute(req)) {
+      if (isCompressed(response)) 
+        return -1L; // avoid lazy-loading of compressed resource that cannot be parsed in parallel
+      return checkRangeSupport(uri, response);
+    }
+  }
+
+  static boolean isCompressed(HttpResponse response) {
+    Header contentTypeHeader = response.getFirstHeader(HttpHeaders.CONTENT_TYPE);
```
satai, line 121: Doesn't this code scream "use applicative functors" to you? :-D

satai (approved):

h2o-persist-http/src/main/java/water/persist/PersistHTTP.java
michalkurka: no, it doesn't even whisper ;)

PR #4910, opened 2020-09-12
[PUBDEV-7683] Remote XGBoost: Improve performance of matrix data transfer

honzasterba:
data upload between clusters with large dataset (2gb in memory)
original implementation: ~90sec
this branch: ~16sec


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -109,7 +109,7 @@ else if (f.vec(i-catCols).isInt())
      * @param weightsVector Vector with row weights, possibly null
      * @return A sum of chunk lengths. Possibly zero, if there are no chunks or the chunks are empty.
      */
-    private static long sumChunksLength(int[] chunkIds, Vec vec, Vec weightsVector, int[] chunkLengths) {
+    public static long sumChunksLength(int[] chunkIds, Vec vec, Vec weightsVector, int[] chunkLengths) {
```
Pscheidl, line 112: This methods itself, originally written by me I believe, could be a major slow down, as it iterates of whole vec in a sequential way. As I was the one to introduce the problem, I offer to make it distributed @honzasterba .

Pscheidl: Nevermind, I forgot this is only done for local chunks. :man_facepalming: 

Could be calculated beforehand for the whole cluster anyway, if possible and not calculate it each round. Something like this:

```java
    private static class ChunksLenTask extends MRTask<ChunksLenTask> {
        public int[] chunkLengths;

        ChunksLenTask(final int nChunks) {
            chunkLengths = new int[nChunks];
            Arrays.fill(chunkLengths, -1);
        }

        @Override
        public void map(Chunk c0, Chunk weightsChunk) {
            chunkLengths[c0.cidx()] = c0._len;
            int lastNonZeroIndex = weightsChunk.nextNZ(-1, true);

            while (weightsChunk.nextNZ(lastNonZeroIndex, true) != -1) {
                chunkLengths[c0.cidx()]--;
                lastNonZeroIndex = weightsChunk.nextNZ(lastNonZeroIndex, true);
            }

        }

        @Override
        public void reduce(ChunksLenTask mrt) {
            for (int i = 0; i < mrt.chunkLengths.length; i++) {
                if (mrt.chunkLengths[i] != -1) {
                    chunkLengths[i] = mrt.chunkLengths[i];
                }
            }
        }

        public long nonZeroChunkLength() {
            return ArrayUtils.suml(chunkLengths);
        }
    }
```

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/RemoteMatrixLoader.java
```diff
@@ -0,0 +1,104 @@
...
+        final SparseMatrix matrix;
+
+        RemoteSparseMatrix(SparseMatrixDimensions dims) {
+            this.dims = dims;
+            this.matrix = SparseMatrixFactory.allocateCSRMatrix(dims);
+        }
+
+        @Override
+        MatrixLoader.DMatrixProvider make() {
+            return SparseMatrixFactory.toDMatrix(matrix, dims, data.actualRows, data.shape, data.resp, data.weights, data.offsets);
```
koniecsveta, line 30: is it somehow ensured that _data_ is defined when this method is called?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/task/XGBoostUploadMatrixTask.java
```diff
@@ -1,42 +1,375 @@
...
+                featChunks[i] = _frame.vecs()[i].chunkForChunkIdx(chunk);
+            }
+            SparseMatrixChunk chunkData = new SparseMatrixChunk(chunkIdx, rowHeaderSize, (int) dataSize);
+            int dataIndex = 0;
+            int rowHeaderIndex = 0;
+            for(int i = 0; i < respChunk._len; i++) {
+                if (weightChunk != null && weightChunk.atd(i) == 0) continue;
+                chunkData.rowHeader[rowHeaderIndex++] = nonZeroCount;
+                _actualRows[chunkIdx]++;
+                for (int j = 0; j < _di._cats; ++j) {
```
koniecsveta, line 349: it may sound pedantic, but even though there is no difference in ++j vs j++ in the for loop incrementation is there any reason why not to follow single convention? :)

honzasterba: there is not, it should always be j++ as the convention dictates

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/RemoteMatrixLoader.java
honzasterba: yes, this should be called only after the matrix was completely uploaded and matrixData was called for this matrix

koniecsveta (approved):

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/DenseMatrixFactory.java
```diff
@@ -51,8 +51,40 @@ protected DenseDMatrixProvider(
...
+                    }
+                }
+            }
+            return true;
+        }
 
-        public DenseDMatrixProvider() {}
+        @Override
+        public int hashCode() {
+            return super.hashCode() + Objects.hash(data);
```
michalkurka, line 86: 2 comments:
1) BigDenseMatrix doesn't implement hashCode function - is that an issue
2) Using Objects.hash, in this case, is inefficient, a new array to wrap the single element needs to be created, it then essentially just does something like `data == null ? 0 : 31 * data.hashCode()`

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/matrix/MatrixLoader.java
```diff
@@ -43,19 +45,23 @@ public final DMatrix get() throws XGBoostError {
...
         }
 
         @Override
-        public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
-            actualRows = in.readLong();
-            response = (float[]) in.readObject();
-            weights = (float[]) in.readObject();
-            offsets = (float[]) in.readObject();
+        public int hashCode() {
+            int result = Objects.hash(actualRows);
```
michalkurka, line 60: This is a very expensive implementation of hashcode, number of rows + optionally maybe first few elements of the array should be good enough

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/task/XGBoostUploadMatrixTask.java
```diff
@@ -1,42 +1,373 @@
...
+            matrixData.actualRows = dense(
+                client, chunks, nRows, nRowsByChunk, weightVec, offsetsVec, responseVec, modelInfo.dataInfo(),
+                matrixData.resp, matrixData.weights, matrixData.offsets
+            );
+        }
+        client.uploadObject(_modelKey, RemoteXGBoostUploadServlet.RequestType.matrixData, matrixData);
+        LOG.debug("Matrix upload finished in " + ((System.currentTimeMillis() - start) / 1000d));
+    }
+
+    public static class MatrixData implements Serializable {
```
michalkurka, line 101: Since this holds data - Iced serialization is much more efficient, why are we using Serializables?

michalkurka (approved):
LGTM

some nitpicking about hashcode implementations ;)

honzasterba:
the hashcode implementations are there just to keep the equals/hashCode contract, they are never used and probably never will be

PR #4911, opened 2020-09-13
PUBDEV-7766: Add context_path in h2o.init in python

skumagai:
Link to JIRA ticket: https://0xdata.atlassian.net/browse/PUBDEV-7766

This PR adds context_path as a new argument in h2o.init in python.
It touches h2o/h2o.py, h2o/backend/server.py and h2o/backend/connection.py.
While I try to provide a unit test, I'm not sure what's already done is sufficient or acceptable.
Documentation is not touched so far.

Comments welcome.


maurever:
Hi @skumagai . Thank you for your contribution. We currently don't have the capacity to assist you in completing this task. We decided that this change is not critical for improving our codebase, so we have closed this PR.

PR #4914, opened 2020-09-15
PUBDEV-7720 fill rulefit model metrics by glm model metrics

koniecsveta:


h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
```diff
@@ -101,6 +101,12 @@ public void testBestPracticeExample() {
 
             System.out.println("RuleFit r2: " + RuleFitScoringInfo.scored_train._r2);
             System.out.println("GLM r2: " + GLMScoringInfo.scored_train._r2);
+
+            System.out.println("RuleFit AUC:" +  rfModel.auc());
```
Pscheidl, line 105: Would any asserts be useful ? What if rfModel.auc() returns nonsense ?

koniecsveta: If models auc would change to some nonsense then already previous asserts would've failed:
```
assertEquals(ruleFitConfusionMatrix.accuracy(),0.7868601986249045,1e-4);
assertEquals(ruleFitConfusionMatrix.specificity(),0.8207663782447466,1e-4);
assertEquals(ruleFitConfusionMatrix.recall(),0.732,1e-4);
```

I've added 
`System.out.println("RuleFit AUC:" +  rfModel.auc());`

+- just to demonstrate functionality

Pscheidl: I see. There was no such thing in the PR and name of the PR doesn't imply there are such asserts already present. If yes, then all is fine :+1:

honzasterba (approved):

PR #4917, opened 2020-09-15
PUBDEV-7753 - Disable REST API on non-leader K8S nodes

Pscheidl:
Still WIP/DRAFT, looks easy to do at first sight as by default, the APIs are enabled and then the API is stopped being served by H2OKubernetesEmbeddedConfig when leader nodes are detected. As `notifyAboutCloudSize` is called in `Paxos` at line 129 after a leader node has been selected, this should be safe to do, as newly added nodes are being disabled as this callback is called and leader node remains open.

Let's see how this works in practice in our test pipeline (temporarily added Kubernetes tests to pipeline tests). Also, I'm thinking about how to test it. One way is to introduce special diagnostic Kubernetes API (this is not H2O API), as it's a lot of hacky work to expose each pod separately via ingress (a service has to be created for each pod separately and then the corresponding ingress only for diagnostics has to be linked with the service - problem is the names have to be obtained dynamically, as we have no control over K8S naming)

### DIFF
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -1846,7 +1846,7 @@ final public class H2O {
    */
   public static void startServingRestApi() {
     if (!H2O.ARGS.disable_web) {
-      NetworkInit.h2oHttpView.acceptRequests();
+      NetworkInit.h2oHttpView.setAcceptRequests(true);
     }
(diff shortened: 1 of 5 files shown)


Pscheidl:
Current WIP solution is NOT working, I'll have to come up with a less cheap solution.

PR #4921, opened 2020-09-16
[PUBDEV-7769] allow GPU backend when _build_tree_one_node is set

honzasterba:


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/exec/LocalXGBoostExecutor.java
```diff
@@ -104,7 +104,7 @@ public LocalXGBoostExecutor(XGBoostModel model, Frame train) {
     private RabitTrackerH2O setupRabitTracker(int numNodes) {
         // XGBoost seems to manipulate its frames in case of a 1 node distributed version in a way 
         // the GPU plugin can't handle Therefore don't use RabitTracker envs for 1 node
-        if (H2O.CLOUD.size() > 1) {
+        if (numNodes > 1) {
```
Pscheidl, line 107: You could also be interesting in look at line 50 in the same file: `nodes = new boolean[H2O.CLOUD.size()];`

Pscheidl (approved):

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/exec/LocalXGBoostExecutor.java
honzasterba: thats a different story, that code is used when remotexgboost is executed

PR #4922, opened 2020-09-16
PUBDEV-7745: Add default values for all arguments of modeling functions in User Guide

hannah-tillman:
Added default values for modeling functions in user guide per https://0xdata.atlassian.net/jira/software/c/projects/PUBDEV/issues/PUBDEV-7745


ledell:
Thanks @hannah-tillman, this a HUGE improvement/effort!!  I pushed some updates, and I have some more remaining... will get back to this on Monday!

PR #4924, opened 2020-09-17
PUBDEV-7720-fix_corner_case_when_setLambdaOptimal()

koniecsveta:
this PR fixes the way a lambda parameter is chosen for GLM in that part of algo, when we are going to fit a GLM to the rule feature set joined with the original feature set. Originally, it was done like what is in getBestLambdaIndex() in this PR, but there were corner cases found, so it was fixed to what is in getBestLambdaIndexCornerCase() in this PR, but this approach is good for corner cases and previous approach is better than this one for not-corner cases, so this PR combines it.

PR #4925, opened 2020-09-17
fix indexing when handling depths for tree ensembles

koniecsveta:
this PR fixes an indexing bug causing that there is one more tree ensemble model trained than it should be during rulefit training

### DIFF
--- a/h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
+++ b/h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
@@ -228,17 +228,18 @@ public class RuleFitTest extends TestUtil {
             
             final Frame fr2 = Scope.track(model.score(fr));
             
-            double[] expectedCoeffs = new double[] {13.54857, 8.37943,  8.33535, 7.78235, 7.62020, -7.57865, -5.59529, 5.54992, -4.04620, -3.73222, -3.66495,
-                    -3.42013, -3.15808, -2.35471, -2.18179, 1.37956, -1.21565, -1.14398, -0.72780, -0.65794,  -0.60032, -0.51938, -0.24730, -0.21409, 0.16232,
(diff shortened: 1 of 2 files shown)


h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
```diff
@@ -228,17 +228,18 @@ public void testCarsRules() {
             
             final Frame fr2 = Scope.track(model.score(fr));
             
-            double[] expectedCoeffs = new double[] {13.54857, 8.37943,  8.33535, 7.78235, 7.62020, -7.57865, -5.59529, 5.54992, -4.04620, -3.73222, -3.66495,
-                    -3.42013, -3.15808, -2.35471, -2.18179, 1.37956, -1.21565, -1.14398, -0.72780, -0.65794,  -0.60032, -0.51938, -0.24730, -0.21409, 0.16232,
-                    0.15663, 0.11327, 0.09523, -0.02568, -0.02156, 0.00606, 0.00080,  -0.00059, 0.00000, 0.00000, -0.00000, -0.00000};
+            double[] expectedCoeffs = new double[] {13.50732, 8.37949, 8.33540, 7.78257, -7.57849, 7.51095, 5.65917, -5.59525, -4.04595, -3.73260, -3.66489,
```
Pscheidl, line 231: If these are not manually calculated, then this is test fitting to me. Anyway, good indicator if change.

Pscheidl (approved):

h2o-algos/src/test/java/hex/rulefit/RuleFitTest.java
koniecsveta: I was comparing results with meghans notebook (https://github.com/h2oai/h2o-tutorials/blob/8df6b492afa172095e2595922f0b67f8d715d1e0/best-practices/explainable-models/rulefit.py). However, because of solving corner case during lambda search for GLM (now will be aligned with notebook again https://github.com/h2oai/h2o-3/pull/4924), some results slightly diverged what I considered valid when using different lambda. But I found out it was not caused by that but by this indexing issue.

honzasterba (approved):

PR #4927, opened 2020-09-17
PUBDEV-7778: minimal integration of TargetEncoding into AutoML

sebhrusen:
see ticket https://0xdata.atlassian.net/browse/PUBDEV-7778 for details

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
@@ -45,7 +45,6 @@ public class TargetEncoding implements PreprocessingStep {
         TargetEncoderParameters params = new TargetEncoderParameters();
         params._train = amlTrain._key;
         params._response_column = _aml.getBuildSpec().input_spec.response_column;
-        params._ignored_columns = Arrays.stream(amlTrain.names()).filter(col -> !teColumns.contains(col)).toArray(String[]::new);
         params._keep_original_categorical_columns = false;
         params._blending = true;
         params._noise = 0;
@@ -74,6 +73,10 @@ public class TargetEncoding implements PreprocessingStep {
                 });
             }
         }
+        String[] keep = params.getNonPredictors();
+        params._ignored_columns = Arrays.stream(amlTrain.names())
+                .filter(col -> !teColumns.contains(col) && !ArrayUtils.contains(keep, col))
+                .toArray(String[]::new);
 
         TargetEncoder te = new TargetEncoder(params, _aml.makeKey(getType(), null, false));
         _teModel = te.trainModel().get();
--- a/h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
@@ -17,8 +17,7 @@ import water.fvec.Vec;
 import water.rapids.ast.prims.advmath.AstKFold;
 import water.util.ArrayUtils;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
 
 public class TargetEncoding implements PreprocessingStep {
     
@@ -28,6 +27,7 @@ public class TargetEncoding implements PreprocessingStep {
     private TargetEncoderPreprocessor _tePreprocessor;
     private TargetEncoderModel _teModel;
     private final List<Completer> _disposables = new ArrayList<>();
+    private int _cardinalityThreshold = 0;  // the minimal cardinality for a column to be TE encoded. 
 
     public TargetEncoding(AutoML aml) {
         _aml = aml;
@@ -40,9 +40,12 @@ public class TargetEncoding implements PreprocessingStep {
 
     @Override
     public void prepare() {
+        Frame amlTrain = _aml.getTrainingFrame();
+        Set<String> teColumns = selectColumnsToEncode(amlTrain);
         TargetEncoderParameters params = new TargetEncoderParameters();
-        params._train = _aml.getTrainingFrame()._key;
+        params._train = amlTrain._key;
         params._response_column = _aml.getBuildSpec().input_spec.response_column;
+        params._ignored_columns = Arrays.stream(amlTrain.names()).filter(col -> !teColumns.contains(col)).toArray(String[]::new);
         params._keep_original_categorical_columns = false;
         params._blending = true;
         params._noise = 0;
@@ -112,6 +115,28 @@ public class TargetEncoding implements PreprocessingStep {
             _tePreprocessor = null;
             _teModel = null;
         }
+    }
+
+    public void setCardinalityThreshold(int cardinalityThreshold) {
+        _cardinalityThreshold = cardinalityThreshold;
+    }
+
+    private Set<String> selectColumnsToEncode(Frame fr) {
+        Set<String> encode = new HashSet<>();
+        if (_cardinalityThreshold > 0) {
+            for (int i = 0; i < fr.names().length; i++) {
+                if (fr.vec(i).cardinality() >= _cardinalityThreshold) encode.add(fr.name(i));
+            }
+        } else {
+            encode.addAll(Arrays.asList(fr.names()));
+        }
+        return encode;
+    }
+
+    public static final String[] ignoredColumns(final Frame frame, final String... keep) {
+        Set<String> ignored = new HashSet(Arrays.asList(frame.names()));
+        ignored.removeAll(Arrays.asList(keep));
+        return ignored.toArray(new String[ignored.size()]);
     }
 
     TargetEncoderPreprocessor getTEPreprocessor() {
--- a/h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
@@ -1,6 +1,9 @@
 package ai.h2o.automl.preprocessing;
 
 import ai.h2o.automl.AutoML;
+import ai.h2o.automl.AutoMLBuildSpec;
+import ai.h2o.automl.AutoMLBuildSpec.AutoMLBuildControl;
+import ai.h2o.automl.AutoMLBuildSpec.AutoMLInput;
 import ai.h2o.targetencoding.TargetEncoder;
 import ai.h2o.targetencoding.TargetEncoderModel;
 import ai.h2o.targetencoding.TargetEncoderModel.DataLeakageHandlingStrategy;
@@ -22,6 +25,7 @@ import java.util.*;
 public class TargetEncoding implements PreprocessingStep {
     
     static String TE_FOLD_COLUMN_SUFFIX = "_te_fold";
+    private static final Completer NOOP = () -> {};
     
     private AutoML _aml;
     private TargetEncoderPreprocessor _tePreprocessor;
@@ -42,24 +46,28 @@ public class TargetEncoding implements PreprocessingStep {
     public void prepare() {
         Frame amlTrain = _aml.getTrainingFrame();
         Set<String> teColumns = selectColumnsToEncode(amlTrain);
+        if (teColumns.isEmpty()) return;
+
+        AutoMLInput amlInput = _aml.getBuildSpec().input_spec;
+        AutoMLBuildControl amlBuild = _aml.getBuildSpec().build_control;
         TargetEncoderParameters params = new TargetEncoderParameters();
         params._train = amlTrain._key;
-        params._response_column = _aml.getBuildSpec().input_spec.response_column;
+        params._response_column = amlInput.response_column;
         params._keep_original_categorical_columns = false;
         params._blending = true;
         params._noise = 0;
-        params._seed = _aml.getBuildSpec().build_control.stopping_criteria.seed();
+        params._seed = amlBuild.stopping_criteria.seed();
         
         if (_aml.isCVEnabled()) {
             params._data_leakage_handling = DataLeakageHandlingStrategy.KFold;
-            params._fold_column = _aml.getBuildSpec().input_spec.fold_column;
+            params._fold_column = amlInput.fold_column;
             if (params._fold_column == null) {
                 //generate fold column
                 Frame train = new Frame(params.train());
                 Vec foldColumn = createFoldColumn(
                         params.train(), 
                         FoldAssignmentScheme.Modulo,
-                        _aml.getBuildSpec().build_control.nfolds,
+                        amlBuild.nfolds,
                         params._response_column,
                         params._seed
                 );
@@ -85,6 +93,8 @@ public class TargetEncoding implements PreprocessingStep {
 
     @Override
     public Completer apply(Model.Parameters params) {
+        if (_tePreprocessor == null) return NOOP;
+        
         params._preprocessors = (Key<ModelPreprocessor>[])ArrayUtils.append(params._preprocessors, _tePreprocessor._key);
         
         Frame train = new Frame(params.train());
@@ -126,20 +136,18 @@ public class TargetEncoding implements PreprocessingStep {
 
     private Set<String> selectColumnsToEncode(Frame fr) {
         Set<String> encode = new HashSet<>();
-        if (_cardinalityThreshold > 0) {
-            for (int i = 0; i < fr.names().length; i++) {
-                if (fr.vec(i).cardinality() >= _cardinalityThreshold) encode.add(fr.name(i));
-            }
-        } else {
-            encode.addAll(Arrays.asList(fr.names()));
+        for (int i = 0; i < fr.names().length; i++) {
+            if (fr.vec(i).cardinality() >= _cardinalityThreshold) encode.add(fr.name(i));
         }
-        return encode;
-    }
 
-    public static final String[] ignoredColumns(final Frame frame, final String... keep) {
-        Set<String> ignored = new HashSet(Arrays.asList(frame.names()));
-        ignored.removeAll(Arrays.asList(keep));
-        return ignored.toArray(new String[ignored.size()]);
+        AutoMLInput amlInput = _aml.getBuildSpec().input_spec;
+        List<String> nonPredictors = Arrays.asList(
+                amlInput.weights_column,
+                amlInput.fold_column,
+                amlInput.response_column
+        );
+        encode.removeAll(nonPredictors);
+        return encode;
     }
 
     TargetEncoderPreprocessor getTEPreprocessor() {
--- a/h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
@@ -10,6 +10,7 @@ import hex.Model;
 import hex.Model.Parameters.FoldAssignmentScheme;
 import hex.ModelPreprocessor;
 import water.DKV;
+import water.Iced;
 import water.Key;
 import water.fvec.Frame;
 import water.fvec.Vec;
@@ -21,13 +22,20 @@ import java.util.List;
 
 public class TargetEncoding implements PreprocessingStep {
     
-    private transient AutoML _aml;
-    private transient TargetEncoderPreprocessor _tePreprocessor;
-    private transient TargetEncoderModel _teModel;
-    private final transient List<Completer> _disposables = new ArrayList<>();
+    static String TE_FOLD_COLUMN_SUFFIX = "_te_fold";
+    
+    private AutoML _aml;
+    private TargetEncoderPreprocessor _tePreprocessor;
+    private TargetEncoderModel _teModel;
+    private final List<Completer> _disposables = new ArrayList<>();
 
     public TargetEncoding(AutoML aml) {
         _aml = aml;
+    }
+
+    @Override
+    public String getType() {
+        return PreprocessingStepDefinition.Type.TargetEncoding.name();
     }
 
     @Override
@@ -54,7 +62,7 @@ public class TargetEncoding implements PreprocessingStep {
                         params._seed
                 );
                 DKV.put(foldColumn);
-                params._fold_column = params._response_column+"_te_fold";
+                params._fold_column = params._response_column+TE_FOLD_COLUMN_SUFFIX;
                 addFoldColumn(train, params._fold_column, foldColumn, params._train.toString());
                 params._train = train._key;
                 _disposables.add(() -> {
@@ -64,7 +72,7 @@ public class TargetEncoding implements PreprocessingStep {
             }
         }
 
-        TargetEncoder te = new TargetEncoder(params);
+        TargetEncoder te = new TargetEncoder(params, _aml.makeKey(getType(), null, false));
         _teModel = te.trainModel().get();
         _tePreprocessor = new TargetEncoderPreprocessor(_teModel);
     }
@@ -79,6 +87,9 @@ public class TargetEncoding implements PreprocessingStep {
         if (addFoldColumn) {
             addFoldColumn(train, foldColumn, _teModel._parms._train.get().vec(foldColumn), params._train.toString());
             params._train = train._key;
+            params._fold_column = foldColumn;
+            params._nfolds = 0; // to avoid confusion or errors
+            params._fold_assignment = FoldAssignmentScheme.AUTO; // to avoid confusion or errors
         }
         
         return () -> {
@@ -93,7 +104,24 @@ public class TargetEncoding implements PreprocessingStep {
     public void dispose() {
         for (Completer disposable : _disposables) disposable.run();
     }
-    
+
+    @Override
+    public void remove() {
+        if (_tePreprocessor != null) {
+            _tePreprocessor.remove(true);
+            _tePreprocessor = null;
+            _teModel = null;
+        }
+    }
+
+    TargetEncoderPreprocessor getTEPreprocessor() {
+        return _tePreprocessor;
+    }
+
+    TargetEncoderModel getTEModel() {
+        return _teModel;
+    }
+
     private static void addFoldColumn(Frame fr, String name, Vec foldColumn, String keyPrefix) {
         fr.add(name, foldColumn);
         if (fr._key == null) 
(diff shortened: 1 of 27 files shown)


h2o-py/h2o/automl/autoh2o.py
```diff
@@ -67,14 +67,15 @@ def __init__(self,
                  include_algos=None,
                  exploitation_ratio=0,
                  modeling_plan=None,
+                 preprocessing=None,
                  monotone_constraints=None,
-                 algo_parameters=None,
```
sebhrusen, line 71: as we wanted this being removed in 3.32... doing it here as it's mainly cosmetics

h2o-py/tests/testdir_algos/automl/pyunit_automl_preprocessing.py
```diff
@@ -0,0 +1,42 @@
...
+
+def test_target_encoding_with_nfolds_cv():
+    ds = import_dataset()
+    aml = H2OAutoML(project_name="py_target_encoding",
+                    max_models=10,
+                    preprocessing=['targetencoding'],
+                    seed=1)
+    aml.train(y=ds.target, training_frame=ds.train, leaderboard_frame=ds.test)
+    print(aml.leaderboard)
+
```
sebhrusen, line 38: more tests to come

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Override
+    public void prepare() {
+        TargetEncoderParameters params = new TargetEncoderParameters();
+        params._train = _aml.getTrainingFrame()._key;
+        params._response_column = _aml.getBuildSpec().input_spec.response_column;
+        params._keep_original_features = false;
+        params._blending = true;
+        params._noise = 0;
```
sebhrusen, line 40: differ from what I wrote in the ticket: 
- decided to use default blending params instead so that it behaves better by default with categories with "rare" occurences.
- and disabled noise as it is considered difficult to tune correctly: we can't expect the "auto" noise to be good in most cases.

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
```diff
@@ -0,0 +1,126 @@
...
+        
+        if (_aml.isCVEnabled()) {
+            params._data_leakage_handling = DataLeakageHandlingStrategy.KFold;
+            params._fold_column = _aml.getBuildSpec().input_spec.fold_column;
+            if (params._fold_column == null) {
+                //generate fold column
+                Frame train = new Frame(params.train());
+                Vec foldColumn = createFoldColumn(
+                        params.train(), 
+                        FoldAssignmentScheme.Modulo,
```
sebhrusen, line 51: using Modulo because that's we use for CV in AutoML: hardcoded, @erin wasn't it something we wanted to change?

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
```diff
@@ -0,0 +1,179 @@
...
+
+public class TargetEncoding implements PreprocessingStep {
+    
+    static String TE_FOLD_COLUMN_SUFFIX = "_te_fold";
+    
+    private AutoML _aml;
+    private TargetEncoderPreprocessor _tePreprocessor;
+    private TargetEncoderModel _teModel;
+    private final List<Completer> _disposables = new ArrayList<>();
+    private int _cardinalityThreshold = 0;  // the minimal cardinality for a column to be TE encoded.
```
sebhrusen, line 30: @ledell this one's for you :)

h2o-automl/src/test/java/ai/h2o/automl/preprocessing/TargetEncodingTest.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+    
+    
+    @Test
+    public void test_automl_run_with_target_encoding_enabled() {
+        try {
+            Scope.enter();
+            AutoMLBuildSpec autoMLBuildSpec = new AutoMLBuildSpec();
+            Frame fr = parse_test_file("./smalldata/titanic/titanic_expanded.csv"); Scope.track(fr);
```
sebhrusen, line 177: @erin, @goldentom42 if your goal is just to try different small datasets with different thresholds, and if you have an IDE like IntelliJ, then you can just change the dataset here , run the test, it prints the leaderboard in the console...

you can also change the threshold in `TargetEncoding.java` and rerun the test.

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
ledell: @sebhrusen Thank you!

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
ledell: @sebhrusen We put this here originally (I think) because we wanted all the models to have the exact same folds (even if seed was not set).  However, I think the Stacked Ensembles should be the same (or slightly more robust even), if we use different CV folds because there may be higher variance across the base models.

I guess the real question is -- do we need the CV folds used in TE to be the same folds used in the base models?  If so, then we should keep modulo; if not, then we can remove it.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
```diff
@@ -255,6 +254,10 @@ V getFormattedValue() {
     @API(help="The list of modeling steps to be used by the AutoML engine (they may not all get executed, depending on other constraints).",
             level = API.Level.expert)
     public StepDefinitionV99[] modeling_plan;
+    
+    @API(help="The list of preprocessing steps to run. Only 'targetencoding' is currently supported.",
```
ledell, line 258: Do we want to change "targetencoding" to "target_encoding" for readability?

h2o-py/h2o/automl/autoh2o.py
```diff
@@ -117,11 +118,9 @@ def __init__(self,
         :param exploitation_ratio: The budget ratio (between 0 and 1) dedicated to the exploitation (vs exploration) phase. By default, the exploitation phase is disabled (exploitation_ratio=0) as this is still experimental; to activate it, it is recommended to try a ratio around 0.1. Note that the current exploitation phase only tries to fine-tune the best XGBoost and the best GBM found during exploration.
         :param modeling_plan: List of modeling steps to be used by the AutoML engine (they may not all get executed, depending on other constraints).
           Defaults to None (Expert usage only).
+        :param preprocessing: List of preprocessing steps to run. Only 'targetencoding' is currently supported.
```
ledell, line 121: https://github.com/h2oai/h2o-3/pull/4927/files#r491242730

h2o-r/h2o-package/R/automl.R
```diff
@@ -48,11 +48,9 @@
 #'        Defaults to NULL, which means that all appropriate H2O algorithms will be used, if the search stopping criteria allow. Optional.
 #' @param exploitation_ratio The budget ratio (between 0 and 1) dedicated to the exploitation (vs exploration) phase. By default, the exploitation phase is disabled (exploitation_ratio=0) as this is still experimental; to activate it, it is recommended to try a ratio around 0.1. Note that the current exploitation phase only tries to fine-tune the best XGBoost and the best GBM found during exploration.
 #' @param modeling_plan List. The list of modeling steps to be used by the AutoML engine (they may not all get executed, depending on other constraints). Optional (Expert usage only).
+#' @param preprocessing List. The list of preprocessing steps to run. Only 'targetencoding' is currently supported.
```
ledell, line 51: Same comment as above: https://github.com/h2oai/h2o-3/pull/4927/files#r491242730

h2o-r/h2o-package/R/automl.R
```diff
@@ -48,11 +48,9 @@
...
 #' @param monotone_constraints List. A mapping representing monotonic constraints.
 #'        Use +1 to enforce an increasing constraint and -1 to specify a decreasing constraint.
-#' @param algo_parameters List. A list of param_name=param_value to be passed to internal models. Defaults to none (Expert usage only).
```
ledell, line 53: The `algo_parameters` param seems a bit more dangerous to remove here than in Python... Since it was fully documented in R and there was nothing indicating that it was an experimental feature. Hmmm...  It was not documented in the online user guide, but most R folks just look at the R docs.

h2o-r/h2o-package/R/automl.R
```diff
@@ -100,16 +98,26 @@ h2o.automl <- function(x, y, training_frame,
...
                        export_checkpoints_dir = NULL,
-                       verbosity = "warn")
+                       verbosity = "warn",
+                       ...)
 {
+  dots <- list(...)
+  algo_parameters <- NULL  
+  for (arg in names(dots)) {
+    if (arg == 'algo_parameters') {
+      algo_parameters <- dots$algo_parameters
```
ledell, line 116: Is the goal to keep the `algo_parameters` hidden for developer/experimental use only?  (I know we have had discussions about whether it makes sense to eventually promote this to a fully supported param, or to replace it with a different algo customization approach & param...)  Maybe we should at least add a warning to let people know either they should not use this parameter and/or may be deprecated in the future?

sebhrusen: @ledell 
> Is the goal to keep the algo_parameters hidden for developer/experimental use only?

Correct.

> Should we add a warning to let people know either they should not use this parameter ?

I've never seen anything like that, the param was never documented, was disabled by default (how to enable it was also not documented), and its syntax is not trivial (only way to know how to use it would be to search in unit tests) so imo it's enough to signify that its usage is not recommended if you don't know what you're doing.
And btw, due to the mentioned constraints I can barely imagine anyone using it in production.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
sebhrusen: I think both work as it is represented as `TargetEncoding` enum on the backend, then it should accept `targetencoding`, `target_encoding`, `TargetEncoding` and probably a few more.
I'll change the value in the generated doc though.

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
sebhrusen: > do we need the CV folds used in TE to be the same folds used in the base models?

With current integration of TE, this is necessary to use the same folds: this allows to build the CV models without  introducing data leakage from the target encoded features...  quickly, it allows to behave as if for each CV model, we were training a new TE model on the cv-train data, and apply it separately then to both cv-train and cv-test for out-of-fold predictions.

goldentom42:
Hi @sebhrusen, 

Trying H2OAutoML with Target encoding failed for me using [porto seguro competition dataset](https://www.kaggle.com/c/porto-seguro-safe-driver-prediction/data). 

I built AutoML from the repo like so (just like @ledell indicated): 
```
git clone git@github.com:h2oai/h2o-3.git
cd h2o-3
git checkout seb_pubdev-7778
git pull
./gradlew clean
./gradlew -x test
./install-clients.sh
```

H2OAutoML works fine without TE : 
```
train['target'] = train['target'].asfactor()
train_cols = train.columns
# ID is useless for training
train_cols.remove('id')
if 'oof_preds' in train_cols:
    train_cols.remove('oof_preds')
    
aml = H2OAutoML(
    nfolds=4, 
    max_runtime_secs=None,
    max_runtime_secs_per_model=None,
    max_models=5,
    sort_metric='AUC',
    seed=12
)
aml.train(
    x=train_cols,
    y='target',
    training_frame=train,
    fold_column=None,
    weights_column=None,
    validation_frame=None,
    leaderboard_frame=None,
    blending_frame=None
)
```
Leaderboard displays this:
image

However using TE : 
```
aml_te = H2OAutoML(
    nfolds=4, 
    max_runtime_secs=None,
    max_runtime_secs_per_model=None,
    max_models=5,
    sort_metric='AUC',
    preprocessing=['targetencoding'],
    seed=12
)
aml_te.train(
    x=train_cols,
    y='target',
    training_frame=train,
    fold_column=None,
    weights_column=None,
    validation_frame=None,
    leaderboard_frame=None,
    blending_frame=None
)
```

I get this error : 
```
AutoML progress: |███ (failed)

08:14:08.422: XGBoost_1_AutoML_20200920_081344 [XGBoost def_1] failed: java.lang.AssertionError:  null vec: $04ff3c000000ffffffff$nfs://home/goldentom/kaggle/porto/porto_train.csv; index: 0

---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-17-faa2018cc0e4> in <module>
     16     validation_frame=None,
     17     leaderboard_frame=None,
---> 18     blending_frame=None
     19 )

~/kaggle/osis/venv/lib/python3.6/site-packages/h2o/automl/autoh2o.py in train(self, x, y, training_frame, fold_column, weights_column, validation_frame, leaderboard_frame, blending_frame)
    474         poll_updates = ft.partial(self._poll_training_updates, verbosity=self._verbosity, state={})
    475         try:
--> 476             self._job.poll(poll_updates=poll_updates)
    477         finally:
    478             poll_updates(self._job, 1)

~/kaggle/osis/venv/lib/python3.6/site-packages/h2o/job.py in poll(self, poll_updates)
     76             if (isinstance(self.job, dict)) and ("stacktrace" in list(self.job)):
     77                 raise EnvironmentError("Job with key {} failed with an exception: {}\nstacktrace: "
---> 78                                        "\n{}".format(self.job_key, self.exception, self.job["stacktrace"]))
     79             else:
     80                 raise EnvironmentError("Job with key %s failed with an exception: %s" % (self.job_key, self.exception))

OSError: Job with key $03017f00000132d4ffffffff$_bd2ba3e986dc7beb111abad9fa57477e failed with an exception: water.exceptions.H2OConcurrentModificationException: Rollups not possible, because Vec was deleted: $04ff3c000000ffffffff$nfs://home/goldentom/kaggle/porto/porto_train.csv
stacktrace: 
water.exceptions.H2OConcurrentModificationException: Rollups not possible, because Vec was deleted: $04ff3c000000ffffffff$nfs://home/goldentom/kaggle/porto/porto_train.csv
	at water.fvec.RollupStats.get(RollupStats.java:341)
	at water.fvec.RollupStats.get(RollupStats.java:371)
	at water.fvec.Vec.rollupStats(Vec.java:912)
	at water.fvec.Vec.isBinary(Vec.java:320)
	at ai.h2o.automl.modeling.XGBoostSteps.prepareModelParameters(XGBoostSteps.java:34)
	at ai.h2o.automl.modeling.XGBoostSteps$XGBoostModelStep.prepareModelParameters(XGBoostSteps.java:55)
	at ai.h2o.automl.modeling.XGBoostSteps$2.startJob(XGBoostSteps.java:125)
	at ai.h2o.automl.ModelingStepsExecutor.submit(ModelingStepsExecutor.java:77)
	at ai.h2o.automl.AutoML.learn(AutoML.java:622)
	at ai.h2o.automl.AutoML.run(AutoML.java:422)
	at ai.h2o.automl.H2OJob$1.compute2(H2OJob.java:33)
```

For info, there is no categoricals (no feature flagged as factor) In that dataset.
Will try with factors.

goldentom42:
Perfectly worked with factors. So maybe just need a check here so that we don't error.

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
goldentom42: Hopefully I won't be off topic here.
Best practice for Target Encoding is nested CV to avoid any sort of leakage and that would like this in plain python : 

```
# NOTE : categorical_list contains the list of categorical features
folds = KFold()
nested_folds = KFold()

for trn_, val_ in folds.split(train_frame):
    # Standard fold split
    trn_x, trn_y = train_frame[trn_], y[trn_]
    val_x,  val_y = train_frame[val_], y[val_]

    # Nested CV 
    cat_te_trn = np.zeros((trn_x.shape[0], len(cat_features))
    cat_te_val = np.zeros((val_x.shape[0], len(cat_features))
    for nested_trn_, nested_val_ in nested_folds.split(trn_x):
        nested_trn_x, nested_trn_y = trn_x[nested_trn_], trn_y[nested_trn_]
        nested_val_x, nested_val_y = trn_x[nested_val_], trn_y[nested_val_]
        
        for i_c, cat in enumerate(cat_features):
            model = TEModel().fit(nested_trn_x[cat], nested_trn_y)
            cat_te_trn[nested_val_, i_c] = model.transform(nested_val_x[cat])
            cat_te_val[val_, i_c] += model.transform(val_x[cat]) / folds.n_splits

     # Now you can use cat_te_tran and cat_te_val to train the model
     ...
```
Hopefully I did not get the code wrong.

goldentom42: To create `cat_te_val` you could fit a TEModel on `trn_x` directly instead of averaging btw.

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
sebhrusen: @goldentom42 if you want to change the cardinality, it's here

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
ledell, line 258:
```suggestion
    @API(help="The list of preprocessing steps to run. Only 'target_encoding' is currently supported.",
```

h2o-r/h2o-package/R/automl.R
ledell: In R, it's self documented (you're right that we didn't document it anywhere, including user guide though).  That was my only concern.  It's okay to remove it.

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
ledell: @sebhrusen Regarding the modulo folds (so that we can ensure same folds used in CV and modeling) – another option to handle this is to add an error if the user does not set the `seed` arg when TE is turned on.  If the user does set the seed, then we can ensure same folds across TE and modeling, right?  

Maybe i am not thinking about this correctly, but if a user explicitly sets the seed, and also turns on TE... they will get different folds than if they had just set seed and not turned on TE, right?  That will make runs difficult to compare (TE on/off), so that's why I am wondering if it's better to force the user to set a seed when using TE (or providing a `fold_column`)?  This would force the user to be reproducible when TE is turned on. 

Thoughts?

h2o-r/h2o-package/R/automl.R
```diff
@@ -100,16 +98,26 @@ h2o.automl <- function(x, y, training_frame,
...
                        monotone_constraints = NULL,
-                       algo_parameters = NULL,
                        keep_cross_validation_predictions = FALSE,
                        keep_cross_validation_models = FALSE,
                        keep_cross_validation_fold_assignment = FALSE,
                        sort_metric = c("AUTO", "deviance", "logloss", "MSE", "RMSE", "MAE", "RMSLE", "AUC", "AUCPR", "mean_per_class_error"),
                        export_checkpoints_dir = NULL,
-                       verbosity = "warn")
+                       verbosity = "warn",
+                       ...)
```
ledell, line 110: Adding the `...` without documenting above causes the following warning in CRAN check.  I will push a fix. 

```
* checking Rd \usage sections ... WARNING
Undocumented arguments in documentation object 'h2o.automl'
  ‘...’
```

ledell: Added here: https://github.com/h2oai/h2o-3/pull/4927/commits/87c743d8c040ba45f661105ec241d46b26cbda7f

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
ledell: @sebhrusen Did you want to change the Java parameter definition to use `"target_encoding"` instead of `"targetencoding"` too?  Or is the value hardcoded to "targetencoding" on the back-end?

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
```diff
@@ -48,6 +50,9 @@ public void prepare() {
         Frame amlTrain = _aml.getTrainingFrame();
         Set<String> teColumns = selectColumnsToEncode(amlTrain);
         if (teColumns.isEmpty()) return;
+        
+        _aml.eventLog().warn(Stage.FeatureCreation,
+                "Target Encoding integration in AutoML is in experimental stage, the models obtained with this feature can not be downloaded as MOJO for production yet.");
```
ledell: When running this from the client (this is from R), the print-out of this warning goes very wide and the AutoML status bar gets printed over the top of it (it seems)... can we have multi-line warning messages and/or force a newline before the AutoML status bar gets printed again?  Or is this caused by the fact that AutoML status bar starts, then this TE warning prints, and then it goes back to status bar (and thus we can't fix it...)?

It looks like this:
```
18:59:03.228: Target Encoding integration in AutoML is in an experimental stage, the models obtained with this feature can not yet be downloa  |======================================================================| 100%
Total AutoML runtime (secs): 3341
```

h2o-r/h2o-package/R/automl.R
ledell: Resolved.

h2o-py/h2o/automl/autoh2o.py
ledell: Resolved.

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
ledell: @sebhrusen Did you want to change the Java parameter definition to use `"target_encoding"` instead of `"targetencoding"` too?  Or are you keeping it "targetencoding" because the enum name is `"TargetEncoding"` on the back-end?  (If so, should we just change to `"TargetEncoding"` here for Java users?)

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
ledell: @sebhrusen I created a ticket here to revisit whether or not we need modulo folds (vs forcing a seed to be set): https://0xdata.atlassian.net/browse/PUBDEV-7801

h2o-automl/src/main/java/ai/h2o/automl/preprocessing/TargetEncoding.java
sebhrusen: it's a display issue with all event logs when the progress bar is on.
I can't do anything immediately, I need to reconsider this globally, for example:
1.  save the progress bar state
2. display event logs
3. restore the progress bar
Would be a fun ticket when I'm bored.

ledell: Thanks @sebhrusen I created a JIRA here: https://0xdata.atlassian.net/browse/PUBDEV-7802

ledell (approved):
This looks good for a v1.0.  Performance improvements can be addressed in a follow-up ticket: https://0xdata.atlassian.net/browse/PUBDEV-7795

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLBuildSpecV99.java
```diff
@@ -255,6 +254,10 @@ V getFormattedValue() {
     @API(help="The list of modeling steps to be used by the AutoML engine (they may not all get executed, depending on other constraints).",
             level = API.Level.expert)
     public StepDefinitionV99[] modeling_plan;
+    
+    @API(help="The list of preprocessing steps to run. Only 'target_encoding' is currently supported.",
+            level = API.Level.secondary)
+    public PreprocessingStepDefinitionV99[] preprocessing;
```
mn-mikke, line 260: @sebhrusen Is there any special reason for introducing a dedicated class for this enum? GLM uses the enum directly: https://github.com/h2oai/h2o-3/blob/a8a67c3623da87b5cd814c2ee970c6c1132a2049/h2o-algos/src/main/java/hex/schemas/GLMV3.java#L98

I will have to implement an exception for this in Sparkling Water.

sebhrusen: sorry, your link point to enum `GLMParameters.Family`, I don't see the connection with my changes here.

sebhrusen: if you mean: why did I create `ProcessingStepDefinitionV99`, it's because I will add more parameters in the future (this is just minimal first integration here) for example to configure TE, and we also have plans for more preprocessors...
So having this class now allows more flexibility and backwards compatibility in the future.

mn-mikke: @sebhrusen 

> why did I create ProcessingStepDefinitionV99

Yes, exactly. I used the `rand_family` parameter just as an example how it's done elsewhere. (Propagation of an array of enum values from a client to  the h2o backend).

> for example to configure TE
Thanks for giving me the context! It make sense to me. Just one more question... If you add some parameters to  `ProcessingStepDefinitionV99`, can i expect that type of [the `preprocessing` parameter](https://github.com/h2oai/h2o-3/pull/4927/files#diff-9f262b275056f042a5247e16d4bf59c9R27) in Python client will change from a string array to something more complex?

sebhrusen: correct, clients like Py and R, will support   a simple syntax (string) and an advanced one (dictionary): currently, when user is passing a string, it's actually already translated into a dictionary internally, so that later we will just open the dictionary syntax to the end user.

PR #4929, opened 2020-09-17
PUBDEV-7781: Clarified AUTO default value for categorical_encoding

hannah-tillman:


hannah-tillman:
fixed with https://github.com/h2oai/h2o-3/pull/4992

PR #4930, opened 2020-09-18
[PUBDEV-7706] Unignore ignored columns that are used

honzasterba:
- improves flow UX
- makes our REST API easier to use
- is backwards compatible, only eliminates error states

### DIFF
--- a/h2o-core/src/main/java/hex/ModelBuilder.java
+++ b/h2o-core/src/main/java/hex/ModelBuilder.java
@@ -1244,11 +1244,15 @@ abstract public class ModelBuilder<M extends Model<M,P,O>, P extends Model.Param
 
     // Drop explicitly dropped columns
     if( _parms._ignored_columns != null ) {
-      _train.remove(_parms._ignored_columns);
-      if( expensive ) Log.info("Dropping ignored columns: "+Arrays.toString(_parms._ignored_columns));
+      Set<String> ignoreColumnSet = new HashSet<>(Arrays.asList(_parms._ignored_columns));
+      Collection<String> usedColumns = _parms.getUsedColumns(tr._names);
+      ignoreColumnSet.removeAll(usedColumns);
+      String[] actualIgnoredColumns = ignoreColumnSet.toArray(new String[0]);
+      _train.remove(actualIgnoredColumns);
+      if (expensive) Log.info("Dropping ignored columns: " + Arrays.toString(actualIgnoredColumns));
     }
 
-    if(_parms._checkpoint != null){
+    if(_parms._checkpoint != null) {
       if(DKV.get(_parms._checkpoint) == null){
           error("_checkpoint", "Checkpoint has to point to existing model!");
       }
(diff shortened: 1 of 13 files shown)


h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1243,8 +1243,16 @@ public void init(boolean expensive) {
...
     if( _parms._ignored_columns != null ) {
-      _train.remove(_parms._ignored_columns);
-      if( expensive ) Log.info("Dropping ignored columns: "+Arrays.toString(_parms._ignored_columns));
+      Set<String> ignoreColumnSet = new HashSet<>(Arrays.asList(_parms._ignored_columns));
+      if (_parms._response_column != null) ignoreColumnSet.remove(_parms._response_column);
+      if (_parms._fold_column != null) ignoreColumnSet.remove(_parms._fold_column);
+      if (_parms._offset_column != null) ignoreColumnSet.remove(_parms._offset_column);
+      if (_parms._weights_column != null) ignoreColumnSet.remove(_parms._weights_column);
+      Collection<String> otherUsedColumns = getAdditionalUsedColumns();
+      for (String otherUsedColumn : otherUsedColumns) ignoreColumnSet.remove(otherUsedColumn);
```
satai, line 1252: removeAll

michalkurka:
I don't like this approach personally but I understand why it is done.

The main issue I have with this is that this approach is really a hit-and-miss. People need to remember to implement the getAdditionalUsedColumns and they need to do it correctly (which is hard to test IMHO).

What will happen when I have a column used eg. in GLM interactions and ignored in feature columns?

h2o-core/src/main/java/hex/ModelBuilder.java
honzasterba: fixed

honzasterba:
> I don't like this approach personally but I understand why it is done.

this basically duplicates what is done in Py and R clients, I dont like the code duplication as well, but its better solution than replicate the logic one more time in Flow

> The main issue I have with this is that this approach is really a hit-and-miss. People need to remember to implement the getAdditionalUsedColumns and they need to do it correctly (which is hard to test IMHO).

well not exactly, if you dont, only Flow will be a little harder to use, which is easy to fix

> What will happen when I have a column used eg. in GLM interactions and ignored in feature columns?

I am not sure if interactions are the same case as coxph(start,stop) columns, but if they are user will just get an error that they are ignoring a column that is necessary

h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -526,13 +526,6 @@ def _requires_training_frame(self):
         """
         return True
 
-    def _additional_used_columns(self, parms):
```
honzasterba, line 529: simmilar cleanup could be done in R client

koniecsveta:
LGTM - will you clean up also R client in this PR?

honzasterba:
> LGTM - will you clean up also R client in this PR?

probably yes, if this approach is approved by @michalkurka

wendycwong (approved):
LGTM.  Hope Michalk will turn around too.

michalkurka (approved):
Thank you!

PR #4931, opened 2020-09-18
PUBDEV-7720 deep clone GLM model metrics with RF model and RF input f…

koniecsveta:
…rame

this is also essential for RF benchmarks to not to fail

### DIFF
--- a/h2o-algos/src/main/java/hex/rulefit/RuleFit.java
+++ b/h2o-algos/src/main/java/hex/rulefit/RuleFit.java
@@ -258,11 +258,15 @@ public class RuleFit extends ModelBuilder<RuleFitModel, RuleFitModel.RuleFitPara
             }
         }
         
-        void fillModelMetrics(RuleFitModel model, GLMModel glmModel){
+        void fillModelMetrics(RuleFitModel model, GLMModel glmModel) {
             model._output._validation_metrics = glmModel._output._validation_metrics;
             model._output._training_metrics = glmModel._output._training_metrics;
             model._output._cross_validation_metrics = glmModel._output._cross_validation_metrics;
             model._output._cross_validation_metrics_summary = glmModel._output._cross_validation_metrics_summary;
+
+            for (Key<ModelMetrics> modelMetricsKey : glmModel._output.getModelMetrics()) {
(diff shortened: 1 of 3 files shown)


koniecsveta:
> If RuleFit model has a separate lifetime than the original model, then this makes sense.

it's needed for h2o.performance(rf_model, newdata = rf_frame) to work, because 

```
res <- .h2o.__remoteSend(method = "POST", .h2o.__MODEL_METRICS(model@model_id, 
            newdata.id), .params = parms)

```
returns model metric keys which are filtered by rf_model id and rf_frame id (in ModelMetricsHandler.java).

If I want to have GLMs metrics exposed as RF metrics, I need to place them into RF output with RF model key and RF frame key, otherwise .h2o.__remoteSend() result will be empty and  h2o.performance() will fail because of filtering empty list and trying to access its element [[1]]:

```
model_metrics <- Filter(function(mm) {
            mm$frame$name == newdata.id
        }, res$model_metrics)[[1]] 
```

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
```diff
@@ -258,11 +258,15 @@ public void computeImpl() {
...
         
-        void fillModelMetrics(RuleFitModel model, GLMModel glmModel){
+        void fillModelMetrics(RuleFitModel model, GLMModel glmModel) {
             model._output._validation_metrics = glmModel._output._validation_metrics;
             model._output._training_metrics = glmModel._output._training_metrics;
             model._output._cross_validation_metrics = glmModel._output._cross_validation_metrics;
             model._output._cross_validation_metrics_summary = glmModel._output._cross_validation_metrics_summary;
+
+            for (Key<ModelMetrics> modelMetricsKey : glmModel._output.getModelMetrics()) {
+                model.addModelMetrics(modelMetricsKey.get().deepCloneWithDifferentModelAndFrame(model, model._parms._train.get()));
```
honzasterba, line 268: calling _tain.get() every time is probably not very efficient

koniecsveta: fixed b20af84

PR #4932, opened 2020-09-18
PUBDEV-7796: Model explainability interface for H2O models and AutoML objects in R/Python

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7796

WIP: Near final version, but still making edits

## Explainability API 

Documenting the new methods (for feedback / naming):

Python methods in the `h2o.explain` module:

These take a single model or (including an AutoML object or leaderboard slice) as input:
```
explain        
explain_row        
```

Takes a list of models (including an AutoML object or leaderboard slice) as input:
```
varimp_heatmap          
model_correlation_heatmap        
pd_multi_plot
```

Takes a single model as input:
```
residual_analysis_plot
shap_explain_row_plot
shap_summary_plot
pd_plot
ice_plot
```

R has the same functions names, but with the `h2o.*` prefix.


h2o-r/h2o-DESCRIPTION.template
```diff
@@ -57,7 +57,12 @@ Suggests: ggplot2,
           data.table (>= 1.9.8),
           rgl (>= 0.100.19),
           plot3Drgl (>= 1.0.1),
-          survival
+          survival,
+          DT,
```
tomasfryda, line 61: `DT`, `htmltools`, `plotly` are required for html output of explanation.
`IRdisplay`, `repr` is required for html rendering in Jupyter notebook with R kernel. (Those packages are dependencies of R jupyter kernel).

Without adding them to `Suggests`, R check as CRAN would complain.

h2o-r/h2o-DESCRIPTION.template
```diff
@@ -103,3 +108,4 @@ Collate:
          'zzz.R'
```
tomasfryda, line 108: I'm wondering if the `zzz.R` shouldn't be the last file in the list.

sebhrusen:
@tomasfryda 2 new files of ~2000 lines that should be reviewed at once? any trick to give us some courage?

h2o-py/h2o/mli/_mli.py
```diff
@@ -0,0 +1,1738 @@
...
+
+
+class NumpyFrame:
+    """
+    Simple class that very vaguely emulates Pandas DataFrame.
+    Main purpose is to keep parsing from the List of Lists format to numpy.
+    This class is meant to be used just in the MLI context.
+    Due to that fact it encodes the factor variables similarly to R/pandas -
+    factors are mapped to numeric column which in turn makes it easier to plot it.
+    """
```
sebhrusen, line 140: yet another dataframe format?
Wondering if it's worth introducing this when 9x% of our users (and 100% of conda users) will have pandas in their Python environment.

tomasfryda: Yes, I agree. I would be much happier using pandas instead of my "numpy frame".

tomasfryda:
@sebhrusen I don't know if it helps but significant part of that 2k loc are just comments. Other thing that might help is that the code is not very complex, after all it's just plotting code consisting of getting the data to the right shape and plotting it. 

Python code is slightly more complicated since I couldn't use pandas and scipy (so there is a very simple implementation of density estimation - used just for the jitter in the SHAP summary plot, hierarchical clustering, and the NumpyFrame).

ledell:
@navdeep-G @pramitchoudhary Can you take a look at the descriptions here and suggest/commit any changes/additions? https://github.com/h2oai/h2o-3/blob/e6178699f2dfbed20ea370c54d37e606d9b2f3a1/h2o-py/h2o/mli/_mli.py#L69

h2o-r/h2o-DESCRIPTION.template
ledell: Yeah, let's move it to the end (it used to be there).

h2o-r/h2o-DESCRIPTION.template
```diff
@@ -103,3 +108,5 @@ Collate:
...
          'tf-idf.R'
          'rulefit.R'
+         'mli.R'
```
ledell, line 111: Can we name this file, `explain.R` so it's more obvious what it contains?  I don't think the "MLI" acronym is universally known/used and this code is focused more on explainability vs interpretability.

h2o-r/scripts/h2o-r-test-setup.R
```diff
@@ -186,7 +186,7 @@ function() {
                     "kvstore.R", "frame.R", "targetencoder.R", "astfun.R","automl.R", "import.R", "parse.R", "export.R", "models.R", "edicts.R",
                     "coxph.R", "coxphutils.R", "gbm.R", "glm.R", "gam.R", "glrm.R", "kmeans.R", "deeplearning.R", "randomforest.R", "generic.R",
                     "naivebayes.R", "pca.R", "svd.R", "locate.R", "grid.R", "word2vec.R", "w2vutils.R", "stackedensemble.R", "rulefit.R",
-                    "predict.R", "xgboost.R", "isolationforest.R", "psvm.R", "segment.R", "tf-idf.R")
+                    "predict.R", "xgboost.R", "isolationforest.R", "psvm.R", "segment.R", "tf-idf.R", "mli.R")
```
ledell, line 189: Also change to `"explain.R"`.

h2o-r/h2o-package/R/mli.R
```diff
@@ -0,0 +1,2209 @@
...
+#' @return List of outputs with class "explanation"
+#' @export
+h2o.explain <- function(object,
+                        test_frame,
+                        columns_of_interrest = NULL,
+                        include_explanations = "ALL",
+                        exclude_explanations = character(),
+                        top_n_features = 5,
+                        best_of_family = TRUE,
+                        user_overrides = list()) {
```
ledell, line 1787: If this is just for plot parameter overrides, then maybe we should use a parameter name like `plot_overrides` or `plot_arguments`?

h2o-r/h2o-package/R/mli.R
```diff
@@ -0,0 +1,2209 @@
...
+#' Generate explanations for \code{object} on \code{test_frame}.
+#'
+#' @param object One of the following: an H2OAutoML, an H2OAutoML Leaderboard slice, a model, a list of models.
+#' @param test_frame An H2OFrame.
+#' @param columns_of_interrest A vector of feature names. If specified, plot just these columns.
+#' @param include_explanations If specified, do only the specified explanations.
+#'   (Mutually exclusive with exclude_explanations)
+#' @param exclude_explanations Exclude specified explanations.
+#' @param top_n_features If \code{columns_of_interrest} is missing, use top n features.
+#' @param best_of_family If True, explain only best of family models
```
ledell, line 1774: @tomasfryda Can you explain this parameter a bit more -- can you fill in the missing pieces:  "If True, explain only the best model from each algorithm class; if False ... "

h2o-r/h2o-package/R/mli.R
```diff
@@ -0,0 +1,2209 @@
...
+  })
+}
+
+######################################## Explain ###################################################
+
+#' Generate explanations for \code{object} on \code{test_frame}.
+#'
+#' @param object One of the following: an H2OAutoML, an H2OAutoML Leaderboard slice, a model, a list of models.
+#' @param test_frame An H2OFrame.
+#' @param columns_of_interrest A vector of feature names. If specified, plot just these columns.
```
ledell, line 1769:
```suggestion
#' @param columns_of_interest A vector of feature names. If specified, plot just these columns.
```

ledell: This is spelled wrong, and will need to be changed elsewhere too.

h2o-r/h2o-package/R/mli.R
```diff
@@ -0,0 +1,2209 @@
...
+#' @param include_explanations If specified, do only the specified explanations.
+#'   (Mutually exclusive with exclude_explanations)
+#' @param exclude_explanations Exclude specified explanations.
+#' @param top_n_features If \code{columns_of_interrest} is missing, use top n features.
```
ledell, line 1773:
```suggestion
#' @param top_n_features If \code{columns_of_interest} is missing, create plots only with the top n columns (where applicable).  Defaults to 5.
```

h2o-r/h2o-package/R/mli.R
```diff
@@ -0,0 +1,2209 @@
...
+######################################## Explain ###################################################
+
+#' Generate explanations for \code{object} on \code{test_frame}.
+#'
+#' @param object One of the following: an H2OAutoML, an H2OAutoML Leaderboard slice, a model, a list of models.
+#' @param test_frame An H2OFrame.
+#' @param columns_of_interrest A vector of feature names. If specified, plot just these columns.
+#' @param include_explanations If specified, do only the specified explanations.
+#'   (Mutually exclusive with exclude_explanations)
+#' @param exclude_explanations Exclude specified explanations.
```
ledell, line 1772:
```suggestion
#' @param exclude_explanations Exclude specified model explanations.
```

h2o-r/h2o-package/R/mli.R
```diff
@@ -0,0 +1,2209 @@
...
+}
+
+######################################## Explain ###################################################
+
+#' Generate explanations for \code{object} on \code{test_frame}.
+#'
+#' @param object One of the following: an H2OAutoML, an H2OAutoML Leaderboard slice, a model, a list of models.
+#' @param test_frame An H2OFrame.
+#' @param columns_of_interrest A vector of feature names. If specified, plot just these columns.
+#' @param include_explanations If specified, do only the specified explanations.
```
ledell, line 1770:
```suggestion
#' @param include_explanations If specified, return only the specified model explanations.
```

ledell: Will need to be changed elsewhere too.

h2o-r/h2o-package/R/mli.R
ledell: Will need to be changed elsewhere too.

h2o-r/h2o-DESCRIPTION.template
ledell: It's ok to add them to Suggests.

h2o-r/h2o-DESCRIPTION.template
ledell: Yes, let's move it there, thanks!

h2o-py/h2o/mli/_mli.py
```diff
@@ -0,0 +1,1738 @@
...
+    yield
+    if progress:
+        h2o.show_progress()
+
+
+class NumpyFrame:
+    """
+    Simple class that very vaguely emulates Pandas DataFrame.
+    Main purpose is to keep parsing from the List of Lists format to numpy.
+    This class is meant to be used just in the MLI context.
```
ledell, line 142: I would refrain from using the term, "MLI", since many people don't know that and we are focused on "explainability" right now.  I wonder if we should rename these modules?  `./explain/explain.py`?

ledell:
If we rename file objects from "mli" to "explain",

and could we have aliases so that this:
```
from h2o.explain import explain, explain_row
```
could be written as :

```
from h2o import explain, explain_row
```
or could that not work?

h2o-r/h2o-package/R/mli.R
tomasfryda: will change to `plot_overrides`

h2o-r/h2o-package/R/mli.R
tomasfryda: thanks!

h2o-r/h2o-package/R/mli.R
tomasfryda: That parameter now applies only to the PDP plot - by default it's showing just the best of family. 
I wonder if I should remove the parameter from the `h2o.explain` - it would still be overridable by the `plot_overrides`.
Or I could use just best of family for the rest of the "multi-model" explanations - model correlation heat map and variable importance heat map - now it uses top 20 models which seems to me to be better so I would probably go for the first option.

What do you think @ledell ?

tomasfryda:
> If we rename file objects from "mli" to "explain",
> 
> could we have aliases so that this:
> 
> ```
> from h2o.explain import explain, explain_row
> ```
> 
> could be written as :
> 
> ```
> from h2o import explain, explain_row
> ```
> 
> or could that not work?

@ledell Renamed it to explain. And `from h2o import explain, explain_row` works as well as `h2o.explain_row(...)` and of course the longer version works too (e.g., `h2o.explain.explain_row(...)`)

h2o-r/h2o-package/R/mli.R
tomasfryda: > I wonder if I should remove the parameter from the `h2o.explain` - it would still be overridable by the `plot_overrides`.

Just noticed that in Python version I don't expose this parameter in `h2o.explain` (but the user can still choose to show all models by using the `plot_overrides`)

tomasfryda: I removed this parameter and made it smart as is in python: 
When explaining AutoML, show just best of family to get some general overview, but when explaining a list of models assume that the user actually handpicked the models and actually wants to see them all.

This is now a default behaviour but it is still overridable by using the `plot_overrides`.

h2o-py/h2o/__init__.py
```diff
@@ -61,12 +61,31 @@ def readTxtFromWhl(name, fallback):
...
+        "explain",
+        "explain_row",
+        "variable_importance_heatmap",
+        "model_correlation_heatmap",
+        "partial_dependences"
+    ]
+
+    register_explain_methods()
+except ImportError:
+    pass
```
sebhrusen, line 90: are you using a try-except because `explain` module imports optional dependencies?
If so, ar those optional dependencies used by all methods you want to expose? I mean, if possible, I'd rather do imports  for optional dependencies inside the methods.

h2o-py/h2o/explain/_explain.py
```diff
@@ -0,0 +1,1763 @@
...
+        self.content = content
+        self.level = level
+
+    def _repr_html_(self):
+        return "<h{level}>{content}</h{level}>".format(level=self.level, content=self.content)
+
+    def _repr_markdown_(self):
+        return "\n\n{} {}".format("#" * self.level, self.content)
+
+    def __repr__(self):
```
sebhrusen, line 66: usually, better implement `__str__` for printing and keep `__repr__` for a parseable representation.

h2o-py/h2o/__init__.py
tomasfryda: @sebhrusen Well everything that I expose outside the module uses `matplotlib` and `matplotlib` is optional. I can move it inside the methods if you still think it would be better.

h2o-py/h2o/explain/_explain.py
tomasfryda: Good point! I must have confused myself with the IPython's _repr methods.

h2o-py/h2o/__init__.py
sebhrusen: if all methods are concerned, then your implementation makes sense.
Just wanted to be sure that we're not preventing users from accessing features that would still work in their env.

h2o-py/h2o/__init__.py
```diff
@@ -61,12 +61,31 @@ def readTxtFromWhl(name, fallback):
...
+    # Export explain functions that are useful for lists of models
+    from h2o.explain import explain, explain_row, variable_importance_heatmap, model_correlation_heatmap, \
+        partial_dependences, register_explain_methods
+
+    __all__ += [
+        "explain",
+        "explain_row",
+        "variable_importance_heatmap",
+        "model_correlation_heatmap",
+        "partial_dependences"
```
pramitchoudhary, line 85: `partial_dependences` Is it not ok to rename this to `partial_dependence_plots`?

ledell: @pramitchoudhary I thought the same thing at first... but it's actually just a single plot with multiple lines and not multiple "plots".  The current name, conceptually, makes sense -- my only concern is that people will get confused about whether it's "dependences" or "dependencies".  Also good to note that we have an existing function called `.partial_plot()` which takes a single model and a column (or columns) to generate a PD plot.

h2o-py/h2o/explain/_explain.py
```diff
@@ -0,0 +1,1763 @@
...
+        columns_of_interest=None,  # type: Optional[List[str]]
+        top_n_features=5,  # type: int
+        include_explanations="ALL",  # type: Union[str, List[str]]
+        exclude_explanations=[],  # type: Union[str, List[str]]
+        plot_overrides=dict(),  # type: Dict
+        figsize=(16, 9),  # type: Tuple[float]
+        render=True,  # type: bool
+        qualitative_colormap="Dark2",  # type: str
+        sequential_colormap="RdYlBu_r"  # type: str
+):
```
sebhrusen, line 1446: - `columns_of_interest`: isn't it a bit verbose? why not just `columns`?
- `top_n_features`: serves same purpose as previous param, but we use "columns" in one and "features" in the other. Besides, they're mutually exclusive. Can't they be merged into one param? `features=5` or `features=['col_1', 'col_2']`?
- `plot_overrides`: in matplotlib world, don't we just usually use a `kwargs` to forward  those?
- `plot_overrides`, `exclude_explanations`: careful with mutable defaults (here {}, []), I always need to check that this param is not modified in the impl.
- `include_explanations`, `exclude_explanations`: just an idea, as those explanations are predefined, what aout a single  `explanations` param accepting None (=ALL) or instances of:
```
class Explanations():
     def __init__(self, leaderboard=None, ....):
            # all params default to None
            # if all set to None -> include all
            # otherwise, includes only the ones with value True, or excludes the ones with value False.

#usage
explanations=Explanations(leaderboard=True, model_correlation_heatmap=True)
explanations=Explanations(model_correlation_heatmap=False)
```
Benefits: discoverability+autocompletion for the possible explanations, include+exclude in one param.

sebhrusen: same for `explain_row`

h2o-py/h2o/__init__.py
pramitchoudhary: Got it. Then why not just call it `partial_dependence_plot` so that its clear one is not getting the values but the actual plot and if one is returning the values then we call it `partial_dependence`?

Also, if the `explain` functions are coming from a different namespace then is it not ok to call it `partial_plot()` for consistency?

h2o-py/h2o/__init__.py
```diff
@@ -61,12 +61,31 @@ def readTxtFromWhl(name, fallback):
...
+
+try:
+    # Export explain functions that are useful for lists of models
+    from h2o.explain import explain, explain_row, variable_importance_heatmap, model_correlation_heatmap, \
+        partial_dependences, register_explain_methods
+
+    __all__ += [
+        "explain",
+        "explain_row",
+        "variable_importance_heatmap",
```
pramitchoudhary, line 83: I don't have a strong opinion at this point but thinking from the top  of my mind,
do we need to name function based on the viz it's generating rather than just a `type` param?
For e.g. `variable_importance(type='heatmap', ...)`. 
That way if a single model is passed then it generates a `bar` plot and if a list of models is passed then it generates a heatmap.

If it's too much effort right now, we make modifications for future releases.

h2o-py/h2o/explain/_explain.py
tomasfryda: > * `columns_of_interest`: isn't it a bit verbose? why not just `columns`?
> * `top_n_features`: serves same purpose as previous param, but we use "columns" in one and "features" in the other. Besides, they're mutually exclusive. Can't they be merged into one param? `features=5` or `features=['col_1', 'col_2']`?

we should probably keep the names same in both python and R, right? I can change the behaviour in both. What do you think @ledell?


> * `plot_overrides`: in matplotlib world, don't we just usually use a `kwargs` to forward  those?

this used to be called user overrides, it is used to change parameters such as how many features to plot in one particular type of plot so it's not really the thing that I would use `kwargs` (e.g., line type, color etc)

> * `plot_overrides`, `exclude_explanations`: careful with mutable defaults (here {}, []), I always need to check that this param is not modified in the impl.
 
yep, both are read-only

> * `include_explanations`, `exclude_explanations`: just an idea, as those explanations are predefined, what about a single  `explanations` param accepting None (=ALL) or instances of:
> 
> ```
> class Explanations():
>      def __init__(self, leaderboard=None, ....):
>             # all params default to None
>             # if all set to None -> include all
>             # otherwise, includes only the ones with value True, or excludes the ones with value False.
> 
> #usage
> explanations=Explanations(leaderboard=True, model_correlation_heatmap=True)
> explanations=Explanations(model_correlation_heatmap=False)
> ```
> 
> Benefits: discoverability+autocompletion for the possible explanations, include+exclude in one param.

I like this idea, not sure if I can or should do similar thing in R. Also I'm not sure if we want the method/function arguments differ that much between R and python. Any thoughts @ledell ?

h2o-py/h2o/__init__.py
tomasfryda: I think using `partial_plot()` could be confusing. (When using jupyter, I often found myself googling the function (just the function name not fully qualified name) instead of waiting for the jupyter's documentation to pop up.)

pramitchoudhary: `pdp_plot`?

h2o-py/h2o/__init__.py
tomasfryda: I like the idea, that way we could also do a box plot for multiple models with the same API etc.

I would probably keep it the way it is right now (not much time before release) and in future make a function/method `variable_importance` that would dispatch depending on the input if that sounds reasonably.

pramitchoudhary:
> @navdeep-G @pramitchoudhary Can you take a look at the descriptions here and suggest/commit any changes/additions?
> 
> https://github.com/h2oai/h2o-3/blob/e6178699f2dfbed20ea370c54d37e606d9b2f3a1/h2o-py/h2o/mli/_mli.py#L69

@ledell the description looks good to me.
Just minor stuff,
1. 
`pdp="Partial dependence plot (PDP)`. Should we add the keyword `one-way` explicitly to the description? to indicate we are talking about One-way PDPs. And use the word global explanation?

2. `ice_row`: will it possible to use the word local explanation?

h2o-py/h2o/__init__.py
pramitchoudhary: sounds good.

sebhrusen: if `variable_importance_heatmap` is the only one supported currently, you can still rename it to `variable_importance` and implement the dispatch later, right? This would prevent you from having to implement deprecations later..

navdeep-G (approved):
I cannot comment on the API too much (I will leave that to the rest of you), but in terms of functionality, I approve.

h2o-py/h2o/explain/_explain.py
tomasfryda: @ledell 
- Changed the `columns_of_interest` and `top_n_features` to one parameter `features` with behaviour as Seb described above in both Python and R.

- In python changed the implementation of explanation to use Seb's idea. Usage:
```
import h2o
from h2o.explain import Explanations, RowExplantations

aml.explain(test, explanations=Explanations(leaderboard=False))
aml.explain_row(test, 1, explanations=RowExplanations(ice=True))
```

In R, it stays the same as before (include_explanations/exclude_explanations). I can easily revert this commit or change somehow R's behaviour.

ledell: Let's change the name `features` to `columns` for consistency with other H2O functions.  We can keep the support for integers (to trigger top N) if that seems better, but I am also not opposed to having an extra arg for the latter.

ledell: @tomasfryda @sebhrusen Regarding changing the format of the explanation args in Python and converting into a single `explanations` parameter -- it sounds like the main benefits are:

- remove a parameter
- autocompletion

It's a handy way of storing that information, but I also see some drawbacks:

- Inconsistency between R and Python (adds complexity / duplicates documentation).
- User has two extra imports
- This is not an established way of specifying params in H2O Python module (it could be in the future, but we don't have anything like that that we know of, so far, so it might seem a bit too unique to some folks...?)
- It requires bout 2x the amount of typing because now you have to put True or False in addition to the name.

I honestly don't love the whole `include_*` `exclude_*` thing, but it's got some "existing conventions" weight to it.

tomasfryda: @ledell seems like you'd prefer the include/exclude thing. I dropped the commit for now but I'm keeping the patch handy in case we would like to use the `Explanation` object after all...

ledell:
Here's some code snippets that can be used to try out the interface, if anyone is interested.  cc  @navdeep-G @pramitchoudhary 


### R 

```
        library(h2o)
        h2o.init()

        y <- "response"
        x <- setdiff(names(train), y)

        train[, y] <- as.factor(train[, y])
        test[, y] <- as.factor(test[, y])

        # Run AutoML
        aml <- h2o.automl(x = x, y = y, 
                          training_frame = train,
                          max_models = 10,
                          seed = 1)

        # Explain leader model & compare with all AutoML models                  
        exa <- h2o.explain(aml, test)
        exa

        # Explain a single H2O model (e.g. leader model from AutoML)
        exm <- h2o.explain(aml@leader, test)
        exm
```

### Python:

```
        import h2o
        from h2o.explain import explain, explain_row

        h2o.init()

        train = h2o.import_file("https://s3.amazonaws.com/erin-data/higgs/higgs_train_10k.csv")

        x = train.columns
        y = "response"
        x.remove(y)

        train[y] = train[y].asfactor()
        test[y] = test[y].asfactor()
        
        # Run AutoML
        aml = H2OAutoML(max_models=10, seed=1)
        aml.train(x=x, y=y, training_frame=train)

        # Explain leader model & compare with all AutoML models 
        exa = aml.explain(test)

        # Explain a single H2O model (e.g. leader model from AutoML)
        exm = aml.leader.explain(test)
```

scripts/jenkins/jenkinsfiles/Jenkinsfile
```diff
@@ -47,7 +47,7 @@ try {
               // clear the folder
               deleteDir()
               // checkout H2O-3
-              retryWithTimeout(60, 3) {
+              retryWithTimeout(120, 3) {
```
tomasfryda, line 50: Revert before merging

h2o-py/h2o/__init__.py
ledell: We decided to break this into two functions, one that takes a single model, `h2o.pd_plot()` and one that takes multiple models (as a list or AutoML object), `h2o.pd_multi_plot()`.

ledell:
Ready to merge, waiting for tests.

ledell (approved):

PR #4935, opened 2020-09-21
[core] remove noop arg nochecks

honzasterba:


Pscheidl (approved):
We could get rid of that constructor once and for all, yet killing only a fraction of the persian army is also a "good start" :wink: Thanks.

michalkurka:
I think this is a good change because the usages of the special no-check constructor don't actually need to skip the checks for perf reasons and some of them are outright wrong.

michalkurka (approved):

PR #4937, opened 2020-09-21
Remove deprecated matplotlib.use argument

tomasfryda:
`matplotlib.use` `warn` argument was deprecated in Matplotlib 3.1 and is removed in the current version (Matplotlib 3.3).

Just deleting it shouldn't change behavior (other than being able to use the newer matplotlib versions).

See changes in matplotlib's code:
Current version: https://matplotlib.org/_modules/matplotlib.html#use
Older version: https://matplotlib.org/3.2.1/_modules/matplotlib.html#use

### DIFF
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -3776,7 +3776,7 @@ class H2OFrame(Keyed):
             try:
                 import matplotlib
                 if server:
-                    matplotlib.use("Agg", warn=False)
+                    matplotlib.use("Agg")
                 import matplotlib.pyplot as plt
             except ImportError:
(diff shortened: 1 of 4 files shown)


ledell (approved):
Looks good, thanks @tomasfryda!

sebhrusen (approved):
for the irony of it: was it printing a warning message because the`warn` parameter, that we use to disable warnings, is deprecated? :)

tomasfryda:
> for the irony of it: was it printing a warning message because the`warn` parameter, that we use to disable warnings, is deprecated? :)

@sebhrusen  yes, at least for matplotlib  3.2.2 it was.

PR #4938, opened 2020-09-21
PUBDEV-7720 add accessor method for grabbing the rule importance info…

koniecsveta:
…rmation

### DIFF
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -1713,7 +1713,15 @@ class ModelBase(h2o_meta(Keyed)):
         """DEPRECATED. Use :meth:`scoring_history` instead."""
         return self.scoring_history()
 
+    def rule_importance(self):
+        """
+        Retrieve rule importances for a Rulefit model
 
+        :return: H2OTwoDimTable
+        """
+        if self._model_json["algo"] != "rulefit":
+            raise H2OValueError("This function is available for Rulefit models only")
+        return self._model_json["output"]['rule_importance']
 
 
 
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -1713,15 +1713,6 @@ class ModelBase(h2o_meta(Keyed)):
         """DEPRECATED. Use :meth:`scoring_history` instead."""
         return self.scoring_history()
 
-    def rule_importance(self):
-        """
-        Retrieve rule importances for a Rulefit model
-
-        :return: H2OTwoDimTable
-        """
-        if self._model_json["algo"] != "rulefit":
-            raise H2OValueError("This function is available for Rulefit models only")
-        return self._model_json["output"]['rule_importance']
 
 
 
--- a/h2o-r/h2o-package/R/models.R
+++ b/h2o-r/h2o-package/R/models.R
@@ -5054,20 +5054,3 @@ h2o.reset_threshold <- function(object, threshold) {
     return(NULL)
   }
 }
-
-#'
-#' Retrieve rule importances for a Rulefit model
-#'
-#' @param object An \linkS4class{H2OTable} object.
-#' @export
-h2o.get_rule_importance <- function(object) {
-    o <- object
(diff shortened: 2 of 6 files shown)


h2o-r/h2o-package/R/models.R
```diff
@@ -5054,3 +5054,20 @@ h2o.reset_threshold <- function(object, threshold) {
     return(NULL)
   }
 }
+
+#'
+#' Retrieve rule importances for a Rulefit model
+#'
+#' @param object An \linkS4class{H2OTable} object.
+#' @export
+h2o.get_rule_importance <- function(object) {
```
Pscheidl, line 5063: Is the naming in line with the "universe" conventions we should follow ? snake case is fine I guess, but what about the `get` prefix ? @sebhrusen is the API guy.

koniecsveta: yesyes I put this into api changes channel and waiting for response there. Did this in accordance with ntrees_actual / h2o.get_ntrees_actual but may change according to response.

sebhrusen: yes, to make code more portable between Py and R, we're now trying to use snake case in both languages.

koniecsveta: @sebhrusen  and is it valid, that in py it is named rule_importance and in R there is get prefix : get_rule_importance ? or should it be the same?

sebhrusen: sorry, I read Pavel's question too quickly.
Ideally, for R properties, we shouldn't need any getter... I don't think we provide a getter for other model slots in R, do we? Do we do it here mainly to add the warning message?

sebhrusen: we created a hook for `stackedensemble` that allows to customize models for a specific algo:
see `kvstore.R`, line 266.
I think this could be used to add a slot to the RuleFit model only.

sebhrusen: otherwise, I don't see the benefit of having a new accessor in R, as it doesn't seem to add much value compared with the existing slot

h2o-py/h2o/model/model_base.py
```diff
@@ -1713,7 +1713,15 @@ def score_history(self):
         """DEPRECATED. Use :meth:`scoring_history` instead."""
         return self.scoring_history()
 
+    def rule_importance(self):
```
honzasterba, line 1716: do we need to put this here? I think for python it should be possible to add only the RuleFitEstimator class via custom snippet

h2o-py/h2o/model/model_base.py
```diff
@@ -1713,7 +1713,15 @@ def score_history(self):
...
+        """
+        Retrieve rule importances for a Rulefit model
 
+        :return: H2OTwoDimTable
+        """
+        if self._model_json["algo"] != "rulefit":
+            raise H2OValueError("This function is available for Rulefit models only")
+        return self._model_json["output"]['rule_importance']
```
sebhrusen, line 1724: if it should be made available only for Rulefit, it's much better to add this accessor in `h2o-bindings/bin/custom/python/gen_rulefit.py`:

```
def class_extensions():
    def rule_importance(self):
        #logic goes here
extensions = dict(
    __class__=class_extensions
)
```

koniecsveta: will fix

koniecsveta: e0efd1a

h2o-py/h2o/model/model_base.py
koniecsveta: e0efd1a

honzasterba (approved):
lgtm

would also like to see R tests showing how this is accessed in R

sebhrusen (approved):

koniecsveta:
> lgtm
> 
> would also like to see R tests showing how this is accessed in R

it is here: h2o-r/tests/testdir_algos/rulefit/runit_rulefit_titanic.R

PR #4940, opened 2020-09-21
PUBDEV-7791: Use hostname as a certificate alias to lookup machine-specific certificate

michalkurka:


h2o-core/src/main/java/water/init/HostnameGuesser.java
```diff
@@ -369,4 +369,18 @@ private HostnameGuessingException(Exception e) {
...
+    String hostname = address.getHostName();
+    if (! address.getHostAddress().equals(hostname)) {
+      return hostname;
+    }
+    // we don't want to return IP address (because of a security policy of a particular customer, see PUBDEV-5680)
+    hostname = System.getenv("HOSTNAME");
+    if ((hostname == null) || hostname.isEmpty()) {
+      hostname = "localhost";
+    }
+    System.out.println("WARN: Proxy IP address couldn't be translated to a hostname. Using environment variable HOSTNAME='" + hostname + "' as a fallback.");
```
honzasterba, line 382: would move this to else branch and provide better message for when env HOSTNAME is not available

honzasterba (approved):

PR #4944, opened 2020-09-22
PUBDEV-7720 avoid makeCopy and deepCopy

koniecsveta:


h2o-algos/src/main/java/hex/rulefit/RuleFit.java
```diff
@@ -248,7 +247,7 @@ public void computeImpl() {
                 for (Key key : keys) {
                     DKV.remove(key);
                 }
-                pathsFrame.remove();
+                DKV.remove(pathsFrame._key);
```
michalkurka, line 250: you can also look into method Frame#deleteTempFrameAndItsNonSharedVecs

honzasterba (approved):

PR #4945, opened 2020-09-22
PUBDEV-7319: TE integration to ModelBuilder (no mojo)

sebhrusen:
Same PR as https://github.com/h2oai/h2o-3/pull/4888 without the Mojo support.


Pscheidl (approved):

honzasterba (approved):
side note: why do we have csv files in test-dirs? should these rather live in smalldata?

sebhrusen:
@honzasterba these csv files are golden TE transformations for each problem type (binary, multiclass, regression) and each leakage strategy. Any change in TE that changes the columns order or how the encoded values are computed should break the tests relying on those golden.

michalkurka (approved):
Thank you!

PR #4947, opened 2020-09-23
PUBDV-7787: Use HTTPS for all URLs in R package

hannah-tillman:
https://0xdata.atlassian.net/browse/PUBDEV-7787


ledell (approved):
This is perfect, thank you @hannah-tillman!

PR #4948, opened 2020-09-23
[mojoland] run MOJOLand gradle with jdk8

honzasterba:


michalkurka:
there is a also gradle property `org.gradle.java.home`

PR #4952, opened 2020-09-24
PUBDEV-7786: Model explainability user guide page

ledell:
https://0xdata.atlassian.net/browse/PUBDEV-7786

- New Model Explainability page in the User Guide
- wine quality regression example
- code examples for all functions
- first draft complete with images

To review:

```
cd h2o-3/h2o-docs/src/product
make clean
make html
open _build/html/explain.html
```


h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,205 @@
...
+Explaination Plotting Functions 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+TO DO: Let's put examples of each function and the plot, in the order in which they appear in the ``h2o.explain()`` output.  Let's also show how to customize the plots.
+
+.. tabs::
+   .. code-tab:: r R
+
+        # Residual analysis plot for an AutoML object
+        ra_plot <- h2o.residual_analysis(aml, test)
```
tomasfryda, line 185: For current version change to `ra_plot <- h2o.residual_analysis(aml@leader, test)`. 

Residual analysis works on single model, I could change that to be able to invoke it on H2OAutoML object and just use the `leader` model in that case but I think it is better to enforce the user to be explicit (especially here since the residual_analysis can show very different things even in top models).

I can change the utility functions to be able to be invoked on H2OAutoML if you think it would be better.

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,205 @@
...
+        ra_plot
+
+   .. code-tab:: python
+
+        # Residual analysis plot for an AutoML object
+        ra_plot = aml.residual_analysis(test)
```
tomasfryda, line 191: Should be `ra_plot = aml.leader.residual_analysis(test)`.

The residual_analysis in python works on subclasses of `H2ORegressionModel` which H2OAutoML isn't AFAIK even if we do regression problem.

I can change the utility functions to be able to be invoked on H2OAutoML if you think it would be better.

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,205 @@
...
+        train[, y] <- as.factor(train[, y])
+        test[, y] <- as.factor(test[, y])
+
+        # Run AutoML for 20 base models
+        aml <- h2o.automl(x = x, y = y, 
+                          training_frame = train,
+                          max_models = 20,
+                          seed = 1)
+
+        # Explain all AutoML models
```
tomasfryda, line 107: I don't really know how to rephrase this but it won't try to explain all models, it tries to give some overview

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,205 @@
...
+        test[, y] <- as.factor(test[, y])
+
+        # Run AutoML for 20 base models
+        aml <- h2o.automl(x = x, y = y, 
+                          training_frame = train,
+                          max_models = 20,
+                          seed = 1)
+
+        # Explain all AutoML models                  
+        exa <- h2o.explain(aml, test)
```
tomasfryda, line 108: In R, the explanation has semantics similar to `ggplot2`, i.e., if we want to show it we have to use `print(exa)` or in case of Jupyter `repr(exa)`. In interactive use, it gets printed automatically since R REPL automatically prints, Rmarkdown also prints the value, and Jupyter calls `repr` on the value.

I would just add another line:
```
print(exa)
```

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,205 @@
...
+
+        # Explain a single model (top AutoML model)                 
+        exm <- h2o.explain(aml@leader, test)
```
tomasfryda, line 111: `print(exm)`

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,205 @@
...
+    - ``"model_correlation_heatmap"``
+    - ``"shap_summary"``
+    - ``"pdp"``
+    - ``"ice"``
+
+- **top_n_features**: If ``columns_of_interest`` is missing, create plots only with the top n columns (where applicable).  Defaults to 5.
+
+- **best_of_family**: If True, explain only best of family models.
+
+- **user_overrides**: Overrides for individual explanations, e.g. ``list(shap_summary_plot = list(top_n_features = 50))`` in R.
```
tomasfryda, line 65: changed to `plot_overrides`

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,205 @@
...
+    - ``"variable_importance"``
+    - ``"variable_importance_heatmap"``
+    - ``"model_correlation_heatmap"``
+    - ``"shap_summary"``
+    - ``"pdp"``
+    - ``"ice"``
+
+- **top_n_features**: If ``columns_of_interest`` is missing, create plots only with the top n columns (where applicable).  Defaults to 5.
+
+- **best_of_family**: If True, explain only best of family models.
```
tomasfryda, line 63: This parameter just got removed to make the R version behaviour same as python's:

When explaining AutoML, show just best of family to get some general overview, but when explaining a list of models assume that the user actually handpicked the models and actually wants to see them all.

(still overridable by using the `plot_overrides`)

h2o-docs/src/product/explainability.rst
ledell: No worries, it's perfectly fine as is (taking a single model only).  I will update the example.

h2o-docs/src/product/explainability.rst
ledell: It's fine to take single model for now, thanks!

h2o-docs/src/product/explainability.rst
ledell: Thanks, I will adjust the phrasing.

ledell:
I made some updates based on @tomasfryda's suggesting.  It still could use some more content -- after we make a final decision on the dataset to use (let's use something more interesting than Higgs...), then we should run the code and add screenshots of the plots to the user guide.  The page should be fairly visual.

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,213 @@
...
+        train[y] = train[y].asfactor()
+        test[y] = test[y].asfactor()
+        
+        # Run AutoML for 20 base models
+        aml = H2OAutoML(max_models=20, seed=1)
+        aml.train(x=x, y=y, training_frame=train)
+
+        # Explain leader model & compare with all AutoML models 
+        exa = aml.explain(test)
+        print(exa)
```
tomasfryda, line 145: Remove `print` in python.

Reasons mentioned above.

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,213 @@
...
+
+        # Explain a single H2O model (e.g. leader model from AutoML)
+        exm = aml.leader.explain(test)
+        print(exm)
```
tomasfryda, line 149: Remove `print` in python.

Reasons mentioned above.

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,213 @@
...
+
+        # Residual analysis plot for an AutoML object
+        ra_plot <- h2o.residual_analysis(aml@leader, test)
+        print(ra_plot)
+
+   .. code-tab:: python
+
+        # Residual analysis plot for an AutoML object
+        ra_plot = aml.leader.residual_analysis(test)
+        print(ra_plot)
```
tomasfryda, line 200: Remove `print` in python.

Reasons mentioned above.

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,213 @@
...
+        # Explain an AutoML object
+        exa <- h2o.explain(aml, test)
+        print(exa)
+
+
+   .. code-tab:: python
+
+        # Explain a model
+        exm = model.explain(test)
+        print(exm)
```
tomasfryda, line 29: Remove `print` in python.

In python, the `explain` methods have `render` argument when True (default) it will get plotted else it won't and it will just return an ordered dictionary.

So there is no need to `print` it and it shouldn't be used (python's `print` has different semantics than R's, `print` should just print out plain text).

The difference between Python and R is due to multiple technical reasons, e.g., `matplotlib` doesn't like more active plots.

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,213 @@
...
+
+        # Explain an AutoML object
+        exa = aml.explain(test)
+        print(exa)
```
tomasfryda, line 33: Remove `print` in python.

Reasons mentioned above.

tomasfryda (changes requested):

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,213 @@
...
+Parameters
+~~~~~~~~~~
+
+TO DO: This was created looking at the R interface, may need some updates for Python.  
+
+- **object**: (R only) One of the following: an H2O model, a list of models, an H2OAutoML object, or an H2OAutoML leaderboard slice.
+
+- **test_frame**: An H2OFrame of test data.
+
+- **columns_of_interest**: A vector of column names. If specified, create plots only with these columns (where applicable).
```
tomasfryda, line 47: together with `top_n_features` changed to `columns`:
```
#' @param columns A vector of column names to create plots only with or a number of columns to pick
#'                 using variable importance (where applicable).
```

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,213 @@
...
+    - ``"confusion_matrix"``
+    - ``"residual_analysis"``
+    - ``"variable_importance"``
+    - ``"variable_importance_heatmap"``
+    - ``"model_correlation_heatmap"``
+    - ``"shap_summary"``
+    - ``"pdp"``
+    - ``"ice"``
+
+- **top_n_features**: If ``columns_of_interest`` is missing, create plots only with the top n columns (where applicable).  Defaults to 5.
```
tomasfryda, line 65: removed

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,223 @@
...
+- **columns**: If ``columns`` is missing, create plots only with the top n columns (where applicable).  Defaults to 5.
+
+- **plot_overrides**: Overrides for individual explanations, e.g. ``list(shap_summary_plot = list(top_n_features = 50))`` in R. 
+
+Notes
+~~~~~
+
+Our roadmap for improving the explainability interface is [here](https://0xdata.atlassian.net/jira/software/c/projects/PUBDEV/issues/PUBDEV-7806?filter=allissues).
+
+We are looking for feedback from users.  If you have any feature requests, please file a
```
tomasfryda, line 77: Incomplete sentence

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,223 @@
...
+        train[y] = train[y].asfactor()
+        test[y] = test[y].asfactor()
+        
+        # Run AutoML for 20 base models
+        aml = H2OAutoML(max_models=10, seed=1)
+        aml.train(x=x, y=y, training_frame=train)
+
+        # Explain leader model & compare with all AutoML models 
+        exa = aml.explain(test)
+        display(exa)
```
tomasfryda, line 155: The `display(exa)` will display the explanation for the second time. The `aml.explain()` will display it since some matplotlib backends don't make it easy to show closed figure and not closing it would make Jupyter automatically showing it.

So if you don't provide `explain` function with `render=False`, it will render it when it runs. 

We could either add a comment saying that `display` will show it for second time or remove the `display`.

The other option could be `aml.explain(test, render=False)` but I would discourage putting it to the example since it makes it hard to show the figures in CLI and could confuse the users.

h2o-docs/src/product/explainability.rst
```diff
@@ -0,0 +1,223 @@
...
+
+        # Explain a single H2O model (e.g. leader model from AutoML)
+        exm = aml.leader.explain(test)
+        display(exm)
```
tomasfryda, line 159: see the comment above

h2o-docs/src/product/explainability.rst
ledell: @tomasfryda Thanks for the reminder.. ill update to be clear about this.

h2o-docs/src/product/explainability.rst
ledell: thx

h2o-docs/src/product/explainability.rst
ledell: fixed

h2o-docs/src/product/explainability.rst
ledell: fixed

h2o-docs/src/product/explainability.rst
ledell: fixed

h2o-docs/src/product/explainability.rst
ledell: fixed

h2o-docs/src/product/explainability.rst
ledell: done.

h2o-docs/src/product/explainability.rst
ledell: done

h2o-docs/src/product/explainability.rst
ledell: TO DO: We are going to change this back to two paramters and I will need to update this

h2o-docs/src/product/explain.rst
```diff
@@ -0,0 +1,369 @@
...
+The ``H2OExplanation`` is a named list in R and an ordered dictionary in Python, so explanations can be accessed by their name (same names as in ``include_explanations`` and ``exlude_exlpanations``).
+
+
+When ``h2o.explain()`` is provided a list of models, the following explanations will be generated by default:
+
+- Leaderboard (compare all models)
+- Confusion Matrix for Leader Model (classification only)
+- Residual Analysis for Leader Model (regression only)
+- Variable Importance of Top Base (non-Stacked) Model 
+- Variable Importance Heatmap (compare all models)
```
tomasfryda, line 188: `varimp_heatmap` is unable to compare Stacked Ensembles since they don't have the `varimp`

h2o-docs/src/product/explain.rst
```diff
@@ -0,0 +1,369 @@
...
+
+There are a number of individual plotting functions that are used inside the ``explain()`` function.  Some of these functions 
+
+Takes a list of models (including an AutoML object or leaderboard slice) as input:
+
+.. tabs::
+   .. code-tab:: r R
+
+        h2o.varimp_plot()
+        h2o.varimp_heatmap()
```
tomasfryda, line 216: inconsistencies in parentheses - some function have them here and some don't

h2o-docs/src/product/explain.rst
```diff
@@ -0,0 +1,369 @@
...
+
+Takes a list of models (including an AutoML object or leaderboard slice) as input:
+
+.. tabs::
+   .. code-tab:: r R
+
+        h2o.varimp_plot()
+        h2o.varimp_heatmap()          
+        h2o.model_correlation_heatmap        
+        h2o.pdp_multi_plot
```
tomasfryda, line 218: it should be `h2o.pd_multi_plot` (the second `p` would stand for "plot")

h2o-docs/src/product/explain.rst
```diff
@@ -0,0 +1,369 @@
...
+
+   .. code-tab:: python
+
+        varimp_plot
+        varimp_heatmap          
+        model_correlation_heatmap        
+        pdp_multi_plot
```
tomasfryda, line 225: `h2o.pd_multi_plot`

h2o-docs/src/product/explain.rst
```diff
@@ -0,0 +1,369 @@
...
+   .. code-tab:: r R
+
+        h2o.varimp_plot()
+        h2o.varimp_heatmap()          
+        h2o.model_correlation_heatmap        
+        h2o.pdp_multi_plot
+
+   .. code-tab:: python
+
+        varimp_plot
```
tomasfryda, line 222: `varimp_plot` is defined for models, not AutoML (user would have to do `aml.leader.varimp_plot()`)

h2o-docs/src/product/explain.rst
ledell: thx

h2o-docs/src/product/explain.rst
ledell: thx

h2o-docs/src/product/explain.rst
ledell: thx

h2o-docs/src/product/explain.rst
ledell: thx

h2o-docs/src/product/explain.rst
ledell: thx

ledell:
@tomasfryda I added lots of new content to `explain.rst` and images (R/ggplot2 version only) which demonstrate the output for all the plotting functions.  Please let me know if you see anything else!  Otherwise, we should be good to go.

ledell:
Hi @hannah-tillman if you have time, please take a quick look at the new Explainability page.  If you see anything weird, please let me know!

ledell:
@tomasfryda @hannah-tillman I decided to merge becuase the tests were green and I wanted to get it into master.  If you see anything weird/wrong, please let me know and I'll do a quick hotfix PR to address anything else.

PR #4957, opened 2020-09-24
PUBDEV-7787: Use HTTPS for all URLs in R package

hannah-tillman:
@satai thanks for catching this error earlier: made the changes to the source files this time and built h2o-3


ledell (approved):
Great catch, thanks!

satai (approved):

PR #4962, opened 2020-09-25
PUBDEV-7720 set min_rule_length and max_rule_length default values to 3

koniecsveta:
since results of tests on big data (= ml-benchmarks results) are good (= AUC and logloss similar to AUC and logloss of underlying tree ensemble on the same data) even for setup with min_rule_length = 3 and max_rule_length = 3, default values of these parameters will be set to 3, 3 instead of 1, 10. 1, 10 can be considered as overkill and consuming much more memory, when results are already good with 3, 3.

### DIFF
--- a/h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
+++ b/h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
@@ -45,11 +45,11 @@ public class RuleFitModel extends Model<RuleFitModel, RuleFitModel.RuleFitParame
         // the algorithm to use to generate rules. Options are "DRF" (default), "GBM"
         public Algorithm _algorithm = Algorithm.AUTO;
 
-        // minimum length of rules. Defaults to 1.
-        public int _min_rule_length = 1;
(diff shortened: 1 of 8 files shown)


honzasterba:
why not 1,3? how is 3,3 better than 1,3?

koniecsveta:
> why not 1,3? how is 3,3 better than 1,3?

3,3 means you will train 1 tree model of depth 3, extract information from it, add it to train frame and put into GLM
1,3 means you will train 3 tree models: 1st of depth 1, 2nd of depth 2, 3rd of depth 3, extract them, add to train and put into GLM.

by min_rule_length and max_rule_length you define interval of depths and for each depth you train a tree. The bigger the interval, the more models -> rules you have.. but rules which are too short or too long can worsen final interpretability, because they are too wide or too specific / depends also on data...

so I think can be also 1, 3 / can add now to benchmarks when I will be removing 1, 10... to have lets say 1,3; 3,3 and 1,5

michalkurka (approved):

PR #4966, opened 2020-09-27
set cardinality threshold to 25 for TE integration in AutoML

sebhrusen:
Based on benchmarks, we changed the cardinality threshold for which Target Encoding is triggered in AutoML to 25 categories (instead of 10).

https://0xdata.atlassian.net/browse/PUBDEV-7778


ledell (approved):
Thanks!

PR #4968, opened 2020-09-27
PUBDEV-7808 - Helm chart

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-7808

Implemented a basic Helm chart for H2O, with:
- Configurable H2O-related env vars
- Configurable resources with minimalistic defaults (not sure if 1 CPU is that minimalistic, but whatever). Let's not force users fill those unless they really want to.
- Node selectors applicable
- Name not generated or verified, this is beyond HELM. Modifiable (Helm standard). Up to user to provide a unique name.
- Namespace is taken from Kubeconfig by Helm by default :wink: . This is much desired behavior. Overridable of course.
- Parallel pod management used - this is a must, otherwise H2O startup will be slow.

H2O-related changes:
- Added some additional logging into H2O embedded config for better debugging if something goes wrong (constraint logging)

**Additional TODOs**
- [x] Log constrants used in H2O Kubernetes embedded config
- [x] Use this HELM chart in existing H2O Kubernetes test suite. Requires new K8S test docker image version - there is no helm in the existing one.
- [x] Implemented basic HELM test for users to use (optional)
- [x] With each release, replace `appVersion` in `values.yaml` with latest H2O app. This app version will then be used as docker image tag.
- [x] With each release, push the respective HELM chart into our self-hosted repo. There is no such repository yet, creation in process. The script to do that could be fun, as we'll need helm in place to push to the repository. Will require docker image update probably.
- [x] Make sure requests and limits are always the same, for reproducibility.

@bilcus This is early draft, I'll try to polish most of the points mentioned above during tomorrow. This is for you guys to have the ability to raise issues/requirements early :wink: . Any feedback (incl. positive -everything is fine) is OK.

@michalkurka I'd like this to go into 3.32.0.1 if possible.

@mn-mikke SW will have to use H2O as well. Maybe we could create H2O Helm lib chart as well - maybe not even for internal usage. Probably won't be solved in this PR. WDYT ?


h2o-k8s/tests/clustering/h2o-service.yaml
```diff
@@ -59,7 +59,7 @@ spec:
       port: 80
       targetPort: 54321
 ---
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
```
Pscheidl, line 62: Outdated apiVersion was used. Fixed.

h2o-helm/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v2
+name: h2o
+description: H2O open source machine learning platform official Helm chart
+
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 1.0.0
```
satai, line 10: Just a question - should be this numbers separated to our version numbers?

Pscheidl: The version of the Helm chart could be aligned with H2O release. Good point. cc @michalkurka .

Would be great to see it start with 3.32.0.1, but why not ?

michalkurka: yes, I agree with @satai - would be good to keep those 2 in synch - the versions should be the sae

Pscheidl: We'll only have one then.

michalkurka: yeah, that is what we proposed internally here: https://h2oai.slack.com/archives/C0BGE8ZJP/p1600973747002400?thread_ts=1600970951.001100&cid=C0BGE8ZJP

if you think this approach has issue - we should revisit the discussion

mn-mikke:
> @mn-mikke SW will have to use H2O as well. Maybe we could create H2O Helm lib chart as well - maybe not even for internal usage. Probably won't be solved in this PR. WDYT ?

It would nice if your helm charts covered also manually deployed external backend of SW. SW external backend requires an a small SW library to be on the classpath of each H2O node. I'm not sure whether it's possible to append an library via a helm chart option or not.

If helm is a long term strategy for h2o-3 deployment, the separate chart lib might handy and SW  could use it for automatically started external backend. I'm quite not happy with the current solution in SW that YAML files are copy-pasted to SW codebase. A chart lib sounds like a better solution.

h2o-helm/templates/statefulset.yaml
```diff
@@ -0,0 +1,59 @@
...
+    metadata:
+      labels:
+    {{- include "h2o-helm.selectorLabels" . | nindent 8 }}
+    spec:
+      securityContext:
+      {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.name }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          command: ["/bin/bash", "-c", "java -XX:+UseContainerSupport -XX:MaxRAMPercentage={{ .Values.h2o.memoryPercentage }} -jar /opt/h2oai/h2o-3/h2o.jar"]
```
bilcus, line 25: Can you add optional javaArgs and h2oArgs values?
This way you can pass additional args to Java/H2O without needing to modify the Chart.

h2o-helm/values.yaml
```diff
@@ -0,0 +1,57 @@
...
+  tag: ""
+  
+h2o:
+  nodeCount: 1
+  memoryPercentage: 50
+  lookupTimeout: 180 # Three minutes by default are very generous amount of time for clustering with parallel pod management
+  kubernetesApiPort: 8080
+  
+
+imagePullSecrets: []
```
bilcus, line 19: Looks like it is unused in the Chart

h2o-helm/templates/statefulset.yaml
Pscheidl: Removed.

h2o-helm/values.yaml
Pscheidl: Removed.

h2o-helm/templates/statefulset.yaml
Pscheidl: I've put in fully customizable command, as discussed. A command needs to be fully overridable in case custom docker image is used. Would introduce additional complexity especially for users if custom args were also present - they'd be confused when and what is actually enabled :wink:

Pscheidl:
Test integration complete and successful. Wrong port mapping was to blame.

Now, the release part must be done:

- Replace version before release in `Chart.yaml` with current H2O version being released
- Do helm package
- Upload the resulting archive to the S3 folder `https://s3.console.aws.amazon.com/s3/buckets/h2oai-charts`. @bilcus already pointed https://charts.h2o.ai/ to the S3 bucket.
- Make sure the link with directions is included in the Download page.

@hannah-tillman we'll need to change the docs and inform about Helm existence. I'll try to create a PR and you'll then transalate it to proper english please :wink:

Pscheidl:
Waiting for newest dev docker images to be built after rebase: http://mr-0xc1:8080/view/H2O-3/job/docker-images/job/h2o-3-dev-images/63/

h2o-helm/Chart.yaml
Pscheidl: I don't. It's fine.

h2o-k8s/tests/clustering/docker-init-script.sh
```diff
@@ -14,10 +14,11 @@ export KUBECONFIG="$(k3d get-kubeconfig --name='k3s-default')"
...
-kubectl apply -f h2o-service-subst.yaml
-kubectl wait --for=condition=available --timeout=600s deployment.apps/h2o-deployment -n default
-rm h2o-service-subst.yaml
+envsubst < testvalues-template.yaml >> testvalues.yaml
+helm install -f testvalues.yaml h2o $H2O_BASE/h2o-helm --kubeconfig $KUBECONFIG --dry-run # Shows resulting YAML
+helm install -f testvalues.yaml h2o $H2O_BASE/h2o-helm --kubeconfig $KUBECONFIG
+helm test h2o
+kubectl get ingresses
 kubectl describe pods
 timeout 120s bash h2o-cluster-check.sh
```
Pscheidl, line 23: The cluster check actually also checks the ingress - that's why it is kept. This also makes sure the `helm test` is coherent with test pipeline checks and vice versa.

docker/jenkins-images/Dockerfile-release
```diff
@@ -19,5 +19,11 @@ RUN \
     conda update conda && \
     conda clean --all --yes
 
+RUN curl https://baltocdn.com/helm/signing.asc | apt-key add - \
```
Pscheidl, line 22: http://mr-0xc1:8080/view/H2O-3/job/docker-images/job/h2o-3-dev-images/63/

docker/jenkins-images/Dockerfile-k8s-test
```diff
@@ -11,7 +11,13 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s
 	&& chmod +x ./kubectl \
 	&& mv ./kubectl /usr/local/bin/kubectl
 
-RUN curl -s https://raw.githubusercontent.com/rancher/k3d/master/install.sh | bash
+RUN curl -s https://raw.githubusercontent.com/rancher/k3d/main/install.sh | TAG=v1.7.0 bash
```
Pscheidl, line 14: http://mr-0xc1:8080/view/H2O-3/job/docker-images/job/h2o-3-dev-images/63/

docker/jenkins-images/Dockerfile-k8s-test
```diff
@@ -11,7 +11,13 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s
...
+
+RUN curl https://baltocdn.com/helm/signing.asc | apt-key add - \
```
Pscheidl, line 16: http://mr-0xc1:8080/view/H2O-3/job/docker-images/job/h2o-3-dev-images/63/

scripts/jenkins/groovy/defineTestStages.groovy
```diff
@@ -51,6 +51,12 @@ def call(final pipelineContext) {
     [
       stageName: 'Java 8 Smoke', target: 'test-junit-smoke-jenkins', javaVersion: 8, timeoutValue: 20,
       component: pipelineContext.getBuildConfig().COMPONENT_JAVA
+    ],
+    [
+      stageName: 'Kubernetes', target: 'test-h2o-k8s', timeoutValue: 20,
```
honzasterba, line 56: this is to be removed before merge correct?

honzasterba (approved):

scripts/jenkins/jenkinsfiles/Jenkinsfile-Release
```diff
@@ -180,6 +180,36 @@ try {
...
+                        // Helm automatically installs latest chart if version is unspecified by the user, therefore there is no need for a latest tag
+                        insideDocker([], pipelineContext.getBuildConfig().AWSCLI_IMAGE, pipelineContext.getBuildConfig().DOCKER_REGISTRY, pipelineContext.getBuildConfig(), 3, 'HOURS') {
+                            if (!params.TEST_RELEASE) {
+                                sh """
+                                    cd ${env.BUILD_NUMBER_DIR}/h2o-helm/
+                                    sed -i 's/version: [0-9]*\.[0-9]*\.[0-9]*/version: ${env.PROJECT_VERSION}/' Chart.yaml
+                                    sed -i 's/appVersion: [0-9]*\.[0-9]*\.[0-9]*/version: ${env.PROJECT_VERSION}/' Chart.yaml
+                                    helm package .
+                                    mkdir h2o-3
+                                    cp h2o-${env.PROJECT_VERSION}.tgz h2o-3/
```
michalkurka, line 198: should this be eg. h2o-3/nightly vs h2o-3/stable?

Pscheidl: That's one way to deal with nightly vs stable - to have multiple folders. Could be.

Pscheidl: We'll have two types of charts then - with two different names - the actuallly difference is made in the chart name in `Chart.yaml`, the folders are only relevant for `index.yaml` in the repository itself.

Pscheidl: I'm not sure we want nightlies. This would bloat the repository index with lots of entries. If someone wants nightly H2O-3, they have the chance to do that with the existing helm chart - just change the tag name. A chart is not specifically tied to a single H2O version.

bilcus (approved):
LGTM

scripts/jenkins/jenkinsfiles/Jenkinsfile-Release
michalkurka: ok, I thought would be good to have to test the release process but OK

scripts/jenkins/groovy/defineTestStages.groovy
Pscheidl: Yes.

honzasterba (approved):

scripts/jenkins/jenkinsfiles/Jenkinsfile-Release
Pscheidl: Resolved by having extra Helm repo just for nightlies.

Pscheidl:
This is it, it works this way in rel-zermelo-rc1
http://mr-0xc1:8080/view/H2O-3/job/h2o-3-release-pipeline/job/rel-zermelo-rc1/9/

cc @michalkurka @bilcus @honzasterba

PR #4971, opened 2020-09-29
fix build - mr-0xg1 is the new name of build machine

satai:


scripts/jenkins/groovy/buildConfig.groovy
```diff
@@ -336,12 +336,15 @@ class BuildConfig {
 
   static enum JenkinsMaster {
     C1, // indicates we are running under mr-0xc1 master - master or nightly build
+    G1, // indicates we are running under mr-0xg1 master - master or nightly build
     B4  // indicates we are running under mr-0xb4 master - PR build
 
     private static JenkinsMaster findByName(final String name) {
       switch(name.toLowerCase()) {
         case 'c1':
```
honzasterba, line 344: do we need separate settings for both? I think we can just merge these two case: into one and have only one setting

michal-raska: I agree

satai: I would like to see it running now and clean it (the usage of enum is a bit of overkill here, IMO and can be cleaned)

honzasterba (approved):

michal-raska (approved):

PR #4975, opened 2020-09-29
R Jenkins image -  add new suggested packages

tomasfryda:
I am adding 2 packages -  'IRdisplay', 'plotly'.

I also found couple issues with newer packages and R-3.3, so I attempted to fix them and succeeded as far as I can tell.

### DIFF
--- a/docker/scripts/install_R_version
+++ b/docker/scripts/install_R_version
@@ -77,7 +77,7 @@ echo "Installing dependencies for R ${R_VERSION}"
   )) \
 " 2>&1 | tee R.install.1.log
 checkRInstallLog R.install.1.log
-# Latest XGBoost (from source) requires c++14, possibly causing troubles
+# Latest XGBoost (from source) does not compile; see https://0xdata.atlassian.net/browse/PUBDEV-7812
 #/usr/local/R/current/bin/R -e "chooseCRANmirror(graphics=FALSE, ind=1); \
 #  install.packages(c('xgboost'), type='source') \
 #" 2>&1 | tee R.install.2.log
@@ -108,7 +108,12 @@ checkRInstallLog R.install.3.log
   install_version('Hmisc', version = '4.3-0'); \
 " 2>&1 | tee R.install.4.log
 checkRInstallLog R.install.4.log
+
 # dependecies from GitHub
+# The R_REMOTES_UPGRADE=never disables automatic upgrade of dependencies.
+# mainly because of systemfonts that is not supported in newer versions by R 3.3
+# is also a dependency of pkgdown which would try otherwise to upgrade it to a version
+# which doesn't support R < 3.6
 R_REMOTES_UPGRADE=never
 /usr/local/R/current/bin/R -e "chooseCRANmirror(graphics=FALSE, ind=1); library(devtools); \
   install_version('systemfonts', version = '0.2.3'); \
--- a/docker/scripts/install_R_version
+++ b/docker/scripts/install_R_version
@@ -68,7 +68,7 @@ function checkRInstallLog() {
 
 # Install dependencies
 echo "Installing dependencies for R ${R_VERSION}"
-/usr/local/R/current/bin/R -e "chooseCRANmirror(graphics=FALSE, ind=58); \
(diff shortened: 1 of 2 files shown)


docker/scripts/install_R_version
```diff
@@ -60,54 +60,62 @@ activate_R_${R_VERSION}
...
+  if grep -E '(packages .* are not available)|(installation of package .* had non-zero exit status)' $log_file; then
     echo "Errors detected in R package installation."
     exit 1
   fi
 }
 
 # Install dependencies
 echo "Installing dependencies for R ${R_VERSION}"
-/usr/local/R/current/bin/R -e "chooseCRANmirror(graphics=FALSE, ind=58); \
+/usr/local/R/current/bin/R -e "chooseCRANmirror(graphics=FALSE, ind=1); \
```
tomasfryda, line 71: set mirror to "0-Cloud" which does "automatic redirection to servers worldwide, currently sponsored by Rstudio"

At the time of this change using mirror 58 I wasn't able to download some packages. But it is possible that it was just temporarily down.

docker/scripts/install_R_version
```diff
@@ -60,54 +60,62 @@ activate_R_${R_VERSION}
...
 # Hmisc 4.3-0 is the last to support survival 2.44
-/usr/local/R/current/bin/R -e "chooseCRANmirror(graphics=FALSE, ind=58); library(devtools); \
+/usr/local/R/current/bin/R -e "chooseCRANmirror(graphics=FALSE, ind=1); library(devtools); \
   install_version('ROCR', version = '1.0-7'); \
   install_version('Hmisc', version = '4.3-0'); \
 " 2>&1 | tee R.install.4.log
 checkRInstallLog R.install.4.log
 # dependecies from GitHub
-/usr/local/R/current/bin/R -e "chooseCRANmirror(graphics=FALSE, ind=58); library(devtools); \
+R_REMOTES_UPGRADE=never
```
tomasfryda, line 112: Needed to prevent pkgdown updating its dependencies (systemfonts) that don't support R3.3

docker/scripts/install_R_version
```diff
@@ -60,54 +60,62 @@ activate_R_${R_VERSION}
...
-    'ade4', 'e1071', 'ggplot2', 'LiblineaR', 'optparse', 'uuid', 'usethis', 'rgl', 'plot3Drgl' \
+    'ade4', 'e1071', 'ggplot2', 'LiblineaR', 'optparse', 'uuid', 'usethis', 'rgl', 'plot3Drgl',  \
+    'IRdisplay', 'plotly' \
   )) \
 " 2>&1 | tee R.install.1.log
 checkRInstallLog R.install.1.log
-# Latest XGBoost (from source)
-/usr/local/R/current/bin/R -e "chooseCRANmirror(graphics=FALSE, ind=58); \
-  install.packages(c('xgboost'), type='source') \
+# Latest XGBoost (from source) requires c++14, possibly causing troubles
```
tomasfryda, line 80: @honzasterba It might not be related to C++14 but was the only obvious change between the two versions.

I'm not sure if it is OK not using the latest version but the latest doesn't compile and I didn't have much time to explore if I can upgrade gcc etc.

docker/scripts/install_R_version
satai: I like the fact, that 1 is 0 :-D

satai:
LGTM but I am probably not the right person to see issues with this one.

docker/scripts/install_R_version
honzasterba: please create a JIRA for this

docker/scripts/install_R_version
honzasterba: 1) this can be done via instal_version(..., upgrade="never")
2) if you want to add this for all installs please add the above comment to the file so we know why it is there in the future

docker/scripts/install_R_version
tomasfryda: Done. https://0xdata.atlassian.net/browse/PUBDEV-7812

honzasterba (approved):

PR #4978, opened 2020-09-29
rfit benchmarks adjustment

koniecsveta:


h2o-algos/src/main/java/hex/rulefit/RuleFit.java
```diff
@@ -188,8 +189,12 @@ public void computeImpl() {
                 Key[] keys = new Key[depths.length];
                 // prepare rules
                 if (RuleFitModel.ModelType.RULES_AND_LINEAR.equals(_parms._model_type) || RuleFitModel.ModelType.RULES.equals(_parms._model_type)) {
+                    long startAllTreesTime = System.nanoTime();
                     for (int modelId = 0; modelId < depths.length; modelId++) {
+                        long startModelTime = System.nanoTime();
                         SharedTreeModel treeModel = trainTreeModel(_parms._algorithm, depths[modelId]);
+                        long endModelTime = System.nanoTime() - startModelTime;
+                        Log.debug("Tree model n." + modelId + " trained in " + ((double)endModelTime) / 1E9 + "s.");
```
honzasterba, line 197: ideally use Log4j Logger

scripts/jenkins/groovy/defineTestStages.groovy
```diff
@@ -270,7 +270,7 @@ def call(final pipelineContext) {
     ],
     [
       stageName: 'Rulefit Benchmark', executionScript: 'h2o-3/scripts/jenkins/groovy/benchmarkStage.groovy',
-      timeoutValue: 120, target: 'benchmark', component: pipelineContext.getBuildConfig().COMPONENT_ANY,
+      timeoutValue: 500, target: 'benchmark', component: pipelineContext.getBuildConfig().COMPONENT_ANY,
```
honzasterba, line 273: I think we should think about how many different benchmarks we are running for RuleFit and ideally keep the time inder 2hours, we have little free time on the benchmark machine as it is. WDYT @michalkurka ?

koniecsveta: we have way more test cases in rulefit than in other stages. Im working on running tree ensembles building in parallel, so would be good to run this at least once like this - to be able to evaluate the improvement.

In current state it timeouted after 51 minutes with 20 cases done (depths 3 to 3 and 1 to 3), with 21st nearly done (redhat, depths 1 to 5).

Would put this back after parallelization and if it still timeouts then don't do all 30 test cases but only like 10 or split to multiple stages (like its being done for other algos) and continue optimizing

honzasterba: do you expect more thant 50% speedup? because if 20 tests took 2 hours and those were the "faster" ones I think we can expect the total time to be close to 4 hours, that would mean your optimsations would have to bring it down to half to fit into 2h

koniecsveta: I have no idea, need to do it, run it and see. The next option is to not to test 1 to 5 as we have good accuracy already with "faster" ones to not to waste time on benchmark machine. But I thought would be good to experiment also with slower parametrisation before deciding to not to have it as a test case at all - that's what we have it for, isn't it?

honzasterba: I think its OK to do experiments like this in a separate branch, but once you push this into master the results for the benchmarks will be there forever and if you decide to remove some because they take too long than the old results will always be there, I think we should put into master only benchmarks that we are quite sure we want to run forever

koniecsveta: okok will do this in a separate branch

koniecsveta: reverted and done in separate branch

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
koniecsveta: done

honzasterba (approved):

PR #4979, opened 2020-09-29
Fix race condition in Model::adaptTestForTrain

tomasfryda:
This race condition occurs when doing PDP calculation during
StackedEnsemble's metalearner (GLM) scoring.

Fails on `assert !_write_lock;` in `water.util.IcedHashMapBase.put`.

This PR mitigates the issue - it has been tested locally and on Jenkins in Model explainability PR (https://github.com/h2oai/h2o-3/pull/4932) where this issue first appeared.

Stack Trace:
```
java.lang.RuntimeException: java.lang.AssertionError
	at water.Futures.blockForPending(Futures.java:88)
	at
hex.PartialDependence$PartialDependenceDriver.compute2(PartialDependence.java:263)
	at
jsr166y.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:104)
Caused by: java.lang.AssertionError
	at water.util.IcedHashMapBase.put(IcedHashMapBase.java:69)
	at hex.Model.adaptTestForTrain(Model.java:1424)
	at hex.Model.adaptTestForTrain(Model.java:1300)
	at hex.Model.score(Model.java:1588)
	at
hex.ensemble.StackedEnsembleModel.predictScoreImpl(StackedEnsembleModel.java:172)
	at hex.Model.score(Model.java:1599)
	at hex.Model.score(Model.java:1582)
	at
hex.PartialDependence$PartialDependenceDriver$CalculatePdpPerBin.compute2(PartialDependence.java:400)
	... 6 more
```

The problem is somehow connected with `Model._toDelete` and `adaptTestForTrain`.
I tried multiple things to get rid of it but the only solution that worked is to pass a new `IcedHashMap` to the `adaptTestForTrain`, `putIfAbsent` things from `_toDelete`  and then reassign it back to `_toDelete`.
I also tried cloning it as @Pscheidl suggested but that didn’t help.

Possibly related to https://github.com/h2oai/h2o-3/pull/4969 .

### DIFF
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -1310,7 +1310,8 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
   }
   
   public String[] adaptTestForTrain(Frame test, boolean expensive, boolean computeMetrics, boolean catEncoded) {
-    return adaptTestForTrain(
+    final IcedHashMap<Key, String> toDelete = new IcedHashMap<>();
+    String[] warnings = adaptTestForTrain(
             test,
             _output._origNames,
             _output._origDomains,
@@ -1321,9 +1322,14 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
             computeMetrics,
             _output.interactionBuilder(),
             getToEigenVec(),
-            _toDelete,
+            toDelete,
             catEncoded
     );
+    for (Map.Entry<Key, String> entry: _toDelete.entrySet())
+      toDelete.putIfAbsent(entry.getKey(), entry.getValue());
+
+    _toDelete = toDelete;
+    return warnings;
   }
 
   /**
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -1310,8 +1310,7 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
   }
   
   public String[] adaptTestForTrain(Frame test, boolean expensive, boolean computeMetrics, boolean catEncoded) {
-    final IcedHashMap<Key, String> toDelete = new IcedHashMap<>();
-    String[] warnings = adaptTestForTrain(
+    return adaptTestForTrain(
             test,
             _output._origNames,
             _output._origDomains,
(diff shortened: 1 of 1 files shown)


h2o-core/src/main/java/hex/Model.java
```diff
@@ -1321,9 +1322,14 @@ public double lift_top_group() {
...
             _output.interactionBuilder(),
             getToEigenVec(),
-            _toDelete,
+            toDelete,
             catEncoded
     );
+    for (Map.Entry<Key, String> entry: _toDelete.entrySet())
+      toDelete.putIfAbsent(entry.getKey(), entry.getValue());
+
+    _toDelete = toDelete;
```
honzasterba, line 1331: why not _toDelete.putAll(toDelete)?

tomasfryda: That fails the same way on `assert !_write_lock;`. I'm not sure but it seem to me that the potential issue comes from overwriting something that exists in the map while it is in `IcedHashMapBase::write_impl`. (I'm guessing that is the reason why `putIfAbsent()` doesn't check the `_write_lock`.)

Pscheidl: Problem are the parallel operations on top of one map. Where do they come from ? In my case clone helps. Note to self: See how Clone is implemented. This seems either like a:

1. Bug in ModelBuilder
2. Intended state but poor H2O locking mechanism (this structure should be guaranteed to be used lock-free, but it's not).

honzasterba: If this is about concurent modifications of _toDelete than I dont think this is the right way to go because this can throw away modifications done to _toDelete in parallel. Overall I understand this is a workaround to get something fixed but I would much rather see a proper fix to this problem, even though I understand this might be very difficult.

tomasfryda: Thanks, that's a very good point! I am still not able to determine what's actually causing it but I was able to make a different fix that shouldn't suffer the "throwing away modifications" issue.

In `adaptTestForTrain`, I changed `toDelete.put` to `toDelete.putIfAbsent`.  That shouldn't have any negative side effect as far as I can tell. 

As far as I can tell the issue isn't about concurrent modification, it seem to be about being modified while being read in `IcedHashMapBase::write_impl`.

Docstring of `write_impl` says:
```
  // Since this is a *concurrent* hashtable, writing it whilst its being
  // updated is tricky.  If the table is NOT being updated, then all is written
  // as expected.  If the table IS being updated we only promise to write the
  // Keys that existed at the time the table write began.  If elements are
  // being deleted, they may be written anyways.  If the Values are changing, a
  // random Value is written.
```

So to me seems that this fix should be ok => if the new entry isn't written in current run of `write_impl` it should be written the next time. Since  `putIfAbsent` just adds, there shouldn't be any inconsistencies.

honzasterba (approved):

michalkurka:
Alternative solution: https://github.com/h2oai/h2o-3/pull/4993

PR #4980, opened 2020-09-30
PUBDEV-7720 build tree ensembles for rulefit in parallel

koniecsveta:
this can be done by the same approach as it is used during cv and was used here: https://github.com/h2oai/mli/pull/38/files#diff-3a6961d6c17b4ccf721d40674f1b9a0cR147

### DIFF
--- a/h2o-algos/src/main/java/hex/rulefit/RuleFit.java
+++ b/h2o-algos/src/main/java/hex/rulefit/RuleFit.java
@@ -5,7 +5,6 @@ import hex.genmodel.utils.DistributionFamily;
 import hex.glm.GLM;
 import hex.glm.GLMModel;
 import hex.schemas.TreeV3;
-import hex.tree.SharedTree;
 import hex.tree.SharedTreeModel;
 import hex.tree.TreeHandler;
 import hex.tree.drf.DRF;
@@ -193,11 +192,9 @@ public class RuleFit extends ModelBuilder<RuleFitModel, RuleFitModel.RuleFitPara
                 // prepare rules
                 if (RuleFitModel.ModelType.RULES_AND_LINEAR.equals(_parms._model_type) || RuleFitModel.ModelType.RULES.equals(_parms._model_type)) {
                     long startAllTreesTime = System.nanoTime();
-                    SharedTree[] builders = getBuilders(_parms._algorithm, depths);
-                    builders[0].bulkBuildModels("rulefit-tree-ensemble", _job, builders, builders[0].nModelsInParallel(depths.length), 0  /*no job updates*/);
                     for (int modelId = 0; modelId < depths.length; modelId++) {
                         long startModelTime = System.nanoTime();
-                        SharedTreeModel treeModel = DKV.getGet(builders[modelId].dest());
+                        SharedTreeModel treeModel = trainTreeModel(_parms._algorithm, depths[modelId]);
                         long endModelTime = System.nanoTime() - startModelTime;
                         LOG.info("Tree model n." + modelId + " trained in " + ((double)endModelTime) / 1E9 + "s.");
                         treeModels.add(treeModel);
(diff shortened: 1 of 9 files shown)


h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -454,7 +454,7 @@ public void run() {
    * Each model can override this logic, based on parameters, dataset size, etc.
    * @return How many models to train in parallel during cross-validation
    */
-  protected int nModelsInParallel(int folds) {
+  public int nModelsInParallel(int folds) {
```
honzasterba, line 457: this is only ment for cross validation, not when building the main model

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
```diff
@@ -280,23 +283,29 @@ void fillModelMetrics(RuleFitModel model, GLMModel glmModel) {
...
+                    DKV.put(builders[modelId]._job);
+                    Scope.track_generic(builders[modelId]._job);
+                } else {
+                    // TODO XGB
+                    throw new RuntimeException("Unsupported algorithm for tree building: " + _parms._algorithm);
+                }
             }
-
-            return treeModel;
+           return builders;
```
honzasterba, line 308: indent

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
```diff
@@ -280,23 +283,29 @@ void fillModelMetrics(RuleFitModel model, GLMModel glmModel) {
...
+                    ((SharedTreeModel.SharedTreeParameters) builders[modelId]._parms)._max_depth = depths[modelId];
+                    DKV.put(builders[modelId]._job);
+                    Scope.track_generic(builders[modelId]._job);
+                } else if (algorithm.equals(RuleFitModel.Algorithm.GBM)) {
+                    builders[modelId] = new GBM((GBMModel.GBMParameters) params);
+                    ((SharedTreeModel.SharedTreeParameters) builders[modelId]._parms)._max_depth = depths[modelId];
+                    DKV.put(builders[modelId]._job);
+                    Scope.track_generic(builders[modelId]._job);
+                } else {
+                    // TODO XGB
```
honzasterba, line 304: why is there XGB todo in rulefit?

h2o-core/src/main/java/water/Job.java
```diff
@@ -97,7 +97,7 @@ public Job(Key<T> key, String clz_of_T, String desc, String warningStr) {
   private boolean stopped() { return   _end_time != 0; }
 
   // Simple state accessors; public ones do a DKV update check
-  public long start_time()   { update_from_remote(); assert !created(); return _start_time; }
+  public long start_time()   { update_from_remote();/* assert !created()*/; return _start_time; }
```
honzasterba, line 100: I dont think this is acceptable

h2o-core/src/main/java/hex/ModelBuilder.java
koniecsveta: Yesyes.. discussed this with @michalkurka, that it should be ok to use it in this case. I just think I should put folds = 1. Or do you think I should do some different method for rulefit case based on this one?

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
koniecsveta: because it is planned to use also XGB as algo for rule generation for rulefit but it is not yet done. this TODO was in trainTreeModel() which now won't exist so it's now here in respective place.

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
koniecsveta: fixed

h2o-core/src/main/java/hex/ModelBuilder.java
koniecsveta: now there is `nTreeEnsemblesInParallel(int numDepths)`

h2o-core/src/main/java/water/Job.java
koniecsveta: fixed now

honzasterba (approved):
lgtm

PR #4983, opened 2020-09-30
[r] use RCurl to download jar

honzasterba:
- allows us to set longer timeout

### DIFF
--- a/h2o-r/h2o-package/R/connection.R
+++ b/h2o-r/h2o-package/R/connection.R
@@ -801,8 +801,9 @@ h2o.clusterStatus <- function() {
   # md5_check <- readLines(tcon, n = 1)
   # close(tcon)
   md5_file <- tempfile(fileext = ".md5")
-  download.file(md5_url, destfile = md5_file, mode = "w", cacheOK = FALSE, quiet = TRUE)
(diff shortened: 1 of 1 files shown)


michalkurka (approved):

h2o-r/h2o-package/R/connection.R
```diff
@@ -801,8 +801,9 @@ h2o.clusterStatus <- function() {
   # md5_check <- readLines(tcon, n = 1)
   # close(tcon)
   md5_file <- tempfile(fileext = ".md5")
-  download.file(md5_url, destfile = md5_file, mode = "w", cacheOK = FALSE, quiet = TRUE)
+  .downloadFile(md5_url, dest = md5_file, mode = "w")
   md5_check <- readLines(md5_file, n = 1L)
+    cat(md5_check)
```
michalkurka, line 806: please fix the indentation if you want to leave the print out

honzasterba: removed

ledell (approved):
I did not hear back from CRAN as to whether this would solve the issue that they observed, but I think it should solve the issue, so let's merge it and have it included in the 3.32.0.1 release.

PR #4984, opened 2020-10-01
PUBDEV-7817: Speed up model_summary computations in AutoML Stacked Ensembles in R

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7817 -- R Stacked Ensemble: model_summary - use algorithm from model_id if present, if not get the model and look inside

Helps R implementation of SE. Python doesn't have the same problem - possibly gets model_summary only when requested.

This is a working fix that helps the performance of the `h2o.explain` when stacked ensembles have a lot of base models.

Ideal solution is described in this follow-up [JIRA](https://0xdata.atlassian.net/browse/PUBDEV-7807), this is a quick fix that helps speed up the explanation for small datasets with a lot of base models (titanic, max_runtime_secs = 120 s) by about one fifth (measured only twice).


ledell (approved):
This is great, thanks!

PR #4986, opened 2020-10-01
Compile H2O against servlet 3.1 (used internally in Jetty 9 anyway)

Pscheidl:
As reported by @maurever , we have different servlet API versions inside H2O.

I've taken a look into h2o-jetty-9 and this carries verison 3.1 along. As H2O runs on Jetty 9 completely, there is no point in keeping the tests and webserver i-face on Servlet 3.0.1.

This could ease the setup in IDEs as well. WDYT ?

Screenshot from 2020-10-01 16-36-57

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -67,6 +67,7 @@ defaultHiveExecVersion=1.1.0
 defaultWebserverModule=h2o-jetty-9
 jetty8version=8.2.0.v20160908
 jetty9version=9.4.11.v20180605
+servletApiVersion=3.1.0
 
 # MOJO2 Integration
 mojo2version=2.4.10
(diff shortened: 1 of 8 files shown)


satai (approved):

satai:
It would be nice to have the version in a property but it's a story for an other time I guess.

Pscheidl:
If others approve it, I'll make the property in this very PR @satai .

honzasterba (approved):

michalkurka (approved):

gradle.properties
```diff
@@ -67,6 +67,7 @@ defaultHiveExecVersion=1.1.0
 defaultWebserverModule=h2o-jetty-9
 jetty8version=8.2.0.v20160908
 jetty9version=9.4.11.v20180605
+servletApiVersion=3.1.0
```
honzasterba, line 70: 👍

honzasterba (approved):

PR #4994, opened 2020-10-02
Demonstrate that adaptTestToTrain can lead to memory leaks

michalkurka:


h2o-core/src/test/java/hex/ModelBuilderTest.java
```diff
@@ -231,6 +234,58 @@ public void testScoreReorderedDomain() {
...
+
+      m.score(test).delete();
+
+      Value modelValue = H2O.STORE.get(m._key);
+      if (modelValue != null && killLocalPOJO) {
+        // kill the local POJO (and with it any state that was not propagated to DKV)
+        modelValue.freePOJO();
+      }
+    } finally {
+      if (train != null)
```
honzasterba, line 280: why not use scope?

honzasterba (approved):

tomasfryda (approved):

h2o-core/src/test/java/hex/ModelBuilderTest.java
michalkurka: Because that can hide issues, explicit key management is clearer when you are working with leaks IMHO.

PR #4996, opened 2020-10-04
Fix missing tkinter in py27 small test

tomasfryda:
py27-small test is failing on `master` because of the new explain-related test. The issue is the default backend used by `matplotlib 2`  which is missing dependencies (`TkAgg` depends on `TkInter` (`python-tk` package) which we don't have in Jenkins environment and I don't think it makes sense to add it there since the tests are "headless".)

So the proposed change is to set default backend to `Agg`, which is a non-GUI backend and it is used in `test-py-demos`.

### DIFF
--- a/scripts/jenkins/groovy/defineTestStages.groovy
+++ b/scripts/jenkins/groovy/defineTestStages.groovy
@@ -67,10 +67,6 @@ def call(final pipelineContext) {
       image: "${pipelineContext.getBuildConfig().DOCKER_REGISTRY}/opsh2oai/h2o-3/dev-jdk-8:${pipelineContext.getBuildConfig().DEFAULT_IMAGE_VERSION_TAG}"
     ],
     [
-            stageName: 'Py2.7 Small', target: 'test-pyunit-small', pythonVersion: '2.7',
-            timeoutValue: 90, component: pipelineContext.getBuildConfig().COMPONENT_PY
-    ],
-    [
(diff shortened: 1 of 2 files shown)


scripts/jenkins/groovy/defineTestStages.groovy
```diff
@@ -66,6 +66,10 @@ def call(final pipelineContext) {
       timeoutValue: 10, hasJUnit: false, component: pipelineContext.getBuildConfig().COMPONENT_PY,
       image: "${pipelineContext.getBuildConfig().DOCKER_REGISTRY}/opsh2oai/h2o-3/dev-jdk-8:${pipelineContext.getBuildConfig().DEFAULT_IMAGE_VERSION_TAG}"
     ],
+    [
```
tomasfryda, line 69: Remove before merge

tomasfryda:
With the fix the Py2.7 Small fails only on `pyunit_coxph_concordance.py` and Py3.7 Small on ` pyunit_pubdev_6438_2D_pdp.py` - the same two tests as in `master` branch, i.e., I don't see any negative influence caused by the fix.

Pscheidl (approved):
Thanks.

honzasterba (approved):

satai (approved):

maurever (approved):
Looks good. Thanks!

koniecsveta (approved):

PR #5000, opened 2020-10-05
fix failing pyunit_sklearn_classification_all_estimators and pyunit_s…

koniecsveta:
…klearn_all_estimators

### DIFF
--- a/h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
+++ b/h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
@@ -108,13 +108,15 @@ public class RuleFitModel extends Model<RuleFitModel, RuleFitModel.RuleFitParame
 
     @Override
     public Frame score(Frame fr, String destination_key, Job j, boolean computeMetrics, CFuncRef customMetricFunc) throws IllegalArgumentException {
-        Frame linearTest = new Frame(Key.make("paths_frame" + destination_key));
+        final Frame linearTest = new Frame(Key.make("paths_frame" + destination_key));
         if (fr.vec(_parms._response_column) != null)
             linearTest.add(_parms._response_column,fr.vec(_parms._response_column));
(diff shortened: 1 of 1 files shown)


h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
```diff
@@ -114,7 +114,9 @@ public Frame score(Frame fr, String destination_key, Job j, boolean computeMetri
         if (_parms._weights_column != null && fr.vec(_parms._weights_column) != null)
             linearTest.add(_parms._weights_column,fr.vec(_parms._weights_column));
         
-        Frame adaptFrm = new Frame(fr);//Key.make());
+        Frame adaptFrm = new Frame(Key.make());
```
Pscheidl, line 117: final ?

Pscheidl (approved):

honzasterba (approved):

h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
```diff
@@ -108,13 +108,15 @@ public RuleFitModel(Key<RuleFitModel> selfKey, RuleFitParameters parms, RuleFitO
 
     @Override
     public Frame score(Frame fr, String destination_key, Job j, boolean computeMetrics, CFuncRef customMetricFunc) throws IllegalArgumentException {
```
michalkurka, line 110: Model#score is a public method that is not expected to run in the context of an existing Scope. it is called from REST API (ModelMetricsHandler).

You can only use Scope if you are also managing it, this code doesn't appear to do that.

Same issue on line 133.

michalkurka (changes requested):
Please see inline comments

michalkurka:
Fix proposed here: https://github.com/h2oai/h2o-3/pull/5007

PR #5001, opened 2020-10-05
Fix __grab_values method in 2D PDP plot

maurever:
Fix failing pyunit_pubdev_6438_2D_pdp.py on the master branch

It is probably only a bug where `xlab` and `labels` values were mistaken. In  `labels`, there are not unique values, so the error is thrown. Interestingly it starts to fail right now. Probably the method `set_yticklabels` respected nonunique values before.


honzasterba (approved):

tomasfryda (approved):

michalkurka (approved):

maurever:
Unrelated tests failed, so I will merge this. Thanks!

PR #5002, opened 2020-10-05
PUBDEV-7825 failing coxph performance() with event that is factor

satai:
failing coxph performance() with event that is factor

Reason - in such a case CoxPH model output thinks about itself that it's an output of an classification lagorithm. (An method from parent is used and if `event_column` is factor it's logic returns `true`)


satai:
cc @grigory93

michalkurka (approved):

PR #5005, opened 2020-10-05
Add code examples for all the new explainability plotting functions in Python and R (PUBDEV-7824, PUBDEV-7821)

tomasfryda:
I added examples to the python and R docstrings for the public functions/methods. 

Related JIRAs:
https://0xdata.atlassian.net/browse/PUBDEV-7824
https://0xdata.atlassian.net/browse/PUBDEV-7821

### DIFF
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -95,7 +95,8 @@ class Description:
                                     "By default, the models and variables are ordered by their similarity.",
         model_correlation_heatmap="Model correlation matrix shows correlation between prediction of the models. "
                                   "For classification, frequency of same predictions is used. By default, models "
-                                  "are ordered by their similarity.",
+                                  "are ordered by their similarity. Interpretable models, such as GAM, GLM, and RuleFit"
+                                  " are highlighted using red colored text.",
         shap_summary="SHAP summary plot shows contribution of features for each instance. The sum "
                      "of the feature contributions and the bias term is equal to the raw prediction "
                      "of the model, i.e., prediction before applying inverse link function.",
@@ -550,24 +551,19 @@ def shap_summary_plot(
     >>>
     >>> h2o.init()
     >>>
-    >>> # Import the titanic dataset into H2O:
-    >>> titanic = h2o.import_file("https://h2o-public-test-data.s3.amazonaws.com/smalldata/titanic/titanic_expanded.csv")
+    >>> # Import the wine dataset into H2O:
+    >>> f = "https://h2o-public-test-data.s3.amazonaws.com/smalldata/wine/winequality-redwhite-no-BOM.csv"
+    >>> df = h2o.import_file(f)
     >>>
-    >>> # Set the predictors and the response
-    >>> predictors = [
-    >>>    "pclass", "sex", "age", "sibsp", "parch", "ticket",
-    >>>    "fare", "cabin", "embarked", "home.dest",
-    >>>    "cabin_type", "family_size", "family_type", "title"
-    >>> ]
-    >>> response = "survived"
-    >>> titanic[response] = titanic[response].asfactor()
+    >>> # Set the response
+    >>> response = "quality"
     >>>
     >>> # Split the dataset into a train and test set:
-    >>> train, test = titanic.split_frame([.75])
+    >>> train, test = df.split_frame([.75])
     >>>
     >>> # Train a GBM
     >>> gbm = H2OGradientBoostingEstimator()
-    >>> gbm.train(x=predictors, y=response, training_frame=train)
+    >>> gbm.train(y=response, training_frame=train)
     >>>
     >>> # Create SHAP summary plot
     >>> gbm.shap_summary_plot(test)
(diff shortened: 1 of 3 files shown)


ledell:
@tomasfryda I noticed in [`def explain`](https://github.com/h2oai/h2o-3/blob/32b0a826594081d0b9ddf6ddcb89cbacf5c675bd/h2o-py/h2o/explanation/_explain.py#L1942) and [`def explain_row`](https://github.com/h2oai/h2o-3/blob/32b0a826594081d0b9ddf6ddcb89cbacf5c675bd/h2o-py/h2o/explanation/_explain.py#L2214), that it says `models` is `H2OAutoML object or H2OModel`.  Should we add "list of models" here?  This is officially supported, right?


```
    :param models: H2OAutoML object or H2OModel
```

ledell:
It's helpful in examples to use generic arg names as possible to make it easy for people to swap out our dataset with theirs.  Here's some changes I would make:

```
titanic = h2o.import_file("https://h2o-public-test-data.s3.amazonaws.com/smalldata/titanic/titanic_expanded.csv")
```
becomes:
```
f = "https://h2o-public-test-data.s3.amazonaws.com/smalldata/titanic/titanic_expanded.csv"
df = h2o.import_file(f)
```
If you can avoid hardcoding the predictor columns, I think that would be a slight improvement too.  (They are hardcoded in Py docstrings but not R).

Also, since we are only trying to demo `h2o.explain()` instead of AutoML, I would lower the number of models in the AutoML run to something minimal (where SE still wins).  Maybe for consistency, we could just use the same exact example from the User Guide.  The Wine example?

tomasfryda:
> @tomasfryda I noticed in [`def explain`](https://github.com/h2oai/h2o-3/blob/32b0a826594081d0b9ddf6ddcb89cbacf5c675bd/h2o-py/h2o/explanation/_explain.py#L1942) and [`def explain_row`](https://github.com/h2oai/h2o-3/blob/32b0a826594081d0b9ddf6ddcb89cbacf5c675bd/h2o-py/h2o/explanation/_explain.py#L2214), that it says `models` is `H2OAutoML object or H2OModel`. Should we add "list of models" here? This is officially supported, right?
> 
> ```
>     :param models: H2OAutoML object or H2OModel
> ```

Yes, they are supported, I will fix that tomorrow.

tomasfryda:
> It's helpful in examples to use generic arg names as possible to make it easy for people to swap out our dataset with theirs. Here's some changes I would make:
> 
> ```
> titanic = h2o.import_file("https://h2o-public-test-data.s3.amazonaws.com/smalldata/titanic/titanic_expanded.csv")
> ```
> 
> becomes:
> 
> ```
> f = "https://h2o-public-test-data.s3.amazonaws.com/smalldata/titanic/titanic_expanded.csv"
> df = h2o.import_file(f)
> ```
> 
> If you can avoid hardcoding the predictor columns, I think that would be a slight improvement too. (They are hardcoded in Py docstrings but not R).
> 
> Also, since we are only trying to demo `h2o.explain()` instead of AutoML, I would lower the number of models in the AutoML run to something minimal (where SE still wins). Maybe for consistency, we could just use the same exact example from the User Guide. The Wine example?

I can lower the amount of the models in AutoML, but it is good to note that `varimp_heatmap` doesn't work for stacked ensembles so I would go with at least 7 models, so it looks nice (5 models in the `varimp_heatmap`).

ledell:
R CMD check is currently failing with a valid error:

```
[2020-10-05T21:37:08.996Z] > pdp <- h2o.pd_multi_plot(aml, test, column = "age")
[2020-10-05T21:37:08.996Z] Error in h2o.pd_multi_plot(aml, test, column = "age") : 
[2020-10-05T21:37:08.996Z]   Column was not found in the provided data set!
[2020-10-05T21:37:08.996Z] Execution halted
```

The dataset was updated to Wine data, but still using Titanic colname.  I'll fix.

ledell (approved):
Tests passing now, thanks!

h2o-py/h2o/explanation/_explain.py
```diff
@@ -876,7 +923,30 @@ def pd_plot(
...
+    >>>
+    >>> # Split the dataset into a train and test set:
+    >>> train, test = df.split_frame([.75])
+    >>>
+    >>> # Train a GBM
+    >>> gbm = H2OGradientBoostingEstimator()
+    >>> gbm.train(y=response, training_frame=train)
+    >>>
+    >>> # Create partial dependence plot
+    >>> gbm.pd_plot(test, column="age")
```
ledell:
```suggestion
    >>> gbm.pd_plot(test, column="alcohol")
```

h2o-py/h2o/explanation/_explain.py
```diff
@@ -974,7 +1044,30 @@ def pd_multi_plot(
...
+    >>>
+    >>> # Split the dataset into a train and test set:
+    >>> train, test = df.split_frame([.75])
+    >>>
+    >>> # Train an H2OAutoML
+    >>> aml = H2OAutoML(max_models=7)
+    >>> aml.train(y=response, training_frame=train)
+    >>>
+    >>> # Create a partial dependence plot
+    >>> aml.pd_multi_plot(test, column="age")
```
ledell:
```suggestion
    >>> aml.pd_multi_plot(test, column="alcohol")
```

ledell (approved):
I just made a few changes to the pydocs (one bug, one two edits to match R).

ledell:
Tests already passed.  Thanks @tomasfryda!

PR #5006, opened 2020-10-05
PUBDEV-7541: Update the AutoML description doc string

tomasfryda:
I updated the doc strings with what was provided in the JIRA (https://0xdata.atlassian.net/browse/PUBDEV-7541).
The only difference is that I used part of the last sentence from python's description that describes the SEs:
```
AutoML then trains two Stacked Ensemble models, one of all the models, and one of only the best models of each kind.
```


ledell (approved):
Looks good!

PR #5013, opened 2020-10-06
Improve varimp consolidation

tomasfryda:
An attempt to remove the dependency on `newdata` for `h2o.varimp_heatmap`. 

```
Consolidation works in the following way:
  1. if varimp variable is in x => add it to consolidated_varimps
  2. for all remaining varimp variables:
      1. find the longest prefix of varimp variable that is in x and add it to the consolidated varimp
      2. if there was no match, throw an error
  3. normalize the consolidated_varimps so they sum up to 1
```

Downsides:
if we have a feature `ab` with category `cd.ef` and another feature `ab.cd` that is also categorical then it is impossible with the amount of information that the algo has to decide to which variable the varimp `ab.cd.ef` should be added to.

When I would add the information from the dataset then the amount of possible errors would decrease to situations like:
if we have a feature `ab` with category `cd.ef` and another feature `ab.cd` that is also categorical and has category `ef` then we are unable to decide to which variable varimp `ab.cd.ef` belongs to.

### DIFF
--- a/h2o-docs/src/product/explain.rst
+++ b/h2o-docs/src/product/explain.rst
@@ -343,12 +343,12 @@ Variable importance heatmap shows variable importance across multiple models. So
 .. tabs::
    .. code-tab:: r R
 
-        va_plot <- h2o.varimp_heatmap(aml, test)
+        va_plot <- h2o.varimp_heatmap(aml)
         va_plot
 
    .. code-tab:: python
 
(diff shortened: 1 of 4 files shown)


ledell (approved):
This looks good to me.  I'd rather remove the dependency on `newdata` to start and tolerate the edge cases and then add the required data/categorical metadata to the model so we can easily retrieve it from there (in a fix release).

PR #5017, opened 2020-10-06
Fix rendering of the factor levels

tomasfryda:
`h2o.shap_explain_row_plot` rendered level index instead of the actual level.


ledell (approved):
Great catch, thanks!

PR #5019, opened 2020-10-06
Update descriptions in the explainability API

tomasfryda:
Copied new version of the strings from the user guide.

I have one question about R/Python residual analysis and that is if we want to have the following sentence there. I assume that was in the user guide because of the wine data set. I have nothing against keeping it here but it's better to ask.
```
Note that if you see "striped" lines of residuals, that is an artifact of having an integer valued (vs a real valued) response variable.
```


ledell:
@tomasfryda I'm ok to delete it or keep it... Since it's already there and working, we can keep it for now!  Thanks for checking.

ledell (approved):

PR #5024, opened 2020-10-07
[xgboost] update excluded libs after 1.2.0 upgrade

honzasterba:


michalkurka (approved):

mn-mikke (approved):
LGTM, thanks @honzasterba ! Also confirm that SW compiles against this branch.

PR #5026, opened 2020-10-07
Fix race condition in TransformWrappedVec

michalkurka:
2 threads trying to ini the chunks could end-up in inconsistent state

### DIFF
--- a/h2o-core/src/main/java/water/fvec/TransformWrappedVec.java
+++ b/h2o-core/src/main/java/water/fvec/TransformWrappedVec.java
@@ -32,7 +32,7 @@ import water.rapids.ast.params.AstNum;
 public class TransformWrappedVec extends WrappedVec {
 
   private final Key<Vec>[] _masterVecKeys;
-  private transient Vec[] _masterVecs;
+  private volatile transient Vec[] _masterVecs;
   private final TransformFactory<?> _tf;
 
   public TransformWrappedVec(Key<Vec> key, int rowLayout, TransformFactory<?> fact, Key<Vec>[] masterVecKeys) {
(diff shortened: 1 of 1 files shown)


h2o-core/src/main/java/water/fvec/TransformWrappedVec.java
```diff
@@ -75,11 +75,24 @@ public Vec makeVec() {
...
 
   @Override public Chunk chunkForChunkIdx(int cidx) {
-    Chunk[] cs = new Chunk[_masterVecKeys.length];
-    if( _masterVecs==null )
-      _masterVecs = new Vec[_masterVecKeys.length];
-    for(int i=0; i<cs.length;++i)
-      cs[i] = (_masterVecs[i]!=null?_masterVecs[i]:(_masterVecs[i] = _masterVecKeys[i].get())).chunkForChunkIdx(cidx);
+    Vec[] masterVecs = _masterVecs;
+    if (masterVecs == null) {
+      masterVecs = new Vec[_masterVecKeys.length];
```
Pscheidl, line 80: I'd suggest requiring the keys not to be null in the constructor.

```patch
Index: h2o-core/src/main/java/water/fvec/TransformWrappedVec.java
IDEA additional info:
<+>UTF-8
--- h2o-core/src/main/java/water/fvec/TransformWrappedVec.java	(revision 9c6e827a7b9f3a8e9539c231ce6bb9d0441e54b8)
+++ h2o-core/src/main/java/water/fvec/TransformWrappedVec.java	(date 1602086403059)
@@ -6,6 +6,8 @@
 import water.rapids.Env;
 import water.rapids.ast.params.AstNum;
 
+import java.util.Objects;
+
 /**
  * This wrapper pushes a transform down into each chunk so that
  * transformations will happen on-the-fly. When wrapped and there
@@ -37,6 +39,7 @@
 
   public TransformWrappedVec(Key<Vec> key, int rowLayout, TransformFactory<?> fact, Key<Vec>[] masterVecKeys) {
     super(key, rowLayout, null);
+    Objects.requireNonNull(masterVecKeys);
     _tf = fact;
     _masterVecKeys = masterVecKeys;
     DKV.put(this);
```

Pscheidl (approved):

PR #5029, opened 2020-10-07
Fix K8S service name propagation

Pscheidl:
Problem: Service name was different than the name propagated into `H2O_KUBERNETES_SERVICE_DNS` env var. It should've been full name, but it's only chart name.

Also gave the service name `-service` postfix. Easy to find it on various logs and Helm takes care about naming anyway.

### DIFF
--- a/h2o-helm/templates/ingress.yaml
+++ b/h2o-helm/templates/ingress.yaml
@@ -35,7 +35,7 @@ spec:
           {{- range .paths }}
           - path: {{ . }}
             backend:
-              serviceName: {{ $fullName }}-service
+              serviceName: {{ $fullName }}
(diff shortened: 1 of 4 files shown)


Pscheidl:
Before this PR is merged, Kubernetes must be removed from smoke tests.

michalkurka (approved):

Pscheidl:
Ok, so no. The service prefix will be deleted. One reason - helm strips the name automatically to 63 characters and for extra long names, the extra `-service` could cause failures. But the fix will remain.

honzasterba (approved):

Pscheidl:
Works

Pscheidl:
K8S from smoke tests removed.

PR #5032, opened 2020-10-08
AstUnique: abandon using GroupBy-style implementation

michalkurka:
Reworked implementation doesn't rely on mutable structures and is able
to achieve 30%-50% speed-up on 10m of unique data points.

### DIFF
--- a/h2o-core/src/main/java/water/fvec/task/UniqOldTask.java
+++ b/h2o-core/src/main/java/water/fvec/task/UniqOldTask.java
@@ -1,0 +1,53 @@
+package water.fvec.task;
+
+import water.MRTask;
+import water.fvec.Chunk;
+import water.fvec.Vec;
+import water.rapids.ast.prims.mungers.AstGroup;
+import water.util.IcedHashMap;
+
+@Deprecated // in favor of UniqTask - kept for experimenting only
+public class UniqOldTask extends MRTask<UniqOldTask> {
+  public IcedHashMap<AstGroup.G, String> _uniq;
+
+  @Override
+  public void map(Chunk[] c) {
+    _uniq = new IcedHashMap<>();
+    AstGroup.G g = new AstGroup.G(1, null);
+    for (int i = 0; i < c[0]._len; ++i) {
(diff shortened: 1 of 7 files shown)


h2o-core/src/main/java/water/fvec/task/UniqOldTask.java
```diff
@@ -0,0 +1,53 @@
+package water.fvec.task;
+
+import water.MRTask;
+import water.fvec.Chunk;
+import water.fvec.Vec;
+import water.rapids.ast.prims.mungers.AstGroup;
+import water.util.IcedHashMap;
+
+@Deprecated // in favor of UniqTask - kept for experimenting only
+public class UniqOldTask extends MRTask<UniqOldTask> {
```
Pscheidl, line 10: I'd kill it right away. No need to experiment with it. If this is to demonstrate a technical concept, we're better rename & move it. No big deal though.

h2o-core/src/main/java/water/fvec/task/UniqOldTask.java
```diff
@@ -0,0 +1,53 @@
...
+        l = r;
+        r = _uniq;
+      }  // larger on the left
+      for (AstGroup.G rg : r.keySet()) l.putIfAbsent(rg, "");  // loop over smaller set
+      _uniq = l;
+      t._uniq = null;
+    }
+  }
+
+  public Vec toVec() {
```
Pscheidl, line 39: :+1: :pray:

Pscheidl (approved):

h2o-core/src/main/java/water/fvec/task/UniqOldTask.java
michalkurka: Yeah, I know what you mean... I kept it for experiments for our support team. Given the current uncertainty whether there is a possible issue with unique implementation - I want to give power-users a chance to experiment with old vs new on a nightly build and be able to easily prove there is an improvement. I will remove once these experiments are done.

PR #5034, opened 2020-10-08
[PUBDEV-7793] implement deserialization of monotone constraints

honzasterba:
…m in MOJO

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
@@ -1,0 +1,21 @@
+package hex.genmodel.attributes.parameters;
+
+public class KeyValue {
+    
+    public final String key;
+    public final double value;
+
+    public KeyValue(String key, double value) {
+        this.key = key;
+        this.value = value;
+    }
+
+    public String getKey() {
+        return key;
+    }
+
+    public double getValue() {
+        return value;
+    }
+
+}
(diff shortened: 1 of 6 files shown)


Pscheidl (approved):

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
```diff
@@ -0,0 +1,21 @@
+package hex.genmodel.attributes.parameters;
+
+public class KeyValue {
```
mn-mikke, line 3: Can't we use `water.Iced.KeyValue` instead or is there any limitation?

h2o-test-support/src/main/java/water/TestUtil.java
```diff
@@ -1371,4 +1374,21 @@ public Vec createRandomBinaryWeightsVec(final long len, final int randomSeed) {
     return weightsVec;
   }
 
+  public static GenModel toMojo(Model model, String testName, boolean readModelMetaData) {
+    final String filename = testName + ".zip";
+    StreamingSchema ss = new StreamingSchema(model.getMojo(), filename);
+    try {
+      FileOutputStream os = new FileOutputStream(ss.getFilename());
+      ss.getStreamWriter().writeTo(os);
```
mn-mikke, line 1382: if this line throws an exception, does `os` get closed?

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
Pscheidl: This is MOJO

h2o-test-support/src/main/java/water/TestUtil.java
honzasterba: good catch, improved

mn-mikke (approved):
LGTM

Pscheidl (approved):

PR #5036, opened 2020-10-08
rulefit demo [nocheck]

koniecsveta:


h2o-py/demos/rulefit_demo.ipynb
```diff
@@ -0,0 +1,1544 @@
...
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "import h2o\n",
+    "from h2o.estimators import H2ORuleFitEstimator, H2ORandomForestEstimator\n",
+    "\n",
+    "# init h2o cluster\n",
+    "h2o.init(strict_version_check=False, url=\"http://192.168.59.147:54321\")"
```
honzasterba, line 113: would keep just `h2o.init()`

honzasterba (approved):
minor comment
what about adding an example flow as well?

honzasterba:
this should probably go to release branch

koniecsveta:
Hm I've noticed a problem with displayment of long rules, they are displayed like this:
`(sex in {female}) & (sibsp < 3.5 or sibsp is NA) & (pclass in {1, ...`
hence not readable..  checked also in flow and there the displayment is ok...

michalkurka:
Comparing the model performance can be tricky but for the purposes of this demo (which is understood to be just an illustration of the concept) you should use the test dataset (that you don't use in any part of the training) instead of performance on the training frame.

koniecsveta:
will remove this comparison

michalkurka (approved):
thank you

PR #5037, opened 2020-10-08
Instruct users to helm install latest version in the repository on do…

Pscheidl:
…wnload page


h2o-dist/index.html
```diff
@@ -682,7 +682,7 @@ <h2>Setup H<sub>2</sub>O on Kubernetes using Helm</h2>
                     <p><a href="https://helm.sh/" target="_blank">Helm</a> can be used to deploy H2O into a kubernetes cluster. Helm requires the KUBECONFIG environment variable to be set up properly, or stating the kubeconfig destination explicitly. Please refer to Helm's documentation
                     for further information.</p>
                 <p class="terminal" id="to_copy11">
-helm repo add h2o https://charts.h2o.ai --version SUBST_PROJECT_VERSION
+helm repo add h2o https://charts.h2o.ai
```
Pscheidl, line 685: The version would be substituted for `3.33.0.1` and similar, while Helm charts don't have the leader `3.x` and have to follow semver2.

Also, on the download page, there would always be latest version anyway. Omitting the version gives our users the latest version.

honzasterba (approved):

PR #5038, opened 2020-10-08
[PUBDEV-7819] fix counting of non zero rows in chunks

honzasterba:
https://0xdata.atlassian.net/browse/PUBDEV-7819

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
@@ -13,6 +13,7 @@ import water.fvec.Vec;
 import water.util.VecUtils;
 
 import java.util.Arrays;
+import java.util.Optional;
 
 import static water.H2O.technote;
 import static water.MemoryManager.malloc4f;
@@ -70,7 +71,7 @@ public class XGBoostUtils {
         final Vec weightVec = frame.vec(weight);
         final Vec offsetsVec = frame.vec(offset);
         final int[] nRowsByChunk = new int[chunks.length];
-        final long nRowsL = sumChunksLength(chunks, responseVec, weightVec, nRowsByChunk);
+        final long nRowsL = sumChunksLength(chunks, responseVec, Optional.ofNullable(weightVec), nRowsByChunk);
         if (nRowsL > Integer.MAX_VALUE) {
             throw new IllegalArgumentException("XGBoost currently doesn't support datasets with more than " +
                     Integer.MAX_VALUE + " per node. " +
@@ -102,18 +103,22 @@ public class XGBoostUtils {
     }
 
     /**
-     * Counts a total sum of chunks inside a vector. Only chunks present in chunkIds are considered.
+     * Counts a total sum of chunks inside a vector. Only chunks present in chunkIds are counted.
+     * If a weights vector is provided, only rows with non-zero weights are counted.
      *
-     * @param chunkIds Chunk identifier of a vector
+     * @param chunkIds Chunk ids to consider during the calculation. Chunks IDs not listed are not included.
      * @param vec      Vector containing given chunk identifiers
-     * @param weightsVector Vector with row weights, possibly null
+     * @param weightsVector Vector with row weights, possibly an empty optional
+     * @param chunkLengths Array of integers where the lengths of the individual chunks will be added. Initialization to an array of 0's is expected.
      * @return A sum of chunk lengths. Possibly zero, if there are no chunks or the chunks are empty.
      */
-    public static long sumChunksLength(int[] chunkIds, Vec vec, Vec weightsVector, int[] chunkLengths) {
+    public static long sumChunksLength(int[] chunkIds, Vec vec, Optional<Vec> weightsVector, int[] chunkLengths) {
+        assert chunkLengths.length == chunkIds.length;
+        
         for (int i = 0; i < chunkIds.length; i++) {
             final int chunk = chunkIds[i];
-            if (weightsVector != null) {
-                Chunk weightVecChunk = weightsVector.chunkForChunkIdx(chunk);
+            if (weightsVector.isPresent()) {
+                final Chunk weightVecChunk = weightsVector.get().chunkForChunkIdx(chunk);
                 int nzIndex = 0;
                 do {
                     if (weightVecChunk.atd(nzIndex) != 0) chunkLengths[i]++;
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
@@ -112,20 +112,17 @@ public class XGBoostUtils {
     public static long sumChunksLength(int[] chunkIds, Vec vec, Vec weightsVector, int[] chunkLengths) {
         for (int i = 0; i < chunkIds.length; i++) {
             final int chunk = chunkIds[i];
-            chunkLengths[i] = vec.chunkLen(chunk);
-            if (weightsVector == null)
-                continue;
-
-            Chunk weightVecChunk = weightsVector.chunkForChunkIdx(chunk);
-            if (weightVecChunk.atd(0) == 0) chunkLengths[i]--;
-            int nzIndex = 0;
-            do {
-                nzIndex = weightVecChunk.nextNZ(nzIndex, true);
-                if (nzIndex < 0 || nzIndex >= weightVecChunk._len) break;
-                if (weightVecChunk.atd(nzIndex) == 0) chunkLengths[i]--;
-            } while (true);
+            if (weightsVector != null) {
+                Chunk weightVecChunk = weightsVector.chunkForChunkIdx(chunk);
+                int nzIndex = 0;
+                do {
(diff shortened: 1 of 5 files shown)


Pscheidl (approved):

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -112,20 +112,17 @@ else if (f.vec(i-catCols).isInt())
...
-            Chunk weightVecChunk = weightsVector.chunkForChunkIdx(chunk);
-            if (weightVecChunk.atd(0) == 0) chunkLengths[i]--;
-            int nzIndex = 0;
-            do {
-                nzIndex = weightVecChunk.nextNZ(nzIndex, true);
-                if (nzIndex < 0 || nzIndex >= weightVecChunk._len) break;
-                if (weightVecChunk.atd(nzIndex) == 0) chunkLengths[i]--;
-            } while (true);
+            if (weightsVector != null) {
+                Chunk weightVecChunk = weightsVector.chunkForChunkIdx(chunk);
```
Pscheidl, line 116: For @michalkurka :  Here, the total amount of all elements in the chunks is loaded.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -112,20 +112,17 @@ else if (f.vec(i-catCols).isInt())
...
+                int nzIndex = 0;
+                do {
+                    if (weightVecChunk.atd(nzIndex) != 0) chunkLengths[i]++;
```
Pscheidl, line 119: For @michalkurka Here, only the zero elements should be subtracted. However, if you have a careful look, the logic here (introduced by me originally), it only subtracts the non-zero elements. Which means the final count is WRONG. (in the original code, honza's code does it vice versa and it works).

Pscheidl:
I have troubles reproducing this locally, even with the provided python script. With dense matrix.

Pscheidl:
Couldn't reproduce locally, even with the python script provided.

It worked for dense chunks, as the implementation of `nextNz` is actually this:

```java
public int nextNZ(int rid, boolean onlyTrueZero) { return rid + 1; }
```

and this actually forces the existing code to check every row for zeros. However, if there is a sparse Chunk `CXIChunk`, the implementation looks as follows:

```java
  @Override public final int nextNZ(int rid, boolean onlyTrueZero) {
    return onlyTrueZero && _isNA ? rid + 1 : nextNZ(rid);
  }
```

This suggests the key to reproducibility is creating this sparse chunk.

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostUtilsTest.java
```diff
@@ -314,6 +309,42 @@ public void testSparsematrixInit_dimensions_test() throws XGBoostError {
...
+        assertEquals(dataVec.length(), nonZeroRows);
+
+        // With all weights as 1, all rows should be counted
+        final Vec allWeightsOne = Vec.makeCon(1d, 1000);
+        localChunksLengths = new int[localChunksLengths.length];
+        long weightedNonZeroRows = XGBoostUtils.sumChunksLength(localChunkIds, dataVec, Optional.of(allWeightsOne), localChunksLengths);
+        assertEquals(dataVec.length(), weightedNonZeroRows);
+        
+        // Mixed weights - only the non-zero weight rows should be counted
+        final Vec mixedWeights = Vec.makeCon(0d, 1000);
```
Pscheidl, line 333: This part tests the root cause of this problem:

All chunks with one exception have trivial implementation of `nextNz` - just give the next element. Might not always be correct, but that's how it works.

```java
public int nextNZ(int rid){ return rid + 1;}
```

The problem was simple - the original code, while seemingly having bad logic, worked in most cases, as it inspected all chunks one by one anyway, due to the implementation of `nextNz` in most chunks.

The problem arises with `CXIChunk` - the zero-sparse variant. It's `nextNz` method is implemented as follows:

```java
  @Override public final int nextNZ(int i){
    int x = findOffset(i);
    if(x < 0) x = -x-1-_elem_sz;
    _previousOffset = x += _elem_sz;
    return getId(x);
  }
```

This made the original logic return nonsensical numbers, as it actually skipped values.

I was unable to reproduce this behavior every time with the given test (see https://0xdata.atlassian.net/browse/PUBDEV-7819), probably due to different chunking of the dataset on my machine. Anyway, this unit test simulates the bad behavior all the time and actually tests the behavior is correct.

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostUtilsTest.java
```diff
@@ -688,6 +719,16 @@ private static void comparePreds(final float[][] nativePreds, final Vec h2oPreds
   @RunWith(Parameterized.class)
   public static final class XGBoostSparseMatrixTest extends XGBoostUtilsTest {
 
+    @BeforeClass
+    public static void beforeClass(){
```
Pscheidl, line 723: No need to run both test classes as parametrized, when only one actually is parametrized.

h2o-test-support/src/main/java/water/TestUtil.java
```diff
@@ -1361,9 +1359,9 @@ public static void writeFrameToCSV(String fileNameWithPath, Frame h2oframe, bool
    * @param randomSeed Seed for the random generator (for reproducibility)
    * @return An instance of {@link Vec} with binary weights (either 0.0D or 1.0D, nothing in between).
    */
-  public Vec createRandomBinaryWeightsVec(final long len, final int randomSeed) {
+  public static Vec createRandomBinaryWeightsVec(final long len, final int randomSeed) {
     final Vec weightsVec = Vec.makeZero(len, Vec.T_NUM);
-    final Random random = new Random(randomSeed);
+    final Random random = new RandomUtils.H2ORandomRNG(randomSeed);
```
Pscheidl, line 1364: @koniecsveta Please review, found this "on the go" - a minor thing.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
```diff
@@ -102,30 +103,31 @@ else if (f.vec(i-catCols).isInt())
...
-            Chunk weightVecChunk = weightsVector.chunkForChunkIdx(chunk);
-            if (weightVecChunk.atd(0) == 0) chunkLengths[i]--;
-            int nzIndex = 0;
-            do {
-                nzIndex = weightVecChunk.nextNZ(nzIndex, true);
-                if (nzIndex < 0 || nzIndex >= weightVecChunk._len) break;
-                if (weightVecChunk.atd(nzIndex) == 0) chunkLengths[i]--;
-            } while (true);
+            if (weightsVector.isPresent()) {
+                final Chunk weightVecChunk = weightsVector.get().chunkForChunkIdx(chunk);
```
michalkurka, line 121: Can there be empty chunks?

h2o-test-support/src/main/java/water/TestUtil.java
michalkurka, line 1364: I would rather go through the factory method `getRNG` - I know this is just a test but it is better to follow the good patterns here as well to prevent proliferating this code into the production codebase.

Pscheidl: Definitely a good idea. Fixed !

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostUtils.java
Pscheidl: Great catch. What do you mean by empty chunks ? If you mean both empty, than this will work: I've added a test to demonstrate it. https://github.com/h2oai/h2o-3/pull/5038/commits/7de743c10ab986f076129385854cde456692f281

It is assumed both vectors have the same layout and thus both chunks are of the same size (0).

michalkurka: The test works but only coincidentally. Empty chunks are typically represented by a constant chunk (eg C0LChunk) - these don't have any range checks - just return a constant, this approach thus has 2 issues:

- constant empty chunk can be represented by any constant - it is only a coincidence it is 0,
- there can be cases where empty chunks will not be represented by a constant chunk.

I will push a test that shows the issue.

Pscheidl: Added a trivial fix + added one more variant of the test.

My schedule is full of meetings now, but I"ll try to make this process parallel as more refined and push that variant as well (we can always go back).

michalkurka: I think this fix is fine - no need for parallelization. This is not where we can gain much time.

michalkurka (approved):
Thank you for fixing it @honzasterba and @Pscheidl

Pscheidl:
Waiting for tests: http://mr-0xg1:8080/job/h2o-3-xgb-smoke-pipeline/job/honza%2FPUBDEV-7819%2Fxgb_assertion_error/16/

PR #5042, opened 2020-10-09
Rulefit: airlines demo

koniecsveta:
this is basic "short-training-time parametrization", will run some "longer-training-time" in case they would give nicer rules, but that will take time


koniecsveta:
This PR contains airlines demo run used for gartner 2021 demo purposes. Could be merged or could be closed.

PR #5044, opened 2020-10-12
fix rounding error in effective_parameters tests

koniecsveta:


h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -4357,3 +4357,9 @@ def assertCoefEqual(regCoeff, coeff, coeffClassSet, tol=1e-6):
         diff = abs(val1-val2)
         print("val1: {0}, val2: {1}, tol: {2}".format(val1, val2, tol))
         assert diff < tol, "diff {0} exceeds tolerance {1}.".format(diff, tol)
+
+def almost_equal(x, y, tol=1e-6):
```
Pscheidl, line 4361: Just name `is_equal` please.

Pscheidl (approved):

h2o-py/tests/pyunit_utils/utilsPY.py
Pscheidl: Equals is probably better.

Pscheidl: Also, limit the arg type if possible in python.

koniecsveta: fixed

Pscheidl (approved):

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -4357,3 +4357,12 @@ def assertCoefEqual(regCoeff, coeff, coeffClassSet, tol=1e-6):
         diff = abs(val1-val2)
         print("val1: {0}, val2: {1}, tol: {2}".format(val1, val2, tol))
         assert diff < tol, "diff {0} exceeds tolerance {1}.".format(diff, tol)
+
+def equals(x, y, tol=1e-6):
+    if isinstance(x, (int, float)) & isinstance(y, (int, float)):
+        if abs(x-y) > tol:
+            assert False, "Value {0} and value {1} do not agree.".format(x, y)
```
satai, line 4364: `agree` seems not to be the right word here?

satai: OH, too late, never mind.

satai:
btw you can `import numpy.testing` (already available, no need to make a new image to get it) that has `assert_almost_equal` that works for arrays too.

PR #5049, opened 2020-10-13
PUBDEV-7836: Missing syntax highlight Random Grid search Python example

hannah-tillman:


maurever (approved):
Thanks @hannah-tillman  we will see. :)

PR #5050, opened 2020-10-15
PUBDEV-7739: feature interactions and importance for XGB and GBM

koniecsveta:
!!! TODO dont forget to rebase onto rel branch in the end


h2o-core/src/main/java/hex/Model.java
```diff
@@ -2997,4 +2997,8 @@ protected boolean isFeatureUsedInPredict(int featureIdx) {
   public boolean isDistributionHuber() {
     return _parms._distribution == DistributionFamily.huber;
   }
+
+  public TwoDimTable[][] getFeatureInteractionsTable(int maxInteractionDepth, int maxTreeDepth, int maxDeepening) {
+    return null;
```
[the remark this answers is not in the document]
koniecsveta: FeatureInteractionsCollector is in h2o-algos and it has `Map<String, FeatureInteraction> getFeatureInteractions(...`  declared (so it can't be moved to core), not `TwoDimTable[][] getFeatureInteractionsTable(...`, so it can't be used like this. I will use H2O.unimpl()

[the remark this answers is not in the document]
honzasterba: Then the FeatureInteractionsCollector ifc serves no purpose and can be removed

any reason it is in h2o-algos? I think it would make sense to either move everything to core or to algos so the code can be cleaner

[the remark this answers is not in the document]
koniecsveta: As a next step, I will do the same also for gbm but the calculation will be different. Thats why I did FeatureInteractionsCollector, but can remove now. Not sure about moving everything to core, the topic seems to me h2o-algo like, thats why I put it to algo. I can create interface in core so that `Model.java#getFeatureInteractionsTable(.. `is not needed

h2o-core/src/main/java/hex/FeatureInteractions.java
```diff
@@ -0,0 +1,223 @@
...
+
+import org.apache.commons.lang.mutable.MutableInt;
+import water.util.TwoDimTable;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class FeatureInteractions {
+    
+    private HashMap<String, FeatureInteraction> map;
```
honzasterba, line 11: can be final?

h2o-core/src/main/java/hex/FeatureInteractions.java
```diff
@@ -0,0 +1,223 @@
...
+
+    public FeatureInteractions() {
+        this.map = new HashMap<>();
+    }
+
+    public static void merge(FeatureInteractions leftFeatureInteractions, FeatureInteractions rightFeatureInteractions) {
```
honzasterba, line 17: why static? syntax fi.mergeWith(other) looks nicer

h2o-core/src/main/java/hex/FeatureInteractions.java
```diff
@@ -0,0 +1,223 @@
...
+                leftFeatureInteractions.put(currEntry.getKey(), currEntry.getValue());
+            }
+        }
+    }
+
+    public int maxDepth() {
+        return Collections.max(this.entrySet(), Comparator.comparingInt(entry -> entry.getValue().depth)).getValue().depth;
+    }
+    
+    public TwoDimTable[] getFeatureInteractions() {
```
honzasterba, line 48: `FeatureInteractions.getFeatureInteractions` is not very descriptive, what about `getAsTable()`?

h2o-core/src/main/java/hex/SplitValueHistogram.java
```diff
@@ -0,0 +1,36 @@
+package hex;
+
+import org.apache.commons.lang.mutable.MutableInt;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+
+public class SplitValueHistogram {
+    
+    private TreeMap<Double, MutableInt> map;
```
honzasterba, line 10: can be final

h2o-r/h2o-package/R/xgboost.R
```diff
@@ -599,3 +599,46 @@ h2o.xgboost.available <- function() {
...
+#' h2o.init()
+#' boston <- h2o.importFile(
+#'        "https://s3.amazonaws.com/h2o-public-test-data/smalldata/gbm_test/BostonHousing.csv",
+#'         destination_frame="boston"
+#'         )
+#' boston_xgb <- h2o.xgboost(training_frame = boston, y = "medv", seed = 1234)
+#' feature_interactions <- h2o.xgboost.feature_interaction(boston_xgb)
+#' }
+#' @export
+h2o.xgboost.feature_interaction <- function(model, max_interaction_depth = 100, max_tree_depth = 100, max_deepening = -1) {
```
honzasterba, line 632: are other models going to be supported in the future? if yes, than we should create a more generic api

koniecsveta: xgboost and gbm

honzasterba: Since the API and back-end will be the same for both xgboost and gbm I think we should have just one function for both. But I am not sure. Maybe @ledell can provide some feedback?
Also I think the norm now is to use camelCase for naming R APIs.

honzasterba (commented):
overall lgtm, only docs are missing, are you panning another PR?

h2o-r/h2o-package/R/xgboost.R
koniecsveta: the requirement was to do this for XGB as it is done in XGBFI package (this part is clear), but XGBFI isn't intended for GBM and I am not sure whether it can be adjusted to use it. For gbm feature importance there is a paper on how to measure it by calculating H-statistic what is a different approach, so if used, might be that the gbm function will have different signature. So I am not sure whether to prepare more generic API in case it will be possible to adjust XGBFI approach to gbm - when it may happen that I will not use it in the end. In case I will, I can always add more generic and deprecate and after some time remove this specific xgb one. I am not sure what approach will be used for gbm right now.

koniecsveta:
> overall lgtm, only docs are missing, are you panning another PR?

docs will be done in https://h2oai.atlassian.net/browse/PUBDEV-7877

h2o-core/src/main/java/hex/FeatureInteractions.java
```diff
@@ -0,0 +1,307 @@
+package hex;
+
+import hex.genmodel.algos.tree.SharedTreeNode;
+import org.apache.commons.lang.mutable.MutableInt;
+import water.util.TwoDimTable;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class FeatureInteractions {
```
honzasterba, line 10: would love to see some unit tests for this class

koniecsveta: 1e91e7a

honzasterba (approved):

michalkurka:
Thank you for adding this feature - great work!

PR #5055, opened 2020-10-16
PUBDEV-7842: Fix SE's handling of training frame

tomasfryda:
https://0xdata.atlassian.net/browse/PUBDEV-7842

### Problem Description
SE removes columns that might be used by base models. This happens in `ModelBuilder::ignoreBadColumns`. The edge case that @mn-mikke and @michalkurka reported is caused by using default implementation of `ModelBuilder::canLearnFromNAs` (=> `false`), which is overridden for tree-based models.

Other methods that might cause similar problems in future:
* `ModelBuilder::ignoreStringColumns`
* `ModelBuilder::ignoreConstColumns`
* `ModelBuilder::ignoreUuidColumns`

Note that all mentioned methods are `protected`.

###  Current Solution

Since I need to be able to infer the values of these methods from already built base models, I propose to move the implementation of these methods to `Model.Parameters` so that the value can depend on user specified parameters and is available for both `ModelBuilder` and `Model` (which would be useful in context of Stacked Ensembles).

 Note: I still haven't spend much time in `ModelBuilder` so I might be missing something important here.


h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
```diff
@@ -1406,4 +1408,31 @@ private void familyAndLinkInferenceTestHelper(
...
+            output = basicEnsemble("./smalldata/junit/cars.csv",
+                    null,
+                    new StackedEnsembleTest.PrepData() {
+                        int prep(Frame fr) {
+                            fr.remove("name").remove();
+                            Vec acVec = fr.anyVec().makeCon(Math.PI); // this is what use to break SE (SE ignored this column, DRF did not)
+                            Scope.track(acVec);
+                            acVec.setNA(0);
+                            fr.insertVec(0, "almost_constant", acVec);
+                            return ~fr.find("economy (mpg)");
```
sebhrusen, line 1427: can we get rid of those confusing-hair-pulling-wtf-is-it-useful-here bitwise operator?
I tried to run `StackedEnsembleTest` after removing those as well as the make-things-even-more-confusing-and-useless statement:
```java
            if( idx < 0 ) idx = ~idx;
```
and all tests are passing...

sebhrusen (commented):
something I'm not sure to understand: in SE, all base models are already trained, so we shouldn't not really care about about what they supported during training or not as we will only use the base models to ma predictions (especially on the training frame itself).
And every model should be able to make a prediction for a row if one predictor is NA.
So basically, couldn't we just say that SE can always learn from NAs?
In this case, we would just need to have `StackedEnsembleDriver` to override
```
  protected boolean canLearnFromNAs() {
    return true;
  }
```

michalkurka (changes requested):
Agree with Seb: SE shouldn't try to replicate all the logic that is used to drop columns on the base learners. IMHO it should inspect the base learners and collect all the used columns, and exclude these from the "ignore" logic.

michalkurka:
> I'm even wondering if SE should ignore any column at all.

@sebhrusen that would be a bad thing for productionizing of models, you definitely do not want to force users to provide data that are not actually needed by the model - obtaining this data might be expensive - eg. large text columns

sebhrusen:
> you definitely do not want to force users to provide data that are not actually needed by the model 

@michalkurka, do you mean that any column that is not skipped is then required by the MOJO?
This is an interesting aspect that I didn't consider in my recent changes in TE. Indeed, when using `EasyPredictionModelWrapper`, the `RowToRawDataConverter` will just assign NAs to features not provided by the user, so I don't see exactly where is the problem, outside having a few additional NAs values in the `rawData` array.

michalkurka:
@sebhrusen 

There are 2 aspects - what is actually required and what will users do. After the model is developed it is put in production (usually by a production team), model documentation (including used features) will be used to develop a feature pipeline that will be fed into the model. If users don't know what features are actually used, they will end-up feeding it everything - including stuff that is not actually needed. That can increase the scoring time. Because getting features might also be expensive (see the example I noted above with a large text feature).

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
```diff
@@ -1048,56 +1049,6 @@ public void testMissingFoldColumn_blendingFrame() {
     }
   }
 
-  @Test
-  public void testInvalidFoldColumn_trainingFrame() {
```
tomasfryda, line 1052: Since SE now overrides the ignore logic so it keeps columns used by base models/metalearner, this test has no function since it assumes that SE will ignore invalid (String) column which is used by a base model.

Similar test that tests situation when we are missing fold column is precedes this one.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -57,6 +59,42 @@ public boolean isSupervised() {
...
+      usedColumns.addAll(Arrays.asList(model._parms.getNonPredictors()));
+      if (model._output._origNames != null)
+        usedColumns.addAll(Arrays.asList(model._output._origNames));
+      else
+        usedColumns.addAll(Arrays.asList(model._output._names));
+    }
+
+    usedColumns.addAll(Arrays.asList(_parms.getNonPredictors()));
+
+    if (usedColumns.size() < _train._names.length) {
```
sebhrusen, line 80: looks like this condition makes too strong assumptions: we don't require all base models to have been trained with the same training frame (I mean using the same predictors). The only requirement somehow, is that the training frame used to train the SE should have enough columns to allow each model to give predictions, knowing that models don't usually need all their training predictors to produce predictions.

This being said, we can imagine cases were `usedColumns > train.length` but you could still be able to remove columns from train, so I would just remove the condition.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -57,6 +59,42 @@ public boolean isSupervised() {
...
+      int[] toDelete = IntStream
+              .range(0, _train._names.length)
+              .filter(columnIdx -> !usedColumns.contains(_train._names[columnIdx]))
+              .toArray();
+
+      _removedCols = Arrays.stream(toDelete)
+              .mapToObj(columnIdx -> _train._names[columnIdx])
+              .collect(Collectors.toCollection(HashSet<String>::new));
```
sebhrusen, line 88: `Collectors.toSet()` should work

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -99,6 +100,16 @@ public void initMetalearnerParams(Metalearner.Algorithm algo) {
...
+    @Override
+    public String[] getNonPredictors() {
+      HashSet<String> nonPredictors = new HashSet<>();
+      nonPredictors.addAll(Arrays.asList(super.getNonPredictors()));
+      if (null != _metalearner_parameters)
+        nonPredictors.addAll(Arrays.asList(_metalearner_parameters.getNonPredictors()));
+      if (null != _metalearner_fold_column)
+        nonPredictors.add(_metalearner_fold_column);
+      return nonPredictors.toArray(new String[0]);
+    }
```
sebhrusen, line 112: so you're using my hint for another PR in this one? I first got confused when I saw this change :)

I have a doubt about the necessity to include
```
      if (null != _metalearner_parameters)
        nonPredictors.addAll(Arrays.asList(_metalearner_parameters.getNonPredictors()));
```
can you please justify?

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
tomasfryda: Thanks!

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
tomasfryda: This is very good point. I completely missed that we can use different training frames. Thanks!

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
tomasfryda: Yes, sorry, I usually read through the comments and often I find something new so I want to use it.

My assumption is that there can be a metalearner that would use a different column as a non predictor which would be otherwise removed.
 
For example, I am not sure if HGLM's `random_columns`[1] are included in the predictors so I delegated the responsibility to GLM to decide what's needed. 

(This assumes that the individual algos override the `getNonPredictors` when needed.)

I know this is poor justification, I will look at it on Monday to make sure if there actually some use case for that if condition.

[1] https://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/algo-params/random_columns.html

sebhrusen: Even if your justification makes sense for edge cases, I don't think we support those anyway as the training frame given to the metalearner will contain only the predictions columns from the base models + the ones listed in `StackedEnsemble.addMiscColumnsToLevelOneFrame`.

For clarity, I'm just expecting that the `non-predictors` are exactly those in `addMiscColumnsToLevelOneFrame`.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
tomasfryda: `ModelBuilder._removedCols` is declared as `HashSet` and AFAIK `Collectors.toSet()` doesn't guarantee that I will get a hash set so for now I decided to stick with this.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
tomasfryda: OK, I removed the condition and renamed `addMiscColumnsToLevelOneFrame` to `addNonPredictorsToLevelOneFrame` make its purpose more clear.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
sebhrusen: I see. The declaration of `_removedCols` in `ModelBuilder` as `HashSet` instead of `Set` is most probably unjustified (as well as the `public` visibility) but well…

Now that you mention `_removedCols`, wouldn't it make sense to change lines 81 onwards to use `FilterCols` instead? This would avoid to logic copied from there.
```java
      new FilterCols(npredictors) {
        @Override protected boolean filter(Vec v, String name) {
             return !usedColumns.contains(name);
        }
      }.doIt(_train,"Dropping bad and constant columns: ",expensive);
```
This just requires to modify `FilterCols` to change the signature of `filter` method: given the implementation, it should be very simple to pass the `name` as well.

tomasfryda: Thanks for the suggestion!

sebhrusen (approved):
@tomasfryda Sorry, forgot about this PR. Looks good, thanks!

sebhrusen:
@tomasfryda please update the target branch.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -99,6 +100,14 @@ public void initMetalearnerParams(Metalearner.Algorithm algo) {
     
     public final Frame blending() { return _blending == null ? null : _blending.get(); }
 
+    @Override
+    public String[] getNonPredictors() {
```
michalkurka, line 104: I know this method is not new but this is the first time I noticed it:

Model builder defines methods numSpecialCols and specialColNames, I think the naming needs to be unified.

sebhrusen: @michalkurka I think I created it for TE. I probably didn't notice the `specialCol__` properties because those are available only at ModelBuilder level, whereas I needed those also for scoring. Another small difference is that it includes the response column, but I still agree that it should be unified.
Creating a ticket for this.

sebhrusen: https://h2oai.atlassian.net/browse/PUBDEV-8078

michalkurka (approved):
LGTM, thank you @tomasfryda

PR #5056, opened 2020-10-19
[PUBDEV-7846] Finalization API with Leaving Frames Locked

mn-mikke:
The newly introduced method is required for proper locking of H2O Frames during conversion from Spark Data Frames to H2O Frames in Sparkling Water. After the frame is finalized, SW still does some [post processing](https://github.com/h2oai/sparkling-water/blob/958c9a4210995ab7f3ec0139119ebc51b154fac5/extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala#L37).

### DIFF
--- a/h2o-core/src/main/java/water/fvec/Frame.java
+++ b/h2o-core/src/main/java/water/fvec/Frame.java
@@ -1078,8 +1078,15 @@ public class Frame extends Lockable<Frame> {
   }
 
   // Build real Vecs from loose Chunks, and finalize this Frame.  Called once
+  // after any number of [create,close]NewChunks. This method also unlocks the frame.
+  void finalizePartialFrame(long[] espc, String[][] domains, byte[] types) {
+    finalizePartialFrameAndLeaveLocked(espc, domains, types);
+    unlock();
+  }
+  
+  // Build real Vecs from loose Chunks, and finalize this Frame.  Called once
   // after any number of [create,close]NewChunks.
-  void finalizePartialFrame( long[] espc, String[][] domains, byte[] types ) {
+  void finalizePartialFrameAndLeaveLocked(long[] espc, String[][] domains, byte[] types) {
     // Compute elems-per-chunk.
     // Roll-up elem counts, so espc[i] is the starting element# of chunk i.
     int nchunk = espc.length;
@@ -1111,7 +1118,6 @@ public class Frame extends Lockable<Frame> {
       DKV.put(_keys[i],vec,fs);
     }
     fs.blockForPending();
-    unlock();
   }
 
   // --------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1077,9 +1077,16 @@ static void closeNewChunks(NewChunk[] nchks) {
...
+  // after any number of [create,close]NewChunks. This method also unlocks the frame.
+  void finalizePartialFrame(long[] espc, String[][] domains, byte[] types) {
+    finalizePartialFrameAndLeaveLocked(espc, domains, types);
+    unlock();
+  }
+  
   // Build real Vecs from loose Chunks, and finalize this Frame.  Called once
   // after any number of [create,close]NewChunks.
-  void finalizePartialFrame( long[] espc, String[][] domains, byte[] types ) {
+  void finalizePartialFrameAndLeaveLocked(long[] espc, String[][] domains, byte[] types) {
```
Pscheidl, line 1089: Would be a lot nicer to rename this method (if this can be done) to `finalizePartialFrame` and introduce a new argument `boolean lock` - this way, the method would be generic enough. 

I mean, if we can't change the signature of this method, we might always leave it as-is, introduce a new one with the parameter and let the old one call the new one with `true` value which means "lock it, please".

mn-mikke: ok np, will do. I don't have any preferences.

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1077,9 +1077,15 @@ static void closeNewChunks(NewChunk[] nchks) {
...
+  // Build real Vecs from loose Chunks, and finalize this Frame.  Called once
+  // after any number of [create,close]NewChunks. This method also unlocks the frame.
+  void finalizePartialFrame(long[] espc, String[][] domains, byte[] types) {
+    finalizePartialFrame(espc, domains, types, false);
+  }
+  
   // Build real Vecs from loose Chunks, and finalize this Frame.  Called once
   // after any number of [create,close]NewChunks.
-  void finalizePartialFrame( long[] espc, String[][] domains, byte[] types ) {
+  void finalizePartialFrame(long[] espc, String[][] domains, byte[] types, boolean leaveLocked) {
```
michalkurka, line 1088: The default for unlocking is to unlock, it is IMHO better to have a true (=posive case) as a default. I would thus name it `unlock` with default value `true`.

michalkurka (approved):

h2o-core/src/main/java/water/fvec/Frame.java
mn-mikke: ok, np. will change

Pscheidl (approved):

PR #5057, opened 2020-10-19
PUBDEV-7851: Add user friendly warning when missing ggplot2 in explainability module

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7851

Stops the execution with message such as:
```Function "h2o.model_correlation_heatmap" requires ggplot2 version 3.0.0 or higher.```
and
```Function "h2o.pd_multi_plot" requires ggplot2 version 3.3.0 or higher.```
since it uses the `ggplot2::expansion`.


ledell (approved):
Awesome, thank you!

PR #5058, opened 2020-10-20
PUBDEV-7832: Add information about how early stopping is used to tune models in the User Guide

hannah-tillman:
https://h2oai.atlassian.net/browse/PUBDEV-7832

So, instead of adding this information to the training models page, I thought it would be more cohesive to keep all the early stopping info in one place in the docs. This can easily be changed if you would rather it go beneath the regression section of the training models page, though. Please, let me know if anything else needs updated.


ledell:
Oh, interesting... I had not seen the "early stopping" page before.  I think it makes sense to add this information there.  

I also think we should probably clean up the `data-science.rst` page a little bit (maybe get rid of "Common" section), because early stopping is specifically related to supervised algorithms.  It's not a "common" algorithm itself, it's just an aspect of how the supervised algorithms are trained.  So the link to it should probably be moved to the `training-models.rst` page.  I think we can also get rid of the link to the `quantile()` function because it's not really an "algorithm" either... it's just a computation that can be done on a dataset.  Let me take a look again tomorrow and I can suggest some edits.  Thanks!

ledell:
@hannah-tillman I made some changes to `early-stopping.rst`.

ledell (approved):
This looks good now!  We can follow-up with some changes to `data-science.rst` either on this PR tomorrow, or do that on a separate PR.

hannah-tillman:
> This looks good now! We can follow-up with some changes to `data-science.rst` either on this PR tomorrow, or do that on a separate PR.

Cool! So now we just need a new home for the ``quantile()`` page. 

For early stopping, I shifted the file out of the data-science folder (just into the normal product folder) and placed the link to the file in the training-models page under the supervised regression section. I had to reformat some of the headers to fit it--let me know if you want anything updated.

PR #5059, opened 2020-10-20
PUBDEV-7606: add _varimp to GLM model output.

wendycwong:
This PR completes the work for JIRA: https://h2oai.atlassian.net/jira/software/c/projects/PUBDEV/issues/PUBDEV-7644?filter=myopenissues

GLM provides standardized coefficients which can be massaged into varimp by Python and R clients.  Following the implementation of other H2O algos like deeplearning, GBM and others, I have added the field _varimp to GLMOutput.  In addition, I enabled GLM mojo to be able to use the generic framework of mojos and therefore be able to access variable importances to generic mojo as well.

I added Python/R client tests to make sure the varimp implementation is correct.

I added to Pavel's generic mojo model test to make sure variable importances can be accessed through mojo.

### DIFF
--- a/h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
+++ b/h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
@@ -6,13 +6,13 @@ import water.MemoryManager;
 import water.api.API;
 import water.api.schemas3.ModelOutputSchemaV3;
 import water.api.schemas3.ModelSchemaV3;
-
 import water.api.schemas3.TwoDimTableV3;
 import water.util.ArrayUtils;
 import water.util.TwoDimTable;
 
 import java.util.Arrays;
-import java.util.Comparator;
+
+import static water.util.ArrayUtils.sort;
 //import water.util.DocGen.HTML;
 
 public class GLMModelV3 extends ModelSchemaV3<GLMModel, GLMModelV3, GLMModel.GLMParameters, GLMV3.GLMParametersV3, GLMOutput, GLMModelV3.GLMModelOutputV3> {
@@ -30,6 +30,9 @@ public class GLMModelV3 extends ModelSchemaV3<GLMModel, GLMModelV3, GLMModel.GLM
 
     @API(help="Standardized Coefficient Magnitudes")
     TwoDimTableV3 standardized_coefficient_magnitudes;
+
+    @API(help = "Variable Importances", direction = API.Direction.OUTPUT, level = API.Level.secondary)
+    TwoDimTableV3 variable_importances;
 
     @API(help="Lambda minimizing the objective value, only applicable with lambda search or when arrays of alpha and " +
             "lambdas are provided")
@@ -58,6 +61,11 @@ public class GLMModelV3 extends ModelSchemaV3<GLMModel, GLMModelV3, GLMModel.GLM
       if(impl.get_global_beta_multinomial() == null)
         return this; // no coefificients yet
       String [] names = impl.coefficientNames().clone();
+      int len = names.length-1;
+      String [] names2 = new String[len]; // this one decides the length of standardized table length
+      int[] indices = new int[len];
+      for (int i = 0; i < indices.length; ++i)
+        indices[i] = i;
       // put intercept as the first
       String [] ns = ArrayUtils.append(new String[]{"Intercept"},Arrays.copyOf(names,names.length-1));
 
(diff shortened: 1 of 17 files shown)


h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -254,6 +253,22 @@ public int compare(Integer o1, Integer o2) {
     }
   } // GLMModelOutputV2
 
+  public static void calculateVarimpBase(double[] magnitudes, Integer[] indices, double[] betaNorm) {
+    for (int i = 0; i < magnitudes.length; ++i) {
+      magnitudes[i] = (float) betaNorm[i];
+      if (magnitudes[i] < 0) magnitudes[i] *= -1;
+    }
+    
+    Arrays.sort(indices, new Comparator<Integer>() {
```
michalkurka, line 262: You might want to take a look at ArrayUtils#sort - it does the same thing: Sort an integer array of indices based on values

Most of this method won't be needed then.

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -254,6 +253,22 @@ public int compare(Integer o1, Integer o2) {
     }
   } // GLMModelOutputV2
 
+  public static void calculateVarimpBase(double[] magnitudes, Integer[] indices, double[] betaNorm) {
+    for (int i = 0; i < magnitudes.length; ++i) {
+      magnitudes[i] = (float) betaNorm[i];
```
michalkurka, line 258: is there a reason to convert to float?

wendycwong: For some reason VarImp class only takes float[] for values and String[] for names.

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
wendycwong: Good idea.  Will do!

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -138,6 +128,23 @@ public int compare(Integer o1, Integer o2) {
       return this;
     }
 
+    public static void calculateVarimpMultinomial(double[] magnitudes, Integer[] indices, double[][] betaNorm) {
+      for (int i = 0; i < betaNorm.length; ++i) {
+        for (int j = 0; j < betaNorm[i].length; ++j) {
+          double d = betaNorm[i][j];
+          magnitudes[j] += d < 0 ? -d : d;
```
michalkurka, line 135: Is this correct? Should we instead consider the aggregated coefficient (= move this to the outer loop)?

I don't know - asking because it is not clear to me.

wendycwong: Here the first for loop is over all classes.

The second for loop  is over all coefficients for a fixed class.  

Line 135 is basically adding the magnitudes of say coefficient j of all classes together.  It is aggregating coefficients over all classes and should be correct.  

Say, we have 3 classes and 5 coefficients.  Then, for class 0, it will go over all 5 coefficients and add the magnitude of the 5 coefficients for class 0.  Next, it will do the same for class 1, add the 5 coefficients into magnitude... and so on.  Does this make sense?

michalkurka: I see, so I misunderstood the code - it is actually doing what I thought it should be doing based on your description. Thank you @wendycwong 

Could you add a (java) unit test for this method - it should be relatively easy to do (since this is a static method).

wendycwong: Sure.

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -31,6 +30,9 @@
     @API(help="Standardized Coefficient Magnitudes")
     TwoDimTableV3 standardized_coefficient_magnitudes;
 
+    @API(help="Variable Importances", direction=API.Direction.OUTPUT, level = API.Level.secondary)
+	    TwoDimTableV3 variable_importances;
```
honzasterba, line 34: indnet

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
wendycwong: @michalk:  I check

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
wendycwong: Done.  Good catch!

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
wendycwong: Done

h2o-algos/src/test/java/hex/glm/GLMBasicTestMultinomial.java
```diff
@@ -50,6 +52,36 @@ public static void cleanUp() {
...
+    }
+    calculateVarimpMultinomial(magnitudes, indices, betaNorm);  // calculate magnitude and generate indices of descending sort
+    
+    for (int classInd = 0; classInd < numClass; classInd++) { // manually generate coefficient magnitudes
+      for (int coeffInd = 0; coeffInd < coeffLen; coeffInd++) {
+        magnitudes_check[coeffInd] = magnitudes_check[coeffInd] + Math.abs(betaNorm[classInd][coeffInd]);
+      }
+    }
+    // check to make sure magnitude calculation is correct
+    assert Arrays.equals(magnitudes, magnitudes_check) : "Magnitude calculation for multinomial is incorrect.";
```
michalkurka, line 79: Please junit assert instead of Java assert (they can be disabled!).

michalkurka (approved):
LGTM 

(one comment about test use of assert)

h2o-algos/src/test/java/hex/glm/GLMBasicTestMultinomial.java
wendycwong: Sure

wendycwong: Good point.  Done!

PR #5060, opened 2020-10-21
Don't fail if javadoc is not present for non-release builds

satai:


satai:
The reason javadoc can be absent: https://github.com/h2oai/h2o-3/commit/087e771366a766390527ff9c62404a4b0cfc5c23#diff-f84b31486d438ea87b55ce5d9a4e3b36fc72c39183bdcecd15a144b9e00a4d45R26

PR #5061, opened 2020-10-21
fix failing build - release at CI needs to make javadoc

satai:
caused by 087e771366a766390527ff9c62404a4b0cfc5c23

Javadoc was not created when nexus upload was not enabled by jenkins parameters. Then `cp: cannot stat 'h2o-core/build/docs/javadoc': No such file or directory` appeared.

### DIFF
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -23,7 +23,7 @@ artifacts {
     toPublish sourcesJar
 }
 
-if (project.ext.isRelease) {
+if (project.ext.isMakeJavadoc) {
     java {
         withJavadocJar()
     }
--- a/build.gradle
+++ b/build.gradle
@@ -196,6 +196,7 @@ allprojects {
 
     ext {
         isRelease = rootProject.hasProperty("doRelease")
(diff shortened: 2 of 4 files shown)


gradle/publish.gradle
```diff
@@ -23,7 +23,7 @@ artifacts {
     toPublish sourcesJar
 }
 
-if (project.ext.isRelease) {
+if (project.ext.isMakeJavadoc) {
```
Pscheidl, line 26: The `isMakeJavadoc` is only true when the `makeJavadoc` parametr is put inside. This is wrong, as we need to make JavaDoc during release - with the doRelease parameter passed. Let's see the old code - there was also conditional JavaDoc - let's examine the conditions used.

Pscheidl (changes requested):

gradle/publish.gradle
satai: if you want to make `doRelease` to imply `makeJavadoc` you can easily add 

```groovy
         isMakeJavadoc = rootProject.hasProperty("makeJavadoc") || rootProject.hasProperty("doRelease")
``` 

The problem was, that `doRelease` depends on nexus upload parameter.

PR #5064, opened 2020-10-22
PUBDEV-7841 don't overwrite existing frame when data loaded again fro…

satai:
https://h2oai.atlassian.net/browse/PUBDEV-7841

### DIFF
--- a/h2o-core/src/main/java/water/jdbc/SQLManager.java
+++ b/h2o-core/src/main/java/water/jdbc/SQLManager.java
@@ -2,13 +2,13 @@ package water.jdbc;
 
 import water.*;
 import water.fvec.*;
-import water.parser.BufferedString;
 import water.parser.ParseDataset;
 import water.util.Log;
 
-import java.math.BigDecimal;
 import java.sql.*;
+import java.util.Objects;
 import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.atomic.AtomicLong;
 
 public class SQLManager {
 
@@ -30,6 +30,19 @@ public class SQLManager {
 
   private static final String TMP_TABLE_ENABLED = H2O.OptArgs.SYSTEM_PROP_PREFIX + "sql.tmp_table.enabled";
 
+  private static AtomicLong NEXT_TABLE_NUM = new AtomicLong(0);
+  
+  static Key<Frame> nextTableKey(String prefix, String postfix) {
+    Objects.requireNonNull(prefix);
+    Objects.requireNonNull(postfix);
+
+    final long num = NEXT_TABLE_NUM.incrementAndGet();
+    final String s = prefix + "_" + num +  "_" + postfix;
+    final String withoutWhiteChars = s.replaceAll("\\W", "_");
+
+    return Key.make(withoutWhiteChars);
+  }
+  
   /**
    * @param connection_url (Input)
    * @param table (Input)
@@ -46,7 +59,7 @@ public class SQLManager {
       final Boolean useTempTable, final String tempTableName,
       final SqlFetchMode fetchMode, final Integer numChunksHint) {
 
-    final Key<Frame> destination_key = Key.make((table + "_sql_to_hex").replaceAll("\\W", "_"));
+    final Key<Frame> destination_key = nextTableKey(table, "sql_to_hex");
     final Job<Frame> j = new Job<>(destination_key, Frame.class.getName(), "Import SQL Table");
 
     final String databaseType = getDatabaseType(connection_url);
--- a/h2o-core/src/test/java/water/jdbc/SqlManagerTest.java
+++ b/h2o-core/src/test/java/water/jdbc/SqlManagerTest.java
@@ -1,0 +1,49 @@
+package water.jdbc;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import water.Key;
+import water.Keyed;
+import water.fvec.Frame;
+import water.runner.CloudSize;
+import water.runner.H2ORunner;
+
+import java.util.Set;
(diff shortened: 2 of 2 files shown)


h2o-core/src/main/java/water/Key.java
```diff
@@ -299,7 +299,37 @@ public static String rand() {
...
+   * @param postfix - postfix of the new key. Should not be null. If it contains non-word characters,
+   *                they are converted to underscores.
+   *
+   *  @return Desired Key that looks like prefix_RANDOM_postfix 
+   */
+  public static <P extends Keyed> Key<P> makeRandom(String prefix, String postfix) {
+    assert prefix != null;
+    assert postfix != null;
+    
+    final String s = prefix + "_" + rand()  + (postfix.isEmpty() ? "" : ("_" + postfix));
```
Pscheidl, line 316: The internet says modern Java transforms this to `StringBuilder` anyway, so we're good and it's readable.

h2o-core/src/main/java/water/Key.java
```diff
@@ -299,7 +299,37 @@ public static String rand() {
...
+    final String withoutWhitechars = s.replaceAll("\\W", "_");
```
Pscheidl, line 317: Great catch !

Pscheidl (approved):

h2o-core/src/main/java/water/Key.java
```diff
@@ -299,7 +299,37 @@ public static String rand() {
...
+   * 
+   * @param prefix - prefix of the new key. Should not be null. If it contains non-word characters, 
+   *               they are converted to underscores.
+   * @param postfix - postfix of the new key. Should not be null. If it contains non-word characters,
+   *                they are converted to underscores.
+   *
+   *  @return Desired Key that looks like prefix_RANDOM_postfix 
+   */
+  public static <P extends Keyed> Key<P> makeRandom(String prefix, String postfix) {
+    assert prefix != null;
```
Pscheidl, line 313: I'd even do Objects.requireNotNull here !

h2o-core/src/main/java/water/Key.java
```diff
@@ -299,7 +299,37 @@ public static String rand() {
...
+  /** Factory making an random Key from two Strings
+   * 
+   * @param prefix - prefix of the new key. Should not be null. If it contains non-word characters, 
+   *               they are converted to underscores.
+   * @param postfix - postfix of the new key. Should not be null. If it contains non-word characters,
+   *                they are converted to underscores.
+   *
+   *  @return Desired Key that looks like prefix_RANDOM_postfix 
+   */
+  public static <P extends Keyed> Key<P> makeRandom(String prefix, String postfix) {
```
Pscheidl, line 312: (Optional) make args final for absolute perfection.

h2o-core/src/main/java/water/Key.java
satai: Yes. It was terrible, then it was StringBuffer and now it's StringBuilder.

h2o-core/src/main/java/water/Key.java
satai: Honestly, it was not mine, I have seen this in other places in our codebase.

h2o-core/src/main/java/water/Key.java
satai: Good idea, updated.

h2o-core/src/main/java/water/Key.java
satai: Let's make it even better :-D

h2o-core/src/main/java/water/Key.java
```diff
@@ -299,7 +300,37 @@ public static String rand() {
...
+   *                they are converted to underscores.
+   *
+   *  @return Desired Key that looks like prefix_RANDOM_postfix 
+   */
+  public static <P extends Keyed> Key<P> makeRandom(final String prefix, final String postfix) {
+    Objects.requireNonNull(prefix);
+    Objects.requireNonNull(postfix);
+    
+    final String s = prefix + "_" + rand()  + (postfix.isEmpty() ? "" : ("_" + postfix));
+    final String withoutWhitechars = s.replaceAll("\\W", "_");
```
michalkurka, line 318: This behavior doesn't belong in this class (changes requested).

h2o-core/src/main/java/water/jdbc/SQLManager.java
```diff
@@ -46,7 +46,7 @@
       final Boolean useTempTable, final String tempTableName,
       final SqlFetchMode fetchMode, final Integer numChunksHint) {
 
-    final Key<Frame> destination_key = Key.make((table + "_sql_to_hex").replaceAll("\\W", "_"));
+    final Key<Frame> destination_key = Key.makeRandom(table, "sql_to_hex");
```
michalkurka, line 49: Instead of using a partially randomized key, I think it is cleaner to follow a similar pattern as we use for naming models - just use an increasing sequence of numbers for a suffix, see H2O#calcNextUniqueModelId for inspiration. 

it could be even very simple, something like

    private static AtomicLong nextTableNum = new AtomicLong(0);
    ...
    final Key<Frame> destination_key = Key.make((table.replaceAll("\\W", "_") + "_sql_to_hex_" + nextTableNum.addAndGet(1)));

The advantage definitely is that it is clear in what order the keys were created - which could be useful in debugging.

Pscheidl (approved):

h2o-core/src/main/java/water/jdbc/SQLManager.java
satai: OK, you can have a look.

michalkurka (approved):

PR #5065, opened 2020-10-22
[PUBDEV-7852] Add Constructor to PreviewParseWriter with Domains as Parameter

mn-mikke:
This ticket help to get rid of [IcedHashMapWrapper](https://github.com/h2oai/sparkling-water/blob/b2f2d55009845a4be9e5d9f0fd3d9b63140addca/extensions/src/main/scala/water/parser/IcedHashMapWrapper.java) in Sparkling Water. 

The plan is to pass `domains` converted to lists [here](https://github.com/h2oai/sparkling-water/blob/958c9a4210995ab7f3ec0139119ebc51b154fac5/extensions/src/main/scala/water/parser/CategoricalPreviewParseWriter.java#L26). The conversion will be done via `java.util.Arrays.asList` which creates just an wrapper around the input array.

### DIFF
--- a/h2o-core/src/main/java/water/parser/PreviewParseWriter.java
+++ b/h2o-core/src/main/java/water/parser/PreviewParseWriter.java
@@ -6,6 +6,8 @@ import water.Iced;
 import water.fvec.Vec;
 import water.util.ArrayUtils;
 import water.util.IcedHashMap;
+import java.util.Collection;
+import java.util.Collections;
 
 /** Class implementing ParseWriter, on behalf ParseSetup
  * to examine the contents of a file for guess the column types.
@@ -18,7 +20,7 @@ public class PreviewParseWriter extends Iced implements StreamParseWriter {
   protected int _invalidLines;
   private   String []   _colNames;
   protected String [][] _data = new String[MAX_PREVIEW_LINES][];
-  private IcedHashMap<String,String>[] _domains;  //used in leiu of a HashSet
+  private Collection<String>[] _domains;
   int [] _nnums;
   int [] _nstrings;
   int [] _ndates;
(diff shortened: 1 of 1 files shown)


h2o-core/src/main/java/water/parser/PreviewParseWriter.java
```diff
@@ -18,7 +20,7 @@
   protected int _invalidLines;
   private   String []   _colNames;
   protected String [][] _data = new String[MAX_PREVIEW_LINES][];
-  private IcedHashMap<String,String>[] _domains;  //used in leiu of a HashSet
+  private Collection<String>[] _domains;
```
honzasterba, line 23: why not just String[][]?

h2o-core/src/main/java/water/parser/PreviewParseWriter.java
```diff
@@ -49,9 +62,9 @@ private void setColumnCount(int n) {
       _ndates = new int[n];
       _nnums = new int[n];
       _nempty = new int[n];
-      _domains = new IcedHashMap[n];
+      _domains = new Collection[n];
       for(int i = 0; i < n; ++i)
-        _domains[i] = new IcedHashMap<>();
+        _domains[i] = Collections.newSetFromMap(new IcedHashMap<>());
```
honzasterba, line 67: you are saying that the goal of this PR is to remove the use of IcedHashMap, but yet it it still being used, I dotn get the benefit of moving just here

h2o-core/src/main/java/water/parser/PreviewParseWriter.java
mn-mikke: `PreviewParseWriter` needs to use internally a hash map or hash set for effective insertion in the [`addStrCol`](https://github.com/h2oai/h2o-3/pull/5065/files#diff-1f9b998f1c157b03f6fb652a49fa1460a79ec7a30e569563579f0d7326d027b4L107) method and calculation of distinct column values (a domain of a categorical column).

In this PR, I want to extend `PreviewParseWriter` for another use case. In SW use case, [the domain is already known](https://github.com/h2oai/sparkling-water/blob/68a8a90d4e3c9242fb5109d58be3585da2da89ea/extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala#L99), and doesn't have to be calculated via the `addStrCol` method. In SW, I just want to pass the domain to `PreviewParseWriter`  as a string array and avoid allocation of a big hash map when domain cardinality is high.

I've chosen `Collection` as a type for domains since it's a common ancestor for a hash set and a list which will be backed by the original string array representing the domain.

h2o-core/src/main/java/water/parser/PreviewParseWriter.java
mn-mikke: See the comment above. Here I just need to have a HashSet. I couldn't have used `IcedHashSet` since it doesn't allow string to be a generic parameter.

mn-mikke:
Replaced by https://github.com/h2oai/h2o-3/pull/5086

PR #5066, opened 2020-10-22
WIP add h2 support to tests

satai:
WIP


michalkurka:
Why is derby not enough? Derby is already used in the tests

PR #5068, opened 2020-10-22
Allow experimentation with multi-gpu deployments

michalkurka:


Pscheidl (approved):
Reviewed code. I don't have much context for this change.

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -169,7 +169,7 @@
       if (! hasGPU(_parms._gpu_id))
         error("_backend", "GPU backend (gpu_id: " + _parms._gpu_id + ") is not functional. Check CUDA_PATH and/or GPU installation.");
 
-      if (H2O.getCloudSize() > 1 && !_parms._build_tree_one_node)
+      if (H2O.getCloudSize() > 1 && !_parms._build_tree_one_node && !allowMultiGPU())
```
mn-mikke, line 172: Shouldn't the below error message updated as well?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
```diff
@@ -252,6 +252,14 @@
     PlattScalingHelper.initCalibration(this, _parms, expensive);
   }
 
+  static boolean allowMultiGPU() {
+    return H2O.getSysBoolProperty("xgboost.multinode.gpu.enabled", false);
+  }
+
+  private static boolean gpuCheckEnabled() {
```
mn-mikke, line 259: nit: Is the GPU check related to multinode or is it generic. The name of the method sounds generic, but the property name is related to multinode,

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -268,7 +268,7 @@ public void initActualParamValuesAfterOutputSetup(boolean isClassifier, int ncla
   public static XGBoostParameters.Backend getActualBackend(XGBoostParameters p, boolean verbose) {
     Consumer<String> log = verbose ? LOG::info : LOG::debug;
     if ( p._backend == XGBoostParameters.Backend.auto || p._backend == XGBoostParameters.Backend.gpu ) {
-      if (H2O.getCloudSize() > 1 && !p._build_tree_one_node) {
+      if (H2O.getCloudSize() > 1 && !p._build_tree_one_node && !XGBoost.allowMultiGPU()) {
```
mn-mikke, line 271: ditto

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
michalkurka: @mn-mikke I wouldn't think so - if multigpu is enabled then no one will see this error. Or?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
michalkurka: Yes, good idea - I need to fix the property name

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
michalkurka: imho the same

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
mn-mikke: This just really minor comment I won't be against if you leave it as it is.

Just If I saw this message, i would understand that there is no way how to enable it for distributed mode, but it will be possible with `allowMultiGPU` option, won't it?

michalkurka: I think I see where the misunderstanding is -> multi gpu currently requires distributed (multiple nodes even on a single machine). So the error will never be shown until it is really distributed.

mn-mikke: Do you mean by 'distributed' the option `_parms._build_tree_one_node`?

michalkurka: No, for multi-gpu to work (even on a single node) you will need 2+ instances of H2O.

So the error will only be shown if someone has 2+ instances and `allowMultiGPU = false` (which is the default), all assuming the is regular model build not single-node tree building.

I am starting to think that you want to modify the error message to include `_parms._build_tree_one_node`? Is that so? Because just for the change I made - I don't see how I would change the message :)

Maybe if you could propose a new error message - it would be clearer to me.

honzasterba (approved):
lgtm

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoost.java
mn-mikke: The message `GPU backend is not supported in distributed mode.` sounds to me that GPU is not supported for distrtibuted method at all.

I have been wondering if we change it to something like: `GPU backend for the distributed mode is currently disabled. It can be enabled by setting the property 'xgboost.multinode.gpu.enabled' to true.'`

Anyway if you think that the original message is sufficient, I'm happy with that. I don't want to bother you with such a minor thing. I assume there are more important things to solve.

honzasterba: since `xgboost.multinode.gpu.enabled` is experimental I dont think we want to mention it in our error messages and encourage people using it.

michalkurka: @mn-mikke got it, this is now meant for internal experimentation and for controlled public experimentation but generally, users shouldn't be turning it on

so I think keeping the message as-is makes sense, thanks for the comment and the discussion!

mn-mikke: > this is now meant for internal experimentation and for controlled public experimentation but generally, users shouldn't be turning it on

ok, no problem. makes sense

mn-mikke (approved):

PR #5069, opened 2020-10-22
PUBDEV-7853: Fix learning curve for GLM

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7853

GLM's learning curve plot wasn't working when used with lambda search or HGLM.  This is simple fix that just adds those 2 cases. 

To make the learning curve nice (just one line in case a user provides an array of `alpha`s which happens in AutoML), I had to add `alpha` to scoring history for lambda search to be able to plot just the "winning" `alpha` value. 

I will be happy to hear any suggestions.

This is accompanied with the fix in the flow: https://github.com/h2oai/h2o-flow/pull/171

### DIFF
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -399,6 +399,8 @@ class ModelBase(h2o_meta(Keyed)):
         model = self._model_json["output"]
         if "scoring_history" in model and model["scoring_history"] is not None:
             return model["scoring_history"].as_data_frame()
+        if "glm_scoring_history" in model and model["glm_scoring_history"] is not None:
+            return model["glm_scoring_history"].as_data_frame()
         print("No score history for this model")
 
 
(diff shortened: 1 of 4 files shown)


h2o-py/h2o/model/model_base.py
```diff
@@ -990,12 +990,28 @@ def _plot(self, timestep, metric, server=False):
         scoring_history = self.scoring_history()
         # Separate functionality for GLM since its output is different from other algos
         if self._model_json["algo"] == "glm":
-            # GLM has only one timestep option, which is `iterations`
-            timestep = "iterations"
+            if self.actual_params.get("lambda_search"):
+                allowed_timesteps = ["iteration", "duration"]
+                allowed_metrics = ["deviance_train", "deviance_test"]
+                # When provided with multiple alpha values, scoring history contains history of all...
```
wendycwong, line 996: Do you want to add deviance_xval to allowed_metrics?

wendycwong (changes requested):

h2o-py/h2o/model/model_base.py
tomasfryda: Good point! Thank you @wendycwong .
I also noticed that sometimes I get these columns that are empty (in my experiment): 
`training_rmse`, `training_logloss`, `training_r2`, `training_auc`, `training_pr_auc`, `training_lift`, `training_classification_error`

Should I add them as well? Possibly with `validation_*`?

wendycwong: @tomasfryda:  Check the other algos (GBM, deeplearning) and see what they all plot.  I think being consistent with them will be great.  I was wondering about how you deal with those metrics that are empty like deviance_test or deviance_xval.  If they are empty, don't throw an error, just skip plotting them if possible.

wendycwong (approved):
LGTM!

h2o-py/h2o/model/model_base.py
tomasfryda: @wendycwong The thing is that "deeplearning", "xgboost", "drf", "gbm" use the `training_*` and `validation_*`. Just GLM differs and apparently GAM is not supported but it seems to be easy to add [1].

In this PR, I am just trying to fix the plotting function and keep it backwards compatible so I am not changing behavior (e.g., throwing an error or skipping metrics).
Also this PR  is a great way to learn more about the scoring history (which is not implemented in stacked ensemble)). I will still work on some learning curve related functionality (for the explainability) which might be implemented as a different function/method so I wanted this to be fixed.


[1] To enable plotting GAM's learning curve I can just add into `ModelBase.scoring_history`
```
        if "glm_scoring_history" in model and model["glm_scoring_history"] is not None:
            return model["glm_scoring_history"].as_data_frame()
```
and change the condition for `GLM` to allow `GAM` to be plotted. But I am not sure if it is meaningful for `GAM`s,  especially because of the "non-standard" name of scoring history. WDYT?

wendycwong: @tomasfryda:

I like your suggestion.  GAM uses GLM to train the model and it makes sense to use the glm_scoring_history.  Do as you deem fit for the other algos as well.  Thank you, Wendy

ledell (approved):
Thanks for the fix!

wendycwong:
@wendycwong I clicked through flow to find out if I can easily add the learning curve there as well but it seems that GAM is not available, i.e., can't be selected using the menu or buildModel. Is this intentional or is it a bug?

@tomf:
Regarding the learning curve for GAM, it is hiding in the GLM model.  I did not bother to copy it over to GAM.  However, if you feel it is necessary for GAM to copy it over to itself, please feel free to do so.  Does this answer your question regarding the learning curve for GAM?  Please let me know.  Wendy

PR #5075, opened 2020-10-26
[PUBDEV-7843] build XGB CV models on multiple GPUs in parallel

honzasterba:


build.gradle
```diff
@@ -221,6 +221,9 @@ subprojects {
                 url "${localNexusLocation}/public"
             }
         } else {
+            maven {
```
koniecsveta, line 224: I think it would be better to place this part common for both if else branches before if statement and negate the condition

h2o-extensions/xgboost/build.gradle
```diff
@@ -1,7 +1,7 @@
 description = "H2O XGBoost support"
 
 ext {
-    xgboost4jVersion = "1.2.0.7"
+    xgboost4jVersion = "1.2.0.6-honza-gpu_id_fail-SNAPSHOT"
```
koniecsveta, line 4: is this correct?

honzasterba: this is temporary

build.gradle
honzasterba: will be removed once proper xgb is released, before merging this

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -674,63 +674,17 @@ public Vec cv_AssignFold(int N) {
...
-  /**
-   * Runs given model builders in bulk.
-   *
-   * @param modelType text description of group of models being built (for logging purposes)
-   * @param job parent job (processing will be stopped if stop of a parent job was requested)
-   * @param modelBuilders list of model builders to run in bulk
-   * @param parallelization level of parallelization (how many models can be built at the same time)
-   * @param updateInc update increment (0 = disable updates)
-   */
-  public static void bulkBuildModels(String modelType, Job job, ModelBuilder<?, ?, ?>[] modelBuilders,
```
michalkurka, line 690: This method was originally built for CV models only and then made public for MLI. I don't think it is used in MLI anymore.

h2o-core/src/main/java/hex/BulkModelBuilder.java
```diff
@@ -0,0 +1,86 @@
+package hex;
+
+import water.H2O;
+import water.Job;
+import water.util.Log;
+
+/**
+ * Runs given model builders in bulk.
+ */
+public class BulkModelBuilder {
```
michalkurka, line 10: I would rather specialize this class just for CV models building. It still contains references to CV, eg.

     Log.info("Skipping build of last " + (N - i) + " out of " + N + " " + modelType + " CV models");

Moreover, it has certain assumptions about what models are being built. The main assumption is that they will  be somewhat similar in terms of how long it will take to build them. If you have models that take a dramatically different time to build it will become suboptimal in terms of how it is using resources.

h2o-core/src/main/java/hex/BulkModelBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+                stopAll(submodel_tasks);
+                throw new Job.JobCancelledException();
+            }
+            Log.info("Building " + modelType + " model " + (i + 1) + " / " + N + ".");
+            prepare(modelBuilders[i]);
+            submodel_tasks[i] = H2O.submitTask(modelBuilders[i].trainModelImpl());
+            if (++nRunning == parallelization) { //piece-wise advance in training the models
+                while (nRunning > 0) try {
+                    int waitForTaskIndex = i + 1 - nRunning;
+                    submodel_tasks[waitForTaskIndex].join();
```
michalkurka, line 57: This is the reason why this method is not efficient in the general case of model building. If the model you are waiting for happens to take a much longer time then the rest then you will be underutilizing you resources: you have one you are waiting for and n-1 already completed. In the case of GPUs it is especially bad because these are not a shared resource as compared to CPUs.

I would rename this class to CVModelBuilder or such and/or modify the javadoc.

michalkurka:
@honzasterba very cool! minor comments posted - I would just clarify the use of BulkModelBuilder - it works really well mainly for CV but in generic case, it can be sub-optimal.

h2o-core/src/main/java/hex/BulkModelBuilder.java
michalkurka: Take a look at ModelBuilderHelper#trainModelsParallel for a generic and efficient parallel model builder.

It should be somewhat clear when to use "BulkModelBuilder" and when `trainModelsParallel`. Maybe ModelBuilderHelper needs to be refactored? And we could have CVModelBuilder and ParallelModelBuilder?

WDYT?

h2o-core/src/main/java/hex/ModelBuilder.java
honzasterba: @dvorka please confirm the above or review if this change would break MLI

dvorka: @navdeep-G ^

h2o-core/src/main/java/hex/BulkModelBuilder.java
honzasterba: renamed

h2o-core/src/main/java/hex/BulkModelBuilder.java
honzasterba: improved things a bit, bigger refactor might be in order

h2o-core/src/main/java/hex/ModelBuilder.java
navdeep-G: This is not used in MLI

navdeep-G (approved):

michalkurka (approved):
Thank you Honza!

PR #5076, opened 2020-10-26
PUBDEV-7855: Make varimp feature consolidation more robust

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7855

This PR removes possible incorrect mapping between varimp and original column name which was done by matching the longest prefix. Now it generates possible column names from the column name and its domains thus removing possibility of a mismatch due to a common prefix.

However, there is still possibility of having a mismatch when the names of varimp aren't unique.

For example:
```
library(h2o)
df <- data.frame(y=rep_len(1:5, 1000))
df$a <- as.factor(c("b.c", "d"))
df$a.b <- as.factor(letters[1:5])
df$y <- df$y * c(1, 10)

glm <- h2o.glm(y="y", training_frame = as.h2o(df))
h2o.varimp(glm)
```

=>
```
variable relative_importance scaled_importance   percentage
1    a.b.c        1.330686e+01      1.000000e+00 2.278348e-01
2      a.d        1.330686e+01      1.000000e+00 2.278348e-01
3    a.b.e        1.060822e+01      7.971997e-01 1.816298e-01
4    a.b.a        1.060822e+01      7.971997e-01 1.816298e-01
5    a.b.d        5.287787e+00      3.973731e-01 9.053541e-02
6    a.b.b        5.287787e+00      3.973731e-01 9.053541e-02
7    a.b.c        1.953338e-14      1.467919e-15 3.344429e-16
```

(see the first and the last line of the varimp)

### DIFF
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -421,6 +421,17 @@ class NumpyFrame:
             yield col, self.get(col, with_categorical_names)
 
 
+
+def _get_domain_mapping(model):
+    """
+    Get a mapping between columns and their domains.
+
+    :return: Dictionary containing a mapping column -> factors
+    """
+    output = model._model_json["output"]
+    return dict(zip(output["names"], output["domains"]))
+
+
 def _shorten_model_ids(model_ids):
     import re
     regexp = re.compile(r"(.*)_AutoML_[\d_]+((?:_.*)?)$")  # nested group needed for Py2
@@ -1349,14 +1360,25 @@ def _consolidate_varimps(model):
     consolidated_varimps = {k: v for k, v in varimp.items() if k in x}
     to_process = {k: v for k, v in varimp.items() if k not in x}
 
+    domain_mapping = _get_domain_mapping(model)
+    encoded_cols = ["{}.{}".format(name, domain)
+                    for name, domains in domain_mapping.items()
+                    if domains is not None
+                    for domain in domains + ["missing(NA)"]]
+    if len(encoded_cols) > len(set(encoded_cols)):
+        for x in set(encoded_cols):
+            encoded_cols.remove(x)
+        raise RuntimeError("Ambiguous encoding of the column x category pairs: {}".format(set(encoded_cols)))
+
+    varimp_to_col = {"{}.{}".format(name, domain): name
+                     for name, domains in domain_mapping.items()
+                     if domains is not None
+                     for domain in domains + ["missing(NA)"]
+                     }
     for feature in to_process.keys():
-        col_parts = feature.split(".")
-        for i in range(1, len(col_parts) + 1)[::-1]:
-            if ".".join(col_parts[:i]) in x:
-                column = ".".join(col_parts[:i])
-                consolidated_varimps[column] = consolidated_varimps.get(column, 0) + to_process[
-                    feature]
-                break
+        if feature in varimp_to_col:
+            column = varimp_to_col[feature]
+            consolidated_varimps[column] = consolidated_varimps.get(column, 0) + to_process[feature]
         else:
             raise RuntimeError("Cannot find feature {}".format(feature))
 
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -1,6 +1,7 @@
 # -*- encoding: utf-8 -*-
 
 import random
+import warnings
 from contextlib import contextmanager
 from collections import OrderedDict, Counter, defaultdict
 
@@ -1366,9 +1367,10 @@ def _consolidate_varimps(model):
                     if domains is not None
                     for domain in domains + ["missing(NA)"]]
     if len(encoded_cols) > len(set(encoded_cols)):
+        duplicates = encoded_cols[:]
         for x in set(encoded_cols):
-            encoded_cols.remove(x)
-        raise RuntimeError("Ambiguous encoding of the column x category pairs: {}".format(set(encoded_cols)))
+            duplicates.remove(x)
+        warnings.warn("Ambiguous encoding of the column x category pairs: {}".format(set(duplicates)))
 
     varimp_to_col = {"{}.{}".format(name, domain): name
                      for name, domains in domain_mapping.items()
--- a/h2o-r/h2o-package/R/explain.R
+++ b/h2o-r/h2o-package/R/explain.R
@@ -134,6 +134,16 @@ with_no_h2o_progress <- function(expr) {
   return(!.get_algorithm(model) %in% c("stackedensemble", "naivebayes"))
 }
 
+#' Get a mapping between columns and their domains
+#' @param model an h2o model
+#' @return list containing a mapping from column to its domains (levels)
+.get_domain_mapping <- function(model) {
+  domains <- model@model$domains
+  names(domains) <- model@model$names
+  return(domains)
+}
+
+
 #' Shortens model ids if possible (iff there will be same amount of unique model_ids as before)
 #' @param model_ids character vector
 #' @return character vector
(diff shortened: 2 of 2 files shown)


h2o-r/h2o-package/R/explain.R
```diff
@@ -504,19 +504,28 @@ with_no_h2o_progress <- function(expr) {
...
+      for (domain in c("missing(NA)", domains[[col]])) {
+        col_domain_mapping[[paste0(col, ".", domain)]] <- col
       }
     }
-    if (!found)
+  }
+
+  if (anyDuplicated(names(col_domain_mapping))) {
+    dups <- duplicated(names(col_domain_mapping)) | duplicated(names(col_domain_mapping), fromLast = TRUE)
+    stop("Ambiguous encoding of the column x category pairs: ", col_domain_mapping[dups])
```
tomasfryda, line 519: Unfortunately, this can happen:
```
library(h2o)
df <- data.frame(y=rep_len(1:5, 1000))
df$a <- as.factor(c("b.c", "d"))
df$a.b <- as.factor(letters[1:5])
df$y <- df$y * c(1, 10)

glm <- h2o.glm(y="y", training_frame = as.h2o(df))
h2o.varimp(glm)
```
=>
```
variable relative_importance scaled_importance   percentage
1    a.b.c        1.330686e+01      1.000000e+00 2.278348e-01
2      a.d        1.330686e+01      1.000000e+00 2.278348e-01
3    a.b.e        1.060822e+01      7.971997e-01 1.816298e-01
4    a.b.a        1.060822e+01      7.971997e-01 1.816298e-01
5    a.b.d        5.287787e+00      3.973731e-01 9.053541e-02
6    a.b.b        5.287787e+00      3.973731e-01 9.053541e-02
7    a.b.c        1.953338e-14      1.467919e-15 3.344429e-16
```

(see the first and the last line of the varimp)

ledell (approved):
This looks good, thanks!

sebhrusen:
@tomasfryda why not using sth like brackets if prefix is not enough?
in you example, you would obtain  sth like:
```
variable relative_importance scaled_importance   percentage
1    a[b.c]        1.330686e+01      1.000000e+00 2.278348e-01
2      a[d]        1.330686e+01      1.000000e+00 2.278348e-01
3    a.b[e]        1.060822e+01      7.971997e-01 1.816298e-01
4    a.b[a]        1.060822e+01      7.971997e-01 1.816298e-01
5    a.b[d]        5.287787e+00      3.973731e-01 9.053541e-02
6    a.b[b]        5.287787e+00      3.973731e-01 9.053541e-02
7    a.b[c]        1.953338e-14      1.467919e-15 3.344429e-16
```
avoiding any conflict, i think even in the most exotic situations...

tomasfryda:
Thanks @sebhrusen for the suggestion! I like it, however, it has some problems particularly with R client. R strongly prefers usage of syntactically valid names which makes usage of almost any non-alphanumeric character problematic[1] (only something like this is ok: `[A-Za-z_][A-Za-z0-9\._]*`)

Anyway, I would like to include this fix in a minor release so I would prefer not to do user visible changes in it.

[1] https://stat.ethz.ch/R-manual/R-patched/library/base/html/make.names.html

sebhrusen:
@tomasfryda are you using those names as attrributes as well? I thought these were just values for display.

So you call `make.names(df$variable)` somewhere to generate those values?

h2o-r/h2o-package/R/models.R
```diff
@@ -5069,3 +5069,12 @@ h2o.reset_threshold <- function(object, threshold) {
     return(NULL)
   }
 }
+
+#' Get a mapping between columns and their domains
+#' @param model an h2o model
+#' @return list containing a mapping from column to its domains (levels)
+.h2o.__get_domain_mapping <- function(model) {
+  domains <- model@model$domains
+  domains <- stats::setNames(domains, model@model$names)
```
sebhrusen, line 5078: how different is it from usual
```
names(domains) <- model@model$names
```

tomasfryda:
@sebhrusen I don't think I call directly `make.names` anywhere, however, I make plots and I think `ggplot2` calls it (or some `tidyverse`/`rlang` alternative of it) and possibly other libraries as well (when I tried it I had some issues with `stats::reshape`) and I do the varimp consolidation in the R client code. The problem is that even in base R some functions call `make.names` by default, e.g., `data.frame` will call `make.names`, while `as.data.frame` doesn't.

And unfortunately in R it doesn't really help with the ambiguity when encoded with the `make.names`.
```
> make.names(c("a[b.c]", "a.b[c]",  "a b c "))
[1] "a.b.c." "a.b.c." "a.b.c."
```

I tried to implement this change in java and in python it was quite straightforward to make it work (at least the explainability module) but not in R (but I can investigate the issue further if needed).

h2o-r/h2o-package/R/models.R
tomasfryda: There isn't any difference except for the extra function call. Your version is better.

It used to be:
```
return(stats::setNames(domains, model@model$names))
```

and then I remembered that not everybody enjoys lisp and parentheses so I decided to make it more readable by splitting it in to two lines.

sebhrusen:
I understand that as soon as `make.names` is applied somewhere, then some confusion or even conflicts may appear.
Just saying that if we can at least avoid those in the table, we should not refrain.

tomasfryda:
@sebhrusen I changed this PR back to a state where I use just the `variable.category` names and I keep it strictly in the `explain` module. 
I still have a backup of this PR where I used `variable{category}` names but @michalkurka had some concerns about backward compatibility and other potential issues.

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1349,14 +1360,25 @@ def _consolidate_varimps(model):
...
 
+    domain_mapping = _get_domain_mapping(model)
+    encoded_cols = ["{}.{}".format(name, domain)
+                    for name, domains in domain_mapping.items()
+                    if domains is not None
+                    for domain in domains + ["missing(NA)"]]
+    if len(encoded_cols) > len(set(encoded_cols)):
+        for x in set(encoded_cols):
+            encoded_cols.remove(x)
+        raise RuntimeError("Ambiguous encoding of the column x category pairs: {}".format(set(encoded_cols)))
```
sebhrusen, line 1371: So we prefer raising an error rather than fixing the encoding syntax? Interesting…

A simple compatible way would be to externalize the encoding format, ie. expose it as a global variable:
```
varimp_col_format="{name}[{domain}]"
```
and tell users that they should set `h2o.explanation.varimp_col_format="{name}.{domain}"` if they have compatibility issues...
@michalkurka ?

The problem with raising an error here, is that is doesn't really help the user: are we basically asking him to rename the columns of his training data and retrain his model? This is asking a lot for something that can be fixed easily directly on client side.

tomasfryda: > This is asking a lot for something that can be fixed directly on client side.

@sebhrusen I don't think this could be solved on client side. The problem is created on the backend -  we don't validate if the encoded name is unique.

Imagine situation where we have a categorical column `A` with categories `a`,`b`, and `a.b` and another categorical column `A.a` with categories `a`, and `b`. The backend will produce varimp for some algos with names like `A.a.b` and `A.a.b` and on client I have to decode it. Unfortunately, varimp is sorted by variable importance and not the actual order of `names`, so we cannot easily decode it on backend either.

Using `"{name}[{domain}]"` would have the same issues if we can have square brackets in the `name` and `domain`, e.g., `name = "A[b]", domain = "c"` would end up encoded the same way as `name = "A", domain ="b][c"`. And for cases like that, there's the exception.

Or I am missing something that can be done on client side for that?

michalkurka: I think a proper solution is on the backend: the backend should be able to decode even ambiguous encodings because it knows the logic of how they were created (the order of the columns is critical for decoding ambiguous)

throwing an error doesn't seem like the best solution to me

I think the best solution here would be if varimp method also returned the original feature name as a new column in the output

we can keep this PR but turn the error into a warning instead - to inform user we are not doing it right (for now)

sebhrusen: @tomasfryda I agree that there's no unique syntax that will work universally: this is the reason why I suggest to give the possibility for the user to chose a format that will work for his data.

My preference for brackets instead of dots as the default is just a question of probability: it's much more common to have a single dot in a column name than `][`…
But I wouldn't mind keeping "{name}.{domain}" as the default.
My point is simply that if we detect a conflict, we could print a warning (or maybe even an error) suggesting to change  the global varimp format to avoid the conflict.

> I think the best solution here would be if varimp method also returned the original feature name as a new column in the output

yes, having raw information in addition to the encoded one is always better.

tomasfryda: Looking at XGBoostNativeVarImp parsing[1], I am not sure how easy it will be to get the original feature name in that case and also it looks to me that XGBoost merges the conflicting encoded variables:

```r
library(h2o)
df <- data.frame(y=rep_len(1:5, 1000))
df$a <- as.factor(c("b.c", "d"))
df$a.b <- as.factor(letters[1:5])
df$y <- df$y * c(1, 10)

xgb <- h2o.xgboost(y="y", training_frame = as.h2o(df))
h2o.varimp(xgb)
```
=>
```
Variable Importances: 
  variable relative_importance scaled_importance percentage
1    a.b.c       386479.750000          1.000000   0.702795
2    a.b.a        99022.585938          0.256217   0.180068
3    a.b.b        58419.015625          0.151157   0.106232
4    a.b.d         5996.760742          0.015516   0.010905
5    a.b.e            0.000029          0.000000   0.000000

```

```r
library(h2o)
df <- data.frame(y=rep_len(1:5, 1000))
df$a <- as.factor(c("b.c", "d"))
df$ab <- as.factor(letters[1:5]) # NOTE: the only difference is in the name on this line
df$y <- df$y * c(1, 10)

xgb <- h2o.xgboost(y="y", training_frame = as.h2o(df))
h2o.varimp(xgb)
```
=>
```
Variable Importances: 
  variable relative_importance scaled_importance percentage
1    a.b.c       359073.187500          1.000000   0.652958
2     ab.a        99022.585938          0.275773   0.180068
3     ab.b        58419.015625          0.162694   0.106232
4     ab.c        27406.578125          0.076326   0.049838
5     ab.d         5996.760742          0.016701   0.010905
6     ab.e            0.000029          0.000000   0.000000
```

Personally, I'd rather change the errors to warnings and then gradually improve the code base than getting stuck at trying to make the @michalkurka's solution (for that I'd create a new JIRA). Is that ok with @sebhrusen and @michalkurka ? 

[1] https://github.com/h2oai/h2o-3/blob/master/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/predict/XGBoostNativeVariableImportance.java#L70-L99

michalkurka: absolutely - the good first step is to issue a warning (basically in this PR just error -> warning)

This will give us time to figure out the next step and the proper solution.

michalkurka (approved):

michalkurka:
thank you @tomasfryda 

any comments @sebhrusen?

sebhrusen (approved):
Good to me until  we have backend fix. Thanks @tomasfryda

PR #5077, opened 2020-10-26
XGB: Add NCCL to docker build image

michalkurka:


docker/xgb_build/Dockerfile-gpu-centos
```diff
@@ -60,6 +60,11 @@ RUN \
     /usr/sbin/install_python_source_centos 3.7.0 /usr/src/Setup.dist.patched && \
     /usr/sbin/install_python_source_centos 3.8.0
 
+# Install the NCCL library
+RUN \
```
honzasterba, line 64: is this statically linked so we do not need to add it to runtime images?

michalkurka: it is statically linked

honzasterba (approved):

PR #5080, opened 2020-10-27
Cleans-up dead code in Vec

michalkurka:


h2o-core/src/main/java/water/fvec/Vec.java
```diff
@@ -258,7 +258,6 @@ public Vec( Key<Vec> key, int rowLayout, String[] domain, byte type ) {
     assert key._kb[0]==Key.VEC;
     assert domain==null || type==T_CAT;
     assert T_BAD <= type && type <= T_TIME; // Note that T_BAD is allowed for all-NA Vecs
-    setMeta(type,domain);
```
mn-mikke, line 261: Shouldn't the constructor and `makeCopy(String[] domain)` have the same behavior?

michalkurka: Same behavior as what? Please give me an example.

`makeCopy(String[] domain)` is sometimes used to strip domain out of a categorical Vector and convert it to just domain indices. This behavior should stay IMHO.

mn-mikke: I see, thanks for your explanation!

> we have this constructor corresponding and logically consistent with makeCopy(byte type, String[] domain)

In this case, shouldn't the constructor throw an exception when `type==T_CAT` and `domani==null` as it's specified [here](https://github.com/h2oai/h2o-3/pull/5080/files#diff-e8795931cbf3b4f83c6c675244ee51c1224e8bf56fc8d940664a68e049b8009eR475)? Or vice versa (`makeCopy(String[] domain, byte type)` won't throw an exception).

michalkurka: I think there are 2 dimensions to this question:
1) Whether to use exception or assert
2) How should the condition look like

ad 1) There is no written rule but the old core traditionally uses asserts (like this one https://github.com/h2oai/h2o-3/pull/5080/files#diff-e8795931cbf3b4f83c6c675244ee51c1224e8bf56fc8d940664a68e049b8009eL259) instead of exceptions. This is applied mostly to the "kernel-space code" (not to user-space code). In the user-facing methods, we prefer to use exceptions because we cannot rely on extensive testing the user will do and we cannot rely on assertions being enabled. You cannot really easily tell what is kernel-space and what is user-space. But usages of the function are usually good guidance for a novice (you will see makeCopy is almost exclusively used in Ast-classes). The constructor of Vec is expected to be used by core only and rarely other code.

ad 2) The constructor already has an appropriate assertion here: https://github.com/h2oai/h2o-3/pull/5080/files#diff-e8795931cbf3b4f83c6c675244ee51c1224e8bf56fc8d940664a68e049b8009eL259. Please note that this allowed the vec to be either T_BAD and T_NUM and still have a domain. This is both okay. For the T_NUM case it is less obvious but it is the case of InteractionWrappedVec - see the third condition of https://github.com/h2oai/h2o-3/pull/5080/files#diff-e8795931cbf3b4f83c6c675244ee51c1224e8bf56fc8d940664a68e049b8009eR221

The reason why the condition in `makeCopy` is different is that the result cannot be eg. InteractionWrappedVec and we thus don't need to worry about all possible cases.

mn-mikke: ok, thanks for giving me the context.

mn-mikke (approved):

PR #5082, opened 2020-10-28
Fix warning from old matplotlib.use

tomasfryda:
This should stop emitting the following warning:
```
This call to matplotlib.use() has no effect because the backend has already been chosen; matplotlib.use() must be called *before* pylab, matplotlib.pyplot, or matplotlib.backends is imported for the first time.
```
from **r_suite.pyunit_vec_scaler_comparisions_medium.py**.

I wasn't able to reproduce the issue locally - I ran just the one test so there wasn't anything that would set a backend before it so I suppose this happens only when the tests aren't completely isolated?

Nevertheless, this might still be shown to the user.

Newer versions (3.2+) of `matplotlib` don't have the `warn` parameter in `matplotlib.use`, versions 3.1 higher deprecates it and default `warn` to `False`.

The solution would be prettier if it would be contained in a simple function, any ideas where to put it?

### DIFF
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -27,6 +27,7 @@ from h2o.expr import ExprNode
 from h2o.group_by import GroupBy
 from h2o.job import H2OJob
 from h2o.utils.config import get_config_value
+from h2o.utils.ext_dependencies import get_matplotlib_pyplot
 from h2o.utils.shared_utils import (_handle_numpy_array, _handle_pandas_data_frame, _handle_python_dicts,
                                     _handle_python_lists, _is_list, _is_str_list, _py_tmp_key, _quoted,
                                     can_use_pandas, quote, normalize_slice, slice_is_normalized, check_frame_id)
@@ -3773,21 +3774,8 @@ class H2OFrame(Keyed):
         hist = H2OFrame._expr(expr=ExprNode("hist", self, breaks))._frame()
 
         if plot:
-            try:
-                import matplotlib
-                if server:
-                    # Matplotlib version 3.1 and higher deprecates/doesn't have `warn` parameter
-                    version = matplotlib.__version__.split(".")
-                    if int(version[0]) < 3 or \
-                            (int(version[0]) == 3 and
-                             int(version[1]) < 1):
-                        matplotlib.use("Agg", warn=False)
-                    else:
-                        matplotlib.use("Agg")
-                import matplotlib.pyplot as plt
-            except ImportError:
-                print("ERROR: matplotlib is required to make the histogram plot. "
-                      "Set `plot` to False, if a plot is not desired.")
+            plt = get_matplotlib_pyplot(server)
+            if plt is None:
                 return
 
             hist["widths"] = hist["breaks"].difflag1()
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -3776,7 +3776,14 @@ class H2OFrame(Keyed):
             try:
                 import matplotlib
                 if server:
-                    matplotlib.use("Agg")
+                    # Matplotlib version 3.1 and higher deprecates/doesn't have `warn` parameter
+                    version = matplotlib.__version__.split(".")
+                    if int(version[0]) < 3 or \
+                            (int(version[0]) == 3 and
+                             int(version[1]) < 1):
+                        matplotlib.use("Agg", warn=False)
+                    else:
+                        matplotlib.use("Agg")
                 import matplotlib.pyplot as plt
             except ImportError:
                 print("ERROR: matplotlib is required to make the histogram plot. "
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -11,6 +11,7 @@ from h2o.exceptions import H2OValueError
 from h2o.job import H2OJob
 from h2o.utils.metaclass import BackwardsCompatible, Deprecated as deprecated, h2o_meta
 from h2o.utils.compatibility import viewitems
+from h2o.utils.ext_dependencies import get_matplotlib_pyplot
 from h2o.utils.shared_utils import can_use_pandas
 from h2o.utils.typechecks import I, assert_is_type, assert_satisfies, Enum, is_type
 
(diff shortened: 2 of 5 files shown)


h2o-py/h2o/frame.py
```diff
@@ -3776,7 +3776,14 @@ def hist(self, breaks="sturges", plot=True, **kwargs):
...
                 if server:
-                    matplotlib.use("Agg")
+                    # Matplotlib version 3.1 and higher deprecates/doesn't have `warn` parameter
+                    version = matplotlib.__version__.split(".")
+                    if int(version[0]) < 3 or \
+                            (int(version[0]) == 3 and
+                             int(version[1]) < 1):
+                        matplotlib.use("Agg", warn=False)
+                    else:
+                        matplotlib.use("Agg")
```
sebhrusen, line 3786: do we always use the same pattern when loading `matplotlib`? Can't we externalize this in a separate function?

sebhrusen: also for fun:
```
from distutils.version import LooseVersion as lv

if lv(version) < lv("3.1"):
    ...
```

sebhrusen: btw, you  write:
> Newer versions (3.2+) of matplotlib don't have the warn parameter

so it looks like the comparison operator should be `<=` instead of `<`

sebhrusen: > The solution would be prettier if it would be contained in a simple function, any ideas where to put it?

yeah, I agree.
The only module with similar function is probably `h2o.utils.shared_utils`: there's already way to many things there but well... it also contains functions like `can_use_pandas`.

Maybe it's worth creating a `h2o.utils.imports` module (or `h2o.utils.ext_dependencies`)  and move all those external modules detection logic + custom imports there.

michalkurka: I think the original code is already a copy&paste, IMHO all we need to do is to

take `_get_matplotlib_pyplot` from `ModelBase`, put it in a utility class and use everywhere.

WDYT?

tomasfryda: Thanks for the suggestions! I looked into using `warnings` related functions to suppress it (something that @michalkurka suggested on slack) and there is couple issues with that, namely:

- adding a filter is modifying a global state, that can cause some issues when working with threads/asyncio
- some packages modify the warnings filter (that could override the filter that we would setup) or reroute the warning to logger (which still causes it to be shown in, e.g., Jupyter) (e.g. some versions of tensorflow)
- the warning messages differ between versions (at least between 2.x.x and 3.x.x) so I would either suppress all `UserWarnings` or go through the versions to find out if and in what they differ.

and while we could use `warnings.catch_warnings` context manager to make the filter more local, it would still have undefined behavior when using with multiple threads.

So I think the best approach is not to emit the warning at all by detecting the version and using the appropriate arguments.

sebhrusen: > and while we could use warnings.catch_warnings context manager to make the filter more local, it would still have undefined behavior when using with multiple threads.

maybe you're overthinking this, aren't you? I don't see many use cases where users would use multithread for plotting, doesn't sound reasonable, and it's definitely not the case in Jupyter/ipython environment.

tomasfryda: I guess I am. So you think suppressing the warning using filter would be a better solution?

(The warning is emitted when providing `server=True` where I guess it is a bit more likely to have multiple threads than just during regular plotting, but you are right that it would be a very unlikely scenario.)

h2o-py/h2o/model/model_base.py
```diff
@@ -1713,20 +1714,6 @@ def score_history(self):
...
-        # noinspection PyUnresolvedReferences
-        import matplotlib
-        if server: matplotlib.use("Agg")
-        # noinspection PyUnresolvedReferences
-        import matplotlib.pyplot as plt
-        return plt
-    except ImportError:
-        print("`matplotlib` library is required for this function!")
-        return None
-
```
sebhrusen, line 1729: wouldn't it make sense to move the other import utility functions below to the same py module?
I know it's out of scope somehow, but it would improve consistency.

h2o-py/h2o/frame.py
sebhrusen: if the warning is emitted exactly when calling `matplotlib.use(...)`, then the current solution looks good to me.

The context manager approach is a bit cleaner, but I don't have a strong preference here. I'd validate both.

h2o-py/h2o/model/model_base.py
tomasfryda: I agree that it would make sense and it would be nicer to have more consistent package but a simple grep reveals that except for `matplotlib`, there aren't many repeating "try-import" patterns (on this branch):

```
$ cd h2o-py/h2o 
$ rg -U '\s*try:[\w\W]*?(except|finally)' . | perl -pe 's/^.*?:\s*(.*?)(?:\s*#.*)?$/\1/g' | rg '(import|from) ' | sort |uniq -c |sort -nr

      2 import matplotlib.pyplot as plt
      1 raise RuntimeError('Cannot import pandas')
      1 import winkerberos as kerberos
      1 import subprocess
      1 import pandas
      1 import numpy as np
      1 import matplotlib.figure
      1 import matplotlib
      1 import kerberos
      1 import h2o.explanation
      1 import gssapi
      1 import IPython.display
      1 import IPython
      1 from termios import TIOCGWINSZ
      1 from struct import unpack
      1 from sklearn.externals.funcsigs import Parameter, signature
      1 from requests.auth import AuthBase
      1 from mpl_toolkits.mplot3d import Axes3D
      1 from matplotlib import cm
      1 from ipykernel.zmqshell import ZMQInteractiveShell
      1 from inspect import Parameter, signature
      1 from h2o.explanation import explain, explain_row, varimp_heatmap, model_correlation_heatmap, \
      1 from fcntl import ioctl
      1 from distutils.version import LooseVersion
      1 from dis import _unpack_opargs
      1 from ..utils.shared_utils import can_use_numpy, can_use_pandas
```

The fact that those imports are mostly used once makes me think that it's not worth doing in this PR.

maurever (approved):
Looks good to me. Thanks @tomasfryda.

sebhrusen (approved):
Thx Tomaš, looks good

PR #5084, opened 2020-10-29
[PUBDEV-7866] Fix parsing of GLM.rand_family params in MOJO JSON

honzasterba:
- also make MOJO JSON parsing only log errors, not throw exceptions

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/attributes/ModelJsonReader.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/attributes/ModelJsonReader.java
@@ -5,6 +5,8 @@ import hex.genmodel.*;
 import hex.genmodel.attributes.parameters.ColumnSpecifier;
 import hex.genmodel.attributes.parameters.KeyValue;
 import hex.genmodel.attributes.parameters.ParameterKey;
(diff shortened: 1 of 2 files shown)


mn-mikke (approved):
LGTM

h2o-genmodel/src/main/java/hex/genmodel/attributes/ModelJsonReader.java
```diff
@@ -392,7 +401,8 @@ private static Object convertJsonObject(final JsonObject convertFrom) {
                 convertFrom.get("value").getAsDouble()
             );
         } else {
-            throw new UnsupportedOperationException(String.format("Object not supported: \n %s ", convertFrom.toString()));
+            LOG.warn(String.format("Error reading MOJO JSON: Object not supported: \n %s ", convertFrom.toString()));
```
mn-mikke, line 404: nit: Shouldn't this case be logged as error?

honzasterba: fixed

Pscheidl (approved):

PR #5086, opened 2020-10-30
[PUBDEV-7852] Refactor PreviewParseWriter to 'guessTypes' Method to be Used Independently

mn-mikke:
This PR helps to get rid of [IcedHashMapWrapper](https://github.com/h2oai/sparkling-water/blob/b2f2d55009845a4be9e5d9f0fd3d9b63140addca/extensions/src/main/scala/water/parser/IcedHashMapWrapper.java) in Sparkling Water.

### DIFF
--- a/h2o-core/src/main/java/water/parser/PreviewParseWriter.java
+++ b/h2o-core/src/main/java/water/parser/PreviewParseWriter.java
@@ -131,16 +131,25 @@ public class PreviewParseWriter extends Iced implements StreamParseWriter {
 
   @Override public void setIsAllASCII(int colIdx, boolean b) {}
 
-  public byte[] guessTypes() {
-    IDomain[] domains = new IDomain[_domains.length];
-    for (int i = 0; i < _domains.length; i++) {
-      IcedHashMap<String,String> sourceDomain = _domains[i];
-      domains[i] = new IDomain() {
+  public byte[] guessTypes() { 
+    byte[] types = new byte[_ncols];
+    for (int i = 0; i < _ncols; ++i) {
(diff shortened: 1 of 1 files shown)


honzasterba (approved):
lgtm

h2o-core/src/main/java/water/parser/PreviewParseWriter.java
```diff
@@ -131,86 +131,102 @@ public void addNAs(int colIdx, int nrow) {
...
+    
+    // categorical or string?
+    // categorical with 0s for NAs
+    if (nzeros > 0
+        && ((nzeros + nstrings) >= nonemptyLines) //just strings and zeros for NA (thus no empty lines)
+        && (domain.size() <= 0.95 * nstrings)) { // not all unique strings
+      return Vec.T_CAT;
+    }
+    // categorical mixed with numbers
+    if (nstrings >= (nnums+nzeros) // mostly strings
```
Pscheidl, line 222: Unnecessary parentheses + wrong formatting (minor issue). Feel free to keep the parentheses in place if you feel like it, please. This is only a matter of personal taste.

mn-mikke: @Pscheidl It was like this before, I just changed the variable names. I will fix the parentheses and spaces around `+`, but will leave it splitted the condition into two lines due to preserving the comments.

mn-mikke: On second thought, I've decided to leave the parentheses as they are. They are used like this on more places within this class. If the usage is inappropriate, we could fix it in a different PR, just dedicated to this problem.

PR #5089, opened 2020-10-30
[PUBDEV-7859] save frame in chunks

honzasterba:
... this is the first part of many PRs leading to fault-tolerant Grid search in SW.

This code is necessary to quicky dump and re-load Frames


h2o-core/src/main/java/water/fvec/persist/FramePersist.java
```diff
@@ -0,0 +1,111 @@
...
+
+        private void writeChunks(AutoBuffer autoBuffer, Chunk[] chunks) {
+            for (Chunk c : chunks) {
+                autoBuffer.put(c);
+            }
+        }
+    }
+
+    public static Frame loadFrom(Key<Frame> key, String uri) {
+        uri = sanitizeUri(uri);
```
koniecsveta, line 77: I would inline this method call and rather avoid rewriting uri parameter, since it is used only on next line, would do just `URI metaUri = getMetaUri(key, sanitizeUri(uri));`.

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public FramePersist(Frame frame) {
+        this.frame = frame;
+    }
+    
+    private static class FrameMeta extends Iced<FrameMeta> {
+        Key<Frame> key;
+        String[] names;
+        Key<Vec>[] keys;
+        Vec[] vecs;
```
michalkurka, line 29: How does this save Vec.ESPC?

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    private static URI getDataUri(String metaUri, int cidx) {
+        return FileUtils.getURI(metaUri + "_" + H2O.SELF.index() + "_" + cidx);
+    }
+
+    public void saveTo(String uri) {
+        uri = sanitizeUri(uri);
+        URI metaUri = getMetaUri(frame._key, uri);
+        write(metaUri, ab -> {
```
michalkurka, line 50: it would be safer to first have the chunks loaded and after that load the frame envelope, otherwise something can start working with the partial frame before the chunks are actually available

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    public static Frame loadFrom(Key<Frame> key, String uri) {
+        uri = sanitizeUri(uri);
+        URI metaUri = getMetaUri(key, uri);
+        FrameMeta meta = read(metaUri, AutoBuffer::get);
+        Vec con = Vec.makeConN(meta.vecs[0].length(), meta.vecs[0].nChunks());
+        new LoadChunksTask(metaUri.toString(), meta.vecs).doAll(con);
+        for (Vec v : meta.vecs) DKV.put(v);
```
michalkurka, line 82: I would do this in parallel (asynchronous DKV.put)

Futures fs = new Futures():
for 
   DKV.put(v, fs);
fs.blockForPending();

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        @Override
+        public void map(Chunk c) {
+            PersistUtils.read(getDataUri(metaUri, c.cidx()), ab -> readChunks(ab, c.cidx()));
+        }
+
+        private int readChunks(AutoBuffer autoBuffer, int cidx) {
+            for (Vec v : vecs) {
+                DKV.put(v.chunkKey(cidx), new Value(v.chunkKey(cidx), autoBuffer.get()));
```
michalkurka, line 105: note to self: this will end up being a local put thanks to the MRTask so I don't think we need parallel put

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
honzasterba: good catch, just found it by my self, pushed fix

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
honzasterba: fxied

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
```diff
@@ -0,0 +1,189 @@
...
+import water.util.FileUtils;
+
+import java.net.URI;
+
+import static water.fvec.persist.PersistUtils.*;
+
+public class FramePersist {
+
+    static {
+        TypeMap.onIce(FrameMeta.class.getName());
```
michalkurka, line 17: I guess this is needed because this Iced class is never actually used in DKV? Is that correct? Can you add a small explanation as a comment to the code? Using the construct is usually limited to the very-core classes.

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
```diff
@@ -0,0 +1,189 @@
...
+
+        private final String metaUri;
+
+        SaveChunksTask(String metaUri) {
+            this.metaUri = metaUri;
+        }
+
+        @Override
+        public void map(Chunk[] cs) {
+            PersistUtils.write(getDataUri(metaUri, cs[0].cidx()), ab -> writeChunks(ab, cs));
```
michalkurka, line 106: the progress of the job is never incremented

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
```diff
@@ -0,0 +1,189 @@
...
+                int nchunk = meta.espc.length-1;
+                con = Vec.makeConN(nrow, nchunk);
+                new LoadChunksTask(metaUri, meta.vecs).doAll(con).join();
+            } finally {
+                if (con != null) con.remove();
+            }
+            Futures fs = new Futures();
+            int rowLayout = Vec.ESPC.rowLayout(meta.vecs[0]._key, meta.espc);
+            for (Vec v : meta.vecs) {
+                v._rowLayout = rowLayout;
```
michalkurka, line 152: This could be imho encapsulated in the FrameMeta class. (writeAll_impl, readAll_impl)

h2o-py/h2o/frame.py
```diff
@@ -2184,8 +2184,32 @@ def get_frame_data(self):
         >>> iris = h2o.import_file("http://h2o-public-test-data.s3.amazonaws.com/smalldata/iris/iris_wheader.csv")
         >>> iris.get_frame_data()
         """
-        return h2o.api("GET /3/DownloadDataset", data={"frame_id": self.frame_id, "hex_string": False, "escape_quotes" : True})
+        return h2o.api(
+            "GET /3/DownloadDataset", 
+            data={"frame_id": self.frame_id, "hex_string": False, "escape_quotes": True}
+        )
 
+    def save(self, path, force=True):
```
michalkurka, line 2192: regarding the load/save naming - is it appropriate for the native format? also for binary models, we have importModel/exportModel

h2o-core/src/test/java/water/fvec/persist/FramePersistTest.java
```diff
@@ -0,0 +1,113 @@
...
+            Frame f = Scope.track(new Frame(Key.make(), new Vec[] { v }));
+            DKV.put(f);
+            TwoDimTable origData = f.toTwoDimTable(0, (int) v.length(), false);
+            File dest = temp.newFolder();
+            new FramePersist(f).saveTo(dest.getAbsolutePath(), false).get();
+            f.remove(true);
+            Scope.track(FramePersist.loadFrom(f._key, dest.getAbsolutePath()).get());
+            Frame loaded = DKV.get(f._key).get();
+            TwoDimTable loadedData = loaded.toTwoDimTable(0, (int) loaded.numRows(), false);
+            assertTwoDimTableEquals(origData, loadedData);
```
michalkurka, line 69: why is the comparison done on 2d table? can we use assertFrameEquals instead? you just need to deepclone the original Frame

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
honzasterba: as far as I understand any class that I want to be de-serializable by Icer need to be registerd in the TypeMap, if I dont do this I get an error when trying to desrialize because there is no ID for this class in the TypeMap, maybe I am missing something else

h2o-core/src/test/java/water/fvec/persist/FramePersistTest.java
honzasterba: I just wanted to have a purely local, cluster independent representation of the data. making a deep clone of the frame sounds reasonable too

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
michalkurka: > if I dont do this I get an error when trying to desrialize because there is no ID for this class in the TypeMap, maybe I am missing something else

The classloader should take care of adding FrameMeta to typemap when the class is loaded. The typemap is then persisted and onIce is called when read back to H2O.

Maybe worth bit of debugging what is going on.

michalkurka: My theory is the one above - because the object is never written anywhere, the icer is not needed and typemap not updated. AutoBuffer first starts writing the type map and only after that you actually use the icer and thus update the typemap.

honzasterba: strange, this actually only happens in the JUnit test, in normal scenario this is not needed, but the first junit test fails when reading the written file with
`java.lang.ClassNotFoundException:  BAD
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:355)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at water.Weaver.classForName(Weaver.java:61)
	at water.TypeMap.theFreezable(TypeMap.java:249)
	at water.TypeMap.newFreezable(TypeMap.java:232)
	at water.AutoBuffer.get(AutoBuffer.java:770)
	at water.fvec.persist.PersistUtils.read(PersistUtils.java:26)
	at water.fvec.persist.FramePersist.loadFrom(FramePersist.java:114)`
this is not reproducible with python

honzasterba: I have dug throught the code and you are exactly right, sir, and I understand what your are saying, I have yet to find a solution.

honzasterba: clarification: i can reproduce this with python tests

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
honzasterba: fixed

h2o-py/h2o/frame.py
honzasterba: actually we have `h2o.load_model` and `h2o.save_model` in py, I am surprised we dont have `model.save()` that would the the only discrepancy with this API, should I move `frame.save()` to `h2o.save_frame()`? having an instance method on frame seemed more natural

h2o-core/src/main/java/water/fvec/persist/FramePersist.java
honzasterba: writeAll_impl, readAll_impl are specific to Keyed, FrameMeta is not keyed and I dont want to put it into DKV which AutoBuffer.getKey does, I could add method to FrameMeta that would do the ab.get() and then do the rowLayout post-processing of loaded vecs but that adds another for-loop iterating through all the vecs and the code does not seem cleaner, WDYT?

scripts/jenkins/hadoop/start.sh
```diff
@@ -86,13 +86,13 @@ hadoop jar h2o-hadoop-*/h2o-${hadoopVersion}-assembly/build/libs/h2odriver.jar \
     ${disownArgs} \
     -notify ${notifyFile} \
     > ${driverLogFile} 2>&1 &
-for i in $(seq 24); do
+for i in $(seq 36); do
```
honzasterba, line 89: since the timeout is 360s we should wait atleast that

michalkurka (approved):

Pscheidl (approved):

koniecsveta (approved):

PR #5093, opened 2020-11-03
Avoid managed-block when it could lead to a deadlock

michalkurka:


h2o-core/src/main/java/water/RPC.java
```diff
@@ -240,19 +240,30 @@ private V result() {
...
+    final boolean doManagedBlock = canDoManagedBlock();  
     if( _done ) return result(); // Fast-path shortcut, or throw if exception
     // Use FJP ManagedBlock for this blocking-wait - so the FJP can spawn
     // another thread if needed.
-    try { ForkJoinPool.managedBlock(this); } catch( InterruptedException ignore ) { }
+    try {
+      if (doManagedBlock)
+        ForkJoinPool.managedBlock(this);
+      else
+        block();
```
sebhrusen, line 265: to my understanding, the `doManagedBlock` condition applies for both managed block or "simple block", in the  sense that  `block` still calls `wait`, so you can still have a thread with higher priority waiting on `_dt` with lower priority, which can lead to deadlock...

That's why the else condition should be non-blocking imo, possibly using active waiting.

michalkurka: They do both call wait but the managed block will also replace the thread in the thread pool. So if a higher priority thread is waiting on a `_dt` with a lower priority it can create another higher priority thread that would be running instead of the lower-priority thread that we are waiting for.

Active waiting vs simple block is imho the same thing in this sense because it means there is no higher priority thread created.

Anyway - something is still not right because the affected test still failed ;)

sebhrusen: gotcha, so here by avoiding managed block if current priority is higher, we're just "forcing" the lower priority thread to complete first.

> Active waiting vs simple block is imho the same thing in this sense

agree then.

michalkurka: Yes, that is my thinking - we are giving up resources to the lower priority thread since the higher priority thread is in WAITING state.

Pscheidl (approved):
This is nicely explained in the code itself.

PR #5095, opened 2020-11-03
Split nightly jobs into 2 groups

michalkurka:
One super-group that runs everything (old nightly) - once a night
One smaller nightly group that runs selected jobs multiple times at night

The goal of re-running multiple times is to find rare bugs

### DIFF
--- a/scripts/jenkins/groovy/defineTestStages.groovy
+++ b/scripts/jenkins/groovy/defineTestStages.groovy
@@ -9,6 +9,7 @@ def call(final pipelineContext) {
   def MODE_KERBEROS_CODE = 6
   def MODE_HADOOP_MULTINODE_CODE = 7
   def MODE_MASTER_CODE = 10
+  def MODE_NIGHTLY_REPEATED_CODE = 15
   def MODE_NIGHTLY_CODE = 20
   def MODES = [
     [name: 'MODE_PR', code: MODE_PR_CODE],
@@ -20,6 +21,7 @@ def call(final pipelineContext) {
     [name: 'MODE_SINGLE_TEST', code: MODE_SINGLE_TEST_CODE],
     [name: 'MODE_BENCHMARK', code: MODE_BENCHMARK_CODE],
     [name: 'MODE_MASTER', code: MODE_MASTER_CODE],
+    [name: 'MODE_NIGHTLY_REPEATED', code: MODE_NIGHTLY_REPEATED_CODE],
     [name: 'MODE_NIGHTLY', code: MODE_NIGHTLY_CODE]
   ]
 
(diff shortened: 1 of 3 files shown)


scripts/jenkins/groovy/defineTestStages.groovy
```diff
@@ -330,68 +315,99 @@ def call(final pipelineContext) {
...
     [
-      stageName: 'Py2.7 Small', target: 'test-pyunit-small', pythonVersion: '2.7',
-      timeoutValue: 90, component: pipelineContext.getBuildConfig().COMPONENT_PY
+      stageName: 'R3.3 AutoML', target: 'test-r-automl', rVersion: '3.3.3',
+      timeoutValue: 125, component: pipelineContext.getBuildConfig().COMPONENT_R
     ],
     [
-      stageName: 'Py3.6 Small', target: 'test-pyunit-small', pythonVersion: '3.6',
-      timeoutValue: 90, component: pipelineContext.getBuildConfig().COMPONENT_PY
+      stageName: 'Kubernetes', target: 'test-h2o-k8s', timeoutValue: 20,
```
Pscheidl, line 350: Kubernetes is fast, yet stable and deterministic. Good it's here.

Pscheidl (approved):

scripts/jenkins/jenkinsfiles/Jenkinsfile-Nightly
```diff
@@ -39,7 +39,7 @@ try {
             if (pipelineContext == null) {
               def initPipelineContext = load('h2o-3/scripts/jenkins/groovy/initPipelineContext.groovy')
               pipelineContext = initPipelineContext(scmEnv, 'MODE_NIGHTLY', true)
-              // Execute nightly for master at 22:XX, 0:XX, 2:XX, 4:XX and 6:XX
+              // Execute nightly for master at 22:XX
```
honzasterba, line 42: only change to comment, not the actual schedule

michalkurka: muhaha :) thanks @honzasterba

honzasterba (approved):

scripts/jenkins/groovy/defineTestStages.groovy
```diff
@@ -330,68 +315,99 @@ def call(final pipelineContext) {
...
     ],
     [
-      stageName: 'Java 10 JUnit', target: 'test-junit-10-jenkins', pythonVersion: '2.7', javaVersion: 10,
+      stageName: 'Java 14 JUnit', target: 'test-junit-14-jenkins', pythonVersion: '2.7', javaVersion: 14,
       timeoutValue: 180, component: pipelineContext.getBuildConfig().COMPONENT_JAVA, additionalTestPackages: [pipelineContext.getBuildConfig().COMPONENT_PY]
     ],
     [
-      stageName: 'Java 11 JUnit', target: 'test-junit-11-jenkins', pythonVersion: '2.7', javaVersion: 11,
-      timeoutValue: 180, component: pipelineContext.getBuildConfig().COMPONENT_JAVA, additionalTestPackages: [pipelineContext.getBuildConfig().COMPONENT_PY]
+      stageName: 'Py3.6 Single Node', target: 'test-pyunit-single-node', pythonVersion: '3.6',
```
sebhrusen, line 322: Py 3.6 really?
Getting lost with which Py version is tested where, but I thought we wanted to have:
- Py2.7 (legacy)
- Py3.5 (min  Py3 support)
- Py3.8 (max Py3 support: note that 3.9 is not official but the images for 3.9 are not created afaik).

michalkurka: This is unchanged, the only change is that I split the big job into 2.

We can modify the Python version in another PR to keep it separated.

sebhrusen: ok, I don't see Py 3.5 tested anywhere so I guess that 3.6 is min Py version support now. Did it change recently?

sebhrusen: looks like this minimal version support was changed by @honza last month for xgboost wheel.
Just want us to be aware that we can't claim official support for 3.5 anymore if we don't run any test.

> We can modify the Python version in another PR to keep it separated.

agree, just noticed this because of the highlighting

michalkurka: @sebhrusen I think we should have the Python version testing strategy written somewhere - in simple terms so that we can cross-check that the build system is actually doing that.

Also instead of using "pythonVersion: '3.6'" we could use "pythonVersion: PY_MOST_COMMON" / PY_MINIMAL_3 to be more expressive and have it all at one place.

sebhrusen (approved):

scripts/jenkins/groovy/defineTestStages.groovy
sebhrusen: yes, good suggestion, same with R.

michalkurka: It is complicated right now - at least what we want to accomplish should be written so that anyone can understand and don't change anything that is intentional.

PR #5098, opened 2020-11-04
PUBDEV-7876: NPE when starting grid with SequentialWalker in AutoML exploitation phase

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-7876

### DIFF
--- a/h2o-algos/src/test/java/hex/grid/SequentialWalkerTest.java
+++ b/h2o-algos/src/test/java/hex/grid/SequentialWalkerTest.java
@@ -1,0 +1,129 @@
+package hex.grid;
+
+import hex.ScoreKeeper;
+import hex.ScoreKeeper.StoppingMetric;
+import hex.grid.HyperSpaceSearchCriteria.StoppingCriteria;
+import hex.tree.gbm.GBMModel;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import water.Job;
+import water.Key;
+import water.Scope;
+import water.fvec.Frame;
+import water.runner.CloudSize;
+import water.runner.H2ORunner;
+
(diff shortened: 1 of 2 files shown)


sebhrusen:
@km2018 FYI

wendycwong (approved):
Looks great1

h2o-algos/src/test/java/hex/grid/SequentialWalkerTest.java
```diff
@@ -0,0 +1,129 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static water.TestUtil.parse_test_file;
+
+@RunWith(H2ORunner.class)
+@CloudSize(1)
+public class SequentialWalkerTest {
+    
+    @Test
+    public void test_SequentialWalker() {
```
tomasfryda, line 29: I noticed that in tests we have both camel case and  mixed snake-case with camel-case.
Is there any reason to have those two conventions? Any specific use-cases when to pick the former or the latter?

tomasfryda (approved):
Looks good to me. Thank you!

h2o-algos/src/test/java/hex/grid/SequentialWalkerTest.java
sebhrusen: personally for test names, I use snake case for improved readability, to avoid
`testThisFeatureWhenWeEnableThisAndThat`.
I personally prefer a billion times `test_this_feature_when_we_enable_this_and_that` that looks like a proper sentence...

Also the main use for test names is for test reports, where readability is the most important. 
To be clear, we don't have strong naming conventions for tests: imo standard Java camel case is just not relevant here.

PR #5099, opened 2020-11-05
Rulefit: do not remove bad columns

koniecsveta:
This fixes the bug on rulefit during benchmark test case with springleaf dataset.

Each tree node remembers column id of the feature it is splitting on.
When constructing condition which is representing a tree node, this column id is used as condition's feature id.
When creating GLM's training frame this feature id is used during condition evaluation on the training set rows to map the result to correct place in chunk.
When in between of this we drop some bad columns, mapping between training frame and Condition#featureIndex/Node#colId gets broken and I have IndexOutOfBoundsException.

### DIFF
--- a/h2o-algos/src/main/java/hex/rulefit/RuleFit.java
+++ b/h2o-algos/src/main/java/hex/rulefit/RuleFit.java
@@ -90,6 +90,11 @@ public class RuleFit extends ModelBuilder<RuleFitModel, RuleFitModel.RuleFitPara
         // if (expensive && error_count() == 0) checkMemoryFootPrint();
     }
 
+    @Override
+    protected void ignoreBadColumns(int npredictors, boolean expensive) {
+        // Do not remove bad columns - we need them for correct GLM train frame creation
+    }
+
     private void initTreeParameters() {
         if (_parms._algorithm == RuleFitModel.Algorithm.GBM) {
             treeParameters = new GBMModel.GBMParameters();
--- a/h2o-algos/src/main/java/hex/rulefit/RuleFit.java
+++ b/h2o-algos/src/main/java/hex/rulefit/RuleFit.java
@@ -85,15 +85,12 @@ public class RuleFit extends ModelBuilder<RuleFitModel, RuleFitModel.RuleFitPara
 
             initTreeParameters();
             initGLMParameters();
+            ignoreBadColumns(separateFeatureVecs(), true);
         }
         //   if (_train == null) return;
         // if (expensive && error_count() == 0) checkMemoryFootPrint();
     }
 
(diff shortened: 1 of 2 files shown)


h2o-algos/src/main/java/hex/rulefit/RuleFit.java
```diff
@@ -90,6 +90,11 @@ public void init(boolean expensive) {
         // if (expensive && error_count() == 0) checkMemoryFootPrint();
     }
 
+    @Override
+    protected void ignoreBadColumns(int npredictors, boolean expensive) {
+        // Do not remove bad columns - we need them for correct GLM train frame creation
```
honzasterba, line 95: this does not seem right, I think you might want to remove bad/constant cols at the beginning of RuleFit

honzasterba (commented):
also adding a test case for this edge-case would be a good idea

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
michalkurka: I think the fact that this fixes an issue might be a sign of something bad happening

@koniecsveta how are frames passed to glm and drf/gbm? do you use trainModelNested?

koniecsveta: no no I don't use trainModelNested and removing bad cols at the beginning didn't help

koniecsveta: like this, it still goes into ModelBuilder#ignoreBadCols() on glm's and drf/gbm's model builder level, by this I just avoid going there on rulefit's model builder level

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
```diff
@@ -181,14 +183,17 @@ public void computeImpl() {
             try {
                 // linearTrain = frame to be used as _train for GLM in 2., will be filled in 1.
                 Frame linearTrain = new Frame(Key.make("paths_frame" + _result));
-                
+                // store train frame without bad columns to pass it to tree model builders
+                Frame trainRebalanced = new Frame(_train);
```
michalkurka, line 187: I don't think "rebalanced" is a good name - it implies the dataset was rebalanced which is not always the case (rebalancing means redistributing the chunks)

it is (from RuleFit's point of view) - "adapted frame"

koniecsveta: 5bd6d19

honzasterba (approved):

michalkurka (approved):

PR #5100, opened 2020-11-05
PUBDEV-7822: Fill in Variable Importance user guide page with info about non-tree-based models

hannah-tillman:
For https://h2oai.atlassian.net/browse/PUBDEV-7822

So, I shifted the "Tree-Based Algorithms" section down to the bottom of the page so it could be next to the added "Non-Tree-Based Algorithms" (also, if you think of a better name for this, let me know). Please let me know if you would like anything updated!

### DIFF
--- a/h2o-docs/src/product/data-science/rulefit.rst
+++ b/h2o-docs/src/product/data-science/rulefit.rst
@@ -110,7 +110,7 @@ Examples
 
 		import h2o
 		h2o.init()
-		from h2o.estimators import H2ORuleFitEstimators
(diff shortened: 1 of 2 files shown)


ledell (approved):
Hi @hannah-tillman, feel free to merge this for the release because it's an improvement over what we have.  However, I think it's still too focused on trees (there's a lot of general tree info not related to varimp...), so maybe let's open a new ticket or keep the ticket open to fill it in more after the holidays!   Thank you!

PR #5101, opened 2020-11-05
A simple workaround for a strange segfault in NewChunk#compress

michalkurka:
This introduces an alternative way of building a chunk byte array when compressing a NewChunk.

This PR is motivated by https://bugs.openjdk.java.net/browse/JDK-8216038

We see segmentation faults when compressing a NewChunk in very specific scenarios. It was reported mainly in AutoML benchmarks, however it is not specific to it. The issue itself is very rare and as far as we know it was not reported anywhere and experienced by a 3rd party.

The issue is localized to the `bufX` method which uses `Unsafe` similar to the other `buf?` methods. The implementation of the function was checked several times and possible edge cases like:
- integer overflow
- concurrent modification
were ruled out.

We have not seen it reproduced on other than Java 11. Working theory is that it is specific to ubuntu builds of openJDK. I was not able to reproduce on Oracle builds (@sebhrusen might have seen it on oracle as well).

@sebhrusen My suggestion:
Activate the alternative code path that avoids Unsafe and use it in AutoML benchmarks (where the issue reproduces). If we don't see the issue happing anymore we can basically conclude that bug in Java is to blame as rest of the code remains unchanged and semantically it is doing the same thing.

If we do see an issue - it should this time be in the Java code and we should see IndexArrayOutOfBounds exception proving that the code is suffering either from a bad array sizing or another issue. This would then help us design a proper solution while keeping Unsafe

### DIFF
--- a/h2o-core/src/main/java/water/fvec/NewChunk.java
+++ b/h2o-core/src/main/java/water/fvec/NewChunk.java
@@ -9,6 +9,8 @@ import water.util.PrettyPrint;
 import water.util.StringUtils;
 import water.util.UnsafeUtils;
 
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.util.Arrays;
 import java.util.BitSet;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


sebhrusen (approved):
Thanks a lot! let's merge this, we'll experiment with it.

michalkurka:
thank you @sebhrusen - should I also add to rel-zermelo or is master okay for experimentation on your side?

sebhrusen:
@michalkurka sorry didn't pay attention to the branch: zermelo is better

michalkurka:
ok, np - moved to rel-zermelo - thanks for you help

PR #5102, opened 2020-11-05
PUBDEV-7774 Implement XGBoost Feature Interaction Constraint

maurever:
Add a new parameter `interaction_constraints` to allow users use 'interaction_constraint' parameter in XGBoost. 

In the future, we would also like to add a possibility to define denied interactions. 

JIRA: https://h2oai.atlassian.net/jira/software/c/projects/PUBDEV/issues/PUBDEV-7774

### DIFF
--- a/h2o-algos/src/main/java/hex/DataInfo.java
+++ b/h2o-algos/src/main/java/hex/DataInfo.java
@@ -836,9 +836,8 @@ public class DataInfo extends Keyed<DataInfo> {
     }
     // now loop over the numerical columns, collecting up any expanded InteractionVec names
     if( _interactions == null ) {
-      final int nums = n-k;
-      int index = 0;
-      for(int i = k; i<nums; i++) {
+      int index = _cats;
+      for(int i = k; i < n; i++) {
         res[i] = index++;
       }
     } else {
--- a/h2o-algos/src/main/java/hex/DataInfo.java
+++ b/h2o-algos/src/main/java/hex/DataInfo.java
@@ -21,7 +21,6 @@ public class DataInfo extends Keyed<DataInfo> {
   public int [] _activeCols;
   public Frame _adaptedFrame;  // the modified DataInfo frame (columns sorted by largest categorical -> least then all numerical columns)
   public int _responses;   // number of responses
-  public int _outpus; // number of outputs
 
   public Vec setWeights(String name, Vec vec) {
     if(_weights)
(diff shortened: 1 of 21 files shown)


maurever:
The PR is ready for review @michalkurka and @sebhrusen. Thank you!

h2o-docs/src/product/data-science/algo-params/interaction_constraints.rst
```diff
@@ -0,0 +1,56 @@
...
+		h2o.init()
+
+		# import the prostate dataset:
+		prostate = h2o.importFile("http://s3.amazonaws.com/h2o-public-test-data/smalldata/prostate/prostate.csv.zip")
+
+		# convert the CAPSULE column to a factor
+		prostate$CAPSULE <- as.factor(prostate$CAPSULE)
+		response <- "CAPSULE"
+
+		# train a model using the monotone_constraints option
```
michalkurka, line 33: not monotone

h2o-docs/src/product/data-science/algo-params/interaction_constraints.rst
```diff
@@ -0,0 +1,56 @@
...
+
+		# import the prostate dataset:
+		prostate = h2o.import_file("http://s3.amazonaws.com/h2o-public-test-data/smalldata/prostate/prostate.csv.zip")
+
+		# convert the CAPSULE column to a factor
+		prostate["CAPSULE"] = prostate["CAPSULE"].asfactor()
+		response = "CAPSULE"
+		seed = 1234
+		
+		# train a model using the monotone_constraints option
```
michalkurka, line 54: same

michalkurka (approved):

h2o-docs/src/product/data-science/algo-params/interaction_constraints.rst
maurever: Fixed, thanks!

h2o-docs/src/product/data-science/algo-params/interaction_constraints.rst
maurever: Fixed, thanks!

maurever:
I added an encoding check. Currently, only `OneHotInternal` encoding is supported. More encodings need more time to implement.

h2o-algos/src/main/java/hex/DataInfo.java
```diff
@@ -808,6 +809,57 @@ public final int nextNumericIdx(int currentColIdx) {
     return res;
   }
 
+  public final int[] coefOriginalColumnIndices() {
```
michalkurka, line 812: This look like a copy&paste&modify of something, is it?

maurever: Yes, it is modified `coefNames()` method. However, I checked the meaning of every lines to be sure, it makes sense. But there probably could be one method for precalculation of both _coefNames and _coefOriginalColumnIndices...

michalkurka: > But there probably could be one method for precalculation of both _coefNames and _coefOriginalColumnIndices...

I would rather keep it on the side, maybe if we moved it out of DataInfo it could be significantly simplified? A lot of logic handled InteractionWrappedVecs which don't make sense in case of tree models.

PR #5104, opened 2020-11-06
PUBDEV-7882: Python docstring examples about cross-validation are not using cross-validation

hannah-tillman:
For: https://h2oai.atlassian.net/browse/PUBDEV-7882


ledell (approved):
Thanks!

PR #5105, opened 2020-11-06
PUBDEV-7838: Add 'without quotes' option for header to exportFile

michalkurka:
https://h2oai.atlassian.net/browse/PUBDEV-7838

### DIFF
--- a/h2o-core/src/main/java/water/api/FramesHandler.java
+++ b/h2o-core/src/main/java/water/api/FramesHandler.java
@@ -243,7 +243,10 @@ public class FramesHandler<I extends FramesHandler.Frames, S extends SchemaV3<I,
   public FramesV3 export(int version, FramesV3 s) {
     Frame fr = getFromDKV("key", s.frame_id.key());
     Log.info("ExportFiles processing (" + s.path + ")");
-    Frame.CSVStreamParams csvParms = new Frame.CSVStreamParams().setSeparator(s.separator);
+    Frame.CSVStreamParams csvParms = new Frame.CSVStreamParams()
+            .setSeparator(s.separator)
+            .setHeaders(s.header)
+            .setQuoteColumnNames(s.quote_header);
     s.job = new JobV3(Frame.export(fr, s.path, s.frame_id.key().toString(), s.force, s.num_parts, s.compression, csvParms));
     return s;
   }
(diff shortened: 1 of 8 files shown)


Pscheidl (commented):
Gentle warning, the [RFC-4180](https://tools.ietf.org/html/rfc4180) says:
```
   5.  Each field may or may not be enclosed in double quotes (however
       some programs, such as Microsoft Excel, do not use double quotes
       at all).  If fields are not enclosed with double quotes, then
       double quotes may not appear inside the fields.  For example:

       "aaa","bbb","ccc" CRLF
       zzz,yyy,xxx
```

This implies that:
1. It is ok to not quote the header and quote the body.
1. No quotes should be present in the header and it is user's responsibility to make sure this is true after this change.

Are we ok with this ? Otherwise the CSV might become unparseable by some parsers if there is an unescaped quote in the header.

michalkurka:
> 1. It is ok to not quote the header and quote the body.
> 2. No quotes should be present in the header and it is user's responsibility to make sure this is true after this change.
> 
> Are we ok with this ? Otherwise the CSV might become unparseable by some parsers if there is an unescaped quote in the header.

It is definitely possible to make a column name that contains a quote, it is hard but doable :) So this concern is indeed valid.

The existing export doesn't handle quote escaping in column names in any way. I think that is not an use case it was thought of. @Pscheidl Please correct my if I am wrong. In this respect I am preserving status-quo.

I would say it is okay to make it users responsibility at this point.

Pscheidl:
I agree. We might change this behavior if the writer is replaced any time in future. That would require a stateful automaton to inspect the csv stream and throw an error if there is a quote. 

I can also find opposite argument - some parsers might actually be able to parse single quotes here and there if set properly. 

So leaving that to user's responsibility seems justified.

Pscheidl (approved):

valenad1:
What if user define _parms._escapeCharacter = " (quote)? From the user perspective I would be curious if it apply also for column names. Maybe I would expect it does. We already have some logic in Frame.java:1626 for escaping.

Pscheidl:
Wild guess: Most of the parsers out there will parse escaped quotes without enclosing quotes as an empty column name. Including ours. This is also what the RFC says. It should be kept in mind that many CSV parsers go beyond the RFC to be as compatible as possible and have many settings - e.g. our own datatable csv parser.

**Example**
There are two columns, one without any quotes (let's simplify to double quotes, could also be single quotes or even tabs in many cases - those truly do exist). `First` and `Secon"d`. 

The **proper** result in the resulting CSV according to the RFC would look this:

```
"First", "Secon""d"
```

However, if enclosing quotes are disabled, then the following format is illegal according to the RFC:

```
First, Secon""d
```

It doesn't matter if the escaping is done or no, it will most likely result in a wrongly parsed CSV, even if the quote is not escaped - that's a separate case handled by various parsers differently.

valenad1 (approved):

michalkurka:
@Pscheidl @valenad1 thanks for the discussion, I think this will be useful in the future when we need to address the issue (both on input and output).

PR #5106, opened 2020-11-06
PUBDEV-7913: Fix as.data.frame.H2OFrame

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7913

It looks like `as.data.frame` converts date incorrectly when converting from h2o frame.

`POSIXct` assumes seconds and H2O provides milliseconds.

This is related to fixing explainability module's handling of date time.
(https://github.com/h2oai/h2o-3/pull/5107)


ledell (approved):
Glad that you noticed this, thanks for fixing it!

sebhrusen (approved):
LG

PR #5107, opened 2020-11-06
PUBDEV-7913: Fix  autoexplain's PDP date handling

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7913
(also fixes https://h2oai.atlassian.net/browse/PUBDEV-8375)

AutoExplain's PDP doesn't work properly on date columns. Both Python and R versions are broken for a different reason.

Python: missing conversion to date
R: incorrect conversion to date

R fix is done in two parts:
(1) ensure that `H2OFrame` subset retains its types (needed to have `POSIXct` which would otherwise turn into numeric which is its machine representation; the rest of the types (numeric, factor) don't seem to need the type information to work properly)
(2) `ggplot2` related fixes

(1) required to modify some R tests to loosen the equality strictness, i.e., not compare attributes (type info is in attributes)

### DIFF
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -199,7 +199,7 @@ class NumpyFrame:
                     levels = set(row[col] for row in df)
                     self._factors[self._columns[col]] = list(levels)
 
-            self._data = np.empty((len(df), len(self._columns)))
+            self._data = np.empty((len(df), len(self._columns)), dtype=np.float64)
             df = [self._columns] + df
         elif isinstance(h2o_frame, h2o.H2OFrame):
             _is_factor = np.array(h2o_frame.isfactor(), dtype=np.bool) | np.array(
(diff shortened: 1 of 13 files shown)


sebhrusen:
@tomasfryda is that PR still relevant? Sorry again for not reviewing this earlier.
Can you please rebase and change the target branch? There have been several changes in `explainability` code since then.

tomasfryda:
@sebhrusen Yes, this is still relevant. I rebased and quickly tested it - it seems to work properly (tested on just one dataset with date time)  but before merging I will run more exhaustive testing (requires looking at the plots and trying different versions of the `matplotlib`/`ggplot2` etc).

ledell (approved):
@tomasfryda The changes look good!  

The R tests are failing, but the test, `runit_PUBDEV_8239.R` is a parquet test, so I don't think it's related to this PR.

tomasfryda:
Thanks @ledell for mentioning the test. The failing test was related - subsetting an h2o frame used to (before this PR) lose the `types` attribute, it doesn't really matter much for anything else than datetime - datetime is represented as numeric and without the `types` attribute we have no way of finding out if the column is numeric or datetime.

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1155,6 +1187,8 @@ def pd_multi_plot(
                                    row_index=row_index, targets=target,
                                    nbins=20 if not is_factor else 1 + frame[column].nlevels()[0])[0])
             encoded_col = tmp.columns[0]
+            if frame.type(column) == "time":
+                tmp[encoded_col] /= 1000 * 3600 * 24  # convert to fractions of days for matplotlib
```
sebhrusen, line 1191: idem

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1287,6 +1321,8 @@ def ice_plot(
                 )[0]
             )
             encoded_col = tmp.columns[0]
+            if frame.type(column) == "time":
+                tmp[encoded_col] /= 1000 * 3600 * 24  # convert to fractions of days for matplotlib
```
sebhrusen, line 1325: idem

sebhrusen (approved):
no major concerns, thanks @tomasfryda

ledell:
@tomasfryda Once the conflicts are fixed, this is ready! :-)

tomasfryda:
Due to the recent changes in PDP and ICE plot in explain module this will require a bit more work (and more importantly testing) so this will not be ready for the 3.36.1.1. release.

koniecsveta (approved):
Thanks!

PR #5108, opened 2020-11-06
PUBDEV-7875 - replace System.out with Log

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-7875

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -24,6 +24,7 @@ import hex.util.LinearAlgebraUtils;
 import hex.util.LinearAlgebraUtils.BMulTask;
 import hex.util.LinearAlgebraUtils.FindMaxIndex;
 import jsr166y.CountedCompleter;
+import org.apache.log4j.Logger;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 import water.*;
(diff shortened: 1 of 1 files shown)


michalkurka (changes requested):
Please see inline comments

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -49,6 +50,9 @@
  * Generalized linear model implementation.
  */
 public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
+
+  private static final Logger LOG = Logger.getLogger(GLM.class);
```
michalkurka, line 54: This class is already using `water.util.Log` to handle logging

We should stick with one - I think in this PR you probably don't want to redo everything. cc: @honzasterba

honzasterba (approved):
👍

h2o-algos/src/main/java/hex/glm/GLM.java
valenad1: Log4j removed and water.util.Log is used

honzasterba:
it seems you based your branch on master and are trying to merge it to zermelo

valenad1:
> 
> 
> it seems you based your branch on master and are trying to merge it to zermelo

Exactly, It is weird that rebase this branch on rel-zermelo did not work as I expected and shows those 58 commit. Now I reset hard on rel-zermelo and cherry pick those two commits.

honzasterba:
rebase will not work in this case, beacause it will take all the commits that are not common and rebase on the new branch, hence duplicating the commits, you can do `git rebase -i new-branch` to see what you are bringing over and keep only what you really want to rebase

michalkurka (approved):
Thank you for making the changes @valenad1

PR #5109, opened 2020-11-07
Add GAM synthetic generation code in Python.

wendycwong:
Karthik is starting work on automatic choosing of scale parameter.  Hence I am adding this one for him to be able to generate data for himself.

### DIFF
--- a/h2o-py/tests/testdir_algos/gam/pyunit_generate_synthetic_GAM_data.py
+++ b/h2o-py/tests/testdir_algos/gam/pyunit_generate_synthetic_GAM_data.py
@@ -1,0 +1,121 @@
+import sys
+sys.path.insert(1,"../../../")
+import h2o
+from tests import pyunit_utils
+from h2o.estimators.gam import H2OGeneralizedAdditiveEstimator as gam
+from random import random
+import numpy as np
+
+# This test will generate synthetic GAM dataset.  If given to a GAM model, it should be able to perform well with 
+# this dataset since the assumptions associated with GAM are used to generate the dataset.  However, pay attention
+# to the data types and you may have to cast enum columns to factors manually since during the save, column types
+# information may be lost.
+#
+# Apart from saving the dataset using h2o.download_csv, remember to save the column  types as
+# np.save('my_file.npy', types_dict) 
+#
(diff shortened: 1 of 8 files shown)


h2o-py/tests/testdir_algos/gam/pyunit_generate_synthetic_GAM_data.py
```diff
@@ -0,0 +1,95 @@
...
+    
+    assert numGamCols <= ncol*realFrac, "Number of real columns {0} should exceed the number of gam columns " \
+                                       "{1}".format(ncol*realFrac, numGamCols) # gam can be only real columns
+    gamDataSet = generate_dataset(family, nrow, ncol, realFrac, intFrac, enumFrac, missing_fraction, factorRange, 
+                                  numericRange, targetFactor, numGamCols)
+    #h2o.download_csv(gamDataSet, "/Users/.../dataset.csv") # save dataset
+    assert gamDataSet.nrow == nrow, "Dataset number of row: {0}, expected number of row: {1}".format(gamDataSet.nrow, 
+                                                                                                     nrow)
+    assert gamDataSet.ncol == (1+ncol), "Dataset number of row: {0}, expected number of row: " \
+                                                          "{1}".format(gamDataSet.ncol, (1+ncol))
```
michalkurka, line 30: since the goal is to create a good dataset for GAM, we should perhaps run GAM on it an check we get a very good accuracy

michalkurka (approved):

h2o-py/tests/testdir_algos/gam/pyunit_generate_synthetic_GAM_data.py
wendycwong: Will do before I merge

wendycwong:
Here is a dataset for Gaussian.  I have attached the training and validation metrics here:
image

image

wendycwong:
The following showcase the training for binomial:
image

The validation metrics:
image

wendycwong:
Here is a run with multinomial:
image


The validation metric:
image

I did not use a big dataset because I don't want to wait forever for the result.  Hence, the performance was lower than binomial or regression.

michalkurka (approved):

PR #5110, opened 2020-11-09
PUBDEV-7884 - Java 15 support

Pscheidl:
https://h2oai.atlassian.net/browse/PUBDEV-7884

At first, tests will fail as images need to be built beforehand :wink: 

Sample output if the lower boundary is changed not to cover version `8` (changed to 10):

```
Only Java versions 10,11,12,13,14,15 are supported, system version is 1.8.0_271
```


Pscheidl:
Attempting to build the images:

http://mr-0xc1:8080/view/H2O-3/job/docker-images/job/h2o-3-dev-images/72/

h2o-core/src/main/java/water/Java.java
```diff
@@ -0,0 +1,61 @@
+package water;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import static water.util.JavaVersionUtils.JAVA_VERSION;
+
+public class Java {
```
honzasterba, line 9: why not make this part of unify with JavaVersionUtils?

honzasterba: would love to see a unit test

docker/Jenkinsfile-build-docker
```diff
@@ -19,7 +19,7 @@ properties([
 
 IMAGE_NAME_PREFIX = 'harbor.h2o.ai/opsh2oai/h2o-3'
 
-JDK_VERSIONS = ['8', '10', '11', '12', '13', '14']
+JDK_VERSIONS = ['8', '10', '11', '12', '13', '14', 15]
```
honzasterba, line 22: how many images this added? probably a few, we really need to take a hard look at that

docker/scripts/java-15-vars.sh
```diff
@@ -0,0 +1,6 @@
+#! /bin/bash
+
+JAVA_INSTALL_VERSION_MAJOR='15'
+JAVA_INSTALL_VERSION=2
+JAVA_INSTALL_PATH=/usr/opt/java-${JAVA_INSTALL_VERSION_MAJOR}-${JAVA_INSTALL_VERSION}/
+JDK_FILE=/tmp/jdk-14.0.2_linux-x64_bin.tar.gz
```
honzasterba, line 6: wrong version

honzasterba: also need to add download command to get the zip from s3

Pscheidl: Oh snap. Thanks. will fix.

h2o-core/src/main/java/water/Java.java
Pscheidl: 1. Wil attempt to add UnitTest - could be possible with `SystemProperties` JUnit lib (we already use that).
2. Because JavaVersionUItils are part of `genmodel` and therefore can't touch: `H2O.OptArgs.SYSTEM_PROP_PREFIX + "debug.allowJavaVersions"` for example. So I chose encapsulating this to a new class in the `water` package to at least not make H2O.java any heavier.

I don't have strong opinions about this, just explaining reasons. Would that make sense ?

docker/Jenkinsfile-build-docker
Pscheidl: This is the most important discussion. Yes, I agree. Should I skip building the images and not add Java 15 to smoke tests ? I can test the checks using plain JUnit tests I believe.

Or, add 15 in this PR and in another PR let's optimize this to the oldest and newest supported only ?

docker/scripts/java-15-vars.sh
Pscheidl: Will fix after we discuss whether we want this or not.

h2o-core/src/main/java/water/Java.java
Pscheidl: Added tests: https://github.com/h2oai/h2o-3/pull/5110/commits/db2b060ad1d71572d9a70ec8e419427338d447f6

h2o-core/src/main/java/water/H2O.java
```diff
@@ -2137,11 +2136,11 @@ public static boolean checkUnsupportedJava() {
       return false;
     }
 
-    // Notes: 
-    // - make sure that the following whitelist is logically consistent with whitelist in R code - see function .h2o.check_java_version in connection.R
-    // - upgrade of the javassist library should be considered when adding support for a new java version
-    if (JAVA_VERSION.isKnown() && !isUserEnabledJavaVersion() && (JAVA_VERSION.getMajor()<8 || JAVA_VERSION.getMajor()>14)) {
-      System.err.println("Only Java 8, 9, 10, 11, 12 and 13 are supported, system version is " + System.getProperty("java.version"));
+    if (!Java.runningOnSupportedVersion()) {
```
michalkurka, line 2139: Isn't a less complex solution the better solution? One issue you can see that this list is not ordered. I would prefer simplicity as more lines of code -> more issues is usually the rule.

Pscheidl: True. But then the issue was there were numbers missing in the error message.

Pscheidl: I even have JUnit for that logic btw.

michalkurka: The other approach is to have the production code simple and add a unit test that checks the expectations.

> I even have JUnit for that logic btw.

I see the unit test but it is still returning a Set.

At some point the message shouldn't be just listing all the Javas. We would want to say "Java 8 and later (up to Java 20) is supported" - eg. instead of all the Javas, or no?

Pscheidl: IDK. That's why I "designed" the holes in the range - unsupported versions in the range. Not sure if this will ever be used, but better safe than sorry. This Java class was supposed to be a generic solution. We could also manually list the versions - as we wish. This is internal to the class and might be changed at any time. 

Having it in the Java.java fike makes this logic testable. The final test a specific version works is the smoke test. But that won't test unallowed versions 

There seems to be too many problems at a time.

- Is it ok to keep it in the class ? To me yes. 
- If yes, would a guaranteed order in the very contract of the versions method be sufficient ? To me this should be done. 
- Callers of that class may decide to display any error message they want. These concerns are clearly separated. Nothing prevents the caller from verifying the versions actually form one or multiple intervals and print those in any desired way ?
- Should we start showing the interval range instead of full list now ? I think why not. 

Hope these questions are more specific to what we're talking about here.

honzasterba (approved):

docker/scripts/java-15-vars.sh
```diff
@@ -0,0 +1,6 @@
+#! /bin/bash
+
+JAVA_INSTALL_VERSION_MAJOR='15'
+JAVA_INSTALL_VERSION=2
+JAVA_INSTALL_PATH=/usr/opt/java-${JAVA_INSTALL_VERSION_MAJOR}-${JAVA_INSTALL_VERSION}/
+JDK_FILE=/tmp/jdk-15.0.2_linux-x64_bin.tar.gz
```
michalkurka, line 6: This version doesn't exist, does it?

michalkurka (changes requested):
This PR doesn't add S3 download here: https://github.com/h2oai/h2o-3/pull/5110/files#diff-606c84166941026c89de1b31075a970ccfe062158c0c77efa300b1be74470c68L59

michalkurka:
@Pscheidl there is some good stuff here - I think in order to move forward we should separate the (java) code changes that refactor the error reporting into its own PR - that we can merge regardless of J15 change.

Ideally we can have 3 PRs:
1. refactoring of java backend
2. new images for Java 15 added
3. test pipelines modified to include Java 15
(4. fixes)

Pscheidl:
That actually adds much more complexity to the whole thing, as we first have to merge the backend refactor, then base the remaining PRs on it (at least the one to add the version 15 to the list). But will do.

docker/scripts/java-15-vars.sh
Pscheidl: That's the pattern there for quite some time. Not sure what it means. Do we have to ask someone e.g. @abal5 to install that for us ? Is this mounted ?

Pscheidl:
1. refactoring of java backend - https://github.com/h2oai/h2o-3/pull/5121
2. new images for Java 15 added - **THIS PR** https://github.com/h2oai/h2o-3/pull/5110
3.test pipelines modified to include Java 15 - https://github.com/h2oai/h2o-3/pull/5122

docker/scripts/java-15-vars.sh
Pscheidl: It doesn't, adding.

Pscheidl: Added - https://s3.console.aws.amazon.com/s3/buckets/artifacts.h2o.ai?region=us-east-1&prefix=releases/oracle/jdk-15/
Commit: https://github.com/h2oai/h2o-3/pull/5110/commits/350a230a7a6cfe2366d34d761c4f9bc8acb059fb
Wanted to name it in a better way, but the shell scripts kind of get in the way, so  I kept the current naming pattern.

Pscheidl:
Turned to draft, as there is the image consolidation to be done.

PR #5111, opened 2020-11-09
[PUBDEV-7887] ability to list all models in the cluster

honzasterba:
- some refactors
- getting actual models objects can be expensive, returning only list of IDs seems like the best choice

### DIFF
--- a/h2o-py/h2o/h2o.py
+++ b/h2o-py/h2o/h2o.py
@@ -967,6 +967,29 @@ def deep_copy(data, xid):
     return duplicate
 
 
+def models():
+    """
+    Retrieve the IDs all the Models.
+
+    :returns: Handles of all the models present in the cluster
+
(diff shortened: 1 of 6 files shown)


koniecsveta (approved):

h2o-py/h2o/h2o.py
```diff
@@ -967,6 +967,29 @@ def deep_copy(data, xid):
     return duplicate
 
 
+def models():
+    """
+    Retrieve the IDs all the Models.
+
+    :returns: Handles of all the models present in the cluster
```
michalkurka, line 974: Flow also shows algo. I think the simple form of just ids is more useful though. Maybe in the future, we could add an option detailed=True and show more than that but for now, this makes the best sense to me.

michalkurka (approved):

PR #5112, opened 2020-11-09
Fix svm drop columns test.

valenad1:
According to call with @wendycwong . Test should handle that skipped_columns cannot be pass for the svm files because skipped_columns are not supported for *.svm.

I simplified data consumption and made the test more transparent.


honzasterba (approved):

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
```diff
@@ -0,0 +1,55 @@
...
+        os.mkdir(tmpdir)
+    savefilenamewithpath = os.path.join(tmpdir, 'out.svm')
+    pyunit_utils.write_H2OFrame_2_SVMLight(savefilenamewithpath, f1)  # write h2o frame to svm format
+
+    try:
+        print("Test upload SVM file")
+        h2o.upload_file(savefilenamewithpath, skipped_columns=[5])
+        assert False, "Test should have thrown an exception due skipped_columns parameter is present"  # should have failed here
+    except H2OResponseError:
+        print("Test OK, finished with H2OResponseError")
```
michalkurka, line 37: Can we narrow down the check? To make sure we are getting the right error?

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
```diff
@@ -0,0 +1,55 @@
...
+    seed = 12345
+
+    f1 = h2o.create_frame(rows=nrow, cols=ncol, real_fraction=0.5, integer_fraction=0.5, missing_fraction=0.2,
+                          has_response=False, seed=seed)
+    f2 = h2o.create_frame(rows=nrow, cols=1, real_fraction=1, integer_fraction=0, missing_fraction=0,
+                          has_response=False, seed=seed)
+    f1.set_name(0, "target")
+    f1 = f2.cbind(f1)
+
+    tmpdir = os.path.normpath(os.path.join(os.path.dirname(os.path.realpath('__file__')), "..", "results"))
```
michalkurka, line 26: it should probably use 

    results_dir = pyunit_utils.locate("results")

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def test_parser_svmlight_column_skip_not_supported():
+    print("Test that functions calling fail if skipped_columns is passed with svm file.")
+    # generate a big frame with all datatypes and save it to svmlight
+    nrow = 10
+    ncol = 10
+    seed = 12345
+
+    f1 = h2o.create_frame(rows=nrow, cols=ncol, real_fraction=0.5, integer_fraction=0.5, missing_fraction=0.2,
```
michalkurka, line 19: I think this is also possible to simplify - we might not even need a response for this test but if we do create_frame has the option to generate a binary response.

I think for this type of test we just need one call to create_frame.

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
```diff
@@ -0,0 +1,55 @@
...
+
+    try:
+        print("Test upload SVM file")
+        h2o.upload_file(savefilenamewithpath, skipped_columns=[5])
+        assert False, "Test should have thrown an exception due skipped_columns parameter is present"  # should have failed here
+    except H2OResponseError:
+        print("Test OK, finished with H2OResponseError")
+        pass
+    except Exception as e:
+        assert False, "Test finish with unexpected exception." + str(e)
```
michalkurka, line 40: Do we need to catch the exception at all? Is this adding anything?

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_large.py
```diff
@@ -1,77 +0,0 @@
...
-        pass
-    try:
-        h2o.import_file(savefilenamewithpath, skipped_columns=skip_even)
-        assert False, "Test should have thrown an exception due to column skipping not supported for svm"  # should have failed here
-
-    except:
-        pass
-
-    # check for correct parsing only
-    checkCorrectSkips(savefilenamewithpath, f1)
```
michalkurka, line 51: Why are we not checking the correct parsing anymore?

wendycwong (approved):

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_large.py
wendycwong: @michalk  There are other tests that check for correct parsing.  I think it is okay to skip in this case.  It makes the test run faster.

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
```diff
@@ -0,0 +1,55 @@
...
+    f1.set_name(0, "target")
+    f1 = f2.cbind(f1)
+
+    tmpdir = os.path.normpath(os.path.join(os.path.dirname(os.path.realpath('__file__')), "..", "results"))
+    if not (os.path.isdir(tmpdir)):
+        os.mkdir(tmpdir)
+    savefilenamewithpath = os.path.join(tmpdir, 'out.svm')
+    pyunit_utils.write_H2OFrame_2_SVMLight(savefilenamewithpath, f1)  # write h2o frame to svm format
+
+    try:
```
wendycwong, line 32: Adam:  Can you add a print like this:

print("Expect Java exception error: skipped_columns not supported for AVRO and SVMlight") 

or something like that so that we won't scream everytime an exception is observed in H2O java logs?


Wendy

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
```diff
@@ -0,0 +1,55 @@
...
+        h2o.upload_file(savefilenamewithpath, skipped_columns=[5])
+        assert False, "Test should have thrown an exception due skipped_columns parameter is present"  # should have failed here
+    except H2OResponseError:
+        print("Test OK, finished with H2OResponseError")
+        pass
+    except Exception as e:
+        assert False, "Test finish with unexpected exception." + str(e)
+
+    try:
+        print("Test import SVM file")
```
wendycwong, line 43: Adam: Same thing here...  Can you add a print like this:

print("Expect Java exception error: skipped_columns not supported for AVRO and SVMlight")

or something like that so that we won't scream everytime an exception is observed in H2O java logs?

Wendy

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
valenad1: The reason why is not used here is most probably to create results directory if not exists. 
`results_dir = pyunit_utils.locate("results")`
give error if results directory does not exists.

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
valenad1: I remove the second call and create only one frame. It works fine but I get warning that I am not able to track and handle.

The output of the test:

```
Test upload SVM file
Test OK, finished with H2OResponseError
Test import SVM file
Test OK, finished with H2OResponseError
Closing connection _sid_bf1c at exit
../..\h2o\h2o.py:780: UserWarning: [Lwater.parser.ParseWriter$ParseErr;@48c7a39e
  warnings.warn(w)
../..\h2o\h2o.py:780: UserWarning: [Lwater.parser.ParseWriter$ParseErr;@7ee8b863
  warnings.warn(w)
H2O session _sid_bf1c closed.
```

do you have any Idea?

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
valenad1: The idea here is to be sure that it test throws only ResponseError. If it throws e. g. TypeError, ValueError then something is wrong in my opinion.

honzasterba: I dont think you need the `except Exception` branch

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
honzasterba: does this happen on jenkins as well? I think that when ran via `run.py` the results dir is always present

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
honzasterba: agree there should be some assertion on the error message, usually this is hidden in `e.args[0].exception_msg`

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_large.py
valenad1: The parsing is cancelled before the job started, there is no result to check, beside check that nothing happened.

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
valenad1: When the folder is present, everything goes well, if not, then it fails with file not found exception. I think, this workaround is because of locale run. I have no problem to replace it with suggested solution.

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
valenad1: All right, I will remove it.

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
```diff
@@ -0,0 +1,50 @@
...
+        assert "skipped_columns are not supported" in str(e.args[0].exception_msg), "Exception message is different"
+
+    try:
+        print("Test import SVM file. "
+              "Expected result is Java exception error: skipped_columns not supported for AVRO and SVMlight")
+        h2o.import_file(savefilenamewithpath, skipped_columns=[5])
+        assert False, "Test should have thrown an exception due skipped_columns parameter is present"  # should have failed here
+    except H2OResponseError as e:
+        print("Test OK, finished with H2OResponseError")
+        assert "skipped_columns are not supported" in e.args[0].exception_msg
```
honzasterba, line 44: Generally prints in tests are not good practice, usually they should be just asserts, prints are pseudo-asserts for the programmer reading the test. I know its common practice in our py/R tests, but just wanted to point out that there are better way to do things.
Also you might want to switch the order of the `print` and `assert` because it will be confusing when the test fails with printout `Test OK` and `assertion error` afterwards.

valenad1: Order changed and I will keep in mind to prefer asserts before print. Thanks

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
valenad1: I use results_dir = pyunit_utils.locate("results")

h2o-py/tests/testdir_parser/pyunit_PUBDEV_5705_drop_columns_parser_svmlight_fail.py
valenad1: @michalkurka is this warning important?

honzasterba: I think it deserves an investigation

valenad1: Fixed. The problem was setup of create_frame method missing_fraction=0.2 cause some missing value in the 0 column and svm parser report warning about it.

honzasterba (approved):

valenad1:
I will create new PR starting from branch rel-zermelo

PR #5114, opened 2020-11-10
PUBDEV-7773 fix pca_impl passed to PCA Mojo

wendycwong:
This PR fixes issues in this JIRA: https://h2oai.atlassian.net/browse/PUBDEV-7773

PCA Mojo does not need to know pca_impl and therefore it is not passed over to PCA Mojo.

I have added it to PCAMojoWriter, PCAMojoModel and PCAMojoReader.


mn-mikke (approved):
LGTM

PR #5115, opened 2020-11-10
PUBDEV-7886: Grid search with parallelism enabled includes CV models

michalkurka:


h2o-core/src/main/java/hex/ParallelModelBuilder.java
```diff
@@ -55,21 +55,25 @@ public void run(final Collection<ModelBuilder> modelBuilders) {
 
     @Override
     public void onModelSuccess(Model model) {
-      try {
-        _callback.onBuildSuccess(model, ParallelModelBuilder.this);
-      } finally {
-        _modelInProgressCounter.decrementAndGet();
+      if (! model._parms._is_cv_model) {
```
Pscheidl, line 58: Maybe this logic could be incorporated one level lower - in the ModelBuilder itself - do not report CV models. Not sure the side ffects are right - the clients should probably decide. So on the other hand, this is the perfect "sweet spot". :+1: 

Thank you @michalkurka .

I can add some JUnit tests if @sebhrusen thinks this solution is allright and that's it.

michalkurka:
I don't think this issue is related:

Hyper-parameter: ntrees, 5
failure_details: Cannot update - Lockable is not write-locked!
failure_stack_traces: java.lang.AssertionError: Cannot update - Lockable is not write-locked!
	at water.Lockable$Update.atomic(Lockable.java:196)
	at water.Lockable$Update.atomic(Lockable.java:191)
	at water.TAtomic.atomic(TAtomic.java:17)
	at water.Atomic.compute2(Atomic.java:56)
	at water.Atomic.fork(Atomic.java:39)
	at water.Atomic.invoke(Atomic.java:31)
	at water.Lockable.update(Lockable.java:186)
	at water.Lockable.update(Lockable.java:183)
	at hex.grid.GridSearch$ModelFeeder.onBuildSuccess(GridSearch.java:218)
	at hex.ParallelModelBuilder$ParallelModelBuiltListener.onModelSuccess(ParallelModelBuilder.java:60)

michalkurka:
I see the issue

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -250,8 +250,6 @@ private void attemptBuildNextModel(final ParallelModelBuilder parallelModelBuild
           reconcileMaxRuntime(grid._key, nextModelParams);
 
           parallelModelBuilder.run(Collections.singletonList(ModelBuilder.make(nextModelParams)));
-        } else {
-          parallelModelBuilder.noMoreModels();
```
sebhrusen, line 254: it's still needed to prevent the `grid.update()` on unlocked grid issue right?
We need a follow up ticket to fix the cancellation scenarios, and distinguish the following cases:
1. the grid is exhausted or early stopped: in this case we wait for the pending models and the parallel builder can be marked as complete only when all models are built or failed.
2. the grid is cancelled (explicit `job.stop` or timeout): in this case we should prevent new modelbuilders to be pushed, and running subjobs should be explicitely stopped as well. The parallel builder can also be marked as complete when the pending jobs are done (failed due to cancellation).

sebhrusen (approved):
Looks good, thanks!
added some comments for a follow-up ticket.
Also looks like we need more tests unit tests for all those edge-case scenarios: timeout, explicit cancellation, early-stopped grid.

h2o-core/src/main/java/hex/grid/GridSearch.java
Pscheidl: Agreed.

Pscheidl (approved):

michalkurka:
@sebhrusen @Pscheidl thank you for the review and the suggestions

I made a Jira to address the identified problems: https://h2oai.atlassian.net/browse/PUBDEV-7895 - will fix in the subsequent fix release

PR #5123, opened 2020-11-12
PUBDEV-7854 script to estimate memory usage for a dataset

satai:
usage: estimate_memory_usage.py [-h] --nrows NROWS --ncols NCOLS [--n-num-cols N_NUM_COLS] [--n-string-cols N_STRING_COLS] [--n-uuid-cols N_UUID_COLS] [--n-cat-cols N_CAT_COLS]
                                [--n-time-cols N_TIME_COLS]

Estimate memory needed to run H2O for a given dataset.

optional arguments:
  -h, --help            show this help message and exit
  --nrows NROWS         count of rows in the dataset (required)
  --ncols NCOLS         count of columns of all types in the dataset (required)
  --n-num-cols N_NUM_COLS
                        count of numeric collumns in the dataset
  --n-string-cols N_STRING_COLS
                        count of string collumns in the dataset
  --n-uuid-cols N_UUID_COLS
                        count of UUID collumns in the dataset
  --n-cat-cols N_CAT_COLS
                        count of cathegorical collumns in the dataset
  --n-time-cols N_TIME_COLS
                        count of time collumns in the dataset


scripts/estimate_memory_usage.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/python3
+
+import argparse
+import math
+
+base_mem_requirement = 32
```
Pscheidl, line 6: Knit: First impression: Why not 42 ? 32 of what ? Megabytes ? `base_mem_gig` would also make sense.

scripts/estimate_memory_usage.py
```diff
@@ -0,0 +1,65 @@
...
+
+known_cols = args.n_num_cols + args.n_string_cols + args.n_uuid_cols + args.n_cat_cols + args.n_time_cols
+
+if (args.ncols < known_cols):
+    print("count of all columns can not be lower then sum of counts of different types of columns")
+    print()
+    parser.print_help()
+    exit(2)
+
+safety_factor=4
```
Pscheidl, line 35: Aren't constants upper case in Python ?
Since Py 3.8, there is final type. Wouldn't bother with that. Just saying.

scripts/estimate_memory_usage.py
```diff
@@ -0,0 +1,65 @@
...
+unknown_size = 8
+unknown_requirement = unknown_cols * args.nrows * unknown_size
+
+num_size = 8
+num_requirement = args.n_num_cols * args.nrows * num_size
+
+string_size = 128
+string_requirement = string_size * args.n_num_cols * args.nrows
+
+uuid_size = 36
```
Pscheidl, line 48: Please note, UUID is parsed by the `ParseUUID.java` class.

From the JavaDoc:
```
   * Attempts to parse the provided {@link BufferedString} as
   * a UUID into hash value in two longs.
```

scripts/estimate_memory_usage.py
```diff
@@ -0,0 +1,65 @@
...
+                   help='count of time collumns in the dataset')
+
+
+args = parser.parse_args()
+
+known_cols = args.n_num_cols + args.n_string_cols + args.n_uuid_cols + args.n_cat_cols + args.n_time_cols
+
+if (args.ncols < known_cols):
+    print("count of all columns can not be lower then sum of counts of different types of columns")
+    print()
```
Pscheidl, line 31: Knit: `\n` at the end of the previous line ?

pastx (approved):
LGTM, cannot judge constants and python coding style.

nit: I don't need that for Steam, but args might be validated for positive number (negative values are accepted)

scripts/estimate_memory_usage.py
satai: 32MB is a bit more then an empty usage measured by VisualVM. I renamed to variable to make the units cleaner.

scripts/estimate_memory_usage.py
satai: Can I use py 3.8 stuff here?

Renamed to look less like Pascal ;)

scripts/estimate_memory_usage.py
satai: It's there now

scripts/estimate_memory_usage.py
satai: > LGTM, cannot judge constants and python coding style.
> 
> nit: I don't need that for Steam, but args might be validated for positive number (negative values are accepted)



> LGTM, cannot judge constants and python coding style.
> 
> nit: I don't need that for Steam, but args might be validated for positive number (negative values are accepted)

satai:
@pastx There is a nice way to do it, thanks for idea.

scripts/estimate_memory_usage.py
satai: Updated

scripts/estimate_memory_usage.py
Pscheidl: Just a note. Noone cares about Python 3.8

michalkurka:
Can we have this script tested?

I think this could be a function of the h2o package and be tested as a part of the large tests. Or it can be standalone but this absolutely needs to have tests if we are going to give this to people.

Pscheidl:
Thought: If this is part of our API, it would usually require the whole H2O client to work - is this desirable ?

Or we might have two ways of distributing this script and test it as a part of our API ?

pastx:
@michalkurka 

- it might be better to give the estimation in GBs instead of MBs. At least when consumed by Steam, not sure about other use cases

- also, I would drop the "MB" in the printed output and move it to the function name, this could be better when the output is machine-processed

satai:
@pastx As you wish. Updated to be in GB, postfix not printed

Pscheidl:
Massive test failure:

```python
Error
See /home/jenkins/slave_dir_from_mr-0xc1/workspace/h2o-3-pipeline_PR-5123/py2.7-smoke/h2o-3/h2o-py/tests/results/testdir_algos_pca_pyunit_arrestsPCA.py.out.txt
Stacktrace
  File "/home/jenkins/slave_dir_from_mr-0xc1/workspace/h2o-3-pipeline_PR-5123/py2.7-smoke/h2o-3/h2o-py/scripts/h2o-py-test-setup.py", line 186, in <module>
    h2o_test_setup(sys.argv)
  File "/home/jenkins/slave_dir_from_mr-0xc1/workspace/h2o-3-pipeline_PR-5123/py2.7-smoke/h2o-3/h2o-py/scripts/h2o-py-test-setup.py", line 139, in h2o_test_setup
    import h2o
  File "/envs/h2o_env_python2.7/lib/python2.7/site-packages/h2o/__init__.py", line 14, in <module>
    from h2o.h2o import (connect, init, api, connection,
  File "/envs/h2o_env_python2.7/lib/python2.7/site-packages/h2o/h2o.py", line 2272
    def estimate_mem_usage(ncols: int, nrows: int, num_cols: int = 0, string_cols: int = 0, cat_cols: int = 0, time_cols: int = 0, uuid_cols: int = 0) -> int:
                                ^
SyntaxError: invalid syntax
```

satai:
@Pscheidl I have removed python 3 specific features

h2o-py/h2o/h2o.py
```diff
@@ -2269,6 +2269,85 @@ def print_mojo(mojo_path, format="json", tree_index=None):
     else:
         raise H2OError("Unable to print MOJO: %s" % output)
 
+def estimate_mem_usage(ncols, nrows, num_cols = 0, string_cols = 0, cat_cols = 0, time_cols = 0, uuid_cols = 0):
+    """
+    Generates and estimate for memory usage in GB.
```
michalkurka, line 2274: I find this to be slightly confusing. We need to keep in mind that memory usage in H2O and required memory to parse a dataset are two different things. Here we are trying to estimate how to size the cluster for a particular dataset. The actual usage will be lower and there will be plenty of memory for munging and modeling building.

IMHO we need to clarify this docustring.

michalkurka: I would even rename this method to something like "estimate_cluster_mem" to make it clear it is not estimating memory usage.

satai: I have renamed it and updated the documentation.

PR #5127, opened 2020-11-12
[PUBDEV-7859] fix frame loading

honzasterba:
- need to create a VecGroup so other vec creation does not overwrite Frame's vecs


michalkurka (approved):
LGTM

Just one question: When the deserialized Vecs are read - they are not installed in DKV by autobuffer, right? I am not sure right now if that could happen

honzasterba:
autobuffer.putKey and getKey does interact with dkv, put/get does not

PR #5129, opened 2020-11-13
[PUBDEV-7859] add ability to resume grid after cluster crash

honzasterba:
https://h2oai.atlassian.net/browse/PUBDEV-7859

### DIFF
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -401,7 +401,7 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
     /** @return the validation frame instance, or null
      *  if a validation frame was not specified */
     public final Frame valid() { return _valid==null ? null : _valid.get(); }
-    
+
     public String[] getNonPredictors() {
         return Arrays.stream(new String[]{_weights_column, _offset_column, _fold_column, _response_column})
                 .filter(Objects::nonNull)
@@ -568,6 +568,29 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
         }
       }
       return usedColumns;
+    }
+
+    @SuppressWarnings("rawtypes")
+    public Set<Key<?>> getDependentKeys() {
+      Field[] fields = Weaver.getWovenFields(getClass());
+      Set<Key<?>> values = new HashSet<>();
+      for (Field f : fields) {
+        f.setAccessible(true);
+        Class<?> c = f.getType();
+        try {
+          Object value = f.get(this);
+          if (value instanceof Key) {
+            values.add((Key) value);
+          } else if (value != null && c.isArray() && c.getComponentType() == Key.class) {
+            Key[] arr = (Key[]) value;
+            for (Key k : arr)
+              if (k != null) values.add(k);
+          }
+        } catch (IllegalAccessException e) {
+          throw new RuntimeException(e);
+        }
+      }
+      return values;
     }
   }
 
--- a/h2o-core/src/main/java/hex/faulttolerance/Recoverable.java
+++ b/h2o-core/src/main/java/hex/faulttolerance/Recoverable.java
@@ -1,0 +1,26 @@
+package hex.faulttolerance;
+
+import water.Key;
+import water.Keyed;
+
+import java.util.Set;
+
+public interface Recoverable<T extends Keyed> {
+
+    /**
+     * @return key of this keyed object
+     */
+    Key<T> getKey();
+
+    /**
+     * @param location directory where this recoverable will be written into a single file
+     * @return path to where data was written
+     */
+    String exportBinary(String location);
+
+    /**
+     * @return list of all keys of objects this recoverable needs to resume operation after recovery
+     */
+    Set<Key<?>> getDependentKeys();
+    
+}
--- a/h2o-core/src/main/java/hex/faulttolerance/Recovery.java
+++ b/h2o-core/src/main/java/hex/faulttolerance/Recovery.java
@@ -1,0 +1,136 @@
+package hex.faulttolerance;
+
+import hex.Model;
+import water.*;
+import water.fvec.Frame;
+import water.fvec.persist.FramePersist;
+import water.fvec.persist.PersistUtils;
+import water.util.FileUtils;
+import water.util.IcedHashMap;
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.*;
+
+public class Recovery<T extends Keyed> {
+
+    public static final String REFERENCES_META_FILE_SUFFIX = "_references";
+
+    public enum ReferenceType {
+        FRAME, KEYED
+    }
+
+    private final String storagePath;
+    private final List<String> writtenFiles = new ArrayList<>(); 
+
+    /**
+     * @param storagePath directory to use as base for recovery snapshots
+     */
+    public Recovery(String storagePath) {
+        this.storagePath = storagePath;
+    }
+    
+    private String recoveryFile(String f) {
+        return storagePath + "/" + f;
+    }
+
+    private String recoveryFile(Key key) {
+        return recoveryFile(key.toString());
+    }
+    
+    public String referencesMetaFile(Recoverable<T> r) {
+        return recoveryFile(r.getKey().toString() + REFERENCES_META_FILE_SUFFIX);
+    }
+
+    /**
+     * Called when the training begins, so that initial state can be persisted
+     * 
+     * @param r a Recoverable to persist
+     */
+    public void onStart(final Recoverable<T> r) {
+        writtenFiles.add(r.exportBinary(storagePath));
+        exportReferences(r);
+    }
+
+    /**
+     * Called by the Recoverable to notify of new model was trained and needs to persisted
+     * 
+     * @param r a Recoverable to update
+     * @param modelKey key of the newly trained model
+     */
+    public void onModel(final Recoverable<T> r, Key<Model> modelKey) {
+        try {
+            String modelFile = recoveryFile(modelKey);
+            modelKey.get().exportBinaryModel(modelFile, true);
+            writtenFiles.add(modelFile);
+            r.exportBinary(storagePath);
+        } catch (IOException e) {
+            // this should not happen since storagePath should be writable because
+            // grid was already written to it
+            throw new RuntimeException("Failed to store model for fault tolerance.", e);
+        }
+    }
+
+    /**
+     * Called by the recoverable that the training was finished successfully. This means that
+     * recovery snapshots (persisted data) is no longer needed and can be deleted.
+     */
+    public void onDone(Recoverable<T> r) {
+        final URI storageUri = FileUtils.getURI(storagePath);
+        for (String path : writtenFiles) {
+            URI pathUri = FileUtils.getURI(path);
+            H2O.getPM().getPersistForURI(storageUri).delete(pathUri.toString());
+        }
+    }
+
+    /**
+     * Saves all of the keyed objects used by this Grid's params. Files are named by objects' keys.
+     */
+    public void exportReferences(final Recoverable<T> r) {
+        final Set<Key<?>> keys = r.getDependentKeys();
+        final IcedHashMap<String, String> referenceKeyTypeMap = new IcedHashMap<>();
+        for (Key<?> k : keys) {
+            persistObj(k.get(), referenceKeyTypeMap);
+        }
+        final URI referencesUri = FileUtils.getURI(referencesMetaFile(r));
+        writtenFiles.add(referencesUri.toString());
+        PersistUtils.write(referencesUri, ab -> ab.put(referenceKeyTypeMap));
+    }
+
+    private void persistObj(
+        final Keyed<?> o,
+        Map<String, String> referenceKeyTypeMap
+    ) {
+        if (o instanceof Frame) {
+            referenceKeyTypeMap.put(o._key.toString(), ReferenceType.FRAME.toString());
+            String[] writtenFrameFiles = new FramePersist((Frame) o).saveToAndWait(storagePath, true);
+            writtenFiles.addAll(Arrays.asList(writtenFrameFiles));
+        } else if (o != null) {
+            referenceKeyTypeMap.put(o._key.toString(), ReferenceType.KEYED.toString());
+            String destFile = storagePath + "/" + o._key;
+            URI dest = FileUtils.getURI(destFile);
+            PersistUtils.write(dest, ab -> ab.putKey(o._key));
+            writtenFiles.add(destFile);
+        }
+    }
+
+    public void loadReferences(final Recoverable<T> r) {
+        final URI referencesUri = FileUtils.getURI(storagePath + "/" + r.getKey() + REFERENCES_META_FILE_SUFFIX);
+        Map<String, String> referencesMap = PersistUtils.read(referencesUri, AutoBuffer::get);
+        final Futures fs = new Futures();
+        referencesMap.forEach((key, type) -> {
+            switch (ReferenceType.valueOf(type)) {
+                case FRAME: 
+                    FramePersist.loadFrom(Key.make(key), storagePath).get();
+                    break;
+                case KEYED:
+                    PersistUtils.read(URI.create(storagePath + "/" + key), ab -> ab.getKey(Key.make(key), fs));
+                    break;
+                default:
+                    throw new IllegalStateException("Unknown reference type " + type);
+            }
+        });
+        fs.blockForPending();
+    }
+
+}
--- a/h2o-core/src/main/java/hex/grid/Grid.java
+++ b/h2o-core/src/main/java/hex/grid/Grid.java
@@ -1,21 +1,18 @@
 package hex.grid;
 
 import hex.*;
+import hex.faulttolerance.Recoverable;
 import water.*;
 import water.api.schemas3.KeyV3;
 import water.fvec.Frame;
-import water.persist.Persist;
+import water.fvec.persist.PersistUtils;
 import water.util.*;
 import water.util.PojoUtils.FieldNaming;
 
 import java.io.IOException;
-import java.io.OutputStream;
 import java.lang.reflect.Array;
 import java.net.URI;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
 
 import static hex.grid.GridSearch.IGNORED_FIELDS_PARAM_HASH;
 
@@ -26,7 +23,7 @@ import static hex.grid.GridSearch.IGNORED_FIELDS_PARAM_HASH;
  *
  * @param <MP> type of model build parameters
  */
-public class Grid<MP extends Model.Parameters> extends Lockable<Grid<MP>> implements ModelContainer<Model> {
+public class Grid<MP extends Model.Parameters> extends Lockable<Grid<MP>> implements ModelContainer<Model>, Recoverable<Grid<MP>> {
 
   /**
    * Publicly available Grid prototype - used by REST API.
@@ -512,19 +509,16 @@ public class Grid<MP extends Model.Parameters> extends Lockable<Grid<MP>> implem
    * Exports this Grid in a binary format using {@link AutoBuffer}. Related models are not saved.
    *
    * @param gridExportDir Full path to the folder this {@link Grid} should be saved to
+   * @return Path of the file written
    * @throws IOException Error serializing the grid.
    */
-  public void exportBinary(final String gridExportDir) throws IOException {
+  public String exportBinary(final String gridExportDir) {
     Objects.requireNonNull(gridExportDir);
-    final String gridFilePath = gridExportDir + "/" + _key.toString();
     assert _key != null;
+    final String gridFilePath = gridExportDir + "/" + _key;
     final URI gridUri = FileUtils.getURI(gridFilePath);
-    final Persist persist = H2O.getPM().getPersistForURI(gridUri);
-    try (final OutputStream outputStream = persist.create(gridUri.toString(), true)) {
-      final AutoBuffer autoBuffer = new AutoBuffer(outputStream, true);
-      writeWithoutModels(autoBuffer);
-      autoBuffer.close();
-    }
+    PersistUtils.write(gridUri, this::writeWithoutModels);
+    return gridFilePath;
   }
 
   /**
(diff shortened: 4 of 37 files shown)


h2o-core/src/main/java/hex/Model.java
```diff
@@ -399,6 +399,26 @@ public long getOrMakeRealSeed(){
...
+    /**
+     * @return all Frames used by these params
+     */
+    public final Map<String, Frame> getAllFrames() {
+      Map<String, Frame> framesMap = new HashMap<>();
+      collectAllFrames(framesMap);
+      return framesMap;
+    }
+    
+    protected void collectAllFrames(Map<String, Frame> map) {
```
Pscheidl, line 412: Knit/opinion based:

- The collection of two basic frames (validation and train) could be part of the method's contract in JavaDoc, **or** there could be yet another method `collectCommonFrames`, which would make it obvious for overriders what is done. and could selectively pick functionaly to override. Calling `super` is fine to me as well.

- MIssingJavaDoc.

h2o-core/src/main/java/hex/Model.java
```diff
@@ -399,6 +399,26 @@ public long getOrMakeRealSeed(){
...
+      collectFrame(map, "train", _train);
```
Pscheidl, line 413: Should the type of the frame be an enum ? Possibly algo-specific ? Java doesn't support enum inheritance, could be done with constants in `FrameType` class or something like that. Just thinking out loud.

h2o-core/src/main/java/water/fvec/persist/PersistUtils.java
```diff
@@ -9,17 +9,17 @@
...
+    public static String sanitizeUri(String uri) {
         if (uri.endsWith("/")) {
             return uri.substring(0, uri.length()-1);
         } else {
             return uri;
         }
     }
 
-    static <T> T read(URI uri, Reader<T> r) {
+    public static <T> T read(URI uri, Reader<T> r) {
```
Pscheidl, line 22:
```suggestion
    public static void write(URI uri, Writer w) throws RuntimeException {
```

h2o-core/src/main/java/water/fvec/persist/PersistUtils.java
```diff
@@ -31,7 +31,7 @@ static String sanitizeUri(String uri) {
         }
     }
 
-    static void write(URI uri, Writer w) {
+    public static void write(URI uri, Writer w) {
```
Pscheidl, line 34:
```suggestion
    public static void write(URI uri, Writer w) throws RuntimeException {
```

h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -538,6 +540,24 @@ public void exportModelsBinary(final String exportDir) throws IOException {
     }
   }
 
+  /**
+   * Saves all of the frames used by this Grid's params. Frames are named by their keys.
+   * @param exportDir directory to export all the frames to
+   */
+  public void exportFramesBinary(final String exportDir) {
+    Map<String, Frame> frames = getParams().getAllFrames();
+    frames.forEach((name, f) -> new FramePersist(f).saveTo(exportDir, true));
```
Pscheidl, line 549: - The `saveTo` operation can potentially fail on a subset of persisted frames. There are runtimeExceptions thrown, which would cause the underlying Job to fail.
- `name` unused
- It is unclear whether this method blocks until all the frames are exported or not. This brings up yet another question - should be block on this operation (`exportFramesBinary`) and verify the output before moving on to next round ?

A rough proposal that both blocks and checks for errors (but only logs the errors, does not act upon them in any way):
```suggestion
    final List<Job<Frame>> crashedFrames = frames.values().stream()
            .map((f) -> new FramePersist(f).saveTo(exportDir, true))
            .filter(job -> {
              job.get(); // Wait for the persist operation ot complete.
              return job.isCrashed();
            })
            .collect(Collectors.toList());
    if(!crashedFrames.isEmpty()){
      crashedFrames.forEach(frame -> {
        Log.err(String.format("Unable to export binary frame '%s'", frame.get()._key.toString()));
      });
    }
```

h2o-core/src/main/java/hex/schemas/GridSearchSchema.java
```diff
@@ -37,21 +39,27 @@
...
 
-  @API(help="Hyperparameter search criteria, including strategy and early stopping directives.  If it is not given, exhaustive Cartesian is used.", required = false, direction = API.Direction.INOUT)
+  @API(help="Hyperparameter search criteria, including strategy and early stopping directives.  If it is not given, " +
+      "exhaustive Cartesian is used.", direction = INOUT)
   public HyperSpaceSearchCriteriaV99 search_criteria;
 
-  @API(help = "Level of parallelism during grid model building. 1 = sequential building (default). 0 for adaptive parallelism." +
-          "Any number > 1 sets the exact number of models built in parallel.")
+  @API(help = "Level of parallelism during grid model building. 1 = sequential building (default). 0 for adaptive " +
+      "parallelism.Any number > 1 sets the exact number of models built in parallel.")
```
Pscheidl, line 56: Thank you.

Pscheidl (commented):
A couple of thoughts/questions.

h2o-core/src/main/java/water/fvec/persist/PersistUtils.java
Pscheidl: Yes, it's only a matter of preference - therefore a suggestion. I found it quite nice to declare exceptions thrown, even if Java doesn't force us to do so. It's your call, no big deal. It might as well stay as-is.

h2o-r/h2o-package/R/export.R
```diff
@@ -337,6 +337,7 @@ h2o.saveModelDetails <- function(object, path="", force=FALSE) {
 #'
 #' @param grid_directory A character string containing the path to the folder for the grid to be saved to.
 #' @param grid_id A chracter string with identification of the grid to be saved.
+#' @param save_frames Save data frames used by this grid as well
```
Pscheidl, line 340: Usually in R API, first the type expected is described. Not always, but most of the times yes. Just noticed it.

Pscheidl (approved):
LGTM, thanks. One minor thing to think about.

h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -538,6 +535,31 @@ public void exportModelsBinary(final String exportDir) throws IOException {
...
   }
 
+  /**
+   * Saves all of the frames used by this Grid's params. Frames are named by their keys.
+   * @param exportDir directory to export all the frames to
+   */
+  public void exportFramesBinary(final String exportDir) {
+    Map<String, Frame> frames = getParams().getAllFrames();
+    frames.values().stream()
+        .map(frame -> new FramePersist(frame).saveTo(exportDir, true))
```
mn-mikke, line 545: just a question: What file systems are supported by `FramePersist`? Are save and load operations distributed? Any who looks after cleaning `exportDir` when training is properly done?

mn-mikke (commented):
How ofter is grid search checkpointed? is it configurable?

What about `initial_weights` and `initial_biases` on `DeepLearningModel`? Do we need to persist them as well? + What `pretrained_autoencoder`?

h2o-algos/src/main/java/hex/gam/GAMModel.java
```diff
@@ -286,6 +287,13 @@ public long progressUnits() {
 
     public long _seed = -1;
 
+    @Override
+    protected void collectAllFrames(Map<String, Frame> map) {
+      super.collectAllFrames(map);
+      collectFrame(map, "plug_values", _plug_values);
```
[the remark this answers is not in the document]
honzasterba: interesting idea, AFAIK reflection does not contain information about generics, we could find all Key typed params but will not be able to find those that are Key<Frame>

h2o-core/src/main/java/hex/grid/Grid.java
honzasterba: all that are supported by rest of H2O for a distributed cluster a distributed FS should be used
save and load are distributed, each node writes/reads only local chunks, hence thre requirement for same cluster size on load
cleaning of exportDir is left up to the user, this is the same behaviour with normal export_checkpoints_dir option

h2o-algos/src/main/java/hex/gam/GAMModel.java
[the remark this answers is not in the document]
Pscheidl: Or force a method on an algo.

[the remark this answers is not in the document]
sebhrusen: @honzasterba see 
https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/lang/reflect/Field.html#getGenericType()
https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/lang/reflect/Method.html#getGenericParameterTypes()
https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/lang/reflect/ParameterizedType.html

all this is available since 1.5 and H2O-3 is actually using those in `ReflectionUtils`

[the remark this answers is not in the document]
honzasterba: interesting, but this still does not help for the case of DeepLearningModel.initial_weights which is Key[], we could just take all Key and Key[] params and do DKV.get() and look at the object type and decide based on that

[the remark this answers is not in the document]
sebhrusen: agree, this works only for code, there's still type erasure for instances of course, so you can't obtain any magic information from objects stored in DKV

h2o-core/src/main/java/hex/grid/Grid.java
mn-mikke: Thanks for your explanation!

h2o-algos/src/main/java/hex/gam/GAMModel.java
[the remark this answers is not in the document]
sebhrusen: after reflection (pun intended), for cases like `Key[]` we could consider that they should be declared as `Key<Frame>[]` arrays, and then reflection would work...

Anyway, as much as I also like @mn-mikke suggestion, it may require too many changes, idk..

h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -540,6 +535,68 @@ public void exportModelsBinary(final String exportDir) throws IOException {
     }
   }
 
+  @SuppressWarnings("rawtypes")
+  private Set<Key> getDependentKeysFromParams() {
```
michalkurka, line 539: IMHO this code doesn't belong there, grid search shouldn't be itself responsible for its own recovery

instead, it should just say what objects or Keys are needed for resuming and propagate it to the layer above

we eventually want any process to be recoverable, I would expect GS to have a Recoverable interface that would include a method `Set<Key> getDependentObjects` - the layer avobe should be responsible for saving and re-loading those object upon restart

h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -540,6 +535,68 @@ public void exportModelsBinary(final String exportDir) throws IOException {
...
+    Field[] fields = Weaver.getWovenFields(getParams().getClass());
```
michalkurka, line 540: Should we instead use ModelParameterSchema class to get fields (and possibly their metadata)?

h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -540,6 +535,68 @@ public void exportModelsBinary(final String exportDir) throws IOException {
...
+   * Saves all of the keyed objects used by this Grid's params. Files are named by objects' keys.
+   *
+   * @param exportDir directory to export all the frames to
+   */
+  public void exportReferencesBinary(final String exportDir) {
+    final Set<Key> keys = getDependentKeysFromParams();
+    final IcedHashMap<String, String> referenceKeyTypeMap = new IcedHashMap<>();
+    keys.stream()
+        .map(Key::get)
+        .map(obj -> persistObj(obj, exportDir, referenceKeyTypeMap))
```
michalkurka, line 587: you keep exporting even after it already crashed because you are throwing an exception and skipping rest of the jobs - this behavior is definitely not ideal

simple sequential write would probably make more sense here

h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -540,6 +535,68 @@ public void exportModelsBinary(final String exportDir) throws IOException {
...
+  private Job persistObj(
+      final Keyed o,
+      final String exportDir,
+      Map<String, String> referenceKeyTypeMap
+  ) {
+    if (o instanceof Frame) {
+      referenceKeyTypeMap.put(o._key.toString(), "frame");
+      return new FramePersist((Frame) o).saveTo(exportDir, true);
+    } else if (o != null) {
+      referenceKeyTypeMap.put(o._key.toString(), "iced");
```
michalkurka, line 570: What would this be?

honzasterba: pretrained_autoencoder in DeepLearning for example

michalkurka: in this case, it would only write DeepLearningModel but not actually the weights and biases (that are in DKV), am I right?

h2o-core/src/main/java/hex/grid/Grid.java
honzasterba: fixed

h2o-core/src/main/java/hex/grid/Grid.java
honzasterba: refactored, please look it over

Pscheidl (approved):
Thanks.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -439,23 +455,23 @@ private void reconcileMaxRuntime(Key<Grid<MP>> gridKey, Model.Parameters params)
       Log.info("Due to the grid time limit, changing model max runtime to: " + params._max_runtime_secs + " secs.");
     }
   }
-
+  
+  private void beforeGridStart(Grid grid) {
+    if (_exportFrames) {
+      attemptGridSave(grid);
+      new Recovery<Grid>(_hyperSpaceWalker.getParams()._export_checkpoints_dir).exportReferences(grid);
```
michalkurka, line 462: @honzasterba IMHO this is mixing two concepts together

Parameter `export_checkpoints_dir` is meant for the _user_ to export models - this can have a different motivation than just recovery.

Fault-tolerance is driven by the software, not the user. I think there should be 2 locations - one for user checkpointed objects (that one is preserved after modeling ends) and one for fault-tolerance objects (that was is cleaned up automatically).

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -67,23 +67,28 @@
...
+  private GridSearch(
+      Key<Grid> gkey, HyperSpaceWalker<MP, ?> hyperSpaceWalker, 
+      final boolean exportFrames, final int parallelism
+  ) {
     assert hyperSpaceWalker != null : "Grid search needs to know how to walk around hyper space!";
     _hyperSpaceWalker = hyperSpaceWalker;
     _result = gkey;
     String algoName = hyperSpaceWalker.getParams().algoName();
     _job = new Job<>(gkey, Grid.class.getName(), algoName + " Grid Search");
+    _exportFrames = exportFrames;
```
michalkurka, line 91: instead of passing this parameter, I think it makes more logical sense to pass the recovery object - where the grid can store recovery information; if this parameter is not passed - nothing happens

h2o-py/h2o/grid/grid_search.py
```diff
@@ -72,17 +76,20 @@ class H2OGridSearch(h2o_meta(Keyed)):
 
 
     def __init__(self, model, hyper_params, grid_id=None, search_criteria=None, export_checkpoints_dir=None,
-                 parallelism=1):
+                 recovery_dir=None, parallelism=1):
```
Pscheidl, line 79: This API requires the user to specify the dir. What if H2O is managed ? I believe in managed runs, like on Kubernetes, an automatic regime would be useful. How are we going to solve this ? By introducing another parameter in future, e.g. `autorecovery` ? 

Could be solved by enums in Py (not sure how to solve this in R, but S4 classes seems to offer a solution) and introduce two modes for a single `recovery` parameter. Pseudocode:

enum Recovery{
Directory(Path directoryToSaveTo),
Automated
}

The automated mode could simply dump it to a well known location based on system, e.g. /var/h2o on linux.

h2o-core/src/main/java/hex/faulttolerance/Recovery.java
```diff
@@ -0,0 +1,136 @@
...
+
+public class Recovery<T extends Keyed> {
+
+    public static final String REFERENCES_META_FILE_SUFFIX = "_references";
+
+    public enum ReferenceType {
+        FRAME, KEYED
+    }
+
+    private final String storagePath;
```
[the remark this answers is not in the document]
honzasterba: I think that is out of scope for this PR, everywhere we use strings to pass URIs so that would have to be a bigger change

h2o-core/src/main/java/water/fvec/persist/PersistUtils.java
```diff
@@ -31,7 +31,7 @@ static String sanitizeUri(String uri) {
...
         final Persist persist = H2O.getPM().getPersistForURI(uri);
         try (final OutputStream outputStream = persist.create(uri.toString(), true)) {
             final AutoBuffer autoBuffer = new AutoBuffer(outputStream, true);
```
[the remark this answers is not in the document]
honzasterba: believe me I tried adding it! but autobuffer does so many things that I gave up

h2o-core/src/main/java/hex/faulttolerance/Recoverable.java
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * @return key of this keyed object
+     */
+    Key<T> getKey();
+
+    /**
+     * @param location directory where this recoverable will be written into a single file
+     * @return path to where data was written
+     */
+    String exportBinary(String location);
```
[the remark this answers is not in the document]
honzasterba: this is not part of public API so, for now I think we are good and since another target is autoML I think one file should be enough, we can alway change this if it is not sufficient

h2o-py/h2o/grid/grid_search.py
honzasterba: This API is intended for purely manual use case - see the tests that use it, basically the user is responsible for starting a new cluster and loading the grid and continuing training.
For K8s and other auto-recoverable envs I would imagine this would be done somewhat automatically by the back-end. So the user would only say the cluster should auto-recover and all this magic will happen behind the scenes.

Pscheidl: I see.  Seems fine to me.

Pscheidl (approved):

h2o-py/tests/testdir_algos/grid/pyunit_grid_fault_tolerance_resume.py
```diff
@@ -0,0 +1,330 @@
...
+    resume_grid(loaded, loaded_train, params)
+    print("Newly grained grid has %d models:" % len(loaded.model_ids))
+    for x in sorted(loaded.model_ids):
+        print(x)
+    assert len(loaded.model_ids) == grid_size, "The full grid was not trained."
+    h2o.remove_all()
+    
+
+def grid_ft_resume_test():
+    test_dl()
```
michalkurka, line 319: is this commented out intentionally?

h2o-py/tests/testdir_algos/grid/pyunit_grid_manual_export_resume_new_hyperspace.py
```diff
@@ -15,12 +15,13 @@ def grid_resume():
...
-    hyper_parameters = OrderedDict()
-    hyper_parameters["learn_rate"] = learn_rate_opts
-    hyper_parameters["ntrees"] = ntrees_opts
+    hyper_parameters = {
+        "learn_rate": learn_rate_opts,
+        "ntrees": ntrees_opts
+    }
     print("GBM grid with the following hyper_parameters:", hyper_parameters)
     
-    export_dir = pyunit_utils.locate("results")
```
michalkurka, line 23: Why this change? It is better to output file to results directory, then they are preserved if anything fails.

h2o-r/h2o-package/R/grid.R
```diff
@@ -37,7 +37,10 @@
 #'        max_models = 42, max_runtime_secs = 28800)} or \code{list(strategy = "RandomDiscrete", stopping_metric = "AUTO", 
 #'        stopping_tolerance = 0.001, stopping_rounds = 10)} or \code{list(strategy = "RandomDiscrete", stopping_metric = 
 #'        "misclassification", stopping_tolerance = 0.00001, stopping_rounds = 5)}.
-#' @param export_checkpoints_dir Directory to automatically export grid in binary form to.
+#' @param export_checkpoints_dir Directory to automatically export grid and its models to.
+#' @param recovery_dir When specified the grid and all necessary data (frames, models) will be saved to this
+#'        directory (use HDFS or other distributed file-system). Should the cluster crash during training, the grid
+#'        can be reloaded from this directory via \code{h2o.load_grid} and training can be resumed
```
michalkurka, line 43: this seems to be called h2o.loadGrid

h2o-core/src/main/java/water/api/schemas3/GridExportV3.java
```diff
@@ -5,13 +5,16 @@
...
+      required = true, direction = API.Direction.INPUT, level = API.Level.critical)
   public String grid_id;
 
-  @API(required = true, direction = API.Direction.INPUT, help = "Path to the directory with saved Grid search",
-          level = API.Level.critical)
+  @API(help = "Path to the directory with saved Grid search",
+      required = true, direction = API.Direction.INPUT, level = API.Level.critical)
   public String grid_directory;
   
+  @API(help = "True if data frames should also be saved",
```
michalkurka, line 16: I think we need to be more clear what is actually saved. Users can easily confuse this for eg. CV holdout predictions or other frames that are created as a _result_ of model training.

h2o-py/tests/testdir_algos/grid/pyunit_grid_manual_export_resume_new_hyperspace.py
honzasterba: I thought only test results are to be written to the results dir, tmp data should be in tempfiles, same for junit

h2o-r/h2o-package/R/grid.R
honzasterba: fixed

h2o-core/src/main/java/water/api/schemas3/GridExportV3.java
honzasterba: good call, I tried to improve the naming and docs

hannah-tillman (approved):

h2o-py/tests/testdir_algos/grid/pyunit_grid_manual_export_resume_new_hyperspace.py
michalkurka: That depends on what is tmp data to you but this use-case is imho similar to eg. checkpointing and those tests use pyunit_utils.locate("results"), example https://github.com/h2oai/h2o-3/blob/master/h2o-py/tests/testdir_algos/xgboost/pyunit_checkpoint_xgboost.py#L30

For data exchange for tests we follow the pyunit_utils.locate("results") method because it has a clear lifecycle.

The temp file will be gone with the docker exiting on Jenkins which could be a bad thing for reproducing a bug.

michalkurka (approved):
Thank you @honzasterba

PR #5134, opened 2020-11-15
PUBDEV-7753 - Disable REST API on non-leader K8S nodes

Pscheidl:
https://h2oai.atlassian.net/browse/PUBDEV-7753

Can be even tried manually by using `-Dsys.ai.h2o.disableNonLeaderApi=true`


h2o-core/src/main/java/water/init/NetworkInit.java
```diff
@@ -30,6 +30,7 @@
  */
 public class NetworkInit {
 
+  public static String DISABLE_NON_LEADER_API = H2O.OptArgs.SYSTEM_PROP_PREFIX + ".disableapinonleader";
```
Pscheidl, line 33: This is a system-level property, not directly user facing. One can always do `-Dsys.ai.h2o.disableapinonleader=true`, but users should not care about it.

h2o-jetty-8/src/main/java/water/webserver/jetty8/Jetty8ServerAdapter.java
```diff
@@ -134,9 +134,10 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
 
   private class GateHandler extends AbstractHandler {
     @Override
-    public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) {
-      final boolean handled = h2oHttpView.gateHandler(request, response);
-      if (handled) {
+    public void handle(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException {
```
Pscheidl, line 137: Jetty 8 module should be gone soon. For now, I suggest we ignore the code duplication over here.

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
```diff
@@ -65,7 +64,7 @@ public void init() {
         if (!runningOnKubernetes) {
             return; // Do not initialize any configuration if H2O is not running in K8S-spawned container.
         }
-
+        System.setProperty(NetworkInit.DISABLE_NON_LEADER_API, "true");
```
Pscheidl, line 67: - There is also the option to manipulate some state directly, and not indirectly via a system property.
- Not sure if embedded config should manipulate any internal state directly. Probably not. One way to solve this is to introduce yet another overridable method in `EmbeddedConfig` to override/set system properties and args alike. At the moment, this would have to be done just for this functionality.
- The only user of thins functionality is now Kubernetes, yet it seems like a good idea to expose this setting as a sysprop for anyone to set it in any environment, if required.

h2o-core/src/main/java/water/init/NetworkInit.java
```diff
@@ -30,6 +30,7 @@
  */
 public class NetworkInit {
 
+  public static String DISABLE_NON_LEADER_API = H2O.OptArgs.SYSTEM_PROP_PREFIX + "disableNonLeaderApi";
```
Pscheidl, line 33: This is a system-level property, not directly user facing. One can always do -Dsys.ai.h2o.disableapinonleader=true, but users should not care about it. Therefore not taking it into H2o.args.

h2o-jetty-8/src/main/java/water/webserver/jetty8/Jetty8ServerAdapter.java
honzasterba: why not put this into the `boolean gateHandler` method? would avoid the duplication

Pscheidl: That would avoid a lot of things actually. Let me see.

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
```diff
@@ -65,7 +65,7 @@ public void init() {
         if (!runningOnKubernetes) {
             return; // Do not initialize any configuration if H2O is not running in K8S-spawned container.
         }
-
+        System.setProperty(H2OHttpViewImpl.DISABLE_NON_LEADER_API, "true");
```
Pscheidl, line 68: - There is also the option to manipulate some state directly, and not indirectly via a system property.
- Not sure if embedded config should manipulate any internal state directly. Probably not. One way to solve this is to introduce yet another overridable method in EmbeddedConfig to override/set system properties and args alike. At the moment, this would have to be done just for this functionality.
- The only user of thins functionality is now Kubernetes, yet it seems like a good idea to expose this setting as a sysprop for anyone to set it in any environment, if required.

h2o-core/src/main/java/water/webserver/H2OHttpViewImpl.java
```diff
@@ -23,6 +24,8 @@
  * This is intended to be a singleton per H2O node.
  */
 public class H2OHttpViewImpl implements H2OHttpView {
+  private static final Logger LOG = Logger.getLogger(H2OHttpViewImpl.class);
+  public static String DISABLE_NON_LEADER_API = H2O.OptArgs.SYSTEM_PROP_PREFIX + "disableNonLeaderApi";
```
Pscheidl, line 28: This is a system-level property, not directly user facing. One can always do -Dsys.ai.h2o.disableapinonleader=true, but users should not care about it. Therefore not taking it into H2o.args.

h2o-core/src/main/java/water/webserver/H2OHttpViewImpl.java
```diff
@@ -23,6 +24,8 @@
  * This is intended to be a singleton per H2O node.
  */
 public class H2OHttpViewImpl implements H2OHttpView {
+  private static final Logger LOG = Logger.getLogger(H2OHttpViewImpl.class);
```
Pscheidl, line 27: Allowed myself to migrate to the newer logger.

h2o-jetty-8/src/main/java/water/webserver/jetty8/Jetty8ServerAdapter.java
Pscheidl: That actually led to a lot of simplifications, as no additional state transfer to webserver impl from core is done. Everything stays "nicely" in core.

honzasterba (commented):
I am not totally sure about using the sysprop for transfering the information from the embedded config to the HTTPView. I think it would be cleaner to have this be a property on the embedded config which is then copied over to H2OHttpConfig

Pscheidl:
@honzasterba exactly as mentioned here: https://github.com/h2oai/h2o-3/pull/5134#discussion_r523955077

Question:
- Should we keep this simple and add a single embedded config property, that is mostly Kubernetes-specific ?
- Should we add a general mechanism to override any properties possible ?

There is a strong push to keep things simple and fast, so I'd go with the first option.

Pscheidl:
I went the way of my preference, let me know what you think about it. To be found in https://github.com/h2oai/h2o-3/pull/5134/commits/c02d5287a5c6e20981ef97c9bd4d385a4b62993f .

Thank you @honzasterba .

h2o-core/src/main/java/water/webserver/H2OHttpViewImpl.java
```diff
@@ -84,6 +88,12 @@ public boolean gateHandler(HttpServletRequest request, HttpServletResponse respo
       if (ServletUtils.isTraceRequest(request)) {
         ServletUtils.setResponseStatus(response, HttpServletResponse.SC_METHOD_NOT_ALLOWED);
         return true;
+      } else if (!isApiEnabledOnThisNode()) {
```
honzasterba, line 91: seems inefficient to call this on every request

h2o-core/src/main/java/water/webserver/H2OHttpViewImpl.java
```diff
@@ -94,6 +104,30 @@ public boolean gateHandler(HttpServletRequest request, HttpServletResponse respo
...
+
+  /**
+   * @return False if API is supposed to be disabled on a non-leader node and this node is a non-leader node. Otherwise true.
+   */
+  public static boolean isApiEnabledOnThisNode() {
+    final boolean disabledOnNonLeaderNode;
+    final AbstractEmbeddedH2OConfig embeddedConfig = H2O.getEmbeddedH2OConfig();
+    if (embeddedConfig == null) {
+      // If there is no embedded config, use only the sysvar, if present.
+      disabledOnNonLeaderNode = Boolean.getBoolean(DISABLE_NON_LEADER_API);
```
honzasterba, line 116: whats the point of this sys prop? easier testing? if there is no point for us I would suggest removing it, its less code, and less code is always better

Pscheidl: No. The point is to have it exposed in case anyone would need it at any time in future - exact use case not specified. I agree, let's delete it.

h2o-core/src/main/java/water/server/ServletMeta.java
```diff
@@ -9,10 +9,21 @@
...
+  /**
+   * Constructs a new instance of {@link ServletMeta} with the alwaysEnabled functionality either turned on or off.
+   * An always enabled servlet is guaranteed to be active on every node, even if the rest of the API is shut down
+   * on a particular node. This is useful for vital, per-node APIs. Typically used by non user-facing functionality.
+   * 
+   * @param contextPath Context path the underlying servlet handles
+   * @param servletClass Specific implementation of the {@link HttpServlet} to handle the context path
+   * @param alwaysEnabled Set to true if this servlet should remain active even the API is turned off on a particular node.
+   */
+  public ServletMeta(final String contextPath, final Class<? extends HttpServlet> servletClass, final boolean alwaysEnabled) {
```
Pscheidl, line 23: This is adjusted mainly due to Sparkling Water using some endpoints for data upload on each node which must absolutely always be enabled, and especially on Kubernetes :wink: 

I was talking to @mn-mikke about this and we both agreed SW should be in control which endpoints remain active and it should be immediately clear from the code. The alternative to this is list the allowed context paths as Strings on H2O's side, but that would easily go out of sync with SW.

h2o-webserver-iface/src/main/java/water/webserver/iface/RequestAuthExtension.java
```diff
@@ -20,4 +20,12 @@
    * @throws ServletException -
    */
   boolean handle(String target, HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException;
+
+  /**
+   * @return True if the given extension is enabled. Otherwise false.
+   */
+  default boolean isEnabled(){
```
Pscheidl, line 27: Not to break compatibility with existing extensions in different repositories.

Pscheidl:
Ready for review @michalkurka @honzasterba @mn-mikke @satai .

Re-worked completely in the end. One of the changes with the ServletMeta class might be a little bit controversial. Described in the code why I believe this is the best solution.

Modification of the constructor instead of adding another one seemed best - not sure if I break any contract with that.

Thank you for your patient reviews @honzasterba. This pushed this PR forward a lot.

h2o-core/src/main/java/water/server/ServletMeta.java
```diff
@@ -9,10 +9,21 @@
 
   private final String _contextPath;
   private final Class<? extends HttpServlet> _servletClass;
+  private final boolean _alwaysEnabled;
 
-  public ServletMeta(String contextPath, Class<? extends HttpServlet> servletClass) {
+  /**
```
michalkurka, line 14: A constructor with default value false for alwaysEnabled would make sense since the default is usually what you want.

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
```diff
@@ -1,20 +1,20 @@
 package water.k8s;
 
-import fi.iki.elonen.NanoHTTPD;
 import water.H2O;
 import water.init.AbstractEmbeddedH2OConfig;
 import water.init.EmbeddedConfigProvider;
+import water.init.NetworkInit;
```
michalkurka, line 6: Unused?

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
```diff
@@ -1,20 +1,20 @@
...
 import water.k8s.api.KubernetesRestApi;
 import water.k8s.lookup.KubernetesDnsLookup;
 import water.k8s.lookup.KubernetesLookup;
 import water.k8s.lookup.LookupConstraintsBuilder;
 import water.util.Log;
+import water.webserver.H2OHttpViewImpl;
```
michalkurka, line 12: Unused?

h2o-core/src/main/java/water/server/ServletMeta.java
Pscheidl: Two constructors already on three parameters. I originally had it like that and removed it. It feels awkward to propose builder pattern anywhere. I made it so in https://github.com/h2oai/h2o-3/pull/5134/commits/8668ae733cabc6cd257c1c1946903d503d85722b .

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
Pscheidl: There was one more. Removed in https://github.com/h2oai/h2o-3/pull/5134/commits/11706dd558cec20d3df29ef3140c9172f0625a98 .

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
Pscheidl: There was one more. Removed in https://github.com/h2oai/h2o-3/pull/5134/commits/11706dd558cec20d3df29ef3140c9172f0625a98 .

scripts/jenkins/groovy/defineTestStages.groovy
```diff
@@ -53,6 +53,12 @@ def call(final pipelineContext) {
     [
       stageName: 'Java 8 Smoke', target: 'test-junit-smoke-jenkins', javaVersion: 8, timeoutValue: 20,
       component: pipelineContext.getBuildConfig().COMPONENT_JAVA
+    ],
+    [
```
Pscheidl, line 57: This will of course disapper before merging.

h2o-k8s/src/main/java/water/k8s/KubernetesRequestFilter.java
```diff
@@ -0,0 +1,63 @@
...
+                .keySet());
+
+    }
+
+    @Override
+    public boolean handle(String target, HttpServletRequest request, HttpServletResponse response) throws IOException {
+        if (H2O.SELF == null) {
+            // If H2O clustering process has not yet finished, disable API on all nodes to prevent premature cluster locking.
+            // Send HTTP 403 - Forbidden and indicate the clustering process has not finished yet.
+            response.sendError(HttpServletResponse.SC_FORBIDDEN, "H2O Clustering on Kubernetes not yet finished. H2O API is inaccessible at the moment.");
```
Pscheidl, line 40: This could be 503 as well.

h2o-core/src/main/java/water/server/ServletMeta.java
michalkurka: I think changes to this class are not a big problem, however, we need to keep in mind it is an extension point and might be used by 3rd parties.

Keeping the simplest constructor makes sense to me. More complex use can have a builder pattern for example.

h2o-k8s/src/main/java/water/k8s/KubernetesRequestFilter.java
```diff
@@ -0,0 +1,63 @@
...
+            return false;
+        } else if (allowedContextPaths.contains(target)) {
+            // If this is not the leader node, yet it is the part of API that should be enabled on every node, indicate this request
+            // not been handled by this filter and do nothing.
+            return false;
+        }
+
+        // If API is disabled on this node and the context path of the request is not listed in allowedPaths,
+        // Then send HTTP 403 - Forbidden and indicate the request has been handled by this filter.
+        response.sendError(HttpServletResponse.SC_FORBIDDEN, "API not accessible on this node while running on Kubernetes, as it is not a leader node.");
```
michalkurka, line 54: Maybe we should say this is a deployment error?

h2o-core/src/main/java/water/server/ServletMeta.java
Pscheidl: Question is whether to introduce it now or not. Again, there will be many opinions. I'd stick with the current two-constructor solution and then introduce a builder pattern. Some might even say there is either the constructor approach **or** builder with no constructors .... often I simply don't know what is wrong.

I'd say currently we're fine. One more and there will be builder together with constructors. Is that something acceptable ?

scripts/jenkins/groovy/defineTestStages.groovy
michalkurka: oh, okay!

h2o-k8s/src/main/java/water/k8s/KubernetesRequestFilter.java
Pscheidl: Why not. Adjusting.

h2o-core/src/main/java/water/server/ServletMeta.java
michalkurka: IMHO feel free to introduce now. As long as the original constructor that would be used in 99.9% scenarios stays - I am fine with anything ;)

h2o-k8s/src/main/java/water/k8s/KubernetesRequestFilter.java
Pscheidl: Is this worth creating a technical note ?

Pscheidl: Re-wroded in https://github.com/h2oai/h2o-3/pull/5134/commits/1174e169e465347682d59538afe9df9244674b87
to `response.sendError(HttpServletResponse.SC_FORBIDDEN, "Kubernetes deployment/configuration error. Request reached a non-leader H2O node.");`

h2o-core/src/main/java/water/server/ServletMeta.java
Pscheidl: I think we're fine now.

Pscheidl: Added one now.

michalkurka (approved):
Thank you @Pscheidl

h2o-core/src/main/java/water/server/ServletMeta.java
```diff
@@ -22,5 +34,43 @@ public String getContextPath() {
   public Class<? extends HttpServlet> getServletClass() {
     return _servletClass;
   }
+  
+  public boolean isAlwaysEnabled(){
+      return _alwaysEnabled;
+  }
+  
+  public static class Builder {
```
honzasterba, line 42: does not seem to be used

PR #5135, opened 2020-11-16
PUBDEV-7861 - standalone R client

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-7861


h2o-r/build.gradle
```diff
@@ -152,11 +152,65 @@ task cpH2OAppJar(type: Copy, dependsOn: [configurations.compile]) {
...
+                return 'h2o.jar'
+            }
+            into "${buildDir}/tmp/h2o-package-client/inst/java"
+        }
+
+        File connectionR = new File([T, "h2o-package", "R", "connection.R"].join(File.separator))
+        File connectionClient = new File([T, "build", "tmp", "h2o-package-client", "R", "connection.R"].join(File.separator))
+
+        def connectionRReplaced = connectionR.text
+                .replaceAll("SUBST_START_COMMAND", "h2o.connect()")
```
honzasterba, line 201: same as above

h2o-r/build.gradle
```diff
@@ -152,11 +162,55 @@ task cpH2OAppJar(type: Copy, dependsOn: [configurations.compile]) {
...
+task cpMainPKG {
+    doLast {
+        file("${buildDir}/h2o-package-main").mkdirs()
+        copy {
+            from "h2o-package"
+            into "${buildDir}/h2o-package-main"
+        }
+        File connectionR = new File([T, "h2o-package", "R", "connection.R"].join(File.separator))
+        File connectionMain = new File([T, "build", "h2o-package-main", "R", "connection.R"].join(File.separator))
+        editConnectionSource("h2o-package-main", "h2o.init()", "After starting H2O, you can use the Web UI" +
```
honzasterba, line 174: declarations above (connectionMain) are unused

valenad1: True 🤦

h2o-r/h2o-package/R/connection.R
```diff
@@ -492,13 +492,12 @@ h2o.clusterStatus <- function() {
     "----------------------------------------------------------------------\n",
     "\n",
     "Your next step is to start H2O:\n",
```
honzasterba, line 494: Now that I read this again, I think we should remove this whole block. Having this print out just
`Your next step is to start H2O: >` seems not very user friendly.
I would either remove this all together or replace with
`You next step is to connect to your H2O cluster: > h2o.connect(url="http://hostname:port")`

h2o-r/build.gradle
```diff
@@ -152,11 +162,54 @@ task cpH2OAppJar(type: Copy, dependsOn: [configurations.compile]) {
...
 
+task cpMainPKG {
+    doLast {
+        file("${buildDir}/h2o-package-main").mkdirs()
+        copy {
+            from "h2o-package"
+            into "${buildDir}/h2o-package-main"
+        }
+
+        editConnectionSource("h2o-package-main", "h2o.init()", "After starting H2O, you can use the Web UI" +
```
honzasterba, line 173: I still don't like that we have this "code" inside of build file. the amount of back-slashes in the strings makes it error prone.
I would at least extract these strings into a `map` at the beggining of the build file.
Ideally I would move these into .txt files somewhere in the h2o-r directory (for example h2o-r/templates/main/start_command.txt, start_command_doc.txt) and add a README into the h2o-r/templates to document how these are used.

h2o-r/h2o-package/R/connection.R
```diff
@@ -348,7 +348,7 @@ h2o.connect <- function(ip = "localhost", port = 54321, strict_version_check = T
 h2o.getConnection <- function() {
   conn <- .attemptConnection()
   if (is.null(conn))
-    stop("No active connection to an H2O cluster. Did you run `h2o.init()` ?")
```
honzasterba, line 351: I would remove the whole `Did you run h2o.init()?` sentence. It makes no sense to end up with "Did you run `` ?"

valenad1: It will always be either with h2o.init() or h2o.connect(). SUBST_START_DOCUMENTATION is replaced with ""

michalkurka:
It might be useful to split this into 2 parts:

1. making the client package
2. modifying the messages

The main thing is to prevent users to accidentally start a local H2O instance (1). Having nice messages is secondary. If we can expedite delivery by doing just (1) for now - then I am all for it.

h2o-r/h2o-package/R/connection.R
honzasterba: My mistake!

h2o-r/h2o-package/R/connection.R
honzasterba: I see my mistake, this will be replaced with h2o.connect, so this is fine.

h2o-r/build.gradle
valenad1: Your suggestion is implemented [here](https://github.com/h2oai/h2o-3/pull/5135/commits/09ef16e32bf1de4296f9aa97933137e68769e645).

I try also use source.properties file to define variables and values. Implementation is on HEAD of this branch.

Please tell me what you think

honzasterba (approved):
lgtm

h2o-r/h2o-package/R/connection.R
```diff
@@ -348,7 +348,7 @@ h2o.connect <- function(ip = "localhost", port = 54321, strict_version_check = T
...
+    stop("No active connection to an H2O cluster. Did you run `SUBST_START_COMMAND` ?")
```
michalkurka, line 351: I think this will definitely work but in my opinion, this is not a very robust solution. I think it would be better if we make it work without modifying files.

I suggest the following solution:

Introduce simple boolean flag "is_client" (naming can be different), populate this to true/false based existence of client_info or similar file in the package (you can also generate a file client_info.r during the build). Use an if/else statement to print out the correct message. Defaults should be false -> the package should work normally if the gradle breaks. That is the most common use case.

This has the advantage that it allows us to introduce argument checks in the h2o.init function when it is actually used in client-only mode.

I think this is a far flexible solution.

valenad1: It is implemented. I wasn't sure about content of client_info.txt so I used dist-readme.txt

michalkurka (approved):
thank you @valenad1

PR #5136, opened 2020-11-16
XGBoost: Fix MOJO loading when interaction_constraints is used

michalkurka:


h2o-genmodel/src/main/java/hex/genmodel/attributes/ModelJsonReader.java
```diff
@@ -314,6 +314,17 @@ private static Object convertBasedOnJsonType(final JsonElement convertFrom, fina
                         }
                         convertTo = arrS;
                         break;
+                    case STRING_ARR_ARR:
+                        final String[][] arrSS = new String[array.size()][];
+                        for (int i = 0; i < array.size(); i++) {
+                            final JsonArray arr2 = array.get(i).getAsJsonArray();
+                            arrSS[i] = new String[arr2.size()];
+                            for (int j = 0; j < array.size(); j++) {
```
mn-mikke, line 322: Shouldn't the for loop condition be `j < arr2.size()`?

Pscheidl (approved):
LGTM completely

h2o-genmodel/src/main/java/hex/genmodel/attributes/ModelJsonReader.java
honzasterba: missing unit test

michalkurka: thx, test improved to check for expected output

honzasterba (approved):

michalkurka:
@mn-mikke thank you for catching the bug - the test worked only thanks to a coincidence

thanks for the review and pointers @honzasterba and @Pscheidl

PR #5137, opened 2020-11-17
PUBDEV-7896 - do not discard Keys in h2o.removeAll() command

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-7896

### DIFF
--- a/h2o-r/h2o-package/R/kvstore.R
+++ b/h2o-r/h2o-package/R/kvstore.R
@@ -126,7 +126,7 @@ h2o.rm <- function(ids, cascade=TRUE) {
     } else if( is.character(xi) ) {
       .h2o.__remoteSend(paste0(.h2o.__DKV, "/",xi), method = "DELETE", .params=list(cascade=cascade))
     } else {
-      stop("input to h2o.rm must be a Keyed instance (e.g. H2OFrame, H2OModel) or character")
+      stop("Input to h2o.rm must be either an instance of H2OModel/H2OFrame or a character")
     }
   }
 
--- a/h2o-r/h2o-package/R/kvstore.R
+++ b/h2o-r/h2o-package/R/kvstore.R
@@ -76,6 +76,10 @@ h2o.removeAll <- function(timeout_secs=0, retained_elements = c()) {
         retained_keys <- append(retained_keys, element@model_id)
       } else if (is.H2OFrame(element)) {
         retained_keys <- append(retained_keys, h2o.getId(element))
+      } else if( is.character(element) ) {
+        retained_keys <- append(retained_keys, element)
+      } else {
+        stop("The 'retained_elements' variable must be a Keyed instance (e.g. H2OFrame, H2OModel) or character")
       }
     }
     
--- a/h2o-r/h2o-package/R/kvstore.R
+++ b/h2o-r/h2o-package/R/kvstore.R
@@ -47,12 +47,12 @@ h2o.ls <- function() {
 #' Remove All Objects on the H2O Cluster
 #'
 #' Removes the data from the h2o cluster, but does not remove the local references.
-#' Retains frames and vectors specified in retained_elements argument.
-#' Retained keys must be keys of models and frames only. For models retained, training and validation frames are retained as well.
+#' Retains models, frames and vectors specified in retained_elements argument.
+#' Retained elements must be instances/ids of models and frames only. For models retained, training and validation frames are retained as well.
 #' Cross validation models of a retained model are NOT retained automatically, those must be specified explicitely.
 #'
 #' @param timeout_secs Timeout in seconds. Default is no timeout.
-#' @param retained_elements Frames and vectors to be retained. Other keys provided are ignored.
+#' @param retained_elements Instances or ids of models and frames to be retained. Combination of instances and ids in the same list is also a valid input.
 #' @seealso \code{\link{h2o.rm}}
 #' @examples
 #' \dontrun{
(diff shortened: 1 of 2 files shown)


honzasterba (commented):
test is missing

h2o-r/h2o-package/R/kvstore.R
```diff
@@ -76,6 +76,10 @@ h2o.removeAll <- function(timeout_secs=0, retained_elements = c()) {
         retained_keys <- append(retained_keys, element@model_id)
       } else if (is.H2OFrame(element)) {
         retained_keys <- append(retained_keys, h2o.getId(element))
+      } else if( is.character(element) ) {
+        retained_keys <- append(retained_keys, element)
+      } else {
+        stop("The 'retained_elements' variable must be a Keyed instance (e.g. H2OFrame, H2OModel) or character")
```
michalkurka, line 82: I think the naming in this method is probably not consistent with rest of the package. On the client-side we usually don't use term "keys", we use "ids" instead. See model id, frame id, functions like h2o.getFrame, h2o.getId...

Thankfully the API of this method calls for "retained_elements" and doesn't use the "key" terminology.

I think we should
- fix the documentation of this method to use "ids" instead of "keys" where appropriate. 
- since we are allowing also ids/keys as a parameter the documentation should reflect that.
- error message should be imho simplified to something like "... be either an  instance of H2OFrame/H2OModel or an id of H2OFrame/H2OModel.

cc: @Pscheidl

michalkurka:
@Pscheidl this is your code, please review @valenad1's changes

h2o-r/h2o-package/R/kvstore.R
Pscheidl: That is corrected, I used the same terminology as the Python client uses to unify it - definitely should've been adapted to R conventions we've got there so far.

Pscheidl:
It's a good move to accept the keys as strings as well. 

- Document the behavior in method's description
- What happens if users add wrong/nonexistent key ? The DKVManager itself should ignore it and just move on. Is this acceptable for users ? I think yes. Accepting only the objects was a double-edged sword and server partially as a protection against that - it's easy to pass a string, yet it's not as easy to pass your own estimator. :wink:

h2o-r/h2o-package/R/kvstore.R
valenad1: Documentation is updated. I slightly update the documentation for h2o.rm also

valenad1:
> 
> 
> Thank you @valenad1 - It's a good move to accept the keys as strings as well.
> 
>     * Document the behavior in method's description

Documented.

> 
>     * What happens if users add wrong/nonexistent key ? The DKVManager itself should ignore it and just move on. Is this acceptable for users ? I think yes. Accepting only the objects was a double-edged sword and server partially as a protection against that - it's easy to pass a string, yet it's not as easy to pass your own estimator. 😉
> 
> 
> For tests, the `runit_retain.R` might be extended to test the new functionality.

I agree, when user makes a typo in string then something can be accidentally lost without any warning. I improve this behavior to check if provided key is in h2o.ls() and if not then do not perform remove. Please check. 

If user provide something different than Frame or Model then it also do not perform remove.

h2o-r/h2o-package/R/kvstore.R
```diff
@@ -76,6 +76,14 @@ h2o.removeAll <- function(timeout_secs=0, retained_elements = c()) {
         retained_keys <- append(retained_keys, element@model_id)
       } else if (is.H2OFrame(element)) {
         retained_keys <- append(retained_keys, h2o.getId(element))
+      } else if( is.character(element) ) {
+        if (is.element(element, h2o.ls()$key)) {
```
michalkurka, line 80: I think we should skip this validation and let the backend handle the error.

If the user is providing a key, they probably know what they are doing. Moreover, this still doesn't cover all cases (key can be valid but it isn't a model/frame).

I would rather keep this simple.

Pscheidl: This validation is indeed done on backend. In `DKVManager`, roughly on line 53:
```java
        throw new IllegalArgumentException(String.format("Given key %s is of type %d. Please provide only Model and Frame keys.", key.toString(), value.type()));
```

There are no clear rules for how much validation we should use. However, I managed to find one more argument for the approach @michalkurka proposes - only backend knows whether a key is a proper key or nonsense. We even have structures like `KeyV3.java`. 

Any exceptions on the backend are "nicely" propagated to the user anyway, so we're kind of safe here.

valenad1: I removed the check.

My idea was more simple and reflect my experience with removeAll function, when I accidentally pass wrong key and remove all instead of saying "Hey user, you want to retain something I don't know, are you sure?". But it is true that if we cannot do it properly then this "half-way frontend" check could cause hard to reproduce bugs.

michalkurka: > My idea was more simple and reflect my experience with removeAll function, when I accidentally pass wrong key and remove all instead of saying "Hey user, you want to retain something I don't know, are you sure?". But it is true that if we cannot do it properly then this "half-way frontend" check could cause hard to reproduce bugs.

@valenad1 Good thinking, I appreciate you trying to go the extra mile.

We also need to keep in mind that the H2O cluster can easily have tens of thousands of models (eg. in the case of Segment Models). The calling h2o.ls for every single retained key gets expensive.

Generally: my guidance is - validate only using the information you already have. If wouldn't need to ask the backend - let the backend do the validation instead.

michalkurka (approved):
Thank you!

Pscheidl (approved):

PR #5138, opened 2020-11-17
[PUBDEV-7902] GridSearch trains already trained models when loaded in…

honzasterba:
…to new cluster

### DIFF
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -507,7 +507,10 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
           try {
             f.setAccessible(true);
             Object value = f.get(this);
-            if (value != null) {
+            if (value instanceof Enum) {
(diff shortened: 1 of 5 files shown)


satai:
LGTM but I can easily miss some consequences here.

koniecsveta (approved):

h2o-core/src/main/java/hex/Model.java
```diff
@@ -507,7 +507,10 @@ public int compare(Field field1, Field field2) {
           try {
             f.setAccessible(true);
             Object value = f.get(this);
-            if (value != null) {
+            if (value instanceof Enum) {
+              // use string hashcode for enums, since they differ for each run
```
Pscheidl, line 511:
```suggestion
              // use string hashcode for enums, otherwise they're differ for each run
```

Pscheidl (approved):
LGTM completely

honzasterba:
waiting for https://github.com/h2oai/h2o-3/pull/5140/files to fix failing tests

PR #5140, opened 2020-11-18
[GLM] fix bug where GLM model would have null _input_params

honzasterba:
Fixes multiple issues:
- computeParams was not locking and updating the Model instance
- GLM is not releasing lock on Model instance in computeImpl but only in onCompletion, makint the above impossible

Contains also removal of some unnecessary code.

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -225,8 +225,6 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
     }
   }
 
-  static class TooManyPredictorsException extends RuntimeException {}
-
   DataInfo _dinfo;
 
   private transient DataInfo _validDinfo;
@@ -894,14 +892,19 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
 
 
     private void doCleanup() {
-      if(_parms._lambda_search && _parms._is_cv_model)
-        Scope.untrack(removeLater(_dinfo.getWeightsVec()._key));
-      if (_parms._HGLM) {
-        Key[] vecKeys = _toRemove;
-        for (int index=0; index < vecKeys.length; index++) {
-          Vec tempVec = DKV.getGet(vecKeys[index]);
-          tempVec.remove();
+      try {
+        if (_parms._lambda_search && _parms._is_cv_model)
+          Scope.untrack(removeLater(_dinfo.getWeightsVec()._key));
+        if (_parms._HGLM) {
+          Key[] vecKeys = _toRemove;
+          for (int index = 0; index < vecKeys.length; index++) {
+            Vec tempVec = DKV.getGet(vecKeys[index]);
+            tempVec.remove();
+          }
         }
+      } catch (Exception e) {
+        Log.err("Error while cleaning up GLM " + _result);
+        Log.err(e);
       }
     }
 
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -2113,7 +2113,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       try {
         doCompute();
       } finally {
-        _model.unlock(_job);
+        if (_model != null) _model.unlock(_job);
       }
     }
     
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -855,10 +855,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
   private void buildModel() {
     _model = new GLMModel(_result, _parms, this, _state._ymu, _dinfo._adaptedFrame.lastVec().sigma(), _lmax, _nobs);
     _model._output.setLambdas(_parms);  // set lambda_min and lambda_max if lambda_search is enabled
-    
-    // clone2 so that I don't change instance which is in the DKV directly
-    // (clone2 also shallow clones _output)
-    _model.clone2().delete_and_lock(_job._key);
+    _model.delete_and_lock(_job);
   }
 
   protected static final long WORK_TOTAL = 1000000;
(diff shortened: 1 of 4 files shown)


h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -897,38 +894,20 @@ private final double lmax(double[] grad) {
...
-      } catch(Throwable t){
-        // nada
       }
     }
+
     private transient Cholesky _chol;
     private transient L1Solver _lslvr;
 
-
-    int [] findZeros(double [] vals){
```
honzasterba, line 920: unused

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -897,38 +894,20 @@ private final double lmax(double[] grad) {
...
-          }
+      if(_parms._lambda_search && _parms._is_cv_model)
+        Scope.untrack(removeLater(_dinfo.getWeightsVec()._key));
+      if (_parms._HGLM) {
+        Key[] vecKeys = _toRemove;
+        for (int index=0; index < vecKeys.length; index++) {
+          Vec tempVec = DKV.getGet(vecKeys[index]);
+          tempVec.remove();
         }
-        if(_model!=null)
```
honzasterba, line 910: ulocking moved to compute method

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -897,38 +894,20 @@ private final double lmax(double[] grad) {
...
-          _model.unlock(_job);
-      } catch(Throwable t){
```
honzasterba, line 912: this seem reckless, I think we want to know if something breaks during cleanup

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2131,6 +2110,14 @@ protected Submodel computeSubmodel(int i,double lambda, double nullDevTrain, dou
     
     @Override
     public void computeImpl() {
+      try {
+        doCompute();
+      } finally {
+        _model.unlock(_job);
```
michalkurka, line 2116: Can _model be null at this point?

honzasterba: no, it is assigned to in init() and never set to anything else

h2o-algos/src/main/java/hex/glm/GLM.java
michalkurka: I think you have a valid point, however, I think this code needs more error handling if this catch is removed, eg.:

Can _dinfo be null? Can the temp vec already be removed from DKV (line 903 possible NPE).

I propose this: 
1) Add the error handling 
2) in onCompletion - invoke doCleanup the regular way
3) in onExceptionalCompletion - we are in an undefined state - any clean-up attempt is by definition best effort only - I would keep the try-catch - and produce a warning in the log

h2o-algos/src/main/java/hex/glm/GLM.java
michalkurka: I think you have a valid point, however, I think this code needs more error handling if this catch is removed, eg.:

Can _dinfo be null? Can the temp vec already be removed from DKV (line 903 possible NPE).

I propose this:

Add the error handling
in onCompletion - invoke doCleanup the regular way
in onExceptionalCompletion - we are in an undefined state - any clean-up attempt is by definition best effort only - I would keep the try-catch - and produce a warning in the log

h2o-algos/src/main/java/hex/glm/GLM.java
michalkurka: sounds good, @sebhrusen can you also take a look?

h2o-algos/src/main/java/hex/glm/GLM.java
honzasterba: you caught me there, and so did the smoke tests, apperently yes, the `_model` can be null here, fixed

h2o-algos/src/main/java/hex/glm/GLM.java
sebhrusen: `doCleanUp` is called in `onExceptionalCompletion` (typically if job is aborted), so `_model` is always null in this case.

What would be cleaner probably is to do the unlock only in `onCompletion`, right after calling `doCleanUp`

sebhrusen: having it moved after compute looks good to me

honzasterba: actually no, all other algos do unlock in `computeImpl` and this is necessary for the fix in `computeParams` to work and be able to acquire the lock again

h2o-algos/src/main/java/hex/glm/GLM.java
honzasterba: you are correct, such a change is out of scope for this PR, hence I have reverted most of it and at least added some logging

sebhrusen: it was `Throwable` really? at least it should be limited to `Exception`

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: _model is instantiated right after init().  If something goes wrong in init(), _model can be null.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -855,10 +855,7 @@ private GLMModel runGLMModel(boolean standardize, Family family, Link link, Key<
   private void buildModel() {
     _model = new GLMModel(_result, _parms, this, _state._ymu, _dinfo._adaptedFrame.lastVec().sigma(), _lmax, _nobs);
     _model._output.setLambdas(_parms);  // set lambda_min and lambda_max if lambda_search is enabled
-    
-    // clone2 so that I don't change instance which is in the DKV directly
-    // (clone2 also shallow clones _output)
-    _model.clone2().delete_and_lock(_job._key);
+    _model.delete_and_lock(_job);
   }
```
wendycwong, line 857: What is the problem changing the instance in the DKV directly?  Is it going to be slower?  I am just curious why this is done by Tomas.

honzasterba: it does not make any sense to me, thats why I am removing it

koniecsveta (approved):

wendycwong (approved):

PR #5144, opened 2020-11-20
PUBDEV-7737 CoxPH Mojo

satai:


h2o-algos/src/main/java/hex/coxph/CoxPHModel.java
```diff
@@ -361,5 +361,20 @@ protected Futures remove_impl(Futures fs, boolean cascade) {
...
 
+  @Override
+  public CoxPHMojoWriter getMojo() {
+    return new CoxPHMojoWriter(this);
+  }
+
+  @Override
+  public boolean haveMojo() {
+    final boolean hasInteraction = _parms.interactionSpec() != null;
+    final boolean hasStratification = false;//_parms._stratify_by != null && _parms._stratify_by.length > 0;
```
honzasterba, line 372: will be fixed?

h2o-algos/src/main/java/hex/coxph/CoxPHMojoWriter.java
```diff
@@ -0,0 +1,61 @@
...
+    writekv("strata_count", strataMap.size());
+    
+    int strataNum = 0;
+    for (AstGroup.G g : strataMap.keySet()) {
+      writekv("strata_" + strataNum, g._gs);
+      strataNum++;
+    }
+  }
+
+  private void writekv2D_doubles(String key, double[][] array) throws IOException {
```
honzasterba, line 53: functions like these are better in the superclass for reuse

honzasterba: what about `writeDoubleArray`?

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoReader.java
```diff
@@ -0,0 +1,81 @@
...
+    
+    if (0 == count) {
+      return 0;
+    } else {
+      final double[] strata = readkv("strata_0");
+      return strata.length;
+    }
+  }
+
+  private double[][] readkv2D_doubles(String keyPrefix) {
```
honzasterba, line 62: probably also already implemented

h2o-py/h2o/model/coxph.py
```diff
@@ -8,6 +8,7 @@ class H2OCoxPHModel(ModelBase):
 
     def formula(self):
         """Survival formula."""
+        print(self._model_json["output"])
```
honzasterba, line 11: to be removed right?

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoModel.java
```diff
@@ -0,0 +1,75 @@
...
+    int i = 0;
+    for (String coefName : _coef_names) {
+      Integer index = namesInList.indexOf(coefName);
+      _coef_indexes[i++] = index;
+    }
+  }
+
+  @Override
+  public double[] score0(double[] row, double[] predictions) {
+    if (null == _coef_indexes) {
```
honzasterba, line 37: ideally Mojo should be "state-less", any reason this cannot be done in constructor? or some post-init function called by the reader?

h2o-algos/src/main/java/hex/generic/GenericModel.java
```diff
@@ -70,7 +70,7 @@ private static MojoModel reconstructMojo(ByteVec mojoBytes) {
             case WordEmbedding:
                 return unsupportedMetricsBuilder();
             case CoxPH:
-                return unsupportedMetricsBuilder();
+                return new ModelMetricsRegressionCoxPH.MetricBuilderRegressionCoxPH("start", "stop", false, new String[0]);
```
Pscheidl, line 73: Missing `GenericModelTest` tests. Re-using other tests is fine. There are tests for mojo "downloadability" and scoring tests against Java.

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
```diff
@@ -335,4 +335,34 @@ private static Frame rebalanceToAllNodes(Frame fr, Key<Frame> rebalancedKey) {
     return fr;
   }
 
+  @Test
+  public void testJavaScoring() {
```
Pscheidl, line 339: Please also add a test case where the MOJO itself is downloaded and imported back into H2O. This could be tested in the `Generic` model tests.

Pscheidl (changes requested):

h2o-algos/src/main/java/hex/coxph/CoxPHModel.java
satai: I can remove this. Thanks for noticing. 

The stratification was an issue with the original version of the code, it works now.

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoModel.java
satai: Does it https://github.com/h2oai/h2o-3/commit/c87713ffaf45e3862e4cc24dd6aa2d16c2e3d743 solve for you?

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
satai: This is tested in python but I can add a test for in Generic too

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoModel.java
honzasterba: I was thinking more of a package private method used by coxph only, adding public method to GenModel kindof kills the purpose, since you could expect this would get called only once

satai: I was thinking about this before, but `readAll` is private and it calls `readModelData` too early (before `      _model._modelAttributes` is available. And `readFrom` is `static`

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
Pscheidl: Generic is different - The generic algo needs to import the MOJO and do the scoring properly. Which means we need at least Generic tests as well. There is extra logic involved.

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoReader.java
satai: It looks like a funny story... it is probably implemented as `read2DArray` for GAM only :-/
So `writeDoubleArray` exists for all the writers but you can't use it anyway unsles you write your own code (that needs sizes of the array).

h2o-algos/src/main/java/hex/generic/GenericModel.java
satai: Downloadabily is in place..

satai: ...and scoring too

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
satai: As you wish ;-)

Pscheidl (approved):

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoReader.java
honzasterba: it should be refactored to be reusable if possible

satai: Some of the code reused and moved to a more generic `ModelMojoReader`.

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
```diff
@@ -335,4 +335,34 @@ private static Frame rebalanceToAllNodes(Frame fr, Key<Frame> rebalancedKey) {
...
+      parms._response_column = "event";
+      parms._ignored_columns = new String[]{"id", "year", "surgery", "transplant"};
+      parms._ties = CoxPHModel.CoxPHParameters.CoxPHTies.efron;
+      
+      dl = new CoxPH(parms).trainModel().get();
+      Frame res = dl.score(fr);
+      assertTrue(dl.testJavaScoring(fr, res, 1e-5));
+      res.remove();
+    } finally {
+      if (fr != null) fr.delete();
```
honzasterba, line 362: why not use Scope?

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
```diff
@@ -674,6 +676,96 @@ public void downloadable_mojo_glm() throws IOException {
             genericModel.getMojo().writeTo(new FileOutputStream(genericModelMojoFile));
             assertArrayEquals(FileUtils.readFileToByteArray(originalModelMojoFile), FileUtils.readFileToByteArray(genericModelMojoFile));
 
+        } finally {
+        Scope.exit();
```
honzasterba, line 680: indent

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
```diff
@@ -674,6 +676,96 @@ public void downloadable_mojo_glm() throws IOException {
...
+            final GenericModel genericModel = trainAndCheck(generic);
+            Scope.track_generic(genericModel);
+
+            // Compare the two MOJOs byte-wise
+            final File genericModelMojoFile = File.createTempFile("mojo", "zip");
+            genericModel.getMojo().writeTo(new FileOutputStream(genericModelMojoFile));
+            assertArrayEquals(FileUtils.readFileToByteArray(originalModelMojoFile), FileUtils.readFileToByteArray(genericModelMojoFile));
+
+        } finally {
+        Scope.exit();
```
honzasterba, line 724: indent

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
```diff
@@ -674,6 +676,96 @@ public void downloadable_mojo_glm() throws IOException {
...
+
+            final boolean equallyScored = genericModel.testJavaScoring(testFrame, genericModelPredictions, 0);
+            assertTrue(equallyScored);
+
+            final Frame originalModelPredictions = originalModel.score(testFrame);
+            Scope.track(originalModelPredictions);
+            assertTrue(TestUtil.compareFrames(genericModelPredictions, originalModelPredictions));
+
         } finally {
         Scope.exit();
```
honzasterba, line 770: indent

honzasterba (approved):
minor comments

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
satai: Historical reasons, I guess. I will update this through the whole CoxPH test in a separate PR.

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
satai: Fixed.

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
satai: Fixed.

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
satai: Fixed.

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoModel.java
```diff
@@ -0,0 +1,60 @@
...
+      result += row[coefIndex] * _coef[i];
+    }
+    
+    result -= lpBase[strataForRow(row)];
+    
+    predictions[0] = result;
+    return predictions;
+  }
+  
+  private int strataForRow(double[] row) {
```
michalkurka, line 48: we would ideally like to avoid memory allocation when scoring each on this API level, any idea how to avoid this?

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoReader.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Override
+  protected void readModelData() throws IOException {
+    _model._x_mean_cat = readRectangularDoubleArray("x_mean_cat");
+    _model._x_mean_num = readRectangularDoubleArray("x_mean_num");
+    _model._coef = readkv("coef");
+    _model._strata = readStrata();
+    _model._strata_len = readStrataLen();
+    _model._coef_indexes = readkv("coef_indexes");
```
michalkurka, line 22: I don't think this is necessary, you can rely on the order of indices that you saved in the model output

imho no other algos need this

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoModel.java
satai: It depends on how score0 can be called. If I am promised that it happens from a single thread sequentially, it's easy, just move the arrraylist as a prealocated instance field.

In theory I can make an easy to understand data structure to replace `Map<List<Integer>, Integer> _strata;` but I am not sure it's worth it. I would replace an GC friendly allocation + deallocation (there is only one local refference to it with no escape) with some more memory jumping.

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
```diff
@@ -675,7 +677,97 @@ public void downloadable_mojo_glm() throws IOException {
...
+    } 
+    
+    /**
+     * Create a CoxPH model and writes a MOJO into a temporary zip file. Then, it creates a Generic model out of that
+     * temporary zip file and re-downloads the underlying MOJO again. The byte arrays representing both MOJOs are tested
+     * to be the same.
+     *
+     */
+    @Test
+    public void downloadable_mojo_cox_ph() throws IOException {
```
maurever, line 691:
```suggestion
    public void downloadableMojoCoxPh() throws IOException {
```

We are in Java, please. :)

maurever (commented):
Just one nitpicking note, otherwise looks good! Thanks @satai.

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
satai: This style goes through the whole GenericModelTest. I will add it in the PR to clean the test: 

https://github.com/h2oai/h2o-3/pull/5158/files#diff-d01f2061f9cc0c8a9e6f33a7a5e9c77df07a8094ba3dcbfe4205d0123a946139

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoModel.java
satai: @michalkurka Your opinion?

honzasterba: there is no guarantee score0 is used in single-threaded environment, I think our MOJOs are supposed to be thread-safe

honzasterba: I dont think a custom data-structure is necessary, maybe just a custom key object for the hashmap that could be initialized with an "oversized" array but still return correct hashCode and equals result just for the strata sub-array

michalkurka:
@satai can you please answer my question here? https://github.com/h2oai/h2o-3/pull/5144#discussion_r530574776

This is my main concern

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoReader.java
satai: I personally prefer to have this one explicit.

To compute indexes based on the knowledge of strata column counts, coef counts... it's possible and I can do it. But this spreads pieces of knowledge how exactly are the columns ordered to one more place. The code is not obvious and any changes will get more painful. The amount of data stored in mojo when storing `_coef_indexes` seems to be minuscule to me.

So - you are right, `_coef_indexes` are not necessary but hey lead to (IMO) better code here. But if your experience says to change this I can do it as a part of PR.

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
maurever: :+1:

maurever (approved):
LGTM. Thanks @satai.

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoModel.java
```diff
@@ -0,0 +1,96 @@
...
+    super(columns, domains, responseColumn);
+
+  }
+
+  @Override
+  public double[] score0(double[] row, double[] predictions) {
+    double result = 0.0;
+
+    final int size = 0 < _strata.size() ? _strata.size() : 1;
+    double[] lpBase = new double[size];
```
michalkurka, line 68: I do not understand this: why is this calculated for all strata and then use just for one of them?

the backend precalculates these values and then just re-uses them in scoring

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
```diff
@@ -335,4 +335,34 @@ private static Frame rebalanceToAllNodes(Frame fr, Key<Frame> rebalancedKey) {
...
+      CoxPHModel.CoxPHParameters parms = new CoxPHModel.CoxPHParameters();
+      parms._calc_cumhaz = true;
+      parms._train           = fr._key;
+      parms._start_column    = "start";
+      parms._stop_column     = "stop";
+      parms._response_column = "event";
+      parms._ignored_columns = new String[]{"id", "year", "surgery", "transplant"};
+      parms._ties = CoxPHModel.CoxPHParameters.CoxPHTies.efron;
+      
+      dl = new CoxPH(parms).trainModel().get();
```
michalkurka, line 357: why is this called "dl"?

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoModel.java
```diff
@@ -0,0 +1,96 @@
...
+    for (int s = 0; s < size; s++) {
+      for (int i = 0; i < _x_mean_cat[s].length; i++)
+        lpBase[s] += _x_mean_cat[s][i] * _coef[i];
+      for (int i = 0; i < _x_mean_num[s].length; i++)
+        lpBase[s] += _x_mean_num[s][i] * _coef[i + _numStart];
+    }
+    
+    for (int i = 0; i < _coef_indexes.length; i++) {
+      final int coefIndex = _coef_indexes[i];
+      result += row[coefIndex] * _coef[i];
```
michalkurka, line 78: Does this work with categorical values as well? This seems suspicious to me as there will be label indices by default in `row`? Is there a test that shows how does it work with categoricals?

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoReader.java
michalkurka: IMHO coxph is very similar to GLM, we should follow the same pattern - it is hard for me to understand what the coef_indices actually are.

I raised a concern about handling of categorical values above - lets get back to this topic after we have clarity how categoricals are treated.

michalkurka:
@satai I added a test to verify scoring on a dataset with categorical data: https://github.com/h2oai/h2o-3/pull/5144/commits/90e709d9ff70dc68ae1ac446b60e09e793832f0a

It doesn't seem to work. I think this needs to be looked into.

michalkurka (changes requested):
Please look into scoring on datasets with categorical data.

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoModel.java
satai: Was precalculated

satai:
@michalkurka code updated for cathegorical data and added tests for that too

h2o-genmodel/src/main/java/hex/genmodel/algos/coxph/CoxPHMojoModel.java
michalkurka: There is a way to do it relatively easily by implementing a simple hashtable, the Stratas are known in advance and there is a (relatively) a small number of them. You can calculate a hashcode for each strata identifier (composite key), sort them and do a binary-search for an incoming row, then you just compare the identifier elements. This can be encapsulated in a tiny hashtable class.

Your code is also auto(un)boxing the integers, however, this will be optimized.

Just an idea.

michalkurka:
@satai looks good, is there a test that is testing MOJO scoring functionality in python (and not depending on generic model)?

PR #5149, opened 2020-11-23
[GLRM] fix de-serialization of GLRM models

honzasterba:
GLRM model fails to load for multiple reasons:
- asymetric write/load functions
- frame.readAll_impl not working properly

### DIFF
--- a/h2o-py/tests/testdir_algos/glrm/pyunit_glrm_grid_user_y.py
+++ b/h2o-py/tests/testdir_algos/glrm/pyunit_glrm_grid_user_y.py
@@ -36,6 +36,9 @@ def glrm_grid_user_y():
     grid.train(x=train.names, training_frame=train, **params)
     print("first grid")
     print(grid)
+    assert len(grid.model_ids) == 2
+    archetypes1 = grid.models[0].archetypes()
+    archetypes2 = grid.models[1].archetypes()
     grid_path = h2o.save_grid(export_dir, grid.grid_id)
     h2o.remove_all()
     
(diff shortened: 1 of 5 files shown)


h2o-py/tests/testdir_algos/glrm/pyunit_glrm_grid_user_y.py
```diff
@@ -0,0 +1,51 @@
...
+    h2o.remove_all()
+    
+    # reimport and train some more
+    train = h2o.H2OFrame(train_data.tolist())
+    initial_y = h2o.H2OFrame(initial_y_data.tolist(), destination_frame="glrm_initial_y")
+    hyper_params["gamma_x"] = [0.1, 1]
+    grid = h2o.load_grid(grid_path)
+    grid.train(x=train.names, training_frame=train, **params)
+    
+
```
wendycwong, line 47: Can you check the model archetypes between original grid model and retrained grid model to see more training actually occurred?

honzasterba: done

wendycwong: Honza:

I was more thinking like you stored the archetypes after line 37:
archetypes1 = grid.models[0].archetypes()
Then, you run till the add and do 
assert archetypes1 != grid.models[0].archetypes().

Need to consider the same model to check if archetypes have changed due to training.

wendycwong (approved):

PR #5150, opened 2020-11-23
PUBDEV-7903: Add new R functions to the pkgdown R docs website

hannah-tillman:
I went through all of the R docs and added any missing `h2o.<function_name>` functions or functions that had more in depth explanations/examples. 

When building the docs, 5 of the functions "error" and aren't added to the Reference page. This is the warning:
`Warning: In '_pkgdown.yml', topic must match a function or concept.
Problem topic: h2o.feature_interaction` for example.

This problem exists for `h2o.feature_interaction`, `h2o.getFutureModel`, `h2o.list_models`, `h2o.target_encode_fit`, `h2o.target_encode_transform`, and `h2o.topBottomN`. I don't know how to fix this, so any recommendations, @ledell, would be greatly appreciated.

Let me know if you want me to remove any of these or to add any of the missing topics in (I can provide the list from the local build if you want).

https://h2oai.atlassian.net/browse/PUBDEV-7903


jangorecki (commented):
Did you try to make reference section produced without having to list all functions, as explained on slack?

hannah-tillman:
> Did you try to make reference section produced without having to list all functions, as explained on slack?

Yeah, I did try that and it does work! However, it pulls _every_ available function from the /h2o-package/R/ folder and I'm not entirely sure that's what we want. @ledell your thoughts?

jangorecki:
Having that removed will reduce future maintenance as we will not need to modify this file each time we add new function.

I think the better approach would be to group different help topics together into common topic. So instead of having manual page for each single function, to have one manual page for couple related functions. For example see base R's `?Math` manual.

ledell:
If I remember correctly, I think we originally had the functions grouped into categories, but then there was limitation to that approach... though I can't recall what the issue or limitation was.  And that's how we ended up with the manual method (a search of Slack history might unveil some clues about what the issues were...).

I am definitely in favor of having this be automated so we don't have to manually add the new functions.  I would be also be ok to merge this for 3.32.0.3 (if we don't fully resolve the automation aspect before .3 goes out). 

I don't see a major issue with having all the functions (how many total are there compared to the current set, btw?).  If we keep the [R Reference Guide](http://docs.h2o.ai/h2o/latest-stable/h2o-r/docs/reference/index.html) that lists all functions, maybe we can create a second page with the grouping (though I don't know how to do that, personally).

jangorecki:
I think it might have been a roxygen limitation, which personally I would avoid in the first place.

hannah-tillman:
> I don't see a major issue with having all the functions (how many total are there compared to the current set, btw?). If we keep the [R Reference Guide](http://docs.h2o.ai/h2o/latest-stable/h2o-r/docs/reference/index.html) that lists all functions, maybe we can create a second page with the grouping (though I don't know how to do that, personally).

@ledell So, there are currently 53 not included in the current set:
`Warning: Topics missing from index: H2OSegmentModels-class, H2OSegmentModelsFuture-class, Keyed-class, aaa, &&, .addParm, .collapse, .consolidate_varimps, .create_leaderboard, .customized_call, .find_appropriate_column_name, .get_algorithm, .get_feature_count, .get_first_of_family, .h2o.__ALL_CAPABILITIES, .h2o.__CREATE_FRAME, .h2o.__DECRYPTION_SETUP, .h2o.__DKV, .h2o.__EXPORT_FILES, .h2o.__FRAMES, .h2o.__IMPORT, .h2o.__JOBS, .h2o.__LOGANDECHO, .h2o.__MODELS, .h2o.__MODEL_BUILDERS, .h2o.__MODEL_METRICS, .h2o.__PARSE_SETUP, .h2o.__RAPIDS, .h2o.__REST_API_VERSION, .h2o.__SEGMENT_MODELS_BUILDERS, .h2o.__W2V_SYNONYMS, .h2o.__checkConnectionHealth, .h2o.doGET, .h2o.doPOST, .h2o.doRawGET, .h2o.doRawPOST, .h2o.doSafeGET, .h2o.doSafePOST, .h2o.is_progress, .h2o.locate, .h2o.perfect_auc, .h2o.primitives, .has_varimp, .interpretable, .is_h2o_model, .is_h2o_tree_model, .is_plotting_to_rnotebook, .leaderboard_for_row, .min_max, .model_ids, .pkg.env, .plot_varimp, .process_models_or_automl, .shorten_model_i [... truncated]`

If these are all cool to have, I'll just remove the Reference section and that'll be that. As for making a "grouping" page, I think I know how to do that just from looking at the _pkgdown.yml file, but I think that'll have to be its own PR because it'll take a little more time to play around with (so probably by the next release after this one).

However, there's also 5 functions included in the reference list (`h2o.topBottomN`, `h2o.target_encode_transform`, `h2o.target_encode_fit`, `h2o.list_models`, `h2o.getFutureModel`, & `h2o.feature_interaction`) that error and don't build on the Reference page, and I have no clue how to fix those. This is the error: `Warning: In '_pkgdown.yml', topic must match a function or concept.`

ledell:
hey @hannah-tillman I just found this open PR while searching for something else.  I just changed the base to master and fixed a minor conflict in the file.  Maybe we merge this for 3.36?

ledell (approved):
LGTM.

michalkurka (approved):

PR #5151, opened 2020-11-23
PUBDEV-7890: adding checkpointing to GLM for IRLSM

wendycwong:
This PR completes the task in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-7890

I added backend support for GLM to support checkpointing.  However, for Gaussian, it only takes iteration 1 to complete the model building.  There really is no such thing as checkpointing.  I implemented checkpointing for this case is to start building the model again using the coefficients from the last model used as the model_id.

To test checkpoint models for GLM, I do the following:
1. Build a GLM model with max_iteration set to 3.  Then, build a checkpoint model starting from the GLM model with max_iteration of 3.  Call this model modelCheckpoint.
2. Build a GLM model and let it finished.  Call this model modelLong.

For binomial, we are trying to solve a constrained optimization problem.  Hence, the KKT condition will be checked and if it is determined that the algo has not converged, a few more iterations will be added.  If you normally let the algo ran into completion, the algo converges.  But if you stop the model building and it has not arrived at a place where the KKT condition is satisfied, there will be a few more iterations added to it.  Now, if you restart your model building again by using checkpointing, you have already diverged from the normal model process.  The final models modelCheckpoint and modelLong will be equal in this case.  They may be close but they are not equal.

When lambda_search is enabled, there will be multiple submodels in _model._output._submodels.  For a checkpoint model, it will restore the last submodel built and try to continue from there instead of building a brand new submodel.  Hence, we need to know when we are in a checkpoint model for the first time.  I used _checkpointIterFirst to denote this.


h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -696,10 +698,65 @@ public void init(boolean expensive) {
...
+      if (_parms.hasCheckpoint()) {
+        if (!_parms._family.equals(Family.gaussian))  // Gaussian it not iterative and therefore don't care
+          _checkPointFirstIter = true;  // mark the first iteration during iteration process of training
+        if (!_parms._solver.equals(Solver.IRLSM))
+          error("_checkpoint", "GLM checkpoint is supported only for IRLSM.  Please specify it " +
+                  "explicitly.  Do not use AUTO or default");
+        Value cv = DKV.get(_parms._checkpoint);
+        CheckpointUtils.getAndValidateCheckpointModel(this, GLMModel.GLMParameters.CHECKPOINT_NON_MODIFIABLE_FIELDS, cv);
+
+        // setting more parameters if needed
```
honzasterba, line 711: dont understand this comment, probably can be removed

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -696,10 +698,65 @@ public void init(boolean expensive) {
...
       }
+
+      if (_parms.hasCheckpoint()) {
+        if (!_parms._family.equals(Family.gaussian))  // Gaussian it not iterative and therefore don't care
+          _checkPointFirstIter = true;  // mark the first iteration during iteration process of training
+        if (!_parms._solver.equals(Solver.IRLSM))
+          error("_checkpoint", "GLM checkpoint is supported only for IRLSM.  Please specify it " +
+                  "explicitly.  Do not use AUTO or default");
+        Value cv = DKV.get(_parms._checkpoint);
+        CheckpointUtils.getAndValidateCheckpointModel(this, GLMModel.GLMParameters.CHECKPOINT_NON_MODIFIABLE_FIELDS, cv);
```
honzasterba, line 709: CHECKPOINT_NON_MODIFIABLE_FIELDS is empty, hence this does nothing, is that intentional?

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -232,12 +233,22 @@ public void update(double [] beta, double[] ubeta, double devianceTrain, double
     _output.setSubmodelIdx(id);
   }
 
+  protected GLMModel deepClone(Key<GLMModel> result) {
```
honzasterba, line 236: the naming is confusing, deepClone should do what the name says, this does something more, should be named like `deepCopyWithoutMetrics`

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -232,12 +233,22 @@ public void update(double [] beta, double[] ubeta, double devianceTrain, double
...
+    return newModel;
+  }
+
   public static class GLMParameters extends Model.Parameters {
 
     public enum MissingValuesHandling {
       MeanImputation, PlugValues, Skip
     }
-
+    static String[] CHECKPOINT_NON_MODIFIABLE_FIELDS = {};
```
honzasterba, line 251: should be definitelly final, probably private as well

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -4287,9 +4287,13 @@ def assertEqualCoeffDicts(coef1Dict, coef2Dict, tol = 1e-6):
     assert len(coef1Dict) == len(coef2Dict), "Length of first coefficient dict: {0}, length of second coefficient " \
                                              "dict: {1} and they are different.".format(len(coef1Dict, len(coef2Dict)))
     for key in coef1Dict:
-        assert abs(coef1Dict[key]-coef2Dict[key]) < tol, "Coefficient for {0} from first dict: {1}, from second dict:" \
-                                                         " {2} and they are different.".format(key, coef1Dict[key],
-                                                                                               coef2Dict[key])
+        val1 = coef1Dict[key]
+        val2 = coef2Dict[key]
+        if not(math.isnan(val1)) and not(math.isinf(val1)) and not(math.isnan(val2)) and not(math.isinf(val2)):
```
honzasterba, line 4292: should we also test that there is equality in the NaN/Inf case (both NaN, both inf)?

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Honza:

I only implemented checkpointing for solver IRLSM.  Hence, I am leaving this here for future implementation that will use this.

h2o-py/tests/pyunit_utils/utilsPY.py
wendycwong: Yes, I think so.

h2o-algos/src/main/java/hex/glm/GLM.java
honzasterba: that would mean that at least _solver param should be amongst those params that should be the same as the ones in the checkpoint model. are you sure there are not other that should remain the same between the checkpoint and the new model?

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: Honza:

I am following the convention of other models like GBM.  If you want it changed, all models should change as well.  Other models with deepClone also perform other task like in GBM here:

 /** Performs deep clone of given model.  */
  protected M deepClone(Key<M> result) {
    M newModel = IcedUtils.deepCopy(self());
    newModel._key = result;
    // Do not clone model metrics
    newModel._output.clearModelMetrics(false);
    newModel._output._training_metrics = null;
    newModel._output._validation_metrics = null;
    // Clone trees
    Key[][] treeKeys = newModel._output._treeKeys;
    for (int i = 0; i < treeKeys.length; i++) {
      for (int j = 0; j < treeKeys[i].length; j++) {
        if (treeKeys[i][j] == null) continue;
        CompressedTree ct = DKV.get(treeKeys[i][j]).get();
        CompressedTree newCt = IcedUtils.deepCopy(ct);
        newCt._key = CompressedTree.makeTreeKey(i, j);
        DKV.put(treeKeys[i][j] = newCt._key,newCt);
      }
    }
    // Clone Aux info
    Key[][] treeKeysAux = newModel._output._treeKeysAux;
    if (treeKeysAux!=null) {
      for (int i = 0; i < treeKeysAux.length; i++) {
        for (int j = 0; j < treeKeysAux[i].length; j++) {
          if (treeKeysAux[i][j] == null) continue;
          CompressedTree ct = DKV.get(treeKeysAux[i][j]).get();
          CompressedTree newCt = IcedUtils.deepCopy(ct);
          newCt._key = Key.make(createAuxKey(treeKeys[i][j].toString()));
          DKV.put(treeKeysAux[i][j] = newCt._key,newCt);
        }
      }
    }
    return newModel;
  }

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Honza: 

Great suggestion.  I will add it.  Thank you so much!

Wendy

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: Private makes it a pain to call it from GLM.java  I did add final.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -62,6 +63,8 @@
   public int[][] _gamColIndices = null; // corresponding column indices in dataInfo
   public static int _totalBetaLen;
   private boolean _earlyStopEnabled = false;
+//  List<Integer> _scoreIterationList = new ArrayList<Integer>(); // keep track of iteration where scoring occurs
```
honzasterba, line 66: commented out code

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -847,11 +902,81 @@ private GLMModel runGLMModel(boolean standardize, Family family, Link link, Key<
     GLMModel model = new GLM(tempParams).trainModel().get();
     return model;
   }
+  
+  // copy from scoring_history back to _sc or _lsc
```
honzasterba, line 906: comment indicates that the method is poorly named

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -847,11 +902,81 @@ private GLMModel runGLMModel(boolean standardize, Family family, Link link, Key<
...
+  private void reinstallSC() {
```
honzasterba, line 907: what about  reinstallScoringHistory()?

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -847,11 +902,81 @@ private GLMModel runGLMModel(boolean standardize, Family family, Link link, Key<
...
+      _lsc._lambdaDevTest.add((double) sHist.get(rowInd, colIndices[5]));
+    }
+  }
+
+  private void restoreSC(TwoDimTable sHist, int[] colIndices) {
+    DateTimeFormatter fmt = DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss");
+    int numRows = sHist.getRowDim();
+    for (int rowInd = 0; rowInd < numRows; rowInd++) {  // if lambda_search is enabled, _sc is not updated
+      _sc._scoringIters.add((Integer) sHist.get(rowInd, colIndices[0]));
+//      _scoreIterationList.add((Integer) sHist.get(rowInd, colIndices[0]));
```
honzasterba, line 940: commented out code

h2o-algos/src/main/java/hex/glm/GLMModel.java
honzasterba: you're correct

h2o-core/src/main/java/water/util/ArrayUtils.java
```diff
@@ -101,6 +101,7 @@ public static double innerProductPartial(double [] x, int[] x_index, double [] y
     return result;
   }
 
+  // Find index of an element out of a sorted array.
```
honzasterba, line 104: why not javadoc?

h2o-core/src/main/java/water/util/ArrayUtils.java
```diff
@@ -433,6 +434,16 @@ public static double avg(long[] nums) {
     return res;
   }
   
+  // locate element in array arr within error tolerance
```
honzasterba, line 437: why not javadoc?

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -696,10 +698,63 @@ public void init(boolean expensive) {
         _parms._gradient_epsilon = _parms._lambda[0] == 0 ? 1e-6 : 1e-4;
         if(_parms._lambda_search) _parms._gradient_epsilon *= 1e-2;
       }
+
+      if (_parms.hasCheckpoint()) {
+        if (!_parms._family.equals(Family.gaussian))  // Gaussian it not iterative and therefore don't care
```
koniecsveta, line 703: comparison order should be reversed to prevent NPE

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -696,10 +698,63 @@ public void init(boolean expensive) {
...
+          _checkPointFirstIter = true;  // mark the first iteration during iteration process of training
+        if (!_parms._solver.equals(Solver.IRLSM))
```
koniecsveta, line 705: same as above

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -847,11 +902,81 @@ private GLMModel runGLMModel(boolean standardize, Family family, Link link, Key<
...
+      int num2Copy = _parms._HGLM || _parms._lambda_search ? colHeaders2Restore.length : colHeaders2Restore.length-2;
+      int[] colHeadersIndex = grabHeaderIndex(scoringHistory, num2Copy, colHeaders2Restore);
+      if (_parms._lambda_search)
+        restoreLSC(scoringHistory, colHeadersIndex);
+      else
+        restoreSC(scoringHistory, colHeadersIndex);
+  }
+  
+  private void restoreLSC(TwoDimTable sHist, int[] colIndices) {
+    DateTimeFormatter fmt = DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss");
```
koniecsveta, line 923: Formatter should be created only once, it is defined on multiple places in this file. I would suggest creating static constant for it.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -847,11 +902,81 @@ private GLMModel runGLMModel(boolean standardize, Family family, Link link, Key<
...
+      }
+    }
+  }
+  
+  private int[] grabHeaderIndex(TwoDimTable sHist, int numHeaders, String[] colHeadersUseful) {
+    int[] colHeadersIndex = new int[numHeaders];
+    String[] colHeaders = sHist.getColHeaders();
+    for (int colInd = 0; colInd < numHeaders; colInd++) {
+      if (colInd == 0) {
+        int indexFound = Arrays.asList(colHeaders).indexOf(colHeadersUseful[colInd]);
```
koniecsveta, line 956: `Arrays.asList(colHeaders)` could be a local variable since it is used in both if branches

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2007,7 +2138,7 @@ private void scoreAndUpdateModel(){
     }
     
     private void scorePostProcessing(Frame train, long t1) {
-      _scoreIterationList.add(_state._iter);
+//      _scoreIterationList.add(_state._iter);
```
koniecsveta, line 2141: commented out code

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2067,15 +2198,27 @@ private void coldStart(double[] devHistoryTrain, double[] devHistoryTest) {
...
-        _model.addSubmodel(sm = new Submodel(lambda, _state.alpha(), getNullBeta(),_state._iter,nullDevTrain, nullDevValid));
-      else {  // this is also the path for HGLM model
-        sm = new Submodel(lambda, _state.alpha(), _state.beta(), _state._iter, -1, -1);// restart from last run
-         if (_parms._HGLM) // add random coefficients for random effects/columns
-          sm.ubeta = Arrays.copyOf(_state.ubeta(), _state.ubeta().length);
-        _model.addSubmodel(sm);
-        if (!_parms._HGLM) // only perform this when HGLM is not used.
-          _state.setLambda(lambda);
+      if(lambda >= _lmax && _state.l1pen() > 0) {
+        if (_parms.hasCheckpoint() && (_parms._alpha.length > 1 || _parms._lambda.length > 1) &&
```
koniecsveta, line 2202: - this condition is the same for both branches, it would be more readable if it would be extracted to a local variable

- comparison order should be reversed for constants to prevent NPE

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Done.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: I renamed the function and the variables.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Yeah, that is better.  Thanks.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Removed.

h2o-core/src/main/java/water/util/ArrayUtils.java
wendycwong: Fixed.

h2o-core/src/main/java/water/util/ArrayUtils.java
wendycwong: Done.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Good catch.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Good point.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Removed.

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -522,7 +522,7 @@ protected boolean checkKKTs() {
       return checkKKTsMultinomial();
     double [] beta = _beta;
     double [] u = _u;
-    if(_activeData._activeCols != null) {
+    if(_activeData._activeCols != null && (_beta.length < (_dinfo.fullN()+1))) {
```
honzasterba, line 525: this condition is duplicated twice, from the code is unclear when does this happen, perhaps putting this check into a method and giving it a name would clear things up

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -51,7 +52,8 @@
 public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
   static NumberFormat lambdaFormatter = new DecimalFormat(".##E0");
   static NumberFormat devFormatter = new DecimalFormat(".##");
-
+  private static final DateTimeFormatter fmt = DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss");
```
honzasterba, line 55: containts in Java should be ALL_CAPS_WITH_UNDESCORES

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -696,10 +693,63 @@ public void init(boolean expensive) {
...
       }
+
+      if (_parms.hasCheckpoint()) {
+        if (!Family.gaussian.equals(_parms._family))  // Gaussian it not iterative and therefore don't care
+          _checkPointFirstIter = true;  // mark the first iteration during iteration process of training
+        if (!Solver.IRLSM.equals(_parms._solver))
+          error("_checkpoint", "GLM checkpoint is supported only for IRLSM.  Please specify it " +
+                  "explicitly.  Do not use AUTO or default");
+        Value cv = DKV.get(_parms._checkpoint);
+        CheckpointUtils.getAndValidateCheckpointModel(this, _parms.CHECKPOINT_NON_MODIFIABLE_FIELDS, cv);
```
honzasterba, line 704: referecing a static field from a non-static context, you can do just a static import here, much nicer

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -696,10 +693,63 @@ public void init(boolean expensive) {
...
+      }
       buildModel();
     }
   }
 
+  // copy over parameters from _model to _state for checkpointing
+  // jest of this method is to restore the _state to be the same as before
+  private void copyCheckModel2StateP1() {
```
honzasterba, line 712: could be more english `copyCheckpointModelToStateP1`
what does the `P1` mean? is it `part 1`, I dont see a part two.
also, since this is about modifying ComputationState, would the method be better placed in something like `ComputationState.restoreFromCheckpoint`? such change would allow you to unit-test this method

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -847,11 +897,78 @@ private GLMModel runGLMModel(boolean standardize, Family family, Link link, Key<
...
+              "convergence"};
+      int num2Copy = _parms._HGLM || _parms._lambda_search ? colHeaders2Restore.length : colHeaders2Restore.length-2;
+      int[] colHeadersIndex = grabHeaderIndex(scoringHistory, num2Copy, colHeaders2Restore);
+      if (_parms._lambda_search)
+        restoreLSC(scoringHistory, colHeadersIndex);
+      else
+        restoreSC(scoringHistory, colHeadersIndex);
+  }
+  
+  private void restoreLSC(TwoDimTable sHist, int[] colIndices) {
```
honzasterba, line 917: also probably more suited to be inside `LambdaSearchScoringHistory.restoreFromChecpoint` with a unit test

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -847,11 +897,78 @@ private GLMModel runGLMModel(boolean standardize, Family family, Link link, Key<
...
+      _lambdaSearchScoringHistory._scoringTimes.add(fmt.parseMillis((String) sHist.get(rowInd, colIndices[1])));
+      _lambdaSearchScoringHistory._lambdaIters.add((int) sHist.get(rowInd, colIndices[0]));
+      _lambdaSearchScoringHistory._lambdas.add(Double.valueOf((String) sHist.get(rowInd, colIndices[2])));
+      _lambdaSearchScoringHistory._lambdaPredictors.add((int) sHist.get(rowInd, colIndices[3]));
+      _lambdaSearchScoringHistory._lambdaDevTrain.add((double) sHist.get(rowInd, colIndices[4]));
+      _lambdaSearchScoringHistory._lambdaDevTest.add((double) sHist.get(rowInd, colIndices[5]));
+    }
+  }
+
+  private void restoreSC(TwoDimTable sHist, int[] colIndices) {
```
honzasterba, line 929: also probably better moved to `ScoringHistory.restoreFromCheckpoint`

h2o-algos/src/main/java/hex/glm/GLM.java
honzasterba: note removed

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1418,27 +1533,35 @@ private void fitLSM(Solver s){
...
-          if (!_state._lsNeeded && (Double.isNaN(gram.likelihood) || _state.objective(gram.beta, gram.likelihood) > _state.objective() + _parms._objective_epsilon)) {
+          if ((!_state._lsNeeded && (Double.isNaN(gram.likelihood) || _state.objective(gram.beta, gram.likelihood) > 
+                  _state.objective() + _parms._objective_epsilon)) && !_checkPointFirstIter) {
             _state._lsNeeded = true;
           } else {
-            if (!firstIter && !_state._lsNeeded && !progress(gram.beta, gram.likelihood)) {
-              Log.info("DONE after " + (iterCnt-1) + " iterations (1)");
+            if (!firstIter && !_state._lsNeeded && !progress(gram.beta, gram.likelihood) && !_checkPointFirstIter) {
+              Log.info("DONE after " + (iterCnt - 1) + " iterations (1)");
+              _model._betaCndTemp = s == Solver.COORDINATE_DESCENT ? COD_solve(gram, _state._alpha, _state.lambda())
```
honzasterba, line 1552: will this impact performance? this computation is done elswhere as well, but the result computed here is not used, could it be done so this is computed only once?

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1101,6 +1113,7 @@ public Submodel(double lambda , double alpha, double [] beta, int iteration, dou
   public final long    _nullDOF;
   public final double    _ySigma;
   public final long      _nobs;
+  public double[] _betaCndTemp;  // store temporary beta coefficients for checkpointing purposes
```
honzasterba, line 1116: confusing name, `temp` indicates something volatile, that is to be thrown away, what about `betaCoefsCheckpoint`?

h2o-algos/src/main/java/hex/glm/ComputationState.java
michalkurka: @wendycwong can you please check if this condition is correct also for the case when the intercept is disabled (should it always be fullN + 1)?

wendycwong: Honza:  Will do .

Michalk:  Yes, will check as well.

wendycwong: @michalk:  dinfo.fullN() will return the number of predictors.  So, that will not change if you choose to have intercept or not.  When _parms._intercept = false, this will not change the number of coefficients.  It will still be number of predictors + intercept.  However, the intercept value will be zero.  Therefore, the condition is correct regardless of _parms._intercept = true or false.

wendycwong: However, when we denormalize the coefficients, the intercept may not be zero and this is correct.

wendycwong: @honzasterba:  I removed the first condition.

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: @honzasterba 

I changed the name to betaCndCheckpoint because it stored the last betaCnd.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: This is done only once when the fitting is completed.  Yes, I think it will add one ADMM.solve.  I will make check the performance speed in the benchmarking.

wendycwong: It will add one more ADMM_solve as it is done when it exits.  Will add performance test on my branch to measure the effect.

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: Changed name to _betaCndCheckpoint to store the latest state of _betaCnd

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: added junit test.

maurever (approved):
LGTM, thanks @wendycwong .

wendycwong:
Performance test:  I have compared this branch and rel-zermelo and here are the results.  Positive change means increase in time while negative change means decrease in time:



image

wendycwong:
Here is the performance with more runs:
image

PR #5152, opened 2020-11-24
PUBDEV-7916 - Expose single quotes setting on import/upload file

Pscheidl:
https://h2oai.atlassian.net/browse/PUBDEV-7916

There is a setting to manually switch the CSV parser to single-quoted regime, however it is not exposed to the user. It should be exposed and turned off by default to keep backwards compatibility.


Pscheidl:
Tests use new datasets, just triggered sync. Might fail initially and need to be re-run.

h2o-py/h2o/frame.py
```diff
@@ -449,7 +451,7 @@ def _parse_raw(self, setup):
         p = {"destination_frame": None,
              "parse_type": None,
              "separator": None,
-             "single_quotes": None,
+             "single_quotes": setup["single_quotes"],
```
michalkurka, line 454: Why is this necessary? Line 467 should already handle that

    p.update({k: v for k, v in viewitems(setup) if k in p})

Doesn't it?

Pscheidl: Yes this is redundant.

michalkurka:
@Pscheidl I am not sure if exposing the single quote parameter is the most natural thing to do.

Pandas API uses `quotechar` - which makes a lot of sense to me, see
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html

We could do the same thing, but restrict the allowed values only to a single quote and double quote. Java backend can keep using single_quote parameters as before.

Another reason why not to expose directly is Pandas parameter `doublequote` which is similar to yours but has a different interpretation.

michalkurka:
I checked what `datatable` does and they seem to have also a restricted set of input characters:

quotechar: '"' | "'" | "`"
    The character that was used to quote fields in the CSV file. By
    default the double-quote mark `'"'` is assumed.

https://github.com/h2oai/datatable/blob/49798fa551cc790b513d4d52e662cfd8e612f479/src/core/read/py_fread.cc#L160

Pscheidl:
This was the quickest  way to do it possible.  The tests are still valid.
From RFC POV, only single and double quotes are accepted.

In practice, the RFC is not so widely adapted and there are lots of weird CSVs with tabs as delimiters and whatever on ASCII table as token enclosing char. 

I believe we'll pay not that big price by switching to char specification. The state machine itself is not adapted for that and will have to be adjusted. No too much work, but needs to be done with caution. After that, we might send this switch to the depths of hell. Also, there is some foundation of tests we can rely on due to our activities in the past. 

What you say is a more generic solution, leaving doors open. Leaving the switch in place would require either deprecation or checks for mutual exclusivity with the generic parameter, if added any time in future.

sebhrusen:
@Pscheidl @michalkurka 
I also prefer somethling like `quotechar`.
Do we think it would be possible to have this set to 'AUTO' by default?

Ideally we should be able to detect single or double quote in most cases, in which case 'AUTO' makes sense, and this still leaves the possibility for the user to force the quote character is the parser detected wrong.

`pandas` does do autodetect, but the python Arff parser does, I'll try to find how it does it.

michalkurka:
> What you say is a more generic solution, leaving doors open. Leaving the switch in place would require either deprecation or checks for mutual exclusivity with the generic parameter if added any time in future.

To be clear - I am suggesting this: 
- adapt client API to use a generic Pandas-style parameter that will only accept double quotes and single quotes,
- convert the external client API to internal backend API when making the REST call,
- leave Java backend as-is without any modifications.

I don't know if this is a lot of work on the client-side but the advantage is clearer user API consistent with Pandas.

sebhrusen:
Also, please note that just having single quote vs double quote exposed to client is very small use for the benchmarks...
If at least H2O would fail when trying to load an arff/csv file with the wrong quote character, then we could just try both, but here the parsing looks ok at first sight, so no reason to try a different quote char.

Pscheidl:
Doing only the client part is even easier. 

@sebhrusen  - auto is already there and will be default if no char is specified.

Pscheidl:
Done as @michalkurka said, please review.

h2o-py/h2o/frame.py
sebhrusen: reminder

h2o-py/h2o/h2o.py
```diff
@@ -386,18 +387,20 @@ def upload_file(path, destination_frame=None, header=0, sep=None, col_names=None
     assert_is_type(col_names, [str], None)
     assert_is_type(col_types, [coltype], {str: coltype}, None)
     assert_is_type(na_strings, [natype], {str: natype}, None)
+    assert_is_type(quotechar, str, None)
```
sebhrusen, line 390:
```suggestion
    assert_is_type(quotechar, None, U("'", '"'))
```

h2o-py/h2o/h2o.py
```diff
@@ -457,6 +461,7 @@ def import_file(path=None, destination_frame=None, parse=True, header=0, sep=Non
     assert_is_type(col_types, [coltype], {str: coltype}, None)
     assert_is_type(na_strings, [natype], {str: natype}, None)
     assert_is_type(partition_by, None, [str], str)
+    assert_is_type(quotechar, str, None)
```
sebhrusen, line 464:
```suggestion
    assert_is_type(quotechar, None, U("'", '"'))
```

h2o-py/h2o/h2o.py
```diff
@@ -759,13 +766,18 @@ def parse_setup(raw_frames, destination_frame=None, header=0, separator=None, co
     assert_is_type(column_types, [coltype], {str: coltype}, None)
     assert_is_type(na_strings, [natype], {str: natype}, None)
     assert_is_type(partition_by, None, [str], str)
+    assert_is_type(quotechar, str, None)
```
sebhrusen, line 769:
```suggestion
    assert_is_type(quotechar, None, U("'", '"'))
```

h2o-py/h2o/h2o.py
```diff
@@ -759,13 +766,18 @@ def parse_setup(raw_frames, destination_frame=None, header=0, separator=None, co
...
     check_frame_id(destination_frame)
 
     # The H2O backend only accepts things that are quoted
     if is_type(raw_frames, str): raw_frames = [raw_frames]
 
+    if quotechar not in ["\"", "'", None]:
+        raise ValueError("Only single quotes (') and double quotes (\") are allowed.")
```
sebhrusen, line 776: unnecessary if using my suggestion with assert

h2o-py/h2o/h2o.py
```diff
@@ -759,13 +766,18 @@ def parse_setup(raw_frames, destination_frame=None, header=0, separator=None, co
...
+
     # temporary dictionary just to pass the following information to the parser: header, separator
-    kwargs = {"check_header": header, "source_frames": [quoted(frame_id) for frame_id in raw_frames]}
+    kwargs = {"check_header": header, "source_frames": [quoted(frame_id) for frame_id in raw_frames],
+              "single_quotes": True if quotechar == "'" else False}
```
sebhrusen, line 780: ```True if quotechar == "'" else False``` = ```quotechar == "'"```

h2o-py/h2o/h2o.py
```diff
@@ -436,6 +439,7 @@ def import_file(path=None, destination_frame=None, parse=True, header=0, sep=Non
         directory.
     :param skipped_columns: an integer list of column indices to skip and not parsed into the final frame from the import file.
     :param custom_non_data_line_markers: If a line in imported file starts with any character in given string it will NOT be imported. Empty string means all lines are imported, None means that default behaviour for given format will be used
+    :param quotechar: A hint for the parser which character to expect as quoting character. None (default) means autodetection.
```
sebhrusen, line 442: can we add to those docstrings that it accepts only None (default=Auto), single quote, double quote?

h2o-r/h2o-package/R/import.R
```diff
@@ -78,23 +79,25 @@
...
+                   custom_non_data_line_markers=custom_non_data_line_markers, partition_by, quotechar)
 }
 
 
 #' @rdname h2o.importFile
 #' @export
 h2o.importFolder <- function(path, pattern = "", destination_frame = "", parse = TRUE, header = NA, sep = "",
                              col.names = NULL, col.types=NULL, na.strings=NULL, decrypt_tool=NULL, skipped_columns=NULL,
-                             custom_non_data_line_markers=NULL, partition_by=NULL) {
+                             custom_non_data_line_markers=NULL, partition_by=NULL, quotechar=NULL) {
```
sebhrusen, line 93: in R, I think we can directly enforce supported values in the signature:
```
quotechar=c("'", '"')
```

h2o-py/h2o/h2o.py
Pscheidl: Is this Py2 compatible ?

h2o-py/h2o/h2o.py
Pscheidl: You probably mean `assert_is_type(quotechar,None, U("'", '"'))`

h2o-r/h2o-package/R/import.R
Pscheidl: 1. Should be `c(NULL, "'", '"')`, provided this was possible.
2. This sets the default to the vector of these elements, nothing else. No validation effect observed.

Pscheidl:
@sebhrusen IDK what this reminded is supposed to mean: https://github.com/h2oai/h2o-3/pull/5152#discussion_r531623244

The parse setup phase decides on the usage of single or double quotes (or is given the instructions from the user), but the results from that phase must be carried over to the parsing phase.

h2o-py/h2o/h2o.py
Pscheidl: added

h2o-py/h2o/h2o.py
Pscheidl: changed

h2o-py/h2o/h2o.py
Pscheidl: removed

h2o-py/h2o/h2o.py
Pscheidl: used

h2o-py/h2o/h2o.py
Pscheidl: used

h2o-py/h2o/h2o.py
Pscheidl: used

h2o-py/h2o/frame.py
Pscheidl: I guess I misunderstood this. This is to carry the results from parseSetup phase done by `ParseSetupHandler` to the actual parsing phase handled by ParseHandler.

h2o-r/h2o-package/R/import.R
sebhrusen: yes sorry, validation is not done automatically, you need to apply `match.arg`.
Anyway, this is idiomatic R apparently (we use it in various places in the code
```
> fun <- function(p=c("foo", "bar")) {
+     p <- match.arg(p)
+     p
+ }
> fun("foo")
[1] "foo"
> fun(NULL)
[1] "foo"
> fun("baz")
Error in match.arg(p) : 'arg' should be one of “foo”, “bar”
```

sebhrusen: so in this case, you can just put the default valud first in the propositions

Pscheidl: Sorry, doesn't work with `NULL` in the vecr for me:

```
h2o.importFile <- function(path, destination_frame = "", parse = TRUE, header=NA, sep = "", col.names=NULL,
                           col.types=NULL, na.strings=NULL, decrypt_tool=NULL, skipped_columns=NULL,
                           custom_non_data_line_markers=NULL, partition_by=NULL, quotechar=c(NULL, "'", '"')) {
  quotechar <- match.arg(quotechar)
```

It will automatically use the first non-null item.

```
> h2o.importFile(path = "/home/pavelp/Downloads/single_quotes_mixed.csv")
Called from: eval(expr, p)
Browse[1]> n
debug at ~/projects/h2o-3/h2o-r/h2o-package/R/import.R#84: h2o.importFolder(path, pattern = "", destination_frame = destination_frame, 
    parse, header, sep, col.names, col.types, na.strings = na.strings, 
    decrypt_tool = decrypt_tool, skipped_columns = skipped_columns, 
    custom_non_data_line_markers = custom_non_data_line_markers, 
    partition_by, quotechar)
Browse[2]> quotechar
[1] "'"
```

Screenshot from 2020-11-27 16-28-29

If that's a big problem, can we stick to the form of validation that is already present there and the rest of the method uses it ? I'm not against improvements, but this doesn't seem to be very functional and the rest of the surrounding code uses the same validation anyway.

sebhrusen: you don't need `NULL` in the `c()`, you can just put the double quote (default) first in the suggestions

Pscheidl: I do need NULL because that's the maybe-in-future-to-be-done AUTO mode ? Sure, I don't need it RIGHT NOW, but then it would have to be re-written in future anyway. Which it probably won't, so you're technically correct.

sebhrusen: well, if one day we have an "auto" mode, we can always call it "auto" instead of NULL and put it first, as NULL will automatically be converted to "auto"...

Pscheidl: Which removes the possibility to use "auto" as a quote :D Sure, we might do that. I was just sticking to the defaulte behavior. I don't care. We might adjust the Python client as well.

Pscheidl: Still, the parameter is named `quotechar`. Logically, when quotechar is set to anything, it means the quotechar is that thing set. If it's null, it means no quotechar is set explicitly and the system should decide. But my logic might be too exact.

sebhrusen: I don't want to discuss the sex of angels here, as I think that both approaches (NULL/None vs. 'auto') are valid given that no one would seriously consider 'auto', first as a char, second as a delimiter, without mentioning that the method documentation explicitely mentions which characters are allowed.

My point is just about using idiomatic R when possible and h2o code tends to use `c(...)`-style of param whenever the value choices are predefined.

Pscheidl: Leaving that to None in Python and using AUTO in R makes most sense to me then.

sebhrusen (approved):
LG, thanks for this fix

michalkurka (approved):
thank you for implementing the suggestions

PR #5153, opened 2020-11-24
Enable providing metalearner parameters for NaiveBayes and XGBoost

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7907

Passing parameters for NaiveBayes and XGBoost metalearner wasn't implemented.

### DIFF
--- a/h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
+++ b/h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
@@ -4,6 +4,7 @@ import com.google.gson.reflect.TypeToken;
 import hex.ensemble.Metalearner.Algorithm;
 import hex.ensemble.StackedEnsemble;
 import hex.ensemble.StackedEnsembleModel;
+import hex.naivebayes.NaiveBayesModel;
 import hex.tree.gbm.GBMModel;
 import hex.tree.drf.DRFModel;
 import hex.deeplearning.DeepLearningModel;
@@ -22,6 +23,7 @@ import water.api.schemas3.FrameV3;
 import com.google.gson.Gson;
 import water.fvec.Frame;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
@@ -148,8 +150,8 @@ public class StackedEnsembleV99 extends ModelBuilderSchema<StackedEnsemble,Stack
           }
         }
         
-        ModelParametersSchemaV3 paramsSchema;
-        Model.Parameters params;
+        ModelParametersSchemaV3 paramsSchema = null;
+        Model.Parameters params = null;
         switch (metalearner_algorithm) {
           case AUTO:
           case glm:
@@ -170,6 +172,32 @@ public class StackedEnsembleV99 extends ModelBuilderSchema<StackedEnsemble,Stack
           case deeplearning:
             paramsSchema = new DeepLearningV3.DeepLearningParametersV3();
             params = new DeepLearningModel.DeepLearningParameters();
+            break;
+          case naivebayes:
+            paramsSchema = new NaiveBayesV3.NaiveBayesParametersV3();
+            params = new NaiveBayesModel.NaiveBayesParameters();
+            break;
+          case xgboost:
+            try {
+              ClassLoader classLoader = getClass().getClassLoader();
+              paramsSchema = (ModelParametersSchemaV3)  classLoader
+                      .loadClass("hex.schemas.XGBoostV3$XGBoostParametersV3")
+                      .getConstructor()
+                      .newInstance();
+              params = (Model.Parameters) classLoader
+                      .loadClass("hex.tree.xgboost.XGBoostModel$XGBoostParameters")
+                      .getConstructor()
+                      .newInstance();
+            } catch (ClassNotFoundException |
+                     NoSuchMethodException |
+                     IllegalAccessException |
+                     InstantiationException |
+                     InvocationTargetException e) {
+              e.printStackTrace();
+            }
+            if (paramsSchema == null || params == null) {
+              throw new UnsupportedOperationException("Cannot use XGBooost metalearner as the XGBoost extension is not loaded.");
+            }
             break;
           default:
             throw new UnsupportedOperationException("Unknown meta-learner algo: " + metalearner_algorithm);
--- a/h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_metalearner.py
+++ b/h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_metalearner.py
@@ -66,7 +66,7 @@ def stackedensemble_metalearner_test():
         if meta_params:
             assert(se.params['metalearner_nfolds']['actual'] == 3)
 
-        meta = h2o.get_model(se.metalearner()['name'])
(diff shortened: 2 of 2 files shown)


h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
```diff
@@ -171,6 +173,32 @@ public StackedEnsembleParametersV99 fillFromImpl(StackedEnsembleModel.StackedEns
             paramsSchema = new DeepLearningV3.DeepLearningParametersV3();
             params = new DeepLearningModel.DeepLearningParameters();
             break;
+          case naivebayes:
+            paramsSchema = new NaiveBayesV3.NaiveBayesParametersV3();
+            params = new NaiveBayesModel.NaiveBayesParameters();
+            break;
+          case xgboost:
```
tomasfryda, line 180: @sebhrusen Not sure how to solve this: XGBoost is an extension so I think we should import it dynamically but the solution isn't the prettiest. Another solution that occurred to me is to (ab)use the `hex.ensemble.Metalearners.LocalProvider` and add two factory method to return `ModelParametersSchemaV3` and `Model.Parameters`.

Or do you have better ideas how to solve this issue?

sebhrusen: yeah, my fault, I forgot the parameters schema when creating this provider.
I still prefer to have a single entry point to load this dynamically, and `Metalearners` makes sense here.

for `params`, you can already use `Metalearners.createParameters()`:you just need to make it public.
for the schema, I think it's ok to extend the `LocalProvider` with a method instantiating schemas, without being a complete abuse :) Another solution would be to use reflection as all model schemas follow the same convention: `hex.schemas.MyModelV3.MyModelParametersV3` which will work, except potentially for users that would want to plug their own algo...

h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
```diff
@@ -171,6 +173,32 @@ public StackedEnsembleParametersV99 fillFromImpl(StackedEnsembleModel.StackedEns
...
+                      .loadClass("hex.tree.xgboost.XGBoostModel$XGBoostParameters")
+                      .getConstructor()
+                      .newInstance();
+            } catch (ClassNotFoundException |
+                     NoSuchMethodException |
+                     IllegalAccessException |
+                     InstantiationException |
+                     InvocationTargetException e) {
+              e.printStackTrace();
+            }
```
sebhrusen, line 197: oh I see, you tried with reflection.
Yeah, the `Metalearners` approach is definitely more elegant.

btw, no `e.printStackTrace()`, use logger instead.

h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
```diff
@@ -148,32 +151,8 @@ public StackedEnsembleParametersV99 fillFromImpl(StackedEnsembleModel.StackedEns
...
-        ModelParametersSchemaV3 paramsSchema;
-        Model.Parameters params;
-        switch (metalearner_algorithm) {
-          case AUTO:
-          case glm:
-            paramsSchema = new GLMV3.GLMParametersV3();
-            params = new GLMModel.GLMParameters();
-            // FIXME: This is here because there is no Family.AUTO. It enables us to know if the user specified family or not.
-            // FIXME: Family.AUTO will be implemented in https://0xdata.atlassian.net/projects/PUBDEV/issues/PUBDEV-7444
-            ((GLMModel.GLMParameters) params)._family = null;
```
tomasfryda, line 160: This was supposed to be fixed in different PR[1] but I forgot about it and it is handy to have here fixed already so I added the fix to this PR.

[1] https://github.com/h2oai/h2o-3/pull/4760 (now CLOSED PR)

h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_metalearner.py
```diff
@@ -66,7 +66,7 @@ def train_ensemble_using_metalearner(algo, expected_algo):
         if meta_params:
             assert(se.params['metalearner_nfolds']['actual'] == 3)
 
-        meta = h2o.get_model(se.metalearner()['name'])
+        meta = h2o.get_model(se.metalearner().model_id)
```
tomasfryda, line 69: unrelated to this PR, just to remove deprecation warning since the `metalearner()` is not a dictionary anymore and throws deprecation warning when using as dictionary.

h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java
tomasfryda: Thanks!

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -407,7 +402,7 @@ boolean inferDistributionOrFamily(Model aModel) {
     if (Metalearners.getActualMetalearnerAlgo(_parms._metalearner_algorithm) == Metalearner.Algorithm.glm) { //use family
       // FIXME: This is here because there is no Family.AUTO. It enables us to know if the user specified family or not.
```
michalkurka, line 403: @tomasfryda can you please also delete the obsolete comment?

michalkurka (approved):
tiny comment - otherwise looks good to me, thank you!

h2o-algos/src/main/java/hex/ensemble/MetalearnerProvider.java
```diff
@@ -1,10 +1,13 @@
 package hex.ensemble;
 
 import hex.Model.Parameters;
+import water.api.schemas3.ModelParametersSchemaV3;
 
-public interface MetalearnerProvider<M extends Metalearner> {
+public interface MetalearnerProvider<M extends Metalearner, MPS extends ModelParametersSchemaV3> {
```
sebhrusen, line 6: I'm a bit allergic to those dependency loops:
- model API depends on schema API
- schema API depends on model API

and even worse when it's used in generic parameters because it will propagate those dependency loops like a virus…

At the minimum, I would remove it from the generic signature to avoid propagation: we abuse generics beyond necessity sometimes in this project to the point where we often struggle against the compiler to get things right, which should NOT be the purpose of generics (I blame the design of generics itself for that): we should generify code when it becomes useful instead of complexifying it from the beginning (I take my part of responsibility in this…).

Here, we just need `ModelParametersSchemaV3` (technically, we even just need `Schema`, which is in `core` module, that makes it a more tolerable loop): I don't see any reason to ask for more.

h2o-algos/src/main/java/hex/ensemble/Metalearners.java
```diff
@@ -48,11 +50,16 @@ static Algorithm getActualMetalearnerAlgo(Algorithm algo) {
         return algo == Algorithm.AUTO ? Algorithm.glm : algo;
     }
 
-    static Model.Parameters createParameters(String name) {
+    static public Model.Parameters createParameters(String name) {
```
sebhrusen, line 53: the usual convention in java (dating from Sun Microsystems) is to use modifiers in the following order:
```
[public | protected | private] [static] [final | transient] [volatile] [abstract] …
```
Having everyone using this convention makes it easier to read.

PR #5155, opened 2020-11-24
Publish only Python 3.6 and later to Conda

michalkurka:
There is an issue with certifi package that cannot be installed for
2.7 and 3.5 using the conda-forge channel.

A simple fix would be to enable another channel that has the package for
earlier version. Instead of doing that, we decided we will no longer
deploy the older python versions to Conda - they don't seem to be used
and are thus just taking space.

Successful RC build: http://mr-0xc1:8080/view/H2O-3/job/h2o-3-release-pipeline/job/rel-zermelo-rc2/15/

### DIFF
--- a/scripts/jenkins/jenkinsfiles/Jenkinsfile-Release
+++ b/scripts/jenkins/jenkinsfiles/Jenkinsfile-Release
@@ -463,9 +463,10 @@ EOF
                     if (params.BUILD_CONDA) {
 
                         final def condaS3Dir = "${env.S3_ROOT}/${env.BRANCH_NAME}/${currentBuild.number}/Python/Conda"
+                        final def pyVersions = pipelineContext.getBuildConfig().PYTHON_VERSIONS.findAll { Double.parseDouble(it) >= 3.6 }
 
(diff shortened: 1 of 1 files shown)


sebhrusen (approved):
thx!

scripts/jenkins/jenkinsfiles/Jenkinsfile-Release
```diff
@@ -463,9 +463,10 @@ EOF
                     if (params.BUILD_CONDA) {
 
                         final def condaS3Dir = "${env.S3_ROOT}/${env.BRANCH_NAME}/${currentBuild.number}/Python/Conda"
+                        final def pyVersions = pipelineContext.getBuildConfig().PYTHON_VERSIONS.findAll { Double.parseDouble(it) >= 3.6 }
```
honzasterba, line 466: knit: what about putting "3.6" into buildConfig.groovy as something like "CONDA_PYTHON_MIN_VERSION"?

honzasterba (approved):
have minor comment

PR #5157, opened 2020-11-25
[PUBDEV-7921] fix parent dir lookup for hdfs grid imports

honzasterba:


Pscheidl (approved):
Thanks. This is the change we've been talking about earlier. I didn't use Persist but File here.

PR #5159, opened 2020-11-25
[GRID] fix failing glrm test on py2.7

honzasterba:
- unicode strings from json parsing make restarting a grid with generated hyper-params impossible
  - fix this by implementing better serialization for gridsearch params
  - add specific test for gridsearch
- the test was passing only accidentally on py3 since the same grid was built again due to training frame having different key each time
  - fixed by using pre-defined key for training frame

### DIFF
--- a/h2o-core/src/main/java/water/api/GridSearchHandler.java
+++ b/h2o-core/src/main/java/water/api/GridSearchHandler.java
@@ -204,10 +204,8 @@ public class GridSearchHandler<G extends Grid<MP>,
         fields.add(name);
       } catch (NoSuchFieldException e) {
         throw new IllegalArgumentException("Cannot find field '" + name + "'" + " to value " + value, e);
-      } catch (IllegalAccessException e) {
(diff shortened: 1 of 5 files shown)


Pscheidl (approved):
I'm fine with this for now personally.

sebhrusen (approved):
interesting! yeah if it's only for Py2.7, I don't think it's worth applying a proper fix.

sebhrusen:
>  the test was passing only accidentally on py3 since the same grid was built again due to training frame having different key each time

you mean it should have happened on Py3 as well? But Py3 uses unicode everywhere by default... so what's the issue there?

honzasterba:
the issue was that the training frame was created from numpy via H2OFrame(array), this caused a random key to be generated and thus making grid search think that there are different params from last time and re-training the whole grid again, thus doubling the number of models which in turn satisfied the assertion

Pscheidl (approved):

PR #5161, opened 2020-11-27
[log] convert hex.tree to log4j

honzasterba:


h2o-core/src/main/java/water/util/Log.java
```diff
@@ -9,6 +9,8 @@
 import java.util.ArrayList;
 import java.util.Properties;
 
+import static water.util.StringUtils.fixedLength;
```
maurever, line 12: I thought the goal of this PR is get rid off H2O Log. So why you import the fixedLength method here instead of using this method directly from StringUtils in the case where it is needed?

maurever (commented):
Just one question, otherwise looks good.

Pscheidl (approved):

h2o-algos/src/main/java/hex/tree/DTree.java
```diff
@@ -216,12 +226,12 @@ float splat(DHistogram hs[]) {
...
       double n = way==0 ? _n0 : _n1;
       if( n < parms._min_rows ) {
-//        Log.info("Not splitting: too few observations left: " + n);
+        if (LOG.isTraceEnabled()) LOG.trace("Not splitting: too few observations left: " + n);
         return null; // Too few elements
       }
       double se = way==0 ? _se0 : _se1;
       if( se <= 1e-30 ) {
-//        Log.info("Not splitting: pure node (perfect prediction).");
+        LOG.trace("Not splitting: pure node (perfect prediction).");
```
Pscheidl, line 234: You don't if here for JVM optimization but the perf impact is negligible anyway.

honzasterba: the if is only needed if prepring the log message contains some work, like string concat or calling a complex toString() method

h2o-core/src/main/java/water/util/Log.java
honzasterba: I have moved the method to StringUtils from Log because it was used elsewhere and I wanted to remove the dependency on Log.

koniecsveta (approved):

h2o-core/src/main/java/water/util/Log.java
maurever: Oh, I see. So the next step could be to remove Log dependency from all our classes and then this method will not be useful here as well. Cool, thanks.

maurever (approved):

PR #5167, opened 2020-12-03
Fix warning from matplotlib in explain

tomasfryda:
Change warning to be more user friendly.

### DIFF
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -13,9 +13,13 @@ import matplotlib.figure
 try:
     import matplotlib.pyplot as plt
 except ImportError:
+    import warnings
     # Possibly failed due to missing tkinter in old matplotlib in python 2.7
-    matplotlib.use("Agg")
-    import matplotlib.pyplot as plt
+    warnings.warn(
+        "An error occurred while importing matplotlib with backend \"{}\". Trying with to use Agg backend now."
+            .format(matplotlib.get_backend()))
+    from h2o.utils.ext_dependencies import get_matplotlib_pyplot
+    plt = get_matplotlib_pyplot(True)
 
 
 def _display(object):
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -9,17 +9,8 @@ import numpy as np
 import matplotlib
 import matplotlib.colors
 import matplotlib.figure
+from h2o.utils.ext_dependencies import get_matplotlib_pyplot
 
-try:
-    import matplotlib.pyplot as plt
-except ImportError:
-    import warnings
-    # Possibly failed due to missing tkinter in old matplotlib in python 2.7
-    warnings.warn(
-        "An error occurred while importing matplotlib with backend \"{}\". Trying with to use Agg backend now."
-            .format(matplotlib.get_backend()))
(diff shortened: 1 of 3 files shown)


h2o-py/h2o/explanation/_explain.py
```diff
@@ -13,9 +13,13 @@
...
 except ImportError:
+    import warnings
     # Possibly failed due to missing tkinter in old matplotlib in python 2.7
-    matplotlib.use("Agg")
-    import matplotlib.pyplot as plt
+    warnings.warn(
+        "An error occurred while importing matplotlib with backend \"{}\". Trying with to use Agg backend now."
+            .format(matplotlib.get_backend()))
+    from h2o.utils.ext_dependencies import get_matplotlib_pyplot
+    plt = get_matplotlib_pyplot(True)
```
sebhrusen, line 22: not sure I understand why you don't direcly call this last line... I thought this was the goal of this `get_matplotlib_pyplot`.
Or do you absolutely need tkinter? But in this case, why not having sth like
```
def get_matplotlib_pyplot(server, preferred_backend):
```

tomasfryda: @sebhrusen the thing is that as soon as you do `import matplotlib.pyplot`, it will try to initialize the backend. In python 2 the backend is `TkAgg` by default and if it is not present it will fail with an `ImportError`. 

The purpose of `get_matplotlib_pyplot` was to enable setting a headless backend `Agg` without a warning.

I would prefer not changing the backend at all since the user might have already set some backend, e.g., via `%matplotlib inline` in jupyter.

Other option would be to move this logic to every single method that uses `plt` in `explain.py` (not that much work I could do it today), that would stop the warning on import but I think the logic will have to be the same (if we want to have passing tests on Jenkins) and the warning would appear on the first use.

sebhrusen: ok with the tinker issue, but still, what happens in `_explain.py` can happen anywhere the code calls `get_matplotlib_pyplot(False)`, so why not handling the error there?

btw, also noticed a reference to `matplotlib` module in `frame.py` when this is out of context:
```
            version_number = matplotlib.__version__    # line 3793
```

tomasfryda: Good catch and a good point.

I moved the `plt` import to the methods to warn on first use and modified the `get_matplotlib_pyplot` to be able to reraise the error if it happens. I kept the old behavior and added a keyword argument to `raise` - so we don't break anybody's scripts (they'd be checking for a `None` not an error). The new behavior (`raise`ing an error) is used only in the `explain` module.

sebhrusen (approved):
probably better to import matplotlib only on demand as you're doing now.
Thanks for those changes.
Added a minor suggestion.

h2o-py/h2o/utils/ext_dependencies.py
```diff
@@ -9,10 +9,22 @@ def get_matplotlib_pyplot(server):
...
+        try:
+            # noinspection PyUnresolvedReferences
+            import matplotlib.pyplot as plt
+        except ImportError as e:
+            if server:
+                raise e
+            import warnings
+            # Possibly failed due to missing tkinter in old matplotlib in python 2.7
+            warnings.warn(
+                "An error occurred while importing matplotlib with backend \"{}\". Trying with to use Agg backend now."
```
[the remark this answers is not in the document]
tomasfryda: Thanks!

PR #5168, opened 2020-12-03
Fix failing R tests related to date time handling

tomasfryda:
Locally tested, I will merge it if all tests pass.

### DIFF
--- a/h2o-r/tests/runitUtils/utilsR.R
+++ b/h2o-r/tests/runitUtils/utilsR.R
@@ -1024,12 +1024,12 @@ assertCorrectSkipColumns <-
             else if (is.numeric(f1R[rind, ind]) || is.factor(f1R[rind, ind])) {
               if (allFrameTypes[ind] == 'time')
                 expect_true(
-                  abs(f1R[rind, ind] - f2R[rind, skipcount]) < 10,
+                  abs(f1R[rind, ind] - f2R[rind, skipcount]/1000) < 10,
                   info = paste0(
                     "expected: ",
                     f1R[rind, ind],
                     " but received: ",
-                    f2R[rind, skipcount],
+                    f2R[rind, skipcount]/1000,
                     " in row: ",
                     rind,
                     " with column name: ",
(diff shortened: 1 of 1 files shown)


h2o-r/tests/runitUtils/utilsR.R
```diff
@@ -1024,12 +1024,12 @@ assertCorrectSkipColumns <-
             else if (is.numeric(f1R[rind, ind]) || is.factor(f1R[rind, ind])) {
               if (allFrameTypes[ind] == 'time')
                 expect_true(
-                  abs(f1R[rind, ind] - f2R[rind, skipcount]) < 10,
+                  abs(f1R[rind, ind] - f2R[rind, skipcount]/1000) < 10,
```
sebhrusen, line 1027: still, how do you justify that the 2 frames don't have the same value here?
both are the result of a POST to the backend (different API though), and of a call to `as.data.frame` on the payload.

tomasfryda: @sebhrusen Good point, thanks. It turns out this fix was wrong. The issue is that in the test we have 2 frames: `f1` and `f2` and we convert `f1` directly to dataframe and `f2` created by taking the original `f1` saving it to CSV[1] and then loading it and converting to data frame.  And this makes it lose the type information about the date time columns.

At least that's my hypothesis for now. (it works locally)

[1] or some other formats but in the CSV case we lose the type information about the date time  columns

h2o-r/tests/runitUtils/utilsR.R
```diff
@@ -991,6 +991,9 @@ assertCorrectSkipColumns <-
       wholeFrame <<-
         h2o.uploadFile(inputFileName, skipped_columns = skip_columns)
     }
+    # time type gets lost when exporting to, e.g., csv and importing it back
+    for (i in which(sapply(f1R[,names(wholeFrame)], class) == "POSIXct"))
+      attr(wholeFrame, "types")[[i]] <- "time"
```
sebhrusen, line 996: interesting, didn't notice that the test was also writing to a csv file. Yes , in that case, if the timestamp is saved as a float in the CSV, this is lost.
Cool test bug.

sebhrusen (approved):
Nice fix, thanks

PR #5170, opened 2020-12-03
PUBDEV-4940 Uplift trees - restructure DTree and refactor related code

maurever:
Preparation for Uplift trees POC. 

JIRA: https://h2oai.atlassian.net/jira/software/c/projects/PUBDEV/issues/PUBDEV-4940

### DIFF
--- a/h2o-algos/src/main/java/hex/tree/DHistogram.java
+++ b/h2o-algos/src/main/java/hex/tree/DHistogram.java
@@ -302,10 +302,14 @@ public final class DHistogram extends Iced {
     if (_histoType==SharedTreeModel.SharedTreeParameters.HistogramType.Random) {
       // every node makes the same split points
       Random rng = RandomUtils.getRNG((Double.doubleToRawLongBits(((_step+0.324)*_min+8.3425)+89.342*_maxEx) + 0xDECAF*_nbin + 0xC0FFEE*_isInt + _seed));
-      assert _nbin > 1;
-      _splitPts = makeRandomSplitPoints(_nbin, rng);
-    }
-    else if (_histoType== SharedTreeModel.SharedTreeParameters.HistogramType.QuantilesGlobal) {
+      assert(_nbin>1);
+      _splitPts = new double[_nbin];
+      _splitPts[0] = 0;
+      _splitPts[_nbin - 1] = _nbin-1;
+      for (int i = 1; i < _nbin-1; ++i)
+         _splitPts[i] = rng.nextFloat() * (_nbin-1);
+      Arrays.sort(_splitPts);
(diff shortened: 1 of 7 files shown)


honzasterba:
if tests pass I think we should merge this (without the uplift tests) after some cleanup and continue uplift in a new PR

honzasterba (approved):

michalkurka (approved):

h2o-algos/src/main/java/hex/tree/DHistogram.java
```diff
@@ -302,10 +302,14 @@ public void init(final double[] vals) {
     if (_histoType==SharedTreeModel.SharedTreeParameters.HistogramType.Random) {
       // every node makes the same split points
       Random rng = RandomUtils.getRNG((Double.doubleToRawLongBits(((_step+0.324)*_min+8.3425)+89.342*_maxEx) + 0xDECAF*_nbin + 0xC0FFEE*_isInt + _seed));
-      assert _nbin > 1;
-      _splitPts = makeRandomSplitPoints(_nbin, rng);
-    }
-    else if (_histoType== SharedTreeModel.SharedTreeParameters.HistogramType.QuantilesGlobal) {
+      assert(_nbin>1);
```
michalkurka, line 305: this looks strange, this looks like you reverted code I wrote

h2o-algos/src/main/java/hex/tree/DTree.java
```diff
@@ -140,17 +140,16 @@ StringBuilder printLine(StringBuilder sb ) {
   // Records a column, a bin to split at within the column, and the MSE.
   public static class Split extends Iced {
     final public int _col, _bin;// Column to split, bin where being split
-    final DHistogram.NASplitDir _nasplit;
```
michalkurka, line 143: why not final anymore?

michalkurka:
I don't think I can really say I can review it - is the code just reformatted or are there other changes made?

h2o-algos/src/main/java/hex/tree/DHistogram.java
maurever: Maybe I did rebase incorrectly... Will check.

h2o-algos/src/main/java/hex/tree/DTree.java
maurever: Because I initialize Split before validation, and during validation sometimes direction is changed.

maurever:
> I don't think I can really say I can review it - is the code just reformatted or are there other changes made?

There is still one smoke test failing, so this PR is still in progress. I will convert it to the draft and let you know to make a review again when it will be ready. Thank you.

PR #5175, opened 2020-12-08
PUBDEV-7928: Fix custom distribution reuse

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7928

When the following tests are ran in sequence the second one fails: 
-  h2o-py/tests/testdir_algos/gbm/pyunit_PUBDEV_4076_custom_distribution.py          
-  h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_distributions.py

**Hypothesis**
CustomDistribution is a singleton that keeps one custom distribution at a time. When switching between the distributions it can happen that one thread updates static string variable `_distributionDef`  and another thread compares the newly modified `_distributionDef` with the `params._custom_distribution_func`. Since another thread modified it to match the custom distribution func we return the `_instance` which at this point doesn't correspond to the `_distributionDef`.

**Solution**
Make the `_distributionDef` a property of the class instance `_instance`, this way the `_distributionDef` is tightly coupled with the actual `_instance` so we can actually compare  `_instance._distributionDef` with the `params._custom_distribution_func` and expect the correct distribution to be returned.

### DIFF
--- a/h2o-core/src/main/java/hex/DistributionFactory.java
+++ b/h2o-core/src/main/java/hex/DistributionFactory.java
@@ -441,7 +441,7 @@ class CustomDistribution extends Distribution {
     
     private CustomDistributionWrapper _wrapper;
     private static CustomDistribution _instance;
-    public static String _distributionDef;
+    private String _distributionDef;
     
     private CustomDistribution(Model.Parameters params){
         super(params);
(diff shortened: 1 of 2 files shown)


maurever (approved):
Looks good, thank you @tomasfryda!

michalkurka (approved):

PR #5180, opened 2020-12-10
PUBDEV-5923: Cross-validation predictions not saved with exported model

michalkurka:


michalkurka:
@honzasterba can you please take a look at this draft before things further?

h2o-core/src/main/java/hex/ModelExportOptions.java
```diff
@@ -0,0 +1,7 @@
+package hex;
+
+import water.api.StreamWriteOption;
+
+public enum ModelExportOptions implements StreamWriteOption {
```
honzasterba, line 5: class name is usually singular

honzasterba (commented):
I am not too fond of this solution, especially the varargs and merging of arrays, maybe passing around a Set would be more elegant?
overall I think this is an ok way to to

michalkurka:
> I am not too fond of this solution, especially the varargs and merging of arrays, maybe passing around a Set would be more elegant?

The varargs approach has the advantage of backward compatibility. It is inspired by Java NIO API, eg. 

    static long copy(InputStream in, Path target, CopyOption... options)

It is a somewhat similar API to what we are doing and I believe it is user friendly, I agree that merging of the arrays is clunky but users won't be exposed to that.

michalkurka:
@honzasterba thanks for the review and comments

h2o-core/src/main/java/hex/ModelExportOptions.java
michalkurka: thanks for catching the typo, it really looks bad "ModelExportOptions implements StreamWriteOption" :)

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/remote/RemoteXGBoostHandler.java
```diff
@@ -73,13 +76,13 @@ private StreamingSchema streamBytes(byte[] data) {
...
         if (data == null) dataToSend = new byte[0];
         else dataToSend = data;
-        return new StreamingSchema(os -> {
-            try {
-                IOUtils.copyStream(new ByteArrayInputStream(dataToSend), os);
-            } catch (IOException e) {
-                LOG.error("Failed writing data to response.", e);
-                throw new RuntimeException("Failed writing data to response.", e);
-            }
+        return new StreamingSchema((os, options) -> {
```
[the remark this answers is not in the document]
michalkurka: ouch! will fix

michalkurka:
Added tests, cleaned-up API, fixed found issues.

@honzasterba please take another look

h2o-algos/src/test/java/water/ModelSerializationTest.java
```diff
@@ -21,19 +21,18 @@
...
+import water.runner.H2ORunner;
 import water.util.Log;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.*;
 
+@CloudSize(1)
+@RunWith(H2ORunner.class)
 public class ModelSerializationTest extends TestUtil {
```
honzasterba, line 34: can also remove the extends

h2o-core/src/main/java/hex/Model.java
```diff
@@ -2862,25 +2864,26 @@ private InteractionPair(int v1, int v2, String[] v1Enums, String[] v2Enums) {
...
+    
   /**
    * Exports a binary model to a given location.
    * @param location target path, it can be on local filesystem, HDFS, S3...
    * @param force If true, overwrite already existing file
    * @return URI representation of the target location
    * @throws water.api.FSIOException when writing fails
    */
-  public URI exportBinaryModel(String location, boolean force) throws IOException {
+  public final URI exportBinaryModel(String location, boolean force, ModelExportOption... options) throws IOException {
```
honzasterba, line 2880: should we also convert `force` param into a `ModelExportOption`?

h2o-py/h2o/h2o.py
```diff
@@ -515,12 +515,14 @@ def load_grid(grid_file_path):
...
 
-def save_grid(grid_directory, grid_id):
+def save_grid(grid_directory, grid_id, export_cross_validation_predictions=False):
     """
     Export a Grid and it's all its models into the given folder
 
     :param grid_directory: A string containing the path to the folder for the grid to be saved to.
-    :param grid_id: A chracter string with identification of the Grid in H2O. 
+    :param grid_id: A character string with identification of the Grid in H2O.
+    :param export_cross_validation_predictions: A boolean flag indicating whether the models exported from the grid should be saved
```
honzasterba, line 524: maybe also document that default is False

h2o-core/src/main/java/hex/Model.java
michalkurka: I think that would make sense but it would also break the backwards compatibility. I think it is better to keep as-is instead of introducing a backwards compatible version of the method.

h2o-py/h2o/h2o.py
michalkurka: good idea, thank you - added to affected methods

h2o-algos/src/test/java/water/ModelSerializationTest.java
michalkurka: done

h2o-core/src/main/java/hex/Model.java
honzasterba: we could keep the original method and make it call `exportBinaryModel(location, Option.FORCE);` that would make it backwards compatible, to me it seems like the right thing to do, but feel free to ignore

michalkurka: thanks, I tried it your way but it actually made the existing invocations from the code to look worse - I had to do an union of the Force option and the other options. It also made me realize that the options are made more for what goes into the stream/file instead of _where_ it goes. It seems like they are not good for where it goes because that is controlled by the Persist layer and that one takes argument in a different way.

PR #5181, opened 2020-12-10
[fix] add debug logs during beta constraints encoding

koniecsveta:
this will help to investigate why https://github.com/h2oai/sparkling-water/blob/master/py/tests/unit/with_runtime_sparkling/test_glm.py#L180 is failing in SW

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -3162,18 +3162,10 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
     }
 
     private Frame encodeCategoricalsIfPresent(Frame beta_constraints) {
-      FrameUtils.BetaConstraintsEncoder constraintsEncoder = new FrameUtils.BetaConstraintsEncoder(_dinfo.coefNames(), _dinfo.coefOriginalNames());
-      Frame transformedFrame =  constraintsEncoder.doAll(getBetaConstraintsTypesForEncoder(beta_constraints), beta_constraints).outputFrame();
-      transformedFrame.setNames(beta_constraints._names);
+      String[] coefNames = _dinfo.coefNames();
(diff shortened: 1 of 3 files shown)


h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -3027,7 +3027,12 @@ public double applyBounds(double d, int i) {
     }
 
     private Frame encodeCategoricalsIfPresent(Frame beta_constraints) {
-      FrameUtils.BetaConstraintsEncoder constraintsEncoder = new FrameUtils.BetaConstraintsEncoder(_dinfo.coefNames(), _dinfo.coefOriginalNames());
+      Log.debug("Encoding beta constraints...");
+      String[] coefNames = _dinfo.coefNames();
+      Log.debug("_dinfo.coefNames() content: " + Arrays.toString(coefNames));
```
honzasterba, line 3032: how often is this function called? if the number is non-trivial it would be good to prefix the Arrays.toString calls with a check for debug being enabled

koniecsveta: rather will prefix

koniecsveta: 33c8112

mn-mikke (approved):
I confirm that this PR fixes [the failing beta-constraints tests in SW](https://github.com/h2oai/sparkling-water/pull/2461).

honzasterba (commented):
Log calls with toTwoDimTable().toString() and Arrays.toString should be wrapped in ifDebugEnabled

honzasterba (approved):

koniecsveta:
Hi @wendycwong , this problem appeared after resolving this issue: https://h2oai.atlassian.net/browse/PUBDEV-7473 can you please review whether this approach is ok?

wendycwong (approved):
It looks fine for me.

PR #5184, opened 2020-12-11
PUBDEV-7861 - Feedback for standalone r client implementation

valenad1:
I have implemented feedback from Joseph. Only minor changes, I decided to edit hint from h2o.connect() to h2o.connect(ip="hostname", port=number) as @honzasterba suggested in the old PR because h2o.connect() with default setting is more or less the same as h2o.init(). It is better to remind ip and hostname to user in my opinion.

Thank you

https://h2oai.atlassian.net/browse/PUBDEV-7861


michalkurka (approved):
Thank you!

honzasterba (approved):
lgtm

PR #5185, opened 2020-12-11
Edit all conditions with None, clear and refactor model_base.py a bit

maurever:


wendycwong (approved):
Whoa, that is quite some change in format!

michalkurka (approved):
Thank you

PR #5188, opened 2020-12-16
[xgboost] upgrade to 1.3.0

honzasterba:


valenad1:
This PR is not valid since we go directly to xgboost 1.6: https://github.com/h2oai/h2o-3/pull/6664

PR #5190, opened 2020-12-16
PUBDEV-7861 - fix runit_init_https test

valenad1:
In case of any error don't check file location and always return False.

It could be done with "quite = TRUE" but in that case we check for location /client_info.txt. With this implementation we avoid check of any unknown location and directly return False.

### DIFF
--- a/h2o-r/h2o-package/R/constants.R
+++ b/h2o-r/h2o-package/R/constants.R
@@ -159,4 +159,16 @@ assign("LOG_FILE_NAME", NULL,  .pkg.env)
 
 # Client version of R package
 .h2o.__CLIENT_VERSION_FILENAME <- "client_info.txt"
-.h2o.__CLIENT_VERSION <- function() {return(file.exists(paste0(path.package("h2o", quiet = FALSE), .Platform$file.sep, .h2o.__CLIENT_VERSION_FILENAME)))}
+
+# Return True if installed h2o package location contain .h2o.__CLIENT_VERSION_FILENAME file
+# In case of any error return False always
+.h2o.__CLIENT_VERSION <- function() {
+  tryCatch(
+    return(
+      file.exists(
+        paste0(path.package("h2o", quiet = FALSE), .Platform$file.sep, .h2o.__CLIENT_VERSION_FILENAME)
+      )
+    ),
+    error = function(e) return(FALSE)
+  )
+}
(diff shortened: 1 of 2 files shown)


h2o-r/h2o-package/R/constants.R
```diff
@@ -159,4 +159,16 @@ assign("LOG_FILE_NAME", NULL,  .pkg.env)
 
 # Client version of R package
 .h2o.__CLIENT_VERSION_FILENAME <- "client_info.txt"
-.h2o.__CLIENT_VERSION <- function() {return(file.exists(paste0(path.package("h2o", quiet = FALSE), .Platform$file.sep, .h2o.__CLIENT_VERSION_FILENAME)))}
+
+# Return True if installed h2o package location contain .h2o.__CLIENT_VERSION_FILENAME file
+# In case of any error return False always
+.h2o.__CLIENT_VERSION <- function() {
+  tryCatch(
+    return(
```
honzasterba, line 167: which one of these calls throws an error? first I thought its `file.exists` but then I guess it would be `path.package`. why is it throwing an error?

honzasterba (commented):
also why is this a problem now? what caused the issue?

h2o-r/h2o-package/R/constants.R
valenad1: file.existS quietly tells False but path.package("h2o", quiet = FALSE) fails if h2o is not found and quite = False. If quite = True then it return empty string

honzasterba: why does `path.package("h2o", quiet = FALSE)` fail? h2o package should be loaded at this time

valenad1: When you call

```
library(h2o)
path-to-h2o-library <- path.package("h2o", quiet = FALSE)
```
path-to-h2o-library will be filled.

but call without library(h2o)

path-to-h2o-library <- path.package("h2o", quiet = FALSE)

gives you empty path-to-h2o-library or error in path.package call.

honzasterba (approved):

h2o-r/h2o-package/R/constants.R
michalkurka: @valenad1 can you instead use 

```system.file("your/path", package = "h2o")```

?

michalkurka (approved):

h2o-r/h2o-package/R/constants.R
valenad1: Its working! In case of any issue it returns "".

PR #5191, opened 2020-12-16
Assign iteration from the winning submodel to the final model when used with lambda search

tomasfryda:
alpha_best_selected

`number_of_iterations` in model_summary does not correspond to the number of iterations for the given model assuming `scoring_history` contains all the iterations.

After some digging, I noticed that during lambda search, we overwrite the `_state._iter` so when we generate the model_summary we get use the `number_of_iterations` of the last trained (sub)model instead of the winning one.
This PR, contains a trivial fix that does seem to work as can be seen below.
However, the GLM code is quite unfamiliar to me so it's possible that the fix is incorrect.

Rplot01


wendycwong:
During lambda_search, the iteration just goes from 0 to the end of the iteration for the last lambda value tried or earlier if early-stop is enabled.  Hence, if the best_model is what we are looking for, we should just look at the iteration of the selected_model_index.  I think this is sound.

wendycwong (approved):

PR #5195, opened 2020-12-18
Fix missing argument in explain's get_algorithm

tomasfryda:


michalkurka:
Can you please also add a test? Even a smoke test is fine

PR #5204, opened 2020-12-23
PUBDEV-7923: Ignore metadata files when importing from Delta Lake

michalkurka:
https://h2oai.atlassian.net/browse/PUBDEV-7923

### DIFF
--- a/h2o-core/src/main/java/water/persist/PersistManager.java
+++ b/h2o-core/src/main/java/water/persist/PersistManager.java
@@ -40,7 +40,7 @@ public class PersistManager {
   private static final String PROP_ENABLE_HDFS_FALLBACK = SYSTEM_PROP_PREFIX + "persist.enable.hdfs.fallback";
 
   private static final String PROP_IMPORT_FILES_PATTERNS = SYSTEM_PROP_PREFIX + "persist.importFile.patterns";
-  private static final String PROP_IMPORT_FILES_PATTERNS_DEFAULT = 
+  private static final String DEFAULT_IMPORT_FILES_PATTERNS = 
           "^((?!\\/_delta_log\\/).)*$"; // skip "Delta Lake" log files
 
   private final String[] importFilesPatterns;
@@ -109,7 +109,7 @@ public class PersistManager {
     for (int i = 0; i < stats.length; i++) {
       stats[i] = new PersistStatsEntry();
     }
-    importFilesPatterns = System.getProperty(PROP_IMPORT_FILES_PATTERNS, PROP_IMPORT_FILES_PATTERNS_DEFAULT).split(";");
+    importFilesPatterns = System.getProperty(PROP_IMPORT_FILES_PATTERNS, DEFAULT_IMPORT_FILES_PATTERNS).split(";");
 
     if (iceRoot == null) {
       Log.err("ice_root must be specified.  Exiting.");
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/persist/PersistManager.java
```diff
@@ -38,6 +39,12 @@
    * layer forwards the request through HDFS API. */
   private static final String PROP_ENABLE_HDFS_FALLBACK = SYSTEM_PROP_PREFIX + "persist.enable.hdfs.fallback";
 
+  private static final String PROP_IMPORT_FILES_PATTERNS = SYSTEM_PROP_PREFIX + "persist.importFile.patterns";
+  private static final String PROP_IMPORT_FILES_PATTERNS_DEFAULT =
```
Pscheidl, line 43: Inconsistent naming. This is no property and is not configurable.

Pscheidl (approved):

h2o-core/src/main/java/water/persist/PersistManager.java
michalkurka: yeah, I though "PROP_.._DEFAULT" but I see your point, do you have a suggestion?

wdyt about DEFAULT_IMPORT_FILES_PATTERNS?

Pscheidl: `IMPORT_FILES_PATTERNS_DEFAULT` or your option, whatever. This is minor thing.

michalkurka: thank you, changed the name

h2o-core/src/main/java/water/persist/PersistManager.java
```diff
@@ -401,18 +409,33 @@ public void importFiles(String path, String pattern, ArrayList<String> files, Ar
...
     }
 
-    if(pattern != null && !pattern.isEmpty()) {
-      files.retainAll(matchPattern(path,files,pattern)); //New files ArrayList after matching pattern of choice
-      keys.retainAll(matchPattern(path,keys,pattern)); //New keys ArrayList after matching pattern of choice
-      //New fails ArrayList after matching pattern of choice. Only show failures that match pattern
-      if(!fails.isEmpty()) {
-        fails.retainAll(matchPattern(path, fails, pattern));
-      }
+    final String[] patterns = pattern != null && !pattern.isEmpty() ?
```
mn-mikke, line 412: nit: Isn't this condition redundant with respect to the condition at line [415](https://github.com/h2oai/h2o-3/pull/5204/files#diff-658af709365770678f5a8b58440d20a396e79dd4c0c9f54398eee4b48aaefeedR415)?

h2o-core/src/test/java/water/persist/PersistManagerTest.java
```diff
@@ -46,4 +46,37 @@ public void createReturnsBufferedOutputStreamForFiles() throws IOException  {
         }
     }
 
+    @Test
+    public void deltaLogFilesAreFiltered() throws IOException {
+        File dir = tmp.newFolder("to_import");
+        Files.write(new File(dir, "file-a").toPath(), Collections.singleton("file-a"));
+        Files.write(new File(dir, "file-b").toPath(), Collections.singleton("file-a"));
```
mn-mikke, line 53: nit: `Collections.singleton("file-b")`, but it's irrelevant for the purpose of this test

mn-mikke (approved):
LGTM

h2o-core/src/main/java/water/persist/PersistManager.java
michalkurka: it is, its only purpose is to avoid copying the array in case the user-provided pattern is actually empty; but yes, is technically redundant

h2o-core/src/test/java/water/persist/PersistManagerTest.java
michalkurka: right, I see :) will fix

PR #5209, opened 2021-01-04
rename some_methods in TestUtil to match Java camelCase style

satai:
Rename some TestSupport methods to match Java camelStyle

### DIFF
--- a/h2o-algos/src/jmh/java/hex/pca/PCACatOnlyPUBDEV3988Bench.java
+++ b/h2o-algos/src/jmh/java/hex/pca/PCACatOnlyPUBDEV3988Bench.java
@@ -8,6 +8,7 @@ import org.openjdk.jmh.runner.options.Options;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
 import water.DKV;
 import water.Key;
+import water.TestUtil;
 import water.fvec.Frame;
 import water.util.FrameUtils;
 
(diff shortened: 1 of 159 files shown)


maurever (approved):
Thanks @satai !!!

satai:
I will add a bit more, because it's quite common issue with this class :-/

maurever (approved):
Perfect, thanks. There is a lot of mess, every change in this direction is meaningful.

Pscheidl (approved):
Makes sense to me, better for searching through the code, even though after the years i already got used to it.

Went throught the PR, saw renames only. Should break nothing. Approved.

honzasterba (approved):
💯 👍

michalkurka:
This is going to break everyone's' code who is using the testutils. There are in-house examples of users (MLI). I am not convinced this is a step in the right direction. Especially after we had a similar issue in our SW (not in tests but camel case vs snake case): https://h2oai.slack.com/archives/C0BEJUPAN/p1608136973108700

michalkurka:
@satai no need to close right away. Maybe there is a compromise solution, most of the methods changed are not really used by anyone else. But for stuff like parse_test_file (with path and also optionally target key) - I would be careful there and introduce and keep also the older naming around with deprecated annotation.

Then I would check MLI and make sure it works without any code changes on their end. This should be a pretty good indicator we are in a good shape.

Also, we don't want to make more work for ourselves - eg. adding a new test on the release branch should not cause code not to compile on master after an automatic merge, not on a daily basis.

satai:
Continues here:
https://github.com/h2oai/h2o-3/pull/5226

PR #5210, opened 2021-01-04
Fix MathUtils.compare() - don't have false possitives with negative n…

satai:
…umbers

### DIFF
--- a/h2o-core/src/main/java/water/util/MathUtils.java
+++ b/h2o-core/src/main/java/water/util/MathUtils.java
@@ -377,17 +377,34 @@ public class MathUtils {
     return Math.abs(actual - expected) <= (n-1) * Math.ulp(actual) * absum;
   }
 
-  /** Compare 2 doubles within a tolerance
+  /** 
+   * Compare 2 doubles within a tolerance. True iff the numbers difference within a absolute tolerance or
+   * within an relative tolerance.
+   * 
    *  @param a double
    *  @param b double
-   *  @param abseps - Absolute allowed tolerance
-   *  @param releps - Relative allowed tolerance
+   *  @param absoluteTolerance - Absolute allowed tolerance
+   *  @param relativeTolerance - Relative allowed tolerance
    *  @return true if equal within tolerances  */
-  public static boolean compare(double a, double b, double abseps, double releps) {
-    return
-      Double.compare(a, b) == 0 || // check for equality
-      Math.abs(a-b)/Math.max(a,b) < releps ||  // check for small relative error
-      Math.abs(a - b) <= abseps; // check for small absolute error
+  public static boolean compare(double a, double b, double absoluteTolerance, double relativeTolerance) {
+    assert absoluteTolerance >= 0;
+    assert relativeTolerance >= 0;
+    
+    final boolean equal = Double.compare(a, b) == 0;
+    
+    if (equal) {
+      return true;
+    }
+    
+    final double absoluteError = Math.abs(a - b);
+    
+    if (absoluteError <= absoluteTolerance) {
+      return true;
+    }
+    
+    final double relativeError = Math.abs(absoluteError / Math.max(Math.abs(a), Math.abs(b)));
+
+    return relativeError < relativeTolerance;
   }
 
   // some common Vec ops
--- a/h2o-core/src/test/java/water/util/MathUtilsTest.java
+++ b/h2o-core/src/test/java/water/util/MathUtilsTest.java
@@ -1,0 +1,65 @@
+package water.util;
+
+import junit.framework.TestCase;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static water.util.StringUtils.join;
+
+public class MathUtilsTest extends TestCase {
+
+
+    @Test
+    public void testCompare() throws Exception {
+        assertTrue(MathUtils.compare(1,1, 0.1, 0.1));
+        assertTrue(MathUtils.compare(1,1, 0.0, 0.1));
+        assertTrue(MathUtils.compare(1,1, 0.1, 0.0));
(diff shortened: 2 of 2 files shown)


Pscheidl (approved):

h2o-core/src/main/java/water/util/MathUtils.java
```diff
@@ -377,17 +377,24 @@ public static boolean equalsWithinRecSumErr(double actual, double expected, int
...
-  public static boolean compare(double a, double b, double abseps, double releps) {
-    return
-      Double.compare(a, b) == 0 || // check for equality
-      Math.abs(a-b)/Math.max(a,b) < releps ||  // check for small relative error
-      Math.abs(a - b) <= abseps; // check for small absolute error
+  public static boolean compare(double a, double b, double absoluteTolerance, double relativeTolerance) {
+    assert absoluteTolerance >= 0;
+    assert relativeTolerance >= 0;
+    
+    final boolean equal = Double.compare(a, b) == 0;
```
michalkurka, line 393: This deoptimization missed the point of the exact comparison, the exact equality is fast-path optimization to avoid division. By removing the lazy evaluation you are making this comparison redundant, it will be caught either by relative or absolute error check later.

h2o-core/src/main/java/water/util/MathUtils.java
```diff
@@ -377,17 +377,24 @@ public static boolean equalsWithinRecSumErr(double actual, double expected, int
...
+    final double absoluteError = Math.abs(a - b);
+    final double relativeError = Math.abs(absoluteError / Math.max(a, b));
```
michalkurka, line 395: This has a division by zero problem - please fix.

michalkurka (changes requested):
Add a test for Math.max(a, b) == 0

h2o-core/src/main/java/water/util/MathUtils.java
Pscheidl: Maybe it would be a good idea to set this repository to automatically cancel all older reviewes once new changes are pushed. I originally reviewed a much smaller change: https://github.com/h2oai/h2o-3/commit/434467971495454b4f74c21284bb718f283e5212

satai: Anyway I have reordered it

h2o-core/src/main/java/water/util/MathUtils.java
satai: No, it hasn't, IMO. In this case you get ` Double.POSITIVE_INFINITY` as `relativeError`

h2o-core/src/test/java/water/util/MathUtilsTest.java
```diff
@@ -0,0 +1,65 @@
...
+        assertTrue(MathUtils.compare(1, 1.099, 0, 0.1));
+        assertTrue(MathUtils.compare(-1, -0.900, 0.1, 0.1));
+        assertTrue(MathUtils.compare(-1, -1.099, 0, 0.1));
+        
+        assertFalse(MathUtils.compare(1, 0.8999999, 0, 0.1));
+        assertFalse(MathUtils.compare(1,  1.1,0, 0.09));
+        assertFalse(MathUtils.compare(-1, -0.899999, 0, 0.1));
+        assertFalse(MathUtils.compare(-1, -1.100001, 0, 0.1));
+
+        assertFalse(MathUtils.compare(1.0000, 0, 0, 0.1));
```
honzasterba, line 54: is there a point to all the zeroes after 1.?

h2o-core/src/main/java/water/util/MathUtils.java
michalkurka: That is correct, you do not get an error but the division by zero problem still exists.

The point is to compare the relative difference using a scale of the numbers provided. There are different ways of defining relative difference but the one we typically use is to take the absolute difference of the 2 number and divide by the maximum _absolute_ value of the 2 numbers. Because the 2 numbers are not the same (exact quality) there will be no division by zero.

honzasterba (approved):

h2o-core/src/test/java/water/util/MathUtilsTest.java
satai: No, it was just a relic of generating more tests semi-automatically to cover more cases I was thinking about. I have removed it.

h2o-core/src/main/java/water/util/MathUtils.java
satai: I guess I see your point now. Is `final double relativeError = Math.abs(absoluteError / Math.max(Math.abs(a), Math.abs(b)));`
 what you had in mind?

michalkurka: Yes, that is exactly what I had in mind and why I clicked on "request changes". It addresses more than just division by zero, the method is incorrect for negative numbers as well (case of -1e10 and 0.001). This could even change the results of early stopping for metrics that are not zero-bound.

h2o-core/src/main/java/water/util/MathUtils.java
satai: @Pscheidl I try to rerequest :+1: when I do some bigger changes but I forget sometimes :-/ The original commit was mainly my expression of surprise I run into this issue with some minimal tests to show it.

honzasterba (approved):

h2o-core/src/main/java/water/util/MathUtils.java
Pscheidl: That was a generic comment.

h2o-core/src/main/java/water/util/MathUtils.java
satai: @michalkurka anything else to add?

michalkurka (approved):

PR #5211, opened 2021-01-04
Fix random error in GenericModelTest

satai:
testJavaScoring_mojo_cox_ph_categorical sometimes had small relative differences
of big numbers.

### DIFF
--- a/h2o-algos/src/test/java/hex/generic/GenericModelTest.java
+++ b/h2o-algos/src/test/java/hex/generic/GenericModelTest.java
@@ -839,7 +839,7 @@ public class GenericModelTest extends TestUtil {
 
         final Frame originalModelPredictions = originalModel.score(testFrame);
         Scope.track(originalModelPredictions);
-        assertTrue(TestUtil.compareFrames(genericModelPredictions, originalModelPredictions, 0.000001));
+        assertTrue(TestUtil.compareFrames(originalModelPredictions, genericModelPredictions, 0.000001, 0.00001));
     }
 
     @Test
--- a/h2o-test-support/src/main/java/water/TestUtil.java
+++ b/h2o-test-support/src/main/java/water/TestUtil.java
@@ -793,6 +793,10 @@ public class TestUtil extends Iced {
     assertVecEquals("", expecteds, actuals, delta);
   }
 
+  public static void assertVecEquals(Vec expecteds, Vec actuals, double delta, double relativeDelta) {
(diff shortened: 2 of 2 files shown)


michalkurka (approved):

h2o-test-support/src/main/java/water/TestUtil.java
```diff
@@ -1294,11 +1298,28 @@ protected static RowData toRowData(Frame fr, String[] columns, long row) {
...
+   * Compares two frames. Two frames are equal if and only if they contain the same number of columns, rows,
+   * and values at each cell (coordinate) are the same. Column names are ignored, as well as chunks sizes and all other
+   * aspects besides those explicitly mentioned.
+   * 
+   * @param f1 Frame to be compared, not null
+   * @param f2 Frame to be compared, not null
+   * @param delta absolute tolerance
+   * @param relativeDelta relative tolerance
+   * @return True if frames are the same up to tolerance - number of columns, rows & values at each cell.
+   * @throws IllegalStateException If any inequalities are found
```
honzasterba, line 1320: probably throws AssertionError, no?

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
```diff
@@ -839,7 +839,7 @@ private void testJavaScoringCoxPH(Frame trainingFrame, Frame testFrame, String[]
 
         final Frame originalModelPredictions = originalModel.score(testFrame);
         Scope.track(originalModelPredictions);
-        assertTrue(TestUtil.compareFrames(genericModelPredictions, originalModelPredictions, 0.000001));
+        assertTrue(TestUtil.compareFrames(genericModelPredictions, originalModelPredictions, 0.000001, 0.00001));
```
honzasterba, line 842: the order of args should be `originalModelPredictions, genericModelPredictions` since it is exp, actual

h2o-test-support/src/main/java/water/TestUtil.java
satai: Updated

h2o-algos/src/test/java/hex/generic/GenericModelTest.java
satai: Updated

honzasterba (approved):
I think that we should throw AssertionError instead of IllegalArgumentException but I am fine either way

PR #5212, opened 2021-01-04
fix gbm effective parameters sometimes failing

koniecsveta:


h2o-py/tests/testdir_algos/gbm/pyunit_effective_parameters_gbm.py
```diff
@@ -21,10 +21,10 @@ def test_gbm_effective_parameters():
     response = "economy_20mpg"
     train, valid = cars.split_frame(ratios=[.8], seed=1234)
 
-    gbm1 = H2OGradientBoostingEstimator(seed=1234, stopping_rounds=3)
+    gbm1 = H2OGradientBoostingEstimator(seed=1234, stopping_rounds=3, score_tree_interval=5)
     gbm1.train(x=predictors, y=response, training_frame=train, validation_frame=valid)
 
-    gbm2 = H2OGradientBoostingEstimator(seed=1234, stopping_rounds=3, distribution="bernoulli", stopping_metric="logloss",
+    gbm2 = H2OGradientBoostingEstimator(seed=1234, stopping_rounds=3, score_tree_interval=5, distribution="bernoulli", stopping_metric="logloss",
```
michalkurka, line 27: this will fix it, however, are we sure it will actually stop early? it would be good to check that if that is important to the test

koniecsveta: It's not important for this test. The main aim is to test whether actual values of distribution, stopping_metric,                                  histogram_type and categorical_encoding get set to expected default values when they are not on input (=AUTO). And checking logloss so see whether it produces the same model as when they are explicitly listed on input with default values.

maurever (approved):

PR #5213, opened 2021-01-04
Enhanced Red Hat certification script with fixed for various undocume…

Pscheidl:
…nted states

I do not have exact logs of what failed during the last release, yet the script used in [h2o-kubernetes](https://github.com/h2oai/h2o-kubernetes) repository has been fine-tuned to react to various undocumented states of the certification API. This is a direct copy of the [latest version](https://github.com/h2oai/h2o-kubernetes/blob/master/operator/release/red_hat_docker_certification.py) (with typo fixed).

### DIFF
--- a/scripts/jenkins/red_hat/red_hat_docker_certification.py
+++ b/scripts/jenkins/red_hat/red_hat_docker_certification.py
@@ -22,10 +22,18 @@ def has_scan_errors(digest, pid, api_key):
     response = None
     while response is None:
         intermediate_response = requests.get(url=url, headers=headers)
+        print("Intermediate response: {}".format(intermediate_response.status_code))
         if intermediate_response.status_code == 200:
-            response = intermediate_response
+            try:
+                intermediate_response.json()["data"]["results"]
+                response = intermediate_response
+            except (KeyError, TypeError):
+                # Do effectively nothing if image check results are missing in the response
+                # There are various states the Red Hat API responses with 200, yet the payload may differ
+                # causing KeyError or TypeError when trying to access scan results
+                response = None
         else:
-            time.sleep(5)
+            time.sleep(30)
 
     # Iterate over scan results, print them and search for erroneous states
     scan_err_present = False
@@ -58,7 +66,8 @@ def publish(digest, pid, api_key, tag):
 
     response = requests.post(url, headers=headers)
 
-    if response.status_code != 201:
+    # Official API documentation only declares 201, yet HTTP 200 may be returned as well
+    if response.status_code != 201 and response.status_code != 200:
         print("Unable to publish, invalid status code: {}.".format(response.status_code))
         print(response)
         print(response.content)


scripts/jenkins/red_hat/red_hat_docker_certification.py
```diff
@@ -22,10 +22,18 @@ def has_scan_errors(digest, pid, api_key):
     response = None
     while response is None:
         intermediate_response = requests.get(url=url, headers=headers)
+        print("Intermediate response: {}".format(intermediate_response.status_code))
         if intermediate_response.status_code == 200:
-            response = intermediate_response
+            try:
+                intermediate_response.json()["data"]["results"]
+                response = intermediate_response
+            except (KeyError, TypeError):
```
Pscheidl, line 30: Sometimes, the body is missing when this is hit right after the scan is completed. Therefore, those errors are expected. If there are no results in the body, yet the response is `200`, this is considered to be a failure and the API is polled again. After a while, the API will start to return proper response according to documentation.

satai (approved):

scripts/jenkins/red_hat/red_hat_docker_certification.py
honzasterba: what about checking for the key presence instead of catching the error and then logging the response contents?

honzasterba: also, the comment is a good indicator that this should be moved to a function

scripts/jenkins/red_hat/red_hat_docker_certification.py
```diff
@@ -22,10 +22,18 @@ def has_scan_errors(digest, pid, api_key):
...
+                # Do effectively nothing if image check results are missing in the response
+                # There are various states the Red Hat API responses with 200, yet the payload may differ
+                # causing KeyError or TypeError when trying to access scan results
+                response = None
         else:
-            time.sleep(5)
+            time.sleep(30)
```
honzasterba, line 36: any reason why?

scripts/jenkins/red_hat/red_hat_docker_certification.py
```diff
@@ -58,7 +66,8 @@ def publish(digest, pid, api_key, tag):
 
     response = requests.post(url, headers=headers)
 
-    if response.status_code != 201:
+    # Official API documentation only declares 201, yet HTTP 200 may be returned as well
+    if response.status_code != 201 and response.status_code != 200:
```
honzasterba, line 70: how about `response.status_code not in [201, 200]`?

scripts/jenkins/red_hat/red_hat_docker_certification.py
Pscheidl: It takes 30 minutes approximately - up to 4 hours officially. No need to poll it every 5 seconds or so. Or is it ?

scripts/jenkins/red_hat/red_hat_docker_certification.py
Pscheidl: The data might be missing, or just the results might be missing. Or the results might be empty, resulting in the TypeError. This code is like Pokemon - gotta catch them good. I think it's good enough, but sure won't help to traverse it manually.

Agreed with the function, why not.

scripts/jenkins/red_hat/red_hat_docker_certification.py
honzasterba: that is a good reason 👍

Pscheidl:
Closing this branch. There will be a release candidate branch with all the changes, not just this individual change. Won't be merged to zermelo without review.

PR #5214, opened 2021-01-04
Fix fault tolerance test failing with GLRM on jenkins

honzasterba:


h2o-py/tests/testdir_algos/grid/pyunit_grid_fault_tolerance_resume.py
```diff
@@ -283,13 +283,13 @@ def grid_ft_resume(train, grid_id, params, hyper_parameters, start_grid, resume_
     grid = start_grid(grid_id, export_dir, train, params, hyper_parameters)
     grid_in_progress = None
     times_waited = 0
-    while (times_waited < 300) and (grid_in_progress is None or len(grid_in_progress.model_ids) == 0):
```
michalkurka, line 286: 300s seems like a lot for this test, is there a GLMR parameterization that would finish quicker?

honzasterba: each sleep is 0.1s so the max total is 30s

Pscheidl (approved):

PR #5220, opened 2021-01-08
Fix stacked ensemble poor models test

tomasfryda:
This issue is that all but the default GLM metalearner  (not the “AUTO”) provide a model with all classes in prediction.
The default GLM didn’t use to. But after [1] it started to provide similar predictions as all the other metalearners (containing all the classes). To me it looks like [1] fixed the issue so my course of action would be to change the test to expect the same as all the other metalearners.

(happens only on master branch)

[1] [PUBDEV-7843] build XGB CV models on multiple GPUs in parallel (#5075)

### DIFF
--- a/h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_with_poor_models.py
+++ b/h2o-py/tests/testdir_algos/stackedensemble/pyunit_stackedensemble_with_poor_models.py
@@ -112,8 +112,7 @@ def test_models_not_predicting_some_classes_dont_corrupt_resulting_SE_model():
         se.train(data.x, data.y, data.train)
         results = scores_and_preds(se, data.test)
         print(results)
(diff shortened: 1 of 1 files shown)


sebhrusen (approved):
Thanks @tomasfryda

ledell (approved):
Thanks, @tomasfryda!

honzasterba:
I have found the cause of this change in behaviour, see here for a fix: https://github.com/h2oai/h2o-3/pull/5229
If you wish for the GLM model to return more sensible results I would encourage to increase the max_runtime_secs.

michalkurka:
@tomasfryda can this be closed? PR https://github.com/h2oai/h2o-3/pull/5229 should resolve the issue

tomasfryda:
Yes, sorry, I was planning to close this today.

PR #5224, opened 2021-01-12
PUBDEV-4940 implement UpliftRandomForest

maurever:
**JIRA:** 
https://h2oai.atlassian.net/jira/software/c/projects/PUBDEV/issues/PUBDEV-4940

**Sources:**
- https://www.stochasticsolutions.co.uk/pdf/kdd2011late.pdf
- https://stochasticsolutions.com/pdf/sig-based-up-trees.pdf

**Existing implementation:**
- https://github.com/cran/uplift
- https://github.com/uber/causalml

**TODO:**
- [x] implement uplift calculation to find the best split
- [x] implement AUUC metric calculation
- [x] test basic functionality
- [x] plot AUUC
- [x] implement gainsUplift table 
- [x] check early stopping
- [x] improve AUUC calculation accuracy due to binning
- [x] test against UpliftRF and CausalML (predictions, AUUC metric calculation)
- [x] run performance test against DRF 
- [x] run and check GBM benchmarks against this branch (done -> slower GBM and DRF for higher  numbers in `nbins` parameter)
- [x] add smoke tests
- [x] separate uplift tree code from tree algos due to slowdown
- [x] documentation (all doc commits: https://github.com/h2oai/h2o-3/pull/5565/commits/62c62ee34595534ca99ac96360b6eec3f756cef5)
- [x] python smoke test
- [x] improve AUUC plot - add the random curve, change naming
- [x]  Add `auuc_type` to performance to get AUUC for unseen test data

**In this PR:**
- [x] disable cross-validation
- [x] disable weights
- [x] disable early stopping
- [x] grid search not implemented 

**Improvements for future PR:**
- add Q coeficient calculation
- add new parameters: minimal sample size in control and treatment group
- add cross-validation
- add early stopping
- add MOJO
- optimize AUUC calculation
- calculate AUUC exactly for small datasets?
- calculate variable importances
- implement grid search

**Implementation problems:**

- saving treatment and control leaves into two trees instead of one, the reason is the new tree structure should be implemented for this special case of tree -> task for future improving the algorithm, for example when MOJO will be implementing

**Uplift Random Forest API:**

- Uplift trees implementation is currently supported only for binomial classification.
- API for Python:
    - `H2OUpliftRandomForestEstimator` : train a model 
    -  `performance.auuc` : get the default AUUC from the performance object
    - `model.auuc `:  get the default AUUC from the model
    - `performance.auuc_table` : get all types of AUUCs from the performance object
    - `model.auuc_table` :  get all types of AUUCs from the model
    - `plot_uplift `: plot uplift curve or returns values for x and y axis of the plot, you can set a metric type
    - `make_metrics` : make H2OBinomialUpliftMetrics metrics
    
- API for R
   - `h2o.upliftRandomForest `: train a model
   - `h2o.performance` : returns H2OBinomialUpliftMetrics
   -` h2o.auuc `: get the default AUUC
   - `h2o.auuc_table` : get all types of AUUCs
   - `plot.H2OBinomialUpliftMetrics` : plot uplift curve or returns values for x and y-axis of the plot, you can set a metric type
   - `h2o.make_metrics` : make H2OBinomialUpliftMetrics metrics

**Python Uplift curve:**

uplift_curve_gain_python

uplift_curve_lift_python

uplift_curve_qini_python

**R Uplift curve:**

uplift_curve_gain_R

uplift_curve_lift_R

uplift_curve_qini_R


h2o-algos/src/main/java/hex/tree/DTree.java
```diff
@@ -1364,6 +1461,424 @@ static Split findBestSplitPoint(DHistogram hs, int col, double min_rows, int con
...
+    double   wlo[] = MemoryManager.malloc8d(nbins+1);
+    double  wYlo[] = MemoryManager.malloc8d(nbins+1);
+    double wYYlo[] = MemoryManager.malloc8d(nbins+1);
+    double[]  numloTreat = MemoryManager.malloc8d(nbins + 1);
+    double[]  resploTreat = MemoryManager.malloc8d(nbins + 1);
+    double[] numloContr = MemoryManager.malloc8d(nbins + 1);
+    double[]  resploContr = MemoryManager.malloc8d(nbins + 1);
+
+    for( int b = 1; b <= nbins; b++ ) {
+      int id = vals_dim * (b - 1);
```
koniecsveta, line 1521: `b-1` could be stored in local variable since it is used 9-times

h2o-algos/src/main/java/hex/tree/DTree.java
```diff
@@ -1364,6 +1461,424 @@ static Split findBestSplitPoint(DHistogram hs, int col, double min_rows, int con
...
+    double wYNA = hs.wYNA();
+    double wYYNA = hs.wYYNA();
+    double var = (wYYlo[nbins]+wYYNA)*tot - (wYlo[nbins]+wYNA)*(wYlo[nbins]+wYNA);
+    if( ((float)var) == 0f ) {
+      if (LOG.isTraceEnabled()) LOG.trace("can't split " + hs._name + ": var = 0.");
+      return null;
+    }
+
+    // Compute mean/var for cumulative bins from nbins to 0 inclusive.
+    double   whi[] = MemoryManager.malloc8d(nbins+1);
```
koniecsveta, line 1559: `nbins+1` should also be in local variable when it is repeated multiple times

h2o-algos/src/main/java/hex/tree/DTree.java
```diff
@@ -1364,6 +1461,424 @@ static Split findBestSplitPoint(DHistogram hs, int col, double min_rows, int con
...
+    if( tot < 2*min_rows ) {
+      if (LOG.isTraceEnabled()) LOG.trace("can't split " + hs._name + ": min_rows: total number of observations is " + tot);
+      return null;
+    }
+    // If we see zero variance, we must have a constant response in this
+    // column.  Normally this situation is cut out before we even try to split,
+    // but we might have NA's in THIS column...
+    double wYNA = hs.wYNA();
+    double wYYNA = hs.wYYNA();
+    double var = (wYYlo[nbins]+wYYNA)*tot - (wYlo[nbins]+wYNA)*(wYlo[nbins]+wYNA);
```
koniecsveta, line 1552: the computation could be extracted to a method with explanatory name for a better readability

h2o-algos/src/test/java/hex/tree/DHistogramTest.java
```diff
@@ -190,14 +190,14 @@ public void testUpdateHistoWithIntOpt() {
             SharedTreeModel.SharedTreeParameters.HistogramType.UniformAdaptive, 42L, null, null, false);
     histoOpt.init();
 
-    histoOpt.updateHisto(weights, null, dataInt, ys, null, rows, N, 0);
+    histoOpt.updateHisto(weights, null, dataInt, ys, null, rows, N, 0, null);
```
koniecsveta, line 193: I think it would be cleaner to create second `updateHisto` method that would call the one extended with the last parameter `null` and all original usage would remain unchanged

h2o-core/src/main/java/hex/Model.java
```diff
@@ -509,6 +512,12 @@ public long getOrMakeRealSeed(){
      *  if a validation frame was not specified */
     public final Frame valid() { return _valid==null ? null : _valid.get(); }
 
+    public String[] getNonPredictors() {
+        return Arrays.stream(new String[]{_weights_column, _offset_column, _fold_column, _response_column, _treatment_column})
```
koniecsveta, line 516: this could be simplified using ` Stream.of`

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1544,10 +1569,11 @@ protected AdaptFrameParameters makeAdaptFrameParameters() {
     String getOffsetColumn();
     String getFoldColumn();
     String getResponseColumn();
+    String getTreatmentColumn();
     double missingColumnsType();
     int getMaxCategoricalLevels();
     default String[] getNonPredictors() {
-      return Arrays.stream(new String[]{getWeightsColumn(), getOffsetColumn(), getFoldColumn(), getResponseColumn()})
+      return Arrays.stream(new String[]{getWeightsColumn(), getOffsetColumn(), getFoldColumn(), getResponseColumn(), getTreatmentColumn()})
```
koniecsveta, line 1576: this could be simplified using `Stream.of`

h2o-core/src/main/java/hex/ModelMetricsBinomialUplift.java
```diff
@@ -0,0 +1,294 @@
...
+                }
+            }
+            return makeModelMetrics(m, f, preds, resp, weight, treatment);
+        }
+
+        private ModelMetrics makeModelMetrics(final Model m, final Frame f, final Frame preds,
+                                              final Vec resp, final Vec weight, final Vec treatment) {
+            GainsUplift gul = null;
+            AUUC auuc = null;
+            if (_wcount > 0) {
```
koniecsveta, line 227: conditions could be collabsed

h2o-algos/src/main/java/hex/tree/DTree.java
michalkurka: this is method is largely a copy&paste of the original findBestSplitPoint method - lets keep it the same as much as possible for now and then refactor later

h2o-core/src/main/java/hex/ModelMetricsBinomialUplift.java
```diff
@@ -0,0 +1,294 @@
...
+        /**
+         * @param m       Model to calculate GainsUplift for
+         * @param preds   Predictions
+         * @param resp    Actual label
+         * @param weights Weights
+         * @param treatment  Treatment column               
+         * @return An Optional with GainsUplift instance if GainsUplift is not disabled (gainsUplift_bins = 0). Otherwise an
+         * empty Optional.
+         */
+        private Optional<GainsUplift> calculateGainsUplift(Model m, Frame preds, Vec resp, Vec weights, Vec treatment) {
```
koniecsveta, line 265: since only `isPresent` method is called on the result, using `Optional` doesn't make sense and is equivalent to returning `null` or the result directly and performing `null` comparison without an additional wrapper

h2o-algos/src/main/java/hex/tree/DTree.java
koniecsveta: okok 👍

h2o-algos/src/main/java/hex/tree/DTree.java
maurever: Good point, but no refactor in this code is allowed within this PR.

h2o-algos/src/main/java/hex/tree/DTree.java
maurever: Good point, but no refactor in this code is allowed within this PR.

h2o-core/src/main/java/hex/Model.java
maurever: Good point, but no refactor in this code is allowed within this PR.

h2o-algos/src/test/java/hex/tree/DHistogramTest.java
maurever: Good point, will change.

h2o-core/src/main/java/water/util/ArrayUtils.java
```diff
@@ -2219,4 +2219,56 @@ private static boolean isPresentInAllWords(String prefix, String[] words) {
...
+
+  /*
+   * Linear interpolation values in the array with Double.NaN values.
+   * The interpolation always starts from 0. 
+   * The last element of array cannot be Double.NaN.
+   *
+   * @param array input array with Double.NaN values
+   */
+  public static void interpolateLinear(double[] array){
+    assert array.length > 0 && !Double.isNaN(array[array.length-1]);
```
valenad1, line 2231: Can we have some message here? I faced this assert before when I had some input parameter not correct, but I cannot reproduce it in the new version.

michalkurka (approved):
Great work @maurever! Looks good to me - great start - further refactoring will be needed but this is good enough to go into master

h2o-core/src/main/java/water/util/ArrayUtils.java
maurever: There was a bug, so I change the code. That maybe why you are not able to reproduce it. What do you set into input parameters? 

I add a message there, no problem.

koniecsveta (approved):
LGTM 👌  Thanks @maurever !

valenad1 (approved):
Good job! 💯

PR #5227, opened 2021-01-12
Assisted clustering on Kubernetes

Pscheidl:


Pscheidl:
As I was reworking this to fit into a standalone module and separate clouding from the rest of the logic,I found one issue with the logic - we need Kubernetes readiness probes to be active from the h2o-k8s module on k8s ! And two `AbstractEmbeddedConfig`s can't be active. Possible solutions:

1. Make this Kubernetes specific, yet in a separate module producing separate artifact. It would still expose readiness probe APIs.
2. Drop support for older H2O via operator and make this part of h2o-k8s module.
3. Create module hierarchy:

```
                        Readiness probes are defined here            Assisted clustering endpoint is defined here.

                           +-------------------+                   +---------------------------+
Only activated on K8S      |                   |                   |                           |
headless service-based     |  H2O-K8S Module   |                   | Assisted clusering module | Activated only when generic AC is enabled.
clustering.                |                   |                   |                           | Shuts down REST API once clustered.
                           +---------------^---+                   +-------^-------------------+
                                           |                               |
                           Depends on.     |                               | Depends on.
                                           |                               |
                                         +-+-------------------------------+---+
                                         |                                     |
                                         |                                     |
                                         | H2O-K8S Assisted clustering module  |
                                         |                                     |
                                         |                                     |
                                         +-------------------------------------+
                                 Exposes assisted clustering endpoint from AC module
                                 + keeps the REST API open with readiness probe from K8S module exposed.

                                 Only activated when operator enabled K8S assisted clusteirng via env var.
```

Pscheidl:
Today, I realized there is no need for a special kubernetes version of this module with readiness probe enabled. This can be solved simply by observing the H2O cluster and once clustered, the leader node is assigned a special label. The service created afterwards then targets only this one pod with leader node. This also prevents users from targeting non-leader nodes (unless a service is set up manually in a wrong way).

h2o-clustering/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Example
+
+A flatfile with an IPv6 and IPv4 address with ports defined in both cases (mandatory).
+```
+[1200:0000:AB00:1234:0000:2552:7777:1313]:54321
+9.255.255.255:54321
+```
+
+## Distribution
```
michalkurka, line 30: Hmm, I thought it would be better not to include it - that would give use the freedom to upgrade the code without upgrading H2O in deployments.

Pscheidl: A small change. And indeed, it is better. Will do.

michalkurka: We can still include it in the k8s docker images, but I would leave it out of the jar to give us the option to swap them with less effort. WDYT?

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
```diff
@@ -0,0 +1,79 @@
...
+public class AssistedClusteringEmbeddedConfigProvider implements EmbeddedConfigProvider {
+    private static final Logger LOG = Logger.getLogger(AssistedClusteringEmbeddedConfigProvider.class);
+
+    private AssistedClusteringEmbeddedConfig assistedClusteringEmbeddedConfig;
+    private String flatfile;
+    private final Object notification = new Object();
+
+    @Override
+    public void init() {
+        final Consumer<String> flatFileCallback = s -> {
```
michalkurka, line 28: This method is called to initialize the provider, regardless of if the provider is active or not, see javadoc:

    /**
     * Provider initialization. Guaranteed to be called before any other method is called, including the`isActive`
     * method.
     */

This means you shouldn't probably do anything like this in this method, instead, you would want to move this code to `getConfig`

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
```diff
@@ -0,0 +1,79 @@
...
+            // Make sure to set the flatfile first and then notify, as the flatfile is immediately used
+            // to produce a resulting config
+            flatfile = s;
+            synchronized (notification) {
+                notification.notify();
+            }
+        };
```
michalkurka, line 35: I think the consumer-producer code could be simplified by using SynchronousQueue to exchange the flatfile. The advantage is that you can use a timeout on the consumer side and you can occasionally produce logs about how the process is going. I would use XGBoostUpdater#invoke for inspiration.

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.IOException;
+import java.util.Optional;
+import java.util.function.Consumer;
+
+/**
+ * Provides {@link AssistedClusteringEmbeddedConfig} as a result of external assist. The resulting
+ * embedded config provides a flatfile assembled by external service, given to H2O. 
+ * This EmbeddedConfig has no timeout and will wait indefinitely until a flatfifle is received.
+ */
+public class AssistedClusteringEmbeddedConfigProvider implements EmbeddedConfigProvider {
```
michalkurka, line 19: What is the advantage of a "PUSH" design (the node is waiting for being told where are his colleagues) compared to "POLL" design - the node would be asking the operator to get a list of nodes?

The advantage of "POLL":
- not HTTP endpoint needs to be started,
- the implementation on the client-side is simpler - a simple loop with synchronous HTTP rest calls to operators,
- more robust - if there is a network issue, a repeated call will recover from the failure.

Pscheidl: The contract is still there on the polled server as well - it needs to provide a flatfile of given format and it needs to provide it correctly. This just swaps the role transfers the complexity on the other side - what was client is the server and vice versa.

- No HTTP endpoint has to be started (but it has to - only on the polled flatfile provider's side)
- The implementation on the h2O side is simpler, but the system polled is not.
- The flatfile provider may call H2O repeatedly as well in case off network failure (when it's not polled and it's in the role of a client of course). In fact, it is often the assistant who started H2O and is able to act upon network issues (Especially in K8S, where custom controllers are far more powerful with more rights than a plain pod - typically. But not only in K8S). This, from my POV places the burden on the right side. If a Kubernetes operator doesn't receive any request, it'd have to time out and try to do some guesswork and diagnose the problem on it's own. Doable, but seems like making a straightforward point of failure not easy to spot.

The advantage was exatly that. Less stress on the assisting service, more on H2O. Besides this factor, the following was my motivation:
1. The assistant service knows exactly when the flatfile is ready and pushes the ready state. Polling means additional calls in vain. Event-driven means more effective usually (even though nobody cares about that).
2. The API definition is easy to distribute and the client is easy to generate. Making clients implement a server to be polled is harder for the client. I can even give out a simple `curl` example and people will immediately get the API contract.
3. The more complex part is written once on H2O's side. And the multiple clients are much easier to do.

Yet polling comes with advantages: It is doable with pure Java and no additional dependencies (not even NanoHTTP). So we'd only depend on abstractembeddedconfig from h2o-core.

One more thought:
HTTP server to be polled is done in no time in Rust. Fast, reliable, extremely small, disposable once not needed. Much better, safer and faster done than in Java, with proper implementation of graceful shutdowns without requirements to pull in megabytes of dependencies - [see](https://hyper.rs/guides/server/echo/). I'm saying this because this would **have been** actually easier path for me.

I'm not against the polling solution, just described my motivation. There are two perspectives - one is design. Where to we want to place the burden. Second is effectivity. Making it even-driven is naturally more effective and everything occurs just when it needs to, as described above. On the other side, polling is "easier" on H2O, where the implementation is definitely more complex (but hey, we have it done now ?). I'm advocate of both, but like the push a little bit more. Any thoughts ?

michalkurka: Yes, agreed - it is about what side has the complexity. This is a 1-to-many relationship and in my point of view, it is better to keep the "many" side simpler and the "1" side complex from a debuggability point of view. It is also easier to upgrade the service (operator) than to upgrade H2O (which is addressed by having the library not bundled in the jar).

I am fine with either, just wanted to know your reasons. Thank you for explaining.

Pscheidl: Would be nice to have opinions of @mn-mikke (especially but on exclusively from SW POV) and @honzasterba :wink:

h2o-clustering/README.md
Pscheidl: The Kubernetes Docker image could have it, yes. No harm done there. The Operator would have to know though (or it won't, as putting the same classes on CP twice does not harm anyone).

h2o-clustering/src/main/java/water/clustering/api/AssistedClusteringEndpoint.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+        // The text/plain content-type is stored as `postData` by HTTPD in the map.
+        final String postBody = map.get("postData");
+
+        if (postBody != null) {
+            final Lock writeLock = lock.writeLock();
+            try {
+                writeLock.lock();
+                if (flatFileReceived.get()) {
+                    return NanoHTTPD.newFixedLengthResponse(NanoHTTPD.Response.Status.FORBIDDEN, RESPONSE_MIME_TYPE,
```
honzasterba, line 71: FORBIDDEN means something related to authentication, this just a BAD_REQUEST IMO

h2o-clustering/src/main/java/water/clustering/api/GracefulAsyncRunner.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class GracefulAsyncRunner extends NanoHTTPD.DefaultAsyncRunner {
+    private static final Logger LOG = Logger.getLogger(GracefulAsyncRunner.class);
+
+    @Override
+    public void closeAll() {
+        while (getRunning().size() > 0) {
+            synchronized (this) {
+                try {
+                    this.wait();
```
honzasterba, line 24: should there be a timeout in the wait() call to make sure this does not hang indefinitelly?

h2o-clustering/src/main/java/water/clustering/api/GracefulAsyncRunner.java
```diff
@@ -0,0 +1,40 @@
...
+                } catch (InterruptedException e) {
+                    LOG.error(e.getMessage(), e);
```
honzasterba, line 26: do you really want to ignore interrupts? I would think not

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
```diff
@@ -0,0 +1,77 @@
...
+        try (final AssistedClusteringRestApi assistedClusteringRestApi = startAssistedClusteringRestApi(flatFileCallback)
+                .orElseThrow(() -> new IllegalStateException("Assisted clustering Rest API unable to start."))) {
+            try {
+                // Blocks until a flatfile is received. 
+                synchronized (notification) {
+                    notification.wait();
+                }
+                return new AssistedClusteringEmbeddedConfig(flatfile);
+            } catch (InterruptedException e) {
+                LOG.error(e.getMessage(), e);
```
honzasterba, line 72: just `Thread.currentThread().interrupt();` is fine here

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
honzasterba: dont have a strong preference, personally I find the PUSH method more elegant and easier to work with

h2o-clustering/src/main/java/water/clustering/api/GracefulAsyncRunner.java
Pscheidl: I'd say it is not required, as by design the rest of the code just replies and no matter if the reply fails or succeds, the asyncRunner frees the resources and this part is unblocked. To make this futureproof, I've introduced a one minute timeout, which is PLENTY of space.

h2o-clustering/src/main/java/water/clustering/api/GracefulAsyncRunner.java
Pscheidl: I am effectively not ignoring it, the handling of the interrupt means to ignore the graceful shutdown period and do a shutdown immediately (resources are freed). Which is done by the `super.closeAll()` call below. This call will be reached anyway.

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
Pscheidl: True. Modified.

h2o-clustering/src/main/java/water/clustering/api/GracefulAsyncRunner.java
Pscheidl: Oh now I see, I am not respecting the interrupt. Fixed by closing all connections immediately and self-interrupting.

h2o-clustering/src/main/java/water/clustering/api/AssistedClusteringEndpoint.java
Pscheidl: Changed.

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
Pscheidl: The timeout is something I'm not sure we want.

h2o-clustering/src/main/java/water/clustering/api/GracefulAsyncRunner.java
honzasterba: I think you want a timeout here and a short one, because between calling `getRunning().size()` and this.wait() the state can change and you could be waiting forever

Pscheidl: True. Will make it one second - the while cycle is the guard..

Pscheidl:
Found one more possible issue - the callback is blocking call, holding the lock for long, blocking other requests from being answered. Making it parallel.

h2o-clustering/src/main/java/water/clustering/api/GracefulAsyncRunner.java
Pscheidl: Done.

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
Pscheidl: Done. I kept the callback interface inside the REST API to make it more generic and just did the queue on the client of the endpoint. Also made the callback asynchronous, so it's not blocking the REST API.

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
Pscheidl: Moved.

Pscheidl:
All previous comments resolved, resolved a couple of things I found myself with - some asserts missing in tests, added non-blocking callback invocation etc.

I'll adjust one more thing - make a JAR with things packed properly (this code + nanohttpd basically).

Pscheidl:
This is how the produced JAR looks like:

```
.
├── fi
│   └── iki
│       └── elonen
│           ├── InternalRewrite.class
│           ├── NanoHTTPD$1.class
│           ├── NanoHTTPD$AsyncRunner.class
│           ├── NanoHTTPD$ClientHandler.class
│           ├── NanoHTTPD$ContentType.class
│           ├── NanoHTTPD$Cookie.class
│           ├── NanoHTTPD$CookieHandler.class
│           ├── NanoHTTPD$DefaultAsyncRunner.class
│           ├── NanoHTTPD$DefaultServerSocketFactory.class
│           ├── NanoHTTPD$DefaultTempFile.class
│           ├── NanoHTTPD$DefaultTempFileManager.class
│           ├── NanoHTTPD$DefaultTempFileManagerFactory.class
│           ├── NanoHTTPD$HTTPSession.class
│           ├── NanoHTTPD$IHTTPSession.class
│           ├── NanoHTTPD$Method.class
│           ├── NanoHTTPD$Response$1.class
│           ├── NanoHTTPD$Response$ChunkedOutputStream.class
│           ├── NanoHTTPD$Response$IStatus.class
│           ├── NanoHTTPD$Response$Status.class
│           ├── NanoHTTPD$Response.class
│           ├── NanoHTTPD$ResponseException.class
│           ├── NanoHTTPD$SecureServerSocketFactory.class
│           ├── NanoHTTPD$ServerRunnable.class
│           ├── NanoHTTPD$ServerSocketFactory.class
│           ├── NanoHTTPD$TempFile.class
│           ├── NanoHTTPD$TempFileManager.class
│           ├── NanoHTTPD$TempFileManagerFactory.class
│           ├── NanoHTTPD.class
│           ├── router
│           │   ├── RouterNanoHTTPD$1.class
│           │   ├── RouterNanoHTTPD$BaseRoutePrioritizer.class
│           │   ├── RouterNanoHTTPD$DefaultHandler.class
│           │   ├── RouterNanoHTTPD$DefaultRoutePrioritizer.class
│           │   ├── RouterNanoHTTPD$DefaultStreamHandler.class
│           │   ├── RouterNanoHTTPD$Error404UriHandler.class
│           │   ├── RouterNanoHTTPD$GeneralHandler.class
│           │   ├── RouterNanoHTTPD$IndexHandler.class
│           │   ├── RouterNanoHTTPD$InsertionOrderRoutePrioritizer.class
│           │   ├── RouterNanoHTTPD$IRoutePrioritizer.class
│           │   ├── RouterNanoHTTPD$NotImplementedHandler.class
│           │   ├── RouterNanoHTTPD$ProvidedPriorityRoutePrioritizer.class
│           │   ├── RouterNanoHTTPD$StaticPageHandler.class
│           │   ├── RouterNanoHTTPD$UriResource.class
│           │   ├── RouterNanoHTTPD$UriResponder.class
│           │   ├── RouterNanoHTTPD$UriRouter.class
│           │   └── RouterNanoHTTPD.class
│           ├── SimpleWebServer$1.class
│           ├── SimpleWebServer$2.class
│           ├── SimpleWebServer$3.class
│           ├── SimpleWebServer.class
│           ├── util
│           │   └── ServerRunner.class
│           ├── WebServerPlugin.class
│           └── WebServerPluginInfo.class
├── LICENSE.txt
├── META-INF
│   ├── MANIFEST.MF
│   ├── maven
│   │   └── org.nanohttpd
│   │       ├── nanohttpd
│   │       │   ├── pom.properties
│   │       │   └── pom.xml
│   │       ├── nanohttpd-nanolets
│   │       │   ├── pom.properties
│   │       │   └── pom.xml
│   │       └── nanohttpd-webserver
│   │           ├── pom.properties
│   │           └── pom.xml
│   └── nanohttpd
│       ├── default-mimetypes.properties
│       └── mimetypes.properties
└── water
    └── clustering
        ├── api
        │   ├── AssistedClusteringEndpoint.class
        │   ├── AssistedClusteringRestApi.class
        │   └── GracefulAsyncRunner.class
        ├── AssistedClusteringEmbeddedConfig.class
        └── AssistedClusteringEmbeddedConfigProvider.class
```

h2o-clustering/README.md
Pscheidl: As the operator has to do the check anyway, there is no need to do this now. I'd skip this optional step of adding this into the Kubernetes jar. The only user ATM is the operator. If someone goes as far as writing a custom clustering assistant, they can surely add one more `java -cp`

Pscheidl:
@michalkurka @honzasterba Ready for another round of reviews.

I'll think about a quick way to test this on a real thing, including a small Kubernetes cluster (could be done with kubectl, curl and `grep`ed pod IPs).

Pscheidl:
Docker images for K8S tests are built over here: http://mr-0xg1:8080/view/H2O-3/job/docker-images/job/h2o-3-k8s-test-docker-build/6/

h2o-k8s/tests/clustering/docker-init-script.sh
```diff
@@ -9,21 +9,49 @@ fi
...
+# Check H2O is clustered with a script independent from the one in h2o-helm triggered previously with 'helm test'
 timeout 120s bash h2o-cluster-check.sh
-export EXIT_STATUS=$?
+export CLUSTER_EXIT_CODE=$?
 kubectl get pods
 kubectl get nodes
+# To save resources in Jenkins pipeline (e.g. CPUs are limited), remove the H2O cluster deployed via HELM
+helm uninstall h2o 
+# Test assisted clustering regime by adding h2o-clustering artifact into the H2O container using volumes.
+envsubst < h2o-assisted-template.yaml >> h2o-assisted.yaml
```
michalkurka, line 36: Does this mean it is all tested in a single stage?

Would it be better to have different stages? If one thing doesn't work - do the other tests run?

Pscheidl: If one deployment check fails, the exit value is stored and the script continues. Read line 56:

```bash
export EXIT_STATUS=$(if [ "$CLUSTER_EXIT_CODE" -eq 1 ] || [ "$ASSISTED_EXIT_CODE" -eq 1 ]; then echo 1; else echo 0; fi)
exit $EXIT_STATUS
```

This stage's exit code is **decided at the end**, if one of the clustering stages fails. This means both stages are ran. The `kubectl get *` diagnostics in the middle provide a detailed look into the cluster (especially the `kubectl describe *` commands), allowing for detailed overview of the cluster state in case of test failure.

Having separate stages for different ways of clustering would be too much. Also, this test belongs into this suite, as it tests a specific use case - Kubernetes clustering using a generic assisted clustering H2OEmbeddeConfigProvider.

h2o-clustering/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Testing
+
+This module, besides its own test suite, is tested in Kubernetes environment inside the [h2o-k8s](../h2o-k8s/tests/clustering/README.md)
+module.
+
+## Cluster status endpoint
+
+There is an endpoint for H2O Cluster status provided by this module, located via `GET /cluster/status`. 
+If H2O has not clustered yet, the endpoint `HTTP 204` - No content, according to [RFC-2616](https://tools.ietf.org/html/rfc2616#section-10.2.5).
```
Pscheidl, line 46: No content is a clear indication the cluster has not been formed yet. Removes the necessity to parse the JSON on the client just to find out  the clustering status. And it also semantically makes sense - no cluster = no content on cluster status/health.

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
```diff
@@ -0,0 +1,74 @@
...
+        Log.info("Starting assisted clustering REST API services");
+        try {
+            final AssistedClusteringRestApi assistedClusteringRestApi = new AssistedClusteringRestApi(flatFileCallback);
+            assistedClusteringRestApi.setAsyncRunner(new GracefulAsyncRunner());
+            assistedClusteringRestApi.start();
+            Log.info("Assisted clustering REST API services successfully started.");
+            return Optional.of(assistedClusteringRestApi);
+        } catch (IOException e) {
+            Log.err("Unable to start H2O assisted clustering REST API", e);
+            System.exit(1);
```
michalkurka, line 51: use H2O.exit instead of System.exit to make sure logs are preserved

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfig.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public String fetchFlatfile() {
+        return flatfile;
+    }
+
+    @Override
+    public void exit(int status) {
+        System.exit(status);
```
michalkurka, line 39: H2O.exit instead?

h2o-clustering/src/main/java/water/clustering/AssistedClusteringEmbeddedConfigProvider.java
Pscheidl: Done.

h2o-clustering/src/main/java/water/clustering/api/AssistedClusteringEndpoint.java
```diff
@@ -0,0 +1,92 @@
...
+    private final AtomicBoolean flatFileReceived;
+    private final ExecutorService flatFileConsumerCallbackExecutor = Executors.newSingleThreadExecutor();
+
+    public AssistedClusteringEndpoint() {
+        flatFileReceived = new AtomicBoolean(false);
+    }
+
+    @Override
+    public String getText() {
+        throw new IllegalStateException(String.format("Method getText should not be called on '%s'",
```
honzasterba, line 41: H2O.unimpl maybe?

honzasterba (approved):
minor comments

h2o-clustering/README.md
```diff
@@ -0,0 +1,61 @@
...
+unhealthy nodes.
+
+Any node of the cluster can be queried and each will return the same response.
+
+
+Example `curl --location --request GET 'localhost:8080/cluster/status'`:
+
+```json
+{
+  "leader_node": "192.168.0.149:54321",
```
Pscheidl, line 57: @michalkurka @honzasterba  One quick thought before merging:

Do we want the port there ? By convention, the internal comms port will be +1. Not sure if this info will be useufl at the moment, but might be in future. So I'll keep it unless there is anotehr reason not to do so of course.

honzasterba: why not?

Pscheidl: No idea. I was just checking with other my own code. I'm all for it.
Let me fix the rest of the comments and merge.

Pscheidl:
Found a bug by actually using this: The REST API, after the cluster is formed, may return less H2O nodes in the healthy_cluster array. This is caused by immediate call by the client (<1ms). Iff delayed, cluster size is reported properly. Will be probably linked to slow H2O internal state updates. Todo: Check if number of healthy nodes can be made final before stating cluster is in "clustered" status.

h2o-k8s/tests/clustering/h2o-python-clustering-template.yaml
```diff
@@ -0,0 +1,74 @@
...
+ type: ClusterIP
+ clusterIP: None
+ selector:
+   app: h2o-assisted
+ ports:
+   - protocol: TCP
+     port: 80
+     targetPort: 54321
+---
+apiVersion: networking.k8s.io/v1beta1
```
mn-mikke, line 57: Why not `networking.k8s.io/v1`? OpenShift?

h2o-k8s/tests/clustering/h2o-assisted-template.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apps/v1
+kind: Deployment
```
mn-mikke, line 2: So H2O cluster won't use statefull-set anymore?

h2o-clustering/src/main/java/water/clustering/api/AssistedClusteringRestApi.java
```diff
@@ -0,0 +1,80 @@
...
+        } catch (NumberFormatException e) {
+            final String errorMessage = String.format("Unusable port for Assisted clustering REST API to bind to: '%s'", customPort);
+            throw new IllegalArgumentException(errorMessage, e);
+        }
+    }
+
+    @Override
+    public void addMappings() {
+        super.addMappings();
+        addRoute("/clustering/flatfile", AssistedClusteringEndpoint.class, this.flatFileConsumer);
```
mn-mikke, line 57: Just a question to think about (might be irrelevant)... Should this endpoint be secured? Couldn't an enemy call this endpoint with a flat file containing own node?

Pscheidl: Yes, that could happen.

Pscheidl: Can be improved in another iteration.

h2o-k8s/tests/clustering/h2o-python-clustering-template.yaml
Pscheidl: Yes.

h2o-k8s/tests/clustering/h2o-assisted-template.yaml
Pscheidl: Nope. Won't be necessary, ever. Not necessary for this test. One needs to look into the "nature of this test" a little bit more. I need two H2O pods deployed and cluster them in a very simple and straightforward wy using a python script. StatefulSet could be used as well, but it's too heavy and clever. For this test to `not pass` when required, a more stupid deployment is the perfect fit.

PR #5230, opened 2021-01-13
[test] add CV to GLM test to make the models train a little slower

honzasterba:


Pscheidl (approved):
Is this about expected test times ? If yes, then incorporating CV doesn't hurt anyway - always better to benchmark changes in CV as well.

honzasterba:
this is about being able to interrupt the grid before it finishes, since glm on small data trains very fast its sometimes difficult to catch

Pscheidl (approved):

PR #5231, opened 2021-01-14
PUBDEV-7748: interactions support in Target Encoding

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-7748

due to the addition of a new format for columns mapping, and per request from Sparkling Water, this also includes
https://h2oai.atlassian.net/browse/PUBDEV-7848

This PR includes the support for interactions during training and predict, as well as at MOJO level.

*API changes*
clients have one additional param named `columns_to_encode` defaulting to None/NULL.
- If not set, the Target Encoder keeps encoding all the categorical columns present in `x` (by default all cat columns in the training frame).
- if `columns_to_encode` is set, then TE will only encode the columns listed and/or encode the interaction column generated for each non-trivial group.
- if  `columns_to_encode` is set, all columns listed should be also listed in `x` (or in training frame by default if `x` is not set).

*API MOJO*
for simplicity of use by Sparkling Water, the `TargetEncodingMojoModel` exposes an non-official field named `_inoutMapping`: this field lists the encoded columns generated for each column or group of columns provided (through `x` or `columns_to_encode`). At MOJO level, this can also be used to identify the columns of the `TargetEncoderPrediction.transformations` (it corresponds to the inoutMapping items' `to` values in order of iteration).


sebhrusen:
@arnocandel feel free to test this branch for the big data recipe.

Py TargetEncoder has one new param named `columns_to_encode`, it takes a list of columns or groups of columns. In the case of groups, it will apply TE on  the interaction between those columns.
For example:
```
columns_to_encode = ['cat1', ['cat2', 'cat3'], ...]
```

Note that after discussion with @michalkurka, TE still only supports categorical columns, so if you need to create interactions between numerical columns and categorical ones, you'll have to apply binning first on the numerical column using `H2OFrame.cut` method in Py. See https://h2oai.atlassian.net/browse/PUBDEV-7748?focusedCommentId=57722

Looking forward for your feedback, thanks.

h2o-genmodel/src/main/java/hex/genmodel/algos/targetencoder/TargetEncoderMojoModel.java
```diff
@@ -14,12 +14,23 @@ public static double computeLambda(long nrows, double inflectionPoint, double sm
...
+    return nameToIdx;
+  }
+
 
   public final Map<String, Integer> _columnNameToIdx;
   public Map<String, Boolean> _teColumn2HasNAs; // tells if a given encoded column has NAs
   public boolean _withBlending;
   public double _inflectionPoint;
   public double _smoothing;
+  public List<ColumnsToSingleMapping> _inencMapping;
```
sebhrusen, line 32: @mn-mikke is it ok to have it in the MojoModel or would you prefer it in `TargetEncoderPrediction`? as a result of `EasyPredictModelWrapper.predictTargetEncoding`?

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
```diff
@@ -84,48 +85,79 @@ protected boolean defaultDropConsCols() {
...
 
     private IcedHashMap<String, Boolean> buildCol2HasNAsMap() {
       final IcedHashMap<String, Boolean> col2HasNAs = new IcedHashMap<>();
       for (Map.Entry<String, Frame> entry : _target_encoding_map.entrySet()) {
         String teColumn = entry.getKey();
         Frame encodingsFrame = entry.getValue();
-        boolean hasNAs = _parms.train().vec(teColumn).cardinality() < encodingsFrame.vec(teColumn).cardinality(); //XXX: _parms.train().vec(teColumn).naCnt() > 0 ?
+        int teColCard = _parms.train().vec(teColumn) == null ? -1 : _parms.train().vec(teColumn).cardinality();
+        // if teColumn is a transient (generated for interactions), it can't have NAs as they're all encoded.
+        boolean hasNAs = teColCard > 0 && teColCard < encodingsFrame.vec(teColumn).cardinality(); //XXX: _parms.train().vec(teColumn).naCnt() > 0 ?
```
Pscheidl, line 134: Forgotten comment ?

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
```diff
@@ -84,48 +85,79 @@ protected boolean defaultDropConsCols() {
 
     public final TargetEncoderParameters _parms;
     public final int _nclasses;
+    public final ColumnsToSingleMapping[] _input_to_encoding_column; // maps input columns (or groups of columns) to the single column being effectively encoded (= key in _target_encoding_map).
```
Pscheidl, line 88: Knit: These comments belong above `ColumnsToSingleMapping` and `ColumnsMapping` classes IMHO.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java
```diff
@@ -82,6 +73,90 @@ static double computePriorMean(Frame encodings, int targetClass) {
     }
   }
 
+  /**
+   * @return the index of the interaction column for the group, or the index of the column if the group has only one.
+   */
+  static int createFeatureInteraction(Frame fr, String[] colGroup) {
```
Pscheidl, line 79: Knit: Using Optional arguments and one method would be useful.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java
```diff
@@ -82,6 +73,90 @@ static double computePriorMean(Frame encodings, int targetClass) {
...
+  }
+
+  /**
+   * @param interactionDomain the domain of the generated interaction column, if already known, 
+   *                          for example when computing interaction for predictions.
+   * @return the index of the interaction column for the group, 
+   * or the index of the column if the group has only one, 
+   * or -1 if one column in the interaction group is missing.
+   */
+  static int createFeatureInteraction(Frame fr, String[] colGroup, String[] interactionDomain) {
```
Pscheidl, line 90: Knit: I don't like the idea that 80% of TE functionality is in TEHelper. There could easily be `TargetEncoderFeatureInteractions` class at least.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java
```diff
@@ -82,6 +73,90 @@ static double computePriorMean(Frame encodings, int targetClass) {
...
+    Vec interactionCol = new CreateInteractionTask(encoder, interactionDomain)
+            .doAll(new byte[] {interactionType}, interactingVecs)
+            .outputFrame(null, null, new String[][]{interactionDomain})
+            .lastVec();
+    if (interactionType != Vec.T_CAT)
+      interactionCol = VecUtils.toCategoricalVec(interactionCol); // the domain is obtained from CollectDoubleDomain, so it is sorted by numerical value, and then converted to String
+    return interactionCol;
+  }
+
+  private static class CreateInteractionTask extends MRTask<CreateInteractionTask> {
```
Pscheidl, line 127: Knit: Could be extracted to separate class.

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java
```diff
@@ -82,6 +73,90 @@ static double computePriorMean(Frame encodings, int targetClass) {
...
+    final InteractionsEncoder _encoder;
+    final long[] _interactionDomain; // sorted by construction (see createInteractionColumn), or null
+    
+    public CreateInteractionTask(InteractionsEncoder encoder,  String[] interactionDomain) {
+      _encoder = encoder;
+      _interactionDomain = interactionDomain == null ? null : Arrays.stream(interactionDomain).mapToLong(Long::parseLong).toArray();
```
Pscheidl, line 133: Each node could transfer the num-only domaing from strings to longs locally in the map method and drop it immediately as the `map` function ends.

Cons:
- Each node does the work

Pros: 
- Potentially long domain is not serialized and transfered
- Each node waits for the master node to do the calculation anyway + the serialization time

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java
```diff
@@ -82,6 +73,90 @@ static double computePriorMean(Frame encodings, int targetClass) {
...
+        for (int i=0; i<cs.length; i++) {
+          interactingValues[i] = cs[i].isNA(row) ? -1 : (int)cs[i].at8(row);
+        }
+        long val = _encoder.encode(interactingValues);
+        if (val < 0) {
+          nc.addNA();
+        } else if (_interactionDomain == null){
+          nc.addNum(val);
+        } else {
+          int catVal = Arrays.binarySearch(_interactionDomain, val);
```
Pscheidl, line 149: Could speed this up by creating a map of sorted indices to the actual values. No need to binary search afterwards, just access the value quickly.

Pscheidl (commented):
First round of reviewing. Mostly oriented myself in the code, found some knits and have some comments on the `TargetEncoderHelper`'s content - mainly performance-oriented comments.

I'll do one more review round oriented solely on the business logic and performance due to the scale of this PR.

h2o-extensions/target-encoder/src/test/java/ai/h2o/targetencoding/TargetEncoderMojoIntegrationTest.java
```diff
@@ -36,6 +35,7 @@
 import static ai.h2o.targetencoding.TargetEncoderHelper.addKFoldColumn;
 import static hex.genmodel.algos.targetencoder.TargetEncoderMojoModel.computeBlendedEncoding;
 import static hex.genmodel.algos.targetencoder.TargetEncoderMojoModel.computeLambda;
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 
 @RunWith(JUnitQuickcheck.class)
```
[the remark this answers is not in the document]
sebhrusen: This class was already using a runner:
```@RunWith(JUnitQuickcheck.class)```

So it would require some work to get some of the tests working.
What I could do probably, is extracting the tests that don't need quickcheck and move them to a different test class.

File: h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java (hunk 2 above)
Comment (sebhrusen) [reply h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java:134]: the comment was intended to clarify the line above: i'll inline it and rephrase it.

File: h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java (hunk 3 above)
Comment (sebhrusen) [reply h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java:88]: Well, those classes can be technically used for any kind of columns to column(s) mapping, it's only those specific instances that tell what we're mapping.

File: h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java (hunk 4 above)
Comment (sebhrusen) [reply h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java:79]: Not sure I understand what you mean: is it
```
static int createFeatureInteraction(Frame fr, String[] colGroup, String... interactionDomain)
```
If this is what you mean, then I really don't like it here because we expect the `interactionDomain` array being of same length as `colGroup`, and the signature above makes them look totally independant.

File: h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java (hunk 5 above)
Comment (sebhrusen) [reply h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java:90]: yes, not false… I followed the original code structure, but I could extract all MRTasks here into a separate package private class.

File: h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java (hunk 7 above)
Comment (sebhrusen) [reply h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java:133]: Here, only the `long[]` array is serialized, whereas with your suggestion, the `String[]` array would have to be sent to each node… I believe it's almost always faster with the first array.

File: h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java (hunk 8 above)
Comment (sebhrusen) [reply h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderHelper.java:149]: true, should mainly be useful for very large interaction domains: I'll try it.

File: h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java [hunk 10]
```diff
@@ -84,48 +85,79 @@ protected boolean defaultDropConsCols() {
...
-    public TargetEncoderOutput(TargetEncoder b, IcedHashMap<String, Frame> teMap) {
-      super(b);
-      _parms = b._parms;
-      _nclasses = b.nclasses();
+    
+    public TargetEncoderOutput(TargetEncoder te) {
+      this(te, new IcedHashMap<>(), new ColumnsToSingleMapping[]{});
+    }
+    
+    public TargetEncoderOutput(TargetEncoder te, IcedHashMap<String, Frame> teMap, ColumnsToSingleMapping[] columnsToEncodeMapping) {
```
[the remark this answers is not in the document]
sebhrusen: Ok, I see what you  mean by `Optional` parameter now.
However, in Java, as much as I see the use for `Optional` in return values (I think that `Optional` was inspired from FP languages like Haskell that has this concept of `Maybe` return values to be able to chain functions), I don't see the benefit as a parameter. Actually I think the good old method overload in Java is much more clear at saying what is optional or not and I've never seen `Optional` used this way before in JDK or famous libraries.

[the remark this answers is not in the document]
sebhrusen: From B. Goetz himself (yes, a true authority argument :), but that's before all because I agree), (https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555):

---
Of course, people will do what they want. But we did have a clear intention when adding this feature, and it was not to be a general purpose Maybe type, as much as many people would have liked us to do so. Our intention was to provide a limited mechanism for library method return types where there needed to be a clear way to represent "no result", and using null for such was overwhelmingly likely to cause errors.

For example, you probably should never use it for something that returns an array of results, or a list of results; instead return an empty array or list. **You should almost never use it as a field of something or a method parameter.**

**There's nothing wrong with Optional that it should be avoided, it's just not what many people wish it were, and accordingly we were fairly concerned about the risk of zealous over-use.**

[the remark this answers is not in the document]
Pscheidl: Sure, that's what I meant. It's up to you to decide. I'm not pushing it. Not a deal breaker. Deal breaker would be big inefficiency or bad business logic. :wink: 

Monkey see, monkey writes - when I see something during an interview, I point it out. Not everything is a must.

h2o-genmodel/src/main/java/hex/genmodel/algos/targetencoder/TargetEncoderMojoModel.java
mn-mikke: @sebhrusen It could stay here, but what mapping I'm supposed to use? `_inencMapping` or `_inoutMapping`?

sebhrusen: @mn-mikke thanks for pointing that out: I underlined the wrong attribute for you: I'm pretty sure that you need only `_inoutMapping` as its name suggests.

ledell:
Hey @sebhrusen, right now the auto-generated interaction column names use a `~` to separate the original colnames.  Is that an allowable character in a Pandas DataFrame?  In R, that's not allowed, so if you convert a TE-encoded H2OFrame to an R data.frame, it will do an auto-conversion.  

In H2OFrame via R (example colnames): 
```
[158] "apache_3j_diagnosis~icu_id_te"               
[159] "apache_3j_diagnosis~icu_type_te"             
[160] "apache_3j_diagnosis~icu_stay_type_te"        
[161] "apache_3j_diagnosis~icu_admit_source_te"     
[162] "apache_3j_diagnosis~hospital_admit_source_te"
[163] "apache_3j_diagnosis~gender_te"               
[164] "apache_3j_diagnosis~ethnicity_te"
```

After using `as.data.frame()` to convert to an R data.frame:
```
[158] "apache_3j_diagnosis_te"                      
[159] "apache_3j_diagnosis.icu_id_te"               
[160] "apache_3j_diagnosis.icu_type_te"             
[161] "apache_3j_diagnosis.icu_stay_type_te"        
[162] "apache_3j_diagnosis.icu_admit_source_te"     
[163] "apache_3j_diagnosis.hospital_admit_source_te"
[164] "apache_3j_diagnosis.gender_te"                   
```

Is there another unique separator character that we could use (that's supported by R)?  I appreciate that `~` is rare and thus easier to parse names afterwards...

sebhrusen:
@ledell I actually chose `~` for interactions bacause it's ASCII and I saw it used to represent this somewhere (can't remember where though), I can change it, but I don't know with what (I also thought about `*` before, or in the worst case `_x_`).
`pandas` accepts any character for columns names, so no issue there.

In R, looks like the conversion is a restriction imposed by H2O client, R itself accetps the caracter for column names, this works fine:
```
df <- data.frame(matrix(ncol = 3, nrow = 0))
colnames(df) <- c("a", "b", "a~b")
```

sebhrusen:
@ledell switching to `:` separator as it is the standard representation in R formulas and in the `Patsy` library for Python:
https://patsy.readthedocs.io/en/latest/overview.html
https://cran.r-project.org/doc/manuals/R-intro.html#Formulae-for-statistical-models

our `as.dataframe` converter still replaces the `:` with `.`, but this can be avoided using:
```
> df <- h2o.getFrame("Key_Frame__upload_ba7b5c21b20b27225417c861ecd64ea6.hex")
> df
  a b a:b
1 1 2   3
2 4 5   6

[2 rows x 3 columns] 
> as.data.frame(df)
  a b a.b
1 1 2   3
2 4 5   6

> as.data.frame(df, check.names=FALSE)
  a b a:b
1 1 2   3
2 4 5   6
```

Given the few characters allowed to get proper variable names in R (only alphanum + _ + . ), I don't see a nice way to use only those to represent interactions, it would make it ugly for all type of frames on all clients, just to please R dataframe when it doesn't even seem necessary to manipulate the latter, for example, I can still access columns:
```
> adf <- as.data.frame(df, check.names=FALSE)
> adf$a
[1] 1 4
> adf$`a:b`
[1] 3 6
```

ledell:
@sebhrusen Switching to `:` sounds good!  Thanks.

sebhrusen:
@ledell 
> Do you know why our R as.data.frame() switches to period for various characters? 

it uses `read.csv` internally, and the default behaviour of this R function is to convert column names to valid R variables, that's why we can use the `check.names=FALSE` passthrough paramter to `as.data.frame`.

>  I suppose that would be a breaking change if we changed the way that works

I guess so

> If you go the other way in Python (Pandas DF -> H2OFrame) we don't allow the `:`

I really don't know why we impose ourselves such useless restrictions that seems to be only aiming at preventing interoperability with the data science ecosystem.
Column names are just strings all the way down, from pandas to backend abd back to R, I don't see any reason to restrict ourselves there.
It probably comes from R which seems to have a preference for column names that can also be valid variables, but even this isn't strictly necessary as R accepts the backquote syntax for non-valid variable names, and doesn't need any special syntax when using the square brackets to access those columns.

What should I do now? Fix our pandas to H2OFrame converter? Get rid of the `:` separator and have a confusing `_` separator everywhere (also much more likely to conflict)? We're really good at creating useless difficulties just for ourselves.

ledell:
Hey @sebhrusen, I think the warning message is misplaced here.  I specified the `columns_to_encode` param (leaving `x` unset), but I am still getting a warning that the non-categorical columns will be removed from the TE process (none of these are present in my `columns_to_encode` arg).  It must get triggered anytime that `x` is not set, but if the user specifies `columns_to_encode` then this warning is not necessary/incorrect.

Example in R:

```
library(h2o)
h2o.init()

#Import the titanic dataset
f <- "https://s3.amazonaws.com/h2o-public-test-data/smalldata/gbm_test/titanic.csv"
titanic <- h2o.importFile(f)

# Set response column as a factor
response <- "survived"
titanic[response] <- as.factor(titanic[response])

# Split the dataset into train and test
seed <- 1234
splits <- h2o.splitFrame(titanic, seed = seed, ratios = c(0.8))

train <- splits[[1]]
test <- splits[[2]]

# For k_fold strategy we need to provide fold column
train$fold <- h2o.kfold_column(train, nfolds = 5, seed = seed)

# Choose which columns to encode
columns <- list('home.dest', 'cabin', 'embarked', c('home.dest', 'embarked'))

# Train a TE model
target_encoder <- h2o.targetencoder(training_frame = train,
                                    columns_to_encode = columns,
                                    y = "survived",
                                    fold_column = "fold",
                                    data_leakage_handling = "KFold",
                                    seed = seed)
```

Gives warning:

```
Warning message:
In .h2o.processResponseWarnings(res) :
  Removing non-categorical columns found in the list of encoded columns.[fare, ticket, name, pclass, sibsp, body, age, parch, boat].
```

ledell:
Hey @sebhrusen, there's a warning message that should not be happening.  I specified the `columns_to_encode` param (leaving `x` unset), but I am still getting a warning that the non-categorical columns will be removed from the TE process (none of these are present in my `columns_to_encode` arg).  The warning is too eager -- it's getting triggered any time that `x` is not set, but if the user specifies `columns_to_encode` then this warning is not necessary/incorrect.

Example in R:

```
library(h2o)
h2o.init()

#Import the titanic dataset
f <- "https://s3.amazonaws.com/h2o-public-test-data/smalldata/gbm_test/titanic.csv"
titanic <- h2o.importFile(f)

# Set response column as a factor
response <- "survived"
titanic[response] <- as.factor(titanic[response])

# Split the dataset into train and test
seed <- 1234
splits <- h2o.splitFrame(titanic, seed = seed, ratios = c(0.8))

train <- splits[[1]]
test <- splits[[2]]

# For k_fold strategy we need to provide fold column
train$fold <- h2o.kfold_column(train, nfolds = 5, seed = seed)

# Choose which columns to encode
columns <- list('home.dest', 'cabin', 'embarked', c('home.dest', 'embarked'))

# Train a TE model
target_encoder <- h2o.targetencoder(training_frame = train,
                                    columns_to_encode = columns,
                                    y = "survived",
                                    fold_column = "fold",
                                    data_leakage_handling = "KFold",
                                    seed = seed)
```

Gives warning:

```
Warning message:
In .h2o.processResponseWarnings(res) :
  Removing non-categorical columns found in the list of encoded columns.[fare, ticket, name, pclass, sibsp, body, age, parch, boat].
```

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
```diff
@@ -422,18 +472,34 @@ private void reorderColumns(Frame fr) {
     String[] toTheEnd = _parms.getNonPredictors();
     Map<String, Integer> nameToIdx = nameToIndex(fr);
     List<Integer> toAppendAfterNumericals = new ArrayList<>();
-    String[] columns = fr.names();
-    int[] newOrder = new int[columns.length];
+    String[] trainColumns = _parms.train().names();
```
mn-mikke, line 475: As discussed, `TargetEncoderModel` cannot have reference the training frame, otherwise SW won't be able to delete the training frame during the call of SW's `fit` method. This will cause a memory leak since a user can't access frames created inside SW's `fit` method.

mn-mikke (commented):
From SW prospective, the pr looks good, except:
- The reference in `TargetEncoderModel` to the training frame
- Many warning messages of the following characters are present on stdout:
```_train: Removing non-categorical columns found in the list of encoded columns.[VOL, ID, GLEASON, AGE, PSA]``` These columns are no specified in `columns_to_encode`.

File: h2o-genmodel/src/main/java/hex/genmodel/algos/targetencoder/TargetEncoderMojoModel.java [hunk 12]
```diff
@@ -14,12 +14,23 @@ public static double computeLambda(long nrows, double inflectionPoint, double sm
...
+  public List<ColumnsMapping> _inoutMapping;
```
mn-mikke, line 33: @sebhrusen Is there any chance I could get this piece of information also for TE Model (not MOJO) via REST API. I need need it for multinomial cases. (Need to get the list of output columns for an input feature)

sebhrusen: sure, i can make it accessible in the model output

sebhrusen: @mn-mikke this is now exposed in the model output as `input_to_output_columns`

h2o-extensions/target-encoder/src/main/java/ai/h2o/targetencoding/TargetEncoderModel.java
sebhrusen: @mn-mikke this should be fixed by now

h2o-genmodel/src/main/java/hex/genmodel/algos/targetencoder/TargetEncoderMojoModel.java
mn-mikke: Will try. Thanks a lot!

h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderModelOutputV3.java
```diff
@@ -1,7 +1,23 @@
...
+import water.api.API;
 import water.api.schemas3.ModelOutputSchemaV3;
+import water.api.schemas3.SchemaV3;
 
 
-public class TargetEncoderModelOutputV3 extends ModelOutputSchemaV3<TargetEncoderModel.TargetEncoderOutput, TargetEncoderModelOutputV3> { }
+public class TargetEncoderModelOutputV3 extends ModelOutputSchemaV3<TargetEncoderModel.TargetEncoderOutput, TargetEncoderModelOutputV3> {
+    
+    public static class ColumnsMappingV3 extends SchemaV3<ColumnsMapping, ColumnsMappingV3> {
+        String[] from;
```
mn-mikke, line 13: nit: This could be made public

h2o-extensions/target-encoder/src/main/java/hex/schemas/TargetEncoderModelOutputV3.java
```diff
@@ -1,7 +1,23 @@
...
+        String[] to;
+    }
+    
+    @API(help = "Mapping between input column(s) and their corresponding target encoded output column(s). " +
+            "Please note that there can be multiple columns on the input/from side if columns grouping was used, " +
+            "and there can also be multiple columns on the output/to side if the target was multiclass.",
+            direction = API.Direction.OUTPUT)
+    ColumnsMappingV3[] input_to_output_columns;
```
mn-mikke, line 21: nit: This could be made public

mn-mikke (approved):
LGTM!

PR #5233, opened 2021-01-14
PUBDEV-7937 - Expose most important reason codes in Py/R & MOJO

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-7937

More info in Jira.

edit:
Maybe the best to support review is to leave it as it is - without xboost - and do not combine the sorting and output format together in this PR. Current implementation works for Mojo (EasyPredictModelWrapper), DRF, and GBM. For xgboost model gives "Not Implemented yet".


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -17,6 +17,7 @@
 import hex.tree.xgboost.predict.*;
 import hex.tree.xgboost.util.PredictConfiguration;
 import hex.util.EffectiveParametersUtils;
+import org.apache.commons.lang.NotImplementedException;
```
michalkurka, line 20: I would use UnsupportedOperationException

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
```diff
@@ -1,5 +1,7 @@
 package hex.genmodel.attributes.parameters;
 
+import java.util.Comparator;
+
 public class KeyValue {
```
michalkurka, line 5: Does this need to be Serializable? Is it used in any Serializable class?

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
```diff
@@ -22,4 +24,34 @@ public double getValue() {
     public String toString() {
         return "{Key: " + key + ", Value: " + value + "}";
     }
+    
+    public static class AscComparator implements Comparator<KeyValue> {
```
michalkurka, line 28: I think if you have a specific functionality, it would be better to call the outer class something else, eg. FeatureContribution.

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -20,7 +22,9 @@
         P extends SharedTreeModel.SharedTreeParameters,
         O extends SharedTreeModel.SharedTreeOutput
         > extends SharedTreeModel<M, P, O> implements Model.Contributions {
-    
+
+  private transient final ContributionComposer contributionComposer = new ContributionComposer();
```
michalkurka, line 26: do you need this to be an instance variable? I think this should be avoided because you can lose the instance on serialization and deserialization if that ever happens

just make a new instance when you need it

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -30,20 +34,25 @@ public Frame scoreContributions(Frame frame, Key<Frame> destination_key) {
...
   }
 
-  @Override
-  public Frame scoreContributions(Frame frame, Key<Frame> destination_key, Job<Frame> j) {
-    if (_output.nclasses() > 2) {
-      throw new UnsupportedOperationException(
-              "Calculating contributions is currently not supported for multinomial models.");
-    }
-
+  private Frame cleanFrame(Frame frame) {
```
michalkurka, line 37: the columns that a removed are sometimes called "special columns") - I would just call the method removeSpecialColumns

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -53,14 +62,55 @@ public Frame scoreContributions(Frame frame, Key<Frame> destination_key, Job<Fra
...
+    for (int i = 0, topFeatureIterator = 1, bottomFeatureIterator = 1; i < outputSize; i+=2, topFeatureIterator++) {
+      if (topFeatureIterator <= topNAdjusted) {
+        names[i] = "top_feature_" + topFeatureIterator;
+        names[i+1] = "top_value_" + topFeatureIterator;
+      } else {
+        names[i] = "bottom_top_feature_" + bottomFeatureIterator;
+        names[i+1] = "bottom_top_value_" + bottomFeatureIterator;
+        bottomFeatureIterator++;
+      }
+      types[i] = Vec.T_STR;
```
michalkurka, line 91: try to save memory by using categorical output

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -53,14 +62,55 @@ public Frame scoreContributions(Frame frame, Key<Frame> destination_key, Job<Fra
...
+    final String[] contribNames = ArrayUtils.append(adaptFrm.names(), "BiasTerm");
+
+    int topNAdjusted = contributionComposer.checkAndAdjustInput(topN, adaptFrm.names().length);
+    int topBottomNAdjusted = contributionComposer.checkAndAdjustInput(topBottomN, adaptFrm.names().length);
+
+    int outputSize = Math.min((topNAdjusted+topBottomNAdjusted)*2, adaptFrm.names().length*2);
+    String[] names = new String[outputSize];
+    byte[] types = new byte[outputSize];
+
+    for (int i = 0, topFeatureIterator = 1, bottomFeatureIterator = 1; i < outputSize; i+=2, topFeatureIterator++) {
```
michalkurka, line 82: I think this for loop is too complex for me to understand :) it seems like 2 for-loops collapsed into one

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
```diff
@@ -135,6 +139,15 @@ protected ModelMetricsListSchemaV3 schema(int version) {
             "specifying a Compact output format will produce a per-feature contribution", values = {"Original", "Compact"}, json = false)
     public Model.Contributions.ContributionsOutputFormat predict_contributions_output_format;
 
+    @API(help = "Predict the feature contributions - Shapley values - return top_n highest", json = false)
```
michalkurka, line 142: it is not very clear these belong to "predict contributions"

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
```diff
@@ -402,10 +419,14 @@ public void compute2() {
             throw new H2OIllegalArgumentException("Model type " + parms._model._parms.algoName() + " doesn't support calculating Feature Contributions.");
           }
           Model.Contributions mc = (Model.Contributions) parms._model;
-          Model.Contributions.ContributionsOutputFormat outputFormat = null == s.predict_contributions_output_format ?
-                  Model.Contributions.ContributionsOutputFormat.Original : s.predict_contributions_output_format;
-          Model.Contributions.ContributionsOptions options = new Model.Contributions.ContributionsOptions().setOutputFormat(outputFormat);
-          mc.scoreContributions(parms._frame, Key.make(parms._predictions_name), j, options);
+          if (parms._top_n == 0 && parms._top_bottom_n == 0 && !parms._abs) {
```
michalkurka, line 422: this would be more elegant if you used ContributionsOptions - then you wouldn't need 2 branches of execution

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
```diff
@@ -682,6 +683,11 @@ public Frame scoreContributions(Frame frame, Key<Frame> destination_key, Job<Fra
             .outputFrame(destination_key, outputNames, null);
   }
 
+  @Override
+  public Frame scoreContributions(Frame frame, Key<Frame> destination_key, int topN, int topBottomN, boolean abs, Job<Frame> j) {
+    throw new NotImplementedException("Sorting of shap for XGBooost is not yet implemented");
```
michalkurka, line 688: why not?

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ContributionComposer.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param contribNames Contribution corresponding feature names
+     * @param topN Return only #topN highest contributions + bias.
+     * @param topBottomN Return only #topBottomN lowest contributions + bias
+     *                   If topN and topBottomN are defined together then return array of #topN + #topBottomN + bias
+     * @param abs True to compare absolute values of contributions
+     * @return Sorted KeyValue array of contributions of size #topN + #topBottomN + bias
+     *         If topN < 0 || topBottomN < 0 then all descending sorted contributions is returned.
+     */
+    public final KeyValue[] composeContributions(float[] contribs, String[] contribNames, int topN, int topBottomN, boolean abs) {
+        if (topBottomN == 0) {
```
michalkurka, line 25: why all these special cases? can the code be simplified?

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ContributionComposer.java
```diff
@@ -0,0 +1,66 @@
...
+            return composeSortedContributions(contribs, contribNames, topN, new KeyValue.DescComparator(abs));
+        } else if (topN == 0){
+            return composeSortedContributions(contribs, contribNames, topBottomN, new KeyValue.AscComparator(abs));
+        } else if ((topN + topBottomN) >= contribs.length || topN < 0 || topBottomN < 0) {
+            return composeSortedContributions(contribs, contribNames, contribs.length, new KeyValue.DescComparator(abs));
+        }
+        
+        KeyValue[] topSorted = composeSortedContributions(contribs,contribNames, topN, new KeyValue.DescComparator(abs));
```
michalkurka, line 33: can we sort everything just once? and then pick the result?

also it seems like you are operating with n*logn complexity but the problem only needs n*logk (where k is the "topN"); probably not a big deal since n won't be crazy big

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
valenad1: We have a test which failed when ContributionComposer did not implement Serializable. In this case it seems ok but its not harm to implement Serializable so I'll do it

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
valenad1: This class is used for MojoDeserialization, Variable importance sorting and Contributions sorting. I think there nothing special, just pair of string-double values and couple of comparators to sort by values

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
valenad1: Fixed

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
valenad1: Renamed

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
michalkurka: oh, sorry, didn't notice you are actually reusing something that is used for attribute parsing

In that case, using this code for algo execution is not advised. I suggest creating a new class that addresses your specific functionality, eg. FeatureContributions.

At least your comparators definitely don't belong here.

valenad1: All right, thanks, I'll do it

michalkurka: thanks, the classes meant for parsing of info from bundle modeldetails.json should be independent of the regular scoring logic 

maybe it would be better if it was a separate module but that would make build more complicated - something to thing about

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
valenad1: Reworded

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
valenad1: I just noticed that the field is transient, its not harm to make instance if its needed

michalkurka: @valenad1 was it fixed or was it not fixed? Is still see the transient field defined in this class.

Please read my original comment. You are defining a transient field on a class that can be serialized. In that case, you will lose the value of the instance variable if serialized and deserialized.

valenad1: Yes, it is fixed in 66fecf09d262fb1ab87ea714000854ab42400e6c

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -3,6 +3,7 @@
 import hex.genmodel.algos.tree.SharedTreeNode;
 import hex.genmodel.algos.tree.SharedTreeSubgraph;
 import hex.genmodel.algos.tree.*;
+import hex.genmodel.attributes.parameters.KeyValue;
```
michalkurka, line 6: it seems like this is still used, classes in `hex.genmodel.attributes` shouldn't be used in scoring logic

valenad1: I agree, thanks I am working on separating of scoring logic from genmodel class.

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ContributionComposer.java
valenad1: > Can we sort everything just once? and then pick the result?

Yes, I fixed the double sorting.

valenad1: > also it seems like you are operating with nlogn complexity but the problem only needs nlogk (where k is the "topN"); probably not a big deal since n won't be crazy big

In a nutshell, I agree that we can be more clever here but IMHO it is enough.

I am operating with nlogn. I found that best known algorithm has linear time. Algorithm with Heap could be nlogk, but AFAIK you have to run it twice when you want to have #topN and #topBottomN. I rather choose a KISS principle and use java build-in sorting and provide a clear interface for sorting functionality in case of future replacement of this part of algorithm.

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
valenad1: Implemented

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
valenad1: True, I replaced it with three nice to read for loops

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
valenad1: Done

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
valenad1: Separated

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -3,6 +3,7 @@
 import hex.genmodel.algos.tree.SharedTreeNode;
 import hex.genmodel.algos.tree.SharedTreeSubgraph;
 import hex.genmodel.algos.tree.*;
+import hex.genmodel.attributes.parameters.Pair;
```
michalkurka, line 6: please don't put any scoring related logic into hex.genmodel.attributes package; this belongs in algos

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -98,25 +171,61 @@ public void map(Chunk chks[], NewChunk[] nc) {
...
-        }
-        for (int i = 0; i < contribs.length; i++) {
-          contribs[i] = 0;
-        }
-
+        fillInput(chks, row, input, contribs);
         // calculate Shapley values
         _treeSHAP.calculateContributions(input, contribs, 0, -1, workspace);
-        
+        doModelSpecificComputation(contribs);
```
michalkurka, line 177: this is never called because you are overriding method map(Chunk chks[], NewChunk[] nc)

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -98,25 +171,61 @@ public void map(Chunk chks[], NewChunk[] nc) {
...
         // Add contribs to new chunk
         addContribToNewChunk(contribs, nc);
       }
     }
 
+    protected void doModelSpecificComputation(float[] contribs) {/*For children*/}
```
michalkurka, line 183: doesn't need to be defined here

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/FeatureContribution.java
```diff
@@ -0,0 +1,27 @@
...
+        this.columnId = columnId;
+        this.shapleyContribution = shapleyContribution;
+    }
+
+    public Object getKey() {
+        return columnId;
+    }
+
+    public Double getValue() {
+        return shapleyContribution;
```
michalkurka, line 20: this is bad, unnecessary autoboxing

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ContributionComposer.java
```diff
@@ -0,0 +1,80 @@
...
+        if (nAdjusted < contribs.length) {
+            Pair<?,?> bias = sortedContributions[contribs.length-1];
+            sortedContributions = Arrays.copyOfRange(sortedContributions, 0, nAdjusted + 1);
+            sortedContributions[nAdjusted] = bias;
+        }
+        return sortedContributions;
+    }
+    
+    private Pair<?,?>[] sortContributions(float[] contribs, Object[] contribNames, Comparator<? super Pair<?,Double>> comparator) {
+        Pair<?,Double>[] sorted = new FeatureContribution[contribs.length];
```
michalkurka, line 63: I think that creating a new object for each value of the row is expensive.

You could definitely save some memory allocation by using water.ArrayUtils 

   public static void sort(final int[] idxs, final double[] values)
   
or one of its variants.

michalkurka: IMHO this needs to be redone for sure.

h2o-genmodel/src/main/java/hex/genmodel/attributes/VariableImportances.java
```diff
@@ -2,13 +2,12 @@
 
 
 import com.google.gson.JsonObject;
+import hex.genmodel.attributes.comparators.AscPairComparator;
```
michalkurka, line 5: do not use hex.genmodel.attributes package

h2o-genmodel/src/main/java/hex/genmodel/attributes/VariableImportances.java
```diff
@@ -56,12 +55,7 @@ protected static VariableImportances extractFromJson(final JsonObject modelJson)
         for (int i = 0; i < _importances.length; i++) {
             sortedImportances[i] = new KeyValue(_variables[i], _importances[i]);
         }
-        Arrays.sort(sortedImportances, new Comparator<KeyValue>() {
```
michalkurka, line 59: I would leave as is

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
```diff
@@ -1,6 +1,8 @@
 package hex.genmodel.attributes.parameters;
 
-public class KeyValue {
+import java.io.Serializable;
+
+public class KeyValue implements Serializable, Pair<String, Double> {
```
michalkurka, line 5: please do not modify, no reason to do so

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/Pair.java
```diff
@@ -0,0 +1,6 @@
+package hex.genmodel.attributes.parameters;
+
+public interface Pair<K, V> {
```
michalkurka, line 3: not needed, you can always generalize later - currently there is no need to do that

this interface is actually harmful - it is making you do autoboxing of Doubles to double in your API

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -347,7 +350,7 @@ public EasyPredictModelWrapper(GenModel model) {
    * For specific applications (where the kind of model is known and doesn't change), it is recommended to call
    * specific prediction calls like predictBinomial() directly.
    *
-   * @param data A new data point.
+   * @param data A new data point. Unknown or missing column name is treated as a NaN or ignored. Column names are case sensitive.
```
michalkurka, line 353: thanks for the docuchanges!

michalkurka:
@valenad1 I can tell a lot of thought and work went into this PR, I appreciate the thorough test.

I would say improvements still need to be made. Clean-up the APIs (no Pair) to avoid unnecessary autoboxing and autounboxing of double values is suggested.

h2o-core/src/main/java/water/util/ArrayUtils.java
```diff
@@ -101,6 +101,17 @@ public static double innerProductPartial(double [] x, int[] x_index, double [] y
     return result;
   }
 
+  /**
+   * @return Index of element val or -1 if not found. Method is iterating through arr and found value in liner time.
+   */
+  public static<T extends Comparable<T>> int indexOfLinear(T[] arr, T val) {
```
michalkurka, line 107: unused - seems to be used only in tests

valenad1: Removed

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ContributionComposer.java
valenad1: Thank you for the clarification

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
valenad1: Thanks!

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
valenad1: Done

h2o-genmodel/src/main/java/hex/genmodel/attributes/parameters/KeyValue.java
valenad1: KeyValue class is untouched

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostModel.java
valenad1: I'll do it in scope of this task https://h2oai.atlassian.net/browse/PUBDEV-8084

valenad1: https://github.com/h2oai/h2o-3/pull/5491

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
```diff
@@ -3,6 +3,7 @@
 import hex.*;
 import hex.genmodel.utils.DistributionFamily;
 import org.apache.commons.lang.ArrayUtils;
+import org.apache.commons.lang.NotImplementedException;
```
michalkurka, line 6: I would rather use UnsupportedOperationException. It is not common to use the commons.lang version

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
```diff
@@ -136,6 +140,15 @@ protected ModelMetricsListSchemaV3 schema(int version) {
             "specifying a Compact output format will produce a per-feature contribution", values = {"Original", "Compact"}, json = false)
     public Model.Contributions.ContributionsOutputFormat predict_contributions_output_format;
 
+    @API(help = "Only for predict_contributions function - sort Shapley values and return top_n highest (optional)", json = false)
+    public int top_n;
+
+    @API(help = "Only for predict_contributions function - sort Shapley values and return top_bottom_n lowest (optional)", json = false)
+    public int top_bottom_n;
```
michalkurka, line 147: top_bottom_n - that seems confusing to me, why not just bottom_n? isn't there already a naming in another our product how to call this?

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
```diff
@@ -136,6 +140,15 @@ protected ModelMetricsListSchemaV3 schema(int version) {
...
+
+    @API(help = "Only for predict_contributions function - sort absolute Shapley values (optional)", json = false)
+    public boolean abs;
```
michalkurka, line 150: I think this is missing a verb, how about "compare_abs" - is there a precedence of calling this? this must be used somewhere, not necessarily just for contributions

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ContributionsPredictor.java
```diff
@@ -1,12 +1,16 @@
...
+import hex.genmodel.attributes.parameters.FeatureContribution;
 import hex.genmodel.utils.ArrayUtils;
 
+import java.util.Arrays;
+
 public abstract class ContributionsPredictor<E> implements PredictContributions {
   private final int _ncontribs;
   private final String[] _contribution_names;
   private final TreeSHAPPredictor<E> _treeSHAPPredictor;
+  private final ContributionComposer contributionComposer = new ContributionComposer();
```
michalkurka, line 13: Does this need to be instance variable? It doesn't add any value if it is right now - if it wasn't it wouldn't need to be Serializable - one less problem to worry about.

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ContributionsPredictor.java
```diff
@@ -41,5 +45,17 @@ private Object getWorkspace() {
     }
     return workspace;
   }
+
+  @Override
+  public FeatureContribution[] calculateContributions(double[] input, int topN, int topBottomN, boolean abs) {
+    float[] contributions = calculateContributions(input);
+    int[] contributionNameIds = ArrayUtils.range(0, _contribution_names.length -1);
+    int[] sorted = contributionComposer.composeContributions(contributionNameIds, contributions, topN, topBottomN, abs);
```
michalkurka, line 53: just call new ContributionComposer here

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
valenad1: I was also struggling how to name it, bottom_n I am not aware of any similar naming in other product

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
valenad1: I'll try to find similar atribute

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ContributionsPredictor.java
valenad1: True, it can be only instance or the methods could be static, maybe it is the best solution

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
valenad1: Its forgotten import.

Removed.

h2o-genmodel/src/main/java/hex/genmodel/algos/tree/ContributionsPredictor.java
valenad1: I made it only instance variable

michalkurka (approved):
Great work @valenad1 , than you for the contribution

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
valenad1: I am not aware of the similar attribute in h2o-3. 

- The absolute value in python is `abs(number)` -> that is why I use "abs_val" for both Python and R API
- In numpy it is `np.absolute(x)`
- In R it is `abs(value)`

h2o-core/src/main/java/water/api/ModelMetricsHandler.java
valenad1: I'll rename it to `bottom_n`

PR #5234, opened 2021-01-14
[PUBDEV-7859] grid.resume()

honzasterba:
- make grid resumable without providing any extra params
- this means all necessary config and params must be stored with the grid
- another piece to the puzzle of fault tolerance

### DIFF
--- a/h2o-algos/src/test/java/hex/grid/SequentialWalkerTest.java
+++ b/h2o-algos/src/test/java/hex/grid/SequentialWalkerTest.java
@@ -14,11 +14,9 @@ import water.fvec.Frame;
 import water.runner.CloudSize;
 import water.runner.H2ORunner;
 
-import java.util.Arrays;
-import java.util.Comparator;
+import java.util.*;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 import static water.TestUtil.parse_test_file;
 
 @RunWith(H2ORunner.class)
@@ -73,6 +71,27 @@ public class SequentialWalkerTest {
         } finally {
             Scope.exit();
         }
+    }
+
+    @Test
+    public void test_SequentialWalker_getHyperParams() {
+        GBMModel.GBMParameters gbmParameters = new GBMModel.GBMParameters();
+        Object[][] hyperParams = new Object[][] {
+            new Object[] {"_learn_rate", "_score_tree_interval"},
+            new Object[] {       0.5   ,                     2 },
+            new Object[] {       0.2   ,                     4 },
+            new Object[] {       0.1   ,                     8 }
+        };
+        SequentialWalker walker = new SequentialWalker<>(
+            gbmParameters,
+            hyperParams,
+            new GridSearch.SimpleParametersBuilderFactory<>(),
+            new HyperSpaceSearchCriteria.SequentialSearchCriteria()
+        );
+        Map<String, Object[]> exp = new HashMap<>();
+        assertEquals(new HashSet<>(Arrays.asList("_learn_rate", "_score_tree_interval")), walker.getHyperParams().keySet());
+        assertArrayEquals(new Object[] { 0.5, 0.2, 0.1 }, (Object[]) walker.getHyperParams().get("_learn_rate"));
+        assertArrayEquals(new Object[] { 2, 4, 8}, (Object[]) walker.getHyperParams().get("_score_tree_interval"));
     }
     
     @Test
--- a/h2o-core/src/main/java/hex/grid/HyperParameters.java
+++ b/h2o-core/src/main/java/hex/grid/HyperParameters.java
@@ -1,0 +1,79 @@
+package hex.grid;
+
+import water.AutoBuffer;
+import water.H2O;
+import water.Iced;
+
+import java.util.Map;
+import java.util.TreeMap;
+
+/**
+ * Wrapper class to make hyper-parameters serializable with Icer
+ */
+public class HyperParameters extends Iced<HyperParameters> {
+
+    private volatile Map<String, Object[]> values;
+
+    public HyperParameters(Map<String, Object[]> values) {
+        this.values = values;
+    }
+
+    public Map<String, Object[]> getValues() {
+        return values;
+    }
+
+    public final AutoBuffer write_impl(AutoBuffer ab) {
+        writeHyperParamsMap(ab, values);
+        return ab;
+    }
+
+    private void writeHyperParamsMap(AutoBuffer ab, Map<String, Object[]> params) {
+        ab.putInt(params.keySet().size());
+        for (String key : params.keySet()) {
+            ab.putStr(key);
+            Object[] vals = params.get(key);
+            if (vals.length > 0 && vals[0] instanceof Map) {
+                ab.putInt(vals.length);
+                for (int j = 0; j < vals.length; j++) {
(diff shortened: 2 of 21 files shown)


h2o-core/src/main/java/hex/grid/HyperParameters.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.util.Map;
+import java.util.TreeMap;
+
+/**
+ * Wrapper class to make hyper-parameters serializable with Icer
+ */
+public class HyperParameters extends Iced<HyperParameters> {
+
+    private volatile Map<String, Object[]> values;
```
michalkurka, line 15: volatile? Maybe transient instead? to make it clear it is not serialized?

h2o-algos/src/test/java/hex/grid/SequentialWalkerTest.java
```diff
@@ -74,6 +72,27 @@ public void test_SequentialWalker() {
             Scope.exit();
         }
     }
+
+    @Test
+    public void test_SequentialWalker_getHyperParams() {
```
michalkurka, line 77: does the resume work also with parallel grid search?

h2o-py/h2o/grid/grid_search.py
```diff
@@ -354,6 +354,22 @@ def train(self, x=None, y=None, training_frame=None, offset_column=None, fold_co
         parms["x"] = x
         self.build_model(parms)
 
+    def resume(self, recovery_dir=None):
```
michalkurka, line 357: Very cool!

h2o-py/h2o/grid/grid_search.py
```diff
@@ -354,6 +354,22 @@ def train(self, x=None, y=None, training_frame=None, offset_column=None, fold_co
...
+        """
+        Resume previously stopped grid training.
+
+        :param recovery_dir: When specified, the grid and all necessary data (frames, models) will be saved to this
```
Pscheidl, line 361: Why'd recovery dir be defined on resume ? Shouldn't that be part of GS initial params ?

Pscheidl (commented):
- I expected a different design. Expected H2O to have a default "dump folder" where everything is saved automatically and that dump folder is checked on startup for any checkpoints and other configuration. Such a folder could be configurable of course. And it's usage. For starters, it could be enabled on K8S only via a dedicated envvar. I'm just curious why didn't we go this way ?

- Resumes (not autorestart) are testable via JUnit - the deserialized objects can be compared easily with the ones before the algorithm was interrupted. This would test the correctness of the serialization part.

- Should work with parallel grid search nicely, as each model has a separate file. Only tests are missing - would be nice to have one as well.

honzasterba:
> * I expected a different design. Expected H2O to have a default "dump folder" where everything is saved automatically

... yes, this is the next stel in the process, I have the code mostly ready but did not want to make this PR even larger

> * Should work with parallel grid search nicely, as each model has a separate file. Only tests are missing - would be nice to have one as well.

... will add

h2o-py/h2o/grid/grid_search.py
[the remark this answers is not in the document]
honzasterba: out of scope of this PR, this PR just aims to make grid resumable without the need to provide any additional info on resume i.e. the method `resume()`, for user triggered resume it is necessary to indicate whether the resumed grid should also be recoverable since the circumstances might have changed since the last attempt

h2o-core/src/main/java/hex/grid/HyperParameters.java
honzasterba: good catch, fixed

h2o-algos/src/test/java/hex/grid/SequentialWalkerTest.java
honzasterba: modified the test to test this as well, actually discovered a bug, in parallel grid search resume code

michalkurka (approved):

h2o-core/src/main/java/hex/grid/HyperParameters.java
```diff
@@ -0,0 +1,79 @@
...
+import water.H2O;
+import water.Iced;
+
+import java.util.Map;
+import java.util.TreeMap;
+
+/**
+ * Wrapper class to make hyper-parameters serializable with Icer
+ */
+public class HyperParameters extends Iced<HyperParameters> {
```
Pscheidl, line 13: Should we also store version of H2O the parameters are intended to be users with. ? What if someone upgrades H2O and then attempts to load old GS ?

Pscheidl:
GS saves models as they're built. If a new folder is specified in resume function, shouldn't the content of the old folder be copied to the new one ?

h2o-r/tests/testdir_algos/grid/runit_grid_recovery_dir.R
```diff
@@ -22,6 +22,13 @@ test.grid.resume <- function() {
   # check the recovery dir is empty after grid success
   print(list.files(export_dir))
   expect_true(length(list.files(export_dir)) == 0)
+  grid_path <- h2o.saveGrid(export_dir, grid_id, save_params_references=TRUE)
+  h2o.removeAll()
+
+  grid <- h2o.loadGrid(grid_path, load_params_references = TRUE)
+  h2o.resumeGrid(grid@grid_id)
```
Pscheidl, line 29: Test resume with new folder as well ?

h2o-hadoop-common/tests_fault_tolerance/python/test_grid_reload.py
```diff
@@ -60,12 +60,10 @@ def test_frame_reload(self):
             cluster_2 = utils.start_cluster(cluster_2_name)
             h2o.connect(url=cluster_2)
             loaded = h2o.load_grid("%s/%s" % (work_dir, grid_id), load_params_references=True)
-            loaded_train = h2o.H2OFrame.get_frame(train.frame_id)
             print("models after first run:")
             for x in sorted(loaded.model_ids):
                 print(x)
-            loaded.hyper_params = hyper_parameters
-            loaded.train(x=list(range(4)), y=4, training_frame=loaded_train)
+            loaded.resume()
```
Pscheidl, line 66: Test with the new target folder as well ?

h2o-core/src/main/java/hex/grid/HyperParameters.java
honzasterba: this is stored with Icer, that guarantees that only the same version of H2O will be able to load it

h2o-r/tests/testdir_algos/grid/runit_grid_recovery_dir.R
honzasterba: good idea, it is difficult to test with the current code and it would a lot of wasted work, since my next PR will make this much easier. in my next PR I will add such tests

honzasterba:
> GS saves models as they're built. If a new folder is specified in resume function, shouldn't the content of the old folder be copied to the new one ? ( What if there will be another interruption, will H2O be able to load all the models from the new folder ?

very good catch, will work on that

h2o-r/tests/testdir_algos/grid/runit_grid_recovery_dir.R
Pscheidl: Agreed.

Pscheidl (approved):

PR #5238, opened 2021-01-15
Add a missing DKV.put to GLMBasicTestMultinomial

michalkurka:
The test does this:

    _covtype = parse_test_file("smalldata/covtype/covtype.20k.data");
    _covtype.replace(_covtype.numCols()-1,_covtype.lastVec().toCategoricalVec()).remove();

Updating an object, however, it is failing to also update the state in DKV.

This is missing:

    DKV.put(_covtype);

Now, why did it work before my commit/fix? My fix addressed a race
condition in read-unlock. The read-unlock was previously changing
the outside state of the world and updating the Lockable entity
(in this case a Frame) in DKV. That means read_lock; .. unlock; actually
did the missing DKV.put!


sebhrusen (approved):
at least it's the test that was wrong here.
But it also tell us which methods are more likely to be used incorrectly, like this `frame.replace` method in a certain context.
It's also something we could communicate to 3rd parties: a list of code smells.

tomasfryda (approved):
Thank you @michalkurka

michalkurka:
@sebhrusen @tomasfryda thank you for the review and discussion

PR #5239, opened 2021-01-18
PUBDEV-7962: Fix SE's wrong fold column handling

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7962

Since we don't use `fold_column` in SE, but we use `metalearner_fold_column` I adjusted the calculation of `nfeatures` to correspond with the actual number of features. This caused problems with DeepLearning MOJO used within StackedEnsemble MOJO.


michalkurka (approved):

sebhrusen (approved):
Looks good, thanks.

PR #5240, opened 2021-01-18
SW-2460 CoxPH to have ModelMetrics computed for all the trainings

satai:


h2o-algos/src/main/java/hex/coxph/CoxPHModel.java
```diff
@@ -48,6 +57,12 @@
 
     public boolean _calc_cumhaz = true; // support survfit
 
+    /**
+     * If true, computation is performed with jobs that
```
honzasterba, line 61: incomplete sentence

h2o-algos/src/main/java/hex/coxph/CoxPHModel.java
```diff
@@ -32,7 +32,16 @@
...
+    /**
+     * True if model scoring should not be performed during the computation.
+     * This setting is mainly to support tests scenarios with {@link #_single_node_mode} set
+     * to true.
+     * 
+     * If true, no {@link CoxPHOutput#_training_metrics} is not computed and it's null.
+     */
+    boolean _skip_scoring = false;
+    
+    public enum CoxPHTies { efron, breslow;}
```
honzasterba, line 44: extra ;

h2o-algos/src/main/java/hex/coxph/CoxPHModel.java
```diff
@@ -32,7 +32,16 @@
...
 
-    public enum CoxPHTies { efron, breslow }
+    /**
+     * True if model scoring should not be performed during the computation.
+     * This setting is mainly to support tests scenarios with {@link #_single_node_mode} set
+     * to true.
+     * 
+     * If true, no {@link CoxPHOutput#_training_metrics} is not computed and it's null.
+     */
+    boolean _skip_scoring = false;
```
honzasterba, line 42: do you plan to expose this via py/R?

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
```diff
@@ -264,12 +276,15 @@ public void testCoxPHSingleNodeMode() {
       parms._response_column = "event";
       parms._ignored_columns = new String[]{"id", "year", "surgery", "transplant"};
       parms._ties = CoxPHModel.CoxPHParameters.CoxPHTies.efron;
+      parms._skip_scoring = true;
```
honzasterba, line 279: this just makes the test pass? why is it necessary?

michalkurka: +1, should be explained here

h2o-algos/src/main/java/hex/coxph/CoxPH.java
```diff
@@ -657,9 +658,16 @@ public void computeImpl() {
           calcCumhaz_0(model, coxMR);
         }
 
-        if (iterTimer != null)
+        if (iterTimer != null) {
           Log.info("CoxPH Last Iteration: " + iterTimer.toString());
 
+        }
+       
+        if (!_parms._skip_scoring) {
```
michalkurka, line 666: it would be cleaner if the model was updated in DKV before scoring is invoked

h2o-algos/src/main/java/hex/coxph/CoxPH.java
```diff
@@ -579,6 +579,7 @@ public void computeImpl() {
         // The model to be built
         CoxPHModel.CoxPHOutput output = new CoxPHModel.CoxPHOutput(CoxPH.this, dinfo._adaptedFrame, train(), strataMap);
         model = new CoxPHModel(_result, _parms, output);
+        model.setInputParms(_input_parms);
```
michalkurka, line 582: I wonder why do you need this

h2o-algos/src/main/java/hex/coxph/CoxPHModel.java
satai: I don't think it's necessary right now. It' mainly intended for java unit tests.

h2o-algos/src/main/java/hex/coxph/CoxPH.java
satai: Because it's not set there in any other place for CoxPH as far I was searching.

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
satai: Explained.

h2o-algos/src/main/java/hex/coxph/CoxPH.java
honzasterba: this is done in ModelBuilder#computeParameters and should also work for cox, this should not be needed

h2o-algos/src/main/java/hex/coxph/CoxPHModel.java
honzasterba: just out of curiosity, why is this needed for tests?

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
honzasterba: The way I undestand this is that when user uses the params & input above, the algo will crash, is that correct? if yes, I dont thinkg that is right and needs to be fixed

satai: No, the algo will not crash in such a situation.

What caused crash in the test is `sys.ai.h2o.debug.checkRunLocal` system property used in this test, that probably shouln't be used in prod.
I made `_skip_scoring` package private to make it less likely to be used by accident.

h2o-algos/src/main/java/hex/coxph/CoxPH.java
satai: Removed

h2o-algos/src/main/java/hex/coxph/CoxPH.java
satai: Added

honzasterba (approved):

h2o-algos/src/main/java/hex/coxph/CoxPH.java
```diff
@@ -657,9 +657,20 @@ public void computeImpl() {
           calcCumhaz_0(model, coxMR);
         }
 
-        if (iterTimer != null)
+        if (iterTimer != null) {
           Log.info("CoxPH Last Iteration: " + iterTimer.toString());
 
+        }
+        
+        final boolean _skip_scoring = Boolean.parseBoolean(
```
michalkurka, line 665: Use H2O.getSysBoolProperty instead

h2o-algos/src/main/java/hex/coxph/CoxPH.java
```diff
@@ -657,9 +657,20 @@ public void computeImpl() {
...
+                System.getProperty("sys.ai.h2o.debug.skipScoring", Boolean.FALSE.toString())); 
+        
+        if (!_skip_scoring) {
+          model.update();
```
michalkurka, line 669: missing a job parameter

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
```diff
@@ -264,12 +276,18 @@ public void testCoxPHSingleNodeMode() {
       parms._response_column = "event";
       parms._ignored_columns = new String[]{"id", "year", "surgery", "transplant"};
       parms._ties = CoxPHModel.CoxPHParameters.CoxPHTies.efron;
+      
+      // Concordance computation can't support single node so it's disabled for this test
+      System.setProperty("sys.ai.h2o.debug.skipScoring", Boolean.TRUE.toString());
```
michalkurka, line 281: This needs to be in try-finally block or it will not be unset when assertion fails

michalkurka (changes requested):
overall looks good, please resolve inline comments

h2o-algos/src/test/java/hex/coxph/CoxPHTest.java
satai: It's inside try-finally if I look correctly.

michalkurka: my fault! sorry, you are right!

michalkurka (approved):
Thank you @satai

maurever (approved):
LGTM, thanks @satai.

satai:
Was merged as https://github.com/h2oai/h2o-3/pull/5323

PR #4913, opened 2020-09-14
Implement a parallel export to CSV

michalkurka:

PR #5243, opened 2021-01-19
PUBDEV-7953 - Escape quotes by default when writing CSV

Pscheidl:
https://h2oai.atlassian.net/browse/PUBDEV-7953

By default (and the clients use the default) in `CSVStreamParams`, the `_escapeQuotes` is set to false. Which leaves quotes inside string and enum tokens exposed, resulting in invalid CSV written as an output. The correct behavior by default is to escape all quotes.

As [RFC-4180](https://tools.ietf.org/html/rfc4180) says: `Fields containing line breaks (CRLF), double quotes, and commas should be enclosed in double-quotes. This means every line with a quote has to be enclosed in quotes as well. This is done. Some writers enclose all tokens in quotes, some selectively enclose only tokens with the above-mentioned characters in quotes. For simplicity and easiness of this change, we do enclose all enums and strings in quotes if quoting is enabled. No detection of quote presence is done, resulting in a slightly larger CSV.

This change also means that even tokens with no special characters inside will be enclosed in quotes - which is perfectly valid.

### DIFF
--- a/h2o-core/src/main/java/water/fvec/Frame.java
+++ b/h2o-core/src/main/java/water/fvec/Frame.java
@@ -1572,7 +1572,7 @@ public class Frame extends Lockable<Frame> {
     boolean _headers = true;
     boolean _quoteColumnNames = true; 
     boolean _hexString = false;
-    boolean _escapeQuotes = false;
+    boolean _escapeQuotes = true;
     char _separator = DEFAULT_SEPARATOR;
     char _escapeCharacter = DEFAULT_ESCAPE;
 
@@ -1727,7 +1727,6 @@ public class Frame extends Lockable<Frame> {
 
     byte[] getBytesForRow() {
       StringBuilder sb = new StringBuilder();
-      BufferedString tmpStr = new BufferedString();
       for (int i = 0; i < _curChks.length; i++) {
         Vec v = _curChks[i]._vec;
         if (i > 0) sb.append(_parms._separator);
@@ -1738,7 +1737,8 @@ public class Frame extends Lockable<Frame> {
           } else if (v.isUUID()) sb.append(PrettyPrint.UUID(_curChks[i].at16l(_chkRow), _curChks[i].at16h(_chkRow)));
           else if (v.isInt()) sb.append(_curChks[i].at8(_chkRow));
           else if (v.isString()) {
-            final String escapedString = escapeQuotesForCsv(_curChks[i].atStr(tmpStr, _chkRow).toString());
+            final BufferedString unescapedTempStr = new BufferedString();
+            final String escapedString = escapeQuotesForCsv(_curChks[i].atStr(unescapedTempStr, _chkRow).toString());
             sb.append('"').append(escapedString).append('"');
           } else {
             double d = _curChks[i].atd(_chkRow);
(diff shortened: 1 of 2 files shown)


h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1787,7 +1787,7 @@ public int available() {
       Chunk anyChunk = _curChks[0];
 
       // Case 3:  Out of data.
-      if (anyChunk._start + _chkRow == anyChunk._vec.length()) {
+      if (anyChunk._start + _chkRow >= anyChunk._vec.length()) {
```
Pscheidl, line 1790: A bug found. The reader enclosing InputStream, calling the `available` method might actually call the `available` method multiple times. This is a completely valid scenario. For the first time it is called, this stream returns `-1` correctly. Yet the chunk row index increments as well with the call and the state of the stream did not pass the `==` check. Therefore changed to `>=`. 

Naturally tested by the newly introduced test by Apache Common's IOUtils. Taking this change away will crash the test :wink:

h2o-core/src/main/java/water/fvec/Frame.java
```diff
@@ -1738,7 +1737,8 @@ public int getCurrentRowSize() {
           } else if (v.isUUID()) sb.append(PrettyPrint.UUID(_curChks[i].at16l(_chkRow), _curChks[i].at16h(_chkRow)));
           else if (v.isInt()) sb.append(_curChks[i].at8(_chkRow));
           else if (v.isString()) {
-            final String escapedString = escapeQuotesForCsv(_curChks[i].atStr(tmpStr, _chkRow).toString());
+            final BufferedString unescapedTempStr = new BufferedString();
```
honzasterba, line 1740: any reason to move it here? I think the reason this was declared outside was to avoid repeated allocation

Pscheidl: This is a common premature optimization mistake many Java devs believe for whatever reason. Javac is clever enough to optimize this. So I followed the rule of narrowest variable scope possible. I can move it outside if this is a problem for whatever reason. 

https://stackoverflow.com/questions/8803674/declaring-variables-inside-or-outside-of-a-loop

honzasterba (approved):

h2o-core/src/main/java/water/fvec/Frame.java
michalkurka: @Pscheidl this is not true. There is a difference between **declaring** and creating a new instance.

Please keep in mind that I am not arguing whether your change is good/makes an impact or not.

In your modified code the instance will be created per-column (chunk). The BufferedString was designed to be mutable and save you from new object creation. The `atStr` method just updates the pointers on the existing object, no new memory is allocated. This allows for faster iteration over String columns. Obviously the benefits go away as long as you call `toString`.

BTW method getBytesForRow is a huge performance bottleneck, it is critical for good export performance and I have a pending improvement: https://github.com/h2oai/h2o-3/pull/4913/files#diff-ee1c3497b4c32b3ec0b3577c3d1e736b28bac51e97841cb15778610a9255072cR1712

Pscheidl: First, I confused `BufferedString` and `StringBuffer`. Didn't think it was possible to reuse it and didn't look properly in the code. Yes, after looking into `atStr_impl`, I can see the bytes copied anyway. Didn't notice the contract. I was convince we initialize it every time anyway.

Reverted in https://github.com/h2oai/h2o-3/pull/5243/commits/1cb0c48822a4fe7e2e71126fd45f92591f5a798b.

PR #5244, opened 2021-01-20
[PUBDEV-7859] grid search auto recovery

honzasterba:


h2o-core/src/main/java/hex/faulttolerance/Recovery.java
```diff
@@ -13,8 +19,30 @@
...
 
     public static final String REFERENCES_META_FILE_SUFFIX = "_references";
+    public static final String RECOVERY_META_FILE = "recovery.json";
+    
+    public static final String INFO_CLASS = "class";
+    public static final String INFO_RESULT_KEY = "resultKey";
+    public static final String INFO_JOB_KEY = "jobKey";
+
+    public static void autoRecover(String autoRecoveryDir) {
+        if (autoRecoveryDir == null || autoRecoveryDir.length() == 0) {
```
Pscheidl, line 33: `Optional` and `Path`s or `File`s are much more type safe. The problem with Path and File throughout the code is serialization, let's skip that. But using Optional to indicate an argument is Optional seems like a good idea to me.

h2o-core/src/main/java/hex/faulttolerance/Recovery.java
```diff
@@ -13,8 +19,30 @@
 import java.util.*;
 
 public class Recovery<T extends Keyed> {
```
Pscheidl, line 21: This is a main class of its domain, a little JavaDoc with explanation of what it does, how it should be used and why won't hurt.

h2o-core/src/main/java/hex/faulttolerance/Recovery.java
```diff
@@ -13,8 +19,30 @@
...
+    public static final String INFO_CLASS = "class";
+    public static final String INFO_RESULT_KEY = "resultKey";
+    public static final String INFO_JOB_KEY = "jobKey";
+
+    public static void autoRecover(String autoRecoveryDir) {
+        if (autoRecoveryDir == null || autoRecoveryDir.length() == 0) {
+            LOG.debug("Auto recovery dir not configured.");
+        } else {
+            LOG.info("Initializing auto recovery from " + autoRecoveryDir);
+            H2O.submitTask(new H2O.H2OCountedCompleter(H2O.MIN_PRIORITY) {
```
Pscheidl, line 37: This solved the FJ issue, right ?

h2o-core/src/main/java/hex/faulttolerance/Recovery.java
```diff
@@ -13,8 +19,30 @@
...
+    private static final Logger LOG = Logger.getLogger(Recovery.class);
 
     public static final String REFERENCES_META_FILE_SUFFIX = "_references";
+    public static final String RECOVERY_META_FILE = "recovery.json";
+    
+    public static final String INFO_CLASS = "class";
+    public static final String INFO_RESULT_KEY = "resultKey";
+    public static final String INFO_JOB_KEY = "jobKey";
+
+    public static void autoRecover(String autoRecoveryDir) {
```
Pscheidl, line 32: A perfect indicator for a `Recovery` interface and `FileSystemAutoRecovery` implementation - others might come. Strategy pattern. Or could be refactored in future once a new strategy is added. Just noticed this obvious clue.

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfig.java
```diff
@@ -35,6 +37,9 @@ public void notifyAboutEmbeddedWebServerIpPort(InetAddress ip, int port) {
     @Override
     public void notifyAboutCloudSize(InetAddress ip, int port, InetAddress leaderIp, int leaderPort, int size) {
         Log.info(String.format("Created cluster of size %d, leader node IP is '%s'", size, leaderIp.toString()));
+        if (size == cloudSize) {
+            cloudingFinished();
```
Pscheidl, line 41: This change has complex consequences. A normal deployment via "headless service" clustering (not assisted clustering with operator) looks like this: https://github.com/h2oai/h2o-3/tree/master/h2o-k8s - this is the YAML recipe given to our users. There is no persistent volume bound by default when the clustering is done via h2o-k8s module. Which means, **once the pod is down, all data inside the pod are lost**. Helm chart uses the same recipe by default at the moment. 

We have multiple choices:
1. Add PersistentVolume to the documentation + the Helm chart and tell ours users to use this version of YAML. And make this enabled by default. This means, if the user accidentally doesn't define a PersistentVolume for the recovery folder and autorecovery is enabled by default, it won't work, but the client will signal it does.
2. Also add PersistentVolume + the Helm chart and disable this by default. The new YAML will use a dedicated ENV VAR to enable this by default. This will prevent false expectations and failures when newer version of H2O jar is used in combination with older YAML recipe/Helm chart.
3. Leave this disabled by default and support this only via the assisted clustering. Not my preferred option.

I think the ideal solution is as follows:
Add env var to enable/disable autorecovery. Would be present in both `h2o-k8s` and `h2o-clustering` module. In both modules, it will have to be explicitly disabled. In assisted clustering, it makes sense, as there is "expert software" that knows how to operate H2O and has no problem with distinguishing different use cases. In case of h2o-k8s module, this is to counter false expectations when combined with older YAML/Helm.

Pscheidl (commented):
Haven't finished the review just yet, went through the Java part of the code. The code architecture/style related comments are just observations and should be treated as such.

We should focus on the autorecovery being enabled/disabled by default and where. I can help/explain in greater detail if required.

h2o-core/src/main/java/hex/faulttolerance/Recovery.java
honzasterba: yes

h2o-k8s/src/main/java/water/k8s/KubernetesEmbeddedConfig.java
honzasterba: auto-recovery is enabled only when -auto_recovery_dir is set, so I dont think this is a problem

h2o-core/src/main/java/hex/faulttolerance/Recovery.java
honzasterba: done

h2o-core/src/main/java/hex/faulttolerance/Recovery.java
honzasterba: good idea, done

h2o-core/src/main/java/hex/faulttolerance/Recovery.java
honzasterba: sounds like premature refactoring

michalkurka (approved):
Thank you @honzasterba!

PR #5245, opened 2021-01-20
WIP run sorting localy if needed

satai:


michalkurka (changes requested):
Doesn't actually guarantee a local execution because sort is using direct RPC calls, runLocal only work for MRTasks

Alternative solution proposed on Slack

PR #5249, opened 2021-01-21
PUBDEV-8097: H2O won't start with system properties that begin with 'ai.h2o.'

Pscheidl:
There is also an alternative solution based on relocating to a different package, where the problem is described in detail: https://github.com/h2oai/h2o-3/pull/5248

- No need for our users to migrate
- No need to relocate elsewhere
- Cleanest for our users

http://mr-0xg1:8080/view/H2O-3/job/h2o-3-hadoop-smoke-pipeline/job/pavel%252Fhadoop-3-ignore-props/

### DIFF
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -1,7 +1,6 @@
 package water;
 
 import hex.ModelBuilder;
-import hex.faulttolerance.Recovery;
 import jsr166y.CountedCompleter;
 import jsr166y.ForkJoinPool;
 import jsr166y.ForkJoinWorkerThread;
@@ -2173,7 +2172,19 @@ final public class H2O {
       return true;
     }
     return false;
-  } 
+  }
+
+  /**
+   * Any system property starting with `ai.h2o.` and containing any more `.` does not match
+   * this pattern and is therefore ignored. This is mostly to prevent system properties
+   * serving as configuration for H2O's dependencies (e.g. `ai.h2o.org.eclipse.jetty.LEVEL` ).
+   */
+  static boolean isArgProperty(String name) {
+    final String prefix = "ai.h2o.";
+    if (!name.startsWith(prefix))
+      return false;
+    return name.lastIndexOf('.') < prefix.length(); 
+  }
 
   // --------------------------------------------------------------------------
   public static void main( String[] args ) {
@@ -2195,8 +2206,8 @@ final public class H2O {
     // effectively overwriting the earlier args.
     ArrayList<String> args2 = new ArrayList<>(Arrays.asList(args));
     for( Object p : System.getProperties().keySet() ) {
-      String s = (String)p;
-      if( s.startsWith("ai.h2o.") ) {
+      String s = (String) p;
+      if(isArgProperty(s)) {
         args2.add("-" + s.substring(7));
         // hack: Junits expect properties, throw out dummy prop for ga_opt_out
         if (!s.substring(7).equals("ga_opt_out") && !System.getProperty(s).isEmpty())
--- a/h2o-core/src/test/java/water/H2OTest.java
+++ b/h2o-core/src/test/java/water/H2OTest.java
@@ -33,5 +33,19 @@ public class H2OTest {
     assertTrue(id.startsWith("test-desc_test-type"));
     assertTrue(id.endsWith("_" + seq));
   }
+
+  @Test
+  public void testIsArgProperty() {
+      // Any string with a dot after `ai.h2o.` must fail
+      assertFalse(H2O.isArgProperty("noprefixnodots"));
+      assertFalse(H2O.isArgProperty("no.prefix.ever"));
+      assertFalse(H2O.isArgProperty("ai.h2o.anythingwithdotwillfail."));
+      assertFalse(H2O.isArgProperty("ai.h2o.org.eclipse.jetty.LEVEL"));
+      assertFalse(H2O.isArgProperty("ai.h2o.org.eclipse.jetty.util.log.class"));
+      assertFalse(H2O.isArgProperty("ai.h2o.org.eclipse.jetty.util.log.StdErrLog"));
+      assertFalse(H2O.isArgProperty("sys.ai.h2o.nodots"));
+
+      assertTrue(H2O.isArgProperty("ai.h2o.hdfs_config"));
+  }
   
 }


honzasterba:
I don't like this approach. Can lead to more hard-to-catch errors in the future.

h2o-core/src/main/java/water/H2O.java
```diff
@@ -2171,7 +2183,7 @@ public static void main( String[] args ) {
     ArrayList<String> args2 = new ArrayList<>(Arrays.asList(args));
     for( Object p : System.getProperties().keySet() ) {
       String s = (String)p;
-      if( s.startsWith("ai.h2o.") ) {
+      if( s.startsWith("ai.h2o.") && !IGNORED_PROPERTIES.contains(s)) {
```
michalkurka, line 2186: Suggestion: The h2o arguments never have a dot `.` in the name, can we ignore all properties that start with `ai.h2o.` but do have a dot character after this initial prefix?

That would solve this problem for all other cases as well. Not just Jetty.

scripts/jenkins/groovy/hadoopCommands.groovy
```diff
@@ -54,7 +54,7 @@ private GString getCommandHadoop(final stageConfig) {
 private GString getCommandStandalone(final stageConfig) {
     def defaultPort = 54321
     return """
-            java -cp build/h2o.jar:\$(cat /opt/hive-jdbc-cp) water.H2OApp \\
+            java -cp h2o-hadoop-*/h2o-${stageConfig.customData.distribution}${stageConfig.customData.version}-assembly/build/libs/h2odriver.jar:\$(cat /opt/hive-jdbc-cp) water.H2OApp \\
```
michalkurka, line 57: The intention here is to use the regular h2o.jar to connect to Hadoop. We are testing standalone h2o.jar here.

honzasterba: I thought it is recommended to use h2odriver.jar even when not running on hadoop but interacting with HDFS so that matching hadoop libs are used

michalkurka: The original purpose of the test was that the built-in Hadoop classes can be used in most cases (!) to connect to Hadoop as well.

For more complex cases it is absolutely recommended what you are saying. But here we want to test the h2o.jar.

honzasterba: than maybe we should add another stage to test that h2odriver.jar works in standalone as well, wdyt?

michalkurka:
I like this approach better. We already relocated jetty once. Anyone who has adopted to the relocation and changed their logging properties will have to adapt again. I don't think that is reasonable given that there is a simpler solution (this PR) that would solve it not just for Jetty but generally for any such case.

scripts/jenkins/groovy/hadoopCommands.groovy
michalkurka: That makes sense, that case is not covered currently.

Pscheidl: Will do.

h2o-core/src/test/java/water/H2OTest.java
```diff
@@ -33,5 +35,26 @@ private void checkObjectId(String id, int seq) {
     assertTrue(id.startsWith("test-desc_test-type"));
     assertTrue(id.endsWith("_" + seq));
   }
+
+  @Test
+  public void testPropertiesMatching() throws Exception {
+    final Field ignored_properties = H2O.class.getDeclaredField("IGNORED_PROPERTIES");
```
honzasterba, line 41: you could have just made it package-private and not have to use reflection

Pscheidl: Why not ? Done !

h2o-core/src/main/java/water/H2O.java
Pscheidl: Added. It is assumend all arguments have `ai.h2o.` prefix.

scripts/jenkins/groovy/hadoopCommands.groovy
Pscheidl: @honzasterba did this. Done. Thank you.

satai (approved):

h2o-core/src/main/java/water/H2O.java
```diff
@@ -2171,7 +2180,7 @@ public static void main( String[] args ) {
     ArrayList<String> args2 = new ArrayList<>(Arrays.asList(args));
     for( Object p : System.getProperties().keySet() ) {
       String s = (String)p;
-      if( s.startsWith("ai.h2o.") ) {
+      if( s.startsWith("ai.h2o.") && IGNORED_PROPERTIES.matcher(s).matches()) {
```
michalkurka, line 2183: I am having a hard time understanding this. If it starts with "ai.h2o." and matches "ignored properties" pattern? Is that right?

honzasterba: seems it shoud be `! IGNORED_PROPERTIES.matcher(s).matches()`

michalkurka: I think the logic works but should be probably called something else than IGNORED_PROPERTIES, maybe "ARGUMENT_PROPERTIES_PATTERN"

see
    assertTrue(H2O.IGNORED_PROPERTIES.matcher("ai.h2o.hdfs_config").matches());

h2o-core/src/test/java/water/H2OTest.java
```diff
@@ -33,5 +35,18 @@ private void checkObjectId(String id, int seq) {
...
+  public void testPropertiesMatching() throws Exception {
+      // Any string with a dot after `ai.h2o.` must fail
+      assertFalse(H2O.IGNORED_PROPERTIES.matcher("noprefixnodots").matches());
+      assertFalse(H2O.IGNORED_PROPERTIES.matcher("no.prefix.ever").matches());
+      assertFalse(H2O.IGNORED_PROPERTIES.matcher("ai.h2o.anythingwithdotwillfail.").matches());
+      assertFalse(H2O.IGNORED_PROPERTIES.matcher("ai.ai.h2o.org.eclipse.jetty.LEVEL").matches());
+      assertFalse(H2O.IGNORED_PROPERTIES.matcher("ai.h2o.org.eclipse.jetty.util.log.class").matches());
+      assertFalse(H2O.IGNORED_PROPERTIES.matcher("ai.h2o.org.eclipse.jetty.util.log.StdErrLog").matches());
+      
+      assertTrue(H2O.IGNORED_PROPERTIES.matcher("ai.h2o.hdfs_config").matches());
```
michalkurka, line 49: it also matches

"blabla.ai.h2o.something" and "ai.h2o."

Do we want to allow the prefix there?

h2o-core/src/main/java/water/H2O.java
```diff
@@ -2173,7 +2172,19 @@ public static boolean checkUnsupportedJava() {
       return true;
     }
     return false;
-  } 
+  }
+
+  /**
+   * Any system property starting with `ai.h2o.` and containing any more `.` does not match
+   * this pattern and is therefore ignored. This is mostly to prevent system properties
+   * serving as configuration for H2O's dependencies (e.g. `ai.ai.h2o.org.eclipse.jetty.LEVEL` ).
```
mn-mikke, line 2180: `ai.ai.h2o.org.eclipse.jetty.LEVEL` -> `ai.h2o.org.eclipse.jetty.LEVEL` or `sys.ai.h2o.org.eclipse.jetty.LEVEL` like in the test?

mn-mikke (approved):
One question otherwise lgtm

h2o-core/src/main/java/water/H2O.java
michalkurka: good catch! thank you

It should be `ai.h2o.org.eclipse.jetty.LEVEL`

michalkurka:
@mn-mikke thank you for catching the typos

PR #5251, opened 2021-01-21
PUBDEV-7964: Update graphic to include Kubernetes

hannah-tillman:
for: https://h2oai.atlassian.net/browse/PUBDEV-7964


NidhiMehta (approved):

tomkraljevic:
So, if I re-did this diagram today I would take some stuff out.

"Customer Algorithm" on the top box doesn't really make sense.
JavaScript doesn't really make sense.
Excel/Tableau, it turns out, isn't something people actually do.

R/Python/Flow *is* what people do, so keep those.

There are many more algorithms in the GLM/GBM/RF/etc. box.  So the box should reflect that (add '...').

"Customer Algorithm" -> "Customer Algorithm (in Java)"

hannah-tillman:
@tomkraljevic Gotcha--added those updates :)

tomkraljevic (approved):

PR #5255, opened 2021-01-25
PUBDEV-7951: Delta lake fails to import

michalkurka:


h2o-core/src/main/java/water/persist/PersistManager.java
```diff
@@ -772,7 +792,7 @@ public Persist getPersistForURI(URI uri) {
    * @param matchStr The regular expression to use on the string after prefix
    * @return list containing the matching entries
    */
-  private ArrayList<String> matchPattern(String prefix, ArrayList<String> fileList, String matchStr){
+  public ArrayList<String> matchPattern(String prefix, ArrayList<String> fileList, String matchStr){
```
mn-mikke, line 795: nit: Any reason for making this method public?

h2o-core/src/main/java/water/persist/PersistManager.java
```diff
@@ -409,17 +402,44 @@ public void importFiles(String path, String pattern, ArrayList<String> files, Ar
...
+  }
+
+  private void filterMetadataFiles(String path, ArrayList<String> files, ArrayList<String> keys, ArrayList<String> fails) {
+    filterFiles(new DeltaLakeMetadataFilter(), path, files, keys, fails);
+  }
+
+  static class DeltaLakeMetadataFilter implements Function2<String, ArrayList<String>, ArrayList<String>> {
+    @Override
+    public ArrayList<String> apply(String prefix, ArrayList<String> ids) {
+      if (isDeltaLogPath(prefix))
```
mn-mikke, line 417: Can be a prefix also a relative path?

mn-mikke, line 417: An obscure edge case... What if a user folder is `/a/_delta_log/b` and persists data to this folder with delta. IMHO, reading from this folder would fail on files from `/a/_delta_log/b/_delta_log`.

h2o-core/src/main/java/water/persist/PersistManager.java
```diff
@@ -409,17 +402,44 @@ public void importFiles(String path, String pattern, ArrayList<String> files, Ar
...
+        return ids; // users intention is to import these files
+      ArrayList<String> filteredIds = new ArrayList<>(ids.size());
+      for (String id : ids) {
+        try {
+          URI uri = URI.create(id);
+          String path = uri.getPath();
+          if (isDeltaLogPath(path))
+            continue;
+        } catch (Exception e) {
+          Log.trace("Cannot create uri", e);
```
mn-mikke, line 427: Shouldn't the logging level be warning. I assume that trace is off by default.... Or do we want to hide such a messages.

h2o-core/src/main/java/water/persist/PersistManager.java
michalkurka: good point, I wanted to do this to avoid polluting the logs but I will improve it

h2o-core/src/main/java/water/persist/PersistManager.java
michalkurka: good point, will improve

h2o-core/src/main/java/water/persist/PersistManager.java
michalkurka: you are right, I think the use case of trying to read the metadata files directly doesn't make sense and is creating weird edge cases

I will fix it and ignore the files always

mn-mikke (approved):
LGTM

PR #5256, opened 2021-01-26
PUBDEV-7958: Enable returning data from explain's varimp_heatmap and model_correlation_matrix

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7958

This PR enables retrieving data from varimp_heatmap and model_correlation_matrix which are the only two functions that do some computations in the explain module - the rest of the functions/methods use data straight from h2o with only transformations used for plotting. This is also motivated by feedback I got from @meganjkurka (that's why I added you as reviewer, in case I missed something).

## Usage:
Python:
```python
# return pd.DataFrame by default, if not available ImportError will be thrown (only when user uses this function; explain related functions don't use pandas)
df = aml.varimp()
df = aml.model_correlation(frame=test)

# return tuple with numpy.ndarray, column names and row names when pandas is explicitly not used 
varimps, model_ids, variables = aml.varimp(use_pandas=False)
# column names and row names are the same `model_ids` in the same order
corr, model_ids = aml.model_correlation(frame=test_frame, use_pandas=False)
```
R:
```r
varimp = h2o.varimp(object = aml) 
corr = h2o.model_correlation(object = aml, newdata = test)
```

### DIFF
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -1040,7 +1040,7 @@ def pd_plot(
 
 
 def pd_multi_plot(
-        models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, List[h2o.model.model_base]]
+        models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, h2o.H2OFrame, List[h2o.model.model_base]]
         frame,  # type: h2o.H2OFrame
         column,  # type: str
         best_of_family=True,  # type: bool
@@ -1508,8 +1508,8 @@ def _calculate_clustering_indices(matrix):
 
 
 def varimp_heatmap(
-        models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, List[h2o.model.ModelBase]]
-        top_n=20,  # type: int
+        models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, h2o.H2OFrame, List[h2o.model.ModelBase]]
+        top_n=None,  # type: Option[int]
         figsize=(16, 9),  # type: Tuple[float]
         cluster=True,  # type: bool
         colormap="RdYlBu_r"  # type: str
@@ -1526,8 +1526,8 @@ def varimp_heatmap(
     encoded features and return a single variable importance for the original categorical
     feature. By default, the models and variables are ordered by their similarity.
 
-    :param models: H2O AutoML object or list of H2O models
-    :param top_n: use just top n models (applies only when used with H2OAutoML)
+    :param models: H2O AutoML object, leaderboard slice or list of H2O models
+    :param top_n: DEPRECATED. use just top n models (applies only when used with H2OAutoML)
     :param figsize: figsize: figure size; passed directly to matplotlib
     :param cluster: if True, cluster the models and variables
     :param colormap: colormap to use
@@ -1557,7 +1557,15 @@ def varimp_heatmap(
     >>> aml.varimp_heatmap()
     """
     plt = get_matplotlib_pyplot(False, raise_if_not_available=True)
-    varimps, model_ids,  x = varimp(models=models, top_n=top_n, cluster=cluster, use_pandas=False)
+    if isinstance(models, h2o.automl._base.H2OAutoMLBaseMixin):
+        if top_n is not None:
+            import warnings
+            from h2o.exceptions import H2ODeprecationWarning
+            warnings.warn("Setting the `top_n` parameter is deprecated, use leaderboard slice instead, e.g., aml.leaderboard.head({}).".format(top_n), category=H2ODeprecationWarning)
+            models = models.leaderboard.head(top_n)
+        else:
+            models = models.leaderboard.head(20)
+    varimps, model_ids,  x = varimp(models=models, cluster=cluster, use_pandas=False)
 
     plt.figure(figsize=figsize)
     plt.imshow(varimps, cmap=plt.get_cmap(colormap))
@@ -1574,8 +1582,7 @@ def varimp_heatmap(
 
 
 def varimp(
-        models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, List[h2o.model.ModelBase]]
-        top_n=20,  # type: int
+        models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, h2o.H2OFrame, List[h2o.model.ModelBase]]
         cluster=True,  # type: bool
         use_pandas=True  # type: bool
 ):
@@ -1583,23 +1590,16 @@ def varimp(
     """
         Get data that are used to build varimp_heatmap plot.
 
-        :param models: H2O AutoML object or list of H2O models
-        :param top_n: use just top n models (applies only when used with H2OAutoML)
+        :param models: H2O AutoML object, leaderboard slice or list of H2O models
         :param cluster: if True, cluster the models and variables
         :param use_pandas: if True, try to return pandas DataFrame. Otherwise return a triple (varimps, model_ids, variable_names)
         :returns: either pandas DataFrame (if use_pandas == True) or a triple (varimps, model_ids, variable_names)
     """
     if _is_automl_or_leaderboard(models):
         model_ids = _get_model_ids_from_automl_or_leaderboard(models, filter_=_has_varimp)
-        models = [
-            h2o.get_model(model_id)
-            for model_id in model_ids[:min(top_n, len(model_ids))]
-        ]
-    else:
-        top_n = len(models)
+        models = [h2o.get_model(model_id) for model_id in model_ids]
     # Filter out models that don't have varimp
     models = [model for model in models if _has_varimp(model)]
-    models = models[:min(len(models), top_n)]
     if len(models) == 0:
         raise RuntimeError("No model with variable importance")
     varimps = [_consolidate_varimps(model) for model in models]
@@ -1625,9 +1625,9 @@ def varimp(
 
 
 def model_correlation_heatmap(
-        models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, List[h2o.model.ModelBase]]
+        models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, h2o.H2OFrame, List[h2o.model.ModelBase]]
         frame,  # type: h2o.H2OFrame
-        top_n=20,  # type: int
+        top_n=None,  # type: Optional[int]
         cluster_models=True,  # type: bool
         triangular=True,  # type: bool
         figsize=(13, 13),  # type: Tuple[float]
@@ -1641,9 +1641,9 @@ def model_correlation_heatmap(
     For classification, frequency of identical predictions is used. By default, models
     are ordered by their similarity (as computed by hierarchical clustering).
 
-    :param models: H2OAutoML object or a list of H2O models
+    :param models: H2OAutoML object, leaderboard slice or a list of H2O models
     :param frame: H2OFrame
-    :param top_n: show just top n models (applies only when used with H2OAutoML)
+    :param top_n: DEPRECATED. show just top n models (applies only when used with H2OAutoML).
     :param cluster_models: if True, cluster the models
     :param triangular: make the heatmap triangular
     :param figsize: figsize: figure size; passed directly to matplotlib
@@ -1674,7 +1674,15 @@ def model_correlation_heatmap(
     >>> aml.model_correlation_heatmap(test)
     """
     plt = get_matplotlib_pyplot(False, raise_if_not_available=True)
-    corr, model_ids = model_correlation(models, frame, top_n, cluster_models, use_pandas=False)
+    if isinstance(models, h2o.automl._base.H2OAutoMLBaseMixin):
+        if top_n is not None:
+            import warnings
+            from h2o.exceptions import H2ODeprecationWarning
+            warnings.warn("Setting the `top_n` parameter is deprecated, use leaderboard slice instead, e.g., aml.leaderboard.head({}).".format(top_n), category=H2ODeprecationWarning)
+            models = models.leaderboard.head(top_n)
+        else:
+            models = models.leaderboard.head(20)
+    corr, model_ids = model_correlation(models, frame, cluster_models, use_pandas=False)
 
     if triangular:
         corr = np.where(np.triu(np.ones_like(corr), k=1).astype(bool), float("nan"), corr)
@@ -1699,9 +1707,8 @@ def model_correlation_heatmap(
 
 
 def model_correlation(
-        models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, List[h2o.model.ModelBase]]
+        models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, h2o.H2OFrame, List[h2o.model.ModelBase]]
         frame,  # type: h2o.H2OFrame
-        top_n=20,  # type: int
         cluster_models=True,  # type: bool
         use_pandas=True  # type: bool
 ):
@@ -1709,23 +1716,16 @@ def model_correlation(
     """
     Get data that are used to build model_correlation_heatmap plot.
 
-    :param models: H2OAutoML object or a list of H2O models
+    :param models: H2OAutoML object, leaderboard slice or a list of H2O models
     :param frame: H2OFrame
-    :param top_n: show just top n models (applies only when used with H2OAutoML)
     :param cluster_models: if True, cluster the models
     :param use_pandas: if True, try to return pandas DataFrame. Otherwise return a tuple (correlation_matrix, model_ids)
     :returns: either pandas DataFrame (if use_pandas == True) or a tuple (correlation_matrix, model_ids)
     """
     if _is_automl_or_leaderboard(models):
         model_ids = _get_model_ids_from_automl_or_leaderboard(models)
-        models = [
-            h2o.get_model(model_id)
-            for model_id in model_ids[:min(top_n, len(model_ids))]
-        ]
-    else:
-        top_n = len(models)
+        models = [h2o.get_model(model_id) for model_id in model_ids]
     is_classification = frame[models[0].actual_params["response_column"]].isfactor()[0]
-    models = models[:min(len(models), top_n)]
     predictions = []
     with no_progress():
         for idx, model in enumerate(models):
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -1308,7 +1308,7 @@ def ice_plot(
 
 
 def _has_varimp(model):
-    # type: (Union[str, h2o.model.ModelBase]) -> bool
+    # type: (h2o.model.ModelBase) -> bool
     """
     Does model have varimp?
     :param model: model or a string containing model_id
@@ -1319,6 +1319,7 @@ def _has_varimp(model):
     # when a model is stopped sooner than calculating varimp (xgboost can rarely have no varimp).
     output = model._model_json["output"]
     return output.get("variable_importances") is not None
+
 
 def _is_automl_or_leaderboard(obj):
     # type: (object) -> bool
@@ -1355,8 +1356,7 @@ def _get_models_from_automl_or_leaderboard(automl_or_leaderboard, filter_=lambda
     :param filter_: a predicate used to filter model_ids. Signature of the filter is (model) -> bool.
     :return: Generator[h2o.model.ModelBase, None, None]
     """
-    models = map(h2o.get_model, _get_model_ids_from_automl_or_leaderboard(automl_or_leaderboard))
-    return (model for model in models if filter_(model))
+    return filter(filter_, map(h2o.get_model, _get_model_ids_from_automl_or_leaderboard(automl_or_leaderboard)))
 
 
 def _get_xy(model):
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -1349,14 +1349,15 @@ def _get_model_ids_from_automl_or_leaderboard(automl_or_leaderboard, filter_=lam
 
 
 def _get_models_from_automl_or_leaderboard(automl_or_leaderboard, filter_=lambda _: True):
-    # type: (object) -> List[str]
+    # type: (object) -> Generator[h2o.model.ModelBase, None, None]
     """
     Get model ids from H2OAutoML object or leaderboard
     :param automl_or_leaderboard: AutoML
     :param filter_: a predicate used to filter model_ids. Signature of the filter is (model) -> bool.
     :return: Generator[h2o.model.ModelBase, None, None]
     """
-    return filter(filter_, map(h2o.get_model, _get_model_ids_from_automl_or_leaderboard(automl_or_leaderboard)))
+    models = (h2o.get_model(model_id) for model_id in _get_model_ids_from_automl_or_leaderboard(automl_or_leaderboard))
+    return (model for model in models if filter_(model))
 
 
 def _get_xy(model):
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -1058,7 +1058,7 @@ def pd_multi_plot(
     on the response. The effect of a variable is measured in change in the mean response.
     PDP assumes independence between the feature for which is the PDP computed and the rest.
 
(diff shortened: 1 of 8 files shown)


ledell (changes requested):
From our conversation, here's the updated, proposed API (should also work with lists of models):

Python:
```
# return pd.DataFrame by default, if not available ImportError will be thrown (only when user uses this function; explain related functions don't use pandas)
df = aml.varimp()
df = aml.model_correlation(frame=test)

# return tuple with numpy.ndarray, column names and row names when pandas is explicitly not used 
varimps, model_ids, variables = aml.varimp(use_pandas=False)
# column names and row names are the same `model_ids` in the same order
corr, model_ids = aml.model_correlation(frame=test_frame, use_pandas=False)
```

R:
```
varimp = h2o.varimp(object = aml) #are we planning to overload the h2o.varimp function?
corr = h2o.model_correlation(object = aml, newdata = test)
```

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1511,6 +1519,38 @@ def varimp_heatmap(
...
+        top_n=20,  # type: int
+        cluster=True,  # type: bool
+        use_pandas=True  # type: bool
+):
+    # type: (...) -> Union[pandas.DataFrame, Tuple[numpy.ndarray, List[str], List[str]]]
+    """
+        Get data that are used to build varimp_heatmap plot.
+
+        :param models: H2O AutoML object or list of H2O models
+        :param top_n: use just top n models (applies only when used with H2OAutoML)
```
sebhrusen, line 1549: are we already using this param with this semantic?
I don't like that:
- it has a default value, although it applies only to AutoML: I would either assign a (non-None) default value and ensure it always applies, or set it to None and assign the default dynamically if AutoML.
- it's in 2nd position although the param doesn't apply when using lists: we should at least ensure that the first 3 params are always used, given that we can't enforce keywords params with the `*` marker (Py 3 only).
- it works only with the default ordering of the AutoML leaderboard.

I think we would have been better advised to provide a `get_top_n_models` method on AutoML.
For example:
```python
# in H2OAutoMLBase
def get_top_models(self, top_n=20, order_by=None):
    pass

# in def varimp(…):
    if isinstance(models, h2o.automl._base.H2OAutoMLBaseMixin):
        models = models.get_top_models()
```
No need for `top_n` param here anymore, and if user wants to customize the models extraction from AutoML, it's easy to do:
```python
varimp(aml.get_top_models(30, order_by='logloss'), cluster=False)
```
without mentioning that it removes duplicate code from all those places where we use `top_n` in this file.

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1608,6 +1637,48 @@ def model_correlation_heatmap(
...
+        cluster_models=True,  # type: bool
+        use_pandas=True  # type: bool
+):
+    # type: (...) -> Union[pandas.DataFrame, Tuple[numpy.ndarray, List[str]]]
+    """
+    Get data that are used to build model_correlation_heatmap plot.
+
+    :param models: H2OAutoML object or a list of H2O models
+    :param frame: H2OFrame
+    :param top_n: show just top n models (applies only when used with H2OAutoML)
```
sebhrusen, line 1677: see above

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1617,56 +1688,40 @@ def model_correlation_heatmap(
...
-                    corr[i, j] = (predictions[i, :] == predictions[j, :]).mean()
+                    corr[i, j] = (predictions[i] == predictions[j]).mean()[0]
                     corr[j, i] = corr[i, j]
     else:
-        corr = np.corrcoef(predictions)
-
+        try:
+            from io import StringIO
+        except ImportError:
+            from StringIO import StringIO
```
sebhrusen, line 1709: this is standard lib, the import could go on the top of the file: looks a bit noisy here.

sebhrusen: btw the order is reverted: Py 2.7 has both `io.StringIO` (for unicode only) and `StringIO.StringIO` (8 bits strings or unicode).
So you for max compatibility you need to use:
```python
try:
    from StringIO import StringIO  # py2 (first as py2 also has io.StringIO, but only with unicode support)
except:
    from io import StringIO  # py3
```
Also, when doing compatibility imports, please write a short comment for each import telling when each is expected to work.

h2o-r/h2o-package/R/classes.R
```diff
@@ -978,3 +978,75 @@ setMethod("summary", signature("H2OAutoML"), function(object) {
...
+#' f <- "https://s3.amazonaws.com/h2o-public-test-data/smalldata/prostate/prostate_complete.csv.zip"
+#' pros <- h2o.importFile(f)
+#' response <- "GLEASON"
+#' predictors <- c("ID", "AGE", "CAPSULE", "DCAPS", "PSA", "VOL", "DPROS")
+#' aml <- h2o.automl(x = predictors, y = response, training_frame = pros, max_runtime_secs = 60)
+#' h2o.varimp(aml)
+#' }
+#' @export
+setMethod("h2o.varimp", signature("H2OAutoML"), function(object, top_n = 20) {
+  .varimp_matrix(object, top_n = top_n)
```
sebhrusen, line 1051: I guess we could use the same suggestion as for Py here and get rid of the `top_n` param in all those signatures.

sebhrusen (changes requested):
some concerns about what I'll call the nasty `top_n`

h2o-py/h2o/explanation/_explain.py
tomasfryda: Thank you! I didn't know this.

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1302,7 +1304,13 @@ def _has_varimp(model):
     :param model: model or a string containing model_id
     :returns: bool
     """
-    return _get_algorithm(model) not in ["stackedensemble", "naivebayes"]
+    if isinstance(model, h2o.model.ModelBase):
+        # check for cases when variable importance is disabled or
+        # when a model is stopped sooner than calculating varimp (xgboost can rarely have no varimp).
+        output = model._model_json["output"]
+        return "variable_importances" in list(output.keys()) and output["variable_importances"]
```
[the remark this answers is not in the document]
tomasfryda: Thank you for the suggestion!  I'll use `output.get("variable_importances") is not None` as I want a `bool`.

h2o-py/h2o/explanation/_explain.py
tomasfryda: The `top_n` is already used in explain.

The problem with `explain` is that it has different behavior depending on the input -  for automl it will show you mainly the leader model and SHAP from the best tree model, for list of models it assumes that you handpicked them and want to see each of the handpicked models. This would make it hard to use the `get_top_models`. I'm not sure if there is some easy intuitive way of doing this differently.

That being said, I'm not opposed to changing the behavior but it would not be backward compatible and I think it is easier for the users to, e.g., in Jupyter press Shift + Tab and find out about `top_n` than going through the methods(though I think most people would just try to get a subset of a list of models from the leaderboard).

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1512,26 +1551,53 @@ def varimp_heatmap(
...
-        models = [
-            h2o.get_model(model_id)
-            for model_id in model_ids[:min(top_n, len(model_ids))]
-        ]
-    else:
-        top_n = len(models)
+        if top_n is not None:
+            import warnings
+            from h2o.exceptions import H2ODeprecationWarning
+            warnings.warn("Setting the `top_n` parameter is deprecated, use leaderboard slice instead, e.g., aml.leaderboard.head({}).".format(top_n), category=H2ODeprecationWarning)
```
sebhrusen, line 1557:
```suggestion
            warnings.warn("Setting the `top_n` parameter is deprecated, use a leaderboard (sub)frame instead, e.g., aml.leaderboard.head({}).".format(top_n), category=H2ODeprecationWarning)
```

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1609,50 +1668,22 @@ def model_correlation_heatmap(
...
-        corr = corr[order, :]
-        corr = corr[:, order]
-        models = [models[i] for i in order]
+        if top_n is not None:
+            import warnings
+            from h2o.exceptions import H2ODeprecationWarning
+            warnings.warn("Setting the `top_n` parameter is deprecated, use leaderboard slice instead, e.g., aml.leaderboard.head({}).".format(top_n), category=H2ODeprecationWarning)
+            models = models.leaderboard.head(top_n)
+        else:
+            models = models.leaderboard.head(20)
```
sebhrusen, line 1677: same as above? maybe can be extracted to private fn.

h2o-py/h2o/explanation/_explain.py
```diff
@@ -2209,16 +2286,25 @@ def _process_models_input(
...
+    is_aml = False
+    if _is_automl_or_leaderboard(models):
+        is_aml = True
+        models = models.leaderboard if isinstance(models, h2o.automl._base.H2OAutoMLBaseMixin) else models
+        models_to_show = [h2o.get_model(models[0, "model_id"])]
+        multiple_models = models.nrow > 1
+        models_with_varimp = models_to_show if _has_varimp(models_to_show[0]) \
+            else list(islice((h2o.get_model(model)
+                              for model in _get_model_ids_from_automl_or_leaderboard(models, filter_=_has_varimp)
+                              if _has_varimp(h2o.get_model(model))), 1))
```
sebhrusen, line 2299: it's very inefficient to call `h2o.get_model` twice or thrice here (filter=_has_varimp will call get_model), each call makes a request to the backend.
Also, I don't understand why `_has_varimp` appears both in the `filter_` and in the `if` guard.
Finally, usage of `list(islice(..., 1))` is rather unorthodox, you probably just need `next(...)`.
Overall, this piece can be improved for both readability and efficiency.

tomasfryda: The `_get_model_ids_from_automl_or_leaderboard(models, filter_=_has_varimp)` will return model ids of all algos but StackedEnsembles and NaiveBayes (they don't have varimp). 

The ` if _has_varimp(h2o.get_model(model))` is there to check that the model that should have varimp really has it - sometimes xgboost didn't for some reason.

I chose `list(islice(..., 1))` because of the `next([])` behavior - I'd have to surround it with `try-except` block. 

I will split it in to multiple lines, it will save one call to `h2o.get_model` and it should improve readability as well.

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1302,7 +1307,41 @@ def _has_varimp(model):
...
+    # type: (object) -> List[str]
+    """
+    Get model ids from H2OAutoML object or leaderboard
+    :param automl_or_leaderboard: AutoML
+    :param filter_: a predicate used to filter model_ids. Signature of the filter is (model_id: str) -> bool.
+    :return: List[str]
+    """
+    leaderboard = automl_or_leaderboard.leaderboard \
+        if isinstance(automl_or_leaderboard, h2o.automl._base.H2OAutoMLBaseMixin) \
+        else automl_or_leaderboard
```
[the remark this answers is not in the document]
tomasfryda: Thanks, I didn't realize that.

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1302,7 +1307,41 @@ def _has_varimp(model):
...
     :returns: bool
     """
-    return _get_algorithm(model) not in ["stackedensemble", "naivebayes"]
+    if isinstance(model, h2o.model.ModelBase):
+        # check for cases when variable importance is disabled or
+        # when a model is stopped sooner than calculating varimp (xgboost can rarely have no varimp).
+        output = model._model_json["output"]
+        return output.get("variable_importances") is not None
+    else:
+        return _get_algorithm(model) not in ["stackedensemble", "naivebayes"]
```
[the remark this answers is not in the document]
tomasfryda: Cool, I didn't know that!

The problem is that sometimes some models don't have varimp even though the algo supports it, e.g., XGBoost that is stopped sooner than calculating varimp - that's why there is the `output.get("variable_importances") is not None` to check that there really are some results.

The other branch is using `model_id` - if the prefix of the model id is a known algo (`ALGONAME_.*`) use that (transformed to the same name as is the `model.algo`), otherwise `h2o.get_model(model_id).algo`.

The `has_extension(model, 'VariableImportance')` would require me to `get_model`s in order to find if they have varimp and I think it could give me false positives - seems that the mixin is added without looking if the model really has the varimp data.

h2o-py/h2o/explanation/_explain.py
sebhrusen: `next(comprehension, def_val)` doesn't raise exception.
You may need 2 steps, `next` followed with a `if-else` clause but Python has a strong focus on readability, let's use it.

> The if _has_varimp(h2o.get_model(model)) is there to check that the model that should have varimp really has it - sometimes xgboost didn't for some reason.

it still doesn't make much sense to me: you're using the same predicate function in both cases, so why in the first case it could return a "model that should have varimp [but maybe doesn't have it]", whereas the second case ensures that "the model that should have varimp really has it".

tomasfryda: > it still doesn't make much sense to me: you're using the same predicate function in both cases, so why in the first case it could return a "model that should have varimp [but maybe doesn't have it]", whereas the second case ensures that "the model that should have varimp really has it".

The first application  of the `_has_varimp` checks only on the model_id - so it's basically just a coarse filter to get rid of the SE and NaiveBayes. The second application is on the "materialized" model so it can be sure that it does have the varimp. 

I see I should really improve readability and possibly split the has_varimp in to two functions one for model_ids and one for models.

Thank you for mentioning `next(..., default)`, I didn't know that.

h2o-py/h2o/explanation/_explain.py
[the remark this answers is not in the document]
sebhrusen: > The other branch is using model_id - if the prefix of the model id is a known algo (ALGONAME_.*) use that (transformed to the same name as is the model.algo), otherwise h2o.get_model(model_id).algo

We should always avoid using `model_id` to extract information about the model: `model_id` is often entered by the user, and we can't trust the user (copy/paste, change the estimator…).
So yes, you should only rely on the result from `get_model`, and you need to ensure that `get_model` is always called once, i.e never keep ids further that the main entry point.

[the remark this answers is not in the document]
sebhrusen: > that's why there is the output.get("variable_importances") is not None to check that there really are some results

you can keep this additional condition. Was mainly concerned with the algo check:
```return _get_algorithm(model) not in ["stackedensemble", "naivebayes"]```

File: h2o-py/h2o/explanation/_explain.py (hunk 8 above)
Comment (sebhrusen) [reply h2o-py/h2o/explanation/_explain.py:2299]: > I see I should really improve readability and possibly split the has_varimp in to two functions one for model_ids and one for models.

exactly, you can't have this function have a different semantic depending if the model parameter is passed as a model or as a model_id

File: h2o-py/h2o/explanation/_explain.py [hunk 11]
```diff
@@ -1045,7 +1051,7 @@ def pd_multi_plot(
     on the response. The effect of a variable is measured in change in the mean response.
     PDP assumes independence between the feature for which is the PDP computed and the rest.
 
-    :param models: H2O AutoML object, or list of H2O models
+    :param models: H2O AutoML object, slice of leaderboard, or list of H2O models
```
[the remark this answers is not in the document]
tomasfryda: I removed the "slice" from docstrings but kept it in examples in the documentation, is that alright?

https://github.com/h2oai/h2o-3/pull/5256/files#diff-76a0048a26e179ddf1c1293825e7c62b8abe2412a0abdc8ec8cb0f4cca86466fR352-R353

h2o-py/h2o/explanation/_explain.py
```diff
@@ -2225,16 +2309,31 @@ def _process_models_input(
...
+        else:
+            models_to_show = [h2o.get_model(models[0, "model_id"])]
+        if _has_varimp(models_to_show[0]):
+            models_with_varimp = models_to_show
+        else:
+            model_with_varimp = next(_get_models_from_automl_or_leaderboard(models, filter_=_has_varimp), None)
+            if model_with_varimp is None:
+                models_with_varimp = []
+            else:
+                models_with_varimp = [model_with_varimp]
```
sebhrusen, line 2327: nitpick: one assignment better than 2
```suggestion
            models_with_varimp = [] if model_with_varimp is None else [model_with_varimp]
```

h2o-r/h2o-package/R/classes.R
```diff
@@ -986,3 +986,100 @@ setMethod("summary", signature("H2OAutoML"), function(object) {
...
+#' pros <- h2o.importFile(f)
+#' response <- "GLEASON"
+#' predictors <- c("ID", "AGE", "CAPSULE", "DCAPS", "PSA", "VOL", "DPROS")
+#' aml <- h2o.automl(x = predictors, y = response, training_frame = pros, max_runtime_secs = 60)
+#' h2o.varimp(aml@leaderboard[1:5,])
+#' }
+#' @export
+setMethod("h2o.varimp", signature("H2OFrame"), function(object) {
+  if (! "model_id" %in% names(object)){
+    warning("This is not a leaderboard frame.", call. = FALSE)
```
sebhrusen, line 1081: why warning instead of error here?
I'd also make the message more explicit if the user tries to use it with a collection of ids: "h2o.varimp requires a H2OFrame with a 'model_id' column (e.g. AutoML leaderboard)"

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1616,64 +1677,90 @@ def model_correlation_heatmap(
...
         for i in range(len(models)):
             for j in range(len(models)):
                 if i <= j:
-                    corr[i, j] = (predictions[i, :] == predictions[j, :]).mean()
+                    corr[i, j] = (predictions[i] == predictions[j]).mean()[0]
                     corr[j, i] = corr[i, j]
     else:
-        corr = np.corrcoef(predictions)
-
+        corr = np.genfromtxt(StringIO(predictions[0].cbind(predictions[1:]).cor().get_frame_data()),
```
sebhrusen, line 1749: is `predictions != predictions[0].cbind(predictions[1:])` ?

sebhrusen (approved):
Mainly syntactic comments otherwise looks good.

h2o-r/h2o-package/R/explain.R
```diff
@@ -1614,21 +1623,20 @@ h2o.varimp_heatmap <- function(object,
...
 #' For classification, frequency of identical predictions is used. By default, models
 #' are ordered by their similarity (as computed by hierarchical clustering).
 #'
-#' @param object An H2OAutoML object or list of H2O models.
-#' @param newdata An H2O Frame.  Predictions from the models will be generated using this frame, 
+#' @param object A list of H2O models, an H2O AutoML instance, or an H2OFrame with a 'model_id' column (e.g. H2OAutoML leaderboard)..
+#' @param newdata An H2O Frame.  Predictions from the models will be generated using this frame,
 #'                so this should be a holdout set.
-#' @param top_n Integer specifying the number models shown in the heatmap (used only with an 
+#' @param top_n Integer specifying the number models shown in the heatmap (used only with an
```
sebhrusen, line 1636: isn't it deprecated in R? at least in doc.

sebhrusen: does it require to define a generic method like with `varimp`?

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1598,7 +1598,7 @@ def varimp(
         :returns: either pandas DataFrame (if use_pandas == True) or a triple (varimps, model_ids, variable_names)
     """
     if _is_automl_or_leaderboard(models):
-        models = _get_models_from_automl_or_leaderboard(models, filter_=_has_varimp)
+        models = list(_get_models_from_automl_or_leaderboard(models, filter_=_has_varimp))
```
sebhrusen, line 1601: why do you need a list? 
most Py operations accept an iterator: see my other comment about using `filter` (which returns an iterator) instead of generator.
I don't know if all models are always needed here.
If yes, then you can as well fetch them all from the start.
If no, then the iterator will ensure that further operations will fetch models only if needed.

h2o-r/h2o-package/R/classes.R
tomasfryda: So it behaves like the `h2o.varimp(model)` which warns and returns `NULL`. Since it is a generic function I think it should behave the same for different input types and it used to warn so I didn't want to modify that behavior.

h2o-r/h2o-package/R/explain.R
tomasfryda: Good catch, thank you!

h2o-py/h2o/explanation/_explain.py
tomasfryda: Predictions is `List[H2OFrame]`. `predictions[0].cbind(predictions[1:])` is `H2OFrame`.

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1341,16 +1338,27 @@ def _get_model_ids_from_automl_or_leaderboard(automl_or_leaderboard, filter_=lam
...
+def _get_models_from_automl_or_leaderboard(automl_or_leaderboard, filter_=lambda _: True):
+    # type: (object) -> List[str]
+    """
+    Get model ids from H2OAutoML object or leaderboard
+    :param automl_or_leaderboard: AutoML
+    :param filter_: a predicate used to filter model_ids. Signature of the filter is (model) -> bool.
+    :return: Generator[h2o.model.ModelBase, None, None]
+    """
+    models = map(h2o.get_model, _get_model_ids_from_automl_or_leaderboard(automl_or_leaderboard))
+    return (model for model in models if filter_(model))
```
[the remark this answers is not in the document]
tomasfryda: Thank you!

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1309,7 +1314,49 @@ def _has_varimp(model):
     :param model: model or a string containing model_id
     :returns: bool
     """
-    return _get_algorithm(model) not in ["stackedensemble", "naivebayes"]
+    assert isinstance(model, h2o.model.ModelBase)
```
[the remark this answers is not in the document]
tomasfryda: Thank you

h2o-r/h2o-package/R/explain.R
tomasfryda: > does it require to define a generic method like with varimp?

I think it's probably better to deprecate it and motivate users to use the more flexible way - the leaderboard.

h2o-py/h2o/explanation/_explain.py
tomasfryda: This uses all the models with varimp. All this function does is getting varimp from them, transform it so it contains varimp per variable (not the encoded variable) and it makes a matrix / pandas.dataframe if `use_pandas==True`.

h2o-py/h2o/explanation/_explain.py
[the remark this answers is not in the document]
tomasfryda: I just found out that `filter` is eager in py2.7 and `map` as well (they return lists). So I will convert both to generator comprehension so it works lazily for both py2.7 and py3.

https://docs.python.org/2.7/library/functions.html#filter

h2o-py/h2o/explanation/_explain.py
sebhrusen: ok! I made the assumption that `predictions` was already a frame.

h2o-r/h2o-package/R/classes.R
sebhrusen: well, it's not exactly the same scenario:
`h2o.varimp(model)` returns NULL if the model doesn't support variable importance, right?
In the case above, we're talking about the user applying the function to an H2OFrame with wring format.

h2o-py/h2o/explanation/_explain.py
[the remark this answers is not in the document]
sebhrusen: right! forgot about that… damn Py2.7…

h2o-r/h2o-package/R/classes.R
tomasfryda: True. Thank you.

ledell (approved):
Thank you!

PR #5257, opened 2021-01-27
PUBDEV-7862: Add h2o.get_automl alias for h2o.automl.get_automl

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7862

Add an alias of `h2o.get_automl()`, so that we have same function name in R/Py and it's easier to remember.

### DIFF
--- a/h2o-py/h2o/__init__.py
+++ b/h2o-py/h2o/__init__.py
@@ -24,6 +24,8 @@ from h2o.h2o import (connect, init, api, connection,
                      get_timezone, set_timezone, list_timezones,
                      load_dataset, demo, make_metrics, flow, upload_custom_metric, upload_custom_distribution,
                      import_mojo, upload_mojo, print_mojo, load_grid, save_grid)
+
+from h2o.automl import get_automl  # has to be here after "from h2o.h2o" to prevent cycles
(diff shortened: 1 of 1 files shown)


sebhrusen (changes requested):
Personally, I'm against the idea behind the entire PR.

Everything automl-related is already exposed in `h2o.automl` module, which is obvious enough to find.
I think that `h2o` root module should only expose `core` features, and if people start to find one AutoML function in the root module, they will expect to find more/others as well, and soon the concept of hierarchical modules in Python becomes pointless.

There are key differences between R and Py and, modularization is surely one of them. I don't think we should make Py code look R-like for the sake of  pseudo-compatibility: again, all automl functions today are in `h2o.automl`, it's a fact.

In Py, is a user wants to import a function in the current scope s/he just needs to do:
```
from h2o.automl import get_automl
```

tomasfryda:
@sebhrusen makes sense... So I should close this PR, right?  To which state should I put the JIRA (seems like the choices are OPEN, IN PROGRESS, CLOSED)?

ledell:
@sebhrusen That sounds fine to me; your reasoning makes sense.  @tomasfryda Sorry to waste your time.  I closed the JIRA as "Won't Fix".

PR #5258, opened 2021-01-27
Make sure cancelled models are cleaned-up in Grid Search

michalkurka:


h2o-core/src/main/java/water/Keyed.java
```diff
@@ -57,6 +58,13 @@ public static void remove( Key k ) {
     if (val==null) return;
     ((Keyed)val.get()).remove();
   }
+  public static void removeQuietly(Key k) {
+    try {
+      remove(k);
+    } catch (Exception e) {
+      Log.trace("Failed to remove key " + k, e);
```
sebhrusen, line 65: log level looks a bit too low: what about `warn` for the failure+error message, and `debug` for the stacktrace?

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -504,7 +516,28 @@ private Model buildModel(final MP params, Grid<MP> grid, int paramsIdx, String p
     }
 
     // Is there a model with the same params in the DKV?
-    @SuppressWarnings("unchecked") final Key<Model>[] modelKeys = KeySnapshot.globalSnapshot().filter(new KeySnapshot.KVFilter() {
+    Key<Model>[] modelKeys = findModelsByChecksum(checksum);
```
sebhrusen, line 519: this whole block is just refactoring, correct?

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -379,7 +379,19 @@ private void gridSearch(Grid<MP> grid) {
               ScoringInfo.sort(grid.getScoringInfos(), sortingMetric());
             }
           } catch (RuntimeException e) { // Catch everything
-            if (!Job.isCancelledException(e)) {
+            if (Job.isCancelledException(e)) {
+              assert model == null;
```
sebhrusen, line 383: is this assertion always true when using CV?
I think so, but I'm not 100% sure: it only depends on step 7 of CV building (`cv_mainModelScores`) never throwing any `CancelledException`.
I think it's ok, just want confirmation.

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -379,7 +379,19 @@ private void gridSearch(Grid<MP> grid) {
...
+              final long checksum = params.checksum(IGNORED_FIELDS_PARAM_HASH);
+              final Key<Model>[] modelKeys = findModelsByChecksum(checksum);
+              if (modelKeys.length == 1) {
+                Keyed.removeQuietly(modelKeys[0]);
+              } else if (modelKeys.length > 1) {
+                Log.err("Checksum " + checksum + " " +
```
sebhrusen, line 389: looks more like a warning

sebhrusen (commented):
Looks definitely more robust this way: added minor suggestions.

h2o-core/src/main/java/hex/grid/GridSearch.java
michalkurka: Yes, always true. `buildModel` either returns a model or throws an exception. If it throws an exception `model` will not be defined. The code after `buildModel` should never result in JobCancelledException. The assertion makes sure that stays true in the future as well. Otherwise, I would delete the model in DKV while keeping the model object around. Not the worst thing but better be safe.

h2o-core/src/main/java/hex/grid/GridSearch.java
michalkurka: yep, will change - thx

h2o-core/src/main/java/hex/grid/GridSearch.java
michalkurka: yes, no other changes

h2o-core/src/main/java/water/Keyed.java
michalkurka: fixed

sebhrusen (approved):
Thanks @michalkurka , hoping this will definitely close the issue with leaked keys in AutoML timeout test.

PR #5260, opened 2021-01-27
Clean-up GridSearch API and fix issue with working with previous model

michalkurka:


sebhrusen (approved):
300% in favour of this change!

honzasterba (approved):
👍

PR #5271, opened 2021-01-28
Build: Don't let AutoML tests run on micro_21 machines

michalkurka:


sebhrusen (approved):
You could also run Py3.6 on micro-nodes and skip them for Py 3.7.
This way, on Monday we'll have enough runs to compare.

michalkurka:
not needed anymore - https://github.com/h2oai/h2o-3/pull/5274

PR #5274, opened 2021-01-28
Fix infinite loop in CVModelBuilder

michalkurka:


sebhrusen (approved):
Thanks @michalkurka, another great bug killer for this bad bug :)

PR #5282, opened 2021-01-29
CountdownTest fix - make the test less time/scheduling dependent

michalkurka:


sebhrusen (approved):
Thanks for the fix, my test was a bit lousy indeed.

PR #5283, opened 2021-01-29
Switching to a safe version of NewChunk#bufX method due to crashes on Jenkins

michalkurka:
Jenkins builds showed the default implementation of NewChunk#bufX
can sometimes cause segfaults. We have seen similar behavior in AutoML
testing (rarely).

This was reported as a JDK bug, however, due to the use of Unsafe API
the bug was rejected. We are providing an implementation of the `bufX`
method that doesn't rely on Unsafe and is turned on by default in this change.

I investigated the behavior of the original `bufX` implementation and found that Java 11.0.8 doesn't suffer
from this issue.

    java version "11.0.8" 2020-07-14 LTS - OK (60 re-runs)
    java version "11.0.7" 2020-04-14 LTS - FAIL (1 in 3 runs fail)

Steps to reproduce

docker run -it -v $PWD/h2o-3-nightly-pipeline_master/java-11-junit/:/root/junit -v /home/0xdiag/smalldata/:/root/junit/h2o-3/smalldata harbor.h2o.ai/opsh2oai/h2o-3/dev-jdk-11:36 /bin/bash

cd root/junit/h2o-3/h2o-parsers/h2o-parquet-parser
java -version

for i in $(seq 3); do ../../gradlew clean build; done

### DIFF
--- a/h2o-core/src/main/java/water/fvec/NewChunk.java
+++ b/h2o-core/src/main/java/water/fvec/NewChunk.java
@@ -21,7 +21,7 @@ import static water.H2OConstants.MAX_STR_LEN;
 // An uncompressed chunk of data, supporting an append operation
 public class NewChunk extends Chunk {
 
-  private static final boolean DEBUG_SAFE_BUFX = H2O.getSysBoolProperty("debug.safe_bufx", false); 
+  private static final boolean USE_SAFE_BUFX = H2O.getSysBoolProperty("debug.safe_bufx", true); 
 
   private static final int[] EXP10s = new int[Double.MAX_EXPONENT - Double.MIN_EXPONENT + 1];
   private static final double[] INV_POW10s = new double[EXP10s.length];
@@ -1483,7 +1483,7 @@ public class NewChunk extends Chunk {
   // Compute a compressed integer buffer
   private byte[] bufX( long bias, int scale, int off, int log ) {
     byte[] bs = MemoryManager.malloc1((_len <<log)+off);
-    if (DEBUG_SAFE_BUFX && log > 0) {
+    if (USE_SAFE_BUFX && log > 0) {
       return bufX_safe(bs, bias, scale, off, log);
     }
     int j = 0;


h2o-core/src/main/java/water/fvec/NewChunk.java
```diff
@@ -21,7 +21,7 @@
 // An uncompressed chunk of data, supporting an append operation
 public class NewChunk extends Chunk {
 
-  private static final boolean DEBUG_SAFE_BUFX = H2O.getSysBoolProperty("debug.safe_bufx", false); 
+  private static final boolean USE_SAFE_BUFX = H2O.getSysBoolProperty("debug.safe_bufx", true);
```
sebhrusen, line 24: maybe you also want to change the sys property name as well: no one is aware of the current one anyway.

sebhrusen (approved):

h2o-core/src/main/java/water/fvec/NewChunk.java
michalkurka: but the purpose is still for debugging, right? maybe even more so now than before

sebhrusen: is it? I don't see what additional information it provides, and how it can help to figure out what is the issue with the unsafe version.
If you mean `debug` as "removing a bug", then yeah (it's a fix then), but `debug` as "trying to identify a bug" then for me it's not what it's doing.

michalkurka: I mean it in the sense that it is for "debugging"; no user should really be setting that flag.

I see it as a mechanism to revert the behavior back when I am debugging the old code. WDYT?

I can rename but my pattern is to use "debug" for all stuff that should be enabled/disabled in development mode.

sebhrusen: @michalkurka I see what you mean now. Agree that it's only for dev anyway, so I was not about to argue longer :)

PR #5288, opened 2021-01-31
Fix: Make sure onExceptionalCompletion is called non-CV models started from Grid Search

michalkurka:


michalkurka:
@sebhrusen please review, this PR will be followed by https://github.com/h2oai/h2o-3/pull/5289 (in draft currently)

h2o-core/src/main/java/hex/CVModelBuilder.java
```diff
@@ -52,13 +52,19 @@ public void bulkBuildModels() {
...
-                        int waitForTaskIndex = i + 1 - nRunning;
                         submodel_tasks[waitForTaskIndex].join();
                         finished(modelBuilders[waitForTaskIndex]);
                     } catch (RuntimeException t) {
-                        if (rt == null) rt = t;
+                        if (rt == null) {
+                            LOG.info("Exception from CV model #" + waitForTaskIndex + " will be reported as main exception.");
+                            rt = t;
+                        } else {
+                            LOG.warn("CV model #" + waitForTaskIndex + " failed, the exception will not be reported", t);
```
sebhrusen, line 64: out of scope question as it apparently used to work this way, but why do we wait for those loops to complete before throwing the exception.
Can't we also interrupt the loop (force nRunning = 0)
Same for the for-loop below.

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -413,7 +413,7 @@ final public M trainModelNested(Frame fr) {
     if (error_count() > 0)
       throw H2OModelBuilderIllegalArgumentException.makeFromBuilder(this);
     startClock();
-    if( !nFoldCV() ) trainModelImpl().compute2();
+    if( !nFoldCV() ) H2O.submitTask(trainModelImpl()).join();
```
sebhrusen, line 416: good catch! Just curious why it was implemented the way it was: by calling `compute2` directly, was it not completely ignoring the priority queue?

h2o-core/src/main/java/hex/CVModelBuilder.java
sebhrusen: Afterthoughts.
We probably have to keep the loop going, but then it would be to try to abort the running tasks before waiting for their completion: can't we do that the same way we try to abort jobs?

h2o-core/src/main/java/hex/ModelBuilder.java
michalkurka: IMHO this was done in error, not intentionally. I don't really see the purpose and since we have this working this way in the CV case anyway, I think we are fine here.

h2o-core/src/main/java/hex/CVModelBuilder.java
michalkurka: > can't we do that the same way we try to abort jobs?

We could absolutely do that. I am not sure what the motivation was here - maybe to give a cleaner clean-up workflow for the tasks that don't fail.

We cannot really say the caller can do anything with the succeded tasks anyway - there is no guarantee how many of them will be executed.

I think this could be revised. Also, the current way is not leveraging maximum parallelism right now: It starts 4 tasks and waits for them to finish, then it starts the last 5th task (which could really be running after one of the 4 finishes).

I think this is a candidate for a speedup. I also have an idea of running the main model alongside the CV tasks and let it advance when we know we can make additional iteration.

sebhrusen (approved):
Approving, we don't want to implement here the other improvements discussed in the comments anyway.

michalkurka:
@sebhrusen thanks for the review and the idea of how to further improve CVModelBuilder

PR #5289, opened 2021-02-01
Use a barrier to wait for completion of model Driver

michalkurka:


h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -457,6 +456,51 @@ public void run() {
...
 
+  private static class Barrier extends CountedCompleter {
+    @Override public void compute() { }
+  }
+
+  /**
+   * Simple wrapper around model task Driver, its main purpose is to make
+   * sure onExceptionalCompletion is not called after join method finishes (similarly how Job behaves).
+   */
+  class TrainModelTaskController {
```
sebhrusen, line 467: just wondering, do you think it is only needed for this type of task? or could it be useful as a more generic abstraction?
Rephrased: are all other tasks wrapped into a Job, and therefore they don't need this? My concern is that if we need an ad-hoc wrapper here, some other `MRTask` or `H2OCountedCompleter` instances may have a similar need.

sebhrusen (approved):
Nice improvement, I'm just concerned about the ad-hoc wrapper: wondering if we don't need an abstraction at the same level as Job (I can see using this in AutoML instead of polling jobs). This probably requires a separate discussion though.

h2o-core/src/main/java/hex/ModelBuilder.java
michalkurka: Agreed, might be needed in other places as well. 

MRTask should be fine, it has lots of special handling and `onExceptionalCompletion` is declared final. The only other place where I see a possible issue is FrameSplitter, however, that one I only see used in tests.

michalkurka:
> Nice improvement, I'm just concerned about the ad-hoc wrapper: wondering if we don't need an abstraction at the same level as Job (I can see using this in AutoML instead of polling jobs). This probably requires a separate discussion though.

I think you are right, the problem is more general and definitely exists in the codebase. It would be almost better if submitTask had this behavior by default and if you don't want it you would need to opt-out.

However, that would be a bigger change.

PR #5290, opened 2021-02-01
Change the automl eventlog tests to check for time difference instead of hour and day equality

tomasfryda:
It can rarely happen that the test would fail due to hour change between finishing autuml run and getting current time. To solve this I propose checking for time difference. I chose 15 minutes as it should be large enough to not fail due to slow communication and small enough to detect, e.g., wrong handling of time zones (AFAIK smallest time zone difference is 30 mins)

### DIFF
--- a/h2o-py/tests/testdir_algos/automl/pyunit_automl_event_log.py
+++ b/h2o-py/tests/testdir_algos/automl/pyunit_automl_event_log.py
@@ -24,8 +24,8 @@ def test_event_log():
     assert int(aml.training_info['stop_epoch']) > int(aml.training_info['start_epoch'])
     stop_dt = dt.datetime.fromtimestamp(int(aml.training_info['stop_epoch']))
     now = dt.datetime.now()
-    assert stop_dt.day == now.day
-    assert stop_dt.hour == now.hour
+    # test that stop_epoch is time encoded as unix epoch
+    assert abs(stop_dt - now) < dt.timedelta(minutes=1)
     assert abs(int(aml.training_info['duration_secs']) - (int(aml.training_info['stop_epoch']) - int(aml.training_info['start_epoch']))) <= 1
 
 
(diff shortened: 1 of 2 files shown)


h2o-py/tests/testdir_algos/automl/pyunit_automl_event_log.py
```diff
@@ -24,8 +24,7 @@ def test_event_log():
     assert int(aml.training_info['stop_epoch']) > int(aml.training_info['start_epoch'])
     stop_dt = dt.datetime.fromtimestamp(int(aml.training_info['stop_epoch']))
     now = dt.datetime.now()
-    assert stop_dt.day == now.day
-    assert stop_dt.hour == now.hour
+    assert abs(stop_dt - now) < dt.timedelta(minutes=15)
```
sebhrusen, line 27: even if `15 min` looks like an excessive delta (the test itself is supposed to last a few seconds…), I think it's kind of ok, the only goal of the previous assertions was to verify that the`epoch` was in the correct format, here Unix timestamp instead of millis.

The duration assertion was really done below.

sebhrusen: > wrong handling of time zones

unix epoch is always UTC.
Both `dt.datetime.fromtimestamp` and `dt.datetime.now` will return an object on the local timezone, so there shouldn't be any timezone issue.

tomasfryda: I wasn't really sure why there was the comparison of hours and days. I know we have `h2o.setTimezone(...)` so it was the only reason that I could think of for comparing hours (different time zone on client vs h2o) but looking at `ai.h2o.automl.events.EventLogEntry` it seems that we really use UTC no matter what the user sets. 

@sebhrusen Is this change PR ok or should I change anything?

sebhrusen: can you please just add a comment explaining this assertion as at first sight the 15min delta looks like it's not testing much: honestly if it takes 15min between the end of training on backend and the completion on client side, then there is a very serious problem somewhere.
I'd set the delta to `1min`, and mention that it's only verifying that `stop_epoch` was a unix epoch.

tomasfryda: Sure. Thanks!

sebhrusen (approved):

PR #5294, opened 2021-02-02
Fix Grid Search Fault Tolerance test

michalkurka:


h2o-core/src/main/java/water/api/JobsHandler.java
```diff
@@ -51,6 +52,17 @@ public JobsV3 cancel(int version, JobsV3 c) {
       throw new IllegalArgumentException("No job with key " + c.job_id.key());
     }
     j.stop(); // Request Job stop
+    long start = System.currentTimeMillis();
+    Log.info("Waiting for job " + c.job_id.key() + " to finishe execution.");
```
honzasterba, line 56: typo "finishe"

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -328,11 +328,14 @@ private void parallelGridSearch(final Grid<MP> grid) {
 
     if(!startModels.isEmpty()) {
       parallelModelBuilder.run(startModels);
-      parallelModelBuilder.join();
+      parallelModelBuilder.join(); // Warning: keep in mind this is not being executed as a F/J task, it just looks like one
     }
     grid.update(_job);
 
     attemptGridSave(grid);
+    if (_job.stop_requested()) {
```
honzasterba, line 336: at this point the grid is done, what is the point of throwing this exception?

michalkurka: it is not done, only the models that were executed before the user triggered the "cancel"

honzasterba (approved):

PR #5295, opened 2021-02-02
Upgrade Google Guava version in h2o-persist-hdfs to v24.1.1

this:
This PR upgrades the Google Guava transitive dependency version from v11.0.2 to v24.1.1 in order to circumvent security vulnerabilities found in v11.0.2.


this:
Instead of enforcing particular version of a transitive dependency, we will swap the h2o.jar in `h2o` Python module as suggested by @mmalohlava https://github.com/h2oai/h2o-3/pull/5296#issuecomment-771909174

PR #5296, opened 2021-02-02
Upgrade com.fasterxml.jackson.core:jackson-databind version h2o-persist-hdfs to v2.10.10

this:
This PR upgrades the `com.fasterxml.jackson.core:jackson-databind` transitive dependency version from v2.5.3 to v2.10.10 in order to circumvent security vulnerabilities found in v2.5.3.

### DIFF
--- a/h2o-persist-hdfs/build.gradle
+++ b/h2o-persist-hdfs/build.gradle
@@ -14,6 +14,7 @@ dependencies {
         transitive = true
     }
     compile("org.apache.hadoop:hadoop-aws:$defaultHadoopVersion")
+    compile("com.fasterxml.jackson.core:jackson-databind:2.10.0")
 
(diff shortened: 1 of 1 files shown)


mmalohlava:
Proposal for DAI context: bundle our own H2O.jar without HDFS client and inject into H2O-3 Python client.

@michalkurka do you see any danger?

michalkurka:
@mmalohlava I like this idea better, enforcing particular versions of dependencies for libraries we do not directly depend is not ideal

an alternative solution would be to use existing jars built for Hadoop 3.0 (any version that doesn't have the issue) and use it with the python client

this:
Instead of enforcing particular version of a transitive dependency, we will swap the h2o.jar in `h2o` Python module as suggested by @mmalohlava

PR #5301, opened 2021-02-03
Add an option to return k-means centers as a table instead of just a list

michalkurka:


sebhrusen (commented):
@michalkurka is it needed only for Py?
We can consider that this is already more easily accessible in R, but for the sake of consistency we probably want to have the equivalent there as well:
```R
# today, to get the whole table:
centers <- kmeans_model@model$centers
# only the centers:
centers <- h2o.centers(kmeans_model)
```

PR #5302, opened 2021-02-03
PUBDEV-7954: allow .xlsx feature interactions output

koniecsveta:


h2o-py/h2o/model/model_base.py
```diff
@@ -439,6 +441,12 @@ def feature_interaction(self, max_interaction_depth=100, max_tree_depth=100, max
             kwargs["max_deepening"] = max_deepening
             
             json = h2o.api("POST /3/FeatureInteraction", data=kwargs)
+            if path is not None:
+                import pandas as pd
+                writer = pd.ExcelWriter(path, engine='xlsxwriter')
```
michalkurka, line 446: this package is probably not in the docker images

h2o-py/tests/testdir_algos/xgboost/pyunit_feature_interactions.py
```diff
@@ -16,7 +16,14 @@ def xgboost_feature_interactions():
     feature_interactions = model.feature_interaction(2, 100, -1)
     assert isinstance(feature_interactions[0], H2OTwoDimTable)
     assert len(feature_interactions) == 11
-    
+
+    tempdir = "./"
```
honzasterba, line 20: usually for storing test outputs we use `path = pyunit_utils.locate("results")`
also I would call locate once not again on every line

koniecsveta: e9696f4

honzasterba (approved):

h2o-py/tests/testdir_algos/xgboost/pyunit_feature_interactions.py
```diff
@@ -16,7 +16,14 @@ def xgboost_feature_interactions():
     feature_interactions = model.feature_interaction(2, 100, -1)
     assert isinstance(feature_interactions[0], H2OTwoDimTable)
     assert len(feature_interactions) == 11
-    
+
+    path = pyunit_utils.locate("results")
+    feature_interactions = model.feature_interaction(2, 100, -1, path + 'feature_interactions.xlsx')
```
michalkurka, line 21: not that it matters but it is cleaner to use os.path.join

koniecsveta: 3cb4e85

PR #5310, opened 2021-02-08
PUBDEV-7998: Fixed ICE plots keys issue

sergeiten:
As explained in the ticket there is no keys for ICE plot:

https://h2oai.atlassian.net/browse/PUBDEV-7998

Not sure if I fixed this in right way but it's working for me. Made it same as PDP plots.

Screen Shot 2021-02-08 at 10 29 27

### DIFF
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -2143,24 +2143,22 @@ def explain(
         result["ice"]["description"] = display(Description("ice"))
         result["ice"]["plots"] = H2OExplanation()
         for column in columns_of_interest:
-            result["ice"]["plots"][columns] = H2OExplanation()
-            for model in models_to_show:
-                result["ice"]["plots"][columns][model] = H2OExplanation()
-                for target in targets:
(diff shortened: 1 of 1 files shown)


michalkurka:
thank you for the contribution @sergeiten!

tomasfryda (changes requested):
Thank you for investigating and reporting this issue and for creating this PR!

 I believe a proper fix would be replacing `result["ice"]["plots"][columns][model]` with `result["ice"]["plots"][columns][model.model_id]` (lines 2148, 2161, and 2163). This way we will plot one ICE plot for each column and model.

sergeiten:
I have added model id. I believe I should use "column" instead of "columns" so

Instead of:
```
result["ice"]["plots"][columns][model.model_id]
```

I have added:
```
result["ice"]["plots"][column][model.model_id]
```

tomasfryda:
Good catch @sergeiten. Thank you!

tomasfryda (approved):

ledell:
Thank you, @sergeiten!

PR #5311, opened 2021-02-08
Refactor build histogram

honzasterba:


michalkurka:
@honzasterba thank you for the clean-up! I have pending changes in the same code so please wait with merging

honzasterba:
this will need to be re-done, @maurever also has some pending changes in the code

PR #5313, opened 2021-02-08
PUBDEV-7992: Add more information about log files in User Guide

hannah-tillman:
For: https://h2oai.atlassian.net/browse/PUBDEV-7992


ledell (approved):
I added some more text to the Log File Structure section & made some other updates for clarity.  This looks good!  

However, I think we should create a new PR against master and not merge this one.

PR #5314, opened 2021-02-09
PUBDEV-8003: Fix notes about learn.h2o.ai

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8003

Minor documentation issues related to deep learning.

### DIFF
--- a/h2o-algos/src/main/java/hex/deeplearning/DeepLearningModelInfo.java
+++ b/h2o-algos/src/main/java/hex/deeplearning/DeepLearningModelInfo.java
@@ -205,7 +205,7 @@ final public class DeepLearningModelInfo extends Iced<DeepLearningModelInfo> {
         Log.warn(" *) Limit the size of the first hidden layer");
         if (dinfo._cats > 0) {
           Log.warn(" *) Limit the total number of one-hot encoded features with the parameter 'max_categorical_features'");
-          Log.warn(" *) Run h2o.interaction(...,pairwise=F) on high-cardinality categorical columns to limit the factor count, see http://learn.h2o.ai");
+          Log.warn(" *) Run h2o.interaction(...,pairwise=F) on high-cardinality categorical columns to limit the factor count, see http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/deep-learning.html#faq");
         }
         Log.warn("===================================================================================================================================");
       }
--- a/h2o-docs/src/product/data-science/deep-learning.rst
+++ b/h2o-docs/src/product/data-science/deep-learning.rst
@@ -217,7 +217,7 @@ H2O Deep Learning models have many input parameters, many of which are only acce
 
 -  **force_load_balance**: Specify whether to force extra load balancing to increase training speed for small datasets and use all cores. This option is defaults to true (enabled).
 
--  **variable_importances**: Specify whether to compute variable importance. This option is defaults to false (not enabled).
+-  **variable_importances**: Specify whether to compute variable importance. This option is defaults to true.
 
 -  **replicate_training_data**: Specify whether to replicate the entire training dataset onto every node for faster training on small datasets. This option is defaults to true (enabled).
 


h2o-algos/src/main/java/hex/deeplearning/DeepLearningModelInfo.java
```diff
@@ -205,7 +205,7 @@ public DeepLearningModelInfo(final DeepLearningParameters params, Key model_id,
         Log.warn(" *) Limit the size of the first hidden layer");
         if (dinfo._cats > 0) {
           Log.warn(" *) Limit the total number of one-hot encoded features with the parameter 'max_categorical_features'");
-          Log.warn(" *) Run h2o.interaction(...,pairwise=F) on high-cardinality categorical columns to limit the factor count, see http://learn.h2o.ai");
+          Log.warn(" *) Run h2o.interaction(...,pairwise=F) on high-cardinality categorical columns to limit the factor count, see http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/deep-learning.html#faq");
```
tomasfryda, line 208: `h2o.interaction` is the same for both python[1] and R. The only thing not right for python is the `pairwise=F` (`F` should be `False`) which I think is not a big deal and it keeps the message shorter


[1] https://github.com/h2oai/h2o-3/blob/03c73caf76c4eb4069e70490140435f0bc64f79d/h2o-py/h2o/frame.py#L4310

h2o-docs/src/product/data-science/deep-learning.rst
```diff
@@ -217,7 +217,7 @@ H2O Deep Learning models have many input parameters, many of which are only acce
 
 -  **force_load_balance**: Specify whether to force extra load balancing to increase training speed for small datasets and use all cores. This option is defaults to true (enabled).
 
--  **variable_importances**: Specify whether to compute variable importance. This option is defaults to false (not enabled).
+-  **variable_importances**: Specify whether to compute variable importance. This option is defaults to True.
```
tomasfryda, line 220: Defaults to `true` for the last 4 years. https://github.com/h2oai/h2o-3/blob/03c73caf76c4eb4069e70490140435f0bc64f79d/h2o-algos/src/main/java/hex/deeplearning/DeepLearningModel.java#L1617

h2o-docs/src/product/data-science/deep-learning.rst
```diff
@@ -217,7 +217,7 @@ H2O Deep Learning models have many input parameters, many of which are only acce
 
 -  **force_load_balance**: Specify whether to force extra load balancing to increase training speed for small datasets and use all cores. This option is defaults to true (enabled).
 
--  **variable_importances**: Specify whether to compute variable importance. This option is defaults to false (not enabled).
+-  **variable_importances**: Specify whether to compute variable importance. This option is defaults to true.
```
tomasfryda, line 220: Defaults to `true` for the last 4 years. https://github.com/h2oai/h2o-3/blob/03c73caf76c4eb4069e70490140435f0bc64f79d/h2o-algos/src/main/java/hex/deeplearning/DeepLearningModel.java#L1617

h2o-algos/src/main/java/hex/deeplearning/DeepLearningModelInfo.java
ledell: Should we also mention the `max_categorical_features` here?

We could use the text from the FAQ:  We recommend either reducing the number of categorical factor levels upfront (e.g., using h2o.interaction()), or specifying `max_categorical_features` to use feature hashing to reduce the dimensionality.

ledell (approved):

h2o-algos/src/main/java/hex/deeplearning/DeepLearningModelInfo.java
tomasfryda: It is mentioned on the previous line (207) which is actually how I found out about it.

PR #5318, opened 2021-02-11
PUBDEV-8005: Fix bug in varimp calculation in deeplearning with max_categorical_features

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8005

**BEFORE THIS PR**
Varimp calculation failed when `max_categorical_features` was set to lower number than the number of one-hot encoded categories.

This is due to the fact that encoded categories get hashed and binned to have at most `max_categorical_features` categorical features which means that multiple categories can get mapped into the same input neuron - this caused index out of bounds error during varimp calculation.

---

**Current Hacky Solution**
```
let 
   #encoded_cats == _catOffset[_catOffset.length - 1]    // number of "columns" after encoding the categories
in
```
When `max_categorical_features < #encoded_cats` precompute an array `categoryMapping` with length `#encoded_cats` that maps the categories to a subset of the input neuron layer, thus `inputLayer[cats[x]]` changes to`inputLayer[categoryMapping[cats[x]]`.  This is mainly to make it easier for POJO and MOJO implementations as it eliminates the need of including `hex/deeplearning/MurmurHash.java` - since the array is less than or equal to the number of columns it should not cause memory issues and hopefully shouldn't be much slower than the hash calculation itself (I guess it depends mainly on cache hits/misses since the Murmur hash seem to be very easy to calculate).

To create new names for the hashed variables, I join individual variable names with a comma but this doesn't seem to be a good solution, since they can get quite long (see the varimp plot below). Unfortunately, more than one "column" can be mapped to one input neuron (and vice versa) so abbreviation such as `CategoricalVar1.other_levels` is not possible to use here. 

These new variable names are then set to `DataInfo._coefNames`.

dl_varimp

**Issues I could think of**
- Problems with the explanations (one variable importance for multiple categorical variables)

### DIFF
--- a/h2o-algos/src/main/java/hex/deeplearning/DeepLearningModel.java
+++ b/h2o-algos/src/main/java/hex/deeplearning/DeepLearningModel.java
@@ -1064,9 +1064,9 @@ public class DeepLearningModel extends Model<DeepLearningModel,DeepLearningModel
       }
     });
     JCodeGen.toStaticVar(sb, "CATEGORY_MAPPING", model_info.categoryMapping,
-            "Map categories to their hashed location on input layer");
+            "Array that maps categories to their hashed location on input layer");
     JCodeGen.toStaticVar(sb, "MAX_CATEGORICAL_FEATURES",
-            model_info.get_params()._max_categorical_features, "Number of hashed categorical features");
+            model_info.get_params()._max_categorical_features, "Maximum number of categorical features");
 
     return sb;
   }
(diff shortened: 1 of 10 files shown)


tomasfryda:
Pausing the development of this PR. For automl, we have found out that `categorical_encoding=EnumLimited` and setting `max_categorical_levels` work just as well.

If there will be need to fix this, I will be happy to help but since we have more reliable workaround this is low priority for me now.

CC @maurever @wendycwong

wendycwong:
No problem.  Tomas.

PR #5319, opened 2021-02-11
[PUBDEV-7687] docker cleanup

honzasterba:


koniecsveta (approved):

scripts/jenkins/Makefile.jenkins
```diff
@@ -347,17 +319,17 @@ test-junit-7-smoke-jenkins:
 	@$(MAKE) -f $(THIS_FILE) test-junit-7-smoke
 
 test-junit-10-smoke:
-	DOONLY=$$(head -n 1 tests/doOnlyJunitSmokeTestList) ./gradlew h2o-core:test h2o-algos:test $$ADDITIONAL_GRADLE_OPTS
+	ADDITIONAL_TEST_JVM_OPTS="-Dsys.ai.h2o.debug.allowJavaVersions=10" DOONLY=$$(head -n 1 tests/doOnlyJunitSmokeTestList) ./gradlew h2o-core:test h2o-algos:test $$ADDITIONAL_GRADLE_OPTS
```
michalkurka, line 322: Why is this needed? To make sure we have the right java in the docker image?

honzasterba: honestly no idea what it does, was missing in some of the targets so I added it to make things consistent

and yes, this will check that the tests are actually running on the java version we think they are running on, makes sense

michalkurka: ok, it should be removed then

I checked and it adds a user-defined version to the list of supported versions

the reason why it was there was that someone forgot to remove it, it only should be there when you are eg. on a release branch and want to test against a new java and don't want to change it in the code (you don't know it works yet)

so you add this property for say Java 15, merge the PR and let the nightlies confirm there is no issue

then you go back to the code, modify the version check in java and remove this property - which is not needed anymore

docker/scripts/install_python_version
```diff
@@ -19,6 +19,7 @@ fi
 
 # Download requirements files from GitHub
 wget -nv https://raw.githubusercontent.com/h2oai/h2o-3/${H2O_BRANCH}/h2o-py/test-requirements.txt
+wget -nv https://raw.githubusercontent.com/h2oai/h2o-3/${H2O_BRANCH}/h2o-py/test-requirements-py3.txt
```
michalkurka, line 22: I think we should rather have test-requirements.txt for Python 3 (which most people use) and add test-requirements-py2.txt for legacy

=> py3 would be the expected default

docker/scripts/install_python_version
```diff
@@ -43,8 +44,7 @@ for python_version in "${array[@]}"; do
   rm -rf /var/cache/apt/*
 
   echo "###### Installing pip for Python ${python_version} ######"
-  python${python_version} -m easy_install pip
-  #python${python_version} -m pip -U install pip
+  python${python_version} -m easy_install 'pip<21'
```
michalkurka, line 47: I would only do this for Python 2.x, for the Python 3 we want to be moving forward

scripts/jenkins/Makefile.jenkins
honzasterba: ok, will do

docker/scripts/install_python_version
honzasterba: good call

docker/scripts/install_python_version
honzasterba: good call

docker/jenkins-images/Dockerfile-python-base
```diff
@@ -1,15 +1,16 @@
...
   DEBIAN_FRONTEND=noninteractive apt-get install -y \
-        gconf-service libasound2 libatk1.0-0 libatk-bridge2.0-0 libc6 libcairo2 libcups2 libdbus-1-3 libexpat1 libfontconfig1 libgcc1 libgconf-2-4 libgdk-pixbuf2.0-0 \
-        libglib2.0-0 libgtk-3-0 libnspr4 libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libx11-xcb-dev libxcb1 libxcomposite1 libxcomposite-dev libxcursor1 libxcursor-dev libxdamage1 libxext6 \
-        libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 ca-certificates fonts-liberation libappindicator1 libnss3 lsb-release xdg-utils libblas-dev liblapack-dev \
+        gconf-service libasound2 libatk1.0-0 libatk-bridge2.0-0 libc6 libcairo2 libcups2 libdbus-1-3 libexpat1 \
+        libfontconfig1 libgcc1 libgconf-2-4 libgdk-pixbuf2.0-0 libglib2.0-0 libgtk-3-0 libnspr4 libpango-1.0-0 \
+        libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libx11-xcb-dev libxcb1 libxcomposite1 libxcomposite-dev \
+        libxcursor1 libxcursor-dev libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 \
+        ca-certificates fonts-liberation libappindicator1 libnss3 lsb-release xdg-utils libblas-dev liblapack-dev \
```
honzasterba, line 13: only formatting

michalkurka (approved):
Great clean up! Thank you @honzasterba

honzasterba:
there 4 failing tests currently because of updated version of sklear for newer pythons (dependency of causalml), I propose to fix those in separate PR
@michalkurka  @Pscheidl  please give this another look

honzasterba:
there are currently only couple failing py tests testing the H2O with sklearn grid API ingegration
```
Py3.7 Small / pyunit_scal_pca_rf_grid.py
Py3.7 Small / pyunit_scal_svd_rf_grid.py
Py3.7 Small / pyunit_sklearn_classification_search_pipeline.py
Py3.7 Small / pyunit_sklearn_regression_search_pipeline.py
```
proposed fix for these is here: https://github.com/h2oai/h2o-3/pull/5331

scripts/jenkins/groovy/defineTestStages.groovy
```diff
@@ -87,7 +89,7 @@ def call(final pipelineContext) {
     [
       stageName: 'R3.5 Init Java 8', target: 'test-r-init', rVersion: '3.5.3', javaVersion: 8,
```
michalkurka, line 90: question: do we still need the java version specification? and r/py? it is specified in the docker image

michalkurka (approved):

scripts/jenkins/groovy/defineTestStages.groovy
honzasterba: no, once imageSpecifier is present, this info is ignored, but I decided to leave it here since it could be improved further so that the imageSpecifier is not necessary

PR #5320, opened 2021-02-11
GBM: Avoid scoring the last iteration twice when early stopping is enabled

michalkurka:


h2o-algos/src/main/java/hex/tree/SharedTree.java
```diff
@@ -409,10 +409,10 @@ protected final void resumeFromCheckpoint(SharedTree st) {
     protected final void scoreAndBuildTrees(boolean oob) {
       for( int tid=0; tid< _ntrees; tid++) {
         // During first iteration model contains 0 trees, then 1-tree, ...
-        boolean scored = doScoringAndSaveModel(false, oob, _parms._build_tree_one_node);
+        final boolean scored = doScoringAndSaveModel(false, oob, _parms._build_tree_one_node);
         if (scored && ScoreKeeper.stopEarly(_model._output.scoreKeepers(), _parms._stopping_rounds, getProblemType(), _parms._stopping_metric, _parms._stopping_tolerance, "model's last", true)) {
-          doScoringAndSaveModel(true, oob, _parms._build_tree_one_node);
           _job.update(_ntrees-_model._output._ntrees); //finish
+          LOG.info(_model.toString()); // we don't know if doScoringAndSaveModel printed the model or not
```
maurever, line 415: I am just curious, why we don't know if doScoringAndSaveModel printed the model or not?

h2o-algos/src/main/java/hex/tree/SharedTree.java
```diff
@@ -409,10 +409,10 @@ protected final void resumeFromCheckpoint(SharedTree st) {
     protected final void scoreAndBuildTrees(boolean oob) {
       for( int tid=0; tid< _ntrees; tid++) {
         // During first iteration model contains 0 trees, then 1-tree, ...
-        boolean scored = doScoringAndSaveModel(false, oob, _parms._build_tree_one_node);
+        final boolean scored = doScoringAndSaveModel(false, oob, _parms._build_tree_one_node);
         if (scored && ScoreKeeper.stopEarly(_model._output.scoreKeepers(), _parms._stopping_rounds, getProblemType(), _parms._stopping_metric, _parms._stopping_tolerance, "model's last", true)) {
-          doScoringAndSaveModel(true, oob, _parms._build_tree_one_node);
```
maurever, line 414: Is the parameter `finalScoring` still useful after your changes?

h2o-algos/src/main/java/hex/tree/SharedTree.java
michalkurka: Because not every scoring iteration prints out the model, this is to avoid too much output in the log.

We don't know if this condition was true or not:

      final boolean printout = (_parms._score_each_iteration || finalScoring || sinceLastScore > _parms._score_interval);

It is fine though, worst case we will print the model twice as before (which doesn't change with this PR).

h2o-algos/src/main/java/hex/tree/SharedTree.java
michalkurka: Yes, it forces the scoring to run regardless of other criteria. Still needed.

h2o-algos/src/main/java/hex/tree/SharedTree.java
maurever: Ok, thanks for the explanation.

maurever (approved):
Thanks for the PR and comments @michalkurka. It looks good to me.

michalkurka:
@maurever thank you for the review and the questions

PR #5321, opened 2021-02-13
GBM: Skip zero-weighted and NA-labeled data to speed-up CV and fix bugs

michalkurka:
This also fixes:
 - PUBDEV-6356 and PUBDEV-7553 (GBM was considering NA rows in training)

This change should have a positive impact on accuracy for datasets that
have many NAs in response.

### DIFF
--- a/h2o-algos/src/main/java/hex/tree/gbm/GBM.java
+++ b/h2o-algos/src/main/java/hex/tree/gbm/GBM.java
@@ -257,8 +257,16 @@ public class GBM extends SharedTree<GBMModel,GBMModel.GBMParameters,GBMModel.GBM
         Vec[] vecs = new Vec[]{_response};
         if (hasWeightCol())
           vecs = ArrayUtils.append(vecs, _weights);
-        for (int k = 0; k < _nclass; k++)
-          vecs = ArrayUtils.append(vecs, _train.vec(frameMap.nids0Index + k));
+        for (int k = 0; k < _nclass; k++) {
+          Vec nidsVecK = vec_nids(_train, k);
+          if (nidsVecK.min() == DECIDED_ROW) {
+            // classes not present in the training frame are skipped
+            assert _model._output._distribution[k] == 0;
+            assert nidsVecK.isConst();
+            continue;
+          }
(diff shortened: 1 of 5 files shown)


h2o-algos/src/main/java/hex/tree/gbm/GBM.java
```diff
@@ -246,8 +248,69 @@ protected Frame makeValidWorkspace() {
...
+        final int colStart;
+        final Chunk y;
+        final Chunk weights;
+        if (hasWeightCol()) {
+          y = cs[0];
+          weights = cs[1];
+          colStart = 2;
+        } else {
+          y = cs[0];
+          weights = new C0DChunk(1, cs[0]._len);
```
maurever, line 291: Is this chunk initialization necessary? You can edit the condition to ask first if `weights != null` or call `hasWeightColumn()` method instead.. Or am I missing something?

h2o-algos/src/main/java/hex/tree/gbm/GBM.java
```diff
@@ -246,8 +248,69 @@ protected Frame makeValidWorkspace() {
...
+          colStart = 1;
+        }
+        for (int row = 0; row < y._len; row++) {
+          if (weights.atd(row) == 0 || y.isNA(row)) {
+            markedCnt++;
+            for (int c = colStart; c < cs.length; c++) {
+              cs[c].set(row, DECIDED_ROW);
```
maurever, line 298: Cool improvement! :+1:

h2o-algos/src/main/java/hex/tree/gbm/GBM.java
michalkurka: This is a trick used oftentimes in the code at other places to avoid checking for null. It is fast and simplifies the flow of the code.

maurever: Thanks for the explanation. :+1:

maurever (approved):
LGTM. Thanks, @michalkurka.

michalkurka:
Thank you for the review @maurever

PR #5325, opened 2021-02-16
Fix issues with test leaks in GroupByTest and RapidsTest

michalkurka:


h2o-core/src/main/java/water/rapids/Env.java
```diff
@@ -51,6 +52,9 @@
  */
 public class Env extends Iced {
 
+  // for DEVELOPMENT/TESTING only
```
Pscheidl, line 55: Tiny tiny knit: Javadoc would show this description in IDEs as well :wink:

Pscheidl (approved):

h2o-core/src/main/java/water/rapids/Env.java
michalkurka: thank you @Pscheidl !

michalkurka:
thank you for the review and the suggestion @Pscheidl

PR #5327, opened 2021-02-16
[docker] switch cdh/hdp images to cloudera auth repos

honzasterba:


Pscheidl (approved):

michalkurka (approved):
Thank you @honzasterba!

docker/hadoop/cdh/Dockerfile
```diff
@@ -27,28 +29,17 @@ ENV DISTRIBUTION='cdh' \
     HIVE_PACKAGE=${HIVE_PACKAGE:-hive} \
     HIVE_HOME=/usr/lib/${HIVE_PACKAGE}
 
+COPY ${PATH_PREFIX}/scripts/sbin ${PATH_PREFIX}/../common/sbin scripts/install_python_version /usr/sbin/
 RUN \
```
michalkurka, line 33: @honzasterba should the RUN command be executed with the other following 2?

the apt cache clean-up is imho only done by install_python_version

    apt-get clean
    rm -rf /var/cache/apt/*

Are we keeping the caches in the layers created for each RUN?

michalkurka:
@honzasterba this looks really good - thank you

please see inline comment about a apt cache being preserved in older layers

PR #5330, opened 2021-02-17
Update deep learning documentation and log entry

tomasfryda:
Add `enum_limited` to DeepLearning's `categorical_encoding` documentation.
Mention `enum_limited` in warning when there is a lot of categories.


h2o-algos/src/main/java/hex/deeplearning/DeepLearningModelInfo.java
```diff
@@ -204,7 +204,8 @@ public DeepLearningModelInfo(final DeepLearningParameters params, Key model_id,
         Log.warn("Suggestions:");
         Log.warn(" *) Limit the size of the first hidden layer");
         if (dinfo._cats > 0) {
-          Log.warn(" *) Limit the total number of one-hot encoded features with the parameter 'max_categorical_features'");
+          Log.warn(" *) Limit the total number of one-hot encoded features by setting 'categorical_encoding=\"enum_limited\"'");
+          Log.warn(" *) Limit the total number of one-hot encoded features with the parameter 'max_categorical_features' (experimental)");
```
tomasfryda, line 208: Added `(experimental)` here so it reflects this comment[1] and the fact that when used it doesn't work with varimp to make the user to prefer the solution with the `enum_limited`.

[1] https://github.com/h2oai/h2o-3/blob/6573fa73353ba87f8cdebfe25ad3f9a2e9678635/h2o-algos/src/main/java/hex/deeplearning/DeepLearningModel.java#L1662

maurever (approved):
LGTM, thanks @tomasfryda !

wendycwong (approved):

PR #5336, opened 2021-02-19
[test] fix stage config for k8s tests

honzasterba:
- python is not available/needed on k8s images


Pscheidl:
I have Python installed in one of the test containers to run the scripts present on master. Shouldn't be affected.

Pscheidl (approved):

PR #5338, opened 2021-02-19
PUBDEV-8013: Make model@model$parameters$x report actual features

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8013

### DIFF
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -1307,9 +1307,15 @@ def _get_xy(model):
     :param model: H2O Model
     :returns: tuple (x, y)
     """
+    names = model._model_json["output"]["original_names"] or model._model_json["output"]["names"]
     y = model.actual_params["response_column"]
-    x = [feature for feature in model._model_json["output"]["names"]
-         if feature not in y]
+    not_x = [
+                y,
+                # if there is no fold column, fold_column is set to None, thus using "or {}" instead of the second argument of dict.get
+                (model.actual_params.get("fold_column") or {}).get("column_name"),
+                (model.actual_params.get("weights_column") or {}).get("column_name")
+             ] + (model.actual_params.get("ignored_columns") or [])
+    x = [feature for feature in names if feature not in not_x]
     return x, y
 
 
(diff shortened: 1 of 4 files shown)


michalkurka:
@koniecsveta can you please review this once it is ready? you were making changes recently to the surrounding code base

h2o-py/h2o/explanation/_explain.py
```diff
@@ -1307,9 +1307,15 @@ def _get_xy(model):
...
     y = model.actual_params["response_column"]
-    x = [feature for feature in model._model_json["output"]["names"]
-         if feature not in y]
+    not_x = [
+                y,
+                # if there is no fold column, fold_column is set to None, thus using "or {}" instead of the second argument of dict.get
+                (model.actual_params.get("fold_column") or {}).get("column_name"),
+                (model.actual_params.get("weights_column") or {}).get("column_name")
+             ] + (model.actual_params.get("ignored_columns") or [])
+    x = [feature for feature in names if feature not in not_x]
```
tomasfryda, line 1318: I wrote this as a copy of the incorrect R code so I included the fix here to keep it in sync even though the name of the PR suggest it involves just R.

tomasfryda:
Converting back to draft since I forgot about offset column 🤦

h2o-py/h2o/explanation/_explain.py
koniecsveta: can you add a test capturing the bug?

koniecsveta (changes requested):
can you add a test capturing this bug?

h2o-py/tests/testdir_misc/pyunit_explain.py
```diff
@@ -11,6 +11,28 @@
 from h2o.estimators import H2OGradientBoostingEstimator
 from h2o.explanation._explain import H2OExplanation
 
+
```
koniecsveta, line 14: Would call this pyunit_pubdev_8013  and would put inside of thestdir_jira instead of testdir_misc to be inline with the second test.

h2o-py/h2o/explanation/_explain.py
tomasfryda: done

ledell (approved):
Thank you!

koniecsveta (approved):

PR #5343, opened 2021-02-23
PUBDEV-8015: Remove warning in Stacked Ensemble predict function about missing fold_column frame

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8015

Setting `fold_column = metalearner_fold_column` resolves the issue with the warning. It seems to me as a better solution than putting a SE specific logic to `hex.Model#adaptTestForTrain`.

Other approaches that I can think of:
1) use only `fold_column` - @ledell wrote good rationale why to keep it `metalearner_fold_column` https://h2oai.atlassian.net/browse/PUBDEV-7675?focusedCommentId=57565
2) use `metalearner_fold_column` only on r/py client, e.g., setter would set `metalearner_params["fold_column"]` - I don't like it since it creates different API between r/py and java
3) modify `hex.Model#adaptTestForTrain` and add logic specific to SE there

Am I missing some better solution?

### DIFF
--- a/h2o-core/src/main/java/hex/Model.java
+++ b/h2o-core/src/main/java/hex/Model.java
@@ -1541,7 +1541,8 @@ public abstract class Model<M extends Model<M,P,O>, P extends Model.Parameters, 
           vec = test.anyVec().makeCon(defval);
           toDelete.put(vec._key, "adapted missing vectors");
           String str = "Test/Validation dataset is missing column '" + names[i] + "': substituting in a column of " + defval;
-          if (isResponse || isWeights || isFold)
+
+          if (ArrayUtils.contains(parms.getNonPredictors(), names[i]))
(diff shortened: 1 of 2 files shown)


sebhrusen:
Suggestion #4:

modify `Model.adaptTestForTrain` and replace:
```java
          if (isResponse || isWeights || isFold)
            Log.info(str); // we are doing a "pure" predict (computeMetrics is false), don't complain to the user
```
with
```java
     if (ArrayUtils.contains(params.getNonPredictors(),names[i])
            Log.info(str); // we are doing a "pure" predict (computeMetrics is false), don't complain to the user
```

now, we can override `getNonPredictors` in `StackedEnsembleParameters` to include `_metalearner_fold_column`.

imo, the best solution :)

ledell (approved):
It looks like you found a good solution (mentioned by @sebhrusen).   I think it's ready to be merged.

PR #5347, opened 2021-02-23
PUBDEV 7798: Multiple knots at same coordinates

km2018:
JIRA: https://h2oai.atlassian.net/browse/PUBDEV-7798

This pull request solves the issue of GAM creating multiple knots at the same coordinates. This issue occurs when the cardinality of the gam columns is less than the number of knots specified by the user. To solve the problem, a check is added in GAM's validateGamParameters function to confirm that this is not the case. A unit test has been added as well.


h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -233,6 +233,9 @@ private void validateGamParameters() {
         if (!dataset.vec(cname).isNumeric())
           error("gam_columns", "column " + cname + " is not numerical and cannot be used as a gam" +
                   " column.");
+        if (dataset.vec(cname).isInt() && ((dataset.vec(cname).max() - dataset.vec(cname).min() + 1) < _parms._num_knots[index]))
+          error("gam_columns", "column " + cname + " has cardinality lower than the number of knots and cannot be used as a gam" +
+                  " column.");
       }
```
wendycwong, line 239: This is a good first order check.  However, even if this check passes, failures can still occur if say we have 100 rows of 1 and 100 rows of 100 and numknot = 1.  It is necessary but not sufficient.  Since you kept the old verification check, it should not be a problem.

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7798_overlapped_knots.py
```diff
@@ -0,0 +1,36 @@
...
+                                                gam_columns=feature_names,
+                                                scale=[1],
+                                                num_knots=[10],
+                                                )
+    try:
+        h2o_model.train(x=feature_names, y='target', training_frame=train)
+        assert False, "Number of knots validation should have failed"
+    except:
+        print("Error correctly raised when cardinality < num_knots")
+    else:
```
wendycwong, line 29: Karthik:  Please check the actual error message and make sure you got the correct one.  You can do something like this:

    try:
        glm.makeGLMModel(model=m,coefs=coefs)
        assert False, "Should have throw exception of bad coefficient length"
    except Exception as ex:
        print(ex)
        temp = str(ex)
        assert ("Server error java.lang.IllegalArgumentException:" in temp) and \
               ("model coefficient length 9 is different from coefficient provided by user ") in temp, \
            "Wrong exception was received."
        print("coefficient test passed!")

wendycwong (changes requested):

wendycwong (approved):

h2o-algos/src/main/java/hex/gam/GAM.java
```diff
@@ -233,6 +233,9 @@ private void validateGamParameters() {
         if (!dataset.vec(cname).isNumeric())
           error("gam_columns", "column " + cname + " is not numerical and cannot be used as a gam" +
                   " column.");
+        if (dataset.vec(cname).isInt() && ((dataset.vec(cname).max() - dataset.vec(cname).min() + 1) < _parms._num_knots[index]))
```
wendycwong: @km2018 please move this test to below line 255 where _parms._num_knots is defined if users do not specify it.

wendycwong (changes requested):

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7798_overlapped_knots.py
```diff
@@ -0,0 +1,37 @@
...
+    try:
+        h2o_model.train(x=feature_names, y='target', training_frame=train)
+        assert False, "Number of knots validation should have failed"
+    except Exception as ex:
+        exception = str(ex)
+        assert ("H2OModelBuilderIllegalArgumentException" in exception) or \
+               ("has cardinality lower than the number of knots" in exception) or \
+               ("does not have enough values to generate well-defined knots" in exception)
+        print("Error correctly raised when cardinality < num_knots")
+
```
wendycwong, line 33: karthik:  Please make sure your error message is included ("has cardinality lower than the number of knots" in exception).  Can you change this into some kind of and instead or or?

wendycwong (changes requested):

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7798_overlapped_knots.py
```diff
@@ -0,0 +1,39 @@
...
+        exception = str(ex)
+        assert ("H2OModelBuilderIllegalArgumentException" in exception) and \
+               (("has cardinality lower than the number of knots" in exception) or \
+               ("does not have enough values to generate well-defined knots" in exception))
+        print("Error correctly raised when cardinality < num_knots")
+
+if __name__ == "__main__":
+    h2o.init(ip='192.168.1.163', port=54321, strict_version_check=False)
+    pyunit_utils.standalone_test(knots_error())
+else:
```
wendycwong, line 37: remove h2o.init

wendycwong (changes requested):

h2o-py/tests/testdir_algos/gam/pyunit_PUBDEV_7798_overlapped_knots.py
wendycwong: Can you just use and and not use "or".

wendycwong (changes requested):

wendycwong (approved):

PR #5350, opened 2021-02-25
Avoid double-scoring of CV models in AutoML 

michalkurka:


h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -183,23 +183,43 @@ protected void map(int id) {
...
+    }
+
+    @Override
+    public ModelMetrics getOrMakeMetrics(Frame fr, Frame adaptFrm) {
+      return _modelMetrics;
+    }
+
+    @Override
+    public ModelMetrics.MetricBuilder<?> getOrMakeMetricBuilder(Frame adaptFrm) {
+      return StackedEnsembleModel.this.scoreMetrics(adaptFrm);
```
tomasfryda, line 219: I don't think you have to override this method. 

We don't have working implementation `StackedEnsembleModel#makeMetricBuilder` but also we don't have working implementation of `StackedEnsembleModel#score0` which is used by `BigScore#map` and using that by `Model#scoreMetrics`. 

Either way I think this call would not work so I'd rather keep it consistent with the parent class (and eventually make SE more like the rest of the models).

michalkurka: @tomasfryda thanks for the comment, I get it now, there is no CV for Stacked Ensembles as such, there is CV on the meta learner level but not on the outer SE model

this means I can simplify the code quite a bit!

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1859,12 +1861,44 @@ protected Frame predictScoreImpl(Frame fr, Frame adaptFrm, String destination_ke
...
+      _mb = mb;
+      _rawPreds = rawPreds;
+      _outputPreds = outputPreds;
+    }
+    
+    public final Frame getPredictions() {
+      return _outputPreds;
+    }
+
+    public ModelMetrics.MetricBuilder<?> getMetricBuilder(Frame adaptFrm) {
```
tomasfryda, line 1890: `Frame adaptFrm` is unused. Is it a remnant of the previous version (IIRC it was used in `StackedEnsemblePredictScoreResult`)?

michalkurka: good catch, I need to remove this argument as well

tomasfryda (approved):

michalkurka:
@tomasfryda thanks for the review and catching the unnecessary complexity built for SE

PR #5353, opened 2021-03-01
PUBDEV-7735: Add baseline hazard function estimate to CoxPH model

satai:


h2o-algos/src/main/java/hex/coxph/CoxPH.java
```diff
@@ -547,15 +556,59 @@ protected void calcCumhaz_0(CoxPHModel model, final CoxPHTask coxMR) {
...
+        baselineHazardAsFrame.setNames(new String[]{"t", "baseline hazard"});
+        baselineSurvivalAsFrame.setNames(new String[]{"t", "baseline survival"});
+      } else {
+        List<String> names = o._strataMap.entrySet().stream()
+                .sorted(Comparator.comparingInt(e -> e.getValue()._val))
+                .map(Map.Entry::getKey)
+                .map(i -> i._gs)
+                .map(a -> Arrays.stream(a).mapToInt(d -> (int)d).toArray())
+                .map(Arrays::toString)
+                .map(s -> s.replace(']',')').replace('[', '('))
```
michalkurka, line 606: Creative but it would look nicer with a utility method that produces output that you expect (ArrayUtils)

h2o-algos/src/main/java/hex/coxph/CoxPH.java
```diff
@@ -727,6 +783,7 @@ public void computeImpl() {
 
     @Override
     protected void chunkInit(){
+      System.out.println("_time = " + Arrays.toString(_time));
```
michalkurka, line 786: you can also use the profiling feature if that is what you are doing - MRTask.profile()...

michalkurka (approved):
LGTM, thank you @satai

PR #5356, opened 2021-03-02
PUBDEV-8032: Stacked Ensemble: Take logit of cvpreds before training the metalearner for classification

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8032

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -221,10 +221,10 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
       Frame old = DKV.getGet(levelOneKey);
       if (old != null && old instanceof Frame) {
         Frame oldFrame = (Frame) old;
+        oldFrame.write_lock(_job);
         // Remove ALL the columns so we don't delete them in remove_impl.  Their
         // lifetime is controlled by their model.
         oldFrame.removeAll();
-        oldFrame.write_lock(_job);
         oldFrame.update(_job);
         oldFrame.unlock(_job);
       }
@@ -252,8 +252,8 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
 
         Frame oldLOF = levelOneFrame;
         levelOneFrame = _parms._metalearner_transform.transform(_model, levelOneFrame);
-        oldLOF.removeAll();
         oldLOF.write_lock(_job);
+        oldLOF.removeAll();
         oldLOF.update(_job);
         oldLOF.unlock(_job);
       }
@@ -261,7 +261,7 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
       // Add metalearner fold column, weights column to level one frame if it exists
       addMiscColumnsToLevelOneFrame(_model._parms, actuals, levelOneFrame, true);
 
-      levelOneFrame.write_lock(_job);
+      levelOneFrame.delete_and_lock(_job);
       levelOneFrame.unlock(_job);
       Log.info("Finished creating \"level one\" frame for stacking: " + levelOneFrame.toString());
       DKV.put(levelOneFrame);
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -247,18 +247,15 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
         Scope.untrack(baseModelPredictions);
       }
       if (_parms._metalearner_transform != null && _parms._metalearner_transform != StackedEnsembleModel.StackedEnsembleParameters.MetalearnerTransform.NONE) {
-        if ((_parms._metalearner_parameters._distribution != DistributionFamily.bernoulli) &&
-                (!(_parms._metalearner_parameters instanceof GLMModel.GLMParameters) ||
-                ((GLMModel.GLMParameters)_parms._metalearner_parameters)._family != GLMModel.GLMParameters.Family.binomial))
-          throw new H2OIllegalArgumentException("Metalearner transform is supported only for bernoulli distribution!");
+        if (!(_model._output.isBinomialClassifier() || _model._output.isMultinomialClassifier()))
+          throw new H2OIllegalArgumentException("Metalearner transform is supported only for classification!");
+
         Frame oldLOF = levelOneFrame;
-        levelOneFrame = _parms._metalearner_transform.task().doAll(levelOneFrame.numCols(), Vec.T_NUM, levelOneFrame)
-                .outputFrame(levelOneFrame._key, levelOneFrame._names, null);
+        levelOneFrame = _parms._metalearner_transform.transform(_model, levelOneFrame);
         oldLOF.removeAll();
         oldLOF.write_lock(_job);
         oldLOF.update(_job);
         oldLOF.unlock(_job);
-
       }
 
       // Add metalearner fold column, weights column to level one frame if it exists
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -1,15 +1,17 @@
 package hex.ensemble;
 
-import hex.Model;
-import hex.ModelBuilder;
-import hex.ModelCategory;
+import hex.*;
 
 import hex.genmodel.utils.DistributionFamily;
+import hex.genmodel.utils.LinkFunctionType;
+import hex.glm.GLMModel;
 import hex.grid.Grid;
 import water.*;
 import water.exceptions.H2OIllegalArgumentException;
 import water.exceptions.H2OModelBuilderIllegalArgumentException;
+import water.fvec.Chunk;
 import water.fvec.Frame;
+import water.fvec.NewChunk;
 import water.fvec.Vec;
 
 import java.util.*;
@@ -214,7 +216,20 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
         throw new H2OIllegalArgumentException("Base models and prediction arrays are different lengths.");
 
       if (null == levelOneKey) levelOneKey = "levelone_" + _model._key.toString() + "_" + _parms._metalearner_transform.toString();
-      Frame levelOneFrame = new Frame(Key.<Frame>make(levelOneKey));
+
+      // TODO: what if we're running multiple in parallel and have a name collision?
+      Frame old = DKV.getGet(levelOneKey);
+      if (old != null && old instanceof Frame) {
+        Frame oldFrame = (Frame) old;
+        // Remove ALL the columns so we don't delete them in remove_impl.  Their
+        // lifetime is controlled by their model.
+        oldFrame.removeAll();
+        oldFrame.write_lock(_job);
+        oldFrame.update(_job);
+        oldFrame.unlock(_job);
+      }
+
+      Frame levelOneFrame = new Frame(Key.make(levelOneKey));
 
       for (int i = 0; i < baseModels.length; i++) {
         Model baseModel = baseModels[i];
@@ -229,26 +244,27 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
           continue;
         }
         StackedEnsemble.addModelPredictionsToLevelOneFrame(baseModel, baseModelPreds, levelOneFrame);
+        Scope.untrack(baseModelPredictions);
       }
       if (_parms._metalearner_transform != null && _parms._metalearner_transform != StackedEnsembleModel.StackedEnsembleParameters.MetalearnerTransform.NONE) {
-        _parms._metalearner_transform.task.doAll(levelOneFrame);
+        if ((_parms._metalearner_parameters._distribution != DistributionFamily.bernoulli) &&
+                (!(_parms._metalearner_parameters instanceof GLMModel.GLMParameters) ||
+                ((GLMModel.GLMParameters)_parms._metalearner_parameters)._family != GLMModel.GLMParameters.Family.binomial))
+          throw new H2OIllegalArgumentException("Metalearner transform is supported only for bernoulli distribution!");
+        Frame oldLOF = levelOneFrame;
+        levelOneFrame = _parms._metalearner_transform.task().doAll(levelOneFrame.numCols(), Vec.T_NUM, levelOneFrame)
+                .outputFrame(levelOneFrame._key, levelOneFrame._names, null);
+        oldLOF.removeAll();
+        oldLOF.write_lock(_job);
+        oldLOF.update(_job);
+        oldLOF.unlock(_job);
+
       }
+
       // Add metalearner fold column, weights column to level one frame if it exists
       addMiscColumnsToLevelOneFrame(_model._parms, actuals, levelOneFrame, true);
 
-      // TODO: what if we're running multiple in parallel and have a name collision?
-      Frame old = DKV.getGet(levelOneFrame._key);
-      if (old != null && old instanceof Frame) {
-        Frame oldFrame = (Frame) old;
-        // Remove ALL the columns so we don't delete them in remove_impl.  Their
-        // lifetime is controlled by their model.
-        oldFrame.removeAll();
-        oldFrame.write_lock(_job);
-        oldFrame.update(_job);
-        oldFrame.unlock(_job);
-      }
-
-      levelOneFrame.delete_and_lock(_job);
+      levelOneFrame.write_lock(_job);
       levelOneFrame.unlock(_job);
       Log.info("Finished creating \"level one\" frame for stacking: " + levelOneFrame.toString());
       DKV.put(levelOneFrame);
@@ -337,11 +353,11 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
         _model.unlock(_job);
       }
 
-      String levelOneTrainKey = "levelone_training_" + _model._key.toString() + "_" + _parms._metalearner_transform;
+      String levelOneTrainKey = "levelone_training_" + _model._key.toString();
       Frame levelOneTrainingFrame = prepareLevelOneFrame(levelOneTrainKey, _model._parms._base_models, getActualTrainingFrame(), true);
       Frame levelOneValidationFrame = null;
       if (_model._parms.valid() != null) {
-        String levelOneValidKey = "levelone_validation_" + _model._key.toString() + "_" + _parms._metalearner_transform;
+        String levelOneValidKey = "levelone_validation_" + _model._key.toString();
         levelOneValidationFrame = prepareLevelOneFrame(levelOneValidKey, _model._parms._base_models, _model._parms.valid(), false);
       }
 
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -6,10 +6,7 @@ import hex.ModelCategory;
 
 import hex.genmodel.utils.DistributionFamily;
 import hex.grid.Grid;
-import water.DKV;
-import water.Job;
-import water.Key;
-import water.Scope;
+import water.*;
 import water.exceptions.H2OIllegalArgumentException;
 import water.exceptions.H2OModelBuilderIllegalArgumentException;
 import water.fvec.Frame;
@@ -216,7 +213,7 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
       if (baseModels.length != baseModelPredictions.length)
         throw new H2OIllegalArgumentException("Base models and prediction arrays are different lengths.");
 
-      if (null == levelOneKey) levelOneKey = "levelone_" + _model._key.toString();
+      if (null == levelOneKey) levelOneKey = "levelone_" + _model._key.toString() + "_" + _parms._metalearner_transform.toString();
       Frame levelOneFrame = new Frame(Key.<Frame>make(levelOneKey));
 
       for (int i = 0; i < baseModels.length; i++) {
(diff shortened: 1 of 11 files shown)


h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -74,6 +75,58 @@ public void initActualParamValues() {
...
+    public enum MetalearnerTransform {
+      NONE(null),
+      Logit(new MRTask() {
+        @Override
+        public void map(Chunk[] cs) {
+          double p, logit;
+          for (int c = 0; c < cs.length; c++) {
+            for (int i = 0; i < cs[c]._len; i++) {
+              p = cs[c].atd(i);
+              logit = Math.log(p/(1-p));
```
michalkurka, line 87: you can either use existing LogitFunction or just use LogExpUtil.log that prevents issues

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -74,6 +75,58 @@ public void initActualParamValues() {
...
+              cs[c].set(i, Double.isInfinite(logit) ? Double.NaN : logit);  // Algos have troubles dealing with Inf
```
michalkurka, line 88: are you sure it is okay to mutate the frame? are you changing the prediction frames?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
tomasfryda: Thanks for the tip!

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
tomasfryda: Thanks, good point!
No, it is not safe. The frame is created by taking one prediction column (vec) from each base model the predictions can be "CV predictions" in normal stacking mode so that shouldn't be safe. 

There are two uses of that MRTask - training and scoring. I think the mutation during scoring should be OK. During training it could modify cv predictions so it is not OK. 

I tried using `map(Chunk[], NewChunk[])` and `doAll(int numberOfOutputs, byte outputType, Frame fr).outputFrame(Key, String[], String[][])`, but somewhere under the hood `vecs` are deleted during second run - I guess SE specific logic for handling `levelOneFrame`, so I am looking more in to that.

tomasfryda: Good point, thanks! Sorry for my late reply - somehow I replied in "review" mode and I didn't finish the review and noticed it today...

I was modifying prediction frames and also cv predictions - which is not okay... I struggled with missing chunks during second run which I found out was due to reusing the same instance of `MRTask` 🤦 .

Now the goal is to have it in a state to work well enough to be able to benchmark it to find out if those transformations help SE be better.

ledell:
@tomasfryda Is this branch ready to benchmark?  If it works better, then we can make this the default and then add a parameter to the Stacked Ensemble which allows the user to turn it on/off (in a way that would allow us to add future transformations as well such as percentile rank).  `levelone_transform = c("logit", "percentile_rank")` or `metalearner_transform`.

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -216,8 +215,21 @@ private Frame prepareLevelOneFrame(String levelOneKey, Model[] baseModels, Frame
...
+      if (null == levelOneKey) levelOneKey = "levelone_" + _model._key.toString() + "_" + _parms._metalearner_transform.toString();
+
+      // TODO: what if we're running multiple in parallel and have a name collision?
+      Frame old = DKV.getGet(levelOneKey);
+      if (old != null && old instanceof Frame) {
+        Frame oldFrame = (Frame) old;
+        // Remove ALL the columns so we don't delete them in remove_impl.  Their
+        // lifetime is controlled by their model.
+        oldFrame.removeAll();
+        oldFrame.write_lock(_job);
```
michalkurka, line 227: you need to get a write-lock prior to changing a frame, otherwise, you might be changing an instance that is already locked

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -232,24 +244,24 @@ private Frame prepareLevelOneFrame(String levelOneKey, Model[] baseModels, Frame
...
+        Scope.untrack(baseModelPredictions);
+      }
+      if (_parms._metalearner_transform != null && _parms._metalearner_transform != StackedEnsembleModel.StackedEnsembleParameters.MetalearnerTransform.NONE) {
+        if (!(_model._output.isBinomialClassifier() || _model._output.isMultinomialClassifier()))
+          throw new H2OIllegalArgumentException("Metalearner transform is supported only for classification!");
+
+        Frame oldLOF = levelOneFrame;
+        levelOneFrame = _parms._metalearner_transform.transform(_model, levelOneFrame);
+        oldLOF.removeAll();
+        oldLOF.write_lock(_job);
```
michalkurka, line 256: same as above

michalkurka:
It would be good to disable MOJO download for the cases when a transform is applied (as that is currently not supported) - better to have that now rather than to accidentally forget about it and let users have invalid predictions in production.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -74,6 +77,76 @@ public void initActualParamValues() {
...
+          for (int i = 0; i <= N; i++)
+            probs[i] = i/(double)N;
+          if (null == model._output._metalearner_percentile_rank_precomputed_quantiles) {
+            // Compute quantiles for specified probs
+            DKV.put(frame);
+            QuantileModel.QuantileParameters qp = new QuantileModel.QuantileParameters();
+            qp._train = frame._key;
+            qp._probs = probs;
+            QuantileModel qm = new Quantile(qp).trainModel().get();
+            model._output._metalearner_percentile_rank_precomputed_quantiles = qm._output._quantiles;
```
michalkurka, line 113: you are modifying a model without any explicit update mechanism, the update will or won't be preserved depending on the context where this code is running

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
tomasfryda: Thank you! I was wondering about that - this code was just moved couple lines up. This was there for the last 4 years so I was assumed it was correct.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
tomasfryda: Thank you, I really appreciate you spending time on helping me making this PR better.
I hope I fixed it by `model.write_lock()`, the assignment, `model.update()`, `model.unlock()`.

h2o-r/h2o-package/R/stackedensemble.R
```diff
@@ -112,6 +114,7 @@ h2o.stackedEnsemble <- function(x,
                                 metalearner_fold_assignment = c("AUTO", "Random", "Modulo", "Stratified"),
                                 metalearner_fold_column = NULL,
                                 metalearner_params = NULL,
+                                metalearner_transform = c("NONE", "Logit", "PercentileRank"),
```
ledell, line 117: Maybe we should use:  `c("none", "logit", "percentile_rank" )` instead?  That way it can match Python too.

tomasfryda: This is generated from the schema - I don't know if I can influence this.

h2o-genmodel/src/main/java/hex/genmodel/algos/ensemble/StackedEnsembleMojoReader.java
```diff
@@ -17,6 +17,9 @@ protected void readParentModelData() throws IOException {
         int baseModelNum = readkv("base_models_num", 0);
         _model._baseModelNum = baseModelNum;
         _model._metaLearner = getModel((String) readkv("metalearner"));
+        _model._metaLearnerTransform = (String) readkv("metalearner_transform", "NONE");
+        if (!_model._metaLearnerTransform.equals("NONE") && !_model._metaLearnerTransform.equals("Logit"))
```
michalkurka, line 21: this is a good thing to have but we also need to make sure such mojo couldn't even reach the code (by version checking)

h2o-genmodel/src/main/java/hex/genmodel/algos/ensemble/StackedEnsembleMojoModel.java
```diff
@@ -27,12 +38,16 @@ public StackedEnsembleMojoModel(String[] columns, String[][] domains, String res
                     basePreds[i * _nclasses + j] = _baseModels[i]._mojoModel.score0(_baseModels[i].remapRow(row), basePredsRow)[j + 1];
                 }
             }
+            if (_metaLearnerTransform.equals("Logit"))
```
michalkurka, line 41: Since you only have logit transform I think in the reader you could just generate a boolean value to avoid the repeated string comparison.

michalkurka:
Looks good @tomasfryda!

I might have missed it but I didn't see any test that would be calling testJavaScoring to compare results between MOJO and SE model.

h2o-genmodel/src/main/java/hex/genmodel/algos/ensemble/StackedEnsembleMojoModel.java
tomasfryda: Good idea, thanks!

h2o-genmodel/src/main/java/hex/genmodel/algos/ensemble/StackedEnsembleMojoReader.java
tomasfryda: I bumped the minor version of StackedEnsemble MOJO both in the writer and in the reader. I assume the version checking is done by `hex.genmodel.ModelMojoReader#checkMaxSupportedMojoVersion` or am I missing something?

tomasfryda:
Thank you @michalkurka. I added the  `testJavaScoring` to a test (I test it in python as well) and I noticed we some issues with mojos - the tests pass but it complains about empty `model_summary`, `scoring_history` which is expected and then it complains about `base_models` that are stored as keys and they are not supported in `hex.genmodel.attributes.ModelJsonReader#convertJsonObject`. I noticed this has to do something with model details JSON which states that is experimental so I don't know if that's really an issue.

```
WARN water.default: Failed to extract element 'output.model_summary' MojoModel dump.
WARN water.default: Failed to extract element 'output.scoring_history' MojoModel dump.
ERROR water.default: Error reading MOJO JSON: Object not supported: 
ERROR water.default:  {"__meta":{"schema_version":3,"schema_name":"KeyV3","schema_type":"Key<Keyed>"},"name":"Grid_GBM__a7149e701913c2b26283c67edf3f4ba3_model_1616165183688_5642_model_1","type":"Key<Keyed>","URL":null} 
ERROR water.default: Error reading MOJO JSON: Object not supported: 
ERROR water.default:  {"__meta":{"schema_version":3,"schema_name":"KeyV3","schema_type":"Key<Keyed>"},"name":"Grid_GBM__a7149e701913c2b26283c67edf3f4ba3_model_1616165183688_5642_model_2","type":"Key<Keyed>","URL":null} 
```

michalkurka:
@tomasfryda I think it is unlikely related to your change, it is not a critical issue but it is not a nice thing to have either.

Ideally, the MOJO should load without errors and warnings, either it should be fixed or the parsing skipped if it cannot be made to work due to a specific nature of SE models.

Maybe put it somewhere on your list of TODOs. But not needed for this PR.

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
```diff
@@ -74,6 +77,79 @@ public void initActualParamValues() {
...
+            probs[i] = i/(double)N;
+          if (null == model._output._metalearner_percentile_rank_precomputed_quantiles) {
+            // Compute quantiles for specified probs
+            DKV.put(frame);
+            QuantileModel.QuantileParameters qp = new QuantileModel.QuantileParameters();
+            qp._train = frame._key;
+            qp._probs = probs;
+            QuantileModel qm = new Quantile(qp).trainModel().get();
+            model.write_lock();
+            model._output._metalearner_percentile_rank_precomputed_quantiles = qm._output._quantiles;
```
michalkurka, line 114: Is this transform only applied in training? Or in scoring a test set as well?

If it is not just training you would be overwriting the pre-determined quantiles.

michalkurka:
@tomasfryda I have a concern about PercentileRank transform - it seems to me it is overwritting the quantiles learned on the training frame, is that intended?

h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
michalkurka: not true ^^ please disregard

h2o-r/h2o-package/R/stackedensemble.R
ledell: @tomasfryda There are many sets of arg values in Python/R which differ between what's in the schema (we could also change the schema) and what we have in Python/R.  In the past few years, there's been a push in H2O to standardizing arg names on what's good for Python/R (which is lower/snake_case).  https://github.com/h2oai/h2o-3/blob/59b4e6fc81912dc3b067f50346aaa54f2c52faf1/h2o-algos/src/main/java/hex/schemas/StackedEnsembleV99.java#L106

The Python names are the snake case version, so right now it's unnecessarily inconsistent between Python and R.  Are you saying we can update the arg names in Python (after the client API files are generated from the schema), but not in R?: https://github.com/h2oai/h2o-3/blob/59b4e6fc81912dc3b067f50346aaa54f2c52faf1/h2o-py/h2o/estimators/stackedensemble.py#L561

I wanted to see if there's an existing convention for "None" but we have all three versions represented in various parts of the H2O algorithms: `"NONE", "none", "None"`. :-/

h2o-r/h2o-package/R/stackedensemble.R
```diff
@@ -215,6 +220,7 @@ h2o.stackedEnsemble <- function(x,
                                                 metalearner_fold_assignment = c("AUTO", "Random", "Modulo", "Stratified"),
                                                 metalearner_fold_column = NULL,
                                                 metalearner_params = NULL,
+                                                metalearner_transform = c("NONE", "Logit", "PercentileRank"),
```
ledell, line 223: See comments above.

h2o-r/h2o-package/R/stackedensemble.R
```diff
@@ -34,6 +34,8 @@
 #'        classification problems. Must be one of: "AUTO", "Random", "Modulo", "Stratified".
 #' @param metalearner_fold_column Column with cross-validation fold index assignment per observation for cross-validation of the metalearner.
 #' @param metalearner_params Parameters for metalearner algorithm
+#' @param metalearner_transform Transformation used for the level one frame. Must be one of: "NONE", "Logit", "PercentileRank". Defaults to
```
ledell, line 37: See below.

ledell (changes requested):

h2o-r/h2o-package/R/stackedensemble.R
tomasfryda: @ledell thank you for mentioning the target encoding. I found there a way to change the values to snake case. When I wrote that comment, I assumed I would have to modify `h2o-bindings/bin/gen_R.py` which would change the behavior for all the generated enums. In target encoding, I found out that I can change just the one parameter.

But I found out that [1] checks it against CamelCase names from the backend in case insensitive manner so I modified it to get rid of underscores and then match which seems to be compatible with what we do in the backend[2].

I added a test in R to make sure it works also I tested it but it is last minute change so if it seems uncomfortably complicated, I am not against of removing the commit and using it later in a fix release.

[1] https://github.com/h2oai/h2o-3/blob/9ff5e74b185a59a6be5cdbe36959695b326088c1/h2o-r/h2o-package/R/models.R#L337

[2] https://github.com/h2oai/h2o-3/blob/18f3f0f522ef3f00c3f9983624f3280af5124d88/h2o-core/src/main/java/water/util/EnumUtils.java#L29-L39

h2o-r/h2o-package/R/stackedensemble.R
michalkurka:
```
#' @param metalearner_fold_assignment Cross-validation fold assignment scheme for metalearner cross-validation.  Defaults to AUTO (which is
#'        currently set to Random). The 'Stratified' option will stratify the folds based on the response variable, for
#'        classification problems. Must be one of: "AUTO", "Random", "Modulo", "Stratified".
```

just few lines above, seems consistent with what we have

h2o-genmodel/src/main/java/hex/genmodel/algos/ensemble/StackedEnsembleMojoReader.java
michalkurka: you are right, nothing else needed

h2o-r/tests/testdir_algos/stackedensemble/runit_stackedensemble_metalearner_transform.R
```diff
@@ -0,0 +1,63 @@
...
+
+  stack_percentile_rank <- h2o.stackedEnsemble(y = y,
+                                               training_frame = train,
+                                               base_models = list(my_rf, my_xrf),
+                                               metalearner_transform = "percentile_rank",
+                                               seed = 1)
+
+  # Check that the metalearner transform was set properly
+  expect_equal(stack_logit@parameters$metalearner_transform, "Logit")
+  expect_equal(stack_percentile_rank@parameters$metalearner_transform, "PercentileRank")
```
michalkurka, line 47: @tomasfryda @ledell I think this is really strange; people shouldn't see values they cannot actually use - this will make hard re-training the model using the same parameters

sebhrusen: are you talking about the difference in the case?
This has always been a challenge to get both clients and backend aligned on this, already with the metrics...

sebhrusen:
```R
histogram_type = "Random",
nfolds = nfolds,
fold_assignment = "Modulo",
```
those are not snake case besides

michalkurka: > are you talking about the difference in the case?

yes

> already with the metrics

I don't think we should introduce another special case, especially not in this PR.

This can and should be addressed on a more system-level with all use cases taken into consideration.

tomasfryda: Ok, I agree that a more system-level approach would be better especially since we would make special logic for `metalearner_transform` in the bindings generation which basically enforces that it doesn't behave the same way as other parameters which would make it harder in future to unite the "coding style" as we would have to deal with this specific logic separately.

So I'll revert the commit. Should I keep the `NONE` or change it to `None`?

michalkurka: @sebhrusen if the snake case is an issue, can we just remove PercentileRank which is not really needed anyway?

This is getting complicated for a last-minute change.

Changes like that should be kept minimal, I think that is what we agreed to do.

tomasfryda: I kept the `NONE` for now as it is the most frequent variant on current `master` branch.

```bash
~/sources/h2o-3/h2o-r/h2o-package/R$ grep -l None *.R | wc -l 
       8

~/sources/h2o-3/h2o-r/h2o-package/R$ grep -l none *.R | wc -l
       9

~/sources/h2o-3/h2o-r/h2o-package/R$ grep -l NONE *.R | wc -l
      14
```

michalkurka:
@mn-mikke FIY - here is an API change, I assume SW will be fine with it but you should be aware of it

mn-mikke:
@michalkurka Thanks for letting me know! Yeah, SW seems to be unaffected since StackedEnsamble as such is not exposed in SW. The only connection to SE in SW are mojo models produced by AutoML, but `EasyPredictModelWrapper` used for scoring in SW  hasn't changed. So LGTM from SW point of view.

h2o-r/tests/testdir_algos/stackedensemble/runit_stackedensemble_metalearner_transform.R
sebhrusen: @michalkurka The backend already accepts almost anything from the client, including snake-case.
The problem here is the case that we chose to use when exposing backend values to a given client compared with the case we chose to accept from the user on that same client.
- `expect_equal(stack_logit@parameters$metalearner_transform, "Logit")` : here "Logit" is what the backed returns, using the enum name.
- so we should accept the same case on the client.

just in `stackedensemble.R` before this PR:
```R
metalearner_algorithm = c("AUTO", "deeplearning", "drf", "gbm", "glm", "naivebayes", "xgboost"),
metalearner_fold_assignment = c("AUTO", "Random", "Modulo", "Stratified"),
auc_type = c("AUTO", "NONE", "MACRO_OVR", "WEIGHTED_OVR", "MACRO_OVO", "WEIGHTED_OVO")
```
and in `automl.R`:
```R
stopping_metric = c("AUTO", "deviance", "logloss", "MSE", "RMSE", "MAE", "RMSLE", "AUC", "AUCPR", "lift_top_group", "misclassification", "mean_per_class_error"),
```
honestly, this is a complete mess.

I agree that this should be handled globally, but this is already the case on the backend (no pun intended).
This is also fine on Python, because we can just usually lowercase everything before validating.

It's mainly an R issue because of this `variable=c('default', 'choice1', 'choice2')` approach, and I don't see a way to fix this there outside one of:
1. get rid of those `c(…)` in the signature. and validate programmatically like in Py.
2. add all possible cases in those `c(…)` initializers.

Alternative: add a client param to each request, and depending on the client the REST API will format all enums use a preferred case. I can already say that I don't like this solution.

Also, afaik all recent R additions have been using snake case (not only for values but also for function names).

I don't plan to fix this here, just starting a discussion, otherwise we'll still mention this issue in the next PR.

ledell: In R, we can have a strict name choice but then also lazily accept a lower/different case version (or any other mapping) rather than using the strict `match.args()` check. Just as an option.

michalkurka (approved):
LGTM! @tomasfryda thank you for the effort

ledell:
@tomasfryda Everything looks good!  We can fix the case of the `"Logit"` later (when we do a package-wide update) and we can add percentile rank in as a follow-up PR for the fix release.

PR #5358, opened 2021-03-02
Michalk test slow

michalkurka:


michalkurka:
@maurever this test didn't show any issue since its creation and takes 1800s, given that it is part of the repeated pipeline - we can now decrease the number of GBMs to a much lower

valenad1:
I would suggest to rename the test file for example to runit_pubdev_3847_train_several_gbm_models_with_different_seeds_large.R

maurever (approved):
I agree with you @michalkurka. And I also agree with renaming the test, thanks @valenad1 for the suggestion.

michalkurka:
@valenad1 thanks for the suggestion, renamed to runit_pubdev_3847_gbm_random_attack_large.R (random attack is an established term for training models with randomized parameters)

PR #5361, opened 2021-03-05
PUBDEV-8021: Add support for refreshing HDFS delegation tokens

michalkurka:


michalkurka:
@mn-mikke still a draft but your comments are appreciated

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+public class HdfsDelegationTokenRefresher implements Runnable {
+
+    public static final String H2O_AUTH_USER = "h2o.auth.user";
+    public static final String H2O_AUTH_PRINCIPAL = "h2o.auth.principal";
+    public static final String H2O_AUTH_KEYTAB = "h2o.auth.keytab";
+    public static final String H2O_AUTH_TOKEN_REFRESHER_ENABLED = "h2o.auth.tokenRefresher.enabled";
+    public static final String H2O_AUTH_TOKEN_REFRESHER_INTERVAL_RATIO = "h2o.auth.tokenRefresher.intervalRatio";
```
mn-mikke, line 31: Is this going to come from some H2O driver parameter? I can't see the origin of the value right now.

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
```diff
@@ -0,0 +1,215 @@
...
+            return;
+        }
+        final long actualInterval;
+        if (interval == Long.MAX_VALUE) {
+            log("Couldn't determine tokens renewal interval, will use 12h", null);
+            actualInterval = 12 * 3600 * 1000; // 12h
+        } else
+            actualInterval = (long) (interval * _intervalRatio);
+        log("Determined token renewal interval = " + interval + "ms. Using actual interval = " + actualInterval + "ms.", null);
+        _executor.scheduleAtFixedRate(this, 0, interval, TimeUnit.MILLISECONDS);
```
mn-mikke, line 95: nit: It seems to me that milliseconds don't bring much benefits in this logic. Maybe seconds could play the same role and not to generate so big numbers in logs.

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
```diff
@@ -0,0 +1,215 @@
...
+                    .stream()
+                    .map(token -> {
+                        try {
+                            long expiresAt = token.renew(PersistHdfs.CONF);
+                            long issuedAt = 0;
+                            TokenIdentifier ident = token.decodeIdentifier();
+                            if (ident instanceof AbstractDelegationTokenIdentifier) {
+                                issuedAt = ((AbstractDelegationTokenIdentifier) ident).getIssueDate();
+                            }
+                            return expiresAt - (issuedAt > 0 ? issuedAt : System.currentTimeMillis());
```
mn-mikke, line 158: What about a presence of one expired token where it's identifier is not `AbstractDelegationTokenIdentifier`. In that case, the method would return a negative number. Can it happen, is it a problem?

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
```diff
@@ -0,0 +1,215 @@
...
+        } catch (IOException | InterruptedException e) {
+            log("Failed to refresh token.", e);
+        }
+    }
+
+    private Credentials refreshTokens(UserGroupInformation tokenUser) throws IOException, InterruptedException {
+        return tokenUser.doAs((PrivilegedExceptionAction<Credentials>) () -> {
+            Credentials creds = new Credentials();
+            try {
+                Token<?>[] tokens = fetchDelegationTokens(_authUser, creds);
```
mn-mikke, line 124: qq: where is the renewing done? Does it still needs to b implemented?

h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
```diff
@@ -1272,11 +1275,14 @@ else if (s.equals("-proxy")) {
       } else if (s.equals("-hivePrincipal")) {
         i++; if (i >= args.length) { usage (); }
         hivePrincipal = args[i];
-      } else if (s.equals("-refreshTokens")) {
-        refreshTokens = true;
+      } else if (s.equals("-refreshTokens") || // for backwards compatibility
```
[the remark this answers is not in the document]
michalkurka: We cannot really deprecate it because Steam needs to be able to support all H2O versions. This means Steam will likely keep using "-refreshTokens" with semantics "refresh hive tokens" (and we shouldn't generate a warning for our own software).

Could change in the future, for now I think leave as-is.

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
```diff
@@ -0,0 +1,215 @@
...
+            log("Couldn't determine tokens renewal interval, will use 12h", null);
+            actualInterval = 12 * 3600 * 1000; // 12h
+        } else
+            actualInterval = (long) (interval * _intervalRatio);
+        log("Determined token renewal interval = " + interval + "ms. Using actual interval = " + actualInterval + "ms.", null);
+        _executor.scheduleAtFixedRate(this, 0, interval, TimeUnit.MILLISECONDS);
+    }
+
+    private static void log(String s, Exception e) {
+        System.out.println("TOKEN REFRESH: " + s);
```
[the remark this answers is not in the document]
michalkurka: This is the same as how we log Hive token refresh, the issue is that this is being done before any logging is configured. This also has the interesting side effect that it doesn't appear in H2O logs - which is a good thing (in this case), it is only accessible to admins who are able to download the yarn logs.

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
michalkurka: This is never populated, h2odriver has a generic mechanism of passing properties and adding them to the configuration. This way user can change the property without the driver really doing anything for it.

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
michalkurka: good point! will look into it

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
michalkurka: The logic works this way: 
- we get a token and renew it (line 153) - which means we are only dealing with a valid token
- if the token isn't an instance of AbstractDelegationTokenIdentifier we will consider the current time as the time of renewal

the renewal intervals are long (hours) so we should never have a negative number here - however that case can be checked as well, will improve

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
michalkurka: this call will always retrieve a fresh set of tokens, regardless if there are existing tokens around because I am passing an empty Credentials object

if I used the actual users Credentials I could get tokens that are about to expire

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
michalkurka: fixed, thank you for the suggestion

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
michalkurka: (it is great you catch these things though!)

michalkurka:
@mn-mikke Please take another look, I added a bunch of tests and incorporated your feedback.

h2o-mapreduce-generic/src/main/java/water/hadoop/h2odriver.java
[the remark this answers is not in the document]
mn-mikke: ok, np

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
```diff
@@ -0,0 +1,230 @@
...
+    public static void setup(Configuration conf, String tmpDir) throws IOException {
+        boolean enabled = conf.getBoolean(H2O_AUTH_TOKEN_REFRESHER_ENABLED, false);
+        if (!enabled) {
+            log("HDFS Token renewal is not enabled in configuration", null);
+            return;
+        }
+        String authUser = conf.get(H2O_AUTH_USER);
+        String authPrincipal = conf.get(H2O_AUTH_PRINCIPAL);
+        String authKeytab = conf.get(H2O_AUTH_KEYTAB);
+        if (authKeytab == null) {
```
mn-mikke, line 45: Shouldn't be here also a similar condition for `authPrincipal` and eventually for `authUser`?

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
```diff
@@ -0,0 +1,230 @@
...
+            Credentials creds = new Credentials();
+            Token<?>[] tokens = fetchDelegationTokens(_authUser, creds);
+            log("Fetched delegation tokens: " + Arrays.toString(tokens), null);
+            return creds;
+        });
+    }
+
+    private UserGroupInformation loginAuthUser() throws IOException, InterruptedException{
+        log("Log in from keytab as " + _authPrincipal, null);
+        UserGroupInformation realUser = UserGroupInformation.loginUserFromKeytabAndReturnUGI(_authPrincipal, _authKeytabPath);
```
mn-mikke, line 148: nit: the `realUser` variable seems to be redundant.

mn-mikke (approved):
Added two comments, otherwise LGTM

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
michalkurka: good point, will fix (authUser can be undefined, in that case we use authPrincipal)

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
michalkurka: yes, intellij agrees with you - however it is easier to debug it then (if I keep both variables)

h2o-persist-hdfs/src/main/java/water/persist/security/HdfsDelegationTokenRefresher.java
michalkurka: fixed, thx

PR #5363, opened 2021-03-09
PUBDEV-8038: Expose compact way of representing contributions in Python

michalkurka:


h2o-py/h2o/model/model_base.py
```diff
@@ -186,12 +186,17 @@ def predict_contributions(self, test_data):
         Note: Multinomial classification models are currently not supported.
 
         :param H2OFrame test_data: Data on which to calculate contributions.
+        :param Enum output_format: Specify how to output feature contributions in XGBoost - XGBoost by default outputs 
+            contributions for 1-hot encoded features, specifying a Compact output format will produce a per-feature
+            contribution. One of: ``"Original"``, ``"Compact"`` (default: ``"Compact"``).
```
g-eoj, line 191: Says default is compact but in code it's original.

g-eoj (approved):
Docstring issue, other than that LGTM. Thanks!

h2o-py/h2o/model/model_base.py
michalkurka: Thank you for catching this. Fixed.

maurever (approved):
Looks good to me. Thanks, @michalkurka!

PR #5364, opened 2021-03-09
Turn off (single node) multigpu support in XGBoost

michalkurka:


neemah2o:
Thanks for the info @michalkurka . I don't have an update from the requestor, but will let you know if things change.

PR #5365, opened 2021-03-09
PUBDEV-8034: Job polling will retry if connection fails

michalkurka:


michalkurka:
Note: this also needs to be ported to rel-zermelo, however, using master as a target is easier for the review

satai:
LGTM if the test is run in isolation

michalkurka:
>LGTM if the test is run in isolation

@satai yes, on the client-side the tests run in isolation - a new process is spawned for each test (file), see https://github.com/h2oai/h2o-3/blob/master/scripts/run.py#L775

michalkurka:
thanks for the review

PR #5368, opened 2021-03-11
XGBoost: Get POJO predictions (even) closer to in-H2O predictions

michalkurka:


wendycwong (approved):
Thank you for fixing it.  You are now using 1e-7 for the Junit tests too.  Everything is now consistent.

PR #5369, opened 2021-03-12
Fix pyunit_glrm_grid_user_y.py

wendycwong:
These compares are failing on some jenkin machines:

    assert grid.models[1].archetypes() != grid.models[2].archetypes()
    assert grid.models[2].archetypes() != grid.models[3].archetypes()

Note that grid.models[1].archetypes() is an array of arrays.  Depending on the Python version on these machines, they may or may not be able to handle this.

Hence, I change it to using our own compare to just compare the first two arrays like this:

    assert not(pyunit_utils.equal_two_arrays(grid.models[1].archetypes()[0], grid.models[2].archetypes()[0], throw_error=False))
    assert not(pyunit_utils.equal_two_arrays(grid.models[2].archetypes()[0], grid.models[3].archetypes()[0], throw_error=False))

The logic is if the first two arrays are not equal, then, the two arrays of arrays will not be equal.


maurever (approved):
Perfect! Thank you, @wendycwong! Looks good to me.

PR #5370, opened 2021-03-12
[Docs] For include / exclude algos options, make it clear that algorithms in list are usable

zainhaq-h2o:
This PR edits the documentation for the include / exclude algos options.

See https://h2oai.slack.com/archives/C04KNHH2H/p1615584988003500 for more info.

chrome_LlYurvV7o9


ledell (approved):
Perfect, thanks!

PR #5372, opened 2021-03-15
PUBDEV-8046: Fix matplotlib import error, unintentional switch to agg, figure cropping

tomasfryda:
Fixes issues mentioned on internal channel regarding usage of the python's `explanation` module and in https://h2oai.atlassian.net/browse/PUBDEV-8046 as the main part regarding `matplotlib.backend` largely overlapped.

Namely:
* Unclear exception when `matplotlib` is missing
* CLI usage problems - Varimp plot and later plots were rendered by "Agg" backend instead of the one that was set before the  `.explain` was called
* Cropping of the legend in some backends (not present in Jupyter)

GUI usage tested on Jupyter and Spyder.
CLI usage tested with `MacOSX` (macOS BigSur) and `TkAgg` (Ubuntu Focal) `matplotlib` backends.

### DIFF
--- a/h2o-py/h2o/explanation/__init__.py
+++ b/h2o-py/h2o/explanation/__init__.py
@@ -1,15 +1,44 @@
 # -*- encoding: utf-8 -*-
 
-from ._explain import varimp_heatmap, model_correlation_heatmap, shap_explain_row_plot, shap_summary_plot,\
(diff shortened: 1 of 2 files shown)


ledell (approved):
Thank you!

michalkurka (approved):

PR #5377, opened 2021-03-16
Weaver clean-up

michalkurka:
This removes dead code and cleans up the code to make IntelliJ happy


h2o-core/src/main/java/water/Weaver.java
```diff
@@ -36,166 +32,49 @@
...
-   * {@link ClassLoader} instances pruned.
-   *
-   * Finally, in order to {@link Weaver#genDelegate(int, Class)} during a class reload, then the
-   * previous {@link Icer} must be {@link CtClass#detach}ed. In addition, the {@link TypeMap#goForGold(int)}
-   * must turn a null for the {@link Icer}.
-   *
-   * @param name class name
-   * @param b bytecode
-   */
-  public static void loadDynamic(final String name, final byte[] b) {
```
sebhrusen, line 150: unused? What was the use-case for this? Plugins/extensions? Custom metrics? The unit test doesn't help…

sebhrusen (approved):
Looks good: just want to  be sure that the `loadDynamic` method is not used by 3rd parties.

h2o-core/src/main/java/water/Weaver.java
michalkurka: This basically lets you modify Iced classes at runtime and re-generate Icers.

To be honest I don't know if this was ever used and how but it definitely is not used for extension points and by 3rd parties. Custom extensions are put on classpath before H2O boots up and classloading happens the regular way.

To my knowledge very safe to remove.

PR #5382, opened 2021-03-17
Relax XGBoost tolerance on GPU when comparing java and native predictions

michalkurka:
As demonstrated in https://github.com/h2oai/h2o-3/pull/5373/files sigmoid transformation
implemented on GPU gives different result than when running on CPU, therefore we pick a lower tolerance

This doesn't affect production results because native xgboost scoring
is only used in test. For all other scoring java scoring is used.

### DIFF
--- a/h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostTest.java
+++ b/h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostTest.java
@@ -498,7 +498,7 @@ public class XGBoostTest extends TestUtil {
       assertEquals(metadataBefore, metadataAfter);
 
       Frame preds = Scope.track(model.score(testFrame));
-      assertTrue(model.testJavaScoring(testFrame, preds, 1e-7));
+      assertJavaScoring(model, testFrame, preds);
       assertEquals(
               ModelMetricsBinomial.make(preds.vec(2), testFrame.vec(response)).auc(),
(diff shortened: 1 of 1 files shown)


h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostTest.java
```diff
@@ -2420,4 +2420,29 @@ private void assertHasValidationError(hex.tree.xgboost.XGBoost builder, String f
     assertTrue(field, builder.validationErrors().contains(error));
   }
 
+  private void assertJavaScoring(XGBoostModel model, Frame testFrame, Frame preds) {
+    double rel_epsilon = Boolean.parseBoolean(confJavaPredict) ?
```
maurever, line 2424:
```suggestion
    double relEpsilon = Boolean.parseBoolean(confJavaPredict) ? 
```

maurever (approved):
Just one nitpicking comment; otherwise looks good. Thank you, you found this issue @michalkurka!

h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostTest.java
michalkurka: applied, thank you

michalkurka:
thank you for checking @maurever

PR #5383, opened 2021-03-17
PUBDEV-8054: Add Java 15 support in R package DESCRIPTION file

ledell:
https://h2oai.atlassian.net/browse/PUBDEV-8054

- Now that Java 15 support is merged, we can update the system requirements in the R DESCRIPTION file.
- Added a drive-by commit to fix the BugReports URL as well.

### DIFF
--- a/h2o-r/h2o-DESCRIPTION.template
+++ b/h2o-r/h2o-DESCRIPTION.template
@@ -41,7 +41,7 @@ License: Apache License (== 2.0)
 URL: https://github.com/h2oai/h2o-3
 BugReports: https://0xdata.atlassian.net/projects/PUBDEV
 NeedsCompilation: no
-SystemRequirements: Java (>= 8, < 15)
+SystemRequirements: Java (>= 8, < 16)
 Depends: R (>= 2.13.0),
          methods,
          stats
--- a/h2o-r/h2o-DESCRIPTION.template
+++ b/h2o-r/h2o-DESCRIPTION.template
@@ -39,7 +39,7 @@ Description: R interface for 'H2O', the scalable open source machine learning
     Word2Vec, as well as a fully automatic machine learning algorithm (H2O AutoML).
 License: Apache License (== 2.0)
 URL: https://github.com/h2oai/h2o-3
-BugReports: https://0xdata.atlassian.net/projects/PUBDEV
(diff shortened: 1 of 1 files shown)


h2o-r/h2o-DESCRIPTION.template
```diff
@@ -39,9 +39,9 @@ Description: R interface for 'H2O', the scalable open source machine learning
     Word2Vec, as well as a fully automatic machine learning algorithm (H2O AutoML).
 License: Apache License (== 2.0)
 URL: https://github.com/h2oai/h2o-3
-BugReports: https://0xdata.atlassian.net/projects/PUBDEV
+BugReports: https://h2oai.atlassian.net/projects/PUBDEV
 NeedsCompilation: no
-SystemRequirements: Java (>= 8, < 15)
+SystemRequirements: Java (>= 8, < 16)
```
michalkurka, line 44: What is the motivation for adding this field? AFAIK this is not a standardized field and has thus no real-life effect.

Is this something we need to do for CRAN?

It seems like another place where we need to keep updating.

tomasfryda: I don't know much about the DESCRIPTION file so I googled and found this https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Writing-portable-packages

> Make sure that any version requirement for Java code is both declared in the ‘SystemRequirements’ field and tested at runtime (not least as the Java installation when the package is installed might not be the same as when the package is run and will not be for binary packages).

So I grepped and we might want to change this as well: 
https://github.com/h2oai/h2o-3/blob/74a03c415e680374e9466db42528f65a9f8c849d/h2o-r/h2o-package/R/connection.R#L580-L592

michalkurka: We do the check for the "certified" latest supported java version in the java code itself. We don't need to check it in R because are able to run the actual java code. On Java 7 and earlier we are not able to do so.

ledell: @michalkurka Yes, this is a requirement for CRAN and they've complained when it was not correct in the past, so we need to keep it up to date.

michalkurka: thank you for clarifying

michalkurka (approved):

PR #5385, opened 2021-03-18
PUBDEV-8038: Allow to use compact output format for contributions in R

michalkurka:


h2o-r/h2o-package/R/models.R
```diff
@@ -838,6 +838,9 @@ h2o.staged_predict_proba <- staged_predict_proba.H2OModel
 #'        desired
 #' @param newdata An H2OFrame object in which to look for
 #'        variables with which to predict.
+#' @param output_format Specify how to output feature contributions in XGBoost - XGBoost by default outputs
+#'                      contributions for 1-hot encoded features, specifying a 'Compact' output format will produce
+#'                      a per-feature contribution. One of: 'Original', 'Compact' (default: 'Original').
```
tomasfryda, line 843: Defaults in R are usually specified as sentence, e.g., `Defaults to Original.`  and the "One of" part is redundant since this is already present in the function declaration.

h2o-r/h2o-package/R/models.R
```diff
@@ -853,12 +856,19 @@ h2o.staged_predict_proba <- staged_predict_proba.H2OModel
 #' h2o.predict_contributions(prostate_gbm, prostate)
 #' }
 #' @export
-predict_contributions.H2OModel <- function(object, newdata, ...) {
+predict_contributions.H2OModel <- function(object, newdata, output_format = c("Original", "Compact"), ...) {
```
tomasfryda, line 859: I don't have strong opinion about this but in explain, we are trying to have lowercase enums (inspired by tidyverse[1]). We should probably have some coding style guide to keep it in sync.

[1] https://style.tidyverse.org

h2o-r/h2o-package/R/models.R
```diff
@@ -853,12 +856,19 @@ h2o.staged_predict_proba <- staged_predict_proba.H2OModel
...
     if (missing(newdata)) {
         stop("predictions with a missing `newdata` argument is not implemented yet")
     }
+    params <- list(predict_contributions = TRUE)
+    if (!missing(output_format)) {
+        if (!(output_format %in% c("Original", "Compact"))) {
+          stop("type must be one of: Original, Compact")
+        }
+        params$predict_contributions_output_format <- output_format
```
tomasfryda, line 868: Wouldn't it be better to use:
```r
params$predict_contributions_output_format <- match.arg(output_format)
```

tomasfryda: or something like `match.arg(arg=tolower(output_format), choices = eval(formals()$output_format))` to keep it case insensitive (assuming the vector in declaration is in lowercase).

h2o-r/h2o-package/R/models.R
michalkurka: Make sense, I followed the example in leaf node assignment

```
   predict_leaf_node_assignment.H2OModel <- function(object, newdata, type = c("Path", "Node_ID"), ...) {
```
we will have to change it there as well

h2o-r/h2o-package/R/models.R
michalkurka: good idea, thanks - I don't think we need to worry about case insensitivity - other places we just use match.arg

h2o-r/h2o-package/R/models.R
michalkurka: Changed, thx

h2o-r/h2o-package/R/models.R
michalkurka: fixed, I made a jira for predict_leaf_node_assignment https://h2oai.atlassian.net/browse/PUBDEV-8055

tomasfryda (approved):
LGTM, thank you @michalkurka!

michalkurka:
@tomasfryda thank you for catching the API naming inconsistencies and your other suggestions

PR #5389, opened 2021-03-20
PUBDEV-7996: single_quote parse setup result param not forwarded to parser

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-7996

### DIFF
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -482,8 +482,9 @@ class H2OFrame(Keyed):
              "blocking": False,
              "column_types": None,
              "skipped_columns":None,
-             "custom_non_data_line_markers": setup["custom_non_data_line_markers"],
-             "partition_by": setup["partition_by"]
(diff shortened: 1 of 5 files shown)


michalkurka (approved):
LGTM - thank you!

Does this mean it never worked in R? Can you please also add a test?

sebhrusen:
@michalkurka It never worked both in Py and in R.

Yes, will write a test for each before merging.

sebhrusen:
@michalkurka tested and ready to merge.
However, the fix for the R test runner discovered 3 bad R tests:
- runit_PUBDEV_7798_overlapped_knots.R. This is a recent test, should have never passed (data -> as.h2o(data)).
- runit_PUBDEV_7860_gam_tp_gaussian_dual_model.R. Also a recent test, seems to fail only once in a while.
- r_suite.runit_lift_xval.R. Old test, but was "passing" in spite of errors with old runner.

PR #5390, opened 2021-03-21
AutoML using fewer neurons in DNN grids

sebhrusen:


ledell:
This PR was extended and merged here: https://github.com/h2oai/h2o-3/pull/5401
So we can close this one.

PR #5392, opened 2021-03-22
AutoML: Fewer neurons in DL grids and a logit transform for classification models

ledell:
https://h2oai.atlassian.net/browse/PUBDEV-8070


ledell:
This was extended here: https://github.com/h2oai/h2o-3/pull/5392

PR #5393, opened 2021-03-22
Remove public final modifier from GLM mojo score0 function

km2018:
This PR fixes a bug found when trying to auto-generate files for GLM in mojoland. When baking new recipes, mojoland sends a GET request for the function `double[] score0(double[] data, double[] preds)`. GLM currently has a `public final` modifier on this function, causing an exception to be thrown when baking new recipes for GLM in mojoland. By removing the `final` modifier from the function, the PR solves the bug.

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/algos/glm/GlmMojoModelBase.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/algos/glm/GlmMojoModelBase.java
@@ -25,7 +25,7 @@ abstract class GlmMojoModelBase extends MojoModel {
   void init() { /* do nothing by default */ }
 
   @Override
-  public final double[] score0(double[] data, double[] preds) {
+  public double[] score0(double[] data, double[] preds) {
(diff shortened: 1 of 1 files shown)


michalkurka:
@km2018 can this be fixed in mojoland instead?

koniecsveta:
@km2018 @michalkurka `score0` is inaccessbile before invocation in MojoApiHandler.java. Issue disappears when it is set to accessible before invocation like this: 
`    public String invoke(Object target, String[] strArgs) throws Exception {
      **method.setAccessible(true);**
      Object retVal = method.invoke(target, parseArgs(strArgs));
      return ret.toString(retVal);
    }
` 
in MojoApi.java  but not sure whether this would be the right fix. Maybe the better question is why it is not accessible when it is `public final` and for example in xgb case it works even with ` public final score0`? WDYT?

michalkurka:
it is probably because it is an abstract class, it needs to be fixed in mojoland to allow for this kind of hierarchy

PR #5395, opened 2021-03-23
Fix runit_lift_xval.R test

michalkurka:
We changed the way cv folds are assigned in https://github.com/h2oai/h2o-3/pull/5345

This affects fold specification that doesn't start from 0, eg. 1, 2, 3

Old way was to use modulo operator resulting in effective fold assignment:
1, 2, 0

New way is to respect the fold order, resulting in fold assignment 0, 1, 2


sebhrusen (approved):
nice one!

PR #5399, opened 2021-03-23
Fixed runit_PUBDEV_7860_gam_tp_gaussian_dual_model.R

wendycwong:
This comparison failed in the test:

    expect_true(abs(mse1-mse2) < 1e-6)

However, mse1, mse2 are around values: 1329035.  Hence, if they only differ by 1 say 1329036, this test will fail.  But differing by 1 is a very small difference:  1/1329035 = 0.0000007524256 < 1e-6.  Hence, I made the change in the comparison as: expect_true(abs(mse1-mse2)/abs(mse1) < 1e-6)


sebhrusen (approved):
relative comparison sounds reasonable, thanks @wendycwong

michalkurka (approved):

PR #5402, opened 2021-03-24
Add metalearner transform documentation

tomasfryda:
Documentation for metalearner transform.

### DIFF
--- a/h2o-docs/src/product/data-science/algo-params/metalearner_transform.rst
+++ b/h2o-docs/src/product/data-science/algo-params/metalearner_transform.rst
@@ -7,9 +7,9 @@
 Description
 ~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


h2o-docs/src/product/data-science/algo-params/metalearner_transform.rst
```diff
@@ -0,0 +1,139 @@
+``metalearner_transform``
+-------------------------
+
+- Available in: Stacked Ensembles
+- Hyperparameter: no
```
tomasfryda, line 5: I'm not really sure if it should be a hyper param or not. Metalearner algorithm is not so I used "no" as well.

hannah-tillman (approved):
super minor updates--builds well and looks good to me :)

ledell (approved):
This is great, thanks!

PR #5408, opened 2021-03-26
Fix Stacked Ensemble with Multinomial GLM

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8076

Fix one of the issues discovered by setting `_alpha=[0.5, 1.0]` in the AUTOMetalearner.


h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
```diff
@@ -1529,4 +1529,137 @@ public void testMetalearnerTransformWorks() {
...
+
+
+
+        } finally {
+            if (training_frame != null) training_frame.remove();
+            if (gbm != null) {
+                gbm.delete();
+                for (Key k : gbm._output._cross_validation_predictions) k.remove();
+                gbm._output._cross_validation_holdout_predictions_frame_id.remove();
+                gbm.deleteCrossValidationModels();
```
sebhrusen, line 1645: Looks like copy/paste from old code, this should be enough now:
```suggestion
                gbm.delete();
```
as it cascades deletion by default.

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
```diff
@@ -1529,4 +1529,137 @@ public void testMetalearnerTransformWorks() {
...
+            if (preds != null) preds.delete();
+
+            Set<Frame> framesAfter = new HashSet<>(framesBefore);
+            framesAfter.removeAll(Arrays.asList(Frame.fetchAll()));
+
+            Assert.assertEquals("finish with the same number of Frames as we started: " + framesAfter, 0, framesAfter.size());
+
+            if (stackedEnsembleModel != null) {
+                stackedEnsembleModel.delete();
+                stackedEnsembleModel._output._metalearner.delete();
```
sebhrusen, line 1660: same as above

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
```diff
@@ -1529,4 +1529,137 @@ public void testMetalearnerTransformWorks() {
...
+            if (drf != null) {
+                drf.delete();
+            }
+
+            if (preds != null) preds.delete();
+
+            Set<Frame> framesAfter = new HashSet<>(framesBefore);
+            framesAfter.removeAll(Arrays.asList(Frame.fetchAll()));
+
+            Assert.assertEquals("finish with the same number of Frames as we started: " + framesAfter, 0, framesAfter.size());
```
sebhrusen, line 1656: is it part of this test? specific to this use case? Otherwise it looks like a smuggled test.

sebhrusen (approved):
Looks good for me.
@wendycwong can you please confirm and have a quick look at the GLM changes? Thanks.

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
tomasfryda: Thank you @sebhrusen! I didn't realize that it cascades, I copied it from the old code without much thought.

h2o-algos/src/test/java/hex/ensemble/StackedEnsembleTest.java
tomasfryda: smuggled test, I removed it now. Thanks!

wendycwong (approved):
LGTM

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1913,19 +1913,30 @@ public GLMMojoWriter getMojo() {
...
+    return beta()[featureIdx + offset] != 0;
+}
 
+  @Override
+  protected boolean isFeatureUsedInPredict(int featureIdx) {
+    if (_output.isMultinomialClassifier()) {
+      int outputClasses =  beta().length / _output.coefficientNames().length;
+      boolean used = false;
+      for (int i = 0; i < outputClasses; i++)
+        used |= isFeatureUsedInPredict(featureIdx, i * _output.coefficientNames().length);
```
michalkurka, line 1936: you can stop when you find feature is not used

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1913,19 +1913,30 @@ public GLMMojoWriter getMojo() {
...
       return false;
     } else {
       featureIdx += _output._dinfo._numOffsets[0] - _output._dinfo._catOffsets.length + 1;
     }
-    return beta()[featureIdx] != 0;
+    return beta()[featureIdx + offset] != 0;
+}
 
+  @Override
+  protected boolean isFeatureUsedInPredict(int featureIdx) {
```
michalkurka, line 1931: is there an easier way of implementing this?

I don't know... for example calling dataInfo.innerProduct on an instance of a dummy row where only feature X is "hot" and the beta coefs?

my concern is that is the is might not be always correct or it might become obsolete, eg. what about interactions - the coefficient for feature could be zero but it might still be used in interactions

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: Good call @michalkurka, he can stop once he finds one coefficient that is not zero!

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1913,19 +1913,30 @@ public GLMMojoWriter getMojo() {
...
+    if (_output.isMultinomialClassifier()) {
+      int outputClasses =  beta().length / _output.coefficientNames().length;
```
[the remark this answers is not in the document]
wendycwong: Yes, beta() includes the intercepts.  Therefore, he cannot use the nclass.

h2o-algos/src/main/java/hex/glm/GLMModel.java
tomasfryda: Good point, thanks!

h2o-algos/src/main/java/hex/glm/GLMModel.java
tomasfryda: @michalkurka I understand the concern and I tried to find some other simpler way but I didn't find one. 

I could use the `innerProduct` but since the problem is the transformation from the "feature space" to $R^n$ the way it is written I think I would have to create an array of `Chunk`s (the dummy row) and use specific values to encode categorical variables etc and then use `extract(Dense|Sparse)Row` to do the transformation (so that I don't replicate the logic). 

I also looked at how `score0` works and it seems to work similarly to what I wrote there.

Good point about the interactions - I wasn't able to find a good solution so for now I check if the interactions are set and if they are I will just say all features are used for prediction - better a bit slower than incorrect solution. Note that this scenario is very unlikely since this method is used only in SE and AFAIK to use interactions the user would have to know what names are used in level one frame.

Did I miss some easier solution? Maybe @wendycwong would know?

PR #5409, opened 2021-03-26
PUBDEV-8077 fix autoML error when an alpha array is used for GLM

wendycwong:
This PR fixed the problem in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8077

The bug is in the following if condition:
        if (!_parms._lambda_search || (_parms._generate_scoring_history) && !_insideCVCheck))
          updateProgress(true);

It should be:
        if ((!_parms._lambda_search || (_parms._generate_scoring_history)) && !_insideCVCheck)
          updateProgress(true);

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/Metalearners.java
+++ b/h2o-algos/src/main/java/hex/ensemble/Metalearners.java
@@ -1,6 +1,5 @@
 package hex.ensemble;
 
-import hex.DistributionFactory;
 import hex.Model;
 import hex.ModelBuilder;
 import hex.ensemble.Metalearner.Algorithm;
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -2682,7 +2682,7 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
           _state.updateState(beta, gginfo._likelihood);
         else
           _state.updateState(beta, gginfo);
-        if (!_parms._lambda_search || (_parms._generate_scoring_history && !_insideCVCheck))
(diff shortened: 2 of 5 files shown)


h2o-py/tests/testdir_algos/automl/pyunit_PUBDEV_8077_automl_args_tomaf.py
```diff
@@ -0,0 +1,33 @@
...
+from tests import pyunit_utils
+from h2o.automl import H2OAutoML
+
+"""
+This test is from Tomas Fryda.  He discovered that if GLM alpha array is specified to be an array, the following test
+will crash.  I fixed the bug and this test should run to completion.  No assert statement is needed for this test.
+"""
+max_models = 2
+
+def automl_tomf():
```
tomasfryda, line 14: This test is dependent on a setting in stacked ensemble that isn't the default yet and so this test would be useless if the `_alpha` is not set in `SE`.

Normally, I would ask you if you could make a test that doesn't depend on AutoML/SE but since the fix is so simple, I think we don't need to add a test for this (although it would be nice to have bigger test coverage for all the algos).

tomasfryda (changes requested):
Thank you @wendycwong! The fix makes sense to me and I think it is a correct one. There are couple things that I would change/remove so we don't modify/depend on the behavior of SE. Please see my comments.

h2o-py/tests/testdir_algos/automl/pyunit_PUBDEV_8077_automl_args_tomaf.py
wendycwong: @tomasfryda I just added 3 tests to test that alpha array works when generate_scoring_history is on or off.

tomasfryda (approved):
Looks good to me @wendycwong. Thank you!

PR #5411, opened 2021-03-29
PUBDEV-8023: Add support for importing encrypted zip files containing CSVs

michalkurka:


h2o-core/src/main/java/water/rapids/ast/prims/internal/AstRunTool.java
```diff
@@ -0,0 +1,44 @@
...
+    private static final String TOOLS_PACKAGE = "water.tools.";
+
+    @Override
+    public String[] args() {
+        return new String[]{"tool_class", "tool_parameters"};
+    }
+
+    @Override
+    public int nargs() {
+        return 1 + 2;
```
satai, line 21: run_tool =1 ;tool_class =1; tool_parameters = 1;
return run_tool + tool_class + tool_parameters;
;-)

or ?
run_tool =1;
return run_tool + args().length

h2o-r/tests/testdir_parser/runit_parse_encrypted_zip.R
```diff
@@ -0,0 +1,30 @@
...
+
+  keystore_path <- locate("smalldata/extdata/keystore.jks")
+  keystore <- h2o.importFile(keystore_path, parse = FALSE)
+
+  decrypt_tool <- h2o.decryptionSetup(keystore, key_alias = "secretKeyAlias",
+                                      password = "Password123", cipher = "AES/ECB/PKCS5Padding")
+
+  args <- c(keystore_path, "JCEKS", "secretKeyAlias", "Password123", "AES/ECB/PKCS5Padding", 
+            crimes_path_zip, crimes_path_zip_aes)
+  tool_result <- h2o.rapids(sprintf('(run_tool "EncryptionTool" ["%s"])', paste(args, collapse = '", "')))
```
mn-mikke, line 18: What is the purpose of EncryptionTool in this test? it doesn't seem to be involved in `h2o.importFile(crimes_path_zip_aes, decrypt_tool = decrypt_tool)`

h2o-core/src/main/java/water/parser/ParseDataset.java
```diff
@@ -814,16 +814,16 @@ private FVecParseWriter makeDout(ParseSetup localSetup, int chunkOff, int nchunk
           localSetup = ParserService.INSTANCE.getByInfo(localSetup._parse_type).setupLocal(vec,localSetup);
           // Zipped file; no parallel decompression;
           InputStream bvs = vec.openStream(_jobKey);
-          ZipInputStream zis = new ZipInputStream(bvs);
+          InputStream dec = decryptionTool.decryptInputStream(bvs);
```
mn-mikke, line 817: Just want to make sure if I understand it correctly: if encryption is disabled, `decryptionTool` is an instance of `NullDecryptionTool`?

michalkurka: @mn-mikke yes, there is always an instance of the "decryption tool"

h2o-r/tests/testdir_parser/runit_parse_encrypted_zip.R
michalkurka: This is a way of encrypting the zip file on the fly when the test runs. Basically, I added a simple tool to encrypt files and I am invoking it through a rapid expression. That way the process is self-documenting as opposed to storing an encrypted file on S3 in smalldata.

h2o-core/src/main/java/water/parser/ParseDataset.java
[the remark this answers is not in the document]
michalkurka: good point, they are only closed on a successful competition 

to be fair - closing is really a NO-OP in this scenario, because no resources are really freed by invoking close but it should be cleaned-up, given that it was already there - is it okay to address that in a new PR? @mn-mikke ?

h2o-core/src/main/java/water/rapids/ast/prims/internal/AstRunTool.java
michalkurka: drop in a bucket :) it is just how the rapids are done, IMHO keeping it consistent is better, wdyt?

h2o-core/src/main/java/water/parser/ParseDataset.java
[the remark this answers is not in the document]
mn-mikke: Ok, np. Let's keep refactoring changes separated.

h2o-r/tests/testdir_parser/runit_parse_encrypted_zip.R
mn-mikke: ok, got it. Thanks for your eplanation!

mn-mikke (approved):
LGTM

michalkurka:
@mn-mikke, @satai  thanks for the review and comments

(merging after discussing with Ondrej in person)

PR #5412, opened 2021-03-29
XGBoost: Added support for extracting platform-native libraries to a directory

michalkurka:
```
XGB_LIB_DIR=$PWD/xgb_lib_dir

mkdir -p $XGB_LIB_DIR
java -cp build/h2o.jar water.tools.XGBoostLibExtractTool $XGB_LIB_DIR

java -Djava.library.path=$XGB_LIB_DIR -jar build/h2o.jar
```

This is meant to be used in R:
1. H2O Jar is downloaded
2. native libraries extracted in package directory (alongside the jar)
3. h2o.init is modified to set `java.library.path` to the directory that holds the libraries, no data is written to a temporary directory or anywhere else

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/util/NativeLibrary.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/util/NativeLibrary.java
@@ -16,9 +16,9 @@
 package hex.tree.xgboost.util;
 
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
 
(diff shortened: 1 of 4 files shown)


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/util/NativeLibrary.java
```diff
@@ -112,27 +107,52 @@ private ClassLoader getClassLoader() {
...
+    }
+  }
+
+  private static URL getLibResource(String libPath, ClassLoader classLoader) {
+    return classLoader.getResource(libPath);
+  }
+
+  private static File extract(String libPath, URL libResource)
       throws IOException, IllegalArgumentException {
+    assert libResource != null;
```
valenad1, line 155: Some message about why it should not be null, what we expect here?

valenad1 (approved):
It looks good to me, just one minor thing...

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/util/NativeLibrary.java
michalkurka: fixed, thank you!

PR #5415, opened 2021-04-01
Remove redundant code in example of upload-data documentation page

valenad1:


hannah-tillman (approved):

h2o-docs/src/product/data-munging/importing-data.rst
```diff
@@ -12,13 +12,7 @@ Refer to the `Supported File Formats <http://docs.h2o.ai/h2o/latest-stable/h2o-d
 
 .. tabs::
    .. code-tab:: r R
-	
-		# To import airlines file from H2O’s package:
```
michalkurka, line 16: is it possible the author intended to show how to import the iris_wheader.csv dataset that is bundled in R package?

valenad1: Yes, it is possible, firstly I thought that is only copy paste mistake because the python example shows only those two imports and I didn't realize that R package cointains some dataset. The first version was:

```
	# To import small iris data file from H2O’s package:
	> library(h2o)
	> h2o.init(nthreads=-1)
	> irisPath <- system.file("extdata", "iris.csv", package="h2o")
	> iris.hex <- h2o.importFile(path = irisPath, destination_frame = "iris.hex")
```

and after that was change to airlines.

valenad1: I fixed it to use iris from h2o package

michalkurka (approved):

PR #5417, opened 2021-04-01
Fix runit_export_file test

michalkurka:
We cannot expect identical frames because we are using a different double
serialization on the backend
- for as.data.frame we use Double.toHexString(d)
- for exportFile we use RyuDouble.doubleToString(d)

This can produce differences on 1e-16


sebhrusen (approved):
Thanks

PR #5419, opened 2021-04-01
PUBDEV-8072 correct use of remove_collinear_columns for GLM

wendycwong:
This PR resolves the issue in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8072

The implementation is correct.  However, the documentation is not clear.  I have asked Hannah to fix it.  In addition, remove_collinear_columns is only used when solver=IRLSM.  Hence, I added a warning message to inform the user that remove_collinear_columns is not used if the solver is not IRLSM.

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -2757,6 +2757,11 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
   
   private Solver defaultSolver() {
     Solver s = Solver.IRLSM;
+    if (_parms._remove_collinear_columns) { // choose IRLSM if remove_collinear_columns is true
+      Log.info(LogMsg("picked solver " + s));
+      _parms._solver = s;
+      return s;
+    }
     int max_active = 0;
     if(_parms._family == Family.multinomial )
       for(int c = 0; c < _nclass; ++c)
--- a/h2o-algos/src/main/java/hex/glm/GLMModel.java
+++ b/h2o-algos/src/main/java/hex/glm/GLMModel.java
@@ -304,6 +304,23 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
     public boolean _generate_scoring_history = false; // if true, will generate scoring history but will slow algo down
     
     public void validate(GLM glm) {
+      if (_remove_collinear_columns) {
+        if (!(Solver.IRLSM.equals(_solver) || Solver.AUTO.equals(_solver)))
+          glm.warn("remove_collinear_columns", "remove_collinear_columns only works when IRLSM (or " +
+                  "AUTO which will default to IRLSM when remove_collinear_columns=true) is chosen as the solver.  " +
+                  "Otherwise, remove_collinear_columns is not enabled.");
+
+        if (_lambda_search) {
+          glm.warn("remove_collinear_columns", "remove_collinear_columns should only be used with " +
+                  "no regularization, i.e. lambda=0.0.  It is used improperly here with lambda_search.  " +
+                  "Please disable lambda_search and set lambda=0.");
+        } else if (_lambda != null) {
+          boolean nonZeroLambda = Arrays.stream(_lambda).sum() > 0;
+          if (nonZeroLambda)
+            glm.warn("remove_collinear_columns", "remove_collinear_columns should only be used with " +
+                    "no regularization, i.e. lambda=0.0.  It is used improperly here.  Please set lambda=0.");
+        }
+      }
       if (_solver.equals(Solver.COORDINATE_DESCENT_NAIVE) && _family.equals(Family.multinomial))
         throw H2O.unimpl("Naive coordinate descent is not supported for multinomial.");
       if ((_lambda != null) && _lambda_search)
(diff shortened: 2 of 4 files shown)


h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -304,6 +304,9 @@ protected GLMModel deepClone(Key<GLMModel> result) {
     public boolean _generate_scoring_history = false; // if true, will generate scoring history but will slow algo down
     
     public void validate(GLM glm) {
+      if (_remove_collinear_columns && !Solver.IRLSM.equals(_solver))
+        glm.warn("remove_collinear_columns", "remove_collinear_columns only works when IRLSM is " +
```
michalkurka, line 308: I think it is very sensible to make this a warning on the fix release branch. This way it doesn't break existing scripts.

However, should this be an error in master (so that next major) doesn't let user train such model?

wendycwong: @michalkurka I struggled with this one too.  Making it an error may be too destructive and can break a lot of codes.  Let me ask the team.   Wendy

wendycwong:
The team has decide to go with warning messages.  Here are the three warning messages captured:

1. when lambda_search is enabled:

RuntimeWarning: remove_collinear_columns should only be used with no regularization, i.e. lambda=0.0.  It is used improperly here with lambda_search, L2 penalty when alpha does not equal to 1.

2. when lambda is set to 0.1

RuntimeWarning: remove_collinear_columns should only be used with no regularization, i.e. lambda=0.0.  It is used improperly here when lambda > 0 with L2 penalty when alpha does not equal to 1.

3. when coordinate_descent is chosen as the solver: 

RuntimeWarning: remove_collinear_columns only works when IRLSM (or AUTO which will default to IRLSM) is chosen as the solver.  Otherwise, remove_collinear_columns is not enabled.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -304,6 +304,24 @@ protected GLMModel deepClone(Key<GLMModel> result) {
     public boolean _generate_scoring_history = false; // if true, will generate scoring history but will slow algo down
     
     public void validate(GLM glm) {
+      if (_remove_collinear_columns) {
+        if (!(Solver.IRLSM.equals(_solver) || Solver.AUTO.equals(_solver)))
+        glm.warn("remove_collinear_columns", "remove_collinear_columns only works when IRLSM (or " +
+                "AUTO which will default to IRLSM) is chosen as the solver.  Otherwise, remove_collinear_columns is " +
```
ledell, line 310:
```suggestion
                "AUTO which will default to IRLSM when the number of predictors is small) is chosen as the solver.  Otherwise, remove_collinear_columns is " +
```

ledell: @wendycwong IRLSM is only the default sometimes, so maybe it's better to clarify that.  The docs doesn't say what the cutoff for the number of predictors is to switch from IRLSM to L-BFGS, so I just kept it generic, "number of predictors is small".  What do you think?

ledell (approved):

g-eoj:
@wendycwong is `when alpha does not equal to 1.` the warning we want to give? My understanding is would be better to just avoid regularization altogether.

h2o-algos/src/main/java/hex/glm/GLMModel.java
wendycwong: @ledell I actually set IRLSM as the default solver if remove_collinear_columns = true regardless of anything else.  The logic given to me by michalk is this:  if the user doesn't choose the solver, that means s/he does not care and we will choose.  Hence, if remove_collinear_columns = true, it will dictate the solver type to IRLSM regardless of any other parameters settings.  If remove_collinear_columns = false, it will choose the solver according to the rules that are already there.  What do you think of how the solver is chosen?

wendycwong: @ledell This is the new warning message now:

RuntimeWarning: remove_collinear_columns only works when IRLSM (or AUTO which will default to IRLSM when remove_collinear_columns=true) is chosen as the solver.  Otherwise, remove_collinear_columns is not enabled.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -2757,6 +2757,10 @@ private boolean updateEarlyStop() {
   
   private Solver defaultSolver() {
     Solver s = Solver.IRLSM;
+    if (_parms._remove_collinear_columns) { // choose IRLSM if remove_collinear_columns is true
+      _parms._solver = s;
+      return s;
```
michalkurka, line 2762: @wendycwong this early return is causing that the picked solver is not being logged, see: https://github.com/h2oai/h2o-3/pull/5419/files#diff-bfdd0a804deb47b13d4ca209feaa9c26d2845ef8192f44c23bc4d5ab6caac585R2779

wendycwong: I added Log.info to log the new solver.  Is this what you are looking for?

wendycwong: Good point.

wendycwong:
@g-eoj 
You are right.  My warning is too complicated.  Here are the new one:

Lambda_search=True
RuntimeWarning: remove_collinear_columns should only be used with no regularization, i.e. lambda=0.0.  It is used improperly here with lambda_search.  Please disable lambda_search and set lambda=0.

lambda > 0
RuntimeWarning: remove_collinear_columns should only be used with no regularization, i.e. lambda=0.0.  It is used improperly here.  Please set lambda=0.

michalkurka (approved):

PR #5423, opened 2021-04-07
PUBDEV-6364: Add support for saving model to S3 directly (no s3a/s3n needed)

michalkurka:


h2o-persist-s3/src/main/java/water/persist/PersistS3.java
```diff
@@ -169,8 +150,77 @@ public InputStream open(String path) {
...
+            LOG.warn("Temporary file `" + _file.getAbsolutePath() + "` was not deleted. Please delete manually.");
+          }
+        }
+      }
+    }
+  } 
+
+  static class CallbackFileOutputStream extends FileOutputStream {
+    private final Object closeLock = new Object();
+    private volatile boolean closed = false;
```
koniecsveta, line 205: `volatile` is redundant when accessing only from synchronized block

michalkurka: good catch, this is just copy&paste from FileOutputStream; they are doing the same thing - I can remove it but I would rather keep it identical

@koniecsveta WDYT?

koniecsveta: :+1:

koniecsveta:
LGTM, Thanks!

koniecsveta (approved):

michalkurka:
thank you for the review @koniecsveta

PR #5425, opened 2021-04-09
PUBDEV-8087 custom escape character

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8087

### DIFF
--- a/h2o-core/src/main/java/water/api/schemas3/ParseV3.java
+++ b/h2o-core/src/main/java/water/api/schemas3/ParseV3.java
@@ -4,6 +4,7 @@ import water.Iced;
 import water.api.API;
 import water.api.schemas3.KeyV3.FrameKeyV3;
 import water.api.ParseTypeValuesProvider;
+import water.parser.ParseSetup;
 
 public class ParseV3 extends RequestSchemaV3<Iced, ParseV3> {
   // Input fields
@@ -69,4 +70,7 @@ public class ParseV3 extends RequestSchemaV3<Iced, ParseV3> {
   // Zero if blocking==false; row-count if blocking==true
   @API(help="Rows", direction=API.Direction.OUTPUT)
   public long rows;
+
+  @API(help="One-character string used to escape other characters.", direction=API.Direction.INOUT)
+  public byte escapechar = ParseSetup.DEFAULT_ESCAPE_CHAR;
 }
--- a/h2o-core/src/main/java/water/api/schemas3/ParseV3.java
+++ b/h2o-core/src/main/java/water/api/schemas3/ParseV3.java
@@ -71,6 +71,6 @@ public class ParseV3 extends RequestSchemaV3<Iced, ParseV3> {
   @API(help="Rows", direction=API.Direction.OUTPUT)
   public long rows;
 
-  @API(help="One-character string used to escape other characters.", direction=API.Direction.INOUT)
+  @API(help="One ASCII character used to escape other characters.", direction=API.Direction.INOUT)
   public byte escapechar = ParseSetup.DEFAULT_ESCAPE_CHAR;
 }
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -445,28 +445,29 @@ class H2OFrame(Keyed):
 
 
     def _import_parse(self, path, pattern, destination_frame, header, separator, column_names, column_types, na_strings,
-                      skipped_columns=None, custom_non_data_line_markers=None, partition_by=None, quotechar=None):
+                      skipped_columns=None, custom_non_data_line_markers=None, partition_by=None, quotechar=None, escapechar=None):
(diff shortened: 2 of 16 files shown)


h2o-core/src/main/java/water/api/schemas3/ParseV3.java
```diff
@@ -69,4 +70,7 @@
   // Zero if blocking==false; row-count if blocking==true
   @API(help="Rows", direction=API.Direction.OUTPUT)
   public long rows;
+
+  @API(help="One-character string used to escape other characters.", direction=API.Direction.INOUT)
+  public byte escape_char = ParseSetup.DEFAULT_ESCAPE_CHAR;
```
michalkurka, line 75: Should we actually make it a String in the API and then validate its properties?

h2o-py/h2o/frame.py
```diff
@@ -445,28 +445,29 @@ def type(self, col):
 
 
     def _import_parse(self, path, pattern, destination_frame, header, separator, column_names, column_types, na_strings,
-                      skipped_columns=None, custom_non_data_line_markers=None, partition_by=None, quotechar=None):
+                      skipped_columns=None, custom_non_data_line_markers=None, partition_by=None, quotechar=None, escape_char=None):
```
michalkurka, line 448: Should probably be consistent with naming of `quotechar`, Pandas also uses spelling `escapechar`

michalkurka:
@sebhrusen can you please also take a look?

koniecsveta:
R client in progress

h2o-core/src/main/java/water/api/schemas3/ParseV3.java
koniecsveta: can be, I did it like byte since separator is also byte

michalkurka: then it is probably better to be consistent, was just thinking out loud

h2o-py/h2o/h2o.py
```diff
@@ -766,6 +770,7 @@ def parse_setup(raw_frames, destination_frame=None, header=0, separator=None, co
     :param custom_non_data_line_markers: If a line in imported file starts with any character in given string it will NOT be imported. Empty string means all lines are imported, None means that default behaviour for given format will be used
     :param partition_by: A list of columns the dataset has been partitioned by. None by default.
     :param quotechar: A hint for the parser which character to expect as quoting character. Only single quote, double quote or None (default) are allowed. None means automatic detection.
+    :param escape_char: One-character string used to escape other characters. '\' by default.
```
sebhrusen, line 773: Now that this is exposed, I'm wondering if it should not be None by default (char=0 on Java side) to be in sync with pandas.

koniecsveta: can be, but then there will be needed changes everywhere in tests where we are parsing something escaped by the escapechar which was hardcoded in h2o until now ('\\')

koniecsveta:
Flow in progress

sebhrusen (commented):
@koniecsveta the changes look good.

However, as mentioned earlier, for both internal consistency (the `escapechar=None` in Py is a bit confusing as it actually means `escapechar='\\'`, so how do we disable escaping?), and compatibility with pandas (for which escaping is disabled by default), I would disable escaping by default.
Which just means use `DEFAULT_ESCAPE_CHAR = 0`, and as you said, it may just require to explicitely set `escapechar='\\'` on a couple of existing tests, but that's fine as it was a very recent change.

h2o-py/h2o/h2o.py
sebhrusen: correct, but it's only 2 or 3 of them: the ones I created in this PR: https://github.com/h2oai/h2o-3/pull/5388/files

koniecsveta:
flow: https://github.com/h2oai/h2o-flow/pull/173

h2o-core/src/main/java/water/api/schemas3/ParseV3.java
```diff
@@ -69,4 +70,7 @@
   // Zero if blocking==false; row-count if blocking==true
   @API(help="Rows", direction=API.Direction.OUTPUT)
   public long rows;
+
+  @API(help="One-character string used to escape other characters.", direction=API.Direction.INOUT)
+  public byte escapechar = ParseSetup.DEFAULT_ESCAPE_CHAR;
```
sebhrusen, line 75: `byte` at API level? I'm amazed that it works.
In that case, maybe help should be updated to "One ASCII character used to..."

sebhrusen (approved):

h2o-core/src/main/java/water/api/schemas3/ParseV3.java
koniecsveta: we have also `public byte separator`, so did it in the same way / same validation

PR #5427, opened 2021-04-09
fix failing cran check

koniecsveta:


michalkurka (approved):
Thank you @koniecsveta

PR #5428, opened 2021-04-09
ITA-911: Script to validate Pull Request metadata using a corresponding JIRA ticket

michalkurka:


gradle/prCheck.gradle
```diff
@@ -0,0 +1,95 @@
...
+    def jsonSlurper = new JsonSlurper()
+    def pullDetails = jsonSlurper.parseText(new URL("https://api.github.com/repos/h2oai/h2o-3/pulls/$changeId").getText())
+    def title = pullDetails.title
+    println ""
+    println "Checking pull request ($changeId): $title"
+    println ""
+    def hexdevMatcher = title =~ /.*HEXDEV-[0-9]{4,5}.*/
+    def itaMatcher = title =~ /.*ITA-[0-9]{3,5}.*/
+    if (hexdevMatcher.matches() || itaMatcher.matches()) {
+        println "HEXDEV and ITA jiras are not validated"
```
valenad1, line 59: Any hint for developer how to fix it? Or where to seek solution.

gradle/prCheck.gradle
```diff
@@ -0,0 +1,95 @@
...
+        return []
+    }
+    def pubdevMatcher = title =~ /.*(PUBDEV-[0-9]{4,5}).*/
+    if (!pubdevMatcher.matches()) {
+        return ["Pull Request ${changeId} doesn't have associated JIRA ticket. Please add jira id in the title of the PR"]
```
valenad1, line 64: I would add information: Title of PR should contain string 'PUBDEV-ticketNumber'.

valenad1: I see it in GOOD_PR_PROCESS but it could help to remind it here

gradle/prCheck.gradle
```diff
@@ -0,0 +1,95 @@
...
+import com.github.ricksbrown.cowsay.Cowsay
+
+def GOOD_PR_PROCESS = """
+How to pass this check?
+    1. Always have a Jira before you start working on a task (otherwise you cannot know what branch to start from).
+    2. Make sure the Jira has a 'Fix Version' defined. If it doesn't consult the team what is a good fix version for your change.
+    3. If you are working on a Jira with 'Fix Version' ending with '.1' you are working on a change that should go to a master branch. 
+       All other changes should go to the (current) fix release branch. Such branches are prefixed 'rel-'. 
+    4. Put the Jira to 'In Progress' state.
+    5. Make a code change. Include Jira ticket ID in message of the main commit (this is not checked).
```
valenad1, line 24: What is main commit? I would write ... of the all commits ...

gradle/prCheck.gradle
```diff
@@ -0,0 +1,95 @@
...
+    println ""
+    println "Checking pull request ($changeId): $title"
+    println ""
+    def hexdevMatcher = title =~ /.*HEXDEV-[0-9]{4,5}.*/
+    def itaMatcher = title =~ /.*ITA-[0-9]{3,5}.*/
+    if (hexdevMatcher.matches() || itaMatcher.matches()) {
+        println "HEXDEV and ITA jiras are not validated"
+        return []
+    }
+    def pubdevMatcher = title =~ /.*(PUBDEV-[0-9]{4,5}).*/
```
valenad1, line 62: Just point to think about: Maybe we can be more strict here a say all PR will start with 'PUBDEV-{number}: some desct text'.

gradle/prCheck.gradle
```diff
@@ -0,0 +1,95 @@
...
+    if (fixVersionValue.endsWith('.1')) {
+        if (baseBranch != 'master')
+            return ["Pull Request ${changeId} is linked to jira ticket https://h2oai.atlassian.net/browse/${jiraId}. " +
+                    "This jira is marked for release in {$fixVersionValue}. This release will be done from a master branch. Change the base branch of the PR to 'master'."]
+    } else {
+        if (!baseBranch.startsWith('rel-')) {
+            return ["Pull Request ${changeId} is linked to jira ticket https://h2oai.atlassian.net/browse/${jiraId}. " +
+                    "This jira is marked for release in {$fixVersionValue}. This release will be done from a fix release branch. Change the base branch to a correspond to the right fix release branch."]
+        }
+    }
```
valenad1, line 93: Just point to think about. What about the task description? We can check if it is not empty, you can always write a link to the jira at least.

valenad1, line 93: Just point to think about. What about the task assignee? Assure specific user can be pain but maybe check if assignee != 'New H2O Bugs' could be helpful.

valenad1 (commented):
Great change to help the PR review process! I have added a couple of suggestions.

A general question: Can this script be triggered with the PR edit or I need to rerun build to fix the bugs?

gradle/prCheck.gradle
michalkurka: Jira then makes a links for each of the commits

gradle/prCheck.gradle
michalkurka: This doesn't produce an error, they are not just validated because they would require a private access to JIRA and they are not really part of releases so all good.

gradle/prCheck.gradle
michalkurka: Good point, that wouldn't hurt

gradle/prCheck.gradle
michalkurka: I am not sure what you mean here. Task description where? Please elaborate

michalkurka: That is a very good point, I might add it in the future. For now the main goal is to make sure the versions are there. This was a pain point in the last release where I had to go manually over the jiras and check if they were done or not.

Making sure the assignee is not wrong would be nice but not necessary for changelog. Moreover, when you click on 'In Progress' it changes the assignee to you.

But very good point and I think I can add it in the future. For now I want to keep it minimal to avoid false positives.

gradle/prCheck.gradle
michalkurka: Yep, we can add in the future. But this is usually not the issue. I am trying to make sure here the jira link is there in this PR.

gradle/prCheck.gradle
michalkurka: Done

gradle/prCheck.gradle
michalkurka: I changed the wording to "one of your commits" to avoid confusing term "main commit"

michalkurka:
> A general question: Can this script be triggered with the PR edit or I need to rerun build to fix the bugs?

You don't need to re-run the whole build. There is an option to re-run the just failed checks.

I think it doesn't matter that it is a bit annoying - it should motivate people to do it right to begin with :)

gradle/prCheck.gradle
valenad1: Sorry, I meant PR description. We can check if the PR description is not empty.

michalkurka: Yeah, I almost never manually fill the description. I put the important stuff in the commit messages (where it really matters). So I would be the victim of this check :)

Something to think about if this proves to be useful.

valenad1: :)

valenad1 (approved):
Thanks

michalkurka:
@valenad1 thanks for the review and the new ideas you gave me for future improvements

PR #5429, opened 2021-04-09
hack to peek into DKV from client

sebhrusen:
Allows to explore DKV from Py client using:
```python
h2o.cluster().get_dkv()
```

### DIFF
--- a/h2o-core/src/main/java/water/HeartBeatThread.java
+++ b/h2o-core/src/main/java/water/HeartBeatThread.java
@@ -102,6 +102,7 @@ public class HeartBeatThread extends Thread {
       // persistent KV pairs are stored
       hb.set_free_disk(H2O.getPM().getIce().getUsableSpace());
       hb.set_max_disk (H2O.getPM().getIce().getTotalSpace() );
+      hb.set_dkv(Cleaner.Histo.current(false).dkv());
 
       // get cpu utilization for the system and for this process.  (linux only.)
       LinuxProcFileReader lpfr = new LinuxProcFileReader();
--- a/h2o-core/src/main/java/water/api/schemas3/CloudV3.java
+++ b/h2o-core/src/main/java/water/api/schemas3/CloudV3.java
@@ -2,7 +2,9 @@ package water.api.schemas3;
 
 import water.*;
 import water.api.API;
+import water.util.TwoDimTable;
 
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 public class CloudV3 extends RequestSchemaV3<Iced, CloudV3> {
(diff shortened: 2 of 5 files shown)


h2o-core/src/main/java/water/HeartBeatThread.java
```diff
@@ -102,6 +102,7 @@ public void run() {
       // persistent KV pairs are stored
       hb.set_free_disk(H2O.getPM().getIce().getUsableSpace());
       hb.set_max_disk (H2O.getPM().getIce().getTotalSpace() );
+      hb.set_dkv(Cleaner.Histo.current(false).dkv());
```
michalkurka, line 105: What is the purpose of populating a transient field in the heartbeat structure? Heartbeats are meant to be exchanged between the node. This seems like like a misuse to me.

h2o-core/src/main/java/water/api/schemas3/CloudV3.java
```diff
@@ -187,6 +189,9 @@ public NodeV3() {}
     @API(help="Memory Bandwidth", direction=API.Direction.OUTPUT)
     public double mem_bw;
 
+    @API(help="DKV", direction = API.Direction.OUTPUT)
+    public TwoDimTableV3 dkv;
```
michalkurka, line 193: the cloud api is called frequently, it is not possible to include a structure of unlimited size in this response

h2o-py/h2o/backend/cluster.py
```diff
@@ -249,6 +249,10 @@ def get_status_details(self):
         table = H2OTwoDimTable(cell_values=node_table, col_header=keys, row_header=keys)
         return table
 
+    def get_dkv(self):
+        info = h2o.api("GET /3/Cloud")
+        return [node['dkv'] for node in info._props['nodes']]
```
michalkurka, line 254: you seem to be iterating over all the nodes, however, given the transient nature of the field - it will only be defined on the requesting node

moreover - this API is to dangerous to be directly exposed here

I would imagine this would be its own API, hidden from users

h2o-core/src/main/java/water/HeartBeatThread.java
sebhrusen: it's hack, was used on single node only. No plan to merge this.

michalkurka (commented):
I would introduce a separate endpoint that will collect the needed information without modifying infrastructure classes.

KeySnapshot already has size information about the keys. Perhaps you can expose that information with some minor modification to how KeySnapshot works

sebhrusen:
Not intended for merge.

michalkurka:
> Not intended for merge.

but could be an useful feature if implemented in another way, eg. as a rapid expression utilizing KeySnapshot

something like 
```
 ls -a -l
```
in shell

sebhrusen:
@michalkurka Yes I agree that it could be useful, if thought and designed properly :)
Agree with your suggestions: `KeySnapshot` looks like a better approach for this, I think I didn't realize that `Value._max` is actually the size of the object (maybe if it had been named differently...).
This is could a nice Friday task.

michalkurka:
> I think I didn't realize that Value._max is actually the size of the object (maybe if it had been named differently...).

:)

I am not sure myself how this came to be

PR #5431, opened 2021-04-10
PUBDEV-8100: ModelDescriptor field in MOJO is not Serializable

michalkurka:


michalkurka:
@satai the added serialization check revealed that CoxPH mojo was not serializable if model was trained with stratification, please take a look

satai:
The code LGTM but maybe a test would be nice?

mn-mikke (approved):
LGTM

Strange, I would expect [this SW test](https://github.com/h2oai/sparkling-water/blob/master/ml/src/test/scala/ai/h2o/sparkling/ml/algos/H2OCoxPHTestSuite.scala#L45) to fail in case of the problem with serialization.

satai:
@mn-mikke There is no stratification there, if I look correctly?

mn-mikke:
> @mn-mikke There is no stratification there, if I look correctly?

Ok, got it. Thanks for your explanation!

michalkurka:
> The code LGTM but maybe a test would be nice?
@satai the serialization is now tested in every invocation of testJavaScoring - https://github.com/h2oai/h2o-3/pull/5431/commits/23b31b0b924b6d5a6e6b1aa5ffa23d8ba8221e10

Did you mean that or something else?

satai:
@michalkurka is stratification used there?

michalkurka:
@michalkurka is stratification used there?

yes, that is how I found the failures

michalkurka:
@satai I accidentally merged this already (instead of another PR) - if you have comments please send - I will incorporate in a new one

satai:
OK for me @michalkurka

michalkurka:
@satai thanks for confirming, sorry about the early merge

PR #5433, opened 2021-04-10
PUBDEV-8102: AutoML crashes if model parameter validation fails

michalkurka:
Even though builder.init(expensive = false) shouldn't cause H2OIllegalArgumentException, the reality is different (currently XGBoost does do that).

This PR prevents AutoML from crashing if model builder produces H2OIllegalArgumentException in the parameter validation phase.


sebhrusen (approved):
Thanks you @michalkurka

tomasfryda (approved):
LGTM. Thank you @michalkurka!

PR #5434, opened 2021-04-12
Make sure CloudHandler works in client mode with no members

michalkurka:
Found by @wendycwong in http://mr-0xg1:8080/job/h2o-3-benchmark-pipeline/job/master/603/artifact/gbm-benchmark-client/h2o-3/h2o-client_gbm-client_1_603.log

### DIFF
--- a/h2o-core/src/main/java/water/api/CloudHandler.java
+++ b/h2o-core/src/main/java/water/api/CloudHandler.java
@@ -41,13 +41,15 @@ class CloudHandler extends Handler {
     cloud.locked = Paxos._cloudLocked;
     cloud.internal_security_enabled = H2OSecurityManager.instance().securityEnabled;
 
-    // Fetch and calculate cloud metrics from individual node metrics.
+
+    // set leader
+    H2ONode leader = H2O.CLOUD.leaderOrNull(); // leader might be null in client mode if clouding didn't finish yet
+    cloud.leader_idx = leader == null ? -1 : leader.index();
+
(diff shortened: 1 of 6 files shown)


h2o-core/src/main/java/water/api/CloudHandler.java
```diff
@@ -41,21 +41,21 @@ public CloudV3 status(int version, CloudV3 cloud) {
...
+    cloud.leader_idx = leader == null ? -1 : leader.index();
+
+    // set list of members (might be empty)
     H2ONode[] members = H2O.CLOUD.members();
     cloud.bad_nodes = 0;
     cloud.cloud_healthy = true;
-    
-    H2ONode leader = H2O.CLOUD.leader();
-    cloud.leader_idx = leader == null ? -1 : leader.index();
-    if (null != members) {
```
mn-mikke, line 51: This condition removal looked suspicious, but didn't find a place in H2O-3 codebase where a `null` would be passed to the  constructor of `water.H2O`. So it should be ok then.

mn-mikke (approved):
LGTM

h2o-core/src/main/java/water/api/CloudHandler.java
michalkurka: Yeah, thanks for checking that @mn-mikke.

If it ever could be null, the previous methods that rely on `_memary` would fail anyway. It was just confusing there, better to clean-up such stuff.

PR #5436, opened 2021-04-13
Remove Error by PCA in Max_runtime tests

wendycwong:
I am removing the error raised by PCA in the following tests:

pyunit_pubdev_4702_algo_max_runtime_secs_large.py
runit_pubdev_4702_algo_max_runtime_secs_large.R

I think there is something wrong with PCA and it is not respecting max_runtime_secs.  That is why the tests keep failing.  PCA timing tests will run and we will get a result whether it fails or pass.  If it fails, no error will be generated.  I have written a JIRA for this: https://h2oai.atlassian.net/browse/PUBDEV-8103


michalkurka (approved):
thank you Wendy!

PR #5437, opened 2021-04-13
PUBDEV-8114: GBM - Save memory in CV by sharing 'validation workspace' with training frame

michalkurka:


michalkurka:
@wendycwong please review:

The idea of this PR is that we can share the workspace where temporary predictions are made between training and validation frame in the case of CV. In CV both of the frames are actually identical physical frame - the only difference is weights. This means we can use the "unused space" of the workspace of training frame for validation workspace and vice versa.

IMHO no need for a specific test, the existing test suite already verifies correcness.

wendycwong (approved):

PR #5438, opened 2021-04-14
PUBDEV-8105: Add percentile rank transform 

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8105

This should work roughly like this:
1) for each probability p in {i/n for i in 0 .. n} calculate corresponding quantile q value to create mapping {(p_i, q_i) for i in 0 .. n}  for each column
2) for column in each data point linearly interpolate its percentile rank using the mapping obtained in step (1)
3) the mapping from (1) is persisted and used for prediction and mojos

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
@@ -119,6 +119,8 @@ public class StackedEnsembleStepsProvider
                 if (aml().getResponseColumn().isCategorical()) {
                     // Add logit transform
                     stackedEnsembleParameters._metalearner_transform = StackedEnsembleParameters.MetalearnerTransform.Logit;
+                    //TODO: REMOVE THE FOLLOWING LINE BEFORE MERGING - THIS IS HERE ONLY TO GET BIGGER TEST COVERAGE
+                    stackedEnsembleParameters._metalearner_transform = StackedEnsembleParameters.MetalearnerTransform.PercentileRank;
                 }
 
                 return stackedEnsembleParameters;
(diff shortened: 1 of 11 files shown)


h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
```diff
@@ -119,6 +119,8 @@ protected StackedEnsembleParameters getStackedEnsembleParameters(Key<Model>[] ba
                 if (aml().getResponseColumn().isCategorical()) {
                     // Add logit transform
                     stackedEnsembleParameters._metalearner_transform = StackedEnsembleParameters.MetalearnerTransform.Logit;
+                    //TODO: REMOVE THE FOLLOWING LINE BEFORE MERGING - THIS IS HERE ONLY TO GET BIGGER TEST COVERAGE
+                    stackedEnsembleParameters._metalearner_transform = StackedEnsembleParameters.MetalearnerTransform.PercentileRank;
```
tomasfryda, line 123: ❗ Remove before merging.

h2o-genmodel/src/main/java/hex/genmodel/algos/ensemble/StackedEnsembleMojoModel.java
```diff
@@ -21,9 +38,31 @@ private static double logit(double p) {
         return  x == 0 ? -19 : Math.max(-19, Math.log(x));
     }
 
-    private static void logitTransformRow(double[] basePreds){
+    private static double[] logitTransformRow(double[] basePreds){
```
tomasfryda, line 41: This change in the API might be useful if we will decide to use percentile rank combined with logit transform (cbinded together) so we output more "columns" than there is base models.

PR #5440, opened 2021-04-15
PUBDEV-7948 fix NPE when saving and uploading binary rulefit model

koniecsveta:


h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
```diff
@@ -141,7 +141,9 @@ protected Futures remove_impl(Futures fs, boolean cascade) {
 
     void updateModelMetrics( GLMModel glmModel, Frame fr){
         for (Key<ModelMetrics> modelMetricsKey : glmModel._output.getModelMetrics()) {
-            this.addModelMetrics(modelMetricsKey.get().deepCloneWithDifferentModelAndFrame(this, fr));
+            // what is null here was already added to RF model from GLM submodel during hex.rulefit.RuleFit.RuleFitDriver.fillModelMetrics
```
michalkurka, line 144: Can you explain this commet please?

koniecsveta: yes, when I binary save the rulefit model, h2o saves metrics which are directly on the model, not on glm submodel, so when I upload it and score it and it gets to this part of code, glm's model metrics calculated during training which were not saved will cause here npe, only keys were stored. But when I save them after training directly on rulefit model in hex.rulefit.RuleFit.RuleFitDriver.fillModelMetrics, they get saved during binary save, so I have them already on rulefit model and it does no bad that they were not saved on glm and I can just skip those keys by this new condition. I save on rulefit model only new metrics calculated during the scoring which I executed. This seemed to me as easier way than to modify model serialization. WDYT about this?

michalkurka (approved):

PR #5444, opened 2021-04-15
Make GBM test of multinomial auc early stopping more deterministic

michalkurka:


maurever (approved):
Thank you @michalkurka for this fast and useful fix!

PR #5446, opened 2021-04-16
CoxPH check both efrom and breslow

satai:


michalkurka (approved):
thanks for the improvement!

PR #5447, opened 2021-04-16
PUBDEV-6249: Produce a warning if early stopping is not reproducible

michalkurka:
For tree-based models


michalkurka:
@meganjkurka since this PR is targeting a fix-release - instead of a strict requirement we are introducing a warning to avoid breaking users scripts

Do you think this is enough to resolve the jira or do you think we should produce an error in the next major version?

meganjkurka:
@michalkurka I think warning is enough

wendycwong (approved):
This makes sense.

PR #5448, opened 2021-04-16
PUBDEV-8092: Job polling fail tolerance implementation for R according to the same python implementation in PUBDEV-8034 

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-8092
https://h2oai.atlassian.net/browse/PUBDEV-8034

I followed the python implementation as much as possible. I exclude from repeating the cases if Job is not found in the cluster or 
if function inputs' validation fail. The check is done with string compare though and AFAIK it fails with API version 4.

[PUBDEV-8092-job-poll-patch.txt](https://github.com/h2oai/h2o-3/files/6352393/PUBDEV-8092-job-poll-patch.txt)
[PUBDEV-8092-job-poll-patch-test.txt](https://github.com/h2oai/h2o-3/files/6352398/PUBDEV-8092-job-poll-patch-test.txt)

### DIFF
--- a/h2o-r/h2o-package/R/communication.R
+++ b/h2o-r/h2o-package/R/communication.R
@@ -348,11 +348,12 @@
               parms = parms, method = "POST", ...)
 }
 
-.h2o.doSafeREST <- function(h2oRestApiVersion, urlSuffix, parms, method, fileUploadInfo,autoML=FALSE, ...) {
-  stopifnot(is.character(urlSuffix))
-  stopifnot(is.character(method))
-  if (!missing(fileUploadInfo)) stopifnot(is(fileUploadInfo, "FileUploadInfo"))
-
+.h2o.doSafeREST <- function(h2oRestApiVersion, urlSuffix, parms, method, fileUploadInfo, autoML=FALSE, doValidation=TRUE, ...) {
+  if (doValidation) {
+    stopifnot(is.character(urlSuffix))
+    stopifnot(is.character(method))
+    if (!missing(fileUploadInfo)) stopifnot(is(fileUploadInfo, "FileUploadInfo"))
+  }
   rv = .h2o.doREST(h2oRestApiVersion = h2oRestApiVersion, urlSuffix = urlSuffix,
                    parms = parms, method = method, fileUploadInfo = fileUploadInfo,autoML=autoML, ...)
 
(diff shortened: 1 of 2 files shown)


h2o-r/h2o-package/R/communication.R
```diff
@@ -799,9 +799,42 @@ h2o.list_jobs <- function() {
...
+  i <- 0
+  while (i < 30) {
+    rawResponse <- try(.h2o.doSafeGET(urlSuffix = myJobUrlSuffix))
+    if(class(rawResponse) == "try-error" && jobPoolSucees){
+      error_type <- attr(rawResponse,"condition")
+      if (jobIsRecoverable) {
+        print(sprintf("Job request failed %s, waiting for cluster to restart.", error_type$message))
+        Sys.sleep(10)
+      } else {
+        if(grepl("Job is missing", error_type$message, fixed = TRUE)) {
```
michalkurka, line 815: Interesting, I am not even sure if the python client handles this case - this might be a bug in Python client. Good thinking!

However, I do think we need to detect the state in another way: grepping the message seems fragile.

h2o-r/h2o-package/R/communication.R
```diff
@@ -799,9 +799,42 @@ h2o.list_jobs <- function() {
   df
 }
 
-h2o.get_job <- function(job_key) {
+h2o.get_job <- function(job_key, jobPoolSucees = FALSE, jobIsRecoverable = FALSE) {
```
michalkurka, line 802: spelling errors: pool -> poll; sucees -> success (I think "pool" is in other places as well)

valenad1: Thank you!

h2o-r/h2o-package/R/communication.R
valenad1: We conclude on a call that it is fragile but it is a best effort to skip unnecessary repeat call. In case of this check failure (e.g because of API change), the implementation is still working analogically to python client.

h2o-r/h2o-package/R/communication.R
```diff
@@ -799,9 +799,42 @@ h2o.list_jobs <- function() {
...
+      error_type <- attr(rawResponse,"condition")
+      if (jobIsRecoverable) {
+        print(sprintf("Job request failed %s, waiting for cluster to restart.", error_type$message))
+        Sys.sleep(10)
+      } else {
+        if(grepl("Job is missing", error_type$message, fixed = TRUE)) {
+          print(sprintf("Job is no longer exists: %s", error_type$message))
+          break;
+        }
+        if (grepl("is not TRUE", error_type$message, fixed = TRUE)) {
```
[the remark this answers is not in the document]
valenad1: It is error from stopifnot(...) function. I remove this check and use the logic you proposed on a call

michalkurka (approved):
Nice work, thank you @valenad1

michalkurka:
@valenad1 please rebase on top of rel-zipf

PR #5451, opened 2021-04-19
PUBDEV-8111: Add support for CDP 7.2

michalkurka:


g-eoj (approved):
I don't know how to determine if the artifact version numbers are correct. Other than that, looks good, thanks.

PR #5454, opened 2021-04-21
Fix runit_pubdev_2441_pvalue_w_offset.R failure

wendycwong:
Added debug printing to print out
h2o p-values, r p-values for GLM so that we can see what the problem is.


michalkurka (approved):

michalkurka:
@wendycwong I merged this one - lets see the results in the repeated nightly builds

wendycwong:
Thank you michalk!

PR #5455, opened 2021-04-23
PUBDEV-8120: Fix SE MOJO for cases when sub-model doesn't have the same columns as the metalearner

michalkurka:


tomasfryda (approved):

PR #5459, opened 2021-04-26
PUBDEV-8125: CoxPH - Make efron-method fully deterministic

michalkurka:


h2o-core/src/main/java/water/LocalMR.java
```diff
@@ -38,17 +34,26 @@ public LocalMR(MrFun mrt, int nthreads, H2O.H2OCountedCompleter cc){
...
+    return this;
+  }
+  public boolean isReproducible() {
+    return !_reusePrevTsk;
+  }
   private LocalMR(LocalMR src, LocalMR prevTsk,int lo, int hi) {
     super(src);
     _root = src._root;
-    _prevTsk = prevTsk;
+    _reusePrevTsk = src._reusePrevTsk;
```
maurever, line 47: Is it really necessary to use `Tsk` instead of `Task`? It looks confusing to me, and I cant see the added value of this shorter version. :)

```suggestion
    _reusePrevTask = src._reusePrevTask;
```

maurever (approved):
LGTM, good catch! Just one small improvement.

h2o-core/src/main/java/water/LocalMR.java
michalkurka: This is just to be consistent with the existing naming: _prevTsk -> _reusePrevTsk. I would also have to change the original field name.

This is a private field, the public-facing method has "Task" in it - so imho we are okay there. WDYT @maurever ?

maurever: Ok, let keep it. It is just funny to me. :)

michalkurka: Thanks, at least you get joy from reading the code ;)

PR #5461, opened 2021-04-27
Added support for load balancers to h2o-helm.

TravisCouture:
Added `loadbalancer.yaml` and updated `values.yaml` templates to enable support for load balancers if the k8s deployments supports it.  (Option for ingress still exists).


michalkurka (approved):

michalkurka:
thank you @TravisCouture

PR #5462, opened 2021-04-27
PUBDEV-8118: AutoML improvements: add more regularization to GLM metalearner

ledell:
https://h2oai.atlassian.net/browse/PUBDEV-8118

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
@@ -10,6 +10,7 @@ import hex.KeyValue;
 import hex.Model;
 import hex.ensemble.StackedEnsembleModel;
 import hex.ensemble.StackedEnsembleModel.StackedEnsembleParameters;
+import hex.glm.GLMModel;
 import water.DKV;
 import water.Job;
 import water.Key;
@@ -116,8 +117,8 @@ public class StackedEnsembleStepsProvider
                 stackedEnsembleParameters._metalearner_parameters._keep_cross_validation_models = buildSpec.build_control.keep_cross_validation_models;
                 stackedEnsembleParameters._metalearner_parameters._keep_cross_validation_predictions = buildSpec.build_control.keep_cross_validation_predictions;
                 // add custom alpha in GLM metalearner
-                // not working:
-                //stackedEnsembleParameters._metalearner_parameters._alpha = new double[] {0.5, 1.0};
+                GLMModel.GLMParameters metalearner_params = (GLMModel.GLMParameters)stackedEnsembleParameters._metalearner_parameters;
+                metalearner_params._alpha = new double[]{0.5, 1.0};
 
                 if (aml().getResponseColumn().isCategorical()) {
                     // Add logit transform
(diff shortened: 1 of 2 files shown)


h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
```diff
@@ -115,6 +116,9 @@ protected StackedEnsembleParameters getStackedEnsembleParameters(Key<Model>[] ba
                 stackedEnsembleParameters.initMetalearnerParams();
                 stackedEnsembleParameters._metalearner_parameters._keep_cross_validation_models = buildSpec.build_control.keep_cross_validation_models;
                 stackedEnsembleParameters._metalearner_parameters._keep_cross_validation_predictions = buildSpec.build_control.keep_cross_validation_predictions;
+                // add custom alpha in GLM metalearner
+                GLMModel.GLMParameters metalearner_params = (GLMModel.GLMParameters)stackedEnsembleParameters._metalearner_parameters;
+                metalearner_params._alpha = new double[]{0.5, 1.0};
```
sebhrusen, line 121: do we add it only for AutoML or shouldn't we rather have it for the AUTO metalearner?
wdyt? Also documentation-wise.

sebhrusen: my main concern here is that we introduce 2 levels of AUTO that may confuse the user: the AUTO SE and the AUTO SE in AUTO ML

ledell: My old PR had it for default SE, but I thought it was best to keep this PR isolated & simple.  I do think we could consider updating the default metalearner in SE as well, but I didn't have time to fully think about the repercussions for doing that (e.g. reproducibility, tests, etc), and also it's harder to do that neatly on the client-side since the alpha array is only supported on the backend.

h2o-docs/src/product/automl.rst
```diff
@@ -497,7 +501,7 @@ Random Grid Search Parameters
 
 AutoML performs a hyperparameter search over a variety of H2O algorithms in order to deliver the best model. In the table below, we list the hyperparameters, along with all potential values that can be randomly chosen in the search. If these models also have a non-default value set for a hyperparameter, we identify it in the list as well. Random Forest and Extremely Randomized Trees are not grid searched (in the current version of AutoML), so they are not included in the list below.
 
-**Note**: AutoML does not run a grid search for GLM. Instead AutoML builds a single model with ``lambda_search`` enabled and passes a list of ``alpha`` values. It returns only the model with the best alpha-lambda combination rather than one model for each alpha.
+**Note**: AutoML does not run a standard grid search for GLM (returning all the possible models). Instead AutoML builds a single model with ``lambda_search`` enabled and passes a list of ``alpha`` values. vIt returns only the model with the best alpha-lambda combination rather than one model for each alpha-lambda combination.
```
tomasfryda, line 504: Typo in the last sentence - "vIt" instead of "It"
```suggestion
**Note**: AutoML does not run a standard grid search for GLM (returning all the possible models). Instead AutoML builds a single model with ``lambda_search`` enabled and passes a list of ``alpha`` values. It returns only the model with the best alpha-lambda combination rather than one model for each alpha-lambda combination.
```

tomasfryda (approved):
There is a typo in the documentation, please see the suggestion above. Otherwise it looks good to me. Thank you @ledell !

sebhrusen (approved):

PR #5463, opened 2021-04-27
PUBDEV-8127: AutoML: removing auc and aucpr from default leaderboard columns for multinomial

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8127


ledell (approved):

michalkurka:
@sebhrusen is this ready to go?

sebhrusen:
@michalkurka I think we forgot that one: I don't think so, many tests were failing after those changes iirc.

sebhrusen:
After fixing the tests, it's good to go.

PR #5466, opened 2021-04-29
PUBDEV-8131 Add parameter auc_type to  performance method

maurever:
- [x] Add new parameter to performance method
- [x] Add test in R for multinomial AUC
- [x] Improve Python test for multinomial AUC
- [x] Small improvements of related tests and code
- [ ] Fix Cross-validation metrics summary to not print aucpr (this will be fixed in different PR)

JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8131

### DIFF
--- a/h2o-core/src/main/java/hex/ModelMetricsMultinomial.java
+++ b/h2o-core/src/main/java/hex/ModelMetricsMultinomial.java
@@ -38,16 +38,16 @@ public class ModelMetricsMultinomial extends ModelMetricsSupervised {
     sb.append(" mean_per_class_error: " + (float)_mean_per_class_error + "\n");
     sb.append(" hit ratios: " + Arrays.toString(_hit_ratios) + "\n");
     sb.append(" AUC: "+auc()+ "\n");
-    sb.append(" auc_pr: "+ pr_auc()+ "\n");
+    sb.append(" pr_auc: "+ pr_auc()+ "\n");
     if(_auc.getAucTable() == null){
       sb.append(" AUC table: is not computed because it is disabled (model parameter 'auc_type' is set to AUTO or NONE) or due to domain size (maximum is 50 domains).\n");
-      sb.append(" auc_pr table: is not computed because it is disabled (model parameter 'auc_type' is set to AUTO or NONE) or due to domain size (maximum is 50 domains).\n");
+      sb.append(" pr_auc table: is not computed because it is disabled (model parameter 'auc_type' is set to AUTO or NONE) or due to domain size (maximum is 50 domains).\n");
     } else if(_domain.length <= 20) {
       sb.append(" AUC table: " + _auc.getAucTable()+"\n");
-      sb.append(" auc_pr table: " + _auc.getAucPrTable()+"\n");
+      sb.append(" pr_auc table: " + _auc.getAucPrTable()+"\n");
     } else {
       sb.append(" AUC table: too large to print.\n");
-      sb.append(" auc_pr table: too large to print.\n");
+      sb.append(" pr_auc table: too large to print.\n");
     }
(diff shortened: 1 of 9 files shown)


h2o-core/src/main/java/water/api/ModelMetricsHandler.java
```diff
@@ -260,6 +266,11 @@ public ModelMetricsListSchemaV3 score(int version, ModelMetricsListSchemaV3 s) {
     if (customMetricFunc == null) {
       customMetricFunc = parms._model._parms._custom_metric_func;
     }
+    // set user given auc type, used for scoring a testing data fe. from h2o.performance function
+    MultinomialAucType at = parms._model._parms._auc_type;
```
maurever, line 270: The auc_type is then used via Model parameters, so that is the reason why I do this here. It is the simplest way, how to propagate a new parameter value. Otherwise, I would have changed the parameters and behavior of many other methods in our Java core.

h2o-py/h2o/model/model_base.py
```diff
@@ -341,7 +341,7 @@ def training_model_metrics(self):
         """
         return self._model_json["output"]["training_metrics"]._metric_json
     
-    def model_performance(self, test_data=None, train=False, valid=False, xval=False):
+    def model_performance(self, test_data=None, train=False, valid=False, xval=False, auc_type="none"):
```
michalkurka, line 344: this parameter has only affect when test_data is provided, we should validate that if test_data is None auc_type is "none" and if not produce a warning

h2o-r/h2o-package/R/models.R
```diff
@@ -949,7 +951,7 @@ h2o.feature_frequencies <- feature_frequencies.H2OModel
 #' h2o.performance(model = prostate_gbm_balanced, train = TRUE)
 #' }
 #' @export
-h2o.performance <- function(model, newdata=NULL, train=FALSE, valid=FALSE, xval=FALSE, data=NULL) {
+h2o.performance <- function(model, newdata=NULL, train=FALSE, valid=FALSE, xval=FALSE, data=NULL, auc_type="NONE") {
```
michalkurka, line 954: same as in python

h2o-core/src/main/java/hex/ModelMetricsMultinomial.java
```diff
@@ -38,16 +38,16 @@ public String toString() {
     sb.append(" mean_per_class_error: " + (float)_mean_per_class_error + "\n");
     sb.append(" hit ratios: " + Arrays.toString(_hit_ratios) + "\n");
     sb.append(" AUC: "+auc()+ "\n");
-    sb.append(" auc_pr: "+ pr_auc()+ "\n");
```
michalkurka, line 41: I think there should be a test

michalkurka:
@maurever this seems like 2 different changes in one PR, one feature one bug fix - I think this should have 2 PRs so that it can be properly logged in changelog

h2o-py/h2o/model/model_base.py
maurever: Good point. Thank you @michalkurka, will do.

maurever:
Thank you for your review, @michalkurka. I will split this PR into two different PRs. Good suggestion.

h2o-py/h2o/model/model_base.py
maurever: Done.

h2o-r/h2o-package/R/models.R
maurever: Done.

michalkurka (approved):
Very nice! Thank you @maurever

PR #5469, opened 2021-05-01
Create frame with sequence column

valenad1:
Cleanup for testIDColumnOperationEncoder


h2o-test-support/src/main/java/water/fvec/TestFrameBuilderTest.java
```diff
@@ -253,5 +253,21 @@ public void testBadVec(){
...
+  public void testSequenceIntDataForCol() {
+    Scope.enter();
+    int numRows = 10;
+    try {
+      Frame f = new TestFrameBuilder()
+              .withVecTypes(Vec.T_NUM)
+              .withSequenceIntDataForCol(0, 0, numRows)
+              .build();
+      Scope.track(f);
+      assertVecEquals(Vec.makeSeq(0, numRows), f.vec(0), 0);
```
michalkurka, line 267: this will only be true for low number of rows because it will create a single chunk, generally it is not true

michalkurka (approved):
thank you, good change!

PR #5472, opened 2021-05-03
PUBDEV-8119: Local Build formatting fixes

hannah-tillman:
For: https://h2oai.atlassian.net/browse/PUBDEV-8119

Two things:
1. The "welcome.rst" python lexing fix is only a possibility. I tried a bunch of different formats and this is the only way I've found that keeps all the information and still lexes properly.
2. I am not entirely sure how to fix the duplicate label warning for the mojo/pojo-quickstarts so any guidance would be greatly appreciated.


zainhaq-h2o (approved):
Looks good to me! Also, I didn't get a duplicate label warning when I built the docs locally.

PR #5473, opened 2021-05-04
Make sure testJavaScoring tests all rows, not just a subset [nocheck]

michalkurka:
Originally testJavaScoring was using a fraction of the dataset to test (10%), with this change it uses the full dataset. The reason for this change is that for small synthetic data (few rows) it didn’t actually test any rows.

With this change enabled I found 2 code issues:

- https://h2oai.atlassian.net/browse/PUBDEV-8263
- https://h2oai.atlassian.net/browse/PUBDEV-8264

### DIFF
--- a/h2o-algos/src/test/java/hex/tree/drf/DRFPredictContribsTest.java
+++ b/h2o-algos/src/test/java/hex/tree/drf/DRFPredictContribsTest.java
@@ -1,5 +1,6 @@
 package hex.tree.drf;
 
+import hex.Model;
 import hex.genmodel.algos.tree.SharedTreeNode;
 import hex.genmodel.algos.tree.SharedTreeSubgraph;
 import hex.genmodel.algos.tree.TreeSHAP;
@@ -180,8 +181,13 @@ public class DRFPredictContribsTest extends TestUtil {
                     .doAll(new byte[]{Vec.T_CAT, Vec.T_NUM, Vec.T_NUM}, contributions)
                     .outputFrame(null, new String[]{"predict", "p0", "p1"}, new String[][]{new String[]{"0", "1"}, null, null});
             Scope.track(predsFromContribs);
-            
-            assertTrue(drf.testJavaScoring(fr, predsFromContribs, 1e-5, 1e-7));
+
+            Model.JavaScoringOptions options = new Model.JavaScoringOptions();
+            if (!Boolean.getBoolean("reproduce.PUBDEV-8264")) { // FIXME - works only by chance - fails on full data
+                options._fraction = 0.1;
+            }
+            options._abs_epsilon = 1e-7;
+            assertTrue(drf.testJavaScoring(fr, predsFromContribs, 1e-5, options));
 
             // Now test MOJO scoring
             EasyPredictModelWrapper.Config cfg = new EasyPredictModelWrapper.Config()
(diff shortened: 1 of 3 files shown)


h2o-core/src/main/java/hex/Model.java
```diff
@@ -2609,20 +2609,37 @@ protected SBPrintStream toJavaTransform(SBPrintStream ccsb,
   // is well, false is there are any mismatches.  Throws if there is any error
   // (typically an AssertionError or unable to compile the POJO).
   public boolean testJavaScoring(Frame data, Frame model_predictions, double rel_epsilon) {
-    return testJavaScoring(data, model_predictions, rel_epsilon, 1e-15, 0.1);
```
valenad1, line 2612: I am wandering if `rel_epsilon` in `JavaScoringOptions ` would have any benefit. At least it would be consistent with all the other options if  `rel_epsilon ` would be in `JavaScoringOptions`.

h2o-algos/src/test/java/hex/tree/drf/DRFPredictContribsTest.java
```diff
@@ -180,8 +181,13 @@ public void testScoreContributionsBinomial() throws IOException, PredictExceptio
                     .doAll(new byte[]{Vec.T_CAT, Vec.T_NUM, Vec.T_NUM}, contributions)
                     .outputFrame(null, new String[]{"predict", "p0", "p1"}, new String[][]{new String[]{"0", "1"}, null, null});
             Scope.track(predsFromContribs);
-            
-            assertTrue(drf.testJavaScoring(fr, predsFromContribs, 1e-5, 1e-7));
+
+            Model.JavaScoringOptions options = new Model.JavaScoringOptions();
+            if (!Boolean.getBoolean("reproduce.PUBDEV-8264")) { // FIXME - works only by chance - fails on full data
+                options._fraction = 0.1;
```
valenad1, line 187: I don't understand why this test is not fail on my local nor in the pipeline. You made test more specific and don't switch of MOJO, so I would expect this test to fail deterministically.

valenad1 (approved):
Looks good to me, thanks! Just a couple of suggestions and question why `testScoreContributionsBinomial` isn't failing,

h2o-algos/src/test/java/hex/tree/drf/DRFPredictContribsTest.java
michalkurka: It doesn't fail in the pipeline because it doesn't test any affected rows actually.

h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
```diff
@@ -2732,43 +2732,49 @@ public void testStochasticGBMHoldout() {
...
+      Frame testPreds = gbm.score(test);
+      Scope.track(testPreds);
+      Log.info(test);
+      Log.info(testPreds);
+
+      Assert.assertTrue(gbm.testJavaScoring(train, trainPreds, 1e-15));
+      Model.JavaScoringOptions options = new Model.JavaScoringOptions();
+      options._disable_pojo = !Boolean.getBoolean("reproduce.PUBDEV-8263"); // FIXME - doesn't work unless POJO is disabled
+      Assert.assertTrue(gbm.testJavaScoring(test, testPreds, 1e-15, options));
+      Assert.assertTrue(Math.abs(trainPreds.vec(0).at(0) - -2.5) < 1e-6);
```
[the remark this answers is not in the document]
michalkurka: yep, not ideal to use assertTrue like that

h2o-core/src/main/java/hex/Model.java
michalkurka: rel_epsilon is the only mandatory parameter, the other can be optional

I think you don't want the test to be calling testJavaScoring without specifying/showing in code the precision it is working with - it is good when you see it explicitly

PR #5474, opened 2021-05-05
PUBDEV-8134: log python version before buildClientDist task to better…

koniecsveta:
… diagnose the problem

https://h2oai.atlassian.net/browse/PUBDEV-8134

after d84f6ae:
- setuptools is there during pythonVersion (added pip list there)
- its also there during next task verifyDependencies, because it gets verified by verify_requirements.py and it is listed in conda/h2o/meta.yaml
- added pip list log directly into buildClientDist before running the failing command

it will be something like this https://stackoverflow.com/questions/37341614/modules-are-installed-using-pip-on-osx-but-not-found-when-importing

because

when
ssh jenkins@mr-0xe2
docker run -it --rm -v /home/0xdiag:/home/0xdiag -v /home/jenkins/slave_dir_from_mr-0xc1/workspace/hadoop-multinode-pipeline_master:/app harbor.h2o.ai/opsh2oai/h2o-3-hadoop-hdp-2.2-0xd-edge:84 bash
source envs/h2o_env_python3.6/bin/activate

setuptools is in pip list result but

(h2o_env_python3.6) root@f1ac451dfed6:/envs# python
Python 3.6.12 (default, Aug 18 2020, 02:08:22) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from setuptools import setup, find_packages
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'setuptools'

and

(h2o_env_python3.6) root@f1ac451dfed6:/envs# which -a python
/envs/h2o_env_python3.6/bin/python
/usr/bin/python

and 

from setuptools import setup, find_packages

works when I do it outside of h2o_env_python3.6

### DIFF
--- a/docker/scripts/install_python_version
+++ b/docker/scripts/install_python_version
@@ -57,7 +57,6 @@ for python_version in "${array[@]}"; do
   su jenkins -c "virtualenv -p python${python_version} /envs/h2o_env_python${python_version}"
 
   echo "###### Installing dependencies for Python ${python_version} ######"
-  su jenkins -c ". /envs/h2o_env_python${python_version}/bin/activate && pip uninstall setuptools -y && pip install setuptools" # resinstalling setuptools fixes install issues for sphinx-osexample
   su jenkins -c ". /envs/h2o_env_python${python_version}/bin/activate && grep -v \"^#\" test-requirements.txt | sed -r 's/(.*)/\"\\1\"/' | xargs -n 1 -L 1 pip install"
   if [[ ${major} == "3" ]]; then
     su jenkins -c ". /envs/h2o_env_python${python_version}/bin/activate && grep -v \"^#\" docs-requirements.txt | sed -r 's/(.*)/\"\\1\"/' | xargs -n 1 -L 1 pip install"    


sebhrusen:
Maybe another improvement that we can do:
The docker scripts hierarchy is as follow:
dev-base < dev-jdk-8-base < dev-python-base < dev-python-3.6 < dev-r-base < dev-r-3.4.1 < dev-build-base < dev-build-hadoop

so, when we're building the hadoop image, we should already have the venv for Py 3.6 ready for use, but not Py 2.7.

For that reason, in the Hadoop docker scripts (HDP or CDH), we run:
```bash
# Set required env vars and install Pythons
RUN \
  chmod 700 /usr/sbin/install_python_version && \
  sync && \
  /usr/sbin/install_python_version && \
  /usr/bin/activate_java_8
```
to install Py2.7 AND Py3.6, therefore executing some commands on the Py3.6 venv inherited from the parent image.

I think we should avoid that, and I see 2 possible fixes for `install-python-version`:
1. delete a Py venv if it exists before (re)creating it.
2. do not (re)create a venv if it already exists.

wdyt?

koniecsveta:
> Maybe another improvement that we can do:
> The docker scripts hierarchy is as follow:
> dev-base < dev-jdk-8-base < dev-python-base < dev-python-3.6 < dev-r-base < dev-r-3.4.1 < dev-build-base < dev-build-hadoop
> 
> so, when we're building the hadoop image, we should already have the venv for Py 3.6 ready for use, but not Py 2.7.
> 
> For that reason, in the Hadoop docker scripts (HDP or CDH), we run:
> 
> ```shell
> # Set required env vars and install Pythons
> RUN \
>   chmod 700 /usr/sbin/install_python_version && \
>   sync && \
>   /usr/sbin/install_python_version && \
>   /usr/bin/activate_java_8
> ```
> 
> to install Py2.7 AND Py3.6, therefore executing some commands on the Py3.6 venv inherited from the parent image.
> 
> I think we should avoid that, and I see 2 possible fixes for `install-python-version`:
> 
> 1. delete a Py venv if it exists before (re)creating it.
> 2. do not (re)create a venv if it already exists.
> 
> wdyt?

I would do 2., @michalkurka wdyt?

PR #5477, opened 2021-05-10
Add nightly node labels to nightly stages

hasithjp:
**Introduction**
This new label called  `nightly` is added to limit the number of nodes that can be used for the parallel stages. Because during PST night time, almost all the nodes in Jenkins get occupied by these nightly parallel runs by blocking the other lightweight jobs of devs which will cause a long waiting time for them. Now the nightly parallel stages will be limited only to 30 nodes (`mr-0x11 to mr-0x40`)  out of 47 mr-0x# nodes


hasithjp:
Closing this PR since the new branch is created from the master branch which is not the usual flow. 
Please refer to the correct PR here: https://github.com/h2oai/h2o-3/pull/5478

PR #5478, opened 2021-05-10
Add new nightly node label for nightly stages

hasithjp:
**Introduction**
This new label called  `nightly` is added to limit the number of nodes that can be used for the parallel stages. Because during PST night time, almost all the nodes in Jenkins get occupied by these nightly parallel runs by blocking the other lightweight jobs of devs which will cause a long waiting time for them. Now the nightly parallel stages will be limited only to 30 nodes (`mr-0x11 to mr-0x40`)  out of 47 mr-0x# nodes. We can add the new label to other nodes as well based on requirements

### DIFF
--- a/scripts/jenkins/groovy/buildConfig.groovy
+++ b/scripts/jenkins/groovy/buildConfig.groovy
@@ -143,6 +143,10 @@ class BuildConfig {
     return nodeLabels.getGPUNodeLabel()
   }
 
+  String getNightlyNodeLabel() {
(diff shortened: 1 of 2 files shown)


surenH2oai:
@hasithjp Lets not change h2o-3 pipeline.  We need to free a small set of 2 to 3 machines from DAI pool for PR/smoke builds.

surenH2oai:
please cancel this PR @hasithjp , lets do as we are discussing on the slack thead.

PR #5483, opened 2021-05-11
PUBDEV-8142 - Fix sort util method and enhance tests

valenad1:
There is a bug in ArrayUtils.sort() → increasing attribute is ignored in the insertion sort, indices in test are always array of zeros.


h2o-core/src/main/java/water/util/ArrayUtils.java
```diff
@@ -1548,8 +1548,7 @@ public static void sort(final int[] idxs, final double[] values, int cutoff, int
       Arrays.sort(d, new Comparator<Integer>() {
         @Override
         public int compare(Integer x, Integer y) {
-          return values[x]*increasing < values[y]*increasing ? -1 : 
-                  (values[x]*increasing > values[y]*increasing ? 1 : 0);
+          return Double.compare(values[x] * increasing, values[y] * increasing);
```
michalkurka, line 1551: technically this is changing the behavior - because Double.compare handles also special cases like NaN, -NaN and -0

michalkurka: I would probably leave as-is in this PR, unless you have a reason to change

valenad1: Aaa ok, I'll remove the change, the change is only because of the code style

valenad1: Removed

michalkurka (approved):

PR #5484, opened 2021-05-11
ITA-945 adjust benchmarks training time expected intervals

koniecsveta:
https://h2oai.atlassian.net/browse/ITA-945


michalkurka (approved):
Thank you @koniecsveta

PR #5486, opened 2021-05-11
Updated expected ranges of GAM benchmark results

michalkurka:


michalkurka:
Screen Shot 2021-05-11 at 2 16 22 PM

@wendycwong updated results according to last months runs

PR #5491, opened 2021-05-13
PUBDEV-8084 - Connect XGboost aggregation functionality with sorting functionality

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-8084


h2o-core/src/main/java/hex/Model.java
```diff
@@ -3189,4 +3190,39 @@ protected boolean isFeatureUsedInPredict(int featureIdx) {
   public boolean isDistributionHuber() {
     return _parms._distribution == DistributionFamily.huber;
   }
+
+  public static void composeScoreContritbionTaskMetadata(final String[] names, final byte[] types, final String[][] domains, final String[] originalFrameNames, final Contributions.ContributionsOptions options) {
```
michalkurka, line 3194: there is a typo, 

since this is a static method (and thus not using anything from Model as such) - can it instead be defined in a helper class?

valenad1: Thanks, I was about to implement it in the ContributionComposer but I need to have access to algos package. This place look right to me since for example adaptTestForTrain(..). 

It doesn't need to be static.

If we want to have it separately, then we need to have a new class for it

valenad1: It can be default implementation of Model.Contributions though

michalkurka (approved):
Thank you! Looks great

PR #5492, opened 2021-05-14
PUBDEV-8145: Fix R package based on CRAN feedback

ledell:
https://h2oai.atlassian.net/browse/PUBDEV-8145

### DIFF
--- a/h2o-r/h2o-package/R/connection.R
+++ b/h2o-r/h2o-package/R/connection.R
@@ -485,23 +485,7 @@ h2o.resume <- function(recovery_dir=NULL) {
 .h2o.pkg.path <- NULL
 .h2o.jar.env <- new.env()    # Dummy variable used to shutdown H2O when R exits
 
-.onLoad <- function(lib, pkg) {
-  .h2o.pkg.path <<- file.path(lib, pkg)
-
-  # installing RCurl requires curl and curl-config, which is typically separately installed
-  rcurl_package_is_installed = length(find.package("RCurl", quiet = TRUE)) > 0L
-  if(!rcurl_package_is_installed) {
-    if(.Platform$OS.type == "unix") {
-      # packageStartupMessage("Checking libcurl version...")
-      curl_path <- Sys.which("curl-config")
-      if(!nzchar(curl_path[[1L]]) || system2(curl_path, args = "--version") != 0L)
-        stop("libcurl not found. Please install libcurl\n",
-             "(version 7.14.0 or higher) from http://curl.haxx.se.\n",
-             "On Linux systems you will often have to explicitly install\n",
-             "libcurl-devel to have the header files and the libcurl library.")
-    }
-  }
-}
+# Note: Moved .onLoad() to zzz.R
 
 .onAttach <- function(libname, pkgname) {
   msg = paste0(
--- a/h2o-r/h2o-package/R/connection.R
+++ b/h2o-r/h2o-package/R/connection.R
@@ -826,12 +826,8 @@ h2o.resume <- function(recovery_dir=NULL) {
     # Get MD5 checksum
     md5_url <- paste("https:/", base_url, "h2o.jar.md5", sep = "/")
   }
-  # ttt <- getURLContent(md5_url, binary = FALSE)
-  # tcon <- textConnection(ttt)
-  # md5_check <- readLines(tcon, n = 1)
-  # close(tcon)
   md5_file <- tempfile(fileext = ".md5")
-  download.file(md5_url, destfile = md5_file, mode = "w", cacheOK = FALSE, quiet = TRUE)
+  download.file(url = md5_url, destfile = md5_file, mode = "w", cacheOK = FALSE, quiet = TRUE, timeout = max(1000, getOption("timeout")))
   md5_check <- readLines(md5_file, n = 1L)
   if (nchar(md5_check) != 32) stop("md5 malformed, must be 32 characters (see ", md5_url, ")")
   unlink(md5_file)
@@ -841,7 +837,7 @@ h2o.resume <- function(recovery_dir=NULL) {
   cat("Performing one-time download of h2o.jar from\n")
   cat("    ", h2o_url, "\n")
   cat("(This could take a few minutes, please be patient...)\n")
-  download.file(url = h2o_url, destfile = temp_file, mode = "wb", cacheOK = FALSE, quiet = TRUE)
+  download.file(url = h2o_url, destfile = temp_file, mode = "wb", cacheOK = FALSE, quiet = TRUE, timeout = max(1000, getOption("timeout")))
 
   # Apply sanity checks
   if(!file.exists(temp_file))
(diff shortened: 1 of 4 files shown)


tomasfryda (approved):
LGTM! Thank you @ledell!

michalkurka (changes requested):
download.file doesn't take argument 'timeout'

verified on `R version 3.6.3`

h2o-r/h2o-package/R/connection.R
```diff
@@ -841,7 +838,8 @@ h2o.resume <- function(recovery_dir=NULL) {
   cat("Performing one-time download of h2o.jar from\n")
   cat("    ", h2o_url, "\n")
   cat("(This could take a few minutes, please be patient...)\n")
-  download.file(url = h2o_url, destfile = temp_file, mode = "wb", cacheOK = FALSE, quiet = TRUE)
+  flush.console()
+  download.file(url = h2o_url, destfile = temp_file, mode = "wb", cacheOK = FALSE, quiet = TRUE, timeout = max(1000, getOption("timeout")))
```
michalkurka, line 842: timeout is in seconds, do we really want the user to wait up to 1000 seconds for everything to fail?

ledell: @michalkurka That was the exact timeout duration that CRAN suggested, based on the size of the h2o.jar, so I think that's probably what we should leave it as.

ledell:
Here's what I changed: 
- Set timeout as an option in `.onLoad()`:  `options(timeout = max(1000, getOption("timeout")))`
- Consolidated the two copies of `.onLoad()` (only one was being used, this was a bug).
- Moved the `.onLoad()` into  `zzz.R`, which is the standard location for these functions.  I left `.onAttach()` et al in `connection.R` because there were some variables which were being used in both `.onAttach()` and the other connection.R internal functions and it seemed slightly risky to move.

h2o-r/h2o-package/R/zzz.R
```diff
@@ -1 +1,61 @@
...
+  
+  invisible()
+}
+
+
+# This was moved from connection.R; .onAttach, .onDetach, .Last are still in connection.R
+.onLoad <- function(lib, pkg) {
+  
+  # for file.download, used to download h2o.jar
+  options(timeout = max(1000, getOption("timeout")))
```
tomasfryda, line 34: I'd suggest to use the same pattern as is often used for `par`:

```r
old_options <- options(timeout = max(1000, getOption("timeout")))  # NOTE: old_options is a named list containing only the old values of the changed options
on.exit(options(old_options))
```

This way the timeout is set only within this function scope so we'd need to move the `.h2o.downloadJar()` inside this function. 

I think it shouldn't cause `.h2o.downloadJar()` to be ran more times than it used to but I am not sure; my assumptions are:
>Now it is called on every `source` which would be just once per session and this would change to be called just once per load which should also be just once per session (both statements should be valid unless a user decides to unload the pkg and reload it which should be relatively rare).

jangorecki: I think it is better to move setting option to `.h2o.downloadJar`

ledell: @tomasfryda Are you saying that you want to change things around so that the timeout only applies to the `.h2o.downloadJar()` function?  

@jangorecki Why do you suggest that we move the option setting to `.h2o.downloadJar`?  The same reason above (so it only applies to one function)?

ledell: Here's the current behavior:

```
> library(h2o)

----------------------------------------------------------------------

Your next step is to start H2O:
    > h2o.init()

For H2O package documentation, ask for help:
    > ??h2o

After starting H2O, you can use the Web UI at http://localhost:54321
For more information visit https://docs.h2o.ai

----------------------------------------------------------------------


Attaching package: ‘h2o’

The following objects are masked from ‘package:stats’:

    cor, sd, var

The following objects are masked from ‘package:base’:

    &&, %*%, %in%, ||, apply, as.factor, as.numeric, colnames,
    colnames<-, ifelse, is.character, is.factor, is.numeric, log,
    log10, log1p, log2, round, signif, trunc

> getOption("timeout")
[1] 1000
> detach("package:h2o", unload=TRUE)
[1] "A shutdown has been triggered. "
> getOption("timeout")
[1] 1000
```
It will go back to 60 seconds only with a new R session.

jangorecki: Yes, so current behavior can be improved

tomasfryda: @ledell yes, I would limit side effects like this to the smallest possible extent so we don't affect anything else, e.g., some other library the same session.

michalkurka:
@ledell to pass the "Check Pull Request Metadata" please change any Java file (eg. docuchange, anything)

currently there is a bug that doesn't let you pass the check if you only make a change in R files

h2o-r/h2o-package/R/connection.R
```diff
@@ -758,7 +742,10 @@ h2o.resume <- function(recovery_dir=NULL) {
 # or it calls stop() and does not return.
 #
 # It will download a jar file if it needs to.
+#' @importFrom utils flush.console
 .h2o.downloadJar <- function(overwrite = FALSE) {
+  old_options <- options(timeout = max(1000, getOption("timeout")))
+  on.exit(options(old_options))
```
ledell, line 748: @tomasfryda  Can you explain what this does?  To me, it looks like it will save the "new" timeout option of 1000s (but you call it `old_options`).  And then it sets the timeout to 1000s only when the `.h2o.downloadJar()` exits (which would be after the download instead of before it).  It seems like it does the opposite of what we want it to do?  I might be missing something though.

Would we do something like this instead?
```
old_options <- options(timeout = getOption("timeout")) #save what the user had
options(timeout = max(1000, getOption("timeout"))) #set the 1000s for the download
on.exit(options(old_options)) #return to the previous settings
  ```

ledell: cc @jangorecki Can you confirm?

tomasfryda: @ledell From the options documentation(https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/options)

>For options(), a list of all set options sorted by name. For options(name), a list of length one containing the set value, or NULL if it is unset. **For uses setting one or more options, a list with the previous values of the options changed (returned invisibly).**

ledell: @tomasfryda Ok, that's not what I expected!  Thanks for the reference.  

We can ignore my comments above -- the code does what we want it to.  So I think we are good to merge!

ledell:
@michalkurka I edited Java file, per your request.  Changing the Java file re-ran all the tests, but they were previously green, so I merged before the tests re-ran: http://mr-0xg1:8080/blue/organizations/jenkins/h2o-3-pipeline/detail/PR-5492/3/pipeline

PR #5493, opened 2021-05-14
PUBDEV-7937 - rename parameters

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-7937

merge after: https://github.com/h2oai/h2o-3/pull/5494

### DIFF
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -173,7 +173,7 @@ class ModelBase(h2o_meta(Keyed)):
                     data={"predict_staged_proba": True})
         return h2o.get_frame(j["predictions_frame"]["name"])
 
-    def predict_contributions(self, test_data, output_format="Original", top_n=None, top_bottom_n=None, abs_val=False):
+    def predict_contributions(self, test_data, output_format="Original", top_n=None, bottom_n=None, abs_val=False):
         """
         Predict feature contributions - SHAP values on an H2O Model (only GBM, XGBoost and DRF models).
         
@@ -190,8 +190,8 @@ class ModelBase(h2o_meta(Keyed)):
             contributions for 1-hot encoded features, specifying a Compact output format will produce a per-feature
             contribution. One of: ``"Original"``, ``"Compact"`` (default: ``"Original"``).
         :param top_n: Return only #top_n highest contributions + bias.
-        :param top_bottom_n: Return only #top_bottom_n lowest contributions + bias
-                             If top_n and top_bottom_n are defined together then return array of #top_n + #top_bottom_n + bias
+        :param bottom_n: Return only #bottom_n lowest contributions + bias
+                             If top_n and bottom_n are defined together then return array of #top_n + #bottom_n + bias
         :param abs_val: True to compare absolute values of contributions
         :returns: A new H2OFrame made of feature contributions.
         """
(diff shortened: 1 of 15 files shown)


h2o-py/h2o/model/model_base.py
```diff
@@ -173,7 +173,7 @@ class c there will be a column Tt.Cc (eg. T3.C1 for tree 3 and class 1). The val
                     data={"predict_staged_proba": True})
         return h2o.get_frame(j["predictions_frame"]["name"])
 
-    def predict_contributions(self, test_data, output_format="Original", top_n=None, top_bottom_n=None, abs_val=False):
+    def predict_contributions(self, test_data, output_format="Original", top_n=None, bottom_n=None, abs=False):
```
valenad1, line 176: Hi guys, do you think it is a problem to shadow python abs() function by naming the parameter abs? AFAIK it should have effect only on the current function.

@michalkurka @sebhrusen

sebhrusen: yes, shadowing in limited scope is usually fine if you think that there's no use for the shadowed function there.
Maybe you could rename to `absolute` otherwise?

Just a suggestion:  I don't know if we're ok to break the client API in a non backwards-compatible way?
For python, you could add a decorator to keep backwards compatibility.
In the same way as we have the `@deprecated_property` decorator, we could create sth like the following:
```
@deprecated_params(dict(
    top_bottom_n=bottom_n,
    abs_val=abs
))
def predict_contributions(self, test_data, output_format="Original", top_n=None, bottom_n=None, abs=False):
```
where the keys of the dictionary are the deprecated names and the values the new ones.
wdyt?

michalkurka: > Just a suggestion: I don't know if we're ok to break the client API in a non backwards-compatible way
@sebhrusen this was not released yet, we are still free to change as we find fit

Regarding naming of "abs", I think it would be appropriate to include a verb in the option name. Eg.: "compare_abs" or "abs_compare" - to show what it does. We need to keep in mind that "abs" is only a secondary level parameter - it doesn't affect what predict_contributions return.

valenad1: Regarding the suggestion: I think it is not necessary to have it here. The current API is model.predict_contributions(self, test_data) is working the same was as before

valenad1: I'll make it compare_abs 👍

h2o-py/h2o/model/model_base.py
```diff
@@ -190,9 +190,9 @@ def predict_contributions(self, test_data, output_format="Original", top_n=None,
             contributions for 1-hot encoded features, specifying a Compact output format will produce a per-feature
             contribution. One of: ``"Original"``, ``"Compact"`` (default: ``"Original"``).
         :param top_n: Return only #top_n highest contributions + bias.
-        :param top_bottom_n: Return only #top_bottom_n lowest contributions + bias
-                             If top_n and top_bottom_n are defined together then return array of #top_n + #top_bottom_n + bias
-        :param abs_val: True to compare absolute values of contributions
+        :param bottom_n: Return only #bottom_n lowest contributions + bias
+                             If top_n and bottom_n are defined together then return array of #top_n + #bottom_n + bias
```
sebhrusen, line 194: many tests are using values like `top_n=-1` or `bottom_n=0`: could be document here the semantic of those particular values?
I think I understand `0`, although I don't see why it is used in tests: probably has the same semantic as `None`, in which case it's not necesseray to document that one, but what about `-1`?

michalkurka (approved):

sebhrusen (approved):
Thanks @valenad1 !

h2o-py/h2o/model/model_base.py
valenad1: Good point! Thanks, I'll improve the documentation for Python and R

PR #5494, opened 2021-05-14
PUBDEV-7937 - client mode test fix

valenad1:


h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
```diff
@@ -190,11 +190,15 @@ protected void addContribToNewChunk(float[] contribs, NewChunk[] nc) {
 
   public class ScoreContributionsSortingTask extends ScoreContributionsTask {
 
-    private transient ContributionsOptions _options;
+    private int _topN;
+    private int _bottomN;
+    private boolean _abs;
 
-    public ScoreContributionsSortingTask(SharedTreeModel model, ContributionsOptions options) {
+    public ScoreContributionsSortingTask(SharedTreeModel model, int topN, int bottomN, boolean abs) {
```
michalkurka, line 197: you can still accept ContributionsOptions in the constructor, just extract the fields you need here

this is a better design because of better encapsulation of logic in the class where it belongs

michalkurka:
usually when bug is encountered it would be also good to add a test that makes sure it cannot happen again, in this case it seems like we don't have enough tests that would be running in a truly distributed mode (otherwise this issue would be caught in PR as well)

h2o-algos/src/main/java/hex/tree/SharedTreeModelWithContributions.java
valenad1: True, thanks

valenad1:
> usually when bug is encountered it would be also good to add a test that makes sure it cannot happen again, in this case it seems like we don't have enough tests that would be running in a truly distributed mode (otherwise this issue would be caught in PR as well)

Thanks for the point, I have added the large test which fail in old version (two nodes) and success with current changes.

michalkurka (approved):

PR #5202, opened 2020-12-23
PUBDEV-7139 Variable importance for black-box models

ardKelmendi:
Creating PR for Permutation Feature Importance

JIRA: https://h2oai.atlassian.net/jira/software/c/projects/PUBDEV/issues/PUBDEV-7139

PR #5495, opened 2021-05-14
PUBDEV-7139: Variable importance for black-box models

tomasfryda:
Creating PR for Permutation Feature Importance

JIRA: https://h2oai.atlassian.net/jira/software/c/projects/PUBDEV/issues/PUBDEV-7139

Created this since I was unable to push into ardKelmendi's fork PR: https://github.com/h2oai/h2o-3/pull/5202

-------
### API

API for `model.permutation_importance(...)`  / `h2o.permutation_importance(...)`:
This produces variable importance table (H2OTwoDimTable).
```
python/R (if the name differs between clients)
self/object     A trained supervised H2O model.
frame/newdata   Training frame of the model which is going to be permuted
metric          Metric to be used. One of "AUTO", "AUC", "MAE", "MSE", "RMSE", "logloss", "mean_per_class_error",
                    "PR_AUC".  Defaults to "AUTO".
n_samples       Number of samples to be evaluated. Use -1 to use the whole dataset. Defaults to 10 000.
n_repeats       Number of repeated evaluations. Defaults to 1.
features        Character vector of features to include in the permutation importance. Use NULL to include all.
seed            Seed for the random generator. Use -1 to pick a random seed. Defaults to -1.
use_pandas/     Set true to return pandas data frame.

Returns a H2OTwoDimTable.
```
API for `model.permutation_importance_plot(...)`  / `h2o.permutation_importance_plot(...)`:
This produces plot and returns data.
```
python/R (if the name differs between clients)
self/object     A trained supervised H2O model.
frame/newdata   Training frame of the model which is going to be permuted
metric          Metric to be used. One of "AUTO", "AUC", "MAE", "MSE", "RMSE", "logloss", "mean_per_class_error",
                    "PR_AUC".  Defaults to "AUTO".
n_samples       Number of samples to be evaluated. Use -1 to use the whole dataset. Defaults to 10 000.
n_repeats       Number of repeated evaluations. Defaults to 1.
features        Character vector of features to include in the permutation importance. Use NULL to include all.
seed            Seed for the random generator. Use -1 to pick a random seed. Defaults to -1.
num_of_features The number of features shown in the plot (default is 10 or all if less than 10).
server/         If true set server settings to matplotlib and do not show the plot

Returns a H2OTwoDimTable. 
```


tomasfryda:
I still have to:

1. add more tests and polish the existing ones
2. polish the documentation and doc-strings

Example: https://github.com/h2oai/h2o-3/blob/6c64f162cae0f239be50c797289a777a6986daf6/h2o-py/demos/permutation_var_imp.ipynb

h2o-core/src/main/java/water/rapids/PermutationVarImp.java
```diff
@@ -0,0 +1,216 @@
...
+            if (null != fr && fr != _inputFrame) fr.remove();
+            if (null != shuffledFeature) shuffledFeature.remove();
+        }
+        return result;
+    }
+
+    /**
+     * Get PermutationVarImp
+     *
+     * @param metric    Metric to use to calculate the feature importance
```
maurever, line 138: To be consistent...

```suggestion
     * @param metric    Metric to use to calculate the variable (feature) importance
```

h2o-core/src/main/java/water/rapids/PermutationVarImp.java
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        // Create TwoDimTable having (Relative + Scaled + percentage) importance 
+        return ModelMetrics.calcVarImp(importance, names);
+    }
+
+    /**
+     * Get Repeated Permutation Variable Importance
+     *
+     * @param metric    Metric to use to calculate the feature importance
```
maurever, line 165:
```suggestion
     * @param metric    Metric to use to calculate the variable (feature) importance
```

or 

```suggestion
     * @param metric    Metric to use to calculate the variable importance
```

h2o-core/src/main/java/water/rapids/PermutationVarImp.java
```diff
@@ -0,0 +1,216 @@
...
+    public TwoDimTable getRepeatedPermutationVarImp(String metric, final long n_samples, final int n_repeats ,final String[] features, long seed) {
+        metric = inferAndValidateMetric(metric);
+
+        HashMap<String, Double>[] varImps =  new HashMap[n_repeats];
+        for (int i = 0; i < n_repeats; i++) {
+            varImps[i] = calculatePermutationVarImp(metric, n_samples, features, (seed == -1 ? -1 : seed + i));
+        }
+
+        String[] names = new String[varImps[0].size()];
+        double[/*r*/][/*c*/] importance = new double[varImps[0].size()][n_repeats];
```
maurever, line 181: This could not be very clear. I will rather use whole words if the comments are necessary. 

```suggestion
        double[/*repeat*/][/*colums*/] importance = new double[varImps[0].size()][n_repeats];
```

h2o-core/src/main/java/water/rapids/ast/prims/models/AstPermutationVarImp.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Override
+    public String str()     { return "PermutationVarImp"; }
+
+    @Override
+    public ValFrame apply(Env env, Env.StackHelp stk, AstRoot asts[]) {
+        Model model = stk.track(asts[1].exec(env)).getModel();
+        Frame fr = stk.track(asts[2].exec(env)).getFrame();
+        String metric = stk.track(asts[3].exec(env)).getStr().toLowerCase();
+        long n_samples = (long) stk.track(asts[4].exec(env)).getNum();
```
maurever, line 37: Do you validate the parameters somewhere? For example, if `n_samples`  is higher than 1 and lower than dataset size? Or if the `n_repeats` is not negative. Or if the input features were used for model training...

maurever (changes requested):

h2o-core/src/main/java/water/rapids/PermutationVarImp.java
```diff
@@ -0,0 +1,216 @@
...
+
+        final String[] variables = _inputFrame.names();
+
+        HashSet<String> featuresToCompute = new HashSet<>(Arrays.asList((null != features && features.length > 0) ? features : variables));
+        featuresToCompute.removeAll(Arrays.asList(_model._parms.getNonPredictors()));
+        if (_model._parms._ignored_columns != null)
+            featuresToCompute.removeAll(Arrays.asList(_model._parms._ignored_columns));
+
+        Frame fr = null;
+        if (n_samples > 1) {
```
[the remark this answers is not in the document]
tomasfryda: By useless sampling you mean this?
```{java}
             while (fr.numRows() < 2) {
                 fr.remove();
                 seed += 1;
                 n_samples *= 2;
                 Log.warn("Sampled less than 2 rows, repeating the sampling with seed = "+ seed +" and increasing n_samples to = "+ n_samples +".");
                 fr = MRUtils.sampleFrame(_inputFrame, n_samples, _model._parms._weights_column, seed);
             }
```             

This is due to the fact that the sampler calculates the fraction of the dataset that should be sampled and randomly decides if a given row will be used or not, thus it is possible (but unlikely) that even if you set `n_samples > 2` you will end up with 1.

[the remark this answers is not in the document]
maurever: Oh, I see. Ok, thank you for your explanation. In this case, it is ok.

h2o-core/src/main/java/water/rapids/ast/prims/models/AstPermutationVarImp.java
tomasfryda: Good point, thank you! I completely missed that.

h2o-algos/src/test/java/hex/varimp/PermutationVarImpTest.java
```diff
@@ -0,0 +1,355 @@
...
+        }
+    }
+
+    /**
+     * Test Permutation Variable Importance together with GLM model and 
+     * check whether the PVI values ordered are similar to the 
+     * coefficients from the GLM. Allows +-3 elements to be in different oreder
+     */
+    @Test
+    public void testPermVarImpGLM(){
```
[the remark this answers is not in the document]
tomasfryda: I'd call that a basic sanity check.

Generally, it depends on the way we compute varimp and of course both approaches will be biased in some way (PVI compares effect on the response on potentially impossible points due to the permutation). In general, I'd avoid that but here when we have some fixed dataset and algo it's basically the same thing as comparing the PVI with a fixed list of variable importance entries.

h2o-algos/src/test/java/hex/varimp/PermutationVarImpTest.java
```diff
@@ -0,0 +1,355 @@
...
+            fr.delete();
+            betaConstraints.delete();
+            if (model != null) model.delete();
+        }
+    }
+
+    /**
+     * Sorting a map by value.
+     */
+    public static class MapUtil {
```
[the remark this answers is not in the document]
tomasfryda: Renamed it but I don't know where else to put it. Did you mean like reorder the declaration within the test or move it outside of the test file?

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -863,4 +865,53 @@ public void map(Chunk c, NewChunk nc) {
...
+   * https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle
+   */
+  public static class ShuffleVecTask extends MRTask<ShuffleVecTask> {
+    private final long _seed;
+    public ShuffleVecTask(final long seed) {
+      super();
+      _seed = seed;
+    }
+
+    @Override public void map(Chunk cs, Chunk ncs) {
```
michalkurka, line 879: @tomasfryda this means we never cross the chunk boundary, this could be an issue, wdyt?

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -863,4 +865,53 @@ public void map(Chunk c, NewChunk nc) {
...
+  }
+
+  /**
+   * Randomly shuffle a Vec. 
+   * @param iVec original Vec
+   * @param srcVec a copy of original Vec, to be shuffled
+   * @param seed seed for random generator
+   * @return shuffled Vec
+   */
+  public static Vec ShuffleVec(Vec iVec, Vec srcVec, final long seed) {
```
michalkurka, line 913: method name shouldn't start with a capital letter

michalkurka:
@tomasfryda it doesn't seem like this PR calls `adaptTestToFrame` anywhere

h2o-core/src/main/java/water/rapids/PermutationVarImp.java
```diff
@@ -0,0 +1,217 @@
...
+            fr = MRUtils.sampleFrame(_inputFrame, n_samples, _model._parms._weights_column, seed);
+            while (fr.numRows() < 2) {
```
michalkurka, line 92: hmm, you could just generate n row indices and extract those from the frame using Frame#deepSlice

so for low numbers you could do deepSlice and for big data use the approximate approach

tomasfryda: Good idea, thanks! Is the threshold for low numbers 1000 ok? (`n_samples < 1000` use deepSlice otherwise the sampleFrame)

h2o-core/src/main/java/water/util/VecUtils.java
tomasfryda: Good catch! Indeed, this could cause some issues, e.g., it could make sorted features less important.

Some papers recommend just splitting the dataset in half and swapping those halves, I wonder if we could do something similar - something like subclass `Vec` to `PermutedVec` and override `chunkForChunkIdx` so it permutes local chunks (so we don't have to send chunks between nodes). Or do you know of some better way?

h2o-core/src/main/java/water/rapids/PermutationVarImp.java
```diff
@@ -0,0 +1,217 @@
...
+ * and Morris method; creating TwoDimTable with relative, scaled, and percentage value
+ * TwoDimTable with mean of the absolute value, and standard deviation of all features importance
+ */
+
+public class PermutationVarImp {
+    private final Model _model;
+    private final Frame _inputFrame;
+
+    /**
+     * Constructor that stores the model, frame, response column, variable Strings and
```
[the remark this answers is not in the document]
tomasfryda: Thank you, I will add that.

h2o-algos/src/test/java/hex/varimp/PermutationVarImpTest.java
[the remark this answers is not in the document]
maurever: Ok, thanks, @tomasfryda.

h2o-algos/src/test/java/hex/varimp/PermutationVarImpTest.java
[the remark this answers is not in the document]
maurever: I was thinking of moving it outside of the test file to some utils class, such as TestUtils or something like that. But I am not sure if it is necessary here...

maurever (approved):
Thanks, @tomasfryda, you implemented all my suggestions. It looks good to me now.

h2o-core/src/main/java/water/util/VecUtils.java
michalkurka: I think the biggest issue with the current method will be when you have a big dataset and sample it to a low number. Then some chunks can end up having just a single row and no shuffling would be done.

So I think something will need to be changed.

One thing you could do is to randomly shuffle the (whole) chunks (across the cluster). Eg. send chunk 1 to node X, chunk 2 to node Y...

Then in your procedure you would pick a random chunk from a local node and get a random value from that chunk.

While moving data from node to node is expensive, the more expensive part would actually be scoring per-feature. And the chunk shuffle can be done on the background while feature > 1 is being scored (for the first feature you would have to do it first obviously).

tomasfryda: > Then some chunks can end up having just a single row and no shuffling would be done.

Thanks for pointing this out. Now I have something that should mitigate that issue - basically:

1. In `setupLocal`: permute the array of local chunk indices
2. In `map`: Get the chunkId of the current chunk and using the permuted array of local chunks get another local chunk

This should make the behavior you described possible only if the value of `n_samples` is set too low (to the number of chunks and at the same time there would have to be only one chunk per node) which on the other hand  shouldn't matter since for those situations (`n_samples < 1000`) I use randomly selected rows using the `deepSlice` that you suggested so the user would have to have >1000 nodes and select the same or lower number of samples.

I think this implementation would be sufficient, however, I am also looking into the "shuffle chunks across the nodes" as that would provide us with "statistically nicer" permutation but I am unsure if it would provide any significant benefit as PFI most likely will not be run that many times in a row on one dataset to notice any issues with the statistical properties of this kind of permutation.
 
For now, I have locally implemented a version where I permute the chunks across the nodes and I also permute the content of each chunk (but only within the chunks boundary). But it is important to note that the main point is to break the ties between the rows and the rows of the permuted column which both implementations do.

Another thing I can do is to use your suggestion but I haven't come up with an efficient solution that would ensure we would really end up with permutation as it would be closer to a bootstrap sample (this wouldn't be a problem if the user decides to sample the data).

The next steps can be:
1. use the current implementation (permute local chunks and data within the individual chunks)
2. permute chunks across all the nodes and permute data within the individual chunks
3. use hierarchical random sampling you proposed


Personally, I'd go with (1) as it is the fastest ( (2) seem to be ~30% slower on my test-case on local 4 node cluster) and I think it is good enough (provides enough random permutations if all cases that I could think of, I often seen PFI used several times to get better sense of how confident we can be but the number of runs was somewhere between 0 and 50).

What do you think, @michalkurka ?

michalkurka: @tomasfryda thanks for the analysis

I think it makes sense to go for (1) - permute local chunks and data within the individual chunks

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -879,4 +882,93 @@ public void map(Chunk c, NewChunk nc) {
...
+        arr[j] = old;
+      }
+    }
+
+    @Override
+    public void map(Chunk _cs, NewChunk nc) {
+      Random rng = getRNG(_seed + _cs.start());
+      Chunk cs = _vec.chunkForChunkIdx(_permutatedChunks[Arrays.binarySearch(_localChunks, _cs.cidx())]);
+
+      int[] permutedRows = IntStream.range(0, cs._len).toArray();
```
michalkurka, line 924: probably doesn't matter but typically we don't use stream API inside MRTasks

h2o-core/src/main/java/water/util/VecUtils.java
```diff
@@ -879,4 +882,93 @@ public void map(Chunk c, NewChunk nc) {
...
+
+      for (int row : permutedRows) {
+        if (cs.isNA(row)) {
+          nc.addNA();
+        } else {
+          switch (_cs.vec().get_type()) {
+            case Vec.T_BAD:
+              break;
+            case Vec.T_UUID:
+              nc.addUUID(cs, row);
```
michalkurka, line 935: random access to chunk elements is expensive, it might be cheaper to extract to a dense array - shuffle and write to nc

- just a thought

michalkurka (approved):
Looks very good, thank you @tomasfryda 

please see minor comment(s)

h2o-core/src/main/java/water/util/VecUtils.java
tomasfryda: Thank you for the comment, I'll change that. Is there any technical reason for that? I can think of only being it too easy to do parallel stream operations within a task that should be single threaded.

h2o-core/src/main/java/water/util/VecUtils.java
tomasfryda: Thank you, I briefly looked in to it and it seems to me that it would look something like the following pseudo-code :
```java
if (c.hasNA()) {
  for (int r=0; r < c._len; r++)
    if (!c.isNA(r)) arr[r] = c.at*(r);
} else {
    for (int r=0; r < c._len; r++)
      arr[r] = c.at*(r);
}

arr_indices = 0 ... arr.length;
permute(arr_indices, seed);
permute(arr, seed);

if (c.hasNA()) {
  for (int r=0; r < c._len; r++)
    if (c.isNA(arr_indices[r])) 
      nc.addNA();
    else
      nc.add*(arr[r]);
} else {
  for (int r=0; r < c._len; r++)
    nc.add*(arr[r]);
}
```

If that's so I think I will still be in need of doing random access to check if it was a missing value (unless it was a double). Anyway, I will look into it more tomorrow, maybe I'll find a better way.

h2o-core/src/main/java/water/util/VecUtils.java
michalkurka: In this particular case, I wouldn't use it because it takes away the option to allocate the array using MemoryManager: Whenever you are allocating an array that is not small (= can grow arbitrarily) you should ideally use MemoryManager.

For small arrays located outside of the loop that processes the chunk items it doesn't matter, but it is not really the h2o-style.

h2o-core/src/main/java/water/util/VecUtils.java
michalkurka: I would probably not worry about it in the context of the varimp-calculation. Most of the complexity will be in scoring most likely.

This seems good as-is.

h2o-docs/src/product/data-science/permutation-variable-importance.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Parameters
+~~~~~~~~~~
+
+- **model**: A trained model for which it will be used to score the dataset.
+- **frame**: The dataset to use, both train and test frame are can be reasonable choices but the interpretation differs (see `Should I Compute Importance on Training or Test Data? <https://christophm.github.io/interpretable-ml-book/feature-importance.html#feature-importance-data>`__ from the Interpretable Machine Learning by Christoph Molnar.).
+- **metric**: The metric to be used to calculate the error measure. One of ``AUTO``, ``AUC``, ``MAE``, ``MSE``, ``RMSE``, ``logloss``, ``mean_per_class_error``, ``PR_AUC``. Defaults to ``AUTO``.
+- **n_samples**: The number of samples to be evaluated. Use -1 to use the whole dataset. Defaults to 10 000.
+- **n_repeats**: The number of repeated evaluations. Defaults to 1.
+- **features**: The features to include in the permutation importance. Use None to include all.
```
ledell, line 32: Should we use the term `columns` since all the rest of the column (or list of column) arguments have "column" in the name?  e.g. `fold_column`, `ignore_columns`, `gam_columns`, etc.  I don't think we have an args with "feature" in the name yet.


Looks like `n_repeats` is replicated to [match sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.inspection.permutation_importance.html), and `n_samples` is new to H2O?  I can't remember if we talked about this on Slack or not yet (were there existing arg names for something like n samples in H2O)?

tomasfryda: Thank you for the consistency check! 

There is a thread on api updates channel about the parameter but not about the naming. `n_samples` would be new to h2o. Now when searching I noticed we have `samples` in SHAP plot. Would it be better to use `samples` here as well?

I will change `features` to `columns` when we decide what to do with the `n_samples`.

ledell (approved):
Everything looks good, ready to merge!

PR #5498, opened 2021-05-14
PUBDEV-8135: Expose argument to enable configuring S3 with S3A configuration

michalkurka:


g-eoj (approved):
I don't see any issues although I'm no expert. Thanks!

michalkurka:
Thanks @g-eoj - you are on your way to become an expert ;)

PR #5500, opened 2021-05-17
PUBDEV-8152: remove dependency on colorama

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8152

### DIFF
--- a/h2o-py/h2o/backend/__init__.py
+++ b/h2o-py/h2o/backend/__init__.py
@@ -41,22 +41,10 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 from distutils.version import StrictVersion
 import sys
 
-import colorama
 
 from .cluster import H2OCluster
 from .server import H2OLocalServer
 from .connection import H2OConnection
 from .connection import H2OConnectionConf
 
-if not hasattr(colorama, '__version__'):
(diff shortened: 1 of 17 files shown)


h2o-py/h2o/backend/__init__.py
```diff
@@ -41,22 +41,10 @@
...
-    print("[WARNING] H2O requires colorama module of version 0.3.8 or newer. You have older version .\n"
-          "You can upgrade to the newest version of the module running from the command line\n"
-          "    $ pip%s install --upgrade colorama" % (sys.version_info[0]))
-    sys.exit(1)
-elif StrictVersion(colorama.__version__) < StrictVersion("0.3.8"):
-    print("[WARNING] H2O requires colorama module of version 0.3.8 or newer. You have version %s.\n"
-          "You can upgrade to the newest version of the module running from the command line\n"
-          "    $ pip%s install --upgrade colorama" % (colorama.__version__, sys.version_info[0]))
-    sys.exit(1)
-
```
sebhrusen, line 61: found this very ironic that a library used to print in the client console was checked in the `backend` module…

michalkurka (approved):
Thank you! Looks great to me

cc: @mn-mikke do we need to make a change in pysparkling documentation?

PR #5508, opened 2021-05-24
PUBDEV-7937 - fix python2 sorting test

valenad1:
It turns out that is is python 2 related bug.
The way I sort SHAP in python for check the predict_contributions  is Unstable in python2 because dict(...) is not ordered by insertion as it is in python3.

We already have tests for Java sort(..) stability and I presume that I can count with Frame.names stable order.


michalkurka (approved):
thank you for the fix

PR #5510, opened 2021-05-24
PUBDEV-8146: Add params (with default value) to the signature of estimator's constructors.

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8146
Also added type hints: https://h2oai.atlassian.net/browse/PUBDEV-8107

Using Sphinx syntax for the `__init__` docstring because it's more dense, and some algos really have a lot of parameters!
However, we don't appear to be using Sphinx syntax for the estimator's attributes… (docstring syntax is actually  quite inconsistent accross the h2o package). 
Any concern with that?

Also replaced our usage of `Enum` to list allowed values by the official, and in our case more accurate, `Literal` type (see https://www.python.org/dev/peps/pep-0586/).


h2o-extensions/xgboost/src/main/java/hex/schemas/XGBoostV3.java
```diff
@@ -55,7 +55,6 @@
 
         "score_tree_interval",
         "min_split_improvement", "gamma",
-        "auc_type",
```
sebhrusen, line 58: @maurever this one was duplicated, which created issues with PY code generation.
I kept the one at the end because that's where it is for most algos. Is it better to have here after `gamma`? (I don't really see why)

michalkurka:
h2o/sklearn/__init__.py seems to use 'param_names'

I would probably keep 'param_names' - instead of doing 

```
sig_params = {k:v for k, v in locals().items() if k != 'self' and not k.startswith('__')}
```

it just seems more obvious to me and since it is generated we don't have to worry about it

h2o-py/h2o/estimators/generic.py
```diff
@@ -19,20 +19,29 @@ class H2OGenericEstimator(H2OEstimator):
     """
 
     algo = "generic"
-    param_names = {"model_id", "model_key", "path"}
 
-    def __init__(self, **kwargs):
+    def __init__(self,
+                 model_id=None,  # type: str
+                 model_key=None,  # type: H2OFrame
```
michalkurka, line 25: the type information doesn't seem right here

michalkurka:
the type information seems wrong for model keys (says H2OFrame)

I don't know if this is a difficult change to make but perhaps it would be safer to add the type information in a separate PR - it would be easier to check

h2o-py/h2o/estimators/generic.py
sebhrusen: good catch, for the types, I'm using the type translation that was used for the docstring: there may be some missing conversions: in most cases where we need an H2OFrame, we accept the frame itself or just its key… maybe I should put `Union[str, H2OFrame]` everywhere `H2OFrame` or key is expected. The setter is supposed to extract the key correctly.

sebhrusen:
> h2o/sklearn/init.py seems to use 'param_names'

@michalkurka sklearn wrappers are actually a bit more clever: they are using `param_names` if provided (H2O estimators) and fallback to signature params extraction otherwise (e.g. H2OAutoML).
I think I can actually get rid of the small  `param_names` logic there as well.

sebhrusen:
@michalkurka Is it ok to completely get rid of the support for `Lambda` param name in GLM? Some tests are failing because of this deprecated param.
I will fix the tests but  if we still want to keep the old syntax, I can create a `@deprecated` decorator for parameters now.

michalkurka:
@sebhrusen please deprecate first, we should wait at least 2 major release cycles before we remove

sebhrusen:
@michalkurka `Lambda` is deprecated since 2017 (but without raising any warning)...
TE params are deprecated only since 3.32, so I'll create a @deprecated decorator for those: will be useful in the future  anyway.

michalkurka:
> @michalkurka Lambda is deprecated since 2017 (but without raising any warning)...

yes, there are more like that - eg. col_major in deeplearning - in that case it is not even doing anything

h2o-extensions/xgboost/src/main/java/hex/schemas/XGBoostV3.java
maurever: Good catch @sebhrusen! That is my mistake. I did not remember why I added it twice. It is ok to remove this one. Thank you!

h2o-py/h2o/estimators/aggregator.py
```diff
@@ -19,22 +19,59 @@ class H2OAggregatorEstimator(H2OEstimator):
     """
 
     algo = "aggregator"
-    param_names = {"model_id", "training_frame", "response_column", "ignored_columns", "ignore_const_cols",
-                   "target_num_exemplars", "rel_tol_num_exemplars", "transform", "categorical_encoding",
-                   "save_mapping_frame", "num_iteration_without_new_exemplar", "export_checkpoints_dir"}
 
-    def __init__(self, **kwargs):
+    def __init__(self,
+                 model_id=None,  # type: str
```
tomasfryda, line 24: Since `model_id` can be `None`, its type should be `Optional[str]`.  This also applies in all the other places where the parameter value can be `None`, not just optional attributes[1].

[1] https://docs.python.org/3/library/typing.html#typing.Optional

h2o-py/h2o/estimators/generic.py
michalkurka: I see what is going on... it is a bad naming of the parameter. `model_key` is misleading.

The meaning of model_key is key to a binary Vec (wrapped in Frame) that represents the MOJO bytes. So the type should indeed by H2OFrame.

sebhrusen: OK, here is the distinction:
```
        :param model_key: Key to the self-contained model archive already uploaded to H2O. (default:None).
        :param path: Path to file with self-contained model archive. (default:None).
```
It could (should!) have been a `str` but the client was designed to accept an `H2OFrame` instead for an unknown reason…

sebhrusen: explanation found in `h2o.upload_mojo`:
```python
    response = api("POST /3/PostFile", filename=mojo_path)
    frame_key = response["destination_frame"]
    mojo_estimator = H2OGenericEstimator(model_key = get_frame(frame_key))
```
I still find this very confusing.

michalkurka: Agreed, it is confusing.

sebhrusen:
@michalkurka
> I wonder if it would be of any benefit of actually generating the setter calls for all the parameters. We now have the arguments in the constructor explicitly and we don't need to use the map anymore.

let me try that. As far as I see it, the main benefit is that is allows to get rid of this line that you don't like:
```python
sig_params = {k:v for k, v in locals().items() if k != 'self' and not k.startswith('__')}
```
Otherwise, the generated code will be a bit more verbose but there will be same amount of setters called at runtime.

h2o-bindings/bin/custom/python/gen_drf.py
```diff
@@ -1,4 +1,4 @@
-deprecated = ['offset_column', 'distribution']
+deprecated_params = dict(offset_column=None)
```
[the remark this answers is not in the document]
sebhrusen: The problem I had with this param is that it was deprecated only in comments: the backend is processing the param, so I was not sure that it was "really" deprecated.
I someone can confirm, then I'll be happy to make it fully deprecated (with warnings and so on).

h2o-py/h2o/estimators/aggregator.py
```diff
@@ -19,22 +19,64 @@ class H2OAggregatorEstimator(H2OEstimator):
...
+                 target_num_exemplars=5000,  # type: int
+                 rel_tol_num_exemplars=0.5,  # type: float
+                 transform="normalize",  # type: Literal["none", "standardize", "normalize", "demean", "descale"]
+                 categorical_encoding="auto",  # type: Literal["auto", "enum", "one_hot_internal", "one_hot_explicit", "binary", "eigen", "label_encoder", "sort_by_response", "enum_limited"]
+                 save_mapping_frame=False,  # type: bool
+                 num_iteration_without_new_exemplar=500,  # type: int
+                 export_checkpoints_dir=None,  # type: Optional[str]
+                 ):
+        """
+        :param model_id: Destination id for this model; auto-generated if not specified. (default:None).
```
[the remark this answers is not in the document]
sebhrusen: I used parenthesis because I had  issues with punctuation: some help string ends with dots, some other don't and I didn't want a hacky logic like checking if the sentences ends with a dot or whatever.
I found a workaround though  and fixed our `wrap` function to handle multilines, so now I can sefely add the defaults as the last line.

h2o-bindings/bin/gen_python.py
```diff
@@ -294,7 +317,12 @@ def extend_schema_params(param):
                 # default assignment
```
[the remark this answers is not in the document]
sebhrusen: You put me in a nasty situation with those doc/validation issues: legacy code was actually very inconsistent.
In last commits, I tried to make it better, but this generated logic is hard to get right for all use-cases… and I'm not sure anymore that I should have fixed this here.

sebhrusen:
@michalkurka, @tomasfryda All issues/comments should be addressed now:
- deprecations.
- inconsistencies in parameters type checking.
- parameters type hints.
- Py2.7 support.

Build is green now, please have another look.

h2o-bindings/bin/custom/python/gen_deeplearning.py
```diff
@@ -24,14 +24,14 @@ def __init__(self, **kwargs):
 overrides = dict(
     initial_biases=dict(
         setter="""
-assert_is_type({pname}, None, [H2OFrame, None])
+assert_is_type({pname}, None, [None, str, H2OFrame])
```
tomasfryda, line 27: Isn't one of the `None`s superfluous here? Looks to me like `_check_type` checks `Union[None, Union[None, str, H2OFrame]]`. Same on the line 34.

sebhrusen: checktype should check `Union[None, List[Union[None, str, H2OFrame]]]`

michalkurka:
@sebhrusen there are still commented out tests in the AutoML test suite

sebhrusen:
>  there are still commented out tests in the AutoML test suite

@michalkurka dammit, sorry for that

michalkurka (approved):
@sebhrusen thank you for moving the package forward, all the improvements are very nice

LGTM if there is no backwards compatibility issue

h2o-py/h2o/transforms/decomposition.py
```diff
@@ -123,20 +122,20 @@ def __init__(self, nv=None, max_iterations=None, transform="NONE", seed=None,
...
+            - ``"randomized"``: approximate SVD by projecting onto a random subspace.
 
         :returns: a new H2OSVD model
         """
         super(H2OSVD, self).__init__()
         self._parms = locals()
         self._parms = {k: v for k, v in self._parms.items() if k != "self"}
 
-        assert_is_type(svd_method, Enum("GramSVD", "Power", "GLRM", "Randomized"))
+        assert_is_type(svd_method, None, Enum("gram_s_v_d", "power", "randomized"))
```
tomasfryda, line 136: `gram_s_v_d` looks like a mistake.

sebhrusen: I'm just using the exact same naming as in `estimators/svd.py`: not planning to fix every single nitpick from legacy code here.
Note that `_` are ignored both in type check and on backend side when resolving values to Java enum.

sebhrusen: PCA  supports `GRLM` but SVD apparently doesn't: either it was a mistake (copy/paste from PCA) in this file, or the SVD estimator is missing it.

sebhrusen: in SVDModel:
```java
    public enum Method {
      GramSVD, Power, Randomized
    }
```

tomasfryda (approved):
Looks good to me. Thank you for this great improvement!

PR #5511, opened 2021-05-24
PUBDEV-8150: Added warning message to Java, R, Python clients if use alpha in hyper-parameter

wendycwong:
This PR is the second part to this JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8150.

I have added the warning message to 
- Java
- python client
- R client.

If we add more clients, we will have to add the error and warning message to each individual client.  This is the way gridsearch was implemented.  It has its own set of failure_messages and need the be processed for each client.

### DIFF
--- a/h2o-py/h2o/grid/grid_search.py
+++ b/h2o-py/h2o/grid/grid_search.py
@@ -1,5 +1,8 @@
 # -*- encoding: utf-8 -*-
 from __future__ import division, print_function, absolute_import, unicode_literals
+
+import warnings
+
 from h2o.utils.compatibility import *  # NOQA
 
 import itertools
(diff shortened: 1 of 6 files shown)


wendycwong:
The warning message is like this:

Adding alpha array to hyperparameter runs slower with gridsearch.  This is due to the fact that the algo has to run initialization for every alpha value.  Setting the alpha array as a model parameter will skip the initialization and run faster overall.

Please help me to make it better.  Thank you in advance!

michalkurka (approved):
Thank you so much @wendycwong

h2o-py/h2o/grid/grid_search.py
```diff
@@ -421,6 +424,14 @@ def _handle_build_finish(self, grid, rest_ver=None):
         grid_json = h2o.api("GET /99/Grids/%s" % (grid.dest_key))
         failure_messages_stacks = ""
         error_index = 0
+        # hyper_params = grid_json['hyper_names']
```
michalkurka, line 427: I assume this is not needed anymore (just an initial attempt)

wendycwong: Good catch Michalk.  Fixed!

ledell:
@wendycwong This is great!  I will add a commit to modify the alpha parameter description for R/Python to let people know that an array is allowed (currently there's no information to let the user know that this is supported).

wendycwong:
@ledell Hannah is already working on changing the documentation on alpha.  I asked for her help after I got your JIRA.

ledell:
Ok, @hannah-tillman and I can work on the docs PR separately.  So I'll go ahead and approve this one and then you can merge.

ledell (approved):
Thanks!

PR #5496, opened 2021-05-14
PUBDEV-8109 friedmans and popescus H statistic

koniecsveta:
tested against https://pypi.org/project/sklearn-gbmi/ in hex/tree/FriedmanPopescusHTest.java

PR #5516, opened 2021-05-27
PUBDEV-8174: remove algo-specific logic from Py base classes

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8174

The goal of this PR is to use a proper OO approach in our Python implementation, especially regarding the `estimators` and `models`.
Too often, base classes were executing code specific to their implementations, making code hard to read, painful to debug, and bug-prone.

This PR uses a declarative approach to dynamically add functionalities to algos at estimator or model level.
#### Example
In `gen_drf.py`:
```python
options = dict(
    verbose=True,
    model_extensions=[
        'h2o.model.extensions.ScoringHistory',
        'h2o.model.extensions.VariableImportance',
        'h2o.model.extensions.Trees',
    ],
)
```
tells that DRF supports `verbose` mode during training, and declares some extensions, which are mixin classes dynamically added to the model after training.

Those extensions allow us to group code by functionality, and to expose them only to models that support them.

Finally, algos can have their own implementation of some functionalities, for example in `gen_glm.py`:
```python
options = dict(
    model_extensions=[
        'h2o.model.extensions.ScoringHistoryGLM',
        'h2o.model.extensions.StandardCoef',
        'h2o.model.extensions.VariableImportance',
    ],
)
```

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -108,8 +108,8 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
     };
   }
 
-  @Override public boolean havePojo() { return !_parms._HGLM; }
-  @Override public boolean haveMojo() { return !_parms._HGLM; }
+  @Override public boolean havePojo() { return true; }
+  @Override public boolean haveMojo() { return true; }
 
   private double _lambdaCVEstimate = Double.NaN; // lambda cross-validation estimate
   private int _bestCVSubmodel;  // best submodel index found during cv
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -108,8 +108,8 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
     };
   }
 
-  @Override public boolean havePojo() { return true; }
-  @Override public boolean haveMojo() { return true; }
+  @Override public boolean havePojo() { return !_parms._HGLM; }
+  @Override public boolean haveMojo() { return !_parms._HGLM; }
 
   private double _lambdaCVEstimate = Double.NaN; // lambda cross-validation estimate
   private int _bestCVSubmodel;  // best submodel index found during cv
--- a/h2o-algos/src/main/java/hex/glm/GLMModel.java
+++ b/h2o-algos/src/main/java/hex/glm/GLMModel.java
@@ -1915,13 +1915,13 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
 
   @Override
   public boolean haveMojo() {
-    if (_parms.interactionSpec() == null) return super.haveMojo();
+    if (!_parms._HGLM && _parms.interactionSpec() == null) return super.haveMojo();
     else return false;
   }
 
   @Override
   public boolean havePojo() {
-    if (_parms.interactionSpec() == null) return super.havePojo();
+    if (!_parms._HGLM && _parms.interactionSpec() == null) return super.havePojo();
     else return false;
   }
 
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -24,6 +24,8 @@ from h2o.utils.typechecks import I, assert_is_type, assert_satisfies, Enum, is_t
 class ModelBase(h2o_meta(Keyed)):
     """Base class for all models."""
 
+    _options_ = {}    # dict of options declared in implementation
+
     def __init__(self):
         """Construct a new model instance."""
         self._id = None
@@ -34,7 +36,7 @@ class ModelBase(h2o_meta(Keyed)):
         self._xval_keys = None
         self._parms = {}  # internal, for object recycle
         self.parms = {}  # external
-        self._estimator_type = "unsupervised"
+        self._estimator_type = None
         self._future = False  # used by __repr__/show to query job state
         self._job = None  # used when _future is True
         self._have_pojo = False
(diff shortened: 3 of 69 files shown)


sebhrusen:
@michalkurka feel free to provide an early high-level feedback on this: this is a structural change, and I want to get it right.
Especially, I have concern if we want to expose feature methods only to models that support them, and if we want to keep the  current approach of having for example `varimp_plot` exposed to all models, including those that  don't support it.

Finally, I currently only extracted the code for which there was explicit conditional logic (`if algo=...`), but for example, `varimp` function could also be moved completely to the `VariableImportance` extension: it would be more consistent. My only concern is that I don't have a list of all the methods that could/should be moved.

h2o-py/h2o/transforms/decomposition.py
```diff
@@ -1,16 +1,21 @@
 from h2o.frame import H2OFrame
-from h2o.utils.typechecks import assert_is_type, Enum
+from h2o.utils.mixin import assign
 from ..estimators.estimator_base import H2OEstimator
+from ..estimators import H2OPrincipalComponentAnalysisEstimator, H2OSingularValueDecompositionEstimator
```
sebhrusen, line 4: I couldn't stand this out-of-sync duplicated logic anymore here (I don't think this is much used besides), so refactored this to use a delegate approach.
This now works as the model obtained from the backend can be resolved on any object, here:
```
self._delegate._model = self
```

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -108,8 +108,8 @@ public boolean isSupervised() {
     };
   }
 
-  @Override public boolean havePojo() { return true; }
-  @Override public boolean haveMojo() { return true; }
+  @Override public boolean havePojo() { return !_parms._HGLM; }
+  @Override public boolean haveMojo() { return !_parms._HGLM; }
```
sebhrusen, line 112: @wendycwong can you please confirm that this is ok?

`estimator_base` used to have the following logic that I really wanted to remove:
```python
        if (model_json["algo"]=="glm") and self.HGLM:
            m._have_pojo = False
            m._have_mojo = False
        else:
            m._have_pojo = model_json.get('have_pojo', True)
            m._have_mojo = model_json.get('have_mojo', True)
```
so I figured it would be better that have the values changed on server side.

michalkurka:
I took a quick look and it looks good to me - the concept would actually translate to backend as well. The backend should ideally say what features the algo support instead of it being specified in the Python file.

This way we could have identical functionality in R/Sparkling Water. However, that could be a next step.

> Especially, I have concern if we want to expose feature methods only to models that support them, and if we want to keep the current approach of having for example varimp_plot exposed to all models, including those that don't support it.

My thinking is that the first iteration should remove anything from the algos. It should just move code around to make it more dynamic.

In the next version we can start removing.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: This looks good to me.  Thanks.

wendycwong: Seb:  Can you please move this to GLMModel.java?  Take a look at the haveMojo/Pojo there and proceed accordingly.  Thanks, Wendy

wendycwong (changes requested):

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1915,13 +1915,13 @@ protected PredictScoreResult predictScoreImpl(Frame fr, Frame adaptFrm, String d
...
   public boolean haveMojo() {
-    if (_parms.interactionSpec() == null) return super.haveMojo();
+    if (!_parms._HGLM && _parms.interactionSpec() == null) return super.haveMojo();
     else return false;
   }
 
   @Override
   public boolean havePojo() {
-    if (_parms.interactionSpec() == null) return super.havePojo();
+    if (!_parms._HGLM && _parms.interactionSpec() == null) return super.havePojo();
```
sebhrusen, line 1924: @wendycwong please confirm, I reverted the changes in GLM driver (haveMojo/Pojo always return true).

wendycwong (approved):

h2o-py/h2o/model/model_base.py
```diff
@@ -1506,193 +1409,28 @@ def varimp_plot(self, num_of_features=None, server=False):
...
-        # if H2ODeepLearningEstimator has variable_importances == True
-        elif self._model_json["algo"] == "deeplearning":
-            plt.title("Variable Importance: H2O Deep Learning", fontsize=20)
-            if not server: plt.show()
-        elif self._model_json["algo"] == "glm":
-            plt.title("Variable Importance: H2O GLM", fontsize=20)
-            if not server: 
-                plt.show()            
+        # For now, redirect to h2o.model.extensions.varimp for models that support the feature, and raise legacy error for others.
+        # Later, the method will be exposed only for models supporting the feature.
```
sebhrusen, line 1413: @michalkurka as discussed, keeping the all those features exposed in `ModelBase` for now.
Second step (exposing features only for relevant models), will just be a matter of deleting the `ModelBase` methods and removing the leading underscore to make the implementation "public" in the feature mixin classes.

sebhrusen:
@michalkurka , @koniecsveta : this is ready for merging if you want have a look at it so that Zuzana can use it asap in https://github.com/h2oai/h2o-3/pull/5496

koniecsveta:
> @michalkurka , @koniecsveta : this is ready for merging if you want have a look at it so that Zuzana can use it asap in #5496

so I will just create new mixin for example 'FriedmanPopescusH' and add it to xgboost and gbm's options as another model_extension analogically as FeatureInteracion was done and it should all work, right?

sebhrusen:
> so I will just create new mixin for example 'FriedmanPopescusH' and add it to xgboost and gbm's options as another model_extension analogically as FeatureInteracion was done and it should all work, right?

@koniecsveta yep, that's the idea: feature encapsulated in it's own extension as a mixin class and then, supported algos register the extension.
As it's a new feature, you probably don't want to expose anything in `ModelBase` so that the method should not even be visible for other algos.

sebhrusen:
@michalkurka , @koniecsveta , @tomasfryda 
I integrated the new `HStatistic` feature to this PR, and I would like not having to do so with the next ongoing feature, so I would appreciate some review: build is green, so the review is the missing part, thanks.

michalkurka (approved):
Looks really good - very nice change, thank you @sebhrusen

PR #5517, opened 2021-05-28
PUBDEV-8167: Screeplot fails for PCA in R and Python

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8167

Problem in python was that it always expected `server` in `kwargs` so I made it optional.

In R, the varimp wasn't supported so I added support for it by copying `model$importance` to `model$variable_importances` in the `.h2o.fill_pca` and I also added `h2o.screeplot` which does the same thing as the one in python.

### DIFF
--- a/h2o-py/h2o/model/dim_reduction.py
+++ b/h2o-py/h2o/model/dim_reduction.py
@@ -94,19 +94,17 @@ class H2ODimReductionModel(ModelBase):
         return h2o.get_frame(j["model_metrics"][0]["predictions"]["frame_id"]["name"])
 
 
-    def screeplot(self, type="barplot", **kwargs):
+    def screeplot(self, type="barplot", server=False):
         """
         Produce the scree plot.
 
         Library ``matplotlib`` is required for this function.
 
         :param str type: either ``"barplot"`` or ``"lines"``.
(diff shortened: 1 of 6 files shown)


h2o-py/h2o/model/dim_reduction.py
```diff
@@ -94,19 +94,17 @@ def proj_archetypes(self, test_data, reverse_transform=False):
...
 
-    def screeplot(self, type="barplot", **kwargs):
+    def screeplot(self, type="barplot", server=False):
         """
         Produce the scree plot.
 
         Library ``matplotlib`` is required for this function.
 
         :param str type: either ``"barplot"`` or ``"lines"``.
+        :param bool server: if true set server settings to matplotlib and show the graph
```
sebhrusen, line 104: if server is true, show the graph? are you sure?
```suggestion
        :param bool server: if true set server settings to matplotlib and do not show the graph
```

sebhrusen (approved):
looks good, thanks

h2o-py/h2o/model/dim_reduction.py
tomasfryda: Thanks, I completely missed that. I copied it from varimp_plot so I'll fix that as well.

PR #5518, opened 2021-06-03
PUBDEV-8099: XGBoost: Fix tree reweighting - did not work for larger frames (nChunks>1)

michalkurka:


h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
```diff
@@ -661,6 +661,7 @@ private void checkUpdateAuxTreeWeights(GBMModel gbm, Frame frame) {
     Map<Integer, SharedTreeMojoModel.AuxInfo>[][] originalAuxInfos = visitTrees(
             gbm._output._treeKeysAux, Map.class, CompressedTree::toAuxInfos);
 
+    frame = Scope.track(ensureDistributed(frame, 16));
```
maurever, line 664: Why 16 is the correct constant here? I am just curious...

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/predict/UpdateAuxTreeWeightsTask.java
```diff
@@ -36,16 +36,22 @@ public UpdateAuxTreeWeightsTask(DistributionFamily dist, DataInfo di, XGBoostMod
         if (_dist != DistributionFamily.gaussian && _dist != DistributionFamily.bernoulli) {
             throw new UnsupportedOperationException("Updating tree weights is currently not supported for distribution " + _dist + ".");
         }
+    }
+
+    private double[][] initNodeWeights() {
```
maurever, line 41: :+1:

maurever (approved):
Looks good, thank you @michalkurka. Do you have any idea how to fix this for larger frames?

koniecsveta (approved):

h2o-algos/src/test/java/hex/tree/gbm/GBMTest.java
michalkurka: It guarantees that one node will have at least 3 chunks and it is a good number for this dataset (the chunks won't be empty).

michalkurka:
> Looks good, thank you @michalkurka. Do you have any idea how to fix this for larger frames?

Large frames should be fine on its own, do I understand you correctly?

PR #5519, opened 2021-06-03
PUBDEV-3524: Add xval argument to h2o.confusionMatrix in R

michalkurka:


ledell:
@michalkurka This was missing a reviewer, so I added myself. :-). This looks good!

PR #5520, opened 2021-06-04
PUBDEV-8181: Fix misuse of is in python

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8181

In new python versions, we get syntax warning about using `is` where we should use `==`. The former should be used to compare "pointers" (`id(obj1) == id(obj2)`), while the latter compares the value.


michalkurka (approved):
Thank you

PR #5521, opened 2021-06-04
PUBDEV-8180: uniform Python API to retrieve trained model(s)/leader

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8180

### DIFF
--- a/h2o-py/h2o/grid/grid_search.py
+++ b/h2o-py/h2o/grid/grid_search.py
@@ -29,42 +29,7 @@ from h2o.utils.typechecks import assert_is_type, is_type
 class H2OGridSearch(h2o_meta(Keyed)):
     """
     Grid Search of a Hyper-Parameter Space for a Model
-
-    :param model: The type of model to be explored initialized with optional parameters that will be
-        unchanged across explored models.
-    :param hyper_params: A dictionary of string parameters (keys) and a list of values to be explored by grid
-        search (values).
-    :param str grid_id: The unique id assigned to the resulting grid object. If none is given, an id will
-        automatically be generated.
-    :param search_criteria:  The optional dictionary of directives which control the search of the hyperparameter space.
-        The dictionary can include values for: ``strategy``, ``max_models``, ``max_runtime_secs``, ``stopping_metric``, 
-        ``stopping_tolerance``, ``stopping_rounds`` and ``seed``. The default strategy, "Cartesian", covers the entire space of 
-        hyperparameter combinations. If you want to use cartesian grid search, you can leave the search_criteria 
-        argument unspecified. Specify the "RandomDiscrete" strategy to get random search of all the combinations of 
-        your hyperparameters with three ways of specifying when to stop the search: max number of models, max time, and 
-        metric-based early stopping (e.g., stop if MSE hasn’t improved by 0.0001 over the 5 best models). 
-        Examples below::
-
-            >>> criteria = {"strategy": "RandomDiscrete", "max_runtime_secs": 600,
-            ...             "max_models": 100, "stopping_metric": "AUTO",
-            ...             "stopping_tolerance": 0.00001, "stopping_rounds": 5,
-            ...             "seed": 123456}
-            >>> criteria = {"strategy": "RandomDiscrete", "max_models": 42,
-            ...             "max_runtime_secs": 28800, "seed": 1234}
-            >>> criteria = {"strategy": "RandomDiscrete", "stopping_metric": "AUTO",
-            ...             "stopping_tolerance": 0.001, "stopping_rounds": 10}
-            >>> criteria = {"strategy": "RandomDiscrete", "stopping_rounds": 5,
-            ...             "stopping_metric": "misclassification",
-            ...             "stopping_tolerance": 0.00001}
-    :param export_checkpoints_dir: Directory to automatically export the grid and its models to.
-    :param recovery_dir: When specified, the grid and all necessary data (frames, models) will be saved to this
-        directory (use HDFS or other distributed file-system). Should the cluster crash during training, the grid
-        can be reloaded from this directory via ``h2o.load_grid``, and training can be resumed.
-    :param parallelism: Level of parallelism during grid model building. 1 = sequential building (default). 
-         Use the value of 0 for adaptive parallelism - decided by H2O. Any number > 1 sets the exact number of models
-         built in parallel.
-    :returns: a new H2OGridSearch instance
-
+    
     Examples
     --------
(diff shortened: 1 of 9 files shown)


h2o-py/h2o/grid/grid_search.py
```diff
@@ -77,9 +42,44 @@ class H2OGridSearch(h2o_meta(Keyed)):
...
+                ...             "stopping_tolerance": 0.00001}
+        :param export_checkpoints_dir: Directory to automatically export the grid and its models to.
+        :param recovery_dir: When specified, the grid and all necessary data (frames, models) will be saved to this
+            directory (use HDFS or other distributed file-system). Should the cluster crash during training, the grid
+            can be reloaded from this directory via ``h2o.load_grid``, and training can be resumed.
+        :param parallelism: Level of parallelism during grid model building. 1 = sequential building (default). 
+             Use the value of 0 for adaptive parallelism - decided by H2O. Any number > 1 sets the exact number of models
+             built in parallel.
+        :returns: a new H2OGridSearch instance
+        """
```
sebhrusen, line 82: just moved this doc where it belongs

h2o-py/h2o/utils/metaclass.py
```diff
@@ -201,6 +200,7 @@ def __call__(self, clz):
         clz._bc = self
         new_clz = None
 
+        @wraps(clz.__init__)
```
sebhrusen, line 203: small fix here to ensure that the init signature and doc appears correctly when printing help in ipython/jupyter: noticed this when fixing the GridSearch docstring.

michalkurka:
@sebhrusen looks very good, I have a question about GridSearch vs AutoML

GridSearch returns all models while AutoML returns the leader

how do we justify that? the concept is not dissimilar?

sebhrusen:
@michalkurka I'm glad you ask, because I was wondering for a few minutes what to do with grids.

The main goal of AutoML is to produce one good model, so I think we all agree that it makes returning the leader there.

We could consider that the goal of grids is also to produce one good model at the end, but currently grids doesn't provide any API to select the best model among all, so it appears that our implementation de facto doesn't aim at producing one single model…
We could add this API to grids, but it would be out-of-scope for this ticket, so if we plan to add a model selection API to grids, I would suggest that we keep `train` method for grids as before (not returning anything), and we will make it return a single model only then.

wdyt?

sebhrusen:
@michalkurka second thought: `R` return a grid  instance,  so maybe we should just return `self` in case of grids as well.

ledell:
@sebhrusen 
> We could consider that the goal of grids is also to produce one good model at the end, but currently grids don't provide any API to select the best model among all, so it appears that our implementation de facto doesn't aim at producing one single model… 

I like the idea that you could train and get the top model from a grid in one-line of code, just like AutoML.  I think it's a good idea!  It seems like grid already comes pre-sorted, using a default metric.  In binary classification, its seems to be sorted by logloss.  Usually I use  `.get_grid(sort_by="auc")` but it seems that we could re-use the single-model leader concept here, and then we could use grids in pipelines.

Example (current functionality):

```
import h2o
from h2o.estimators.gbm import H2OGradientBoostingEstimator

h2o.init()

train = h2o.import_file("https://s3.amazonaws.com/erin-data/higgs/higgs_train_10k.csv")
x = train.columns
y = "response"
x.remove(y)
train[y] = train[y].asfactor()
test[y] = test[y].asfactor()



# Specify GBM hyperparameters for the grid
hyper_params = {"learn_rate": [0.01, 0.03],
                "max_depth": [3, 4, 5, 6, 9],
                "sample_rate": [0.7, 0.8, 0.9, 1.0],
                "col_sample_rate": [0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8]}
search_criteria = {"strategy": "RandomDiscrete", "max_models": 3, "seed": 1}

# Train the grid
grid = H2OGridSearch(model=H2OGradientBoostingEstimator(ntrees=10,
                                                        seed=1,
                                                        nfolds=5),
                     hyper_params=hyper_params,
                     search_criteria=search_criteria,
                     grid_id="gbm_grid_binomial")
grid.train(x=x, y=y, training_frame=train)

grid_auc = grid.get_grid(sort_by='auc', decreasing=True)
```


Currently sorted by logloss by default:
```
In [26]: grid
Out[26]:     col_sample_rate learn_rate max_depth sample_rate                  model_ids             logloss
0               0.4       0.03         3         0.7  gbm_grid_binomial_model_3  0.6670120129698885
1               0.2       0.03         4         0.8  gbm_grid_binomial_model_2  0.6746770622641063
2               0.7       0.01         5         0.9  gbm_grid_binomial_model_1  0.6769884389744855

In [28]: grid_auc
Out[28]:     col_sample_rate learn_rate max_depth sample_rate                  model_ids                 auc
0               0.7       0.01         5         0.9  gbm_grid_binomial_model_1  0.7462602720068559
1               0.4       0.03         3         0.7  gbm_grid_binomial_model_3  0.7335792493670467
2               0.2       0.03         4         0.8  gbm_grid_binomial_model_2  0.7140694557755547
```

sebhrusen:
@ledell @michalkurka I look at the backend and I can confirm what Erin says, the models are automatically sorted (for some reason, this logic is done in `GridSchemaV99`, which is really weird…).
And here are the default sorting metrics:
```
  public GridSortBy getDefaultGridSortBy() {
    if (! isSupervised())
      return null;
    else if (_output.nclasses() > 1)
      return GridSortBy.LOGLOSS;
    else
      return GridSortBy.RESDEV;
  }
 ```
Also, the `sort_by` param suggested could easily be added to the constructor: the logic is already there in the backend, it's the `sort_by` param is just not currently used when fetching the grid on the job completion… I won't  change this in this PR though.

Which leaves us with 2 possibilities as `train` result:
1. return the grid itself: justification being that it has a predict method, and metrics methods similar  to the ones provided with a single model, the difference being that those methods predict/compute metrics on all models of the grid.
2. return the first model: justification is to be consistent with AutoML, and estimator APIs.  The first model will always use the default sorting for now (could be improved later if we add the `sort_by` param to the grid constructor). The only glitch here is for unsupervised models: we can't justify picking the first model.

Time to vote, 1 or 2?

sebhrusen:
@michalkurka @ledell are we ok with this one? there's nothing blocking the PR otherwise, last build was Python-green, I just had to trigger a new build for the PR metadata check (now that this is fixed for client-only PRs).

michalkurka (approved):
LGTM - nice change

koniecsveta (approved):

PR #5523, opened 2021-06-04
PUBDEV-8229: Warn if H2O doesn't have enough memory to run XGBoost

michalkurka:
This will need to be activated by specifying a desired ration on
H2O invocation.

Since the problem primarily affects standalone H2O started from Py/R
we should only add this flag to Py/R clients:

Eg.:
-off_heap_memory_ratio 0.75

### DIFF
--- a/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostExtension.java
+++ b/h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostExtension.java
@@ -80,10 +80,11 @@ public class XGBoostExtension extends AbstractH2OExtension {
   public void onLocalNodeStarted() {
     if (!isEnabled())
       return;
+    final double ratio = H2O.ARGS.off_heap_memory_ratio;
     if (H2O.ARGS.off_heap_memory_ratio > 0) {
-      MemoryCheck.Report report = MemoryCheck.runCheck(H2O.ARGS.off_heap_memory_ratio);
+      MemoryCheck.Report report = MemoryCheck.runCheck(ratio);
       if (!report.isOffHeapRequirementMet()) {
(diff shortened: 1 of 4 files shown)


h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostExtension.java
```diff
@@ -75,6 +76,19 @@ public static NativeLibraryLoaderChain getLoader() throws IOException {
...
 
+  @Override
+  public void onLocalNodeStarted() {
+    if (!isEnabled())
+      return;
+    if (H2O.ARGS.off_heap_memory_ratio > 0) {
+      MemoryCheck.Report report = MemoryCheck.runCheck(H2O.ARGS.off_heap_memory_ratio);
+      if (!report.isOffHeapRequirementMet()) {
+        LOG.warn("There doesn't seem to be enough memory available for XGBoost model training, " +
+                "training XGBoost models is not advised. Details: " + report);
```
sebhrusen, line 87: Shouldn't you mention the off-heap memory ratio in the message? So that the users knows what to change?
Also wondering how we could raise those warnings to the client on startup, but it's for another PR.

sebhrusen (approved):
Looks good, so to disable XGB based on memory constraints, AutoML will have to both check that `off_heap_memory_ratio` is big enough and rerun a MemoryCheck on startup  to check if the requirement is met, right?

h2o-extensions/xgboost/src/main/java/hex/tree/xgboost/XGBoostExtension.java
michalkurka: Added the ratio information to the message.

Yes, eventually we should check the preconditions before the algo is started. We could do it in a similar way we check for XGBoost availability.

PR #5525, opened 2021-06-07
Act/Pred in Confusion Matrix

ChuliangXiao:
In H2O, the actual results display in the **rows** and the predictions display in the **columns**


hannah-tillman (approved):

valenad1 (approved):
Thank you for noticing! Your change is consistent with the matrix's label in Flow/Python/R.

`Training Metrics - Confusion Matrix Row labels: Actual class; Column labels: Predicted class`

PR #5526, opened 2021-06-08
DHistogram clean-up: remove unused parameter nbin_cats

michalkurka:
The main goal of this clean-up is to remove nbin_cats from method signatures - it was not actually used and the source of the parameter was parms._nbin_cats instead.

### DIFF
--- a/h2o-algos/src/main/java/hex/tree/ScoreBuildHistogram2.java
+++ b/h2o-algos/src/main/java/hex/tree/ScoreBuildHistogram2.java
@@ -284,9 +284,6 @@ public class ScoreBuildHistogram2 extends ScoreBuildHistogram {
     final DHistogram [] _lh;
 
     AtomicInteger _cidx;
-    private boolean _done;
-
-    public boolean isDone(){return _done || (_done = _cidx.get() >= _cids.length);}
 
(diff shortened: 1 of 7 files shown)


h2o-algos/src/main/java/hex/tree/ScoreBuildHistogram2.java
```diff
@@ -299,7 +296,7 @@ public ComputeHistoThread makeCopy() {
     }
 
     @Override
-    protected void map(int id){
+    protected void map(int UNUSED){
```
maurever, line 299: This change looks weird. I rather keep it as `id` if there is no solution for removing this from the parent class.

maurever (commented):
I am surprised how much dead code is in this part of the tree calculation.  I checked everything, and it looks really unused. But I would rather wait for tests to be sure.

h2o-algos/src/main/java/hex/tree/ScoreBuildHistogram2.java
michalkurka: ok, sure - will remove

michalkurka:
@maurever tests look good - can you please approve? I will remove the problematic commit you pointed out

maurever (approved):
Sure, thank you @michalkurka!

michalkurka:
fixed, thank you @maurever

PR #5528, opened 2021-06-08
Clean-up unused AutoML classes that are calling deprecated methods

michalkurka:


sebhrusen (approved):
Love deleted code! :)
Been thinking about this for a while, thanks for getting it done.

PR #5533, opened 2021-06-10
PUBDEV-8187: set max XGB max_depth to 15 in AutoML

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8187


ledell:
@sebhrusen Fix looks good -- are you planning to test this on the benchmark, or on a subset of the benchmark?

sebhrusen:
@ledell yes, I'll run the benchmark (probably subset) before merging: that's the only reason why the PR is WIP

PR #5534, opened 2021-06-11
PUBDEV-8312: Speed-up GBM by optimizing histogram building

michalkurka:
1. Optimize binning for small-cardinality categoricals
2. Collocate response data by tree-node for faster random access in CPU cache
3. Stop model building if new trees cannot bring an improvement (won't change the model)

### DIFF
--- a/h2o-algos/src/main/java/hex/tree/DHistogram.java
+++ b/h2o-algos/src/main/java/hex/tree/DHistogram.java
@@ -4,6 +4,7 @@ import hex.Distribution;
 import hex.genmodel.utils.DistributionFamily;
 import org.apache.log4j.Logger;
 import water.*;
+import water.fvec.Chunk;
 import water.fvec.Frame;
 import water.fvec.Vec;
 import water.util.ArrayUtils;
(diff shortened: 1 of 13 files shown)


h2o-algos/src/main/java/hex/tree/DHistogram.java
```diff
@@ -331,6 +341,8 @@ else if (_histoType== SharedTreeModel.SharedTreeParameters.HistogramType.Quantil
         _splitPtsFloat[i] = (float) binAt(i);
       }
     }
+    assert !_intOpt || _splitPts == null : "Integer-optimization cannot be enabled when split points are defined";
```
maurever, line 344: You check  `_splitPts == null` here in the assert, but I can't see this check in `useIntOpt()` method. There should be something like (similar to constraints):

```
    if (_splitPts != null) {
      return false;
    }
```

Or am I missing something?

maurever (commented):
Looks good to me. I left just one comment. Also, tests are missing. Otherwise cool improvement!

h2o-algos/src/main/java/hex/tree/DHistogram.java
maurever: I got it, the _splitPts is initialized later (before this assert).

maurever (approved):
Tests look good. Thank you @michalkurka.

michalkurka:
@maurever thank you for your review and the comments you send

PR #5535, opened 2021-06-11
PUBDEV-8146: fix small Py2 issues after introducing deprecation wrappers.

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8146


michalkurka (approved):
phew

PR #5545, opened 2021-06-17
Report H2OIllegalArgumentException without the stacktrace in Python

michalkurka:


h2o-py/h2o/job.py
```diff
@@ -75,6 +75,9 @@ def poll(self, poll_updates=None):
         if self.status == "CANCELLED":
             raise H2OJobCancelled("Job<%s> was cancelled by the user." % self.job_key)
         if self.status == "FAILED":
+            illegal_arg_ex_prefix = 'water.exceptions.H2OIllegalArgumentException: '
+            if self.exception and self.exception.startswith(illegal_arg_ex_prefix):
+                raise H2OValueError(self.exception[len(illegal_arg_ex_prefix):])
```
sebhrusen, line 80: `H2OValueError` maybe confusing for both dev and user, maybe we can introduce something like `H2OBackendValueError` to make things clear to everyone?

sebhrusen: or the existing `H2OResponseError`

Actually, I think that the default error should also be `H2OResponseError`  instead of  `EnvironmentError`

sebhrusen: Also a lot of backend exceptions are subclasses of `H2OIllegalArgumentException`, so your logic won't work for those: `H2OModelBuilderIllegalArgumentException`, `H2OKeyNotFoundArgumentException`, …
And we could also add `H2OIllegalValueException`.

sebhrusen: We should probably improve this on backend, see `JobV3`:
```
  Throwable ex = job.ex();
    if( ex != null ) status = "FAILED";
    exception = ex == null ? null : ex.toString();
    if (ex!=null) {
      StringWriter sw = new StringWriter();
      PrintWriter pw = new PrintWriter(sw);
      ex.printStackTrace(pw);
      stacktrace = sw.toString();
    }
  ```
  there you can more easily identify the exception, format the `exception` string, and decided to fill the stacktrace.
  Also it would work for all clients of course.

PR #5548, opened 2021-06-22
PUBDEV-8190 : fix multinomial remove collinear columns failure

wendycwong:
This PR address the failure reported in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8190

Basically, when remove_collinear_columns=True, the dimension of gradient and coefficients change when predictors are removed.  Hopefully, I have taken care of all the confusion.

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/ComputationState.java
+++ b/h2o-algos/src/main/java/hex/glm/ComputationState.java
@@ -536,7 +536,7 @@ public final class ComputationState {
     double[] betaCopy = new double[_totalBetaLength]; // make sure fullbeta is full length
     if (_beta.length < _totalBetaLength) {
       if (_beta.length == _activeData.activeCols().length*_nclasses) {  // all classes converted
-        int[] activeCols = genActiveColsAllClass(_beta.length, _betaLengthPerClass, _activeData.activeCols(), c);
+        int[] activeCols = genActiveColsAllClass(_beta.length, _betaLengthPerClass, _activeData.activeCols(), _nclasses);
         fillSubRange(_totalBetaLength, 0, activeCols, _beta, betaCopy);
       } else {
         int[] activeCols = genActiveColsIndClass(_beta.length, _betaLengthPerClass, _activeData.activeCols(), c, _nclasses);
--- a/h2o-algos/src/main/java/hex/glm/ComputationState.java
+++ b/h2o-algos/src/main/java/hex/glm/ComputationState.java
@@ -20,6 +20,7 @@ import water.util.MathUtils;
 
 import java.util.Arrays;
 import java.util.Comparator;
+import java.util.stream.IntStream;
 
 import static hex.glm.GLMUtils.calSmoothNess;
 import static hex.glm.GLMUtils.copyGInfo;
@@ -62,6 +63,7 @@ public final class ComputationState {
   double[][][] _penaltyMatrix;
   int[][] _gamBetaIndices;
   int _totalBetaLength; // actual coefficient length without taking into account active columns only
+  int _betaLengthPerClass;
 
   /**
    *
@@ -79,6 +81,7 @@ public final class ComputationState {
             ((Family.multinomial == parms._family || Family.ordinal == parms._family) ? nclasses : 1);
     _alpha = _parms._alpha[0];
     _totalBetaLength = (dinfo.fullN()+1)*_nclasses;
+    _betaLengthPerClass = dinfo.fullN()+1;
     if (_parms._HGLM) {
       _sumEtaSquareConvergence = new double[2];
       if (_parms._calc_like)
@@ -381,6 +384,16 @@ public final class ComputationState {
 
   public DataInfo activeDataMultinomial(int c) {return _activeDataMultinomial != null?_activeDataMultinomial[c]:_dinfo;}
 
+  /**
+   * This method will return a double array that is extracted from src (which includes active and non-active columns)
+   * to only include active columns stated in ids.
+   * 
+   * @param N
+   * @param c
+   * @param ids
+   * @param src
+   * @return
+   */
   public static double [] extractSubRange(int N, int c, int [] ids, double [] src) {
     if(ids == null) return Arrays.copyOfRange(src,c*N,c*N+N);
     double [] res = MemoryManager.malloc8d(ids.length);
@@ -391,7 +404,17 @@ public final class ComputationState {
     return res;
   }
 
-  private static void fillSubRange(int N, int c, int [] ids, double [] src, double [] dst) {
+  /**
+   * This method will extract coefficients from multinomial.  The extracted coefficients are only from one class
+   * and it contains the active and non-active columns.
+   * 
+   * @param N
+   * @param c
+   * @param ids
+   * @param src
+   * @param dst
+   */
+   static void fillSubRange(int N, int c, int [] ids, double [] src, double [] dst) {
     if(ids == null) {
       System.arraycopy(src,0,dst,c*N,N);
     } else {
@@ -404,7 +427,31 @@ public final class ComputationState {
 
   public double [] betaMultinomial(){return _beta;}
 
-  public double [] betaMultinomial(int c, double [] beta) {return extractSubRange(_activeData.fullN()+1,c,_activeDataMultinomial[c].activeCols(),beta);}
+  public double [] betaMultinomial(int c, double [] beta) {
+     return extractSubRange(_activeData.fullN()+1,c,_activeDataMultinomial[c].activeCols(),beta);
+   }
+   
+   public double[] shrinkFullArray(double[] fullArray) {
+     if (_activeData.activeCols() == null)
+       return fullArray;
+     int[] activeColsAllClass = genActiveColsAllClass(_activeData.activeCols().length*_nclasses, 
+             _betaLengthPerClass);
+     return ArrayUtils.select(fullArray, activeColsAllClass);
+   }
+   
+   public int[] genActiveColsAllClass(int activeColsLen, int numBetaPerClass) {
+     int[] activeCols = new int[activeColsLen];
+     int offset = 0;
+     int[] activeColsOneClass = _activeData._activeCols;
+     for (int classIndex=0; classIndex < _nclasses; classIndex++) {
+       int finalOffset = numBetaPerClass*classIndex;
+       int[] activeCols1Class = IntStream.of(activeColsOneClass).map(i->i+finalOffset).toArray();
+       int num2Copy = activeColsOneClass.length;
+       System.arraycopy(activeCols1Class, 0, activeCols, offset, num2Copy);
+       offset += num2Copy;
+     }
+     return activeCols;
+   }
 
   public GLMSubsetGinfo ginfoMultinomial(int c) {
     return new GLMSubsetGinfo(_ginfo,(_activeData.fullN()+1),c,_activeDataMultinomial[c].activeCols());
@@ -436,6 +483,10 @@ public final class ComputationState {
     }
   }
 
+  /***
+   * This method will grab a subset of the gradient for each multinomial class.  However, if remove_collinear_columns is
+   * on, fullInfo will only contains the gradient of active columns.
+   */
   public static class GLMSubsetGinfo extends GLMGradientInfo {
     public final GLMGradientInfo _fullInfo;
     public GLMSubsetGinfo(GLMGradientInfo fullInfo, int N, int c, int [] ids) {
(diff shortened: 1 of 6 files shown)


h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -569,8 +633,7 @@ public int compare(Integer o1, Integer o2) {
   }
 
   protected boolean checkKKTsMultinomial(){
-    if(_activeData._activeCols == null) return true;
-    throw H2O.unimpl();
+    return true;  // todo: implement this properly
```
koniecsveta, line 636: is this ok?

wendycwong: This is supposed to check if KKT conditions are satisfied.  However, all it is doing is checking if activeCols are null or not which basically does nothing here.  While this function is not actually implemented at all, so it is okay here.  I have written a reminder about actually checking it.  The real check has something to do with checking the actual coefficient value changes.

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -404,8 +417,46 @@ private static void fillSubRange(int N, int c, int [] ids, double [] src, double
...
+    }
+    return betaCnd;
+  }
+  
+  public double[] changeArrayWActiveCols(double[] shortenArray, boolean expand) {
+    if (!expand || (shortenArray != null && !(shortenArray.length == _totalBetaLength))) {  // ginfo is supposed to have full length
+      int[] activeCols = new int[shortenArray.length];
+      int offset = 0;
+      for (int classIndex = 0; classIndex < _nclasses; classIndex++) {
+        int[] activeColsOneClass = _activeDataMultinomial[classIndex].activeCols();
```
tomasfryda, line 445: `_activeDataMultinomial` seems not to be initialized. Some tests such as `tests/testdir_jira/runit_pubdev_1042_LBFGS_with_alpha_and_lambda_search.R` fail here with:

```
        at hex.glm.ComputationState.changeArrayWActiveCols(ComputationState.java:445)
        at hex.glm.ComputationState.updateState(ComputationState.java:787)
        at hex.glm.GLM$GLMDriver.fitLBFGS(GLM.java:1811)
        at hex.glm.GLM$GLMDriver.fitModel(GLM.java:2000)
        at hex.glm.GLM$GLMDriver.computeSubmodel(GLM.java:2445)
        at hex.glm.GLM$GLMDriver.doCompute(GLM.java:2581)
        at hex.glm.GLM$GLMDriver.computeImpl(GLM.java:2479)
        at hex.ModelBuilder$Driver.compute2(ModelBuilder.java:246)
        at hex.glm.GLM$GLMDriver.compute2(GLM.java:1156)
        at water.H2O$H2OCountedCompleter.compute(H2O.java:1637)
```

wendycwong: Yes, good catch!  Fixed.  Thank you.

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -720,15 +786,23 @@ public boolean converged(){
...
+    } else {
+      if (_activeData._activeCols == null)
+        System.arraycopy(beta, 0, _beta, 0, beta.length);
+      else
+        System.arraycopy(changeArrayWActiveCols(beta, false), 0, _beta, 0, oldBeta.length);
+    }
+
+      _ginfo = ginfo;
+      _likelihood = ginfo._likelihood;
+      return (_relImprovement = (objOld - objective()) / Math.abs(objOld));
```
tomasfryda, line 804: Weird, in the second iteration (`_iter == 1`) of this failing test [1], we have lower `objOld` than `objective()` on this line. This leads to the test failure since without early stopping we stop prematurely (`relImprovement < 0 < eps`, line 782 in this file (`converged` method)).

Using absolute value of `_relImprovement` mitigates this problem but I think it might not be a proper solution (since there might be some change in behavior).

[1] `testdir_algos/glm/pyunit_PUBDEV_7611_early_stop_glm_multinomial.py`

wendycwong: Tomas:  there are bugs I fixed that caused the coefficients to shift to the wrong place.  Hopefully this current fix will do the job.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1572,25 +1578,62 @@ private void fitIRLSM_multinomial(Solver s) {
...
+            LineSearchSolver ls;
+            if (_parms._remove_collinear_columns && _state._iter < 1)
+              ls = (_state.l1pen() == 0)  // after first iteration over all classes, state._beta, ginfo.gradient are all smaller size
+                      ? new MoreThuente(_state.gslvrMultinomial(c), _state.betaMultinomialFull(c, beta), _state.ginfoMultinomial(c))
+                      : new SimpleBacktrackingLS(_state.gslvrMultinomial(c), _state.betaMultinomialFull(c, beta), _state.l1pen());
+            else if (_parms._remove_collinear_columns && _state._iter > 0)
+              ls = (_state.l1pen() == 0)  // after first iteration over all classes, state._beta, ginfo.gradient are all smaller size
+                      ? new MoreThuente(_state.gslvrMultinomial(c), _state.betaMultinomialFull(c, beta), _state.ginfoMultinomialRCC(c))
+                      : new SimpleBacktrackingLS(_state.gslvrMultinomial(c), _state.betaMultinomialFull(c, beta), _state.l1pen());
+            else
```
tomasfryda, line 1598: To me, it is hard to see what is different between `iter == 0` and `iter > 0`. I'd find something like this more clear since there is a lot of duplication (but it's just a suggestion, maybe other people prefer it your way):
```suggestion
            if (_parms._remove_collinear_columns)
              ls = (_state.l1pen() == 0)  // after first iteration over all classes, state._beta, ginfo.gradient are all smaller size
                      ? new MoreThuente(_state.gslvrMultinomial(c), _state.betaMultinomialFull(c, beta), (_state._iter < 1) ? _state.ginfoMultinomial(c) : _state.ginfoMultinomialRCC(c))
                      : new SimpleBacktrackingLS(_state.gslvrMultinomial(c), _state.betaMultinomialFull(c, beta), _state.l1pen());
            else
```

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -1572,25 +1578,62 @@ private void fitIRLSM_multinomial(Solver s) {
...
+              gram = _state.computeGramRCC(ls.getX(), s);
+            else
+              gram = _state.computeGram(ls.getX(), s);
 
+            long t3 = System.currentTimeMillis();
+            double[] betaCnd;
+            if (_parms._remove_collinear_columns && _state._iter > 0)
+              betaCnd = ADMM_solve(gram.gram, gram.xy);
+            else 
+              betaCnd = ADMM_solve(gram.gram, gram.xy);  // rcc=true, remove inactive col from _state beta, ginfo
```
tomasfryda, line 1626: This looks like an unnecessary condition or bug:
```suggestion
            betaCnd = ADMM_solve(gram.gram, gram.xy);
```

wendycwong: Tomas:  The two are different.  One is calling computeGramRCC and one is calling computeGram.

wendycwong: You are right.  It is the same!

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Tomas:  With rcc = true and iteration = 0, the dimension of beta and gradient per class has full size.  However, after the admm_solve, the beta and gradient per class will shrink.  So, for a class of 3 with full coefficients of 4, you will begin at 12 beta and 12 gradients.  Then,for class 0,  after first admm_solve, you will have beta of size 11 and gradient of size 11 because the first class has shrunk from 4 to 3.  Then, after you take care of class 1 and after admm_solve, the beta/gradient per class shrunk to 10.  For class 2, after admm_solve, beta/gradient shrunk to 9.  This will complete iteration 0.  Now, when you go to iteration 1, you will be dealing with beta/gradient per class of 9.  In addition, the _state.activeDataMultinomial() will have _adaptedFrame containing only of the active columns only while _state.activeDataMultinomial(c) will contain all the predictors in its _adaptedFrame.  I have to use different functions because at the beginning I can use  _state.activeDataMultinomial(c) since no columns have been made inactive.  However, after iteration 0 is done and at iteration 1, I will need to use  _state.activeDataMultinomial() since the only active columns are used.  

I think the best way to understand it is to go through the code line by line.  I used to have one function to deal with both cases but I made so many mistakes and decided to just write separate methods if I can to make things clean rather than using if/else statements in the methods.  Keeping track of all the beta/gradient dimension changes are exhausting....

tomasfryda: I think I got it but if I did you are missing the point of line 1600 in the suggested edits, namely the `(_state._iter < 1) ? _state.ginfoMultinomial(c) : _state.ginfoMultinomialRCC(c)`. The suggested edit was there just to make it more stand out what differs between the iteration 0 and iteration > 0.

It was just a suggestion that should IMO emphasize that this is the thing that is different because  in those 4 lines I couldn't easily spot the difference but then again you probably didn't in the suggested version so I either is probably fine...

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -403,13 +425,84 @@ private static void fillSubRange(int N, int c, int [] ids, double [] src, double
...
+  
+  public double [] betaMultinomial(int c, double [] beta) {
+     return extractSubRange(_activeData.fullN()+1,c,_activeDataMultinomial[c].activeCols(),beta);
+   }
+
+  public double [] betaMultinomialFull(int c, double [] beta) {
+     if (_parms._remove_collinear_columns)
+      return extractSubRange(_betaLengthPerClass,c,_activeDataMultinomial[c].activeCols(),beta);
+     else
+       return extractSubRange(_activeData.fullN()+1,c,_activeDataMultinomial[c].activeCols(),beta);
```
tomasfryda, line 437: Seems to me that _betaLengthPerClass doesn't change and if !_rcc then I think _activeData.fullN() shouldn't change as well.

If that's so, could we simplify the code to use only `_betaLengthPerClass` so we can get rid of the if statement?

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -459,7 +576,10 @@ public GradientInfo getGradient(double[] beta) {
 
   public void setBetaMultinomial(int c, double [] beta, double [] bc) {
     if(_u != null) Arrays.fill(_u,0);
-    fillSubRange(_activeData.fullN()+1,c,_activeDataMultinomial[c].activeCols(),bc,beta);
+    if (_parms._remove_collinear_columns)
+      fillSubRange(_betaLengthPerClass,c,_activeDataMultinomial[c].activeCols(),bc,beta);
+    else
+      fillSubRange(_activeData.fullN()+1,c,_activeDataMultinomial[c].activeCols(),bc,beta);
```
tomasfryda, line 582: Same here, could we have just `fillSubRange(_betaLengthPerClass,c,_activeDataMultinomial[c].activeCols(),bc,beta);`?

h2o-algos/src/main/java/hex/glm/ComputationState.java
```diff
@@ -403,13 +425,84 @@ private static void fillSubRange(int N, int c, int [] ids, double [] src, double
...
   public GLMSubsetGinfo ginfoMultinomial(int c) {
     return new GLMSubsetGinfo(_ginfo,(_activeData.fullN()+1),c,_activeDataMultinomial[c].activeCols());
   }
 
+  public GLMSubsetGinfo ginfoMultinomialRCC(int c) {
+    if (_activeData.fullN() + 1 == _activeData.activeCols().length)
+      return new GLMSubsetGinfo(_ginfo, (_activeData.fullN() + 1), c, IntStream.range(0, 
+              _activeData.activeCols().length).toArray());
+    else
+      return new GLMSubsetGinfo(_ginfo, (_activeData.fullN() + 1), c, _activeData.activeCols());
```
tomasfryda, line 503: Seems to me that we can keep just the `return new GLMSubsetGinfo(_ginfo, (_activeData.fullN() + 1), c, _activeData.activeCols());`. I assume if we have `_activeData.fullN() + 1 == _activeData.activeCols().length` then `IntStream.range(0, _activeData.activeCols().length).toArray() == _activeData.activeCols()`. 

Or is there some hidden reason why it wouldn't work?

tomasfryda: Tried it and found that active cols could be missing some values even if `_activeData.fullN() + 1 == _activeData.activeCols().length` so my hypothesis was false.

h2o-algos/src/main/java/hex/glm/ComputationState.java
wendycwong: actually _activeData.fullN() changes when inactive columns are removed.  I was confused with this thing in the beginning and it takes me a very long time to figure this one out......

h2o-algos/src/main/java/hex/glm/ComputationState.java
wendycwong: Nope, _activeData.fullN() changes if inactive columns are removed.

h2o-algos/src/main/java/hex/glm/GLM.java
wendycwong: Tomas:  Yes, I missed your point but now I get it.  Good catch.  For some reason, git does not allow me to just commit your code.

tomasfryda (approved):
Looks good to me. Thank you @wendycwong!

PR #5549, opened 2021-06-22
PUBDEV-8207: fix test h statistic failing n gpu

koniecsveta:
Assertion error is caused by different resulting model on gpu vs cpu.

Numerically is h statistic tested here: hex/tree/FriedmanPopescusHTest.java.

In XGBoostTest.java and GBMTest.java it is tested that it can be called from the model, so fixing the problem here by relaxing the test assert to expect either NaN or double in range 0 to 1 - meaning that output is in expected range, since I am not testing here against specific reference number.


tomasfryda (approved):
LGTM! Thank you @koniecsveta!

PR #5550, opened 2021-06-22
[NOCHECK] PUBDEV-8210: Fix failing scale_pca_rf test

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8210

Older sklearn versions (that support py2) don't support `_getitem__` on `Pipeline` object so we should use `Pipeline.named_steps[...]`.


maurever (approved):
Great, thanks @tomasfryda you solved it.

PR #5554, opened 2021-06-25
PUBDEV-8196: fix widget injection for ProgressBar in Py2

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8196

### DIFF
--- a/h2o-py/tests/pyunit_utils/utilsPY.py
+++ b/h2o-py/tests/pyunit_utils/utilsPY.py
@@ -25,10 +25,10 @@ import threading
 import urllib.request, urllib.error, urllib.parse
 import uuid # call uuid.uuid4() to generate unique uuid numbers
 
-try:  
-    from io import StringIO  # py3
+try:
+    from StringIO import StringIO  # py2 (first as py2 also has io.StringIO, but without unicode support)
 except:
-    from StringIO import StringIO  # py2
+    from io import StringIO  # py3
     
 
 try:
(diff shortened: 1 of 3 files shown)


tomasfryda (approved):
Thank you @sebhrusen!

h2o-py/tests/pyunit_utils/utilsPY.py
```diff
@@ -25,10 +25,10 @@
 import urllib.request, urllib.error, urllib.parse
 import uuid # call uuid.uuid4() to generate unique uuid numbers
 
-try:  
-    from io import StringIO  # py3
+try:
+    from StringIO import StringIO  # py2 (first as py2 also has io.StringIO, but without unicode support)
```
sebhrusen, line 29:
```suggestion
    from StringIO import StringIO  # py2 (first as py2 also has io.StringIO, but without string support, only unicode)
```

h2o-py/tests/testdir_utils/pyunit_progressbar.py
```diff
@@ -4,9 +4,9 @@
 from __future__ import absolute_import, division, print_function
 
 try:
-    from io import StringIO   # py3
+    from StringIO import StringIO  # py2 (first as py2 also has io.StringIO, but without unicode support)
```
sebhrusen, line 7:
```suggestion
    from StringIO import StringIO  # py2 (first as py2 also has io.StringIO, but without string support, only unicode)
```

PR #5555, opened 2021-06-25
[nocheck] use of math.log2 in Py test

sebhrusen:
FIX test failing intermittently on Py2 when calling unsupported `math.log2` function


maurever (approved):
Good catch, thanks @sebhrusen!

tomasfryda (approved):
Thank you!

PR #5556, opened 2021-06-26
PUBDEV-8172 predict contributions sorting for generic models

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-8172

Still work in progress, but it is working. We can use PredictContributions interface instead of the whole wrapper.

### DIFF
--- a/h2o-algos/src/main/java/hex/generic/GenericModel.java
+++ b/h2o-algos/src/main/java/hex/generic/GenericModel.java
@@ -186,7 +186,7 @@ public class GenericModel extends Model<GenericModel, GenericModelParameters, Ge
         return new GenericScoreContributionsTask(wrapper)
                 .withPostMapAction(JobUpdatePostMap.forJob(job))
                 .doAll(outputNames.length, Vec.T_NUM, adaptFrm)
-                .outputFrame(outputNames, null);
+                .outputFrame(destination_key, outputNames, null);
     }
 
     class GenericScoreContributionsTask extends MRTask<GenericScoreContributionsTask> {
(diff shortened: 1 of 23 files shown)


h2o-algos/src/main/java/hex/generic/GenericModel.java
```diff
@@ -224,6 +256,63 @@ private void predict(Chunk[] cs, NewChunk[] ncs) throws PredictException {
         }
     }
 
+    private class GenericScoreContributionsSortingTask extends MRTask<GenericScoreContributionsSortingTask> {
+        private final int _topN;
+        private final int _bottomN;
+        private final boolean _compareAbs;
+        private PredictContributions predictContributions;
```
valenad1, line 263: @michalkurka We can use PredictContributions interface instead of the whole wrapper.

h2o-algos/src/main/java/hex/generic/GenericModel.java
```diff
@@ -224,6 +256,63 @@ private void predict(Chunk[] cs, NewChunk[] ncs) throws PredictException {
         }
     }
 
+    private class GenericScoreContributionsSortingTask extends MRTask<GenericScoreContributionsSortingTask> {
+        private final int _topN;
+        private final int _bottomN;
+        private final boolean _compareAbs;
+        private transient PredictContributions predictContributions;
```
valenad1, line 263: @michalkurka We can use PredictContributions interface instead of the whole wrapper.

michalkurka:
Closing - we went in a different direction, thank you anyway @valenad1!

PR #5561, opened 2021-07-02
PUBDEV-8220  DRF fails to create SHAP summary plot

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-8220

IMHO drf with binomial_double_trees behave like multinomial models. Multinomial models are not supported.


valenad1:
Outdated - the solution will be different.

PR #5564, opened 2021-07-09
PUBDEV-8230: Fix GenericModel predictions for non-AUTO categorical encodings

michalkurka:


h2o-core/src/main/java/hex/Model.java
```diff
@@ -1430,7 +1484,7 @@ public double lift_top_group() {
    * @param catEncoded Whether the categorical columns of the test frame were already transformed via categorical_encoding
    */
   public static String[] adaptTestForTrain(final Frame test, final String[] origNames, final String[][] origDomains,
```
michalkurka, line 1486: @sebhrusen any inputs/concerns regarding this change?

michalkurka:
Also adding @sebhrusen - please take a look at the part where I am modifying adaptTestToFrame

h2o-core/src/main/java/hex/Model.java
```diff
@@ -1417,6 +1452,25 @@ public double lift_top_group() {
...
+
+  public interface AdaptFrameParameters {
+    Parameters.CategoricalEncodingScheme getCategoricalEncoding();
+    String getWeightsColumn();
+    String getOffsetColumn();
+    String getFoldColumn();
+    String getResponseColumn();
+    double missingColumnsType();
+    int getMaxCategoricalLevels();
+    default String[] getNonPredictors() {
```
sebhrusen, line 1467: as you're touching this and making it part of an interface, do you think it makes sense renaming this method at the same time according to https://h2oai.atlassian.net/browse/PUBDEV-8078 ?

h2o-core/src/main/java/hex/Model.java
sebhrusen: no, looks good! Actually it increases visibility on what this method is doing, limiting which parameters it's using.

h2o-genmodel/src/main/java/hex/genmodel/CategoricalEncoding.java
```diff
@@ -72,8 +72,24 @@
...
+  /**
+   * Does the categorical encoding have any parameters that are needed to correctly interpret it?
+   * Eg.: number of classes for EnumLimited
+   * 
+   * @return Is this encoding parametrized?
+   */
+  public boolean isParametrized() {
+    return _parametrized;
+  }
+
```
sebhrusen, line 94: I was hoping you would also start to extend this interface to make it easier to convert highly-parametrized TE into an implementation of `CategoricalEncoding` ;)

h2o-core/src/main/java/hex/Model.java
```diff
@@ -281,6 +281,16 @@ public void setTrain(Key<Frame> train) {
...
       boolean needsResponse() { return _needResponse; }
+      public static CategoricalEncodingScheme fromGenModel(CategoricalEncoding encoding) {
+        if (encoding == null)
+          return null;
+        final String name = encoding.name();
+        for (CategoricalEncodingScheme scheme : CategoricalEncodingScheme.values()) {
+          if (name.equals(scheme.name())) 
+            return scheme;
+        }
+        throw new UnsupportedOperationException("Unknown encoding " + encoding);
```
sebhrusen, line 292:
```suggestion
        try {
            return Enum.valueOf(CategoricalEncodingScheme.class, encoding.name());
        } catch (IllegalArgumentException iae) {
            throw new UnsupportedOperationException("Unknown encoding " + encoding);
        }
```
You may not even need to translate the `IllegalArgumentException` given that `encoding.name()` already comes from an enum.
Which btw means that it could have been implemented differently: I would have made `CategoricalEncoding` an interface, and add a `getInstace()` or `newInstance()` method to the `CategoricalEncodingScheme` enum.
This way, no need to translate between 2 enums and the separation of responsibility is made clear between the scheme and the implementation.

sebhrusen (commented):
Looks great. I just made a couple of suggestions that you may not want to address here, but I'm ready to approve if you don't plan to use the occasion to make some changes there.

valenad1 (approved):
Great, Thank you! It seems to me, that we should have test that GeneratedModel give same answers as "normal" model at least for this settings.

```
            @Override
            public String getWeightsColumn() {
                return descriptor.weightsColumn();
            }
            @Override
            public String getOffsetColumn() {
                return descriptor.offsetColumn();
            }
            @Override
            public String getFoldColumn() {
                return descriptor.foldColumn();
            }
            @Override
            public String getResponseColumn() {
                return mojoModel.getResponseName();
            }
            @Override
            public double missingColumnsType() {
                return Double.NaN;
            }
            @Override
            public int getMaxCategoricalLevels() {
                return -1; // returned but won't be used
            }
```

We just have test for categorical encoding now. Maybe I would create a Jira not to forget about it.

h2o-genmodel/src/main/java/hex/genmodel/CategoricalEncoding.java
michalkurka: :)

h2o-core/src/main/java/hex/Model.java
michalkurka: @sebhrusen good point - I was going to do it but then I realized this is going on the fix release branch and such API change shouldn't be there, I will do it in a follow-up PR against master, sounds okay?

sebhrusen (approved):

michalkurka:
@sebhrusen thank you for the review and suggested code change

@valenad1 thank you - I will definitely add a test - I am thinking this could be part of a Model.testJavaScoring testing - I will add in a follow-up PR

PR #5566, opened 2021-07-12
PUBDEV-8172 - refactor PredictCsv to use getter for model

valenad1:
Its a side-effect enhancement, from your https://h2oai.atlassian.net/browse/PUBDEV-8172.

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/tools/PredictCsv.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/tools/PredictCsv.java
@@ -36,7 +36,7 @@ public class PredictCsv {
   boolean returnGLRMReconstruct = false; // for GLRM, return x factor by default unless set this to true
   public int glrmIterNumber = -1;  // for GLRM, default to 100.
   // Model instance
-  private EasyPredictModelWrapper model;
+  private EasyPredictModelWrapper modelWrapper;
 
   public static void main(String[] args) {
     // Parse command line arguments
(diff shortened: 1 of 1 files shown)


h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -63,7 +63,7 @@
  */
 public class EasyPredictModelWrapper implements Serializable {
   // These private members are read-only after the constructor.
-  public final GenModel m;
+  private final GenModel m;
```
michalkurka, line 66: I wouldn't change this - it has no actual benefit and can break code for users => not worth it.

michalkurka (changes requested):
I wouldn't change EasyPredictModelWrapper API

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
valenad1: True, thank you, I didn't think of that 🤦 So it will be just about refactoring on the new method if at least this worth it

h2o-genmodel/src/main/java/hex/genmodel/easy/EasyPredictModelWrapper.java
```diff
@@ -63,7 +63,7 @@
  */
 public class EasyPredictModelWrapper implements Serializable {
   // These private members are read-only after the constructor.
-  public final GenModel m;
```
michalkurka, line 66: I think it should stay `final` - any reason to change?

valenad1: 🤦 no..

PR #5567, opened 2021-07-13
PUBDEV-8221 save mojo with custom filename

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-8221

### DIFF
--- a/h2o-py/h2o/model/model_base.py
+++ b/h2o-py/h2o/model/model_base.py
@@ -1017,19 +1017,23 @@ class ModelBase(h2o_meta(Keyed)):
         path = os.path.join(os.getcwd() if path == "" else path, filename)
         return h2o.api("GET /99/Models.mojo/%s" % self.model_id, data={"dir": path, "force": force})["dir"]
 
-    def save_model_details(self, path="", force=False):
+    def save_model_details(self, path="", force=False, filename=None):
         """
         Save Model Details of an H2O Model in JSON Format to disk.
 
-        :param model: The model object to save.
         :param path: a path to save the model details at (hdfs, s3, local)
         :param force: if True overwrite destination directory in case it exists, or throw exception if set to False.
+        :param filename: a filename for the saved model (file type is always .json)
 
         :returns str: the path of the saved model details
         """
         assert_is_type(path, str)
         assert_is_type(force, bool)
-        path = os.path.join(os.getcwd() if path == "" else path, self.model_id + ".json")
+        if filename is None:
+            filename = self.model_id + ".json"
+        else:
+            assert_is_type(filename, str)
+        path = os.path.join(os.getcwd() if path == "" else path, filename)
         return h2o.api("GET /99/Models/%s/json" % self.model_id, data={"dir": path, "force": force})["dir"]
 
     @staticmethod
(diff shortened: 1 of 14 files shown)


h2o-py/h2o/model/model_base.py
```diff
@@ -996,12 +996,12 @@ def download_mojo(self, path=".", get_genmodel_jar=False, genmodel_name=""):
...
         return h2o.api("GET /3/Models/%s/mojo" % self.model_id, save_to=path)
 
-    def save_mojo(self, path="", force=False):
+    def save_mojo(self, path="", filename=None, force=False):
         """
         Save an H2O Model as MOJO (Model Object, Optimized) to disk.
 
-        :param model: The model object to save.
         :param path: a path to save the model at (hdfs, s3, local)
+        :param filename: a filename for the saved model at (hdfs, s3, local)
```
michalkurka, line 1004: I don't think the filesystem matters for a filename

h2o-py/h2o/model/model_base.py
```diff
@@ -996,12 +996,12 @@ def download_mojo(self, path=".", get_genmodel_jar=False, genmodel_name=""):
                 h2o.api("GET /3/h2o-genmodel.jar", save_to=os.path.join(path, genmodel_name))
         return h2o.api("GET /3/Models/%s/mojo" % self.model_id, save_to=path)
 
-    def save_mojo(self, path="", force=False):
+    def save_mojo(self, path="", filename=None, force=False):
```
michalkurka, line 999: save_mojo and download_mojo should have similar capabilities and semantics - is it still the case?

valenad1: Solved on call

h2o-r/h2o-package/R/export.R
```diff
@@ -327,11 +332,12 @@ h2o.save_mojo <- function(object, path="", force=FALSE) {
 #' # h2o.saveModelDetails(object = prostate_glm, path = "/Users/UserName/Desktop", force = TRUE)
 #' }
 #' @export
-h2o.saveModelDetails <- function(object, path="", force=FALSE) {
+h2o.saveModelDetails <- function(object, path="", force=FALSE, filename=paste0(object@model_id,".json")) {
```
michalkurka, line 335: great you noticed this one as well

h2o-r/h2o-package/R/export.R
```diff
@@ -240,12 +241,13 @@ h2o.save_frame <- function(x, dir, force = TRUE) {
 #' # h2o.saveModel(object = prostate_glm, path = "/Users/UserName/Desktop", force = TRUE)
 #' }
 #' @export
-h2o.saveModel <- function(object, path="", force=FALSE, export_cross_validation_predictions=FALSE) {
+h2o.saveModel <- function(object, path="", force=FALSE, export_cross_validation_predictions=FALSE, filename=object@model_id) {
```
michalkurka, line 244: wouldn't this fail in a non-obvious way if object is not actual H2OModel?

it basically means the validation doesn't show the error message if the object is not what is expected

michalkurka (approved):
Thank you - please look into improving the R side - I left a comment

h2o-r/h2o-package/R/export.R
valenad1: True, thanks, I have updated all save and download methods and add test to catch this case 👍

h2o-r/h2o-package/R/export.R
valenad1: 👀

PR #5569, opened 2021-07-13
PUBDEV-8233: Add StackedEnsembles To AutoML's Time Budget

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8233

Currently, StackedEnsembles ignore the time constraint from AutoML which is fine for a lot of situations since the metalearner is usually very fast to train. 

However, in some situations SEs take much more time than expected. This behavior was observed mainly on multinomial tasks. One good example is 1h training on Covertype dataset which very often takes 1h30mins and the main contributing factor is the StackedEnsemble.

The goal of this PR is to speed up automl training in cases where it uses more time than it should, this will possibly influence the performance of the SE models in these cases so another goal is to have SE in the top 5 models in 95+% of the datasets we use for benchmark.

# Results

## Multinomial 
In multinomial tasks, this PR fails to comply with the rule "95% of runs should have SE in the top 5". But I compared it with previous experiments and the amount of runs with SE in the top 5 increased.

To get more robust comparison I made a plot for the ratio of SE presence in top N - N is on x-axis. Apart from using all old runs I also used `logit_on` PR run as that is IIRC the last PR that we benchmarked that got merged (at least on multinomial datasets).

Unknown-2

I also looked at how big the impact from limiting SE run time on logloss is and there doesn't seem to be a big one - in CoverType, I think there might be some increase in logloss since we use less time by order of magnitude. In other cases, especially those where we finish before the time limit runs out, I think the differences in logloss might be just caused by noise. In the table below I show only the increases of logloss:
Screen Shot 2021-07-14 at 14 41 31

Trained SEs per dataset (should be == 20):
Screen Shot 2021-07-15 at 19 31 37

## Binomial

For binomial, I got quite a lot of failures (121 lines in `failures.csv`) but I think the general trend seems to be the same as in the other task.

Screen Shot 2021-07-15 at 18 58 47

Cases where the performance was worse:
Screen Shot 2021-07-15 at 18 59 19

Trained SEs per dataset (should be == 20):
Screen Shot 2021-07-15 at 18 59 37

## Regression
41 lines in `failures.csv` so this task also has some "missing data".
Screen Shot 2021-07-15 at 18 33 38

Cases where the performance was worse:
Screen Shot 2021-07-15 at 18 33 55

Trained SEs per dataset (should be == 20):
Screen Shot 2021-07-15 at 18 34 28

# Summary
Even though we don't have all the data due to some failures, I think it is safe to say that this PR doesn't make AutoML much worse (quite a lot of tasks end up actually tiny bit better) and at the same time it cuts down the longest running runs so I think it's safe to say that we can merge it.

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
@@ -316,7 +316,8 @@ public class StackedEnsemble extends ModelBuilder<StackedEnsembleModel,StackedEn
       List<Frame> baseModelPredictions = new ArrayList<>();
 
       for (Key<Model> k : baseModelKeys) {
-        if (stop_requested())
+        // if training we can stop prematurely due to a timeout but computing validation scores should be allowed to finish
+        if (stop_requested() && isTraining)
           throw new Job.JobCancelledException();
 
         if (_model._output._metalearner == null || _model.isUsefulBaseModel(k)) {
--- a/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
+++ b/h2o-algos/src/main/java/hex/ensemble/StackedEnsembleModel.java
@@ -4,8 +4,6 @@ import hex.*;
 import hex.genmodel.utils.DistributionFamily;
 import hex.genmodel.utils.LinkFunctionType;
 import hex.glm.GLMModel;
-import hex.quantile.Quantile;
-import hex.quantile.QuantileModel;
 import hex.tree.drf.DRFModel;
 import water.*;
 import water.exceptions.H2OIllegalArgumentException;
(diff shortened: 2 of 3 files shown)


sebhrusen:
@tomasfryda Did you find what's preventing SE to not obey the timeout/stop request on some occasion?

tomasfryda:
@sebhrusen yes, it is the GLM. Basically, it will stop after the GLM's "outer" iteration finishes and the problem is that GLM uses `hex.optimization.*` to do the inner iterations so it is not straightforward to check there for `ModelBuilder#stopRequested` (and we would have to make sure it makes sense stopping the model there, e.g., we probably don't want to finish after we have trained `beta`s for just one class etc.). I want to discuss this in more detail with @wendycwong since the GLM code is very branched and it's hard for me to guess how hard it would be to implement this without breaking something.

But there are some more pressing issues with GLM such as the [L-BFGS + non-negative constrains for multinomial tasks](https://h2oai.atlassian.net/browse/PUBDEV-8182) (reason why we end up with no SE in Helena dataset) and IIRC the longest time above the 1h time limit was less than 6mins which is a big improvement when compared with the run without this change which took in total 1h56mins (so 50 mins more).

sebhrusen (approved):
Given the results, looks good for merge. Thank you @tomasfryda.
Please keep the failures from the experiments for further analysis: we shouldn't discard those.

ledell (approved):
This is a good improvement, thank you!

PR #5571, opened 2021-07-14
PUBDEV-8237: Fix h2odriver - make sure no invalid GC options are added on newer Javas

michalkurka:


michalkurka:
@mn-mikke in this PR I hacked the Java 11 support just to have test coverage - however, it should be cleaned up and baked into the docker images instead - https://h2oai.atlassian.net/browse/ITA-997

mn-mikke (approved):

michalkurka:
thank you for the review @mn-mikke

PR #5577, opened 2021-07-20
PUBDEV-8242: Fix splits NA vs Rest to be correctly represented

michalkurka:
Categorical splits NAvsREST are not represented by a bitset in the
CompressedTrees - there is a bunch of places in the code where the bitset is
removed. However, the type of split stays categorical equal = 2 or 3.

This means the scoring logic interprets the split as categorical. While
no bitset is actually parsed from the tree, the scoreTree method can
erroneously re-use a previous bitset split instead of creating a fresh new
one.

This will result in misclassification of the observations in the tree nodes.

### DIFF
--- a/h2o-algos/src/test/java/hex/tree/drf/DRFTest.java
+++ b/h2o-algos/src/test/java/hex/tree/drf/DRFTest.java
@@ -26,6 +26,7 @@ import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileWriter;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static org.junit.Assert.*;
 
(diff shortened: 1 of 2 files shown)


h2o-algos/src/test/java/hex/tree/drf/DRFTest.java
```diff
@@ -1949,6 +1950,91 @@ private void checkDeepLeafNodeAssignmentConsistency(int depth, int extraObservat
...
+              .withColNames("C0", "C1", "response", "weight")
+              .withVecTypes(Vec.T_CAT, Vec.T_CAT, Vec.T_NUM, Vec.T_NUM)
+              .withDataForCol(0, dataCol2)
+              .withDataForCol(1, dataCol1)
+              .withDataForCol(2, response)
+              .withDataForCol(3, weights)
+              .build();
+      f.vec(0).setNA(0);
+
+      System.out.println(f.toTwoDimTable(0, (int) f.numRows(), false));
```
valenad1, line 1987: I don't know if we are ok with souths in tests, I would rather use LOG.info

h2o-algos/src/test/java/hex/tree/drf/DRFTest.java
```diff
@@ -1949,6 +1950,91 @@ private void checkDeepLeafNodeAssignmentConsistency(int depth, int extraObservat
...
+
+      DRFModel model = new DRF(parms).trainModel().get();
+      Scope.track_generic(model);
+
+      Map<Integer, Integer> actualWeights = new HashMap<>();
+      
+      Frame nodeIds = model.scoreLeafNodeAssignment(
+              f, Model.LeafNodeAssignment.LeafNodeAssignmentType.Node_ID, Key.make());
+      nodeIds.add(f);
+      System.out.println(nodeIds.toTwoDimTable(0, (int) f.numRows(), false));
```
valenad1, line 2008: same here

valenad1 (approved):
Looks good to me, thanks!

h2o-algos/src/test/java/hex/tree/drf/DRFTest.java
michalkurka: LOG.info will make it log 2x - once in H2O logs and the second time in stdout. Since we only collect the stdout and the actual logs will roll off anyway - isn't sysout better?

PR #5578, opened 2021-07-21
PUBDEV-7354: improve and consolidate handling of version mismatch between Py and Backend

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-7354 : standard version mismatch handling on `h2o.init`/`h2o.connect`. Defaults to error with `h2o.init` and warning with `h2o.connect` (for backwards compatibility).
We can easily change the default behavior for `h2o.connect` in a future version.

https://h2oai.atlassian.net/browse/PUBDEV-5300 : ensure there's no confusing error message when pinging the cluster due to a version mismatch, even if user disabled the version check. Errors will be raised only if user requires a feature that is not available on backend.


h2o-py/h2o/backend/cluster.py
```diff
@@ -290,16 +223,48 @@ def timezone(self, tz):
...
-
     def list_timezones(self):
         """Return the list of all known timezones."""
         from h2o.expr import ExprNode
         return h2o.H2OFrame._expr(expr=ExprNode("listTimeZones"))._frame()
-
-
-    #-------------------------------------------------------------------------------------------------------------------
+    
+    def check_version(self, strict=False):
```
sebhrusen, line 231: moved from `h2o` module: sits more comfortably here

h2o-py/h2o/schemas/error.py
```diff
@@ -6,153 +6,51 @@
...
-    @payload.setter
-    def payload(self, value):
-        self._payload = value
-
-    def __repr__(self):
+            if k in _ignored_schema_keys: continue
+            if not hasattr(cls, k):
+                # we can add properties dynamically here as they are defined statically on the backend
+                # as properties of water.api.schemas3.H2OErrorV3 (or subclasses)
+                setattr(cls, k, property(partial(cls.__getitem__, name=k)))
```
sebhrusen, line 28: to avoid having to raise an unnecessary error when there is a version mismatch between PY client and backend, the properties are now generated on the fly on the first call, adapting dynamically to the backend schema.
This also avoid developpers to update this code when the schema is modified.

h2o-py/h2o/backend/cluster.py
```diff
@@ -29,42 +31,30 @@ class H2OCluster(object):
...
-            if k in {"__meta", "_exclude_fields", "__schema"}: continue
-            if k in _cloud_v3_valid_keys:
-                obj._props[k] = v
-            else:
-                raise AttributeError("Attribute %s cannot be set on H2OCluster (= %r)" % (k, v))
+            if k in _ignored_schema_keys: continue
+            if not hasattr(cls, k):
+                # we can add properties dynamically here as they are defined statically on the backend
+                # as properties of water.api.schemas3.CloudV3
+                setattr(cls, k, property(partial(cls.__getitem__, name=k)))
```
sebhrusen, line 47: to avoid having to raise an unnecessary error when there is a version mismatch between PY client and backend, the properties are now generated on the fly on the first call, adapting dynamically to the backend schema.
This also avoid developers to update this code when the schema is modified.

g-eoj (approved):
Clever way of setting the properties! I'm not familiar with the h2o-3 project but I don't see any red flags. I'm assuming that any method in `backend` can be changed without concerns of breaking the public api?

h2o-py/h2o/backend/cluster.py
sebhrusen: @michalkurka as it is potentially the most impacting change in this PR: do you validate this?

michalkurka: @sebhrusen I believe the intention here was the exact list on the client so that there is a clear understanding of what each returned property means.

My preference would be to keep the list static and well-defined. IMHO this is a bad property to have "This also avoid developers to update this code when the schema is modified.". If something is added to the cloud response it should definitely fail - why is it added if it is not being used? should it be added in the first place? I think the list of the properties establishes a "protocol" that the client and cluster can rely on.

I think we should definitely fail if expected properties are not found. If there are some additional properties we can consider allowing user to override using the strict_version_check. To override missing properties I think users would need to switch to "god mode".

My opinion :)

sebhrusen: My point is that the version check should be done when connecting to the backend, so any additional error at runtime because the cluster schema on the backend doesn't match the cluster schema on the frontend is just an hindrance: user may still want to use `strict_version_check=True` but in practice it will still fail anyway on the first 'ping'.

So even if the static API for this particular component was generated the same way as for estimators, we would still have this issue. That's why, by generating the properties on the fly, I'm just suggesting to use the REST schema as the single point of reference: I could also make a request to `/3/Metadata/schemas/CloudV3` (or sth like that) to generate the properties from there if you think it's cleaner (I think it is, but the result is equivalent): that's what we do in Flow for example.

sebhrusen: > If something is added to the cloud response it should definitely fail - why is it added if it is not being used? should it be added in the first place? I think the list of the properties establishes a "protocol" that the client and cluster can rely on.

@michalkurka I think it makes sense only if you put the client at the same level as the backend. But I consider the REST API as the only point of reference / contract that both backend and clients should respect.
Given this, if something is available in the schema, the client should adapt.

michalkurka: > strict_version_check=True but in practice it will still fail anyway on the first 'ping'.

it shouldn't fail on with a non-descript message, also I think having an override to allow new unrecognized properties is fine as well

> I think it makes sense only if you put the client at the same level as the backend
yes, exactly my thinking - the client and backend are in normal circumstances "one entity" - if someone wants to use strict_version_check to debug something and mix it  - fine with me but it should at least alert them there is a mismatch on the fundamental information that the client and backend exchanges

Flow is not a client in the same sense as the others are - Flow is inherently tied to the same H2O version, it doesn't exist beyond the same version actually; it is just dynamically doing what eg. "gen_R.py" is doing at compile time

Again - this is my preference - I am not going to force you to do it that way.

sebhrusen: > Again - this is my preference - I am not going to force you to do it that way.

I understand, but I need your always pertinent ;) arguments before deciding what I want to do.

sebhrusen: FYI, I think I'll keep it dynamic, but will extract the properties directly from the `Metadata/schema` API, and also address your concern regarding the error message when client tries to access unknown properties: I agree that in this case the error message should be descriptive and inform that he's trying to use a property that is not available for the specific backend to which it is connected.

h2o-py/h2o/backend/cluster.py
maurever: Why do you think `check_version` is better name than `version_check`? I am just curious. :)

h2o-py/h2o/exceptions.py
```diff
@@ -115,7 +115,10 @@ class H2OConnectionError(H2OSoftError):
 # This should have been extending from Exception as well; however in old code version all exceptions were
 # EnvironmentError's, so for old code to work we extend H2OResponseError from EnvironmentError.
 class H2OResponseError(H2OError, EnvironmentError):
-    """Raised when the server encounters a user error and sends back an H2OErrorV3 response."""
+    """
+    Raised when the server encounters a user error and sends back an H2OErrorV3 response.
+    H2OResponseError should always be constructed with
```
maurever, line 120: Is this sentence complete?

h2o-py/h2o/h2o.py
```diff
@@ -2394,38 +2344,38 @@ def estimate_cluster_mem(ncols, nrows, num_cols = 0, string_cols = 0, cat_cols =
...
+
+    if uuid_cols < 0:
         raise ValueError("uuid_cols can't be a negative number")
-    
-    BASE_MEM_REQUIREMENT_MB = 32
-    SAFETY_FACTOR = 4
-    BYTES_IN_MB = 1024 * 1024
-    BYTES_IN_GB = 1024 * BYTES_IN_MB
+
+    base_mem_requirement_mb = 32
```
maurever, line 2371: Why do you rename these constants? It looks like constants, or am I missing something?

maurever (approved):
I can't see any problem, good job and thanks, @sebhrusen. I have just a few questions which are not crucial.

h2o-py/h2o/backend/cluster.py
sebhrusen: For functions that do something, I think we should always prefer verbs (like `list_timezones` right above) and keep nouns to represent states or abstractions.
Java developers especially tends to abuse nouns: for an old classical blog post about the difficult (and I agree sometimes painful) relation between Java and nouns, I recommend https://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html :)

h2o-py/h2o/exceptions.py
sebhrusen: apparently not, good catch thanks!

h2o-py/h2o/h2o.py
sebhrusen: If those were defined at module level, or even class level, these would be constants and could be all upper-case and respect the Py coding style (https://www.python.org/dev/peps/pep-0008/#constants), but they are defined inside a function and as such, they're just variables, so they don't follow the PEP, and IDEs like IntelliJ keep complaining/underlining, creating visual noise and this works pretty well because it's annoying me and I fix it to respect the PEP :)

maurever: Oh, I did not check if they are in the function! So then you are absolutely correct. :)

h2o-py/h2o/backend/cluster.py
maurever: So in this case, we should rename also the `_strict_version_check()` function. ;) I am just kidding, I understand, what you want to say.

sebhrusen: you know what? I was thinking exactly the same when I noticed it a few minutes later.and I also remember having difficulties to find a name for that (private) function: it surely shows that my Java background is still there in a corner of my brain :)
But to my defense:
- it's a private method, and I care less than for public ones.
- it's actually not doing anything really: the idea is to return the "real" values of the `strict_version_check` parameter after considering the config and/or environment settings.
Maybe it should be an adverb :)

PR #5579, opened 2021-07-21
[nocheck] Remove stacked_ensembles_are_trained_after_timeout test

tomasfryda:
Recent change includes SEs in the automl's `max_runtime_secs` which sometimes (especially on short runs) means that we won't have all or even any SE trained.


valenad1 (approved):
Thank you!

PR #5580, opened 2021-07-21
PUBDEV-8244: Fix CoxPH mojo import - shouldn't fail when printing model

michalkurka:


h2o-py/h2o/model/metrics_base.py
```diff
@@ -196,8 +196,8 @@ def show(self):
             
         if metric_type in types_w_cox:
             print("Concordance score: " + str(self.concordance()))
-            print("Concordant count: " + str(self.concordance()))
-            print("Tied cout: " + str(self.concordance()))
+            print("Concordant count: " + str(self.concordant()))
+            print("Tied cout: " + str(self.tied_y()))
```
sebhrusen, line 200: the previous values were suspicious indeed :))

h2o-py/tests/testdir_generic_model/pyunit_generic_model_mojo_cox_ph.py
```diff
@@ -26,17 +26,25 @@ def test(output_test, x, ties, stratify_by, use_all_factor_levels):
...
-    with Capturing() as original_output:
-        coxph.show()
-        
     coxph.show()
 
-    original_model_filename = tempfile.mkdtemp()
-    original_model_filename = coxph.download_mojo(original_model_filename)
-      
-    model = H2OGenericEstimator.from_file(original_model_filename)
+    mojo_path = tempfile.mkdtemp()
```
sebhrusen, line 31: it's usually recommended to clean up tmp folders automatically:
```python
with pyunit_utils.TemporaryDirectory() as tmpd:
    mojo_path = coxph.download_mojo(tmpd)

    from h2o.estimators import H2OGenericEstimator
    …
```

h2o-py/h2o/estimators/estimator_base.py
```diff
@@ -560,7 +561,7 @@ def _metrics_class(model_json):
             model_class = H2OAnomalyDetectionModel
         elif model_type == "CoxPH":
             metrics_class = H2OModelMetricsRegressionCoxPH
-            model_class = H2OCoxPHModel
+            model_class = H2OCoxPHModel if model_json["algo"] != "generic" else H2OCoxPHMojoModel
```
sebhrusen, line 564: do we keep the "generic" solution for Mojo models for a follow up ticket? or was it an issue mainly for CoxPH?

Meaning… couldn't we override `model_class` globally after this sequence of if-else:
```python
if model_json['algo'] == 'generic':
    model_class = ModelBase   # or define a GenericModel class
```

sebhrusen (commented):
overall looks good, added some minor comments:

h2o-py/h2o/estimators/estimator_base.py
michalkurka: @sebhrusen we cannot do it globally because the more common models like GBM behave okay - that would be a regression in behavior

I do not believe CoxPH is the only model to suffer from an issue but we need to test it first

In this fix, I wanted to be just focused on the issue that was reported.

h2o-py/tests/testdir_generic_model/pyunit_generic_model_mojo_cox_ph.py
michalkurka: @sebhrusen thanks for the suggestion, you are right

I tried your solution but it is not available on the release branch - I used the "results" directory where we typically store files like that

michalkurka:
@sebhrusen please take another look, I fixed the test

sebhrusen (approved):
Looks good!
Would be nice to create a follow-up ticket to handle generic/mojo models consistently, list expectations regarding those models, and also potential issues with current logic.

michalkurka:
@sebhrusen agrees, we have jira to track this - https://h2oai.atlassian.net/browse/PUBDEV-8256

thank you for the review and suggestions

PR #5581, opened 2021-07-22
PUBDEV-8218: fix h2o interaction outcomes

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8218

it was considering some cases as a self-interaction when it was not true


wendycwong (approved):
Great job for catching that we are comparing the wrong thing@

PR #5586, opened 2021-07-26
PUBDEV-8250: Monotone constraints in GBM can cause issues with reproducing the model

michalkurka:
If a dataset has highly correlated columns and monotonicity constraints are used on one of them, independent attempts to build the same model with identical parameters can lead to slightly different models. This is caused by non-determinism in histogram building - this non-determinism is compensated for split points but monotone constraints are not leveraging the same mechanism.

The solution is to avoid any non-determinism when monotone constraints are used.


koniecsveta (approved):

koniecsveta:
is 3.32.1.5 the correct fix release when pr is against master?

michalkurka:
@koniecsveta yes, in this special case, that is correct

only if validated on master it will be backported back to rel-zipf

michalkurka:
@koniecsveta thank you for noticing!

PR #5587, opened 2021-07-27
PUBDEV-8492: fix GLM zero weight prediction error

wendycwong:
This PR fixed the bug caught by Zuzana: https://h2oai.atlassian.net/browse/PUBDEV-8249.

When there is a weight vector and when its value == 0, two things happen:
1. The extractDenseRow will not read in a new row when weight == 0.  However, this problem does not exist for extractSparseRow;
2. During scoring, if weight == 0, all prediction output will be 0.

I remove the check for weight == 0 in both cases and ran Zuzana's test.  It is working now.


wendycwong:
This is expected behavior when weight == 0.  Users should not use a weight column that contains zeros.

PR #5589, opened 2021-07-28
PUBDEV-8239: weird convert to data frame behaviour

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8239

### DIFF
--- a/h2o-core/src/main/java/water/api/DatasetServlet.java
+++ b/h2o-core/src/main/java/water/api/DatasetServlet.java
@@ -32,6 +32,8 @@ public class DatasetServlet extends HttpServlet {
           .setEscapeQuotes(Boolean.parseBoolean(escape_quotes_string));
       InputStream is = dataset.toCSV(parms);
       response.setContentType("application/octet-stream");
+      // todo: get this somehow from dataset
+      response.setCharacterEncoding("UTF-8");
       // Clean up the file name
       int x = f_name.length() - 1;
       boolean dot = false;
--- a/h2o-core/src/main/java/water/api/DatasetServlet.java
+++ b/h2o-core/src/main/java/water/api/DatasetServlet.java
@@ -1,7 +1,5 @@
 package water.api;
 
-import com.ibm.icu.text.CharsetDetector;
-import com.ibm.icu.text.CharsetMatch;
 import water.DKV;
 import water.fvec.Frame;
 import water.server.ServletUtils;
@@ -11,10 +9,9 @@ import water.util.Log;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import java.io.BufferedInputStream;
-import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.charset.Charset;
 
 /**
  */
@@ -34,10 +31,9 @@ public class DatasetServlet extends HttpServlet {
       Frame.CSVStreamParams parms = new Frame.CSVStreamParams();
       if (hex_string != null) parms.setHexString(Boolean.parseBoolean(hex_string));
       if (escape_quotes_string != null) parms.setEscapeQuotes(Boolean.parseBoolean(escape_quotes_string));
-      InputStream is = new BufferedInputStream(dataset.toCSV(parms));
+      InputStream is = dataset.toCSV(parms);
       response.setContentType("application/octet-stream");
-      String encoding = detectEncoding(is);
-      response.setCharacterEncoding(encoding);
+      response.setCharacterEncoding(Charset.defaultCharset().name());
       // Clean up the file name
       int x = f_name.length() - 1;
       boolean dot = false;
@@ -70,23 +66,5 @@ public class DatasetServlet extends HttpServlet {
     } finally {
       ServletUtils.logRequest("GET", request, response);
     }
-  }
-
-  String detectEncoding(InputStream is) throws IOException {
-    String charset = null;
-    if (is.markSupported()) {
-      CharsetDetector cd = new CharsetDetector();
-      is.mark(0);
-      cd.setText(is);
-      CharsetMatch cm = cd.detect();
-      if (cm != null) {
-         charset = cm.getName();
-      }
-    }
-    if (charset == null || !is.markSupported()) {
-      charset = "UTF-8";
-    }
-    is.reset();
-    return charset;
   }
 }
--- a/h2o-core/src/main/java/water/api/DatasetServlet.java
+++ b/h2o-core/src/main/java/water/api/DatasetServlet.java
@@ -31,7 +31,11 @@ public class DatasetServlet extends HttpServlet {
           .setHexString(Boolean.parseBoolean(hex_string))
           .setEscapeQuotes(Boolean.parseBoolean(escape_quotes_string));
       InputStream is = dataset.toCSV(parms);
-      response.setContentType("application/octet-stream");
+      if (uri.equals("/3/DownloadDataset")) {
+        response.setContentType("text/csv");
+      } else {
+        response.setContentType("application/octet-stream");
+      }
       // todo: get this somehow from dataset
       response.setCharacterEncoding("UTF-8");
       // Clean up the file name
--- a/h2o-core/src/main/java/water/api/DatasetServlet.java
+++ b/h2o-core/src/main/java/water/api/DatasetServlet.java
@@ -31,7 +31,7 @@ public class DatasetServlet extends HttpServlet {
           .setHexString(Boolean.parseBoolean(hex_string))
           .setEscapeQuotes(Boolean.parseBoolean(escape_quotes_string));
       InputStream is = dataset.toCSV(parms);
-      if (uri.equals("/3/DownloadDataset")) {
+      if (request.getServletPath().equals("/3/DownloadDataset")) {
         response.setContentType("text/csv");
       } else {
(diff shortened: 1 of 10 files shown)


sebhrusen:
@koniecsveta did you revert all your previous changes? was it just an encoding issue at the end?

koniecsveta:
> @koniecsveta did you revert all your previous changes? was it just an encoding issue at the end?
> 
> yes but this is not ready yet, will do some more commits. Requests lib is wrongly encoding "content" to "text" (text is content in unicode according to this: https://docs.python-requests.org/en/master/api/#requests.Response.text)
> Screenshot 2021-08-02 at 13 36 01
> 
> and that is because it wrongly autodetects character encoding of "content" (because we are not passing any). So we should pass it, so that it doesn't need to be autodetected. And also ContentType in water/api/DatasetServlet.java is weird.

@sebhrusen sry I edited your comment instead of responding to it x)

sebhrusen:
@koniecsveta yes, we could expect that a csv file would be returned with a difference content-type (text/csv for example) but if you plan to change this, I would strongly advise you to test the effect in Flow in various use-cases.
the fact is that in a browser,`application/octet-stream` will force it to ask the user to download the response/file whereas with  a `text` content-type, it will try to render it in the page, leading to a very different user experience.

I think octet-stream is fine for a REST API, what we need to ensure though is that the csv content is always encoded in UTF-8.

sebhrusen:
At the same time we have:
```
    new ServletMeta("/3/DownloadDataset", DatasetServlet.class),
    new ServletMeta("/3/DownloadDataset.bin", DatasetServlet.class),
```
so we could have mapped the `.bin` entry point to `application/octet-stream` and the other to `text/csv`: this would have been consistent and probably expected bahaviour…

h2o-core/src/main/java/water/api/DatasetServlet.java
```diff
@@ -32,6 +32,8 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) {
           .setEscapeQuotes(Boolean.parseBoolean(escape_quotes_string));
       InputStream is = dataset.toCSV(parms);
       response.setContentType("application/octet-stream");
+      // todo: get this somehow from dataset
```
sebhrusen, line 35: I think H2O parser always convert data to `utf-8`: needs confirmation though.

sebhrusen: it may be more complicated than that: https://docs.h2o.ai/h2o/latest-stable/h2o-docs/getting-data-into-h2o.html?highlight=utf

this is true for CSV/arff imports, bot for example, hard to know what happens with JDBC imports…

koniecsveta: thanks!

sebhrusen: most extended ascii encodings (except the always annoying windows-1252 probably) should convert nicely to `utf-8`, the main issue is with `utf-16` and windows then: and it's not clear to  me if H2O supports those encodings today.

koniecsveta: I've found this check: water.parser.ParseSetup#checkEncoding, from this it seems like UTF16 is not supported

sebhrusen: It's not supported from parsers, correct. But we have other types of import, for example jdbc (direct import from RDBS, or also for Hive tables) or Hive import from hdfs who don't require any parsing if I'm correct…

sebhrusen: Apparently the Hive import calls `guessSetup` so it will detect and reject `utf-16`.
Then even if it's not checked in `SQLManager`, I guess we can say that H2O doesn't support `utf-16`.

koniecsveta:
found out that when content type in header is now set to "text/csv" instead of "application/octet-stream", requests autodetection of character encoding works well. Hence it appears that there is no need to solve character encoding detection beforehand on our backend and setting up more specific content type is the only fix needed

h2o-core/src/main/java/water/api/DatasetServlet.java
```diff
@@ -31,7 +31,11 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) {
           .setHexString(Boolean.parseBoolean(hex_string))
           .setEscapeQuotes(Boolean.parseBoolean(escape_quotes_string));
       InputStream is = dataset.toCSV(parms);
-      response.setContentType("application/octet-stream");
+      if (uri.equals("/3/DownloadDataset")) {
```
sebhrusen, line 34: `request.getServletPath()` is more appropriate than `uri` if you're going to use `equals`

sebhrusen (commented):
please also double-check Flow: it may need to use `/3/DownloadDataset.bin` to keep working as before.
My only concern is in the case were flow offers a button/link to download data in csv format, then you'll want the content-type to be `application/octet-stream`

h2o-core/src/main/java/water/api/DatasetServlet.java
sebhrusen: another way would be to subclass `DownloadDataset` with `DownloadDatasetBin` and have a final content type defined for each...

koniecsveta:
> please also double-check Flow: it may need to use `/3/DownloadDataset.bin` to keep working as before.
> My only concern is in the case were flow offers a button/link to download data in csv format, then you'll want the content-type to be `application/octet-stream`

unfortunately, in flow
 `
http://localhost:54321/3/DownloadDataset.bin?frame_id=as_df_err_parquet.hex
`
 neither

`http://localhost:54321/3/DownloadDataset?frame_id=as_df_err_parquet.hex`

save the file correctly (whereas R/py does)

so need to continue fixing

sebhrusen:
> unfortunately, in flow
http://localhost:54321/3/DownloadDataset.bin?frame_id=as_df_err_parquet.hex
[does not] save the file correctly (whereas R/py does)

Did it ever work? With `.bin` it should be equivalent to the legacy behaviour, right?
Do you mean that it doesn't save it with the right encoding or it doesn't prompt the native save popup?

koniecsveta:
Doesnt save it with the right encoding

koniecsveta:
fix for flow is here: https://github.com/h2oai/h2o-flow/pull/177 

It doesn't matter whether there is `/3/DownloadDataset` or `/3/DownloadDataset.bin` - the save popup prompts in both cases, so I left what was there before. But it needs `hex_string=false&escape_quotes=true` as it is in `h2o.frame.H2OFrame.get_frame_data` to save it with the right encoding

koniecsveta:
h2o-py/tests/testdir_jira/pyunit_pubdev_6439.py and h2o-py/tests/testdir_jira/pyunit_pubdev_6394.py failing with content type == "text/csv"

h2o-py/tests/testdir_jira/pyunit_pubdev_8239.py
```diff
@@ -0,0 +1,27 @@
...
+import h2o
+from tests import pyunit_utils
+
+
+def pubdev_8239():
+    df = h2o.import_file("http://h2o-public-test-data.s3.amazonaws.com/smalldata/parser/parquet/as_df_err.parquet", header=1)
+
+    # check that content doesn't become corrupted when downloading line 561:
+    # this was ok
+    assert df.head(rows=560).get_frame_data().__contains__("~{fy?$ZWN")
```
sebhrusen, line 12:
```python
"str" in df.head(rows=560).get_frame_data()
```
is the proper syntax

h2o-py/tests/testdir_jira/pyunit_pubdev_8239.py
```diff
@@ -0,0 +1,27 @@
...
+    # here it become broken in failing case
+    assert df.head(rows=561).get_frame_data().__contains__("~{fy?$ZWN")
+    assert df.get_frame_data().__contains__("~{fy?$ZWN")
+    df3 = df.as_data_frame(use_pandas=False, header=False)
+    assert df3[2][0].__contains__("~{fy?$ZWN")
+    assert len(df3) == df.nrow
```
sebhrusen, line 18: please compare both rows and columns: with the escape char/quotes issue — which can occur when downloading data —, I've seen datasets being entirely stripped away from some columns or with shifted data because of this.
It may be worth comparing entire randow rows.

h2o-r/tests/testdir_parser/runit_PUBDEV_8239.R
```diff
@@ -0,0 +1,15 @@
...
+)))
+source("../../scripts/h2o-r-test-setup.R")
+
+# Tests parsing with skipped columns
+test.parseSkippedColumnsgzip<- function() {
+  data <- h2o.importFile(path = "http://h2o-public-test-data.s3.amazonaws.com/smalldata/parser/parquet/as_df_err.parquet")
+  
+  df <- as.data.frame(data)
+    
+  expect_equal(nrow(data), nrow(df))
```
sebhrusen, line 12: same as for Py

sebhrusen (commented):
Thanks for the painful investigation @koniecsveta !
If this is all working, then, it's good to go.
Added some syntactic comments otherwise.

h2o-py/tests/testdir_jira/pyunit_pubdev_8239.py
koniecsveta: 126f2de

sebhrusen (approved):
Thank you @koniecsveta !

PR #5590, opened 2021-07-28
PUBDEV-8246: Avoid making a split when bin intervals  are within float precision in DRF

michalkurka:


koniecsveta (commented):
LGTM, thanks! 👍

koniecsveta (approved):

PR #5597, opened 2021-07-29
Speed-up merge test that is comparing frames one-by-one [nocheck]

michalkurka:


tomasfryda (approved):
Thank you @michalkurka!

PR #5599, opened 2021-07-30
[nocheck] java8 tests should use jdk8…

sebhrusen:


michalkurka (approved):
Thanks for noticing @sebhrusen 

I think we need to add a check somewhere

michalkurka (approved):

PR #5604, opened 2021-08-02
[DOCS] PUBDEV-8259 - Weight value of 0 warning

zainhaq-h2o:
This PR adds a warning for when users specify weight column to not have any weight value of 0.0.


h2o-core/src/main/java/water/api/schemas3/ModelParametersSchemaV3.java
```diff
@@ -97,7 +97,8 @@
           " Negative weights are not allowed. Note: Weights are per-row observation weights and do not increase the" +
           " size of the data frame. This is typically the number of times a row is repeated, but non-integer values are" +
           " supported as well. During training, rows with higher weights matter more, due to the larger loss function" +
-          " pre-factor.")
+          " pre-factor. If you set weight == 0, the returned prediction frame at that row is zero. To get an accurate" +
+          " prediction, remove all rows with weight == 0.")
```
wendycwong, line 101: zain:

Please do a ./gradlew clean build -x test at the h2o-3 directory.  You will noticed that there will be new generated python and R files that you will need to do git add and git commit.

wendycwong: zain:  please change this 

If you set weight == 0, the returned prediction frame at that row is zero.

to

If you set weight = 0 for a row, the returned prediction frame at that row is zero and this is incorrect.

h2o-docs/src/product/data-science/algo-params/weights_column.rst
```diff
@@ -9,7 +9,10 @@ Description
...
+For scoring, all computed metrics will take the observation weights into account (for Gains/Lift, AUC, confusion matrices, logloss, etc.), so it’s important to also provide the weights column for validation or test sets if you want to up/down-weight certain observations (ideally consistently between training and testing). Note that if you omit the weights, then all observations will have equal weights (set to 1) for the computation of the metrics. For example, a weight of 2 is identical to duplicating a row.
+
+.. warning::
+    If weight is set equal to zero, the returned prediction frame at that row is zero. To get an accurate prediction, ensure that a value of zero is not specified for the weight column.
```
wendycwong, line 15: Please change:

If weight is set equal to zero, the returned prediction frame at that row is zero.

to 

If weight is set equal to zero, the returned prediction frame at that row is zero and this is incorrect.

wendycwong (changes requested):

wendycwong (approved):

PR #5608, opened 2021-08-03
PUBDEV-8262: Fixed test to not use advanced python command not supported by older Python version

wendycwong:
This PR fixed the problem in https://h2oai.atlassian.net/browse/PUBDEV-8262

The command np.random.default_rng(seed) is not supported by older Python version.  I basically just saved the generated train dataset.  During test, just import it back without running the np.random.default_rng command.


michalkurka (approved):
thank you @wendycwong

sebhrusen (approved):
Thanks @wendycwong.
FYI, if you need to use synthetic data in the future, we have access to sklearn generators in tests: https://scikit-learn.org/0.20/modules/classes.html#samples-generator
Linking the minimal version (0.20) also supported by 2.7 in our test builds.

PR #5612, opened 2021-08-06
HEXDEV-0757: Upgrade versions of AWS and GCS libraries

michalkurka:


michalkurka:
@mn-mikke this might affect SW, can you please check SW side is going to be okay?

mn-mikke (approved):
I've GCS import against spark 2.1 and 3.1. Both works. Spark should be independent on AWS libraries and must added on the classpath manually (see https://medium.com/knoldus/apache-spark-read-data-from-s3-bucket-a53a395c3282) thus S3 should be ok as well.

michalkurka:
@mn-mikke thank you for checking!

PR #5615, opened 2021-08-10
Remove JMH dependency from h2o.jar [nocheck]

michalkurka:
- Don't include JMH dependency in H2O jar
- Remove source of an obsolete module


valenad1 (approved):
👍

PR #5616, opened 2021-08-10
HEXDEV-0772: Use regular H2O serialization in external XGBoost

michalkurka:
Instead of Java object serialization

### DIFF
--- a/h2o-core/src/main/java/water/AutoBuffer.java
+++ b/h2o-core/src/main/java/water/AutoBuffer.java
@@ -278,7 +278,8 @@ public final class AutoBuffer implements AutoCloseable {
       typeMap = getAStr();
       assert typeMap != null;
     } else {
-      assert b == 0;
+      if (b != 0)
+        throw new IllegalStateException("Corrupted communication stream: zero byte expected at the beginning.");
     }
     _typeMap = new short[typeMap.length];
     for (int i = 0; i < _typeMap.length; i++)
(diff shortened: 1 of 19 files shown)


h2o-core/src/main/java/water/AutoBuffer.java
```diff
@@ -264,14 +268,21 @@ public AutoBuffer( InputStream is ) {
...
+    if (typeMap == null) {
+      if (b == 0)
+        return;          // No persistence info
+      int magic = get1U();
+      if (b != 0x1C || magic != 0xED) throw new IllegalArgumentException("Missing magic number 0x1CED at stream start");
+      checkVersion(getStr());
+      typeMap = getAStr();
+      assert typeMap != null;
+    } else {
+      assert b == 0;
```
mn-mikke, line 281: Isn't it better to throw a proper exception identifying that the content is malformed for the current use case?

h2o-core/src/main/java/water/AutoBuffer.java
```diff
@@ -1508,7 +1519,28 @@ public static void writeToChannel(AutoBuffer ab, ByteChannel channel) throws IOE
     channel.write(ab._bb);
     ab.clearForWriting(H2O.MAX_PRIORITY);
   }
-  
+
+  public static byte[] serializeBootstrapFreezable(BootstrapFreezable<?> o) {
+    try (ByteArrayOutputStream bos = new ByteArrayOutputStream();
+         AutoBuffer ab = new AutoBuffer(bos, false)) {
+      ab.put(o);
+      ab.close();
```
mn-mikke, line 1527: Is closing need if resources are in the `try` statement?

mn-mikke (commented):
What is the benefit of using H2O serialization over Java serialization?

michalkurka:
> What is the benefit of using H2O serialization over Java serialization?

@mn-mikke please see https://h2oai.atlassian.net/browse/HEXDEV-772 for details

h2o-core/src/main/java/water/AutoBuffer.java
michalkurka: you are right, this should be improved - thanks for the suggestion

h2o-core/src/main/java/water/AutoBuffer.java
michalkurka: because I need to make sure AutoBuffer is closed before calling toByteArray on the ByteArrayOutputStream - I try to improve the code though

michalkurka: changed, please take another look

mn-mikke (approved):
LGTM

michalkurka:
@mn-mikke thank you for your input and discussion about the PR

PR #5622, opened 2021-08-13
POC: Use log4j2 for logging in standalone

michalkurka:


log4j2.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mmalohlava, line 1: shouldn't be this file in resources? to get it on classpath

michalkurka: yep, just for experimenting, thx

PR #5624, opened 2021-08-16
PUBDEV-4940 - test if this test setup is working

valenad1:


valenad1:
Image is built and this proof is no longer make sence

valenad1:
Image is built and this test proof is no longer make sence

PR #5625, opened 2021-08-16
PUBDEV-8293: Use log4j2 as a logger implementation in H2O standalone

koniecsveta:


michalkurka (approved):
thank you for making this work! @koniecsveta

PR #5627, opened 2021-08-17
PUBDEV-8235: add maximum R2 improvement (maxR) to GLM

wendycwong:
This PR completes the task in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8235

I have completed the Java backend for maxrglm, R/Python APIs and unit tests.  I will add a documentation to help with the code review.


wendycwong:
Here is a description of maxR;

Add Maximum R Square Improvement (MAXR) to GLM
I.	Description from SAS website:
The maximum R2 improvement technique does not settle on a single model. Instead, it tries to find the "best" one-variable model, the "best" two-variable model, and so forth, although it is not guaranteed to find the model with the largest R2 for each size.

The MAXR method 
-	begins by finding the one-variable model producing the highest R2. 
-	Then another variable, the one that yields the greatest increase in R2, is added. 
-	Once the two-variable model is obtained, each of the variables in the model is compared to each variable not in the model. For each comparison, the MAXR method determines if removing one variable and replacing it with the other variable increases R2. 
-	After comparing all possible switches, the MAXR method makes the switch that produces the largest increase in R2. 
-	Comparisons begin again, and the process continues until the MAXR method finds that no switch could increase R2. 
-	Thus, the two-variable model achieved is considered the "best" two-variable model the technique can find. 
-	Another variable is then added to the model, and the comparing-and-switching process is repeated to find the "best" three-variable model, and so forth.

II.	H2O MaxRGLM
The SAS MAXR is not guaranteed to return the model with the best R2 value, the H2O MaxRGLM algo is guaranteed to return the model with the best R2 value.  Instead of doing the switch method, I am just going to do the following for each predictor number x:
-	If there are n predictors and we are looking at using x predictors, first generate all possible combinations of x predictors out of the n predictors;
-	For each element in the combination of x predictors, generate the training frame, build the model and look at the r2 value of the model.
-	Combination with the best r2 value is chosen and the predictor names and r2 values are stored in an array.
Doing it this way will guarantee we find the model with the best R2 values for a particular number of predictors used.
The main disadvantage of this model is the computation time.  Unless it is significantly longer than what SAS does, I don’t intend to change it.

h2o-r/h2o-DESCRIPTION.template
```diff
@@ -35,7 +35,7 @@ Description: R interface for 'H2O', the scalable open source machine learning
     unsupervised machine learning algorithms such as Generalized Linear
     Models (GLM), Gradient Boosting Machines (including XGBoost), Random Forests,
     Deep Neural Networks (Deep Learning), Stacked Ensembles, Naive Bayes,
-    Generalized Additive Models (GAM), ANOVA GLM, Cox Proportional Hazards, K-Means, PCA,
+    Generalized Additive Models (GAM), ANOVA GLM, Cox Proportional Hazards, K-Means, PCA, maxrglm,
```
maurever, line 38:
```suggestion
    Generalized Additive Models (GAM), ANOVA GLM, Cox Proportional Hazards, K-Means, PCA, Maximum R GLM (maxrglm),
```

h2o-test-support/src/main/java/water/TestUtil.java
```diff
@@ -192,6 +192,13 @@ public static void checkArrays(double[] expected, double[] actual, double thresh
         assertEquals(expected[i], actual[i], threshold * Math.min(Math.abs(expected[i]), Math.abs(actual[i])));
     }
   }
+
```
maurever, line 195: You can use `Assert.assertArrayEquals()`, or am I wrong?

h2o-py/tests/testdir_algos/maxrglm/pyunit_PUBDEV_8235_maxrglm_gaussian_validation.py
```diff
@@ -0,0 +1,43 @@
...
+import sys
+sys.path.insert(1, "../../../")
+import h2o
+from tests import pyunit_utils
+from h2o.estimators.maxrglm import H2OMaxRGLMEstimator as maxrglm
+
+# test maxrglm works with validation dataset
+def test_maxrglm_validation():
+    d = h2o.import_file(path=pyunit_utils.locate("smalldata/glm_test/gaussian_20cols_10000Rows.csv"))
+    myY = "C21"
```
maurever, line 12: I prefer the Python syntax in the Python tests. It is a trifle, but can help make code much more readable. So it is up to you to change all variables to snake case, but I strongly recommend it. :)

```suggestion
    my_y = "C21"
```

h2o-algos/src/main/java/hex/maxrglm/MaxRGLMModel.java
```diff
@@ -0,0 +1,177 @@
...
+                _model_summary.set(rIndex, colInd++, bestR2Values[rIndex]);
+                _model_summary.set(rIndex, colInd++, String.join(", ", bestModelPredictors[rIndex]));
+            }
+        }
+    }
+
+    @Override
+    protected Futures remove_impl(Futures fs, boolean cascade) {
+        super.remove_impl(fs, cascade);
+    //    Keyed.remove(_output._transformedColumnKey, fs, true);
```
maurever, line 161: Unused code.

h2o-algos/src/main/java/hex/maxrglm/MaxRGLMModel.java
```diff
@@ -0,0 +1,177 @@
...
+        return fs;
+    }
+
+    @Override
+    protected AutoBuffer writeAll_impl(AutoBuffer ab) {
+  //      if (_output._transformedColumnKey != null)
```
maurever, line 167: Unused code.

maurever (commented):
@wendycwong, thank you for this new algo! I have just a few suggestions. Otherwise, it looks good.

maurever (approved):
Looks good now, thank you @wendycwong .

PR #5628, opened 2021-08-18
PUBDEV-8279: Fixed typo in replacing-values.rst code example docs

rwbrown002:
https://h2oai.atlassian.net/browse/PUBDEV-8279

Fix typo in code example at h2o-docs/src/product/data-munging/replacing-values.rst 

Previous example showed the following typo, using an assignment operator (<-) rather than less than symbol (<) to replace values less than 4.4

`df[df[, "sepal_len"] <- 4.4, "sepal_len"] <- 4.6`

### DIFF
--- a/h2o-docs/src/product/data-munging/replacing-values.rst
+++ b/h2o-docs/src/product/data-munging/replacing-values.rst
@@ -22,7 +22,7 @@ This example shows how to replace numeric values in a frame of data. Note that i
 
         # Replace by row mask. The example below searches for value less than 4.4 in the 
         # sepal_len column and replaces those values with 4.6. 
-        df[df[, "sepal_len"] <- 4.4, "sepal_len"] <- 4.6
+        df[df[, "sepal_len"] < 4.4, "sepal_len"] <- 4.6
 
         # Replace using ifelse. Similar to the previous example, 
         # this replaces values less than 4.6 with 4.6. 


michalkurka:
thank you for your contribution @rwbrown002

michalkurka:
@hannah-tillman the changes look good - however - can you please check the consistency of the example as a whole?

h2o-docs/src/product/data-munging/replacing-values.rst
```diff
@@ -17,12 +17,12 @@ This example shows how to replace numeric values in a frame of data. Note that i
         # so in the example below, the value in the 14th row and 2nd column will be set to 2.0.
```
hannah-tillman, line 17: The R example uses (14th row, 2nd column) while the python example uses (15th row, 3rd column)

h2o-docs/src/product/data-munging/replacing-values.rst
```diff
@@ -17,12 +17,12 @@ This example shows how to replace numeric values in a frame of data. Note that i
...
         df[14, 2] <- 2.0
 
-        # Replace a whole column. The example below multiplies all values in the second column by 3. 
+        # Replace a whole column. The example below multiplies all values in the first column by 3. 
         df[, 1] <- 3 * df[, 1]
 
         # Replace by row mask. The example below searches for value less than 4.4 in the
```
hannah-tillman, line 23: The R example searches for values less than 4.4 while the python example searches for values less than 4.6.

hannah-tillman: Which I think is an inconsistency because in line 28, the R example mentions replacing >4.6 values with 4.6.

hannah-tillman (commented):
I found a couple other inconsistencies between the two examples.

maurever:
Hi @rwbrown002 . Thank you for your contribution. We currently don't have the capacity to assist you in completing this task. We decided that this change is not critical for improving our codebase, so we have closed this PR.

PR #5629, opened 2021-08-18
PUBDEV-8241 if and eif documentation [nocheck]

valenad1:
https://h2oai.atlassian.net/browse/PUBDEV-8241

I added a part about anomaly score in out anomaly detection algorithms

### DIFF
--- a/h2o-docs/src/product/data-science/if.rst
+++ b/h2o-docs/src/product/data-science/if.rst
@@ -138,17 +138,17 @@ A point is marked as an anomaly if the score is greater or equal to (1-``contami
 **Score**: the normalized **mean_length**.
 
 .. math::
-    score(mean\_length) = \frac{(maxPathLength - mean\_length)}{(maxPathLength - minPathLength)}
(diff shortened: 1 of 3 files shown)


h2o-docs/src/product/data-science/if.rst
```diff
@@ -117,6 +117,42 @@ Defining an Isolation Forest Model
...
+
+    - ``1`` = Anomaly
+    - ``0`` = Normal point
+
+Point is marked as an anomaly if the score is great or equal to (1-``contamination``)% quantile of the score.
+
+.. math::
+    predict = score >= Q_{score}(1-contamination)
+
+-  **score**: Normalized mean_length
```
maurever, line 137: Why do you use mean_length and then camelcase in maxPathLength and minPathLength? It looks strange.

```suggestion
-  **score**: Normalized meanLength
```

hannah-tillman (approved):
made a few syntax changes :)

h2o-docs/src/product/data-science/if.rst
valenad1: I have changed it to snake case everywhere according to out conversation.

maurever (approved):
LGTM, thanks @valenad1!

PR #5630, opened 2021-08-18
PUBDEV-8276 fix NPE when weight_column is not in training frame

maurever:
JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8276

### DIFF
--- a/h2o-algos/src/main/java/hex/tree/SharedTree.java
+++ b/h2o-algos/src/main/java/hex/tree/SharedTree.java
@@ -155,7 +155,7 @@ public abstract class SharedTree<
     if( !(0. < _parms._col_sample_rate_change_per_level && _parms._col_sample_rate_change_per_level <= 2) )
       error("_col_sample_rate_change_per_level", "col_sample_rate_change_per_level must be > 0" +
               " and <= 2");
-    if (_train != null) {
+    if (_train != null && _train.find(_parms._weights_column) != -1) {
       double sumWeights = _train.numRows() * (hasWeightCol() ? _train.vec(_parms._weights_column).mean() : 1);
       if (sumWeights < 2*_parms._min_rows ) // Need at least 2*min_rows weighted rows to split even once
         error("_min_rows", "The dataset size is too small to split for min_rows=" + _parms._min_rows
--- a/h2o-algos/src/main/java/hex/tree/SharedTree.java
+++ b/h2o-algos/src/main/java/hex/tree/SharedTree.java
@@ -155,8 +155,8 @@ public abstract class SharedTree<
     if( !(0. < _parms._col_sample_rate_change_per_level && _parms._col_sample_rate_change_per_level <= 2) )
       error("_col_sample_rate_change_per_level", "col_sample_rate_change_per_level must be > 0" +
               " and <= 2");
-    if (_train != null && _train.find(_parms._weights_column) != -1) {
-      double sumWeights = _train.numRows() * (hasWeightCol() ? _train.vec(_parms._weights_column).mean() : 1);
+    if (_train != null) {
+      double sumWeights = _train.numRows() * (hasWeightCol() && _weights != null ? _weights.mean() : 1);
       if (sumWeights < 2*_parms._min_rows ) // Need at least 2*min_rows weighted rows to split even once
         error("_min_rows", "The dataset size is too small to split for min_rows=" + _parms._min_rows
                 + ": must have at least " + 2*_parms._min_rows + " (weighted) rows, but have only " + sumWeights + ".");
--- a/h2o-algos/src/main/java/hex/tree/SharedTree.java
+++ b/h2o-algos/src/main/java/hex/tree/SharedTree.java
@@ -156,7 +156,7 @@ public abstract class SharedTree<
       error("_col_sample_rate_change_per_level", "col_sample_rate_change_per_level must be > 0" +
               " and <= 2");
     if (_train != null) {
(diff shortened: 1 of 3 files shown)


h2o-algos/src/main/java/hex/tree/SharedTree.java
```diff
@@ -155,7 +155,7 @@ protected void ignoreInvalidColumns(int npredictors, boolean expensive) {
     if( !(0. < _parms._col_sample_rate_change_per_level && _parms._col_sample_rate_change_per_level <= 2) )
       error("_col_sample_rate_change_per_level", "col_sample_rate_change_per_level must be > 0" +
               " and <= 2");
-    if (_train != null) {
+    if (_train != null && _train.find(_parms._weights_column) != -1) {
```
sebhrusen, line 158: should we not change the implementation of `hasWeightCol()` instead?
Well, maybe not… because once validation is done, we shouldn't need to do this `find` that can be inefficient on wide datasets.

Maybe the condition should be merged with the call to `hasWeightCol()` below though: if `weight_column=null` now, then `sumWeights=nrows` and we don't validate anymore that `nrows` must be `>= 2 * min_rows`.
wdyt?

maurever: The validation of the `_weights_column` is in `ModelBuilder.init()`,  but after calling `error()` it continues to process the whole `init()` method, and after that, it catches the Exception. 

https://github.com/h2oai/h2o-3/blob/7992e88994590f6775ef8f543d52ad64bc3f6d84/h2o-core/src/main/java/hex/ModelBuilder.java#L1081-L1101

But yes, you are correct, we need to merge the condition below, not to skip `_min_rows` check due to missing `_weights_column`. :+1:

maurever: @sebhrusen, WDYT about this: https://github.com/h2oai/h2o-3/pull/5630/commits/2ddc61178895b10fa37c5e0190909f8c6e22ba44?

The `_weights` should be defined from `ModelBuilder.init()`, if it is not initialized and `hasWeightCol()` is true, something is wrong.

sebhrusen: even better yes! looks like the call to `hasWeightsCol()` is not even necessary anymore.

h2o-algos/src/main/java/hex/tree/SharedTree.java
```diff
@@ -156,7 +156,7 @@ protected void ignoreInvalidColumns(int npredictors, boolean expensive) {
       error("_col_sample_rate_change_per_level", "col_sample_rate_change_per_level must be > 0" +
               " and <= 2");
     if (_train != null) {
-      double sumWeights = _train.numRows() * (hasWeightCol() ? _train.vec(_parms._weights_column).mean() : 1);
+      double sumWeights = _train.numRows() * (hasWeightCol() && _weights != null ? _weights.mean() : 1);
```
michalkurka, line 159: Do we need to check for hasWeightCol? IMHO here just need to check `_weight != null` - the other check is redundant

h2o-algos/src/test/java/hex/tree/SharedTreeTest.java
```diff
@@ -142,4 +143,29 @@ private TestFrameBuilder twoVecFrameBuilder() {
           .withDataForCol(1, ar("A", "B", "A", "B", "A", "B"));
   }
 
+  /**
+   * PUBDEV-8276 - 
+   */
+  @Test(expected = H2OModelBuilderIllegalArgumentException.class)
```
michalkurka, line 149: I think the test has to validate what error is actually thrown

or I can go and accidentally delete a line from the test - eg. line 153 (with column names) and the test will start failing for a different reason => it will not be testing anything anymore

michalkurka (approved):
LGTM - thank you - improvements suggested

h2o-algos/src/main/java/hex/tree/SharedTree.java
maurever: Fixed.

h2o-algos/src/main/java/hex/tree/SharedTree.java
maurever: Fixed.

h2o-algos/src/test/java/hex/tree/SharedTreeTest.java
maurever: Fixed.

sebhrusen (approved):
Thanks @maurever !

h2o-py/tests/testdir_algos/gbm/pyunit_gbm_PUBDEV_8276.py
```diff
@@ -15,7 +15,7 @@ def test_weights_column_not_in_train():
         gbm.train(y=-1, training_frame=df)
         assert False, "Model building should fail."
     except H2OResponseError as e:
-        assert "ERRR on field: _weights_column" in e, "Model building should fail with this in message."
+        assert "ERRR on field: _weights_column" in e.__str__(), "Model building should fail with this in message."
```
sebhrusen, line 18:
```suggestion
        assert "ERRR on field: _weights_column" in str(e), "Model building should fail with this in message."
```

maurever: Thanks, @sebhrusen! I had an eclipse of the brain. :D

michalkurka:
thanks for making the changes @maurever

PR #5636, opened 2021-08-19
PUBDEV-8267: Fixed NPE from multiple runs with knots specified in frame

wendycwong:
This PR resolves the issue in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8267

I removed adding the frame for knots from being tracked in Scope.  This actually delete the frame after scope exit.  

Add junit and runit tests to make sure my fix works and there is no leaked keys.


michalkurka (approved):
thank you @wendycwong !

valenad1 (approved):
Looks good to me, thank you!

PR #5637, opened 2021-08-20
PUBDEV-8205: Mention that explain requires supervised models

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8205

### DIFF
--- a/h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
+++ b/h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
@@ -241,7 +241,7 @@ public class XGBoostSteps extends ModelingSteps {
                     aml().eventLog().info(EventLogEntry.Stage.ModelSelection, "Applying learning rate search on best XGBoost: "+bestXGB._key);
                     XGBoostParameters xgBoostParameters = (XGBoostParameters) bestXGB._parms.clone();
                     XGBoostParameters defaults = new XGBoostParameters();
-                    xgBoostParameters._ntrees = 10000; // reset ntrees (we'll need more for this fine tuning)
+                    xgBoostParameters._ntrees = 10000; // reset ntrees (we'll need more for this fine-tuning)
                     xgBoostParameters._max_runtime_secs = 0; // reset max runtime
                     initTimeConstraints(xgBoostParameters, 0); // ensure we have a max runtime per model in the grid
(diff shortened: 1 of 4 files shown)


h2o-automl/src/main/java/ai/h2o/automl/modeling/XGBoostSteps.java
```diff
@@ -241,7 +241,7 @@ protected ModelSelectionStrategy getSelectionStrategy() {
                     aml().eventLog().info(EventLogEntry.Stage.ModelSelection, "Applying learning rate search on best XGBoost: "+bestXGB._key);
                     XGBoostParameters xgBoostParameters = (XGBoostParameters) bestXGB._parms.clone();
                     XGBoostParameters defaults = new XGBoostParameters();
-                    xgBoostParameters._ntrees = 10000; // reset ntrees (we'll need more for this fine tuning)
+                    xgBoostParameters._ntrees = 10000; // reset ntrees (we'll need more for this fine-tuning)
```
tomasfryda, line 244: Just to get past the `Check Pull Request Metadata` test (it requires changes in java to be run).

h2o-py/h2o/explanation/_explain.py
```diff
@@ -2271,7 +2271,7 @@ def explain(
     or a variable importance plot.  Most of the explanations are visual (plots).
     These plots can also be created by individual utility functions/methods as well.
 
-    :param models: H2OAutoML object, H2OModel, or list of H2O models
+    :param models: H2OAutoML object, supervised H2O model, or list of supervised H2O models
```
michalkurka, line 2274: unrelated - would it work on a imported mojo? "GenericModel"?

ledell (approved):
This looks good!  I just pushed a small edit to the explain.rst file which fixes a bug in the example code.

h2o-py/h2o/explanation/_explain.py
tomasfryda: It does but it wasn't intended to. It doesn't plot the SHAP_* related plots as it assumes only tree-based models are able to calculate SHAP and the other plots seem to work correctly. And AFAIK MOJOs don't support `predict_contributions` (at least my quick attempt to do so failed) so "unofficially" I'd say explain supports MOJOs as much as it can but we don't run tests on MOJOs and it depends on the individual parts, e.g., if there is some limitation for MOJOs on PDP calculation it will be here as well etc.

PR #5640, opened 2021-08-23
PUBDEV-8291: GLM re-enable early stopping with coefficient constraints

BlueTea88:
https://h2oai.atlassian.net/browse/PUBDEV-8291

I would like to re-enable early stopping in GLM lambda search whenever coefficient constraints are specified.

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLM.java
+++ b/h2o-algos/src/main/java/hex/glm/GLM.java
@@ -837,8 +837,6 @@ public class GLM extends ModelBuilder<GLMModel,GLMParameters,GLMOutput> {
       BetaConstraint bc = (_parms._beta_constraints != null)?new BetaConstraint(_parms._beta_constraints.get()):new BetaConstraint();
       if((bc.hasBounds() || bc.hasProximalPenalty()) && _parms._compute_p_values)
         error("_compute_p_values","P-values can not be computed for constrained problems");
-      if(bc.hasBounds())
-        _parms._early_stopping = false; // PUBDEV-4641: early stopping does not work correctly with non-negative option
       _state.setBC(bc);
       if(hasOffsetCol() && _parms._intercept) { // fit intercept
         GLMGradientSolver gslvr = gam.equals(_parms._glmType) ? new GLMGradientSolver(_job,_parms, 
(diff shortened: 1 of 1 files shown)


michalkurka:
@BlueTea88 thanks for the contribution

@wendycwong is this okay to do given that https://h2oai.atlassian.net/browse/PUBDEV-4641 is still open - was it already fixed in another jira?

wendycwong:
Michalk:  The JIRA has not been fixed.  This PR can only be merged after that JIRA is done.

BlueTea88:
@michalkurka - just following up, is there anything I can do to help implement this in some way (temporary or otherwise)?  We use a lot of constrained models at work (insurance pricing) and early stopping would reduce fit time by ~35%.  For us, this is quite significant due to the volume of models we fit.  

When we use constraints in our models, we usually only constrain those predictors that require a monotonic relationship with the target (eg. Sum Insured and Excess when modelling claim cost).  And I think because the predictors naturally have a relationship with the target that is in-line with the constraints, the current early stopping algorithm usually gives us the optimal result.

h2o-algos/src/main/java/hex/glm/GLM.java
```diff
@@ -837,8 +837,6 @@ public void init(boolean expensive) {
       BetaConstraint bc = (_parms._beta_constraints != null)?new BetaConstraint(_parms._beta_constraints.get()):new BetaConstraint();
       if((bc.hasBounds() || bc.hasProximalPenalty()) && _parms._compute_p_values)
         error("_compute_p_values","P-values can not be computed for constrained problems");
-      if(bc.hasBounds())
-        _parms._early_stopping = false; // PUBDEV-4641: early stopping does not work correctly with non-negative option
```
wendycwong, line 841: Instead of deleting this, please change to this:

if (bc.hasBounds() && _parms._early_stopping)
  warn("beta constraint and early_stopping", "if both are enabled may degrade model performance.");

wendycwong (changes requested):

h2o-algos/src/main/java/hex/glm/GLM.java
BlueTea88: Thanks @wendycwong, I have added the change

wendycwong:
Andrew:  

I am going to open an internal JIRA to run your code.  It should be fine.  Will approve after Jenkins tests pass.

wendycwong:
Andrew:  Please change the base from h2oai:master to h2oai:rel-zizler

BlueTea88:
@wendycwong - done

wendycwong (approved):

PR #5648, opened 2021-08-25
PUDBDEV-8282: Added documentation for H2O Maximum R Improvements to GLM [nocheck]

hannah-tillman:
For: [PUBDEV-8282](https://h2oai.atlassian.net/browse/PUBDEV-8282) / tied with https://github.com/h2oai/h2o-3/pull/5627

Let me know if you would like any updates!

### DIFF
--- a/h2o-docs/src/product/data-science/maxrglm.rst
+++ b/h2o-docs/src/product/data-science/maxrglm.rst
@@ -4,7 +4,7 @@ Maximum R Square Improvements (MAXR)
 Introduction
 ~~~~~~~~~~~~
 
-The maximum :math:`R^2` improvement technique does not settle on a single model. Instead, it tries to find the "best" one-variable model, the "best" two-variable model, and so forth. However, it is not guaranteed to find the model with the largest :math:`R^2` for each size.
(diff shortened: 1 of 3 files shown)


h2o-docs/src/product/data-science/maxrglm.rst
```diff
@@ -0,0 +1,231 @@
+Maximum R Square Improvements (MAXR)
+------------------------------------
+
+Introduction
+~~~~~~~~~~~~
+
+The maximum :math:`R^2` improvement technique does not settle on a single model. Instead, it tries to find the "best" one-variable model, the "best" two-variable model, and so forth. However, it is not guaranteed to find the model with the largest :math:`R^2` for each size.
+
```
wendycwong, line 8: Hannah:

Maximum R Square improvement method is first implemented in SAS.  In their implementation, they do not guarantee that the best model is found.  However, for us, we do guarantee that.  Please replace the word 

However,

with

The conventional maximum :math:`R^2` improvement technique does not guarantee to find the model with the largest :math:`R^2` for each predictor subset size.  However, our H2O maximum :math:`R^2` improvement technique does guarantee to find the model with the largest :math:`R^2` for each predictor subset size.

wendycwong (changes requested):

PR #5651, opened 2021-08-27
PUBDEV-8299: Add RMarkdown, Jupyter and HTML output example files to H2O Explain user guide

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8299


ledell (approved):
This is great, thank you @tomasfryda!

PR #5653, opened 2021-08-30
Fix failing test learning curve plot [nocheck]

tomasfryda:
In a new version of `ggplot2`, there seems to be a new consistency check that requires to have the same amount of aesthetics (color in this case) as entries in the legend.

This fix removes the unused colors so that `ggplot2` gets only those colors that are actually used.


koniecsveta (approved):
Thanks!

valenad1 (approved):
Thank you 👍

valenad1:
I just realized that this fix should go into rel-zipf :)

PR #5656, opened 2021-08-30
PUBDEV-8304: Update ignored_columns example [nocheck]

hannah-tillman:
For: [PUBDEV-8304](https://h2oai.atlassian.net/browse/PUBDEV-8304) -> I removed the set predictors from the example


achraf-mer (approved):

achraf-mer:
@hannah-tillman very minor: perhaps we can mention that specifying x with ignored_columns is not (or no longer) accepted.
```
Properties x and ignored_columns cannot be specified simultaneously
```
from error log.

maurever:
Good point @achraf-mer, we should mention that specifying x with ignored_columns is not possible. @hannah-tillman, please, add this to the description section.

maurever (approved):
Thanks, @hannah-tillman!

PR #5660, opened 2021-08-31
Fix Isolation Forest grid search test and API [nocheck]

valenad1:
I have noticed some inconsistency in IF grid search test for R.

The R API treat IF as supervised algorithm and flag is_supervised isn't working properly. This PR provide fix for is_supervised flag and change that IF will be unsupervised by default.

I have created second PR in case that we cannot suddenly change IF to unsupervised for gridsearch. This PR only fix is_supervised flag and add test that IF can be used with is_supervised = FALSE flag.


michalkurka (changes requested):
Please do not introduce a flag is_supervised. Note: API changes like this one should be discussed in #h2o-3-api-updates.

valenad1:
> 
> 
> Please do not introduce a flag is_supervised. Note: API changes like this one should be discussed in #h2o-3-api-updates.

I am not introducing new one, flag is_supervised is already there, I am just using it properly

michalkurka:
@valenad1 I am sorry - my bad - I misunderstood the PR.

I am surprised to see the option there - I don't think we should have it

Would there be a solution that doesn't utilize this flag and works out of the box?

valenad1:
> Would there be a solution that doesn't utilize this flag and works out of the box?

I'll figure something out :)

h2o-r/tests/testdir_algos/isofor/runit_isofor_grid_supervised.R
```diff
@@ -21,7 +21,9 @@ test.grid.resume <- function() {
       x = 1:4,
       y = 5,
       training_frame = iris.hex,
-      hyper_params = hyper_parameters
+      is_supervised = TRUE,
```
michalkurka, line 24: Why do we need to set the flag here?

michalkurka (approved):
I think given that the option `is_supervised` already exists and enables the functionality to work - we should get this fix in.

@valenad1 this looks like this was a code change (in models.R) and not just a test change - in this case we should have a jira

h2o-r/tests/testdir_algos/isofor/runit_isofor_grid_supervised.R
valenad1: I made IF unsupervised by default. This test is using the y parameter and then it should be supervised.

valenad1:
> @valenad1 this looks like this was a code change (in models.R) and not just a test change - in this case we should have a jira

Agree, I will make a JIRA

So I merge this fix and after implement the change that will drop is_supervised parameter. Is that correct? @michalkurka

h2o-r/tests/testdir_algos/isofor/runit_isofor_grid_supervised.R
michalkurka: I am very surprised this works - I thought IF can only have a labeled validation frame, not the training frame. Hmm.

michalkurka:
> So I merge this fix and after implement the change that will drop is_supervised parameter. Is that correct? @michalkurka

If I read this PR correctly, the option didn't work before, or at least not in recent years. If so - it was unused - and thus could be gracefully deprecated (= eg. warning in 3.34 and then removal in 3.36)

h2o-r/tests/testdir_algos/isofor/runit_isofor_grid_supervised.R
valenad1: I didn't check the output of the grid search, since I have only used API I didn't think about that. Good point 👍

valenad1: I am trying to say that I didn't change the test, I just focused that the same code will be used with this flag. I will check what is the actual output of the grid search :)

valenad1:
> If I read this PR correctly, the option didn't work before, or at least not in recent years. 

I agree.

> If so - it was unused - and thus could be gracefully deprecated (= eg. warning in 3.34 and then removal in 3.36)

Ok, I'll try to use a different way (somehow it works for python without this parameter) and make this parameter depracated.

michalkurka (changes requested):
We should not expose is_supervised flag

PR #5661, opened 2021-08-31
Fix Isolation Forest grid search test and API [nocheck] - different solution. 

valenad1:
Different solution for https://github.com/h2oai/h2o-3/pull/5660.


valenad1:
We choose different solution.

PR #5662, opened 2021-08-31
PUBDEV-8319: Implement a java-self-check to allow community users run on latest Java

michalkurka:
H2O has a strict process for certifying the product on compatible Java versions. With recent cadence of Java releases this became an issue for community users. We are usually little behind on the latest supported Java.

While there definitely are Java related issues, the fact is that with state of the Java ecosystem H2O can reliably on any modern and maintained JVM implementation.

This PR replaces a strict check for java version with a dynamic check of JVM capabilities. This means community users (users starting H2O from R/Python) should be able to run H2O even on Java versions that were not certified as long as the self-check is passed.

It also means that users can upgrade Java and keep their older version of H2O.

### DIFF
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -27,6 +27,8 @@ import java.io.IOException;
 import java.lang.management.ManagementFactory;
 import java.lang.management.RuntimeMXBean;
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.*;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
@@ -42,7 +44,7 @@ import java.util.concurrent.atomic.AtomicLong;
 final public class H2O {
   public static final String DEFAULT_JKS_PASS = "h2oh2o";
   public static final int H2O_DEFAULT_PORT = 54321;
-
+  
   //-------------------------------------------------------------------------------------------------------------------
   // Command-line argument parsing and help
   //-------------------------------------------------------------------------------------------------------------------
@@ -376,6 +378,8 @@ final public class H2O {
     /** -allow_clients, -allow_clients=true; Enable clients to connect to this H2O node - disabled by default */
     public boolean allow_clients = false;
 
+    public boolean allow_unsupported_java = false;
+
     /** If this timeout is set to non 0 value, stop the cluster if there hasn't been any rest api request to leader
      * node after the given timeout. Unit is milliseconds.
      */
@@ -595,7 +599,11 @@ final public class H2O {
       }
       else if (s.matches("allow_clients")) {
         trgt.allow_clients = true;
-      } else if (s.matches("rest_api_ping_timeout")) {
+      }
+      else if (s.matches("allow_unsupported_java")) {
+        trgt.allow_unsupported_java = true;
+      } 
+      else if (s.matches("rest_api_ping_timeout")) {
         i = s.incrementAndCheck(i, args);
         trgt.rest_api_ping_timeout = s.parseInt(args[i]);
       }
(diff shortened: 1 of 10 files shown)


h2o-core/src/main/java/water/JavaSelfCheck.java
```diff
@@ -0,0 +1,56 @@
...
+    private static boolean checkUnsafe() {
+        final int N = 1024;
+        final Random r = new Random();
+        final double[] doubleData = new double[N];
+        final byte[] doubleBytes = new byte[N * 8];
+        final long[] longData = new long[N];
+        final byte[] longBytes = new byte[N * 8];
+        for (int i = 0; i < N; i++) {
+            doubleData[i] = r.nextDouble();
+            UnsafeUtils.set8d(doubleBytes, i * 8, doubleData[i]);
```
mmalohlava, line 41: Can this fail on something like UnsupportedOperationException from Unsafe itself?

mmalohlava (commented):
Clever! 

Theoretical question: since we are using custom&old jsr166y, do you see there any problems with future versions of Java
?

h2o-core/src/main/java/water/JavaSelfCheck.java
michalkurka: yes, it most likely can - this is accounted for - if we get any exception -> we are not compatible with the Java

michalkurka:
> Theoretical question: since we are using custom&old jsr166y, do you see there any problems with future versions of Java?

this is a good point, it is possible but probably unlikely - however, we can add a test what would use a fork-join pool

h2o-core/src/main/java/water/H2O.java
```diff
@@ -2197,21 +2205,52 @@ public static void gc() {
...
+        } catch (Throwable t) {
+          error = t;
+        }
+        if (checkPassed) {
+          Log.warn("H2O is running on a version of Java (" + System.getProperty("java.version") + ") that was not certified at the time of the H2O release. " + 
+                  "For production use please use a certified Java version.");
+          return false;
+        }
+      }
+      System.err.printf("Only Java versions %d-%d are supported, system version is %s%n",
```
valenad1, line 2245: I would improve this messages. Can we tell user here that he can pass this check with -allow_unsupported_java?

h2o-core/src/main/java/water/H2O.java
```diff
@@ -2197,21 +2205,52 @@ public static void gc() {
...
+        try {
+          checkPassed = dynamicallyInvokeJavaSelfCheck();
+        } catch (Throwable t) {
+          error = t;
+        }
+        if (checkPassed) {
+          Log.warn("H2O is running on a version of Java (" + System.getProperty("java.version") + ") that was not certified at the time of the H2O release. " + 
+                  "For production use please use a certified Java version.");
+          return false;
+        }
```
valenad1, line 2243: We can be more transparent here and tell the user that we know that he forces his java version, but H2O cannot start.

h2o-core/src/main/java/water/H2O.java
michalkurka: In general, I don't want to recommend that. This is only targeted to community users with straightforward use - h2o.init starts the node.

h2o-core/src/main/java/water/H2O.java
michalkurka: @valenad1 I am not sure what you mean here - can you please provide more details?

valenad1 (approved):
It looks good to me but it is far from my Java skills 👎 

I would only improve the messages to have a different one 

- if user use unsupported java and don't allow it -> tell him he can allow unsupported version
- if user use unsupported java and don't pass your check -> tell him that he must use different java because allow_unsupported_java don't work

michalkurka:
Thanks @valenad1 

> if user use unsupported java and don't allow it -> tell him he can allow unsupported version

Maybe in the future, but for now I want to keep it "under the lid". This is intentional. For straightforward use cases - run h2o.init - let it be seamless and work almost everywhere. If you are doing advanced stuff (possibly multinode) - better use the certified/supported java.

>if user use unsupported java and don't pass your check -> tell him that he must use different java because allow_unsupported_java don't work

Good suggestion - will do - we should tell them to use the latest certified java. Thanks @valenad1

h2o-core/src/main/java/water/H2O.java
valenad1: I meant this

> if user use unsupported java and don't pass your check -> tell him that he must use different java because allow_unsupported_java don't work

michalkurka: Will do, thank you!

PR #5665, opened 2021-09-01
[nocheck] adjust delta 

koniecsveta:
this is because this test is from time to time failing on adjusted assert and factor 3 seems not to be big enough for comparision like this

### DIFF
--- a/h2o-algos/src/test/java/hex/tree/TreeHandlerTest.java
+++ b/h2o-algos/src/test/java/hex/tree/TreeHandlerTest.java
@@ -571,8 +571,8 @@ public class TreeHandlerTest extends TestUtil {
             assertThat("Wow, did we optimize tree rendering with language rules? It is now closer to rendering without rules!",
                     durationEnabled, greaterThan(durationDisabled * slowdownFactor));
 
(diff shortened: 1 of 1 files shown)


valenad1 (approved):
Thank you!

h2o-algos/src/test/java/hex/tree/TreeHandlerTest.java
```diff
@@ -571,8 +571,8 @@ public void testSlowTree() {
             assertThat("Wow, did we optimize tree rendering with language rules? It is now closer to rendering without rules!",
                     durationEnabled, greaterThan(durationDisabled * slowdownFactor));
 
-            // duration disabled and "auto" should be similar in this case, 3x factor of tolerance because we are measuring in ms
-            assertEquals(durationDisabled, durationAuto, (double) 3 * durationDisabled);
+            // duration disabled and "auto" should be similar in this case, 10x factor of tolerance because we are measuring in ms
+            assertEquals(durationDisabled, durationAuto, (double) 10 * durationDisabled);
```
valenad1, line 575: Just a though, it would be more clear to me if we hardcode it to equal +-100ms. What do you think @michalkurka

PR #5666, opened 2021-09-01
PUBDEV-8314: ensure that models without metrics are not added to leaderboard

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8314


michalkurka (approved):
Thank you for the quick fix!

PR #5667, opened 2021-09-01
PUBDEV-7138 - fix model summary for R

valenad1:
seed was not type of long and appear as NA in the R api, Python API is ok with both types..


michalkurka (approved):
good catch!

valenad1:
> Good catch. Are you sure, It is only EIF related? Can you check also other algos, please?

I'll check other algos also

PR #5671, opened 2021-09-02
[nocheck] ModelingStepsExecutorTest sometimes hanging on Jenkins?

sebhrusen:
It looks like one of the test may be hanging sometimes… probably due to the infinite loop.


tomasfryda (approved):
Looks good to me!

PR #5673, opened 2021-09-06
PUBDEV-8284 - test only that _scale_pos_weight gives different output [nocheck]

valenad1:
This test is still failing on multinode. What if we just test if the result is different for both cases (cpu, gpu)?


koniecsveta (approved):
In case _scale_pos_weight not working, the test will still indicate it is not working, so I think this is ok 👍 for check pull request stage you either need to reopen the original ticket while it is being checked or add [nocheck] to the PR name

michalkurka (approved):
in the end that is better :)

PR #5674, opened 2021-09-07
PUBDEV-8014: Add num_of_features argument to explain's varimp

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8014

Add ability to limit the number of features shown in the varimp_heatmap.

### DIFF
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -1744,6 +1744,7 @@ def _calculate_clustering_indices(matrix):
 def varimp_heatmap(
         models,  # type: Union[h2o.automl._base.H2OAutoMLBaseMixin, h2o.H2OFrame, List[h2o.model.ModelBase]]
         top_n=None,  # type: Option[int]
+        num_of_features=20,  # type: Optional[int]
         figsize=(16, 9),  # type: Tuple[float]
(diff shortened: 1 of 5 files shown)


sebhrusen:
`num_of_features`: this `of` is a bit ugly, isn't it? not common at API level…

`n_features` (a la sklearn)? `num_features`?
cc: @ledell

ledell:
@sebhrusen I think `num_of_features` was used already somewhere in H2O and that's why it was re-used, is that right, @tomasfryda?

sebhrusen (commented):
Looks good, I just don't like the name of the new param.

sebhrusen:
>  I think `num_of_features` was used already somewhere in H2O

@ledell that's sad, it's one of those words that I would consider as 100% noise: imo API names should always try to be as concise as possible.
Apparently used for `varimp` and `std_coef` only… probably out of scope to rename this then.

sebhrusen (approved):

ledell (approved):
Works as expected, thanks!

ledell:
```{r}
> aml <- h2o.automl(y = "Species", training_frame = as.h2o(iris), max_models = 5)
> h2o.varimp_heatmap(aml, num_of_features = 2)
```
Screen Shot 2022-03-24 at 11 41 31 PM

PR #5676, opened 2021-09-07
Print Anomaly Score and Normalized Anomaly Score in R instead of empty output [nocheck]

valenad1:
I found that print of IF in R do not report any metrics. I fixed that.


michalkurka (approved):
Nice change!, thank you

PR #5677, opened 2021-09-07
PUBDEV-8294 Add warning for non categorical binomial response

maurever:
JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8294

### DIFF
--- a/h2o-core/src/main/java/hex/ModelBuilder.java
+++ b/h2o-core/src/main/java/hex/ModelBuilder.java
@@ -1443,6 +1443,9 @@ abstract public class ModelBuilder<M extends Model<M,P,O>, P extends Model.Param
         if (_parms._check_constant_response && _response.isConst()) {
           error("_response", "Response cannot be constant.");
         }
+        if(_nclass == 1 && _response.toCategoricalVec().cardinality() == 2){
+          warn("_response", "Response is numeric, so the regression model will be trained. However, the cardinality is equaled to two, so if you want to train a classification model, convert the response column to categorical before training.");
+        }
       }
       if (! _parms._balance_classes)
         hide("_max_after_balance_size", "Balance classes is false, hide max_after_balance_size");


michalkurka (changes requested):
Leaking data + missing a test - please see inline coments

h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -1443,6 +1443,9 @@ public void init(boolean expensive) {
         if (_parms._check_constant_response && _response.isConst()) {
           error("_response", "Response cannot be constant.");
         }
+        if(_nclass == 1 && _response.toCategoricalVec().cardinality() == 2){
```
michalkurka, line 1446: this is leaking the transformed vec - it is generally an expensive operation

I think it would be better to limit this warning only to binary response (0/1) and use isBinary

maurever: Sorry, I forgot to set this PR as a draft. 

Thank you for your suggestions. I miss the isBinary function.. Also, I will implement the -1/1 check!

maurever: @michalk for -1/+1 case there is no easy to get a number of zeros from the rollups stats.. You can get a number of zeros from a sparse chunk only, or am I missing something?

michalkurka: @maurever 

_nzCnt,       //count(X!=0) for non-NA values

you can check this value

vec.length() - _nzCnt - _naCnt - this will give you the number of zeros

maurever: Done here: https://github.com/h2oai/h2o-3/pull/5677/commits/ed57a95ecbe17f0f46d74608a5c5b2c1dc1dc582, thank you @michalkurka for your ideas!

h2o-core/src/main/java/water/fvec/Vec.java
```diff
@@ -320,6 +320,16 @@ public boolean isBinary(){
     return rs._isInt && rs._mins[0] >= 0 && rs._maxs[0] <= 1;
   }
 
+  /**
+   * Check binary vector case -1/1
+   * @return true if the vector is consist of -1/1 only
+   */
+  public boolean isBinaryOnes(){
```
michalkurka, line 327: it looks like a constant -1 or +1 columns would be also considered binary - is that intended?

maurever: Good point. If it is not intended, we should also change `isBinary()` method...

michalkurka: I wouldn't change isBinary method without evaluating the impact, there is also no test for neither of the modified methods

for core classes there should always be a unit test

michalkurka (approved):

PR #5685, opened 2021-09-10
Make k8s clustering test work for the default clustering method [nocheck]

michalkurka:


tomasfryda (approved):
LGTM. Thank you @michalkurka!

PR #5694, opened 2021-09-10
PUBDEV-8322: Create a compatibility k8s module that would allow older H2O version run on k8s

michalkurka:
- Make k8s clustering test work for the default clustering method
- Refactor - EmbeddedH2OConfig should decide wheter non-leader node access is ok
- Move KubernetesRequestFilter to h2o-core - could be used outside of k8s
- Fix-up LeaderNodeRequestFilter to be generic (not just for k8s)
- K8S module refactor - separate plumbing code from k8s logic
- Refactor - move k8s h2o integration code into h2o-k8s-int module


h2o-k8s/tests/clustering/docker-init-script.sh
```diff
@@ -6,6 +6,12 @@ if [[ $string == *"@"* ]]; then
   echo "H2O base path contains at sign. Unable to create K3S cluster."
   exit 1
 fi
+
+# Force in-house version of the image in tests
+sed -i.bu 's,everpeace,harbor.h2o.ai/opsh2oai,' $H2O_BASE/h2o-helm/templates/tests/test-connection.yaml
+rm $H2O_BASE/h2o-helm/templates/tests/test-connection.yaml.bu
```
tomasfryda, line 12: I'm not sure if it's wise to modify part of source - it can get annoying when running the test locally.
I would suggest to undo the change afterwards, e.g.
```suggestion
sed -i.bu 's,everpeace,harbor.h2o.ai/opsh2oai,' $H2O_BASE/h2o-helm/templates/tests/test-connection.yaml
trap 'rm $H2O_BASE/h2o-helm/templates/tests/test-connection.yaml; mv $H2O_BASE/h2o-helm/templates/tests/test-connection.yaml{.bu,}' EXIT
```

Or if it is intended to be ran only in a docker container (linux with GNU coreutils) it could be simplified to:
```suggestion
sed -i 's,everpeace,harbor.h2o.ai/opsh2oai,' $H2O_BASE/h2o-helm/templates/tests/test-connection.yaml
```
(Which has the added "benefit" of failing on macOS with BSD coreutils.)

michalkurka: good point, running this locally really could be an issue - I will try to implement your suggestion

h2o-k8s-comp/build.gradle
```diff
@@ -0,0 +1,40 @@
+apply plugin: 'java'
+apply plugin: 'com.github.johnrengelman.shadow'
+
+description = "H2O Kuberentes Compatibility Library - enables older H2O versions to run on K8S"
```
michalkurka, line 4: typo

tomasfryda (approved):

h2o-k8s-int/src/main/java/water/k8s/KubernetesEmbeddedConfigProvider.java
```diff
@@ -0,0 +1,41 @@
+package water.k8s;
+
+import water.init.AbstractEmbeddedH2OConfig;
+import water.init.EmbeddedConfigProvider;
+import java.util.Collection;
+
+/**
+ * A configuration provider for H2O running in Kubernetes cluster. It is able to detected H2O is being ran in K8S
```
mn-mikke, line 8: typo: `ran` -> `run` - past participle?

mn-mikke (approved):
LGTM!

michalkurka:
@mn-mikke @tomasfryda thank you for your review and your time

PR #5758, opened 2021-09-12
Add ability to delete directory in FileUtils::delete()

navdeep-G:
* Add ability to delete directory in FileUtils::delete()
* I had to make this change for a private h2o repo


h2o-core/src/test/java/water/util/FileUtilsTest.java
```diff
@@ -25,6 +28,32 @@
     @Rule
     public EnvironmentVariables environmentVariables = new EnvironmentVariables();
 
+    @Test
+    public void testFileDelete() throws Exception {
+        // Absolute path
+        String absPath = new File(System.getProperty("user.dir")).getAbsolutePath();
```
michalkurka, line 34: This directory should be created in a temporary JUnit space. Using a user's home directory is a bad idea - if there is a bug in the code user can lose data.

See eg. https://github.com/h2oai/h2o-3/blob/879e58d3e5f98064b738189febecd845bf4f87c4/h2o-mapreduce-generic/src/test/java/water/hadoop/clouding/fs/CloudingEventTest.java#L33 and few lines bellow

michalkurka (approved):
thank you!

PR #5759, opened 2021-09-13
PUBDEV-8307: Updates to AutoML user guide FAQ

ledell:
https://h2oai.atlassian.net/browse/PUBDEV-8307

Also made updates to the leaderboard info -- now we are using an image of the leaderboard rather than a table (easier to read).

Screen Shot 2021-09-13 at 10 54 03 PM


tomasfryda (approved):
Looks good to me. Thank you @ledell!

PR #5760, opened 2021-09-13
PUBDEV-8326: Fix matplotlib's 3.4 issue with one value in where argument in fill_between call

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8326

There's an issue with new `matplotlib` which I think was caused by a bug on our side. 

The new version of matplotlib does not support having just one value in `where` argument. We had there `lower < upper` both of which are lists but python's behavior is to produce one boolean (`[1, 1, 1] < [1, 1, 2]` => `True`).

Semantically I think the intention was to provide a boolean array which corresponds to using numpy arrays instead of lists and this also fixes the problem with the new version of `matplotlib`.


maurever (approved):
I can't see any problems with this change. :+1:  Thanks, @tomasfryda, for a quick fix.

michalkurka (approved):

PR #39, opened 2016-04-21
Rel turchin

karimulla78:
This change is [Reviewable](https://reviewable.io/reviews/h2oai/h2o-3/39)

PR #5761, opened 2021-09-13
Downgrade to release image #39 to avoid issue with Sphinx [nocheck]

michalkurka:
Image #40 breaks documentation - see internal discussion

https://h2oai.slack.com/archives/C04KNHH2H/p1631514069021100


valenad1 (approved):
👍

PR #5763, opened 2021-09-14
PUBDEV-8329: don't handle zero weights

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8329

When rfit mojo gets trained, features contains response column.

This is caused by RuleFitMojoModel#nfeatures() what was needed for handling zero weights block in RuleFitMojoModel#score0(). But since https://h2oai.atlassian.net/browse/PUBDEV-8249 got clarified as not a bug, this is not needed and it also causes this bug.

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/algos/rulefit/RuleFitMojoModel.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/algos/rulefit/RuleFitMojoModel.java
@@ -46,15 +46,6 @@ public class RuleFitMojoModel extends MojoModel {
     
     _linearModel.score0(linearModelInput, preds);
     
-    // if current weight is zero, zero the prediction
-    if (_weightsColumn != null) {
-      int weightsId = Arrays.asList(_linearNames).indexOf("linear." + _weightsColumn);
-      double currWeight = test[weightsId];
-      if (currWeight == 0.0) {
-        for (int i = 0; i < preds.length; i++)
-          preds[i] *= currWeight;
-      }
-    }
     
     return preds;
   }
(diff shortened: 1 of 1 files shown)


mn-mikke:
@koniecsveta I need your help to understand the cause of the problem. I assume that `_names` contains all column names of the training dataset including offset column, weight column and response_column.  Since feature columns are listed in `_names` as first, the logic of the `features()` method (see the below code snippet), will return just feature columns if you fixed calculation of `_nfeatures`.

```
@Override public String[] features() {
    return Arrays.copyOf(_names, nfeatures());
  }
 ```
 
 Is it what's going on or do I miss anything?

koniecsveta:
> @koniecsveta I need your help to understand the cause of the problem. I assume that `_names` contains all column names of the training dataset including offset column, weight column and response_column. Since feature columns are listed in `_names` as first, the logic of the `features()` method (see the below code snippet), will return just feature columns if you fixed calculation of `_nfeatures`.
> 
> ```
> @Override public String[] features() {
>     return Arrays.copyOf(_names, nfeatures());
>   }
> ```
> 
> Is it what's going on or do I miss anything?

yes, exactly. I changed how nfeatures() gets calculated so that I have weight column available in mojo scoring, because I thought that this case  https://h2oai.atlassian.net/browse/PUBDEV-8249 is a bug which is being propagated into rulefit. But as it got clarified, it is not a bug and there is no need to do anything with weights during scoring. So the code which I am removing shouldn't be present.

mn-mikke (approved):
@koniecsveta Thanks for your clarification and fixing the problem!

LGTM

michalkurka (approved):

PR #5766, opened 2021-09-15
Revert Jetty upgrade - caused issues SW and in Hadoop smoke tests [nocheck]

michalkurka:
This reverts commit 6aeb46432872d290158e420c0702451ec4a7fe6e.


michalkurka:
FIY - @achraf-mer we found out that upgrading Jetty to lastest causes issues both for SW and in h2o-3 Hadoop tests.

I will revert this for now and  find the latest jetty that still works

mn-mikke (approved):
LGTM, thank you!

PR #5768, opened 2021-09-18
PUBDEV-8335 - upgrade to Gradle 7 to be able to support java 16+

valenad1:
> Follow up after https://github.com/h2oai/h2o-3/pull/5681 and it is precondition for Java 16 and 17 support (https://github.com/h2oai/h2o-3/pull/5769)

Went to the master first and might go to the release after.

### DIFF
--- a/h2o-hadoop-2/assemblyjar.gradle
+++ b/h2o-hadoop-2/assemblyjar.gradle
@@ -1,4 +1,4 @@
-apply plugin: 'java'
+apply plugin: 'java-library'
 apply plugin: 'com.github.johnrengelman.shadow'
 
 description = 'H2O HDFS client shadowjar for Hadoop ' + hadoopVersion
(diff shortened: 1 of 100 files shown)


valenad1:
Hi @michalkurka @mn-mikke, can you please validate my gradle 7 upgrade?

I would recommend to go through commits one by one. I have made an atomic commit for a every single change that caused problems.

@mn-mikke do you have any tips how to identify dependencies that are not supposed to be define with `api` but `implementation` ? I have used `api` because it works :)

@mn-mikke is there anything else that should be check? @michalkurka mentions something about maven and poms :)

h2o-hadoop-2/assemblyjar.gradle
```diff
@@ -15,60 +15,60 @@ if (hasCustomHdfsDep) {
...
+  api project(":h2o-avro-parser")
   // Include GCS persist layer
-  compile project(":h2o-persist-gcs")
+  api project(":h2o-persist-gcs")
   // Include S3 persist layer
-  compile project(":h2o-persist-s3")
+  api project(":h2o-persist-s3")
   // Include HDFS persist layer
-  compile(project(':h2o-persist-hdfs')) {
+  api(project(':h2o-persist-hdfs')) {
```
mn-mikke, line 46: IMHO, all dependecies 
`
compile(project(...)) {
    transitive = false
  }
  `
  could be changed  to:
  `
implementation(project(...)) {
    transitive = false
  }
  `

mn-mikke (approved):
> I have used api because it works.
We did the same when doing the conversion in Sparkling Water. However, implementation is considered as a best practice.

In nutshell, `api` also puts transitive dependencies to compile classpath and `implementation` don't so if H2O code utilize  for example Gson or something from apache commons, the dependencies must be referenced with `implementation` explicitly. The runtime classpath should be the same in both cases. In other words, Shadow plugin should include transitive dependencies to the  fat jar in both cases.

IMHO, it's ok to start with `api` when converting from `compile`. At least, the change will be minimal.

PR #5770, opened 2021-09-18
Correct MAXR GLM Python example typos

neemah2o:
Correct MAXR GLM Python example typo
`estimators` should be `h2o.estimators`
`maxrglm_model.get_best_r2_values()` should be `maxrglm_model.get_best_R2_values()`


neemah2o:
https://h2oai.atlassian.net/browse/PUBDEV-8336

PR #5775, opened 2021-09-21
[PUBDEV-8280] Fix ModelJsonReader.findInJson for Cases When Path Does Not Exist

mn-mikke:
Before this PR, `ModelJsonReader.readTable` throws a null pointer exception if `findInJson` returns `null` (cannot do `null.isJsonNull`). Similarly, `ModelJsonReader.elementExists` returns `true` in this case.

### DIFF
--- a/h2o-genmodel/src/main/java/hex/genmodel/attributes/ModelJsonReader.java
+++ b/h2o-genmodel/src/main/java/hex/genmodel/attributes/ModelJsonReader.java
@@ -483,7 +483,11 @@ public class ModelJsonReader {
             } else break;
         }
 
-        return result;
+        if (result == null) {
+            return JsonNull.INSTANCE;
+        } else {
+            return result;   
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


valenad1 (approved):
It looks good to me, thank you!

h2o-genmodel/src/main/java/hex/genmodel/attributes/ModelJsonReader.java
```diff
@@ -483,7 +484,11 @@ protected static JsonElement findInJson(final JsonElement jsonElement, final Str
             } else break;
         }
 
-        return result;
```
michalkurka, line 486: looks like https://github.com/h2oai/h2o-3/pull/5775/files#diff-4a16d86a307592634a57c770ec83647ad7867d4d373ea304df64f5defc3074cfR475 became redundant

michalkurka (approved):
Thanks for the fix @mn-mikke

h2o-genmodel/src/main/java/hex/genmodel/attributes/ModelJsonReader.java
mn-mikke: yes, you are right. Thank you!

wendycwong (approved):

maurever (approved):

PR #5778, opened 2021-09-22
Remove h2o-k8s-comp from published projects

michalkurka:
This module doesn't produce a jar artefact, it is not meant to be published (only added there by accident)


mn-mikke (approved):
LGTM, Thanks!

PR #5781, opened 2021-09-23
PUBDEV-8343: expose AutoML get_leaderboard as a method in Py

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8343

Had to reorganize the Python AutoML code layout to avoid dependency loops: Py is very sensitive to that (rightly so) with imports done on the top of the file.

`H2OAutoML` class definition moved from `h2o.automl.autoh2o` to `h2o.automl._estimator`, but is still available as both `h2o.automl.H2OAutoML` and `h2o.automl.autoh2o.H2OAutoML` (for legacy reason as old doc was unfortunately using the latter).

Private static methods handling communication with server were moved as private functions to `h2o.automl._base` module.


ledell:
@sebhrusen Will the user still be encouraged to use:

```
```

The current ^^ version seems easy enough to remember so it would be great to just keep that as the standard we use in the docs, examples, etc.

sebhrusen:
@erin
```python
```
is still the recommended way to import automl

ledell (approved):
This looks good, thanks!

tomasfryda (approved):
Looks good to me. Thank you @sebhrusen!

PR #5782, opened 2021-09-23
PUBDEV-8342: show the stacktrace on Py client in case of ServerError

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8342

always print ErrorV3 shortened stacktrace if available.

Rationale is to make it easier for users to report informative issues and for devs to investigate them.

Before:
```
H2OServerError: HTTP 500 Server Error:
Server error java.lang.NullPointerException:
  Error: Caught exception: java.lang.NullPointerException
  Request: None
```
Now:
```
H2OServerError: HTTP 500 Server Error:
Server error java.lang.NullPointerException:
  Error: Caught exception: java.lang.NullPointerException
  Request: None
  Stacktrace: java.lang.NullPointerException
    water.nbhm.NonBlockingHashMap.putIfMatch(NonBlockingHashMap.java:369)
    water.nbhm.NonBlockingHashMap.put(NonBlockingHashMap.java:320)
    ai.h2o.automl.AutoMLSession.getModelingSteps(AutoMLSession.java:76)
    ai.h2o.automl.ModelingStepsRegistry.getOrderedSteps(ModelingStepsRegistry.java:54)
    ai.h2o.automl.AutoML.getExecutionPlan(AutoML.java:330)
    ai.h2o.automl.AutoML.planWork(AutoML.java:359)
    ai.h2o.automl.AutoML.submit(AutoML.java:395)
    ai.h2o.automl.AutoML.startAutoML(AutoML.java:80)
    water.automl.api.AutoMLBuilderHandler.build(AutoMLBuilderHandler.java:15)
```

### DIFF
--- a/h2o-py/h2o/backend/connection.py
+++ b/h2o-py/h2o/backend/connection.py
@@ -815,6 +815,7 @@ class H2OConnection(h2o_meta()):
 
         # Client errors (400 = "Bad Request", 404 = "Not Found", 412 = "Precondition Failed")
         if status_code in {400, 404, 412} and isinstance(data, H2OErrorV3):
+            data.show_stacktrace = False
             raise H2OResponseError(data)
 
         # Server errors (notably 500 = "Server Error")
--- a/h2o-py/h2o/schemas/error.py
+++ b/h2o-py/h2o/schemas/error.py
@@ -21,6 +21,7 @@ class H2OErrorV3(H2OSchema):
         super(H2OErrorV3, self).__init__()
         self.endpoint = None
         self.payload = None
+        self.show_stacktrace = True
 
     def __setitem__(self, key, value):
         if key in self._schema_attrs_.keys():
@@ -37,15 +38,16 @@ class H2OErrorV3(H2OSchema):
             if self.payload[1]: res += "    json: %r\n" % self.payload[1]
             if self.payload[2]: res += "    file: %r\n" % self.payload[2]
             if self.payload[3]: res += "    params: %r\n" % self.payload[3]
-        if self.stacktrace:
+        if self.show_stacktrace and self.stacktrace:
             res += "  Stacktrace: %s\n" % self._format_stacktrace(indent=6)
         return res
     
     def _format_stacktrace(self, indent=4, top=10):
-        if self.stacktrace:
-            return "\n".join(("%s%s" % ((indent*" " if i > 0 else ""), l.strip()) 
-                              for i, l in enumerate(self.stacktrace) 
-                              if i < top))
+        if not self.stacktrace:
+            return ""
+        return "\n".join(("%s%s" % ((indent*" " if i > 0 else ""), l.strip()) 
+                          for i, l in enumerate(self.stacktrace) 
+                          if i < top))
 
 
 class H2OModelBuilderErrorV3(H2OErrorV3):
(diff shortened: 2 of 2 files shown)


koniecsveta (approved):

tomasfryda (approved):

h2o-py/h2o/schemas/error.py
```diff
@@ -21,6 +21,7 @@ def __init__(self):
         super(H2OErrorV3, self).__init__()
         self.endpoint = None
         self.payload = None
+        self.show_stacktrace = True
```
michalkurka, line 24: Do we know what other subclasses are affected? Does this lead to too verbose output in some cases?

h2o-py/h2o/schemas/error.py
```diff
@@ -37,7 +38,16 @@ def __str__(self):
             if self.payload[1]: res += "    json: %r\n" % self.payload[1]
             if self.payload[2]: res += "    file: %r\n" % self.payload[2]
             if self.payload[3]: res += "    params: %r\n" % self.payload[3]
+        if self.show_stacktrace and self.stacktrace:
+            res += "  Stacktrace: %s\n" % self._format_stacktrace(indent=6)
         return res
+    
+    def _format_stacktrace(self, indent=4, top=10):
```
michalkurka, line 45: It seems that H2OModelBuilderErrorV3 does a similar thing - should this code be unified?

h2o-py/h2o/schemas/error.py
sebhrusen: I disable the stacktrace for client errors, as the message should be enough for those.
I consider that all other errors coming from the backend are unexpected and deserve to display the stacktrace.

h2o-py/h2o/schemas/error.py
sebhrusen: Right, it was my original intent, but looks like I forgot then… will use this function in `H2OModelBuilderErrorV3`

sebhrusen: for `H2OModelBuilderErrorV3` I'll render the entire stacktrace though, mainly useful if we have some cases of wrapped exceptions.

h2o-py/h2o/backend/connection.py
```diff
@@ -815,6 +815,7 @@ def _process_response(response, save_to):
 
         # Client errors (400 = "Bad Request", 404 = "Not Found", 412 = "Precondition Failed")
         if status_code in {400, 404, 412} and isinstance(data, H2OErrorV3):
```
sebhrusen, line 817: btw, could be 
```if 400 <= status_code < 500 and …```
don't see the reason why this is limited to specific client errors

michalkurka: I used the stacktrace of 401 once to determine what Jetty version is the user running :)

That is a weird case of course - but generally the stackrace for Unauthorized is helpful in validating that the message is actually generated by H2O (jetty) and not something along the way.

IMHO other 4xx than the listed ones are not common so we don't have to worry about how well they are printed - I would keep as is - definitely for 401 it can save time.

h2o-py/h2o/schemas/error.py
michalkurka: Thanks for explaining

michalkurka (approved):
Thanks for this improvement @sebhrusen 

Unrelated to this - it would also be helpful to include some basic metadata about the h2o cluster in error printouts, eg.: h2o version, cluster stats (number of nodes), OS - this one is for another time

sebhrusen:
@michalkurka created https://h2oai.atlassian.net/browse/PUBDEV-8351 for stats reporting on exception: feel free to complete

PR #5785, opened 2021-09-24
PUBDEV-8349: Fix infinite loop in hex.grid.HyperSpaceWalker.RandomDiscreteValueWalker

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8349

### DIFF
--- a/h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
+++ b/h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
@@ -527,6 +527,8 @@ public interface HyperSpaceWalker<MP extends Model.Parameters, C extends HyperSp
         private Map<String, Object[]> _currentHyperParams = _hyperParams;
         private String[] _currentHyperParamNames = _hyperParamNames;
 
+        private boolean _exhausted = false;
+
         // TODO: override into a common subclass:
         @Override
         public MP nextModelParameters() {
@@ -569,7 +571,8 @@ public interface HyperSpaceWalker<MP extends Model.Parameters, C extends HyperSp
           //
           // _currentPermutationNum is 1-based
           return (_visitedPermutationHashes.size() < _maxHyperSpaceSize &&
-                  (search_criteria().max_models() == 0 || _currentPermutationNum < search_criteria().max_models())
+                  (search_criteria().max_models() == 0 || _currentPermutationNum < search_criteria().max_models()) &&
+                  !_exhausted
           );
         }
 
@@ -589,8 +592,8 @@ public interface HyperSpaceWalker<MP extends Model.Parameters, C extends HyperSp
           
           int[] hyperparamIndices = new int[_currentHyperParamNames.length];
           
-          do {
-            if(_hyperParamSubspaces.length != 0) {
+          for (int j=0; j < Math.max(1e4, _maxHyperSpaceSize); j++) {
+            if (_hyperParamSubspaces.length != 0) {
               _currentSubspace = _random.nextInt(_hyperParamSubspaces.length);
               _currentHyperParams = mergeHashMaps(_hyperParams, _hyperParamSubspaces[_currentSubspace]);
               _currentHyperParamNames = _currentHyperParams.keySet().toArray(new String[0]);
@@ -600,9 +603,12 @@ public interface HyperSpaceWalker<MP extends Model.Parameters, C extends HyperSp
               hyperparamIndices[i] = _random.nextInt(_currentHyperParams.get(_currentHyperParamNames[i]).length);
             }
             // check for aliases and loop if we've visited this combo before
-          } while (_visitedPermutationHashes.contains(integerHash(_currentHyperParams, _currentHyperParamNames, hyperparamIndices, _currentSubspace)));
-          
-          return hyperparamIndices;
+           if (!_visitedPermutationHashes.contains(integerHash(_currentHyperParams, _currentHyperParamNames, hyperparamIndices, _currentSubspace)))
+            return hyperparamIndices;
+          }
+
+          _exhausted = true;
+          return null;
         } // nextModel
 
       }; // anonymous HyperSpaceIterator class
--- a/h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
+++ b/h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
@@ -106,6 +106,32 @@ public interface HyperSpaceWalker<MP extends Model.Parameters, C extends HyperSp
    */
   MP getParams();
 
+  /**
+   * Return estimated grid work.
+   * Can return Long.MAX_VALUE if no estimate is available.
+   * @param maxModels
+   * @return estimate of grid work
+   */
+  default long getGridWork(long maxModels) {
+    HyperSpaceWalker.HyperSpaceIterator<MP> it = iterator();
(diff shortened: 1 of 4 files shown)


h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -589,8 +592,8 @@ public void onModelFailure(Model failedModel, Consumer<Object[]> withFailedModel
           
           int[] hyperparamIndices = new int[_currentHyperParamNames.length];
           
-          do {
-            if(_hyperParamSubspaces.length != 0) {
+          for (int j=0; j < Math.max(1e4, _maxHyperSpaceSize); j++) {
```
tomasfryda, line 595: `max(1e4, _maxHyperSpaceSize)` is the maximum number of unsuccessful attempts to find a new parameter configuration. This can potentially cause it to finish prematurely but to exhaustively go through all possible parameter configuration we have better way (CartesianWalker).

`1e4` was chosen simply because it is high enough to make it unlikely we would not exhaust small parameter space while being low enough to finish within a second. Test I added finds a hash collision after ~874 512 iterations and takes 7s to finish.

sebhrusen: magic numbers should always be stored in a constant variable (possibly documented if not clear), otherwise no one will remember what this means in a few months/

sebhrusen: Also, I don't understand why you need this magic number now that you have the `exhausted` variable, why not:
```java
for (int j=0; j < _maxHyperSpaceSize && !exhausted; j++) {
```

sebhrusen: Also, as discussed, I'm convinced that the following logic to estimate total work in `GridSearch`:
```java
      while (it.hasNext()) {
        try {
          Model.Parameters parms = it.nextModelParameters();
          gridWork += (parms._nfolds > 0 ? (parms._nfolds+1/*main model*/) : 1) *parms.progressUnits();
        } catch(Throwable ex) {
          //swallow invalid combinations
        }
      }
```
should be refactored to have the iteration logic in the `HyperspaceIterator` impl: this loop is meaningless in RGS and causes the original issue.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -600,9 +603,12 @@ public void onModelFailure(Model failedModel, Consumer<Object[]> withFailedModel
...
             }
             // check for aliases and loop if we've visited this combo before
-          } while (_visitedPermutationHashes.contains(integerHash(_currentHyperParams, _currentHyperParamNames, hyperparamIndices, _currentSubspace)));
-          
-          return hyperparamIndices;
+           if (!_visitedPermutationHashes.contains(integerHash(_currentHyperParams, _currentHyperParamNames, hyperparamIndices, _currentSubspace)))
+            return hyperparamIndices;
+          }
+
+          _exhausted = true;
```
sebhrusen, line 610: so basically, you mark it as exhausted on the first collision, which, imo, is a very pessismistic way to mark the entire hyperspace as exhausted.

I'd rather skip the indices if there is a collision, count successive collisions and finally mark the hyperspace as exhausted if the successive collisions are above a certain threshold (e.g. 1/1000 of hyperspace size?).

h2o-test-support/src/main/java/water/test/dummy/DummyModelParameters.java
```diff
@@ -20,4 +20,16 @@ public DummyModelParameters(String msg, Key trgt) {
...
+  public int _max_depth;
+  public double _min_rows;
+  public double _sample_rate;
+  public double _col_sample_rate;
+  public double _col_sample_rate_per_tree;
+  public String _booster;
+  public float _reg_lambda;
+  public float _reg_alpha;
+  public float _scale_pos_weight;
+  public float _max_delta_step;
```
sebhrusen, line 34: please create your own (extend this one) if you need to make this too specific: it was designed to be simple and generic.
If we start adding params for every single algo, it will be a mess quickly.

sebhrusen (changes requested):

h2o-test-support/src/main/java/water/test/dummy/DummyModelParameters.java
tomasfryda: Good point, thanks!

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
tomasfryda: I think you misunderstood the code. This isn't `first collision => exhausted == True`. I don't check for collisions as it would be too expensive (I'd have to compare the parameter configurations not just their hashes to find out if it's collision or not).

The implementation is basically: if you can't find a new configuration within `maxHyperSpaceSize` steps say the space is `exhausted`. There can be multiple collisions and yet the space wouldn't be considered `exhausted` if you can find another new configuration within the `maxHyperSpaceSize` steps.

sebhrusen: oh right, my bad! As this loop is executed each time we need a new set of params, we might encounter previously used params, which is not a collision.

This looks good then, thanks.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
sebhrusen: > Also, I don't understand why you need this magic number now that you have the `exhausted` variable, why not:
> 
> ```java
> for (int j=0; j < _maxHyperSpaceSize && !exhausted; j++) {
> ```

ok, this comment was also related with my misunderstanding below.
And `1e4` is actually an search upper bound for small hyperspaces.

Still, please extract the magic number.

tomasfryda: Thanks! I extracted the magic number and added some comments. I hope they'll make it more clear.

I also refactored this logic: 
```java
while (it.hasNext()) {
        try {
          Model.Parameters parms = it.nextModelParameters();
          gridWork += (parms._nfolds > 0 ? (parms._nfolds+1/*main model*/) : 1) *parms.progressUnits();
        } catch(Throwable ex) {
          //swallow invalid combinations
        }
      }
```
but not into an `Iterator` but into the `Walker` as iterator is stateful and this logic would modify the state of the iterator.

h2o-core/src/main/java/hex/grid/HyperSpaceWalker.java
```diff
@@ -106,6 +106,32 @@
    */
   MP getParams();
 
+  /**
+   * Return estimated grid work.
+   * Can return Long.MAX_VALUE if no estimate is available.
+   * @param maxModels
+   * @return estimate of grid work
+   */
+  default long getGridWork(long maxModels) {
```
sebhrusen, line 115: nitpick: let's name it for what it is, `estimateGridWork` (?) and keep `get__` for property getters.

sebhrusen (commented):
Thanks for the changes!

sebhrusen (approved):
This signs the end of infinity, maybe Giordano Bruno deserved to be burned after all…

Thanks @tomasfryda

PR #5790, opened 2021-09-27
PUBDEV-8354-remove balance_classes from GLM.  

wendycwong:
This PR completes task in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8354

Removed balance_classes, class_sampling_factors and max_after_balance_size for GLM, GAM, MaxRGLM and ANOVAGLM


michalkurka:
@wendycwong I made a mistake - the error is not actually thrown from the backend - this changes things... even though balance_classes doesn't do anything we cannot simply remove - we need to produce a warning instead

hide(...) only works in Flow but doesn't propagate a warning to the backend - which means there might be scripts out there that have balance_classes enabled and we don't want to break those

wendycwong:
Michalk will take over this JIRA as he has more insight on what to do.

PR #5793, opened 2021-09-29
updated sphinx versions

NidhiMehta:
1) Will need to create a new docs docker image (not sure if will be created automatically)
2) Need to make sure pip is using python version 3.5(or later ) (not sure from here - https://github.com/h2oai/h2o-3/blob/master/docker/Dockerfile.dev)
make sure the syntax substitutions etc are working correct  
cc @hannah-tillman


michalkurka:
@valenad1 idea: instead of investigating what is going wrong with the images - maybe it would be easier to have a specific python environment just for building the documentation (and update the build scripts)

PR #5797, opened 2021-09-30
PUBDEV-8357: Improve error messages in Explain module

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8357

**Includes:**
 - Add error message in R when no model with varimp is found and only
   models with varimp are required
 - Speed up Explain tests: use titanic dataset for regression tests in explain
 - Use only the top SE in best of family plots (default value of `all_stackedensembles` changed to `False`)
 - Add error message when user uses string column in pdp related plots (not
   supported by partialPlot)
   
>**NOTE:** String columns are ignored by most models (those that we support in explain) so the models don't use them   (ModelBuilder drops them), but the user can specify to explain all columns even those that are dropped so I added a check to ensure that we provide some error rather then plotting NaNs.

### DIFF
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -471,7 +471,7 @@ def _get_algorithm(model,  treat_xrt_as_algorithm=False):
     return model.algo
 
 
-def _first_of_family(models, all_stackedensembles=True):
+def _first_of_family(models, all_stackedensembles=False):
     # type: (Union[str, h2o.model.ModelBase], bool) -> Union[str, h2o.model.ModelBase]
     """
     Get first of family models
--- a/h2o-py/h2o/explanation/_explain.py
+++ b/h2o-py/h2o/explanation/_explain.py
@@ -428,7 +428,6 @@ class NumpyFrame:
             yield col, self.get(col, with_categorical_names)
 
 
-
 def _get_domain_mapping(model):
     """
     Get a mapping between columns and their domains.
@@ -998,6 +997,9 @@ def pd_plot(
     if target is not None and not isinstance(target, list):
         target = [target]
 
+    if frame.type(column) == "string":
+        raise ValueError("String columns are not supported!")
+
     color = plt.get_cmap(colormap)(0)
     with no_progress():
         plt.figure(figsize=figsize)
@@ -1114,6 +1116,10 @@ def pd_multi_plot(
                 raise ValueError("Only one target can be specified!")
             target = target[0]
         target = [target]
+
+    if frame.type(column) == "string":
+        raise ValueError("String columns are not supported!")
+
     if _is_automl_or_leaderboard(models):
         all_models = _get_model_ids_from_automl_or_leaderboard(models)
     else:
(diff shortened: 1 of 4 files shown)


h2o-py/h2o/explanation/_explain.py
```diff
@@ -2722,6 +2739,14 @@ def explain_row(
             warnings.warn("No model with variable importance. Selecting all features to explain.")
             columns_of_interest = _get_xy(models_to_show[0])[0]
 
+    # Make sure that there are no string columns to explain as they are not supported by pdp
+    # Usually those columns would not be used by algos so this just makes sure to exclude them
+    # if user specifies top_n=float('inf') or columns_of_interest=x etc.
+    dropped_string_columns = [col for col in columns_of_interest if frame.type(col) == "string"]
+    if len(dropped_string_columns) > 0:
+        warnings.warn("Dropping string columns as they are not supported: {}".format(dropped_string_columns))
+        columns_of_interest = [col for col in columns_of_interest if frame.type(col) != "string"]
```
sebhrusen, line 2748: *fancy*: if you  want to avoid the double iteration, one of my favourite utility fn:
```python
def partition(iterable, predicate=bool):
    iter1, iter2 = tee((predicate(it), it) for it in iterable)
    return (it for b, it in iter1 if b), (it for b, it in iter2 if not b)
```

h2o-r/h2o-package/R/explain.R
```diff
@@ -3394,7 +3412,14 @@ h2o.explain_row <- function(object,
...
     }
   }
-
+  # Make sure that there are no string columns to explain as they are not supported by pdp
+  # Usually those columns would not be used by algos so this just makes sure to exclude them
+  # if user specifies top_n = Inf or columns_of_interest = x etc.
+  dropped_string_columns <- Filter(function(col) h2o.getTypes(newdata[[col]])[[1]] == "string", columns_of_interest)
+  if (length(dropped_string_columns) > 1) {
+    warning(sprintf("Dropping string columns as they are unsupported: %s", paste(dropped_string_columns, collapse = ", ")), call. = FALSE)
+    columns_of_interest <- Filter(function(col) h2o.getTypes(newdata[[col]])[[1]] != "string", columns_of_interest)
```
sebhrusen, line 3421: I guess efficient partitioning is also doable in R, but after scratching my head, doesn't look that simple :)

sebhrusen (approved):
LGTM, thanks!

h2o-py/h2o/explanation/_explain.py
```diff
@@ -472,7 +471,7 @@ def _get_algorithm(model,  treat_xrt_as_algorithm=False):
     return model.algo
 
 
-def _first_of_family(models, all_stackedensembles=True):
+def _first_of_family(models, all_stackedensembles=False):
```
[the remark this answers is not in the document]
tomasfryda: Yes, since all other models had only one "best of family" member, stacked ensembles were handled differently so we show all of them which is not preferred now that there's more of them.

ledell (approved):
Great changes, thank you!

h2o-py/h2o/explanation/_explain.py
tomasfryda: Fancy indeed.

But I'd rather keep it as it is - explicit and clear until we drop support for py2.7 - since `filter` or `map` is eager there I  had to check if `tee` is eager or not and I'd prefer source code to be "simpler". Also the double iteration shouldn't be a performance issue here since the types are cached in the client.

PR #5803, opened 2021-10-04
Fix explain tests [nocheck]

tomasfryda:
I was unable to reproduce the behavior locally so I made it more robust, most likely it will solve the issues `explain` currently has - sometimes there is no model in top N that is not a stacked ensemble and it fails on not having an "explainable" model in that leaderboard slice.

### DIFF
--- a/h2o-r/tests/testdir_misc/runit_explain.R
+++ b/h2o-r/tests/testdir_misc/runit_explain.R
@@ -107,10 +107,10 @@ explanation_test_automl_regression <- function() {
 
   # Leaderboard slice works
   # test explanation
-  expect_true("H2OExplanation" %in% class(h2o.explain(aml@leaderboard[1:4,], train)))
+  expect_true("H2OExplanation" %in% class(h2o.explain(aml@leaderboard[-1,], train)))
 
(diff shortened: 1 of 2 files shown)


h2o-r/tests/testdir_misc/runit_explain.R
```diff
@@ -107,10 +107,10 @@ explanation_test_automl_regression <- function() {
 
   # Leaderboard slice works
   # test explanation
-  expect_true("H2OExplanation" %in% class(h2o.explain(aml@leaderboard[1:4,], train)))
+  expect_true("H2OExplanation" %in% class(h2o.explain(aml@leaderboard[-1,], train)))
```
sebhrusen, line 110: it's just to prove that you can use slices, right? No other specific reason to remove the first line in those tests.

sebhrusen (approved):
thanks, looks more robust against future changes now.

h2o-r/tests/testdir_misc/runit_explain.R
tomasfryda: Exactly.

PR #5804, opened 2021-10-04
PUBDEV-8366  copy also hadoop jars without version [nocheck]

valenad1:


michalkurka (approved):

valenad1:
@michalkurka can we also merge it without Check Pull Request Metadata?

michalkurka:
@valenad1 we could but I think the basic build should run at least

Jenkins didn't register the PR yet - I gave it a little nudge :)

should be running now

PR #5805, opened 2021-10-04
PUBDEV-8366  copy also hadoop jars without version [nocheck]

valenad1:


michalkurka:
thanks - you don't need this PR, just please run the automatic merge job after merging https://github.com/h2oai/h2o-3/pull/5804

valenad1:
True

PR #5812, opened 2021-10-05
Expose python packages in release

valenad1:
We can show python packages in release to debug issues with documentation


michalkurka:
@valenad1 sure, no problem

PR #5814, opened 2021-10-06
PUBDEV-8208: generate delegation token for multiple buckets

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8208

1st approach: get token on import file if not already having it

can be run like:

hadoop jar h2odriver.jar -n 1 -mapperXmx 1g -refreshHdfsTokens -refreshS3ATokens -principal ... -keytab ...


koniecsveta:
@michalkurka hadoop smoke pipeline is green 🍏

PR #5819, opened 2021-10-07
PUBDEV-8370: Fix documentation: MSE -> MAE in Performance and Prediction

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8370

In section `Sensitive to Outliers`, I think we compare wrongly RMSE and MSE. From the context it should be comparing RMSE and MAE.

Reasons to think MSE is there by mistake:
- values from the text don't satisfy `MSE == RMSE**2`
- MSE is as robust to outliers as RMSE is
- MAE is robust to outlier and is an `absolute error` which is mentioned in a paragraph above  as well as in the alt. text of the plot


ledell (approved):
Nice catch, yeah I think you're right.

PR #5827, opened 2021-10-13
[PUBDEV-8377] Increase Jetty Idle Timeout to 5 minutes and Make It Configurable

mn-mikke:
Default limit is 30s on Jetty 9 and 200s on Jetty8.

Sparkling Water conversions fails on this timeout when partitions are big and if the source of the data is slow (JDBC connections).

Problem description: https://github.com/h2oai/sparkling-water/issues/2552


krasinski (approved):
personally I think that PR should be merged regardless of possible improvements discussed above

michalkurka:
> personally I think that PR should be merged regardless of possible improvements discussed above

@krasinski I think the timeout should be consistent between the 2 jetty versions - it doesn't make sense to have it different, it makes sense to change this

however, I am thinking that for regular use cases - the shorter timeout (30s) is better - it allows to catch bad behavior in tests

my thinking here is: should we set the timeout to 30s in H2O and increase in SW because there is a specific use case? @mn-mikke @krasinski

michalkurka:
I think the discussion should continue - merged for now:

1. It is good to unify the values
2. It cannot possibly cause issues at this point

If we decide to decrease to 30s we are free to do it

PR #5837, opened 2021-10-19
GH-7325 Prediction consistency check for constrained models

maurever:
Issue: https://github.com/h2oai/h2o-3/issues/7325

The first idea is to implement assert, which checks the parent prediction is in the interval of children's predictions. 

Quantile distribution is excluded, other monotone constraints tests should pass.

### DIFF
--- a/h2o-algos/src/main/java/hex/tree/DTree.java
+++ b/h2o-algos/src/main/java/hex/tree/DTree.java
@@ -712,10 +712,12 @@ public class DTree extends Iced {
         Arrays.fill(_nids, ScoreBuildHistogram.UNDECIDED_CHILD_NODE_ID);
         return;
       }
-      if(cs != null){
+      if(cs != null) {
         int constr = cs.getColumnConstraint(_split._col);
-        if(nid() != 0 && constr != 0) {
-          assert constr * _split._tree_p0 <= constr* parentPred() &&  constr * parentPred() <= constr * _split._tree_p1;
+        if (!cs._dist._family.equals(DistributionFamily.quantile) && nid() != 0 && constr != 0) {
+            assert constr * _split._tree_p0 <= constr * parentPred() && constr * parentPred() <= constr * _split._tree_p1 :
+                    "Parent prediction and children prediction is not consistent. Parent prediction " + constr * parentPred() +
+                            " should be in interval of the children: " + constr * _split._tree_p0 + " " + constr * _split._tree_p1;
         }
       }
       _splat = _split.splat(hs);
--- a/h2o-algos/src/main/java/hex/tree/DTree.java
+++ b/h2o-algos/src/main/java/hex/tree/DTree.java
@@ -126,6 +126,7 @@ public class DTree extends Iced {
     _leaves = tree._leaves;
     _len = tree._len;
     _depth = tree._depth;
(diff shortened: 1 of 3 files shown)


h2o-algos/src/main/java/hex/tree/DTree.java
```diff
@@ -600,9 +600,17 @@ public DecidedNode(UndecidedNode n, DHistogram hs[], Constraints cs) {
         // which might be because all predictor columns are now constant, or
         // because all responses are now constant.
         _splat = Float.NaN;
-        Arrays.fill(_nids,ScoreBuildHistogram.UNDECIDED_CHILD_NODE_ID);
+        Arrays.fill(_nids, ScoreBuildHistogram.UNDECIDED_CHILD_NODE_ID);
         return;
       }
+      if(cs != null) {
```
michalkurka, line 606: it could be useful to have an option to disable the check - for almost all our self-checks we have a mechanism to defeat them using a system property

maurever: Good point! Is it ok now? https://github.com/h2oai/h2o-3/pull/5837/commits/5fe68994f90632c1700a5c19783cbfaae332a080

wendycwong (approved):

h2o-docs/src/product/data-science/algo-params/monotone_constraints.rst
```diff
@@ -11,6 +11,8 @@ A mapping that represents monotonic constraints. Use +1 to enforce an increasing
 
 **Note**: In GBM and XGBoost, this option can only be used when the distribution is ``gaussian``, ``bernoulli``, ``tweedie``. In GBM also ``quantile`` distribution is supported.
 
+You can enable consistency check using the system property: ``sys.ai.h2o.tree.constraintConsistencyCheck=true``. It is disabled by default.
```
valenad1, line 14: Can we specify what is the meaning of consistency here?

h2o-algos/src/main/java/hex/tree/DTree.java
```diff
@@ -709,9 +722,17 @@ public DecidedNode(UndecidedNode n, DHistogram hs[], Constraints cs, GlobalInter
...
+        int constr = cs.getColumnConstraint(_split._col);
+        if (_tree._checkConstraintConsistency && !cs._dist._family.equals(DistributionFamily.quantile) && nid() != 0 && constr != 0) {
```
valenad1, line 730: Can we have a unit test for this?

valenad1: E.g: https://github.com/h2oai/h2o-3/blob/master/h2o-extensions/xgboost/src/test/java/hex/tree/xgboost/XGBoostTest.java#L86

in xgboost its done like this-

maurever: In GBMTest, another parameter has already been used. The constraint check parameter can be used only for models with constraints, so it only makes sense to test it for some specific use cases. Also, it is computationally expensive, so I created only one test case.

valenad1 (approved):
I see you have implemented my suggestion. 

LGTM, thank you!

PR #5839, opened 2021-10-19
Fix Kerberos documentation for standalone [nocheck]

michalkurka:


michalkurka:
@hannah-tillman thank you for the fixes!

PR #5842, opened 2021-10-19
PUBDEV-7973 glm beta constraints not working

wendycwong:
This PR fixes the problems in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-7973.

This is a two part JIRA.  One is to fix the problem with beta constraints not working and the other one is to add additional ways to specify beta constraints per Megan Kurka suggestion

Done:
1. Fixed bugs associated with beta constraints.  I added additional application of beta constraints to the end of LS if it exists, else, it will be applied after the new beta is calculated;
2. I extended beta constraints to be applicable to IRLSM as well as COORDINATE_DESCENT and LBFGS.
3. Added tests to make sure that the coefficients are constraints are limited as desired.
4. I enabled that python users can now specify beta_constraints using a python dict per Megan Kurka suggestions.
5. Added tests to make sure the new way of specifying beta constraints using python dict works.

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/ComputationState.java
+++ b/h2o-algos/src/main/java/hex/glm/ComputationState.java
@@ -724,7 +724,7 @@ public final class ComputationState {
       }
     }
     ADMM.subgrad(_alpha * _lambda, beta, grad);
-    for (int c : activeCols) // set the error tolerance to the highest error og included columns
+    for (int c : activeCols) // set the error tolerance to the highest error of included columns
       if (grad[c] > err) err = grad[c];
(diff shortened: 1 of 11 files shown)


wendycwong:
Finished the performance test and it does not look like the fixes here affected the training time:

image

michalkurka:
@wendycwong can we also add a self-check that will test that the coefficients after training are within the bounds == the algorithm will do the check itself and will never return unexpected results

wendycwong:
@michalk:  Good idea.

michalkurka (approved):
Thank you Wendy - looks great, well tested

michalkurka (approved):

PR #5843, opened 2021-10-20
HEXDEV-0777: fix regex vulnerability [nocheck]

koniecsveta:
Disclaimer:

> The described vector of attack is merely theoretical and doesn't actually affect H2O deployments. The weakness is found in the Python client. In order to exploit it, the server would have to send malicious headers. Because H2O is in full control of the backend and what headers are sent back the weakness can never actually be exploited.  -- @michalkurka 

https://h2oai.atlassian.net/browse/HEXDEV-777

the regular expression is vulnerable to ReDoS, when it contains:
- grouping with repetition
- inside the repeated group:
    - repetition
    - alternation with overlapping

(https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS)

So in our regex  `(?:.*,)*\s*Negotiate\s*([^,]*),?` the evil part is `(?:.*,)*`

I would modify it to `(?:.*,)?`, so that there will be still a non-capturing group (without backreference) matching whatever ending with`,` but expected 'once or none' instead of 'zero or more' times.

This seems safe to do, because the only usage of this regex is this:
```
def _is_spnego_response(self, response):
    header = response.headers.get('www-authenticate', None)
    return self._header_regex.search(header) is not None
```

So it just returns a boolean whether it can find any regular expression pattern in the header. And `_is_spnego_response()` will return the same boolean value also with the new regex, the only difference lays in how it will group the beginning of the pattern (and there is a non-capturing group anyway) and there won't be repetition in repetition.

Tests running here: http://jenkins:8080/view/H2O-3/job/h2o-3-kerberos-smoke-pipeline/job/zuzana%252FHEXDEV-777%252Ffix_vulnerable_regex/

### DIFF
--- a/h2o-py/tests/testdir_utils/pyunit_auth.py
+++ b/h2o-py/tests/testdir_utils/pyunit_auth.py
@@ -6,18 +6,23 @@ from h2o.auth import SpnegoAuth
 
 
 def test_is_spnego_response():
-    invalid1 = "Digest realm='http-auth@example.org', qop='auth, auth-int', algorithm=MD5, nonce='7ypf/xlj9XXwfDPEoM4URrv/xwf94BcCAzFZH4GiTo0v', opaque='FQhe/qaU925kfnzjCev0ciny7QMkPqMAFRtzCUYo5tdS'"
-    valid1 = "Negotiate realm='http-auth@example.org', qop='auth, auth-int', algorithm=MD5, nonce='7ypf/xlj9XXwfDPEoM4URrv/xwf94BcCAzFZH4GiTo0v', opaque='FQhe/qaU925kfnzjCev0ciny7QMkPqMAFRtzCUYo5tdS'"
-    valid2 = "Basic, Negotiate realm='http-auth@example.org', qop='auth, auth-int', algorithm=MD5, nonce='7ypf/xlj9XXwfDPEoM4URrv/xwf94BcCAzFZH4GiTo0v', opaque='FQhe/qaU925kfnzjCev0ciny7QMkPqMAFRtzCUYo5tdS'"
-    invalid2 = "Basic Negotiate realm='http-auth@example.org', qop='auth, auth-int', algorithm=MD5, nonce='7ypf/xlj9XXwfDPEoM4URrv/xwf94BcCAzFZH4GiTo0v', opaque='FQhe/qaU925kfnzjCev0ciny7QMkPqMAFRtzCUYo5tdS'"
-    invalid3 = "Basic, Negotiatesds realm='http-auth@example.org', qop='auth, auth-int', algorithm=MD5, nonce='7ypf/xlj9XXwfDPEoM4URrv/xwf94BcCAzFZH4GiTo0v', opaque='FQhe/qaU925kfnzjCev0ciny7QMkPqMAFRtzCUYo5tdS'"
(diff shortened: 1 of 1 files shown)


michalkurka:
@sebhrusen the original regex seemed suspicious and I googled it, I found it originates from this library: https://github.com/willthames/urllib_kerberos/blob/fdd0d2954905aa82ff9940d350c0ee8787d379b3/lib/urllib_kerberos/__init__.py#L60

the regex seems to be written that way because this library uses the negotiate_value for some reason: https://github.com/willthames/urllib_kerberos/blob/fdd0d2954905aa82ff9940d350c0ee8787d379b3/lib/urllib_kerberos/__init__.py#L71

I didn't study the code in detail but it is using urllib2 as opposed to our requests package - this means the output could be different in their case and we need a simpler regex.

According to specification, the header value should start with "Negotiate" and can have optional parameters - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate

IMHO we should simplify the regex to follow specification.

sebhrusen:
@michalkurka that's what I was suggesting in https://github.com/h2oai/h2o-3/pull/5843#discussion_r732764134
Note that `Negotiate` is not systematically at the beginning though, you can have sth like `WWW-Authenticate: Basic, Negotiate` (multiple challenges), but if the challenge has params, then it starts.

A better suggestion then:
```python
self._header_regex = re.compile(r'(^|,\s*)Negotiate\b', re.I)
```

proof:
```python
In [1]: import re

In [2]: reg = re.compile(r'(^|,\s*)Negotiate\b', re.I)

In [3]: reg.search("Negotiate")
Out[3]: <re.Match object; span=(0, 9), match='Negotiate'>

In [4]: reg.search("Basic, Negotiate")
Out[4]: <re.Match object; span=(5, 16), match=', Negotiate'>

In [5]: reg.search("Basic, Negotiate, Foo")
Out[5]: <re.Match object; span=(5, 16), match=', Negotiate'>

In [6]: reg.search("Basic Negotiate, Foo")

In [7]: reg.search("Negotiate param=value")
Out[7]: <re.Match object; span=(0, 9), match='Negotiate'>

In [8]: reg.search("NegotiateFoo")

In [9]: reg.search("FooNegotiate")

In [10]: reg.search("Basic param=Negotiate")

In [11]:
```

sebhrusen:
@koniecsveta 
> From ABNF, it seems that it should even be possible to have multiple challenges with params after each other.

the regexp above will catch that, although I don't think it's recommended to mix params with challenges at it makes it more difficult to parse for any client.
```python
In [14]: reg.search("Basic realm=\"apps\", type=1, title=\"Login to \\\"apps\\\"\", Negotiate realm=\"simple\"")
Out[14]: <re.Match object; span=(53, 64), match=', Negotiate'>
```

the problem is that with a single regex, there will always be edge-cases for example:
```python
reg.search('Basic realm="apps", title="Foo, Negotiate, Bar"')
#or
reg.search('Custom realm="apps", negotiate="true"')  # this one is easier to catch
```
Without a list of predefined challenges, I don't think we can handle all those cases only with a regex: we need a proper parser pairing the quotes.

I don't see other implementations trying to handle all those edge-cases though: imo the server should avoid doing something stupid as well.

sebhrusen:
this
```python
reg = re.compile(r'(^|,\s*)Negotiate($|[,\s]+)', re.I)
```
would prevent matching on:
```python
In [27]: reg.search('Custom realm="apps", negotiate="true"')

```

regarding the last edge case:
```python
reg.search('Basic realm="apps", title="Foo, Negotiate, Bar"')
```
I'm not convinced we should try to tackle this: it's a provocation from the server :)

koniecsveta:
> ```python
> r'(^|,\s*)Negotiate($|[,\s]+)'
> ```

yes, that one would be 

> Please add tests. It is a good opportunity to cover the affected code with a unit test: a good candidate to me seems the method `_is_spnego_response`.
> 
> The test should have several examples of what is rejected and several examples of what is accepted.
Test added here: bdcd849

PR #5765, opened 2021-09-15
PUBDEV-8133: rulefit multinomial fix

koniecsveta:

PR #5849, opened 2021-10-27
PUBDEV-8383: consolidate rules

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8383
Rulefit returns rules with splits that are unnecessary.  For example, x < 3 & x < 5.

when https://github.com/h2oai/h2o-3/pull/5765 will be merged i will rebase this PR on rel-zizler

### DIFF
--- a/h2o-algos/src/main/java/hex/rulefit/Condition.java
+++ b/h2o-algos/src/main/java/hex/rulefit/Condition.java
@@ -124,17 +124,17 @@ public class Condition extends Iced {
         if (this.type.equals(Type.Categorical)) {
             expandedNumThreshold = -1;
             
-            List<String> expandedlanguageCatTresholdList = new ArrayList<>();
+            List<String> expandedLanguageCatTresholdList = new ArrayList<>();
             List<Integer> expandedCatTresholdList = new ArrayList<>();
-            expandedlanguageCatTresholdList.addAll(Arrays.asList(this.languageCatTreshold));
+            expandedLanguageCatTresholdList.addAll(Arrays.asList(this.languageCatTreshold));
             expandedCatTresholdList.addAll(Arrays.stream(this.catTreshold).boxed().collect(Collectors.toList()));
             for (int i = 0; i < otherCondition.catTreshold.length; i++) {
                 if (!expandedCatTresholdList.contains(otherCondition.catTreshold[i])) {
                     expandedCatTresholdList.add(otherCondition.catTreshold[i]);
-                    expandedlanguageCatTresholdList.add(otherCondition.languageCatTreshold[i]);
+                    expandedLanguageCatTresholdList.add(otherCondition.languageCatTreshold[i]);
                 }
             }
-            expandedlanguageCatTreshold = expandedCatTresholdList.toArray(new String[0]);
+            expandedlanguageCatTreshold = expandedLanguageCatTresholdList.toArray(new String[0]);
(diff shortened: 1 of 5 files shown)


koniecsveta:
fix version is set to 3.36.0.1 but I would put this into release branch and maybe change the fix version? @michalkurka WDYT?

h2o-algos/src/main/java/hex/rulefit/Condition.java
```diff
@@ -104,4 +109,50 @@ public void map(Chunk[] cs, byte[] out) {
             out[iRow] = newVal;
         }
     }
+
+    Condition expandBy(Condition otherCondition) {
+        assert this.type.equals(otherCondition.type);
+        assert this.operator.equals(otherCondition.operator);
+        assert this.featureIndex == otherCondition.featureIndex;
+        assert this.featureName.equals(otherCondition.featureName);
```
maurever, line 117: What about having a class field `condition` with all the information and implementing the `equals` method in the `Condition` class to make only one assertion here? :) 

```suggestion
        assert this.condition.equals(otherCondition);
```

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
```diff
@@ -77,6 +78,7 @@ public RuleFit(boolean startup_once) {
     public void init(boolean expensive) {
         super.init(expensive);
         if (expensive) {
+            _parms.validate(this);
```
maurever, line 81: Maybe can the validation be disabled by default and a user could use system property to enable it? This is the way how we provide checks like this in tree algos. Or how expensive is this check in the context of this algo? WDYT @michalkurka ?

maurever (commented):
Looks good, just a few suggestions.

h2o-algos/src/main/java/hex/rulefit/Condition.java
koniecsveta: Here I am merging 2 conditions with different thresholds, like x < 4 and x <= 6.5 should result in x < 4, so this is not what I would expect from Condition.equals() method. I am expecting here that there will be a difference which I want to merge into one condition. In this part of code it shouldn't happen that  I will have here for example: x < 4 and z < 5 (difference in feture) or x > 5 and x < 8 (difference in operator) or x < 5 and y in {A, B, C} (difference in type and feature). Hence, I put only asserts here. I can implement Conditions.equals() to check equality of 2 conditions through all properties, but currently I don't need this anywhere.

maurever: Oh, I see. Sure, then it is not necessary to create exactly the `equals` function. Maybe some kind of check function, which returns true if everything is ok, otherwise false. But it is up to you.

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
koniecsveta: currently it checks only 1 check: whether min_rule_length is not greater than max_rule_length. Default values are min_rule_length = 3, max_rule_length = 3, so when you set for example max_rule_length = 2 and forget to set also min_rule_length so that is <= 2, rfit will end up with some quite unclear error, so this is to end up with more informative error on what went wrong

maurever: Ok, make sense to have this check here.

maurever (approved):

michalkurka (approved):

PR #5850, opened 2021-10-28
PUBDEV-8386: Enforce relative redirects in Jetty

michalkurka:
Includes original source code of Response class from Jetty 9.4.11 and Jetty 8.2.0


michalkurka:
Thanks for the comments @mn-mikke - I move the configuration to a single place so now it should be shared for all the different builds

mn-mikke (approved):
LGTM (haven't found anything)

PR #5852, opened 2021-11-02
[PUBDEV-8399] Expose Centers on KmeansMojoModel

mn-mikke:
An open-source user (see https://github.com/h2oai/sparkling-water/issues/2666) would like to access centers on Kmeans models.  This PR is prerequisite for https://h2oai.atlassian.net/browse/SW-2639.


michalkurka:
I think it is okay to do this, however, the MOJO these methods are meant to be called only by callers that also use the low-level score0 API (eg. EasyPredictModelWrapper).

These methods don't provide any kind of annotation and don't explain what is actually being returned.

The caller can thus not know whether the clusters are standardized or not. Also has no way of projecting the clusters back to the original space.

Maybe the information should be taken from a place where it is stored in an annotated form and the users can chose between standardized and original.

wendycwong (approved):

mn-mikke:
I will extract centers from the  serialized model output in MOJO json.

PR #5853, opened 2021-11-02
PUBDEV-8385: add saving parameters to plot functions

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8385

this is the first method, will add these parameters also to other methods from customer preference list (in progress)


h2o-py/h2o/model/extensions/std_coef.py
```diff
@@ -4,12 +4,14 @@
 
 class StandardCoef:
 
-    def _std_coef_plot(self, num_of_features=None, server=False):
+    def _std_coef_plot(self, num_of_features=None, server=False, save_plot_path:str=None, **kwargs):
```
michalkurka, line 7: is this python 2 compatible? I wouldn't think so

h2o-py/h2o/model/extensions/std_coef.py
```diff
@@ -110,6 +112,8 @@ def _std_coef_plot(self, num_of_features=None, server=False):
         plt.yticks(pos[0:num_of_features], feature_labels[0:num_of_features])
         plt.tick_params(axis="x", which="minor", bottom="off", top="off",  labelbottom="off")
         plt.title("Standardized Coef. Magnitudes: H2O GLM", fontsize=20)
+        if save_plot_path is not None:
+            plt.savefig(fname=save_plot_path, **kwargs)
         # show plot
         if server:
             plt.show()
```
michalkurka, line 119: @tomasfryda / @sebhrusen what do you think - should we start returning the plot object instead?

then we don't have to pass all the parameters for saving the plot - instead the caller can do it themselves

tomasfryda: I'm not sure if it's necessary. `plt.gcf()`[1] should return the active figure. So in theory, we could just tell users to do something like:

```python
gbm.varimp_plot(server=True)
plt.gcf().savefig("/Users/user/abc.png")
```

As long as the figure wasn't shown/closed (`server=True`), the current figure should be the last one created.

[1] https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.gcf.html

h2o-py/h2o/model/extensions/std_coef.py
sebhrusen: type annotations: definitely not

h2o-py/h2o/model/extensions/std_coef.py
```diff
@@ -4,12 +4,14 @@
...
 
-    def _std_coef_plot(self, num_of_features=None, server=False):
+    def _std_coef_plot(self, num_of_features=None, server=False, save_plot_path:str=None, **kwargs):
         """
         Plot a GLM model"s standardized coefficient magnitudes.
 
         :param num_of_features: the number of features shown in the plot.
         :param server: if true set server settings to matplotlib and show the graph
+        :param save_plot_path: a path to save the plot via using mathplotlib function savefig
+        :param **kwargs: can be used to pass in additional savefig configuration
```
sebhrusen, line 14: if the kwargs are forwarded to a specific function, then why not name it directly `**savefig`

h2o-py/h2o/model/extensions/std_coef.py
sebhrusen: personally, I would **always** return the plot: it offers much more flexibility/power to the end user.

@tomasfryda using `plt.gcf()` is not very natural or friendly API-wise, it's basically one of those weird property of the `plt` module which stacks various plots and then when you call a function like `varimp_plot` you have to know how many plots were created there if you want to manipulate the stack afterwards: it's breaking almost all principles of API design...
It's nice when you own all the manipulations done using `plt` but we can't rely on this when create plots on our side and expect the user to know what we did exactly.

sebhrusen (changes requested):
also if favour of returning plots, the question is shouldn't we do it for all our plotting functions?

h2o-py/h2o/model/extensions/std_coef.py
tomasfryda: > using plt.gcf() is not very natural or friendly API-wise, it's basically one of those weird property of the plt module which stacks various plots and then when you call a function like varimp_plot you have to know how many plots were created there if you want to manipulate the stack afterwards: it's breaking almost all principles of API design...

I agree - it's just a hacky way that works now without any change.

Potential problem with returning figures is when we show them and then return them - some backends `close()` the figure which makes it hard to show it again (or modify and show)[1] which could make it confusing for the users. 

[1] https://stackoverflow.com/questions/31729948/matplotlib-how-to-show-a-figure-that-has-been-closed

koniecsveta:
> also if favour of returning plots, the question is shouldn't we do it for all our plotting functions?

There is a list of 22 plot functions on the support ticket related to this JIRA, listing which functions should support saving on priority. Other plot functions can be unified after these are. However, there are some, like for example `shap_summary_plot` or 'pd_multi_plot' in the list, which already retrun `plt.gcf()`. Is it save to change change this to `plt`? Or better to save by passing `save_plot_path` in this case? @tomasfryda

h2o-py/h2o/model/extensions/std_coef.py
koniecsveta: fixed 
dbd62d0

h2o-py/h2o/model/extensions/std_coef.py
sebhrusen: > Potential problem with returning figures is when we show them and then return them - some backends close() the figure

@tomasfryda any concrete example? What kind of backend does that?

I'm not sure I understand the difference between:
```python
def myfig():
   plt = …
   plt.show()
   return plt
   
fig = myfig()
```
and
```python
def myfig():
   plt = …
   plt.show()

fig = pyplot.gcf()
```
you're suggesting that depending on the chosen matplotlib backend, in the first case `fig` is sometimes already closed, whereas never in the second?
Where did the backend decide call `pyplot.close(plt)` or `pyplot.close(fig)` exactly?

sebhrusen:
@koniecsveta in `shap_summary_plot` for example, `plt` refers to `pyplot` module not to a fig, contrary to the code in `std-coef.py`.
@tomasfryda could have as well used this instead:
```python
fig = plt.figure(figsize=figsize)
…
return fig
```
which btw, if we already do it, is another argument to keep doing it :)

koniecsveta:
> @koniecsveta in `shap_summary_plot` for example, `plt` refers to `pyplot` module not to a fig, contrary to the code in `std-coef.py`. @tomasfryda could have as well used this instead:
> 
> ```python
> fig = plt.figure(figsize=figsize)
> …
> return fig
> ```
> 
> which btw, if we already do it, is another argument to keep doing it :)
> 
> > There is a list of 22 plot functions on the support ticket related to this JIRA, listing which functions should support saving on priority.
> 
> @koniecsveta are you saying that your current changes are just an appetizer before the main meal?

yes yes, will send you a list

sebhrusen:
after DM with @koniecsveta , suggesting to return a container object like this one:
```python
In [17]: class H2OFigure(object):
    ...:     def __init__(self, *args, fig=None):
    ...:         self.figure=fig
    ...:         self._tuple = args
    ...:
    ...:     def __iter__(self):
    ...:         return iter(self._tuple)
    ...:
    ...:     def __getitem__(self, idx):
    ...:         return self._tuple[idx]
    ...:

In [18]: f = H2OFigure([1, 3, 5], [2, 4, 6, 8], fig="whatever")

In [19]: xs, ys = f

In [20]: xs
Out[20]: [1, 3, 5]

In [21]: ys
Out[21]: [2, 4, 6, 8]

In [22]: f.figure
Out[22]: 'whatever'

In [23]: f[0]
Out[23]: [1, 3, 5]
```

allows us to return the figure in a consistent way and a backwards-compatible way.
Also the method won't return the figure directly, so it shouldn't be automatically closed by nasty matplotlib backends.
And this provides a way to extend the return value for those plotting functions in the future.
wdyt?

h2o-py/h2o/model/extensions/std_coef.py
tomasfryda: Sorry, I should have emphasized that my issue is with the `plt.show()`. In my hacky example, I use the `server=True` which doesn't `show()` the image. So to rephrase that hopefully more clearly - if we show the figure and then return it, it can cause some confusion since you might not be able to view it again.

I think I saw this behavior in some IDE like spyder/pycharm/vscode but I don't really recall the specific one, however,  the easiest way to reproduce is to use the `MacOSX` backend as it is the default one on macOS.

```python
import matplotlib
import matplotlib.pyplot as plt
matplotlib.get_backend() #=> 'MacOSX'
fig = plt.figure()
plt.scatter(range(5), range(5)) #=> <matplotlib.collections.PathCollection object at 0x12ce9c340>
plt.show() # fig.show can be used as well but it doesn't manage the GUI event-loop which makes the window partially unresponsive when used from `python` (as opposed to `IPython`)

# close the window with the figure
plt.show() # Doesn't show anything - no warning/error, no window
fig.show() # Doesn't show anything - no warning/error, no window
```

If we want to return the figure, I think we shouldn't `show()` it by default. Jupyter will plot it even if we don't `show()` it (but it doesn't close it) and for other usages like from CLI I think the user should call the `show()`.

tomasfryda:
@koniecsveta I will be happy to help with (visually) testing it (if there will be some bigger change that would require that) - I still have several environments/IDEs on macOS and linux from the time that I was working on the explain module so hopefully that should catch most of the potential issues.

h2o-py/h2o/model/extensions/std_coef.py
sebhrusen: > I know in this case the user closed the figure so it might seem like it shouldn't be an issue for us

right, you're talking about a user action here. Whatever we do, if the user decides to close the picture window after it is shown the first time, then there will be difficulties to render it again.

> Also in past there were backends that were blocking only so the code didn't continue after it shown the figure until the user closed the window and this might also be relevant for python2 (and matplotlib2).

OK, but we never force the user to use `server=True`, it's not even the default behaviour, so I don't see the issue…

> Jupyter will plot it even if we don't show() it 

see https://github.com/h2oai/h2o-3/pull/5853#issuecomment-958028897 : we're now considering returning a container object, so the figure won't be rendered automatically by Jupyter if it's the last statement of a cell. Therefore the user keeps full control over the figure.

koniecsveta:
> @koniecsveta I will be happy to help with (visually) testing it (if there will be some bigger change that would require that) - I still have several environments/IDEs on macOS and linux from the time that I was working on the explain module so hopefully that should catch most of the potential issues.

@tomasfryda, that would be great, I pushed first 3, more to come

h2o-py/h2o/model/extensions/scoring_history.py
```diff
@@ -69,8 +71,11 @@ def scoring_history_plot(self, timestep, metric, server=False):
             plt.title("Training Scoring History")
             plt.ylim(ylim)
             plt.plot(scoring_history[timestep], scoring_history[training_metric])
+        if save_plot_path is not None:
+            plt.savefig(fname=save_plot_path, fig=fig, **savefig)
```
sebhrusen, line 75: @koniecsveta it's ok to handle the basic use-case (both because we're doing it in several methods already, and because it's convenient for the user):
```python
plt.savefig(fname=save_plot_path)
```
but for all advanced usages, that's what the container is for:
```
scoring_history_plot(…).figure.savefig(path, my_custom_params)
```
therefore, I think we should drop those `**savefig` forwarded params: it's very risky because it may prevent us from adding more params to those plot functions in the future.

h2o-py/h2o/plot_result.py
```diff
@@ -0,0 +1,16 @@
+# -*- encoding: utf-8 -*-
+
+class H2OPlotResult(object):
```
sebhrusen, line 3: can you please move this into a `h2o.plot` submodule: the root module is already overloaded, thanks.

h2o-py/h2o/plot_result.py
```diff
@@ -0,0 +1,16 @@
...
+        self._tuple = args
+
+    def __iter__(self):
+        return iter(self._tuple)
+
+    def __getitem__(self, idx):
+        return self._tuple[idx]
+    
+    def tuple(self):
+        return self._tuple
```
sebhrusen, line 16: is this needed anywhere?
see my latest suggestion: https://github.com/h2oai/h2o-3/pull/5853#issuecomment-958028897

if we want to expose axes values explicitly on this container for example, the `*args` is here for convenience and backwards-compatibility, but I would not encourage users to call `result.tuple()`.
Instead we could decide to expose those with named params:
```python
def my_plot(…):
    xs = …
    ys = …
    fig = plt.figure()
    …
    return H2OPlotResult(xs, ys, figure=figure, xs=xs, ys=ys)
    
# consumer side:
xs, ys  = my_plot()  # backwards-compatible
fig = my_plot().figure  # consistent with all H2O plot functions
res = my_plot()
xs = res.xs    # explicit retrieval of x values
```
This way everything is accessible explicitly (as soon as documented in the `:return:` value for a given plot method).

h2o-py/h2o/model/extensions/std_coef.py
tomasfryda: > OK, but we never force the user to use server=True, it's not even the default behaviour, so I don't see the issue…

@sebhrusen Yes, in my opinion, the problem is that the default is `server=False` which makes the figure to `plt.show()`.
```python
if not server:
    plt.show()
```

So the issues would occur by default and the user would have to use `server=False`. It might be just a documentation issue (I remember spending quite a lot of time trying find out where the issue is).

> see #5853 (comment) : we're now considering returning a container object, so the figure won't be rendered automatically by Jupyter if it's the last statement of a cell. Therefore the user keeps full control over the figure.

I wouldn't be so sure. I tried the following in the Jupyter lab and it renders the figure. I am not sure if I am missing something but I have the feeling that this is not what you'd expect.
```python
import matplotlib.pyplot as plt
class Wrapper:
    def __init__(self):
        plt.figure()
        plt.scatter(range(3), range(3))

w=Wrapper()
print(42)
```
Screen Shot 2021-11-03 at 12 29 40

But as mentioned earlier, Jupyter does use by default backend that doesn't close the figure after rendering so it probably won't cause any other issues (other than potentially showing output in wrong cell as well).

sebhrusen: @tomasfryda I agree this parameter name (`server`) and semantic are a bit esoteric… anything but intuitive.
And unfortunately we have to live with it… except if:
```python
@deprecated_params({'server': (lambda server: {'render': not server})})
def my_plot(…, render=True):  # render: new param with better name/semantic
   if render:
      plt.show()
   return H2OPlotResult(…)
```
it should then be natural for users to switch off the `render` param if having any issue.

for Jupyter, this is interesting, I guess its backend is implicitely calling the `plt.gcf().show()` of sth like that then…
but yeah, if it's not closing anything, no issue there.

h2o-py/h2o/model/extensions/scoring_history.py
koniecsveta: we can drop it: ea118b0

h2o-py/h2o/plot_result.py
koniecsveta: it's not. In tests I see it always extracted together as a tuple, so I would leave it without exposing explicit axes

sebhrusen: @koniecsveta you're right, I was wondering about the case where the user would want both the figure and the axes values…
but it appears that about anything can be then retrieved from the figure object, for example:
```python
xs = figure.axes[0].lines[0].get_xdata()
```
so yeah, I guess you don't need to expose the values explicitly then in this case.

There may other use-cases though, like `permutation_importance_plot`: user can't really retrieve the importance object from the figure, so we'll have to return sth like:
```python
def permutation_importance_plot(…):
    …
    return H2OPlotResult(importance, figure=figure, importance=importance) 
```
if we want the user to have access to both naturally, without doing some tricky assignments, then our current generic container doesn't work well :(
```python
imp = permutation_importance_plot(…)  # legacy code, however imp isn't the importance object anymore
imp, = permutation_importance_plot(…)  # this works with new impl though, but no one was using this because it didn't work with legacy code
imp = permutation_importance_plot(…).importance  # also works with new impl
```

in this case, the only way to get the old code working would be to use:
```python
def permutation_importance_plot(…):
    importance = …
    importance.figure = figure # making the assumption that we can extend the importance object
    return importance
```
it works, but problematic from a design perspective…

koniecsveta: yes I just hit on pdp plot , that's another problematic case

koniecsveta: Maybe I will first do design-perspective-problematic approach on all the functions that are problematic and then try to come up with some nicer solution when I will now what everything is problematic

sebhrusen: thinking about using a combination of mixin+descriptors for the `pdp/pip` use-cases, trying it and see if it can be nicely encapsulated

sebhrusen: what about?
```python
class HTuple(tuple): pass
class HList(list): pass
class HStr(str): pass

def get_decorated_plot_result(res, figure=None):
    # list all special/immutable types that we need first
    if res instanceof tuple:
       dec = HTuple(res)
    elif res instanceof list:
       dec = HList(res)
    elif res instanceof str:
       dec = HStr(res)
    else: # should be an H2O instance, should be mutable
       dec = res
    dec.figure = figure
    return dec
```
then in the plot fn:
```python
def my_plot(…):
   …
   return get_decorated_plot_result(old_result, figure=fig)
```

sebhrusen: updated the suggestion above after realizing that it doesn't handle the case where there was no return value (res=None)

h2o-py/h2o/explanation/_explain.py
```diff
@@ -565,7 +567,8 @@ def shap_summary_plot(
     :param colormap: colormap to use instead of the default blue to red colormap
     :param figsize: figure size; passed directly to matplotlib
     :param jitter: amount of jitter used to show the point density
-    :returns: a matplotlib figure object
+    :param save_plot_path: a path to save the plot via using mathplotlib function savefig
+    :returns: object that contains the resulting matplotlib figure (can be accessed like result.figure)
```
sebhrusen, line 571: `like` doesn't sound right in this context: as/using/with ? Same for all docstr below.
```suggestion
    :returns: object that contains the resulting matplotlib figure (can be accessed using result.figure)
```

h2o-py/h2o/explanation/_explain.py
```diff
@@ -4,6 +4,7 @@
 import warnings
 from contextlib import contextmanager
 from collections import OrderedDict, Counter, defaultdict
+from h2o.plot.plot_result import decorate_plot_result
```
sebhrusen, line 7: @koniecsveta may I ask you a favour?

- `plot_result.py` -> `_plot_result.py` : we want ot make it clear to users that it's not safe to use those modules directly, and that they can change any time.
- create module `h2o/plot/__init__.py` which exports `decorate_plot_result`
- then, please always import directly from `h2o.plot`.
- extra bonus: also move `h2o/utils/ext_dependencies.py` to `h2o/plot/_matplotlib.py` and also ensure it's imported directly from `h2o.plot`
At the end some files should have sth like:
```python
from h2o.plot import get_matplotlib_pyplot, decorate_plot_result
```
which is better structured with all plot-related functions from the same module.

h2o-py/h2o/explanation/_explain.py
```diff
@@ -664,7 +667,9 @@ def shap_summary_plot(
     plt.title("SHAP Summary plot for \"{}\"".format(model.model_id))
     plt.tight_layout()
     fig = plt.gcf()
-    return fig
+    if save_plot_path is not None:
+        plt.savefig(fname=save_plot_path)
+    return decorate_plot_result(res=None, figure=fig)
```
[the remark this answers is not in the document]
koniecsveta: d3db906

h2o-py/h2o/explanation/_explain.py
koniecsveta: okok 👌  d3db906

h2o-py/h2o/explanation/_explain.py
```diff
@@ -4,14 +4,14 @@
 import warnings
 from contextlib import contextmanager
 from collections import OrderedDict, Counter, defaultdict
+from h2o.plot import decorate_plot_result, get_matplotlib_pyplot
```
sebhrusen, line 7: also, sorry for being annoying, let's group `h2o` imports at least.
Usually the structure of imports looks like this, from the most generic available to the most specific:
```python
# imports from Py standard lib

# imports from 3rd party libraries

# imports from your project (here h2o)

# local relative imports (from .x import …)
```
not asking you to fix all existing modules, but just for your new imports.

(Note that we use some exceptions in this project, like the `__future__` and other custom `h2o.compat` import for Py2, which are always on top, but it's mainly to get rid of it quickly…

h2o-py/h2o/model/multinomial.py
```diff
@@ -1,13 +1,13 @@
...
 
 import h2o
+from h2o.utils.compatibility import *  # NOQA
 from h2o.utils.typechecks import assert_is_type
-from ..exceptions import H2OValueError
-from ..frame import H2OFrame
 from .extensions import has_extension
 from .model_base import ModelBase
+from ..exceptions import H2OValueError
+from ..frame import H2OFrame
```
sebhrusen, line 10: actually according to my rule: top-down from furthest to closest, this was better before ;)

(the rationale is: if there is a name conflict and one name is overridden by mistake, then the closest should take precedence to follow usual scoping rules, and make it easier to identify the conflict)

h2o-py/h2o/model/ordinal.py
```diff
@@ -66,17 +66,18 @@ def mean_per_class_error(self, train=False, valid=False, xval=False):
...
         """
         Plots training set (and validation set if available) scoring history for an H2OOrdinalModel. The timestep
         and metric arguments are restricted to what is available in its scoring history.
 
         :param timestep: A unit of measurement for the x-axis.
         :param metric: A unit of measurement for the y-axis.
+        :param save_plot_path: a path to save the plot via using mathplotlib function savefig
 
-        :returns: A scoring history plot.
+        :returns: Object that contains the resulting scoring history plot (can be accessed like result.figure).
```
sebhrusen, line 78:
```suggestion
        :returns: Object that contains the resulting scoring history plot (can be accessed using result.figure).
```

h2o-py/h2o/model/regression.py
```diff
@@ -14,14 +14,16 @@ class H2ORegressionModel(ModelBase):
...
         """
         Plots training set (and validation set if available) scoring history for an H2ORegressionModel. The timestep
         and metric arguments are restricted to what is available in its scoring history.
 
         :param timestep: A unit of measurement for the x-axis.
         :param metric: A unit of measurement for the y-axis.
-        :returns: A scoring history plot.
+        :param save_plot_path: a path to save the plot via using mathplotlib function savefig
+
+        :returns: Object that contains the resulting scoring history plot (can be accessed like result.figure).
```
sebhrusen, line 26:
```suggestion
        :returns: Object that contains the resulting scoring history plot (can be accessed using result.figure).
```

h2o-py/h2o/plot/_plot_result.py
```diff
@@ -0,0 +1,26 @@
...
+        dec = _MList(res)
+    elif isinstance(res, dict):
+        dec = _MDict(res)
+    elif isinstance(res, str):
+        dec = _MStr(res)
+    else: # should be an H2O instance, should be mutable
+        dec = res
+    dec.figure = figure
+    return dec
+
```
sebhrusen, line 25: can you please add a test for this, covering all basic use-cases? just to be sure it works exactly as expected, thx

```python
figure = "let's pretend I'm a figure"

res = decorate_plot_result(figure=figure)
assert res.figure == figure

res = decorate_plot_result((3, 5), figure=figure)
assert isinstance(res, tuple)
a, b = res
assert a == 3
assert b == 5
assert res.figure == figure

res = decorate_plot_result([3, 5, 6], figure=figure)
assert isinstance(res, list)
assert res == [3, 5, 6]
assert res.figure == figure

# same with dict

 class Foo(object):
    def __init__(self, bar=None):
        self.bar=bar

res = decorate_plot_result(Foo(bar="baz"), figure=figure)
assert isinstance(res, Foo)
assert res.bar == "baz"
assert res.figure == figure
```

sebhrusen (approved):
thanks @koniecsveta please add the tests for the new plot utility function

h2o-py/h2o/plot/_plot_result.py
koniecsveta: tests are added on places where underlying plot functions are being tested. But can add also test like this

koniecsveta: d535954

h2o-py/h2o/model/dim_reduction.py
```diff
@@ -94,20 +93,23 @@ def proj_archetypes(self, test_data, reverse_transform=False):
...
         :param str type: either ``"barplot"`` or ``"lines"``.
         :param bool server: if true set server settings to matplotlib and do not show the graph
+        :param save_plot_path: a path to save the plot via using mathplotlib function savefig
+        
+        :returns: Object that contains the resulting scree plot (can be accessed like result.figure).
         """
         # check for matplotlib. exit if absent.
         plt = get_matplotlib_pyplot(server)
         if plt is None:
             return
```
michalkurka, line 111: @koniecsveta If we do not return anything - wouldn't that cause strange errors?

cc: @sebhrusen

michalkurka:
@koniecsveta @sebhrusen @tomasfryda LGTM, thank you for all the work that went into it.

Just one question - what should the behavior be if matplotlib is not available - there is at least one place where we return "nothing". Do we need to change this as well in order to have informative error messages?

h2o-py/h2o/model/dim_reduction.py
sebhrusen: we could call `get_matplotlib_pyplot(server, raise_if_not_available=True)` in most of those plots I guess.
I don't see the point of just ignoring here if we can't produce anything.

koniecsveta: I think, if it would, there already would be some. For example  `_plot_pr` already returns "something" but on matplotlib absent it returns anything. But we can add warning on matplotlib absent

sebhrusen: well, `get_matplotlib_pyplot(server)` will still print a message if `matplotlib` is missing, even when it doesn't raise.
I'm not sure I understand the rationale behind that though: when user asks something and we can't do it, we should raise imo. @tomasfryda?

sebhrusen: @koniecsveta do you mean that because some of those plot functions return something (eg. p/r) then we still want to return those values even if the plot can't be drawn?
It was a strange original design indeed: for example, there are much better ways to obtain the p/r than calling the plot function… so here, by not raising, it's as if we privilege the secondary contract (return value) over the first one (drawing a plot).

michalkurka: the thing to keep in mind - prior to this change the functions didn't have much use in scripts, users could see a warning that something is wrong

with this change we are enabling usage in scripts and we need to think about proper error reporting and what we return to the caller

tomasfryda: > I'm not sure I understand the rationale behind that though: when user asks something and we can't do it, we should raise imo. @tomasfryda?
@sebhrusen This was the behavior before refactoring it into the function so the rationale was to be backward compatible. If we are ok to break the backward compatibility I'd be in favor of always raising an error.

michalkurka: we cannot break backwards compatibility in a fix release cycle

can we instead raise the error when the returned object is accessed?

koniecsveta: I think we can, when figure will be private and getter will check whether there was saved some raise_error_flag into it

tomasfryda (changes requested):
Breaks the explain rendering in jupyter. To reproduce try this in jupyter:
```python
import h2o
h2o.connect()
x = [
   "pclass", "sex", "age", "sibsp", "parch", "ticket",
   "fare", "cabin", "embarked", "home.dest",
   "cabin_type", "family_size", "family_type", "title"
]
y = "survived"

df = h2o.import_file("https://h2o-public-test-data.s3.amazonaws.com/smalldata/titanic/titanic_expanded.csv", na_strings=["", " ", "NA"])


df[y] = df[y].asfactor()

train, test = df.split_frame([.75])
aml = H2OAutoML(max_runtime_secs=60)
aml.train(x=x, y=y, training_frame=train)

aml.explain(test);
```

h2o-py/h2o/plot/_plot_result.py
sebhrusen: thanks, I just don't understand why it's tested in a GLM test file...
`h2o/module/` -> `[ tests/testdir_module/pyunit_test_a.py, tests/testdir_module/pyunit_test_b.py, …]`
or at least  `tests/testdir_misc/pyunit_test_module.py`

I know that the legacy file structure is messy and annoying, but it's not a reason to keep using it.

Imagine someone adding a function in that module, do you think this person will search for existing tests in glm?

koniecsveta: will move all of them to pyunit_PUBDEV-8385_add_saving_parameters_to_plot_functions.py

sebhrusen: are you putting `PUBDEV-xxxx` in all your tests file names?
I don't know who came with this first.
When a ticket is closed, it should be… closed. And we shouldn't care why a test was written: a test is always written to test existing code, no matter if a ticket was previously filed for it (because it wasn't tested…) or just because it's code, so whenever possible, it should be tested.

koniecsveta: not to all.. I think it's good to know the ticket number because you can easily find underlying jira -> underlying support ticket -> underlying information why something was done. It can be found in a commit message but this can get lost after some refactoring, so I don't think it's such a bad thing to have a ticket number in a test name.

sebhrusen: but again, you have a `_plot_result.py` module. In one month, you modify or extend this module, will you rather search for `pyunit_plot_result` or `pyunit_pubdev-8385`??
The pb with `pubdev-xxx` in the name is that it makes sense only for the person who writes the test, and only at the time of the PR, for others and later it's just meaningless noise.

koniecsveta:
> Breaks the explain rendering in jupyter. To reproduce try this in jupyter:
> 
> ```python
> import h2o
> from h2o.automl import H2OAutoML
> h2o.connect()
> x = [
>    "pclass", "sex", "age", "sibsp", "parch", "ticket",
>    "fare", "cabin", "embarked", "home.dest",
>    "cabin_type", "family_size", "family_type", "title"
> ]
> y = "survived"
> 
> df = h2o.import_file("https://h2o-public-test-data.s3.amazonaws.com/smalldata/titanic/titanic_expanded.csv", na_strings=["", " ", "NA"])
> 
> 
> df[y] = df[y].asfactor()
> 
> train, test = df.split_frame([.75])
> aml = H2OAutoML(max_runtime_secs=60)
> aml.train(x=x, y=y, training_frame=train)
> 
> aml.explain(test);
> ```

should be ok now

h2o-py/h2o/model/dim_reduction.py
koniecsveta: b9567fe

tomasfryda (approved):
Everything seems to be working correctly. Great job, thank you @koniecsveta !

michalkurka (approved):
Looks great - thank you @koniecsveta and many thanks for suggestions and improvements from @sebhrusen and @tomasfryda

h2o-py/h2o/plot/_plot_result.py
```diff
@@ -0,0 +1,34 @@
...
+class _MTuple(tuple): pass
+class _MList(list): pass
+class _MDict(dict): pass
+class _MStr(str): pass
+class Error(EnvironmentError): pass
+
+def decorate_plot_result(res=None, figure=None):
+    def get_figure():
+        if figure != "RAISE_EXCEPTION_FLAG":
+            return figure
```
sebhrusen, line 13: magic string!
The pythonic way:
```python
RAISE_ON_FIGURE_ACCESS = object()

def get_figure():
     if figure is not RAISE_ON_FIGURE_ACCESS:
         return figure
     ...
     
 # client code
from h2o.plot import RAISE_ON_FIGURE_ACCESS
...
return decorate_plot_result(figure=RAISE_ON_FIGURE_ACCESS)
```
this is safe: no magic string, everything is verified by compiler

h2o-py/h2o/plot/_plot_result.py
```diff
@@ -0,0 +1,34 @@
...
+    elif isinstance(res, list):
+        dec = _MList(res)
+    elif isinstance(res, dict):
+        dec = _MDict(res)
+    elif isinstance(res, str):
+        dec = _MStr(res)
+    else: # should be an H2O instance, should be mutable
+        dec = res
+    dec.figure = get_figure
+    dec.is_decorated_plot_result = True
```
sebhrusen, line 30: I would not expose this attribute "publicly": don't forget that this appears in auto-completion then… and you use this flag only internally.
```suggestion
    dec._is_decorated_plot_result = True
```

As you seem to check this attribute in many places, I'd also encourage to expose a utility function instead: safer, no magic string:
```python
def is_decorated_plot_result(obj):
    return hasattr(obj, "_is_decorated_plot_result")
```

sebhrusen (commented):
great improvements! I have more comments though (I'm here for that :))

h2o-py/h2o/plot/_plot_result.py
```diff
@@ -0,0 +1,34 @@
+# -*- encoding: utf-8 -*-
+# mutable versions of py immutable types
+class _MObject(object): pass
+class _MTuple(tuple): pass
+class _MList(list): pass
+class _MDict(dict): pass
+class _MStr(str): pass
+class Error(EnvironmentError): pass
```
sebhrusen, line 8: can you also use a more specific class name? It feels like you're stealing global `Error` (which doesn't exist but...).
Also `EnvironmentError` is kinda deprecated and is used for non-python errors (e.g. I/O)

I think you could use an existing H2O exception, for example the generic `H2OError`

koniecsveta: d5b08bb

h2o-py/h2o/plot/_plot_result.py
koniecsveta: 52d6678

sebhrusen (approved):
thanks @koniecsveta , please double check the imports, it looks like some are missing

h2o-py/h2o/model/dim_reduction.py
```diff
@@ -94,20 +93,23 @@ def proj_archetypes(self, test_data, reverse_transform=False):
...
+        :param save_plot_path: a path to save the plot via using mathplotlib function savefig
+        
+        :returns: Object that contains the resulting scree plot (can be accessed like result.figure()).
         """
         # check for matplotlib. exit if absent.
         plt = get_matplotlib_pyplot(server)
         if plt is None:
-            return
-
+            return decorate_plot_result(figure=RAISE_ON_FIGURE_ACCESS)
```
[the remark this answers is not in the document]
koniecsveta: yes this is weird, noticed also before, but when I add it there, idea displays it in grey with "from h2o.plot import decorate_plot_result, get_matplotlib_pyplot, RAISE_ON_FIGURE_ACCESS" message and it builds successfully also without this import . But added in 1a66ba6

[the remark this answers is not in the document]
sebhrusen: oh, I see! It's due to the 
```python
from .metrics_base import *
```
this is not usually recommended if the module didn't define the `__all__` variable, as otherwise it will import everything in the module scope, including the variables/functions/classes/modules that were imported in that module…
Nevermind, you can leave the import, we can fix those `*` another time.

h2o-py/tests/testdir_algos/glm/pyunit_plot_functions__add_saving_parameter_and_decorate_plot_result.py
```diff
@@ -0,0 +1,279 @@
...
+    
+    # Build and train a GLM model
+    cars_glm = H2OGeneralizedLinearEstimator()
+    cars_glm.train(x=predictors, y=response_col, training_frame=cars_train, validation_frame=cars_valid)
+
+    # test saving:
+    tmpdir = tempfile.mkdtemp(prefix="h2o-func")
+    path="{}/plot1.png".format(tmpdir)
+    test_plot_result_saving(cars_glm.std_coef_plot(server=True), "{}/plot2.png".format(tmpdir),
+                            cars_glm.std_coef_plot(server=True, save_plot_path=path), path)
```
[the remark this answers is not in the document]
koniecsveta: 44f4c62

h2o-py/tests/testdir_algos/glm/pyunit_plot_functions__add_saving_parameter_and_decorate_plot_result.py
```diff
@@ -0,0 +1,279 @@
...
+    test_plot_result_saving(cars_glm.varimp_plot(server=True),"{}/plot2.png".format(tmpdir),
+                            cars_glm.varimp_plot(server=True, save_plot_path=path), path)
+
+    
+
+
+if __name__ == "__main__":
+    pyunit_utils.standalone_test(tests)
+else:
+    tests()
```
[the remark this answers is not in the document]
koniecsveta: 44f4c62

PR #5858, opened 2021-11-03
PUBDEV-8400:Add seed to tests to make it predictable.

wendycwong:
This JIRA fixes the issue here: https://h2oai.atlassian.net/browse/PUBDEV-8400.

Veronika:  This is the second time you helped me find my intermittent test failures.  Appreciate it.  Thank you so much!  Wendy


maurever (approved):
Thanks, @wendycwong! Even though the timeout causes the failure, the seed is always useful to set!

PR #5860, opened 2021-11-04
AbstractBuildVersion should not be defined in both h2o-core and h2o-genmodel [nocheck]

michalkurka:
h2o-core depends on h2o-genmodel, when the independent jars are on classpath
this results in a classpath clash

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ h2o-r/h2o-package/DESCRIPTION
 gradle-app.setting
 .gradle
 src/main/java/water/BuildVersion.java
+h2o-genmodel/src/main/java/water/genmodel/BuildVersion.java
 .Rproj.user
 .RData
 .Rhistory
--- a/buildSrc/src/main/java/H2OBuildVersion.java
+++ b/buildSrc/src/main/java/H2OBuildVersion.java
@@ -19,6 +19,7 @@ public class H2OBuildVersion {
   // Passed in by caller.
   File _rootDir;
   String _versionFromGradle;
+  String _targetPackage;
 
   // Calculated.
   String _branch;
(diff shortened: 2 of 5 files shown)


michalkurka:
@mn-mikke this can be affecting SW

h2o-genmodel/src/main/java/water/genmodel/BuildVersion.java
```diff
@@ -0,0 +1,10 @@
+package water.genmodel;
+
+public class BuildVersion extends AbstractBuildVersion {
```
achraf-mer, line 3: is this fix temporary or will this file get re-generated automatically on new releases?

mn-mikke:
Based on the quick look into the SW codebase, SW shouldn't be affected.

h2o-genmodel/src/main/java/water/genmodel/BuildVersion.java
michalkurka: thanks for asking - this file shouldn't be committed and I need need to adjust .gitignore as well

michalkurka:
> Based on the quick look into the SW codebase, SW shouldn't be affected.

@mn-mikke In case someone doesn't use the SW assembly jar and instead adds the jar to Spark classpath - there might be a conflict of classes and it might fail if the order of jars is not as expected

h2o-genmodel/src/main/java/water/genmodel/BuildVersion.java
michalkurka: fixed, thanks for pointing it out

achraf-mer (approved):
LGTM, except the SW part, will defer to others. Thanks

koniecsveta (approved):

PR #5861, opened 2021-11-04
Clean-up TypeMap and make sure we get mappings from leader [nocheck]

michalkurka:
- Remove assert with bad side-effects
- Tests shouldn't hang when there is an AssertionError in h2o-core
- Lock cloud before asking leader for a class type mapping
- Dump TypeMap if a leader node is missing a class mapping
- Remove unused method

### DIFF
--- a/h2o-core/src/main/java/water/AutoBuffer.java
+++ b/h2o-core/src/main/java/water/AutoBuffer.java
@@ -782,21 +782,25 @@ public final class AutoBuffer implements AutoCloseable {
   }
 
   public <T extends Freezable> T get() {
-    int id = getInt();
-    if( id == TypeMap.NULL ) return null;
-    if( _is!=null ) {
-      id = remapFrozenId(id);
-    }
-    return (T)TypeMap.newFreezable(id).read(this);
+    return getFreezable(null);
   }
+
   public <T extends Freezable> T get(Class<T> tc) {
+    if (tc == null)
+      throw new IllegalArgumentException("Class argument cannot be null");
+    return getFreezable(tc);
+  }
+
+  @SuppressWarnings("unchecked")
+  private <T extends Freezable> T getFreezable(Class<T> tc) {
     int id = getInt();
     if( id == TypeMap.NULL ) return null;
     if( _is!=null ) {
       id = remapFrozenId(id);
     }
-    assert tc.isInstance(TypeMap.theFreezable(id)):tc.getName() + " != " + TypeMap.theFreezable(id).getClass().getName() + ", id = " + id;
-    return (T)TypeMap.newFreezable(id).read(this);
+    return (T) TypeMap
+            .newFreezable(id, tc)
+            .read(this);
   }
 
   private int remapFrozenId(int id) {
--- a/h2o-core/src/main/java/water/H2O.java
+++ b/h2o-core/src/main/java/water/H2O.java
@@ -2622,5 +2622,18 @@ final public class H2O {
   public static URI downloadLogs(String destinationDir, String logContainer) {
     return LogsHandler.downloadLogs(destinationDir, LogArchiveContainer.valueOf(logContainer));
   }
-  
+
+  /**
+   * Is this H2O cluster running in our continuous integration environment?
+   * 
+   * This information can be used to enable extended error output or force shutdown in case
+   * an error is encountered. Use responsibly.
+   * 
+   * @return true, if running in CI
+   */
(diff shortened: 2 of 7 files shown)


h2o-core/src/main/java/water/TCPReceiverThread.java
```diff
@@ -163,6 +163,10 @@ public void run() {
           // On any error from anybody, close everything
           System.err.println("IO error");
           Log.err("IO error on TCP port " + H2O.H2O_PORT + ": ", e);
+          if (e instanceof Error && H2O.isCI()) {
+            // could be AssertionError, OOM...
+            throw H2O.fail("Encountered an error while running in CI - will terminate to prevent deadlocks", e);
```
sebhrusen, line 168: is the idea to observe this on CI first, and then investigate+fix to avoid deadlocks in production?

h2o-core/src/test/java/water/TypeMapTest.java
```diff
@@ -0,0 +1,32 @@
...
+            TypeMap.printTypeMap(ps);
+        }
+        String output = os.toString();
+        String[] lines = output.split("\n");
+        String[] bootstrapClasses = TypeMap.bootstrapClasses();
+        assertTrue(bootstrapClasses.length >= lines.length);
+        for (int i = 0; i < bootstrapClasses.length; i++) {
+            assertEquals(i + " -> " + bootstrapClasses[i] + " (map: " + i + ")", lines[i]);
+        }
+        TypeMap.printTypeMap(System.out);
```
sebhrusen, line 29: you may want to print to stdout first: it will provide some help if one assertion fails.

sebhrusen (approved):
nice changes, should help to see more clearly through the ice.

h2o-core/src/main/java/water/AutoBuffer.java
```diff
@@ -782,21 +782,25 @@ public AutoBuffer put(Freezable f) {
...
-    if( _is!=null ) {
-      id = remapFrozenId(id);
-    }
-    return (T)TypeMap.newFreezable(id).read(this);
+    return getFreezable(null);
   }
+
   public <T extends Freezable> T get(Class<T> tc) {
+    if (tc == null)
+      throw new IllegalArgumentException("Class argument cannot be null");
```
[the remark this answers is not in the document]
michalkurka: get and get(Class<T>) both delegate to getFreezable

but if the caller chooses to use the version with the class argument, the argument shouldn't be null

Does it make sense?

h2o-core/src/main/java/water/TCPReceiverThread.java
michalkurka: exactly, these kinds of issues frequently manifest themselves in our CI, and rarely we do see stuff like this reported from users

in CI I want to fail right away - because oftentimes people mistake these hangs for timeouts and they are not reported as issues

for non-CI (= production, use) - even if the cluster hangs on one specific algo - it doesn't make it generally unusable - some of these states are recoverable and we definitely don't want to shutdown the cluster that might have some completed work that wasn't saved

h2o-core/src/test/java/water/TypeMapTest.java
michalkurka: good point, thanks

h2o-core/src/main/java/water/H2O.java
```diff
@@ -2622,5 +2622,18 @@ public static URI downloadLogs(String destinationDir, LogArchiveContainer logCon
   public static URI downloadLogs(String destinationDir, String logContainer) {
     return LogsHandler.downloadLogs(destinationDir, LogArchiveContainer.valueOf(logContainer));
   }
-  
+
+  /**
+   * Is this H2O cluster running in our continuous integration environment?
+   * 
+   * This information can be used to enable extended error output or force shutdown in case
+   * an error is encountered. Use responsibly.
```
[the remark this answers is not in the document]
michalkurka: it is a dangerous method if used wrong; after your comment, I removed the public modifier at least

h2o-core/src/main/java/water/AutoBuffer.java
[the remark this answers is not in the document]
michalkurka: yes, it is ensure the caller is not accidentally passing null instead of the class - the only purpose of the class is to be able to check the returned instance is of the expected type

if I start accidentally passing null - I won't trigger the check

PR #5862, opened 2021-11-05
PUBDEV-8208: update documentation

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8208

### DIFF
--- a/h2o-docs/src/product/getting-data-into-h2o.rst
+++ b/h2o-docs/src/product/getting-data-into-h2o.rst
@@ -570,7 +570,7 @@ Example command:
 
 **Note on refreshHiveTokens:** The provided keytab will be copied over to the machine running the H2O Cluster leader node. For this reason, we strongly recommended that both YARN and HDFS be secured with encryption.
 
-**Note on generating the refreshing HDFS delegation tokens:** In case generation of the refreshing HDFS delegation tokens is required, the ``-refreshHdfsTokens`` argument has to be present. In case S3A filesystem is used, additionally, the ``refreshS3ATokens`` argument has to be present.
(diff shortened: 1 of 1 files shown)


h2o-docs/src/product/getting-data-into-h2o.rst
```diff
@@ -556,7 +556,7 @@ Requirements:
 - The ``hiveHost``, ``hivePrincipal`` and optionally ``hiveJdbcUrlPattern`` arguments are present. (See above for details.)  
 - The ``principal`` argument is set with the value of the users's Kerberos principal.
 - The ``keytab`` argument set pointing to the file with the user's Kerberos keytab file.
-- The ``refreshTokens`` argument is present.
+- The ``refreshHiveTokens`` argument is present.
```
michalkurka, line 559: thank you

h2o-docs/src/product/getting-data-into-h2o.rst
```diff
@@ -566,9 +566,11 @@ Example command:
           -nodes 1 -mapperXmx 4G \
           -hiveHost hostname:10000 -hivePrincipal hive/hostname@EXAMPLE.COM \
           -pricipal user/host@DOMAIN.COM -keytab path/to/user.keytab \
-          -refreshTokens
+          -refreshHiveTokens
 
-**Note on refreshTokens:** The provided keytab will be copied over to the machine running the H2O Cluster leader node. For this reason, we strongly recommended that both YARN and HDFS be secured with encryption.
+**Note on refreshHiveTokens:** The provided keytab will be copied over to the machine running the H2O Cluster leader node. For this reason, we strongly recommended that both YARN and HDFS be secured with encryption.
+
+**Note on generating the refreshing HDFS delegation tokens:** In case generation of the refreshing HDFS delegation tokens is required, the ``-refreshHdfsTokens`` argument has to be present. In case S3A filesystem is used, additionally, the ``refreshS3ATokens`` argument has to be present.
```
michalkurka, line 573: > In case S3A filesystem is used, additionally, the ``refreshS3ATokens`` argument has to be present.

This is not always the case, only in specific deployments. I think it should say something like "in specific deployments (eg. on CDP with IDbroker security) you might need to enable S3A token refresh to acquire (and keep refreshing) delegation tokens to access S3 buckets"

koniecsveta: ok, will modify, thanks!

michalkurka (approved):

PR #5863, opened 2021-11-05
PUBDEV-8384: Explain floating point issue for POJOs

michalkurka:


g-eoj (approved):
Informative notebook! I don't see any issues.

michalkurka:
thanks for reading through it @g-eoj

neemah2o:
Thanks for creating this @michalkurka . If this becomes important for more users, it would be good to have a keyword option to change to doubles rather than use `rapids()`. Like `my_gbm.download_pojo(java_output_as_doubles=True)`

michalkurka:
@neemah2o sure, we can expose it directly if there is demand for it - definitely; not a problem

PR #5866, opened 2021-11-08
PUBDEV-8403: Fix printing DRF when there are no OOB samples

michalkurka:
This is not meant to resolve the jira, just to fix the show method on DRF trained without OOB samples

### DIFF
--- a/h2o-py/h2o/model/metrics_base.py
+++ b/h2o-py/h2o/model/metrics_base.py
@@ -128,11 +128,11 @@ class MetricsBase(h2o_meta()):
             print("Mean Residual Deviance: " + str(self.mean_residual_deviance()))
         if metric_type in types_w_logloss:
             print("LogLoss: " + str(self.logloss()))
-        if metric_type in ['ModelMetricsBinomial', 'ModelMetricsBinomialGeneric']:
+        if metric_type in ['ModelMetricsBinomial', 'ModelMetricsBinomialGeneric'] or \
+            metric_type in types_w_mult or \
+            metric_type in ['ModelMetricsOrdinal', 'ModelMetricsOrdinalGeneric']:
             # second element for first threshold is the actual mean per class error
-            print("Mean Per-Class Error: %s" % self.mean_per_class_error()[0][1])
-        if metric_type in types_w_mult or metric_type in ['ModelMetricsOrdinal', 'ModelMetricsOrdinalGeneric']:
-            print("Mean Per-Class Error: " + str(self.mean_per_class_error()))
+            print("Mean Per-Class Error: %s" % self._mean_per_class_error())
         if metric_type in types_w_glm:
             if metric_type == 'ModelMetricsHGLMGaussianGaussian': # print something for HGLM
                 print("Standard error of fixed columns: "+str(self.hglm_metric("sefe")))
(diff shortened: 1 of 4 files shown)


michalkurka:
I think the main issue is that mean_per_class_error has a different semantics of the return value in H2OModelMetricsBinomial - this also creates inconsistency between binomial and eg ordinal models - https://github.com/h2oai/h2o-3/pull/5866/files#diff-706188769f6ea2ae280cab5624635816376bfe8e71b52c098dfd8a88a9ee26f0L134

h2o-py/h2o/model/metrics_base.py
```diff
@@ -128,11 +128,11 @@ def show(self):
             print("Mean Residual Deviance: " + str(self.mean_residual_deviance()))
         if metric_type in types_w_logloss:
             print("LogLoss: " + str(self.logloss()))
-        if metric_type in ['ModelMetricsBinomial', 'ModelMetricsBinomialGeneric']:
+        if metric_type in ['ModelMetricsBinomial', 'ModelMetricsBinomialGeneric'] or \
+            metric_type in types_w_mult or \
+            metric_type in ['ModelMetricsOrdinal', 'ModelMetricsOrdinalGeneric']:
```
sebhrusen, line 133: recommended way for multiline statements in Py is to wrap everything with parenthesis instead of escaping end-lines. Also, starting next line with `or` makes it more clear that it's a continuation of the previous line imo.

h2o-py/h2o/model/metrics_base.py
```diff
@@ -567,6 +569,10 @@ def null_degrees_of_freedom(self):
             return self._metric_json["null_degrees_of_freedom"]
         return None
 
+    # private accessor for mean per-class error - the public version is overridden in H2OBinomialModelMetrics with
+    # a method with different return semantics
```
sebhrusen, line 573: so it means that for binomial tasks, we always print the backend value only, right?
This same name with different semantics is confusing indeed.

Btw, `H2OBinomialModelMetrics` doesn't exist, it's `H2OBinomialModel` mixin in Py and `ModelMetricsBinomial` on backend.

michalkurka: > so it means that for binomial tasks, we always print the backend value only, right?
yes, we print the backend values without re-calculating the same thing

> Btw, H2OBinomialModelMetrics doesn't exist, it's H2OBinomialModel mixin in Py and ModelMetricsBinomial on backend.

I think it does - https://github.com/h2oai/h2o-3/blob/master/h2o-py/h2o/model/metrics_base.py#L850:

```
class H2OBinomialModelMetrics(MetricsBase):
    """
    This class is essentially an API for the AUC object.
    This class contains methods for inspecting the AUC for different criteria.
    To input the different criteria, use the static variable `criteria`.
    """
```

sebhrusen: my bad! this `metrics_base` module has many many classes :)

h2o-py/h2o/model/metrics_base.py
michalkurka: thank you - fixed

sebhrusen (approved):
thanks @michalkurka

PR #5867, opened 2021-11-08
PUBDEV-8404: Classification stopping metrics cannot be used for regression

michalkurka:


h2o-core/src/main/java/hex/ScoreKeeper.java
```diff
@@ -119,31 +119,33 @@ else if (m instanceof ModelMetricsMultinomial) {
...
 
     private final ConvergenceStrategy _convergence;
     private final boolean _lowerBoundBy0;
+    private final boolean _classificationOnly;
 
-    StoppingMetric(ConvergenceStrategy convergence, boolean lowerBoundBy0) {
+    StoppingMetric(ConvergenceStrategy convergence, boolean lowerBoundBy0, boolean classificationOnly) {
       _convergence = convergence;
       _lowerBoundBy0 = lowerBoundBy0;
+      _classificationOnly = classificationOnly;
```
sebhrusen, line 148: shouldn't we use a 3-values variable instead of a boolean: 'classification', 'regression', 'all' ? (for example `deviance` doesn't fit in your changes) or maybe even an array filled with 'binary', 'multiclass', 'regression'? Although I think that we now have a multiclass AUC/AUC_PR.

Otherwise, I like the idea of having it as part of the StoppingMetric enum.

sebhrusen (commented):
I would also make it work for regression.

h2o-core/src/main/java/hex/ScoreKeeper.java
michalkurka: You could use regression metric to early stop a classification task. Did you meant that?

I think the first attempt is just to make it better for classification - we can change further but I was not going to solve it for all cases in this PR (there are cases like anomaly_score & custom)

sebhrusen: > You could use regression metric to early stop a classification task. Did you mean that?

Yeah, as we'll also need this logic in AutoML, I think it would be better to have sth like:
```
metric.isApplicableTo(task_type)

with task_type being an enum
```
Even if you don't solve all cases right now, I think we could use this semantic immediately, and then later focus on the individual metrics: for `custom`, I guess we can always return true as it's the user responsibility. For `anomaly_score`, we could add an `anomaly_detection` task type (currently, only IF fits in this category if I'm correct). My point is that it's a very flexible approach and we can use it in various places, it will always be better than all those `if sth == this or that`.

michalkurka: I understand your reasoning, however, the issue of "is applicable to" is that it seems to provide the final verdict - while the problem can be more nuanced - like special cases

```        
if (isClassifier()) {
          if (_parms._stopping_metric == ScoreKeeper.StoppingMetric.deviance && !getClass().getSimpleName().contains("GLM")) {
            error("_stopping_metric", "Stopping metric cannot be deviance for classification.");
          }
        } else {
```

So I think in terms of making it better - the PR right as-is is a good compromise to me :)

We can refactor it further but in this PR I would just like to essentially extend the black list that was previously coded in model builder.

sebhrusen: Yeah, I understand, the problem is that all those `if-else` leads to (and are caused by) hacky unstructured code, code that is not reusable, which leads to copy/paste with small modifications logic that is not kept in sync with the original, and finally bugs.

Ideally, the example above could probably be rewritten as :
```java
if (this.supportsMetric(_parms._stopping_metric, task_type) { ...

//with default implementation:
protected boolean supportsMetric(StoppingMetric metric, TaskType task) {
    return metric.isApplicableTo(task);
}

//allowing GLM to override `supportsMetric` to always accept deviance 
```

I understand your concern for this PR and I'm approving it now. My own concern is that I want to be able to reuse this logic in AutoML without having to copy-paste a bunch of `if-else` logic  that will always be out-of-sync: there's (or was) a bunch of those already, especially with metrics.

sebhrusen (approved):

h2o-core/src/main/java/hex/ScoreKeeper.java
michalkurka: Thanks for approving, you are right in the problem formulation.

The issue is that the ModelBuilders are can override the checks in whichever way they want - so even if we have this logic outside of MB, the MB can still add an "if stopping_metric = something then" - and everything will work fine - except the common places like AutoML.

That makes me think we need a better design on the MB level - perhaps break down the init method into smaller pieces and make sure there is just one place where you are validating things like stopping metrics and so on.

Something I am not yet sure how to do but I would appreciate your input.

PR #5868, opened 2021-11-09
H2OTree: Demonstrate how to map node ids to decision paths [nocheck]

michalkurka:


g-eoj (commented):
I didn't realize `index()` could be used to index a substring. Neat!

PR #5869, opened 2021-11-09
Check if release is from correct branch [nocheck]

valenad1:


michalkurka (approved):
Great! Thank you @valenad1 

Please add to rel-zizler

PR #5872, opened 2021-11-10
PUBDEV-8361: Fix local build warnings for python documentation

hannah-tillman:
For: [PUBDEV-8361](https://h2oai.atlassian.net/browse/PUBDEV-8361)

Dropped from 141 warnings to only 2. Builds fine locally, but would appreciate a second set of eyes to make sure nothing got upset by the changes.


h2o-py/h2o/automl/_estimator.py
```diff
@@ -195,13 +196,8 @@ def __init__(self,
...
         :param exclude_algos: List the algorithms to skip during the model-building phase. 
-            The full list of options is: 
-                ``"DRF"`` (Random Forest and Extremely-Randomized Trees),
-                ``"GLM"``,
-                ``"XGBoost"``,
-                ``"GBM"``,
-                ``"DeepLearning"``,
-                ``"StackedEnsemble"``. 
+            The full list of options is: ``"DRF"`` (Random Forest and Extremely-Randomized Trees), ``"GLM"``, ``"XGBoost"``, ``"GBM"``,
+            ``"DeepLearning"``, ``"StackedEnsemble"``.
```
sebhrusen, line 200: I don't know how you feel about this, but personally each time there are let's say >3 options like this, I think it's much nicer for the user to have them rendered as some kind of list in the generated doc. Also it allows us to put a description or quick explanation next to some items, without being visually annoying.
I tried the following approaches:
1. proper list:
```suggestion
            The full list of options is:
            
                * ``"DRF"`` (Random Forest and Extremely-Randomized Trees)
                * ``"GLM"``
                * ``"XGBoost"``
                * ``"GBM"``
                * ``"DeepLearning"``
                * ``"StackedEnsemble"`` 
                
```
2. code block: 
```suggestion
            The full list of options is::
            
                "DRF" (Random Forest and Extremely-Randomized Trees)
                "GLM"
                "XGBoost"
                "GBM"
                "DeepLearning"
                "StackedEnsemble" 
                
```
My preference goes for the second one: the rendering is more condensed but visually very easy to identify all the options.  Also note that in this case, there's no need for backsticks to highlight the options. wdyt?

Not asking you to change all the long lists here, but this one could be a good start and serve as an example for further improvements.

h2o-py/h2o/frame.py
```diff
@@ -3941,19 +3944,19 @@ def convert_H2OFrame_2_DMatrix(self, predictors, yresp, h2oXGBoostModel, return_
...
 
         3. Derive the parameters for native XGBoost:
-        nativeParams = h2oModelD.convert_H2OXGBoostParams_2_XGBoostParams()
+        ``nativeParams = h2oModelD.convert_H2OXGBoostParams_2_XGBoostParams()``
 
         4. Train your native XGBoost model and generate a prediction:
-        nativeModel = xgb.train(params=nativeParams[0], dtrain=nativeDMatrix, num_boost_round=nativeParams[1])
-        nativePredict = nativeModel.predict(data=nativeDMatrix, ntree_limit=nativeParams[1].
+        ``nativeModel = xgb.train(params=nativeParams[0], dtrain=nativeDMatrix, num_boost_round=nativeParams[1])``
+        ``nativePredict = nativeModel.predict(data=nativeDMatrix, ntree_limit=nativeParams[1].``
```
sebhrusen, line 3959: What about a single code block for this:
```suggestion
        Follow the steps below to compare H2OXGBoost and native XGBoost::
        
            # 1. Train the H2OXGBoost model with H2OFrame trainFile and generate a prediction:
            h2oModelD = H2OXGBoostEstimator(**h2oParamsD) # parameters specified as a dict()
            h2oModelD.train(x=myX, y=y, training_frame=trainFile) # train with H2OFrame trainFile
            h2oPredict = h2oPredictD = h2oModelD.predict(trainFile)

            # 2. Derive the DMatrix from H2OFrame:
            nativeDMatrix = trainFile.convert_H2OFrame_2_DMatrix(myX, y, h2oModelD)

            # 3. Derive the parameters for native XGBoost:
            nativeParams = h2oModelD.convert_H2OXGBoostParams_2_XGBoostParams()

            # 4. Train your native XGBoost model and generate a prediction:
            nativeModel = xgb.train(params=nativeParams[0], dtrain=nativeDMatrix, num_boost_round=nativeParams[1])
            nativePredict = nativeModel.predict(data=nativeDMatrix, ntree_limit=nativeParams[1].

            # 5. Compare the predictions h2oPredict from H2OXGBoost, nativePredict from native XGBoost.
```
I find the result more appealing visually and users can copy paste it integrally in a Jupyter notebook for example.

sebhrusen (commented):
some suggestion: I think we need to use code blocks more often in our doc.

hannah-tillman:
@sebhrusen I love these suggestions! I updated the two you suggested in this PR and created [this JIRA](https://h2oai.atlassian.net/browse/PUBDEV-8442) for continuing to update readability :)

h2o-py/h2o/automl/_estimator.py
```diff
@@ -195,13 +196,15 @@ def __init__(self,
...
             (as long as the same training frame, or a sample, is used in subsequent runs).
         :param exclude_algos: List the algorithms to skip during the model-building phase. 
-            The full list of options is: 
-                ``"DRF"`` (Random Forest and Extremely-Randomized Trees),
-                ``"GLM"``,
-                ``"XGBoost"``,
-                ``"GBM"``,
-                ``"DeepLearning"``,
-                ``"StackedEnsemble"``. 
+            The full list of options is:
```
sebhrusen, line 199: does it render correctly? It needs to end with a double-colon `::` to start a code block:
```suggestion
            The full list of options is::
```

hannah-tillman: Good catch--I was testing out the other list version as well and thought I put it all back to normal. Thanks!

sebhrusen: also, just noticed this existing use of lists for a similar case in our doc: https://docs.h2o.ai/h2o/latest-stable/h2o-py/docs/frame.html#h2o.H2OFrame.distance

maybe we should use the same format/sphynx syntax everywhere then

hannah-tillman: yes good catch! also slipped in another minor update with the usage examples (thanks to your previous suggestions)

h2o-py/h2o/frame.py
```diff
@@ -3938,25 +3941,25 @@ def convert_H2OFrame_2_DMatrix(self, predictors, yresp, h2oXGBoostModel, return_
...
         column. There can be NAs in any columns.
 
-        Follow the steps below to compare H2OXGBoost and native XGBoost:
+        Follow the steps below to compare H2OXGBoost and native XGBoost::
 
-        1. Train the H2OXGBoost model with H2OFrame trainFile and generate a prediction:
-        h2oModelD = H2OXGBoostEstimator(**h2oParamsD) # parameters specified as a dict()
-        h2oModelD.train(x=myX, y=y, training_frame=trainFile) # train with H2OFrame trainFile
-        h2oPredict = h2oPredictD = h2oModelD.predict(trainFile)
+            1. Train the H2OXGBoost model with H2OFrame trainFile and generate a prediction:
```
sebhrusen, line 3946: if we use a single code block, the idea is to add a `#` in front of the numbered bullets: the benefit is that it makes the items look different from the code itself, and as mentioned earlier, this way the entire block can be copy/paster safely into a Python interpreter.

h2o-py/h2o/frame.py
```diff
@@ -3435,10 +3436,12 @@ def distance(self, y, measure=None):
     def drop_duplicates(self, columns, keep="first"):
         """
         Drops duplicated rows across specified columns.
+        
         :param columns: Columns to compare during the duplicate detection process.
         :param keep: Which rows to keep. Two possible values: ["first", "last"]. The "first" value (default) keeps
          the first row and deletes the rest. The "last" value keeps the last row.
-        :return: A new H2OFrame with rows deduplicated
+         
+        :returns: A new H2OFrame with rows deduplicated
```
sebhrusen, line 3444: are you encouraging us to use `returns` instead of `return`? Afaik both are valid and I tend to use the latter, but for the sake of consistency, I'm ready to use the conjugated version and pay attention to this when reviewing other's PR as well.

sebhrusen (approved):
Thanks @hannah-tillman looks good to me!

h2o-py/h2o/frame.py
hannah-tillman: Okay so when I first started working with the python docs I only ever saw `:returns:` so I thought that it was standard and it was like halfway through doing this update that I learned both are valid. 

I do think moving forward for consistency we should stick to `:returns:`, however. I might tack that on to the PR for py doc improvements, too.

PR #5873, opened 2021-11-10
[nocheck] reorganize AutoML Py tests to be better distributed and faster on Jenkins

sebhrusen:
focusing on Py tests in this PR, will do R separately


h2o-py/tests/testdir_algos/automl/leaderboard/pyunit_automl_leaderboard.py
```diff
@@ -0,0 +1,60 @@
...
+import sys
+
+sys.path.insert(1, os.path.join("..","..",".."))
+from h2o.automl import H2OAutoML
+from tests import pyunit_utils as pu
+
+pu.load_module("_leaderboard_utils", os.path.dirname(__file__))
+pu.load_module("_automl_utils", os.path.join(os.path.dirname(__file__), ".."))
+from _leaderboard_utils import all_algos, check_leaderboard
+from _automl_utils import import_dataset, get_partitioned_model_names
```
sebhrusen, line 12: @michalkurka found a way to load local files in the `tests` subfolders, which allows me to share test functions between test files: it would be nice if we could start using this instead of putting about every function in overloaded `utils.py` and improve organization of new tests.
I can post in slack after merging this.

michalkurka: this is very cool - could there be a heuristic that would load all the "utility" modules automatically to avoid repeating lines 9-10 everywhere?

michalkurka (approved):
Perfect thank you

h2o-py/tests/testdir_algos/automl/leaderboard/pyunit_automl_leaderboard.py
sebhrusen: @michalkurka  good idea, we could automatically load all files starting with underscore in the current folder or in one parent (until `tests` root).
I would still need to pass the current file though so it would look like:
```python
pu.load_utilities(__file__)
from _leaderboard_utils import all_algos, check_leaderboard
from _automl_utils import import_dataset, get_partitioned_model_names
```

sebhrusen: or automatically load all `_xxx` files under `tests` when loading `utils.py`…
Hopefully we wont create thoushands of those… (I'm not afraid of that).

michalkurka: could we get the file from sys.argv? it is not necessary - it doesn't really matter - just an idea

sebhrusen: `sys.argv` would not work as it depends how the test is executed, and by default test are "sourced" using `exec`

sebhrusen: I'm trying the "load all" approach, providing a way to fail early on name conflicts, should be enough

h2o-py/tests/testdir_algos/automl/leaderboard/pyunit_automl_leaderboard_binary.py
```diff
@@ -0,0 +1,82 @@
+from __future__ import print_function
+import os
+import sys
+
+sys.path.insert(1, os.path.join("..","..","..",".."))
+from h2o.automl import H2OAutoML
+from tests import pyunit_utils as pu
+
+from _leaderboard_utils import check_leaderboard, check_model_property
+from _automl_utils import import_dataset, get_partitioned_model_names
```
sebhrusen, line 10: @michalkurka this is as much magic as I can sacrifice to, you pushed my boundaries :)
it's not very pythonic (the provenance of `_automl_utils` in the parent folder is not completely obvious here, although not difficult to find in this case), but I think it is clear enough.

if 2 utility modules with the same name are created, it will fail immediately with a descriptive message, encouraging people to give a more precise name to the module (instead of just `_utils.py`)

tomasfryda (approved):
Looks good to me. Thank you @sebhrusen !

PR #5877, opened 2021-11-11
Speed-up AutoML tests [nocheck]

michalkurka:
This is the first iteration. The next, step is to derive # clouds from # available CPU cores on a specific Jenkins machine.

Many AutoML tests are working on small datasets - eg. prostate - that is a single chunk, so there is no parallelism in model building except for parallelism of CV itself - if you have 5 fold cv for prostate you are running all of them independently in 5 threads, this is followed by the main model which is running 1 thread => having just one cloud would therefore waste time.

Having 2 clouds should help with utilizing more cores.


sebhrusen (approved):
I see: sounds good to me

PR #5879, opened 2021-11-12
[nocheck] Add python 3.7 to prerequisites

tomasfryda:
We are using python 3.7 in automated tests in jenkins (for about 17 months[1]) so I think we should update the prerequisites.

[1] https://github.com/h2oai/h2o-3/blame/rel-zizler/scripts/jenkins/groovy/defineTestStages.groovy#L45

### DIFF
--- a/h2o-dist/index.html
+++ b/h2o-dist/index.html
@@ -554,7 +554,7 @@
             <div id="quickstart-py" style="display:none">
                 <h2>Use H<sub>2</sub>O directly from Python</h2>
 
-                <p>1. Prerequisite:    Python 2.7.x, 3.5.x, or 3.6.x</p>
+                <p>1. Prerequisite:    Python 2.7.x, 3.5.x, 3.6.x, or 3.7.x</p>
                 <p>2. Install dependencies (prepending with `sudo` if needed):</p>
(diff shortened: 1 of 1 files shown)


h2o-dist/index.html
```diff
@@ -554,7 +554,7 @@ <h2>Get started with H<sub>2</sub>O in 3 easy steps</h2>
             <div id="quickstart-py" style="display:none">
                 <h2>Use H<sub>2</sub>O directly from Python</h2>
 
-                <p>1. Prerequisite:    Python 2.7.x, 3.5.x, or 3.6.x</p>
+                <p>1. Prerequisite:    Python 2.7.x, 3.5.x, 3.6.x, or 3.7.x</p>
```
sebhrusen, line 557: can we put sth like 3.5.x to 3.7.x? especially as 3.10.x is released now and we should start to run Jenkins against it as well and I don't see us listing all versions from 3.5

tomasfryda: Sure... Done.

valenad1 (approved):
Thanks! 🚀

sebhrusen (approved):
thx @tomasfryda

PR #5880, opened 2021-11-12
PUBDEV-8405: follow up after add saving parameter

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8405
- added `save_plot_path` parameter also to h2o.frame.H2OFrame.hist since this was not in the list on support ticket, but its also a plot function.
- did alias on functions where there was originally used '`save_to_file'` parameter so that it works also with '`save_plot_path`'

I would not add save_plot_path to varimp_heatmap() and model_correlation_heatmap(), since those already return a plot and there are varimp() and model_correlation() to get data used to build them or do you think it should be added also there?

### DIFF
--- a/h2o-py/h2o/frame.py
+++ b/h2o-py/h2o/frame.py
@@ -26,7 +26,7 @@ from h2o.exceptions import H2OTypeError, H2OValueError, H2ODeprecationWarning
 from h2o.expr import ExprNode
 from h2o.group_by import GroupBy
 from h2o.job import H2OJob
-from h2o.plot import get_matplotlib_pyplot
+from h2o.plot import get_matplotlib_pyplot, decorate_plot_result, RAISE_ON_FIGURE_ACCESS
 from h2o.utils.config import get_config_value
 from h2o.utils.shared_utils import (_handle_numpy_array, _handle_pandas_data_frame, _handle_python_dicts,
                                     _handle_python_lists, _is_list, _is_str_list, _py_tmp_key, _quoted,
@@ -3831,7 +3831,7 @@ class H2OFrame(Keyed):
             expr=ExprNode("table", self, dense))
 
 
-    def hist(self, breaks="sturges", plot=True, **kwargs):
+    def hist(self, breaks="sturges", plot=True, save_plot_path=None, **kwargs):
         """
         Compute a histogram over a numeric column.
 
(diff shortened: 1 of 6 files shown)


tomasfryda:
> I would not add save_plot_path to varimp_heatmap() and model_correlation_heatmap(), since those already return a plot and there are varimp() and model_correlation() to get data used to build them or do you think it should be added also there?

I think it would be nice to have python client consistent so I would add that there too. (`varimp_plot()` has it and we also have `varimp()` to get the data used to build it)

h2o-py/h2o/model/model_base.py
```diff
@@ -1088,6 +1088,8 @@ def partial_plot(self, data, cols=None, destination_key=None, nbins=20, weight_c
         :param targets: Target classes for multiclass model.
         :returns: Plot and list of calculated mean response tables for each feature requested.
         """
+        if save_to_file is None:  # if not supplied by 'real name'
+            save_to_file = kwargs.get('save_plot_path')
```
sebhrusen, line 1092: if the idea is to deprecate `save_to_file` param for consistency (which I approve), we have a decorator for this:
```python
@deprecated_params({'save_to_file': 'save_plot_path'})
def partial_plot(self, data, …, save_plot_path=None, …):
    …
```
for new code, it will promote `save_plot_path` (as `save_to_file` won't appear anymore in the list of params, nor in the doc).
for legacy code, it will translate on the fly `save_to_file` param into `save_to_path`, which I believe is what we want to achieve.

h2o-py/h2o/model/metrics_base.py
```diff
@@ -1419,6 +1419,9 @@ def plot(self, type="roc", server=False, save_to_file=None, plot=True):
         
         """
 
+        if save_to_file is None:  # if not supplied by 'real name'
+            save_to_file = kwargs.get('save_plot_path')
+
```
sebhrusen, line 1424: see comment below

h2o-py/h2o/frame.py
```diff
@@ -3831,17 +3831,19 @@ def table(self, data2=None, dense=True):
             expr=ExprNode("table", self, dense))
 
 
-    def hist(self, breaks="sturges", plot=True, **kwargs):
+    def hist(self, breaks="sturges", plot=True, save_plot_path=None, **kwargs):
```
sebhrusen, line 3834: interesting, strange use of `kwargs` in this method, it is used for a param that is usually public (`server`), I guess we'll never know why :)

sebhrusen (changes requested):

h2o-py/h2o/model/model_base.py
koniecsveta: Thanks! be2b8e3

sebhrusen (approved):
looks good, thx @koniecsveta

PR #5882, opened 2021-11-12
PUBDEV-8407: Fix GAM weight bug

wendycwong:
This PR fixes the problem in : https://h2oai.atlassian.net/browse/PUBDEV-8407

The weight vectors are tracked by Scope and hence when the model building process ends, it will be deleted.
I fixed the error by removing the Scope.track(weight_v).


koniecsveta (approved):

michalkurka (approved):
Thank you @wendycwong

PR #5892, opened 2021-11-16
PUBDEV-8421 add model output ymu and rank to GLMModelOutputV3

wendycwong:
This PR complete the ask in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8421

Added the fields to GLMModelOutputV3.java and make sure they are copied from ModelOutput.
Added python test to check the fields are properly filled.

### DIFF
--- a/h2o-algos/src/main/java/hex/glm/GLMModel.java
+++ b/h2o-algos/src/main/java/hex/glm/GLMModel.java
@@ -1161,6 +1161,7 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
   public static class GLMOutput extends Model.Output {
     Submodel[] _submodels = new Submodel[0];
     DataInfo _dinfo;
+    double[] _ymu;
     public String[] _coefficient_names;
     String[] _random_coefficient_names; // for HGLM
     String[] _random_column_names;
@@ -1229,6 +1230,8 @@ public class GLMModel extends Model<GLMModel,GLMModel.GLMParameters,GLMModel.GLM
     }
     
     public int rank() { return _submodels[_selected_submodel_idx].rank();}
+    
+    public double[] ymu() { return _ymu;}
 
     public boolean isStandardized() {
       return _dinfo._predictor_transform == TransformType.STANDARDIZE;
--- a/h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
+++ b/h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
@@ -132,6 +132,8 @@ public class GLMModelV3 extends ModelSchemaV3<GLMModel, GLMModelV3, GLMModel.GLM
           }
           standardized_coefficient_magnitudes = new TwoDimTableV3();
           standardized_coefficient_magnitudes.fillFromImpl(tdt);
+        } else {
+          rank = 0;
         }
 
       return this;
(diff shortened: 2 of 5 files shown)


h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -53,6 +53,12 @@
 
     @API(help="Starting lambda value used when lambda search is enabled.")
     double lambda_max;
+    
+    @API(help="ymu value")
+    double[] ymu;
+    
+    @API(help="rank")
+    double rank;
```
mn-mikke, line 61: Shouldn't this be an `int`?

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
```diff
@@ -53,6 +53,12 @@
 
     @API(help="Starting lambda value used when lambda search is enabled.")
     double lambda_max;
+    
+    @API(help="ymu value")
```
mn-mikke, line 57: Would it be possible to add more detailed descriptions to new fields? Sparkling Water in one of the future versions will expose these fields on `H2OMOJOModel` and descriptions in `@API` notation will serve as source for auto-generated documentation.

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_8421_check_ymu_rank.py
```diff
@@ -0,0 +1,30 @@
+import sys
+sys.path.insert(1,"../../../")
+import h2o
+from tests import pyunit_utils
+from h2o.estimators.glm import H2OGeneralizedLinearEstimator as glm
+
+# check parameters ymu and rank were passed 
+def test_gaussian_alpha():
```
mn-mikke, line 8: Thanks for adding a test for a new functionality, but I'm not sure whether this functionality is worth of the time spend on CI resources (training a new model because in order to check propagation of two fields).

mn-mikke (approved):
I left few minor comments, otherwise LGTM. Thanks a lot @wendycwong for this PR and your help!!!

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
wendycwong: Yes, good catch

h2o-algos/src/main/java/hex/schemas/GLMModelV3.java
wendycwong: Sure, done.

h2o-py/tests/testdir_algos/glm/pyunit_PUBDEV_8421_check_ymu_rank.py
wendycwong: Let me make it a very short test.

h2o-algos/src/main/java/hex/glm/GLMModel.java
```diff
@@ -1161,6 +1161,7 @@ public void setZValues(double [] zValues, double dispersion, boolean dispersionE
   public static class GLMOutput extends Model.Output {
     Submodel[] _submodels = new Submodel[0];
     DataInfo _dinfo;
+    double[] _ymu;
```
michalkurka, line 1164: I believe this information is redundant - this seems to be the same value as "prior class distribution"

Screen Shot 2021-11-24 at 10 30 55 AM

maybe we should expose that instead?

michalkurka: there is a difference - for regression, this value holds the mean response

wendycwong: @michalk:  @marek needs the two parameters to generate a metric class as follows: 

  public GLMMetricBuilder(String[] domain, double [] ymu, GLMWeightsFun glmf, int rank, boolean computeMetrics, boolean intercept, MultinomialAucType aucType)

That is why he is looking for ymu and rank variable.

You are right that the ymu and the priorClassDist.  They are almost the same except in the case of regression.  That is why I choose to use ymu.  Is this okay?

maurever (approved):
Looks good to me, thanks @wendycwong.

PR #5893, opened 2021-11-16
Clean-up IF grid search tests [nocheck]

michalkurka:
- Extended IF: Fix grid search test - the purpose is not to test parallelism
- Clean-up grid search tests - get rid off some redundant copy&paste


valenad1 (approved):
Thanks!

PR #5896, opened 2021-11-18
PUBDEV-8408: rulefit: label observations that match rules

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8408


h2o-bindings/bin/custom/R/gen_rulefit.py
```diff
@@ -12,6 +12,57 @@
...
+#' splits <- h2o.splitFrame(data = titanic, ratios = .8, seed = 1234)
+#' train <- splits[[1]]
+#' test <- splits[[2]]
+#' 
+#' rfit <- h2o.rulefit(y = response, x = predictors, training_frame = train, validation_frame = test, 
+#' min_rule_length = 1, max_rule_length = 10, max_num_rules = 100, seed = 1, model_type="rules")
+#' h2o.fit_rules(rfit, train, c("M1T0N7, M1T49N7, M1T16N7", "M1T36N7", "M2T19N19"))
+#' }
+#' @export
+h2o.fit_rules <- function(model, frame, rule_ids) {
```
michalkurka, line 43: @koniecsveta can you please validate the naming in #h2o-3-api-updates?

h2o-bindings/bin/custom/R/gen_rulefit.py
```diff
@@ -12,6 +12,57 @@
...
+    o <- model
+    if (is(o, "H2OModel")) {
+        if (o@algorithm == "rulefit"){
+            parms <- list()
```
michalkurka, line 47: Seems like this code could be simplified if we used a rapid expression evaluation

this would avoid the need to write a custom handler and this custom call

koniecsveta: e3e99b0 + 0a103f7

h2o-bindings/bin/custom/R/gen_rulefit.py
```diff
@@ -12,6 +12,51 @@
...
+#' rfit <- h2o.rulefit(y = response, x = predictors, training_frame = train, validation_frame = test, 
+#' min_rule_length = 1, max_rule_length = 10, max_num_rules = 100, seed = 1, model_type="rules")
+#' h2o.fit_rules(rfit, train, c("M1T0N7, M1T49N7, M1T16N7", "M1T36N7", "M2T19N19"))
+#' }
+#' @export
+h2o.fit_rules <- function(model, frame, rule_ids) {
+    o <- model
+    if (is(o, "H2OModel")) {
+        if (o@algorithm == "rulefit"){
+            expr <- sprintf('(rulefit.fit.rules %s %s %s)', model@model_id, h2o.getId(frame), .collapse.char(rule_ids))
```
michalkurka, line 47: it would be better to use

H2OFrame._expr(expr=ExprNode("rulefit.fit.rules", model, False, frame))

something like this - this way you do take the advantage of lazy evaluation of the expression

michalkurka: oh sorry, this was R

so similar to this :)

the suggestion as such applies to Python

h2o-py/h2o/estimators/targetencoder.py
```diff
@@ -26,7 +26,7 @@ class H2OTargetEncoderEstimator(H2OEstimator):
     algo = "targetencoder"
     supervised_learning = True
 
-    @deprecated_params({'k': 'inflection_point', 'f': 'smoothing', 'noise_level': 'noise'})
```
michalkurka, line 29: @sebhrusen looks like the generator doesn't guarantee a stable order?

sebhrusen: nasty! will look at this.
how was it discovered? why the change in TE? does anyone really used all TE args as positional args? do people really enjoy unreadble code?

sebhrusen: actually the decorator leaves the positional args (when passed by the user) as-is, and only remaps named args, so I don't see what you mean exactly.

h2o-py/h2o/estimators/targetencoder.py
```diff
@@ -394,8 +394,8 @@ def seed(self, seed):
         self._parms["seed"] = seed
 
     k = deprecated_property('k', inflection_point)
-    f = deprecated_property('f', smoothing)
     noise_level = deprecated_property('noise_level', noise)
+    f = deprecated_property('f', smoothing)
```
sebhrusen, line 398: what's happening here? how useful is it to move this?

h2o-py/h2o/estimators/targetencoder.py
michalkurka: @sebhrusen the issue is that in this PR target encoder was not touched (at least I don't see it) but we do have a different generated code

sebhrusen: ok ok, totally misunderstood! it's a generator issue, not a decorator issue :))

sebhrusen: ok, I see, I guess the generator is using a simple dict (not ordered), and different versions of Py may read the keys in different order.

sebhrusen: I can test a change and submit it to @koniecsveta to include it in this PR

sebhrusen: this was generated using Py2.7: it should actually fail in the PR build as this one uses Py3 (confirmation: http://mr-0xg1:8080/blue/organizations/jenkins/h2o-3-pipeline/detail/PR-5896/11/pipeline).
@koniecsveta I'll suggest you to just rebuild locally with Py3 before pushing, I'll fix the consistent ordering issue separately, thx.

michalkurka: @sebhrusen I see - not worth the fix then - should we instead add a check in the build script that will require Python 3+?

sebhrusen: that's a good idea: we don't need to support 2.7 for the build phase

koniecsveta: thanks for clarification! I overlooked that I was building outside of my env for h2o3

koniecsveta: rebuilt here f163b9a

h2o-bindings/bin/custom/R/gen_rulefit.py
koniecsveta: thanks! improved here a85735f

h2o-bindings/bin/custom/R/gen_rulefit.py
koniecsveta: `fit_rules` renamed to `eval_rules` based on slack discussion ab4985d

h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
```diff
@@ -173,4 +177,39 @@ public RuleFitMojoWriter getMojo() {
...
+    
+    @Override
+    public Frame fitRules(Frame frame, String[] ruleIds) {
+        Frame adaptFrm = new Frame(frame);
+        adaptTestForTrain(adaptFrm, true, false);
+        
+        List<Rule> rules = new ArrayList<>();
+        List<String> linearRules = new ArrayList<>();
+        for (int i = 0; i < ruleIds.length; i++) {
+            if (ruleIds[i].startsWith("linear.") && isLinearVar(ruleIds[i])) {
```
maurever, line 189: Is the first part of the condition necessary?

h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
```diff
@@ -173,4 +177,39 @@ public RuleFitMojoWriter getMojo() {
...
+            result.add(linearRules.get(i), Vec.makeOne(frame.numRows()));
+        }
+        
+        result = new Frame(Key.make(), result.names(), result.vecs());
+        DKV.put(result);
+        return result;
+    }
+    
+    private boolean isLinearVar(String potentialLinVarId) {
+        List<String> linVarNames = Arrays.asList(glmModel.names()).stream().filter(name -> name.startsWith("linear.")).collect(Collectors.toList());
```
maurever, line 208: IMO the list `linVarNames` could be initialized only once, or am I missing something?

h2o-algos/src/main/java/hex/rulefit/RuleFitUtils.java
```diff
@@ -131,4 +131,14 @@ static Rule consolidateRule(Rule rule, boolean remove_duplicates) {
 
         return transform.toArray(new Rule[0]);
     }
+
+    static String readRuleId(String ruleId) {
+        // if ruleId is in form after deduplication: "M0T0N1, M0T9N56, M9T34N56", meaning contains ", "
```
maurever, line 136: I would move the comment into the method docstring.

h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
koniecsveta: thats a good point, thanks!

h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
koniecsveta: the reason is: if I know that the rule id doesn't start with "linear." then it's not necessary to evaluate isLinearVar() method and loop inside of it. So on easy cases, first part of && is false and it doesn't even evaluate the second part. And most of the cases will be that ruleIds[i] not starting with "linear." and they will fall to else branch right after first part of the condition evaluated.

maurever: Oh, I see. Sure, make sense. :+1:

h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
koniecsveta: 8902392

h2o-algos/src/main/java/hex/rulefit/RuleFitUtils.java
koniecsveta: 8902392

h2o-algos/src/main/java/hex/rulefit/RuleFitModel.java
```diff
@@ -9,8 +9,12 @@
 import water.udf.CFuncRef;
 import water.util.TwoDimTable;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
 
-public class RuleFitModel extends Model<RuleFitModel, RuleFitModel.RuleFitParameters, RuleFitModel.RuleFitOutput> {
+public class RuleFitModel extends Model<RuleFitModel, RuleFitModel.RuleFitParameters, RuleFitModel.RuleFitOutput> implements FitRulesCollector {
```
maurever, line 17: One more note:

The R and Python API call the `predict_rules` method based on the discussion about naming, but in Java, you keep everywhere `FitRules`. Maybe it is only a detail, but I would keep the naming the same everywhere. 

WDYT @koniecsveta ?

koniecsveta: yesyes, will rename also in Java

koniecsveta: b615b24

michalkurka (approved):
Thank you!

PR #5898, opened 2021-11-18
PUBDEV-8425: Add option to mimic "ActiveProcessorCount" for older JVMs

michalkurka:


h2o-core/testMultiNode.sh
```diff
@@ -76,7 +76,8 @@ else
     COVERAGE=""
 fi
 # Command to invoke test.
-JVM="nice $JAVA_CMD $COVERAGE -Xmx${MAX_MEM} -Xms${MAX_MEM} -ea -cp ${JVM_CLASSPATH} -Dsys.ai.h2o.rapids.checkObjectConsistency=true ${ADDITIONAL_TEST_JVM_OPTS}"
+# note: we set sys.ai.h2o.activeProcessorCount=4 because multinode tests are running on the same instance - this helps avoid resource contention and speeds tests up  
+JVM="nice $JAVA_CMD $COVERAGE -Xmx${MAX_MEM} -Xms${MAX_MEM} -ea -cp ${JVM_CLASSPATH} -Dsys.ai.h2o.rapids.checkObjectConsistency=true -Dsys.ai.h2o.activeProcessorCount=4 ${ADDITIONAL_TEST_JVM_OPTS}"
```
mn-mikke, line 80: Shouldn't the number of gc threads be also set?

mn-mikke (commented):
What about [LinuxProcFileReader.getProcessCpusAllowed()](https://github.com/h2oai/h2o-3/blob/879e58d3e5f98064b738189febecd845bf4f87c4/h2o-core/src/main/java/water/util/LinuxProcFileReader.java#L118-L130)? Shouldn't it be also modified?

h2o-core/testMultiNode.sh
michalkurka: In theory, it is a good idea, however, it is currently really not needed as GC is not an issue in tests and it could cause issues with different GC implementations in different java versions that we use in tests (although currently both parallel gc and g1 would support it).

michalkurka:
> What about [LinuxProcFileReader.getProcessCpusAllowed()](https://github.com/h2oai/h2o-3/blob/879e58d3e5f98064b738189febecd845bf4f87c4/h2o-core/src/main/java/water/util/LinuxProcFileReader.java#L118-L130)? Shouldn't it be also modified?

This is on purpose, this method reads the information directly from the OS and bypasses java on Linux. This also means that specifying JVM option `-XX:ActiveProcessorCount` won't affect it (on Linux). It is used only for diagnostic purposes and doesn't affect parallelism in H2O as such.

It should be kept as-is but maybe I need to add a comment explaining why it is done this way.

mn-mikke (approved):
Thanks for answering my questions! LGTM

PR #5899, opened 2021-11-19
[nocheck] avoid use of deprecated imp standard library

sebhrusen:
I checked in the entire project: the `imp` module was used only in those `shared_utils` functions.

### DIFF
--- a/h2o-py/h2o/utils/shared_utils.py
+++ b/h2o-py/h2o/utils/shared_utils.py
@@ -66,7 +66,7 @@ def temp_ctr():
 
 
 def is_module_available(mod):
-    if mod in sys.modules:  # fast track + safer in unusual environments 
+    if mod in sys.modules and sys.modules[mod] is not None:  # fast track + safer in unusual environments 
         return True
     if PY2:
         import imp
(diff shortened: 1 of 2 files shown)


michalkurka (approved):
Thank you @sebhrusen

h2o-py/h2o/utils/shared_utils.py
```diff
@@ -66,36 +65,45 @@ def temp_ctr():
     return _id_ctr
 
 
+def is_module_available(mod):
+    if PY2:
```
sebhrusen, line 69:
```suggestion
    if mod in sys.modules:  # fast track + safer in unusual environments 
        return True
    if PY2:
```

h2o-py/h2o/utils/shared_utils.py
```diff
@@ -66,36 +65,47 @@ def temp_ctr():
     return _id_ctr
 
 
-def can_use_pandas():
-    try:
-        imp.find_module('pandas')
+def is_module_available(mod):
+    if mod in sys.modules:  # fast track + safer in unusual environments
```
sebhrusen, line 69:
```suggestion
    if mod in sys.modules and sys.modules[mod] is not None:  # fast track + safer in unusual environments 
```

PR #5900, opened 2021-11-19
PUBDEV-8439: consolidate rules during extraction

koniecsveta:
https://h2oai.atlassian.net/browse/PUBDEV-8439
this is second approach to PUBDEV- 8383:

- tree traversal during rule extraction has changed: it goes from node to root, not from root to node anymore
- it consolidates rules during extraction, before they are passed to GLM, not after GLM as a post-process anymore

### DIFF
--- a/h2o-algos/src/main/java/hex/rulefit/Rule.java
+++ b/h2o-algos/src/main/java/hex/rulefit/Rule.java
@@ -7,6 +7,7 @@ import water.Iced;
 import water.fvec.Chunk;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 public class Rule extends Iced {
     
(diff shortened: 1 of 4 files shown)


maurever (approved):
LGTM, thanks @koniecsveta.

h2o-algos/src/main/java/hex/rulefit/Rule.java
```diff
@@ -85,43 +86,57 @@ public boolean equals(Object obj) {
...
-        }
-        
-        if (node.isLeaf()) {
-            // create Rule
-            String varName = "M" + modelId + "T" + node.getSubgraphNumber() + "N" + node.getNodeNumber();
+        // filter leaves
+        List<SharedTreeNode> leaves = tree.nodesArray.stream().filter(sharedTreeNode -> sharedTreeNode.isLeaf()).collect(Collectors.toList());
+        // traverse paths
+        for (SharedTreeNode leaf : leaves) {
+            List<Condition> conditions = new ArrayList<>();
```
michalkurka, line 93: it would be a little clearer if there was a version fo the `traversePath` method that doesn't need the empty array on the input

```
void traversePath(SharedTreeNode node, Set<Rule> rules, String varName) {
   traversePath(node, new ArrayList<>(), rules, String varName)
}
```

michalkurka (approved):
Thank you, I think this is definitely a change for the better

h2o-algos/src/main/java/hex/rulefit/Rule.java
koniecsveta: thanks, abef4ad

PR #5901, opened 2021-11-22
PUBDEV-8416: Add stopping parameters to SE

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8416

Up until this commit, users could provide stopping parameters only using `metalearner_params`. This commit exposes them as StackedEnsemble's parameters. 

To make this change backward compatible the code first checks if the parameter has a default value in `metalearner_params` and if it does it will overwrite it using the value that is in Stacked Ensemble's parameters.

In other words, parameters set in `metalearner_params` have higher priority and will be used if specified. 

The problem with this approach is that it is not consistent with what we currently have in SE, e.g.,`weights_column` will always be used from SE's params and not from the `metalearner_params`. I can make the changes both ways to make SE parameter setting consistent. 

Any opinions @sebhrusen @ledell ?

### DIFF
--- a/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
+++ b/h2o-algos/src/main/java/hex/ensemble/Metalearner.java
@@ -10,6 +10,9 @@ import water.Key;
 import water.fvec.Frame;
 import water.util.Log;
 
+import java.lang.reflect.Field;
+import java.util.Objects;
+
 public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Model<M, P, ?>, P extends Model.Parameters> {
 
   /**
@@ -106,6 +109,19 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
 
   abstract B createBuilder();
 
+  protected <P1 extends Model.Parameters, P2 extends Model.Parameters> void setIfNotSet(P1 from, P2 to, String what) {
+    try {
+      P2 defaultParms = (P2) to.getClass().newInstance();
+      Field toField = to.getClass().getField(what);
+      Field fromField = from.getClass().getField(what);
+      if (Objects.equals(toField.get(to), toField.get(defaultParms))) {
+        toField.set(to, fromField.get(from));
+      }
+    } catch (Exception e){
+      // Ignore
+    }
+  }
+
   protected void setCommonParams(P parms) {
     if (parms._seed == -1) { //use _metalearnerSeed only as legacy fallback if not set on metalearner_parameters 
       parms._seed = _metalearnerSeed;
@@ -116,6 +132,12 @@ public abstract class Metalearner<B extends ModelBuilder<M, P, ?>, M extends Mod
     parms._max_runtime_secs = _maxRuntimeSecs;
     parms._weights_column = _model._parms._weights_column;
     parms._offset_column = _model._parms._offset_column;
+
+    // Checking if it was set in "metalearner_params" to keep backward compatibility
+    setIfNotSet(_model._parms, parms, "_stopping_metric");
+    setIfNotSet(_model._parms, parms, "_stopping_rounds");
+    setIfNotSet(_model._parms, parms, "_stopping_tolerance");
+    setIfNotSet(_model._parms, parms, "_score_each_iteration");
   }
 
   protected void setCrossValidationParams(P parms) {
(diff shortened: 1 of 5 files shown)


h2o-py/h2o/estimators/stackedensemble.py
```diff
@@ -155,10 +159,27 @@ def __init__(self,
                random number)
                Defaults to ``-1``.
         :type seed: int
+        :param score_each_iteration: Whether to score during each iteration of model training.
```
michalkurka, line 162: What about score_tree_interval for tree-based models?

h2o-algos/src/main/java/hex/ensemble/Metalearner.java
```diff
@@ -116,6 +132,12 @@ protected void setCommonParams(P parms) {
     parms._max_runtime_secs = _maxRuntimeSecs;
     parms._weights_column = _model._parms._weights_column;
     parms._offset_column = _model._parms._offset_column;
+
+    // Checking if it was set in "metalearner_params" to keep backward compatibility
```
michalkurka, line 136: instead of using reflection it would be better to introduce an API on model Parameters for setting early stopping criteria

in tree-based you have score_tree_interval, in GLM score_iteration_interval 

we could eg. add a method setScoreIterationInteval to Model.Parameters that would set the parameter if it is suppored

h2o-py/h2o/estimators/stackedensemble.py
tomasfryda: Thanks, I forgot about that one.

It makes me wonder if this PR is a good idea. We can already set the parameters via `metalearner_params` and exposing parameters that are applicable only on some meta learners on the SE level doesn't seem like a good idea to me.

sebhrusen: @tomasfryda I thought the ticket was only about `stopping_*` params that define the early stopping feature and that are supported by all metalearners I think.
I'm much less convinced about `score_each_iteration` or `score_tree_interval` that are not really necessary to use early stopping, just optional additions.

tomasfryda: Ok, I thought it should include even these as I think they influence the stopping - to be able to early stop we have to have scored iterations/trees, in other words, it defines the "resolution" used by the early stopping. At least this is how it worked in GLM last time I looked.

I will remove the `score_each_iteration`.

michalkurka: @sebhrusen Without those parameters, the early stopping feature is not reproducible and I would consider it problematic in general. If you don't know how often H2O scores then it is very difficult to correctly set stopping rounds.

sebhrusen: @michalkurka I get your point about reproducibility. My concern is purely semantic, and deals with API choices.

To me the problem is similar to the decision that was made long ago to expose only the `stopping_*` params in 
AutoML: I guess that only those ones have been exposed to AutoML because they are enough to define the early stopping feature. The others depend on each model trained internally. If the user wanted to specify the optional ones for each algo, now s·he could override those using `algo_parameters`. Not ideal, but otherwise we could as well expose any possible algo param as a first-class param.

For SE, I think it's similar: the `stopping_*` are enough to define the feature—a feature otherwise made available to all algos, except SE until now—, but if the user needs to fine tune, s·he will have to consider the specific metalearner used by the SE, and may as well set those params as `metalearner_params`.

sebhrusen: @ledell wdyt?

michalkurka: I think the difference between AutoML and SE is that AutoML by default has predictable scoring intervals and users can reason about it and use early stopping to their advantage.

I am thinking that these parameters are not needed on SE level because without understanding what metalearner is used and what parameters goes in the ability to set the early stopping parameters is limited anyway.

ledell: @sebhrusen @tomasfryda Let's chat on Slack to discuss in more detail and then return to the thread here...

sebhrusen: @michalkurka we agreed on not to expose the params at all, and as a result @tomasfryda will close this PR.
It was based on a misconception that some of the high-level params were already forwarded to the metalearner, but this is actually not the case: only data/input-related params are, not algo-related ones.

Thanks for being watchful on this.

ledell:
After a lengthy discussion, we decided it was cleaner to keep metalearner related params in the `metalearner_params` arg and not expose/duplicate them at the top level.  Instead we will modify the JIRA to tackle the actual issue which is that early stopping params from AutoML are not getting propagated down to the Stacked Ensemble (we will instead pass them through the `metalearner_params` arg directly).

PR #5902, opened 2021-11-22
[nocheck] Adding Py support until 3.10

sebhrusen:
Refactored groovy scripts + docker scripts + makefile in such a way that we can more clearly see which versions are really used and for which purpose.
Py 3.10 is not used it as I want to test all Jenkins builds first, but then it will just be matter of changing a couple of versions lists + build new images.


docker/Jenkinsfile-build-docker
```diff
@@ -20,9 +20,9 @@ properties([
 IMAGE_NAME_PREFIX = 'harbor.h2o.ai/opsh2oai/h2o-3'
 
 JDK_VERSIONS = ['8', '10', '11', '12', '13', '14', '15']
-JDK_VERSIONS_PY_R = ['8', '11', '14', '15'] // stable, last-supported, latest
-PYTHON_VERSIONS = ['2.7', '3.5', '3.6', '3.7', '3.8']
-R_VERSION = ['3.3.3', '3.4.1', '3.5.3', '3.6.2', '4.0.2']
+JDK_VERSIONS_PY_R = ['8', '11', '14', '15'] // first_LTS, latest_LTS, latest (see scripts/jenkins/groovy/buildConfig.groovy)
+PYTHON_VERSIONS = ['2.7', '3.5', '3.7', '3.8'] // legacy, first supported Py3, first active Py3, latest (see scripts/jenkins/groovy/buildConfig.groovy)
+R_VERSIONS = ['3.3.3', '3.4.4', '3.5.3', '3.6.3', '4.0.5', '4.1.2'] // do we really need to build an image for all those versions given that we test only one v3 + one v4?
```
sebhrusen, line 25: I feel like reducing the list of R versions to test: not sure it's worth testing all minor versions since 3.3 + there are now 2 minor 4.x versions available

michalkurka:
@sebhrusen this cannot be merged until new images are built, correct?

sebhrusen:
@michalkurka correct, not before images are built, and I'll probably change the latest Py or R versions if you agree.
Mainly putting this to review also to discuss this version testing strategy.

michalkurka:
ok, got it - would it be possible to apply the same refactoring/restructuring to the existing code base without changing the versions

and then update the versions in a new PR?

this way we can merge the changes eg. in "define test stages" first

@valenad1 can testify that building the docker images will probably be a non-trivial effort as we will need to resolve incompatibilities of packages

sebhrusen:
> would it be possible to apply the same refactoring/restructuring to the existing code base without changing the versions

@michalkurka yes, that's what I tried to do at the end, but I messed up with R versions. For Java+Py it should behave as before though. Will fix the R versions

valenad1:
> @valenad1 can testify that building the docker images will probably be a non-trivial effort as we will need to resolve incompatibilities of packages

Agree, all the changes that can be split from the image build would be better to merged separately.

PR #5906, opened 2021-11-23
[nocheck] fix too lenient logic for Py test utility modules loading

sebhrusen:
tested locally with fake `_foo/bar.py` and `_foo/_bar.py` folder structures.


michalkurka:
looks like this did the trick - thank you @sebhrusen

PR #5845, opened 2021-10-22
PUBDEV-8381: Add Gains/Lift table to Performance & Prediction

hannah-tillman:
For: [PUBDEV-8381](https://h2oai.atlassian.net/browse/PUBDEV-8381)

First draft of gains/lift added to performance and prediction. It's an amalgam of information from gains/lift in the flow documentation and that wonderful doc @wendycwong curated. 

Please let me know what needs to be changed/clarified/updated!

PR #5907, opened 2021-11-24
PUBDEV-8388: Add a plot function for gains/lift in R and Python

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8388

@ledell After looking at https://github.com/h2oai/h2o-3/pull/5845, I still have several API related questions:
* Should the user be able to plot just one of the curves? If so what should be the API, `type = "both" | "gains" | "lift"` or `plot_gains = TRUE`. I am also wondering about the usage of "gains" since in the plot we are using "cummulative capture rate".
* The name `h2o.plot_gainslift` looks inconsistent next to the `h2o.gainsLift`. Should we deprecate the `h2o.gainsLift` and add `h2o.gainslift`?

The plots are basically the same as in JIRA but Y axis doesn't have upper bound since it cutoff some values and I added grid and legend to R. This can be seen here:
glr

### DIFF
--- a/h2o-py/h2o/model/binomial.py
+++ b/h2o-py/h2o/model/binomial.py
@@ -6,6 +6,7 @@ from h2o.exceptions import H2OValueError
 from h2o.utils.typechecks import assert_is_type
 from .extensions import has_extension
 from .model_base import ModelBase
+from ..utils.metaclass import deprecated_params
 
 
 class H2OBinomialModel(ModelBase):
@@ -851,6 +852,20 @@ class H2OBinomialModel(ModelBase):
         >>> gbm.gains_lift(train=True, valid=True, xval=True)
         """
         return self._delegate_to_metrics('gains_lift', train, valid, xval)
+
+    @deprecated_params({'save_to_file': 'save_plot_path'})
+    def plot_gainslift(self, xval=False, server=False, save_plot_path=None, plot=True, **kwargs):
+        """
+        Plot Gains/Lift curves.
+
+        :param xval: if True, use cross-validation metrics
+        :param server: if True, generate plot inline using matplotlib's "Agg" backend.
+        :param save_plot_path: filename to save the plot to
+        :param plot: True to plot curve, False to get a gains lift table
+
+        :returns: Gains lift table + the resulting plot (can be accessed using result.figure())
+        """
+        return self._delegate_to_metrics('plot_gainslift', xval=xval, server=server, save_plot_path=save_plot_path, plot=plot)
 
     def kolmogorov_smirnov(self):
         """
--- a/h2o-py/h2o/model/binomial.py
+++ b/h2o-py/h2o/model/binomial.py
@@ -854,10 +854,11 @@ class H2OBinomialModel(ModelBase):
         return self._delegate_to_metrics('gains_lift', train, valid, xval)
 
     @deprecated_params({'save_to_file': 'save_plot_path'})
-    def plot_gainslift(self, xval=False, server=False, save_plot_path=None, plot=True, **kwargs):
+    def plot_gains_lift(self, type="both", xval=False, server=False, save_plot_path=None, plot=True, **kwargs):
         """
         Plot Gains/Lift curves.
 
+        :param type: one of "both", "gains", "lift". Defaults to "both".
         :param xval: if True, use cross-validation metrics
         :param server: if True, generate plot inline using matplotlib's "Agg" backend.
         :param save_plot_path: filename to save the plot to
@@ -865,7 +866,7 @@ class H2OBinomialModel(ModelBase):
 
         :returns: Gains lift table + the resulting plot (can be accessed using result.figure())
         """
-        return self._delegate_to_metrics('plot_gainslift', xval=xval, server=server, save_plot_path=save_plot_path, plot=plot)
+        return self._delegate_to_metrics('plot_gains_lift', type=type, xval=xval, server=server, save_plot_path=save_plot_path, plot=plot)
 
     def kolmogorov_smirnov(self):
         """
--- a/h2o-py/h2o/model/binomial.py
+++ b/h2o-py/h2o/model/binomial.py
@@ -854,7 +854,7 @@ class H2OBinomialModel(ModelBase):
         return self._delegate_to_metrics('gains_lift', train, valid, xval)
 
     @deprecated_params({'save_to_file': 'save_plot_path'})
-    def plot_gains_lift(self, type="both", xval=False, server=False, save_plot_path=None, plot=True, **kwargs):
+    def plot_gains_lift(self, type="both", xval=False, server=False, save_plot_path=None, plot=True):
         """
         Plot Gains/Lift curves.
 
--- a/h2o-py/h2o/model/binomial.py
+++ b/h2o-py/h2o/model/binomial.py
@@ -854,7 +854,7 @@ class H2OBinomialModel(ModelBase):
         return self._delegate_to_metrics('gains_lift', train, valid, xval)
 
     @deprecated_params({'save_to_file': 'save_plot_path'})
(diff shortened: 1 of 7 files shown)


ledell:
@tomasfryda Good questions:

> * Should the user be able to plot just one of the curves? If so what should be the API, `type = "both" | "gains" | "lift"` or `plot_gains = TRUE`. I am also wondering about the usage of "gains" since in the plot we are using "cummulative capture rate".

Yeah, it seems like the single plot versions are widely used as well.  I am ok with `type = "both" | "gains" | "lift"`.  We use `plot_type` as an [arg](https://github.com/h2oai/h2o-3/blob/45c2cbcff332ebf178c60f1293169f948b3bc405/h2o-py/h2o/explanation/_explain.py#L685) elsewhere to mean something else (actual plot type)... so maybe the arg should be named something like, `display` or `render` or similar?  (e.g. `display = "both"` or `display = "gains_and_lift"`) Or we just keep it simple with `type`.

> * The name `h2o.plot_gainslift` looks inconsistent next to the `h2o.gainsLift`. Should we deprecate the `h2o.gainsLift` and add `h2o.gainslift`?

Agreed.  We are slowly phasing out camelCase in leu of snake_case in all function names, one alias at a time, so that R and Python can be consistent and also to adhere to the Tidyverse R style guide.  So I think it would be ok to make an `h2o.gainslift()` alias and lightly deprecate `h2o.gainsLift()` (phase it out of documentation but still support it indefinitely).

h2o-py/h2o/model/binomial.py
```diff
@@ -852,6 +853,21 @@ def gains_lift(self, train=False, valid=False, xval=False):
         """
         return self._delegate_to_metrics('gains_lift', train, valid, xval)
 
+    @deprecated_params({'save_to_file': 'save_plot_path'})
+    def plot_gains_lift(self, type="both", xval=False, server=False, save_plot_path=None, plot=True, **kwargs):
```
sebhrusen, line 857: what are the `kwargs` for? they're not forwarded

h2o-py/h2o/model/metrics_base.py
```diff
@@ -1428,11 +1428,13 @@ def plot(self, type="roc", server=False, save_plot_path=None, plot=True, **kwarg
         >>> cars_gbm.plot(type="pr")
         
         """
-
+        assert type in ["roc", "pr", "gains_lift"]
```
sebhrusen, line 1431: the `plot` method also has unused `kwargs` param!
I don't get it
We allow user to pass whatever without validation…

h2o-py/h2o/model/metrics_base.py
```diff
@@ -1766,6 +1768,55 @@ def gains_lift(self):
             return self._metric_json['gains_lift_table']
         return None
 
+    @deprecated_params({'save_to_file': 'save_plot_path'})
+    def plot_gains_lift(self, type="both", server=False, save_plot_path=None, plot=True, **kwargs):
```
sebhrusen, line 1772: guess what…
I wouldn't mind if:
1. `kwargs` was then passed to matplotlib somehow to customize plots: shouldn't be necessary though now that we return the figure.
2. it was documented.

h2o-r/h2o-package/pkgdown/_pkgdown.yml
```diff
@@ -122,6 +122,8 @@ reference:
       - h2o.floor
       - h2o.flow
       - h2o.gainsLift
+      - h2o.gains_lift
+      - h2o.gains_lift_plot
```
sebhrusen, line 126: I see that you introduced `h2o.plot_gains_lift` but don't see any `h2o.gains_lift_plot`.
Also, having both would be very confusing to say the least.

h2o-r/h2o-package/R/models.R
```diff
@@ -3580,6 +3581,10 @@ h2o.null_dof <- function(object, train=FALSE, valid=FALSE, xval=FALSE) {
 #' @export
 setGeneric("h2o.gainsLift", function(object, ...) {})
 
+#' @rdname h2o.gainsLift
+#' @export
+h2o.gains_lift <- function(object, ...) h2o.gainsLift(object, ...)
```
sebhrusen, line 3586: same as Py, I don't understand why we went for dots param in all those plot functions here: there's no validation, and the only reason for their existence seems to be able to pass `xval` when applied to a model object.
In return, nowhere do we seem to validate those dots. Either we allow explicit `xval` param everywhere, emitting a warning if passed non-default value to an object that doesn't support it, or we use dots but at the end validate those, checking that each method received only allowed params (empty dots if metrics, only xval if model).

sebhrusen (commented):
some concerns about Py `kwargs` and R dots.
Is it something that was previously discussed with @ledell ?

h2o-py/h2o/model/binomial.py
tomasfryda: Thank you for noticing that. I must have copied the `kwargs` from some other method that used the `@deprecated_params` while assuming that the `kwargs` are used for the deprecated parameter which was a wrong assumption...

h2o-r/h2o-package/pkgdown/_pkgdown.yml
tomasfryda: Thank you for catching that. We renamed the function and I forgot to remove the old name from the `pkgdown`.

h2o-r/h2o-package/R/models.R
tomasfryda: Here I use the `...` just to have the same interface as the generic method that this function should alias.
Rather than fixing this in this PR I think we should instead do it in separate PR (if we decide to) as this pattern (not checking for extra args passed in the dots) is used throughout our R package.

What I did now was to ensure that we pass the dots to all "recursive" calls of `h2o.gainsLift` which makes it eventually fail as the check for the undeclared argument within the dots is checked implicitly by looking at the method signature, e.g., `h2o.gainsLift(model, unused_param)` will call `h2o.gainsLift(model_metrics, unused_param)` and then warns about the unused argument since the signature for `h2o.gainsLift` for model_metrics is just `function(object)`...

h2o-r/h2o-package/pkgdown/_pkgdown.yml
```diff
@@ -244,6 +245,7 @@ reference:
       - h2o.permutation_importance
       - h2o.permutation_importance_plot
       - h2o.pivot
+      - h2o.plot_gains_lift
```
sebhrusen, line 248: I don't like the fact that we changed the naming convention here.
Until now:
```
     - h2o.permutation_importance
     - h2o.permutation_importance_plot
```
(same in Py: `model.varimp` vs. `model.varimp_plot`, `model.std_coef` vs. `model.std_coef_plot` …)
With this PR:
```
- h2o.gains_lift
- h2o.plot_gains_lift
```

We need to be consistent. And actually I prefer the legacy style (easily discovered during autocomplete when first trying to get the `gains_lift`)

sebhrusen (changes requested):
@tomasfryda I'd consider renaming the new methods in Py and R for the reason mentioned below.
cc: @ledell

h2o-r/h2o-package/pkgdown/_pkgdown.yml
tomasfryda: Makes sense, I made the appropriate changes (`plot_gains_lift -> gains_lift_plot`).

ledell: Agreed @sebhrusen, thanks for the name adjustment @tomasfryda.

ledell (approved):
This is good, thank you very much @tomasfryda!

sebhrusen (approved):
Looks good, thx @tomasfryda

PR #5908, opened 2021-11-24
PUBDEV-8427 add maxrglm coeffs function

wendycwong:
This PR completes the task in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8427

I have completed the following functions:
1.  add to coef(), coef_norm() in python to return coefficients and normalized coefficients for maxrglm;
2. add to h2o.coef(), h2o.coef_norm() in R  to return coefficients and normalized coefficients for maxrglm;
3. beta(), getNormBeta(), coefficients(), coefficients(boolean standardize) in Java.

Unit tests in R, Python and Java are added to check correct implementation.

### DIFF
--- a/h2o-bindings/bin/custom/python/gen_maxrglm.py
+++ b/h2o-bindings/bin/custom/python/gen_maxrglm.py
@@ -1,4 +1,66 @@
 def class_extensions():
+    def coef_norm(self, predictor_size=None):
+        """
+        Get the normalized coefficients for all models built with different number of predictors.
+        
+        :param self:
+        :param predictor_size: predictor subset size, will only return model coefficients of that subset size.
+        :return: list of Python Dicts of coefficients for all models built with different predictor numbers
+        """
+        model_ids = self._model_json["output"]["best_model_ids"]
+        if model_ids is None:
+            return None
+        else:
+            model_numbers = len(model_ids)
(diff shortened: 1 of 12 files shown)


h2o-bindings/bin/custom/python/gen_maxrglm.py
```diff
@@ -1,4 +1,44 @@
 def class_extensions():
+    def coef(self):
```
neemah2o, line 2: I think one thing that users might see is that MaxRGLM creates multiple models while GLM makes only one. So they will get a list of coefficients for each model. If they build a pipeline, they will need to be aware of this and know which model/index to select

wendycwong: @neemah2o This is correct and they should expect that because they are building multiple models here.  However, the indices are pretty generic.  If they are looking for best one predictor model, it will be index = 0, for best two predictors model, index=1, and so on and so forth.  I will check and make sure the documentation mentions that.

wendycwong: I think I will add an index allowing them to choose the coefficients of a particular size.  Thanks for the feedback.

maurever (approved):
LGTM, thanks @wendycwong.

PR #5912, opened 2021-11-26
Uprade version of HDP 3.1.1 dependencies [nocheck]

valenad1:


valenad1:
Try branch with different name and newer version of hadoop.

PR #5915, opened 2021-11-29
PUBDEV-8437: Updating MAXR documentation

hannah-tillman:
For: [PUBDEV-8437](https://h2oai.atlassian.net/browse/PUBDEV-8437)

Please let me know if anything is inaccurate or needs further reworking.


michalkurka:
@hannah-tillman thank you for the clean-up, the removed text doesn't describe the procedure that was actually implemented as is thus confusing.

michalkurka (approved):

PR #5916, opened 2021-11-30
Make JavaVersionSupportTest independent from external debug.allowJava versions settings [nocheck]

valenad1:
I found that when I set debug.allowJavaVersions to 16 externally. Then I get failed test in 
`assertFalse(JavaVersionSupport.runningOnSupportedVersion());`

I made this test independent from external set of this property.

Second option would be to change:

```
        @Parameterized.Parameters
        public static Collection<Object[]> data() {
            return Arrays.asList(new Object[][]{
                    {7}, {16}
            });
        }
```

to `{7}, {256}` and avoid conflict between temporally unsupported java version.

### DIFF
--- a/h2o-core/src/test/java/water/JavaVersionSupportTest.java
+++ b/h2o-core/src/test/java/water/JavaVersionSupportTest.java
@@ -1,5 +1,6 @@
 package water;
 
+import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


michalkurka:
thanks, good catch - I think this test doesn't apply in this case - so we should perhaps disable it with Assume.assumeTrue(System.getProperty(H2O.OptArgs.SYSTEM_PROP_PREFIX + "debug.allowJavaVersions") == null)

valenad1:
Thank you for the tip! I change base branch to the rel-zizler

michalkurka (approved):
Thank you @valenad1

PR #5917, opened 2021-11-30
Explicitly state that VarImp is not supported in Uplift [nocheck]

michalkurka:


maurever (approved):
Thanks, @michalkurka, looks good to me. Failed tests look unrelated.

PR #5920, opened 2021-12-01
PUBDEV-8445 fix pvalue tweedie failure

wendycwong:
This PR fixes the issue in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8445

runit_pubdev_2441_pvalue_w_offset_tweedie_R sometimes failed due to the generation of constant columns as the response column.  The generated dataset is very small 10 rows by 4 columns, hence the change of getting constant column is very real.  I fixed this by using one fixed randomly generated dataset instead.


koniecsveta (approved):

valenad1 (approved):
LGTM, thanks!

PR #5921, opened 2021-12-01
PUBDEV-8433 glrm model id error

wendycwong:
This PR solves the issue in JIRA: https://h2oai.atlassian.net/browse/PUBDEV-8433.

When the user sets model_id and representation_name to the same string, there is a collision of model and frame using the same key.  To avoid this, we generate a warning if it happens.

### DIFF
--- a/h2o-algos/src/main/java/hex/glrm/GLRM.java
+++ b/h2o-algos/src/main/java/hex/glrm/GLRM.java
@@ -144,7 +144,7 @@ public class GLRM extends ModelBuilder<GLRMModel, GLRMModel.GLRMParameters, GLRM
       error("_recover_svd", "_recover_svd and _impute_original cannot both be true if _train" +
               " is transformed");
 
-    if (_train == null) return;
+    if (null == _train) return;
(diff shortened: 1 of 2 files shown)


us8945 (approved):

h2o-algos/src/main/java/hex/glrm/GLRM.java
```diff
@@ -159,13 +159,23 @@ public GLRM(boolean startup_once) {
...
+    
     if ((_parms._representation_name != null) && (_parms._loading_name != null)) {
       if (!(_parms._representation_name.equals(_parms._loading_name)))
         warn("_representation_name and _loading_name", "Are not equal.  _representation_name will" +
                 " be used since _loading_name is deprecated.");
     }
     
+    if (_parms._representation_name != null) {
+      if (dest().toString().equals(_parms._representation_name))
+        error("representation_name", "representation_name and model_id cannot use the same string.");
```
koniecsveta, line 171: would do in reversed order to be NPE safe

koniecsveta (approved):

PR #5924, opened 2021-12-02
PUBDEV-8447: Avoid using Random Stream API - creates differences between Java versionsions

michalkurka:
Java 17 brings changes in java Random implementation: https://nipafx.dev/java-random-generator/ - the Stream API changed in a way that produces different results on Java 8 and Java 17.

The goal of H2O's random is to be independent on the Java version - we provide our own implementation, however, we are not immune to changes in Stream API.

We will disable Stream API in this task and rewrite affected pieces of code to use non-Stream alternatives.


valenad1:
> Tests needed

Regarding IsolationTree gaussian.. We already have a test for ghost clusters and I'll run the anomaly detection performance test that I have for EIF on CreditCards

michalkurka:
@valenad1 @tomasfryda  please review

@tomasfryda please pay attention to fix https://github.com/h2oai/h2o-3/pull/5924/files#diff-987222eb4d0ae65eec2c3128c442c4b4c8b8bca725906f7efaa94aac1add51abR114

I changed the behavior when the weights column is provided - previously we could have used even 0-weighted rows to calculate varimp

valenad1 (approved):
Thank you!

PR #5929, opened 2021-12-03
PUBDEV-8448: Permutation varimp doesn't correctly work with weights

michalkurka:

PR #5930, opened 2021-12-03
Fix incorrect use of indexOf in FriedmanPopescusH [nocheck]

michalkurka:
This PR removes method ArrayUtils#indexOf - this method was responsible for 2 bugs:

- https://github.com/h2oai/h2o-3/pull/5929/files#diff-97ab26585a59e421d722a6ddd915c83f9253545352911b44b43e210a0b459590L405
- and the one in H-statistic in this PR

The issue is that the general contract of `indexOf` is to not have any assumptions about the input being sorted.
However, our version does require a sorted method and hence creates confusion and when used on unsorted
data produces bugs (like the 2 above).

In this PR I completely removed the method in favor of `Arrays.binarySearch` and in H-statistic ArrayUtils#find (the input was not sorted)


wendycwong (approved):
Thank you!

PR #5933, opened 2021-12-04
PUBDEV-8222: Infogram R API + plot & updates to Python API + plot

ledell:
R API: https://h2oai.atlassian.net/browse/PUBDEV-8222
- Adds the `plot.H2OInfogram()` function to plot the infogram in R
- Updated documentation for `h2o.infogram()`
- Fixes all bugs in CRAN check

Python API updates:
- Code documentation changes in R & Python
- Changes to Python API (plot, method names, cmi/rel key name)

This PR completes the R/Python API and plotting for Infogram: https://h2oai.atlassian.net/browse/PUBDEV-8075


h2o-r/h2o-package/R/infogram.R
```diff
@@ -0,0 +1,526 @@
...
+                 "ig_x",
+                 "ig_y",
+                 "raw")
+  ggplot2::ggplot(data = df, ggplot2::aes_(~ig_x, ~ig_y)) +
+    ggplot2::geom_point() +
+    ggplot2::geom_polygon(ggplot2::aes(x, y), data = data.frame(
+      x = c(xthresh, xthresh, -Inf, -Inf, Inf, Inf, xthresh),
+      y = c(ythresh, Inf, Inf, -Inf, -Inf, ythresh, ythresh)
+    ), alpha = 0.1, fill = "#CC663E") +
+    ggplot2::geom_path(ggplot2::aes(x, y), data = data.frame(
```
tomasfryda, line 512: same here

ledell: @tomasfryda Can we re-use the same `.data` object here, which was already set to `NULL` and then used in the polygon above?  Or do I need to declare a second `.data2 = NULL`?

h2o-r/h2o-package/R/infogram.R
```diff
@@ -0,0 +1,526 @@
...
+  # use generic names for x, y for easier ggplot code
+  names(df) <- c("column",
+                 "admissible",
+                 "admissible_index",
+                 "ig_x",
+                 "ig_y",
+                 "raw")
+  ggplot2::ggplot(data = df, ggplot2::aes_(~ig_x, ~ig_y)) +
+    ggplot2::geom_point() +
+    ggplot2::geom_polygon(ggplot2::aes(x, y), data = data.frame(
```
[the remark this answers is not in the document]
ledell: Hey @tomasfryda I tried the edit you suggested above, but I am getting an error, after making the following changes:

Screen Shot 2021-12-14 at 2 21 16 PM

```r
> plot(ig)
Error: geom_path requires the following missing aesthetics: x and y
Run `rlang::last_error()` to see where the error occurred.
```
I didn't see any errors when I ran `R CMD check --as-cran` on the original version, but it sounds like you might have some insight on this from working on the explain plot functions?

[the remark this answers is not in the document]
tomasfryda: Sorry, I tried to run the CRAN check as well and it doesn't produce an error but when I renamed the x and y to `x_points` and `y_points` I get
```
* checking R code for possible problems ... NOTE
plot.H2OInfogram: no visible binding for global variable ‘x_points’
plot.H2OInfogram: no visible binding for global variable ‘y_points’
Undefined global functions or variables:
  x_points y_points
```

So somewhere in the h2o code, there is `x` and `y` variable that is accessible from within the plotting function which is why there was no error.

The issue you had is probably caused by `.data2` - in tidyverse the `.data` is a "special" variable and `ggplot2` doesn't actually evaluate it as it is just used to do nicer "quasiquotation". The `aes` tries to be smart and user friendly for interactive use but when programming with it it gets a bit more complicated so the `.data` makes it easier to distinguish if you want to specify a column A (`.data$A`) or a column which name is specified in variable B (`.data[[B]]`).

So the change should be 
```diff
--- a/h2o-bindings/bin/custom/R/gen_infogram.py
+++ b/h2o-bindings/bin/custom/R/gen_infogram.py
@@ -87,6 +87,7 @@ h2o.get_admissible_attributes <- function(model) {
 #' @export
 plot.H2OInfogram <- function(x, ...) {
   .check_for_ggplot2() # from explain.R
+  .data <- NULL
   varargs <- list(...)
   if ("title" %in% names(varargs)) {
     title <- varargs$title
@@ -116,13 +117,13 @@ plot.H2OInfogram <- function(x, ...) {
                  "raw")
   ggplot2::ggplot(data = df, ggplot2::aes_(~ig_x, ~ig_y)) +
     ggplot2::geom_point() +
-    ggplot2::geom_polygon(ggplot2::aes(x, y), data = data.frame(
-      x = c(xthresh, xthresh, -Inf, -Inf, Inf, Inf, xthresh),
-      y = c(ythresh, Inf, Inf, -Inf, -Inf, ythresh, ythresh)
+    ggplot2::geom_polygon(ggplot2::aes(.data$x_points, .data$y_points), data = data.frame(
+      x_points = c(xthresh, xthresh, -Inf, -Inf, Inf, Inf, xthresh),
+      y_points = c(ythresh, Inf, Inf, -Inf, -Inf, ythresh, ythresh)
     ), alpha = 0.1, fill = "#CC663E") +
-    ggplot2::geom_path(ggplot2::aes(x, y), data = data.frame(
-      x = c(xthresh, xthresh, NA, xthresh, Inf),
-      y = c(ythresh,     Inf, NA, ythresh, ythresh)
+    ggplot2::geom_path(ggplot2::aes(.data$x_points, .data$y_points), data = data.frame(
+      x_points = c(xthresh, xthresh, NA, xthresh, Inf),
+      y_points = c(ythresh,     Inf, NA, ythresh, ythresh)
     ), color = "red", linetype = "dashed") +
     ggplot2::geom_text(ggplot2::aes_(~ig_x, ~ig_y, label = ~column),
                        data = df[as.logical(df$admissible),], nudge_y = -0.0325,
```

h2o-r/h2o-package/R/infogram.R
```diff
@@ -0,0 +1,526 @@
...
+  if (!missing(data_fraction))
+    parms$data_fraction <- data_fraction
+  if (!missing(top_n_features))
+    parms$top_n_features <- top_n_features
+  if (!missing(compute_p_values))
+    parms$compute_p_values <- compute_p_values
+
+  if (!missing(infogram_algorithm_params))
+      parms$infogram_algorithm_params <- as.character(toJSON(infogram_algorithm_params, pretty = TRUE))
+  h2o.show_progress() # enable progress bar explicitly
```
tomasfryda, line 265: I might be wrong but if the user sets `h2o.no_progress()` then this will not only do something that the user didn't want but also it will influence the rest of the program.

In explain, I use the following to temporary hide the progress bar, if we really want to ignore users wishes I'd suggest to modify the following snippet and use it so we don't change the behavior for the rest of the program.
```r
with_no_h2o_progress <- function(expr) {
  show_progress <- environment(h2o.no_progress)$
    .pkg.env$
    PROGRESS_BAR
  if (length(show_progress) == 0L || show_progress) {
    on.exit(h2o.show_progress())
  } else {
    on.exit(h2o.no_progress())
  }
  h2o.no_progress()
  force(expr)
}
```

h2o-r/tests/testdir_algos/admissibleml/runit_PUBDEV_8075_core_infogram_erin_fail1_large.R
```diff
@@ -0,0 +1,79 @@
...
+           "balloon_payment_desc",
+           "property_value",
+           "income",
+           "debt_to_income_ratio",
+           "applicant_credit_score_type_desc",
+           "co_applicant_credit_score_type_desc")
+    ig2 <- h2o.infogram(y = y, x = x2, 
+                       training_frame = train,
+                       protected_columns = protected_columns)
+    ig2
```
tomasfryda, line 75: This can deleted.

h2o-r/h2o-package/R/infogram.R
tomasfryda: The `.data` is meant to be reused. The `.data <- NULL` is there just so that CRAN doesn't complain about uninitialized variable.

h2o-r/h2o-package/R/infogram.R
```diff
@@ -0,0 +1,526 @@
...
+      instead.   Using default of 0.1 if not set")
+    }
+  } else { # fair infogram
+    if (!missing(net_information_threshold)) {
+    warning("Should not set net_information_threshold for fair infogram runs, set safety_index_threshold instead.  
+      Using default of 0.1 if not set")
+    }
+    if (!missing(total_information_threshold)) {
+      warning("Should not set total_information_threshold for fair infogram runs, set relevance_index_threshold
+       instead.    Using default of 0.1 if not set")
```
tomasfryda, line 188: The warnings contain probably unwanted whitespaces leading to suboptimally formatted warnings, e.g.,
```
Warning message:
Should not set total_information_threshold for fair infogram runs, set relevance_index_threshold
        instead.    Using default of 0.1 if not set 
```
or 
```
Warning message:
Should not set net_information_threshold for fair infogram runs, set safety_index_threshold instead.  
       Using default of 0.1 if not set 
```

h2o-r/tests/testdir_algos/admissibleml/runit_PUBDEV_8258_core_infogram_breast_cancer_infogram_params_cv_valid.R
```diff
@@ -0,0 +1,47 @@
...
+    compareFrames( relCMITrainCV, relCMITrainVCV, prob=1.0)
+    compareFrames( relCMITrain, relCMITrainVCV, prob=1.0)
+    
+    Log.info("comparing infogram info from validation dataset")
+    relCMIValidV <- infogramModelV@admissible_score_valid
+    relCMIValidVCV <- infogramModelVCV@admissible_score_valid
+    compareFrames(relCMIValidV, relCMIValidVCV, prob=1.0)
+    
+    Log.info("comparing infogram info from cross-validation hold out")
+    browser()
```
tomasfryda, line 41: `browser()` can be deleted.

h2o-bindings/bin/custom/python/gen_infogram.py
```diff
@@ -0,0 +1,323 @@
...
+    if name == 'infogram_algorithm_params':
+        param['type'] = 'KeyValue'
+        param['default_value'] = None
+        return param
+    return None  # param untouched
+
+def class_extensions():
+    def plot(self, train=True, valid=False, xval=False, figsize=(10, 10), title="Infogram", legend_on=True, server=False):
+        """
+        Perform plot function of infogram.  This code is given to us by Tomas Fryda.  By default, it will plot the
```
tomasfryda, line 13: No need to mention me here - it doesn't help the user.

h2o-bindings/bin/custom/python/gen_infogram.py
```diff
@@ -0,0 +1,323 @@
...
+        plt.ylabel(y_label)
+        plt.title(title)
+        fig = plt.gcf()
+        if not server:
+            plt.show()
+        return fig
+        
+    def get_relevance_cmi_frame(self, valid=False, xval=False):
+        """
+        Retreive relevance, cmi information in an H2O frame for training dataset by default
```
tomasfryda, line 117: In the docstrings, I would be a little more verbose - if there is some abbreviation I think we could use the unabbreviated form at least on the first usage, e.g., instead of `cmi` I'd use `conditional mutual information (cmi)`. (Assuming it is the conditional mutual information). 

I'm not sure if `cmi` is well-known abbreviation but to me as a non-native English speaker this is unknown and I think this could save some time for the users.

h2o-bindings/bin/custom/python/gen_infogram.py
```diff
@@ -0,0 +1,323 @@
...
+
+doc = dict(
+    __class__="""
+Given a sensitive/unfair predictors list, Infogram will add all predictors that contains information on the 
+ sensitive/unfair predictors list to the sensitive/unfair predictors list.  It will return a set of predictors that
+ do not contain information on the sensitive/unfair list and hence user can build a fair model.  If no sensitive/unfair
+ predictor list is given, Infogram will return a list of core predictors that should be used to build a final model.
+ Infogram can significantly cut down the number of predictors needed to build a model and hence will build a simple
+ model that is more interpretable, less susceptible to overfitting, runs faster while providing similar accuracy
+ as models built using all attributes.
```
tomasfryda, line 321: Second to last line have unnecessary indentation (could cause some formatting issue in generated docs).

h2o-py/h2o/estimators/infogram.py
```diff
@@ -0,0 +1,1037 @@
...
+            else:
+                self._parms["net_information_threshold"]=net_information_threshold
+
+    @property
+    def total_information_threshold(self):
+        """
+        relevance threshold for core infogram between 0 and 1 that is used to decide whether a predictor's relevance
+        level is high enough to be chosen into the admissible feature set.  If not set, will default to 0.1.
+
+        Type: ``float``, defaults to ``4.940656458e-324``.
```
tomasfryda, line 724: I don't think it is a good idea to have some complicated magic float number as a default especially if we say the default is 0.1 and on the next line we say it's ~5e-324.

In cases where the value is supposed to be positive (or even limited between 0 and 1), I'd use either the default (here 0.1) or -1  since we use it already as the "AUTO" value.

h2o-r/h2o-package/R/infogram.R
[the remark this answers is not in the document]
ledell: Thanks @tomasfryda I will make the update.

h2o-r/h2o-package/R/infogram.R
ledell: I guess I am not sure why we have `h2o.show_progress()` here because it seems like other algorithms don't have it.  @wendycwong did you add this for a particular reason, or was it just copy/pasted from some other algo skeleton code?   I remember I reported some issues with the progress bar a while ago when the R API was first created (I think it wasn't working) so maybe this was added to address that?

h2o-r/tests/testdir_algos/admissibleml/runit_PUBDEV_8075_core_infogram_erin_fail1_large.R
ledell: It prints the object (a list of some data).  I think it's fine to delete it, unless @wendycwong is trying to print the object in the test script so it's easier to debug (in logs after) when it may fail?

h2o-r/h2o-package/R/infogram.R
ledell: Good catch.  Will fix.

h2o-r/tests/testdir_algos/admissibleml/runit_PUBDEV_8258_core_infogram_breast_cancer_infogram_params_cv_valid.R
ledell: Thanks, not sure why that's there.

h2o-bindings/bin/custom/python/gen_infogram.py
ledell: Yeah, the paper used to use "CMI",  but the terminology has changed since the early versions of the paper.  However, the backend Java code has not been updated and uses "CMI", so I am wondering if these client-side utility functions should be updated as well... Maybe something like `get_admissibile_information`?  or `get_admissible_score()` since the data frame object is called `admissible_score`?  @wendycwong

h2o-py/h2o/estimators/infogram.py
ledell: @tomasfryda This is a known issue. I reported it to @wendycwong but it has not been fixed yet (she's working through higher priority items first).  I am not sure why it's being inserted into the clients as the default (when it's not the correct value).  Does anyone know where (I think in Java) this is coming from?  If we can find the source, we can fix it... I was not able to figure out why/where it's being added (in both R and Python client).

h2o-r/h2o-package/R/infogram.R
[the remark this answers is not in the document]
ledell: Pushed a fix.

h2o-r/h2o-package/R/infogram.R
ledell: Pushed a fix.

h2o-r/tests/testdir_algos/admissibleml/runit_PUBDEV_8258_core_infogram_breast_cancer_infogram_params_cv_valid.R
ledell: @wendycwong I noticed that we have the `browser()` command in a [bunch of tests](https://github.com/h2oai/h2o-3/search?l=R&q=browser%28%29).  Do you know what the purpose of this is?  That should open a web browser, so I am not sure what it's function inside the tests is.  I will wait for your reply before removing it...

h2o-bindings/bin/custom/python/gen_infogram.py
ledell: Fixed.

h2o-r/h2o-package/R/infogram.R
ledell: Fixed.

h2o-r/tests/testdir_algos/admissibleml/runit_PUBDEV_8258_core_infogram_breast_cancer_infogram_params_cv_valid.R
ledell: @wendycwong I just saw that you removed `browser()` on this commit, so you can ignore this and we can consider it resolved. https://github.com/h2oai/h2o-3/commit/343713a9a68e59b747e13d1936702de0cb3ab598

h2o-r/tests/testdir_algos/admissibleml/runit_PUBDEV_8075_core_infogram_erin_fail1_large.R
tomasfryda: I think it would print out only in interactive session on REPL. Not from within the function. (But I might be wrong R always surprises me)

h2o-bindings/bin/custom/python/gen_infogram.py
```diff
@@ -0,0 +1,322 @@
...
+
+        rel_cmi_frame_names = rel_cmi_frame.names
+        x_label = rel_cmi_frame_names[3]
+        y_label = rel_cmi_frame_names[4]
+        ig_x_column = 3
+        ig_y_column = 4
+        index_of_admissible = 1
+        features_column = 0
+        x_thresh = self.actual_params["relevance_threshold"]
+        y_thresh = self.actual_params["cmi_threshold"]
```
tomasfryda, line 52: This fails in the `pyunit_PUBDEV_8258_core_infogram_breast_cancer_cv_validFrame_plot.py` test. 

This seems to fix it:
```diff
index 557af1d606..dd3ebbb0cb 100644
--- a/h2o-py/h2o/estimators/infogram.py
+++ b/h2o-py/h2o/estimators/infogram.py
@@ -861,8 +861,8 @@ class H2OInfogram(H2OEstimator):
         ig_y_column = 4
         index_of_admissible = 1
         features_column = 0
-        x_thresh = self.actual_params["relevance_threshold"]
-        y_thresh = self.actual_params["cmi_threshold"]
+        x_thresh = self.actual_params[x_label + "_threshold"]
+        y_thresh = self.actual_params[y_label + "_threshold"]
 
         xmax=1.1
         ymax=1.1
```

h2o-r/h2o-package/R/infogram.R
ledell: I talked to @wendycwong: I will remove it.

h2o-r/tests/testdir_algos/admissibleml/runit_PUBDEV_8075_core_infogram_erin_fail1_large.R
ledell: I will remove it.

h2o-py/h2o/estimators/infogram.py
ledell: @wendycwong said this is fixed on the original PR & CV branch.

h2o-bindings/bin/custom/python/gen_infogram.py
ledell: @wendycwong fixed this on CV branch.

h2o-bindings/bin/custom/python/gen_infogram.py
ledell: @wendycwong and I agreed that I will come up with a mapping of old cmi function names to new names and we will update this.

ledell:
Currently the `.plot()` inside Jupyter will plot twice, we can either add a semi-colon like `.plot();` or we can fix it by wrapping it inside the `decorate_plot_result()` function like we do [here](https://github.com/h2oai/h2o-3/blob/master/h2o-py/h2o/explanation/_explain.py#L679).  I will make this update.

ledell:
I decided to remove this util function from R since we don't have any other the other util functions in R, and I think it's better to just have one way to access things, for documentation purposes (which is accessing directly with the slots in the `H2OInfogram` class. 

Posting it here in case we want to add later:

```
#' Extract the admissible features out of the H2O Infogram Model.
#'
#' @param model A fitted \linkS4class{H2OInfogram} object.
#' @export 
h2o.get_admissible_attributes <- function(model) {
  if (is(model, "H2OInfogram") && (model@algorithm == "infogram"))
    return(model@admissible_features)
}
```

ledell:
@wendycwong I made the following name changes in Python:

- `get_admissible_attributes()` -> `get_admissible_features()`
- `get_relevance_cmi_frame()` -> `get_admissible_score_frame()`

I also changed the keyname for the admissible score frame match the method name (so users can quicky identify the admissible score frame in the output).  These were updated:

```
_model_json["output"]["relevance_cmi_key"]
_model_json["output"]["relevance_cmi_key_valid"]
_model_json["output"]["relevance_cmi_key_xval"]
```

to:
```
_model_json["output"]["admissible_score_key"]
_model_json["output"]["admissible_score_key_valid"]
_model_json["output"]["admissible_score_key_xval"]
```

Lastly, I edit the Infogram param definitions in the schema.

ledell:
@wendycwong this is ready to be merged.

ledell:
@wendycwong Somehow I managed to break the R API... maybe my last commit.  I will have to try to fix it tomorrow. Python is working fine.

```
h2o.init()

# Convert iris dataset to an H2OFrame    
df <- as.h2o(iris)

# Infogram
ig <- h2o.infogram(y = "Species", training_frame = df) 
```


```r
> ig <- h2o.infogram(y = "Species", training_frame = df) 
  |=====================================================| 100%
Error in .newH2OFrame(id, id, -1, -1) : 
  base::is.character(id) is not TRUE
  ```

**Comment by ledell** [comment]
Looks like that rogue `targetencoder.py` file is causing issues.  I should just uncommit that.

http://mr-0xg1:8080/blue/organizations/jenkins/h2o-3-pipeline/detail/PR-5933/10/pipeline
Screen Shot 2021-12-19 at 2 13 35 AM

**Comment by ledell** [comment]
Reverted `targetencoder.py` edit.

File: h2o-bindings/bin/custom/python/gen_infogram.py (hunk 8 above)
Comment (ledell) [reply h2o-bindings/bin/custom/python/gen_infogram.py]: These have been updated.

**Comment by ledell** [comment]
Pushed one last commit:  VSCode missed some of my find/replaces renames of the the R code, so I manually updated classes.R.

**Comment by ledell** [comment]
One last change that I am considering: 

- In the admissible score table, rename `safety_index_raw` and `net_information_raw` columns to simply, `cmi_raw` in both cases.  That ties in the "CMI" concept back in, which is a more mathematical/generic term for safety index / net information.

Very-minimally considering:

- Is there a better name than `admissible_index`?  We don't really want people reducing admissibility to a single index, when it's really a two dimensional concept... Is there a name that better represents this concept without over-simplifying it?

**Comment by ledell** [comment]
The pending tests are related to this (will be resolved when the base branch gets changed to master):
```

[2021-12-20T05:16:27.365Z] * What went wrong:

[2021-12-20T05:16:27.365Z] Execution failed for task ':checkPullRequest'.

[2021-12-20T05:16:27.365Z] > Pull Request 5933 is linked to jira ticket https://h2oai.atlassian.net/browse/PUBDEV-8222. This jira is marked for release in {3.36.0.1}. This release will be done from a master branch. Change the base branch of the PR to 'master'.
```

wendycwong (approved):

ledell:
I added the following commits, originally on this branch https://github.com/h2oai/h2o-3/pull/5969:

- remove `compute_p_values` from infogram
- Raw CMI column changed to `cmi_raw` from domain specific context.
- Reordered some infogram params.

The last one broke the python client because `total_information_threshold` setter requires you to specify the `protected_columns` argument first (I had moved it after to be with the index thresholds...).  So I just reverted part of that commit, now just waiting for tests to go green.

PR #5938, opened 2021-12-06
PUBDEV-8420: AutoML & Grids, rethink+improve exception handling

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8420

AutoML used to ignore all exceptions thrown by model builders or grids by swallowing them early.
However, in some situations, we want to be able to make decisions on some of those errors without having the exception handling cluttered in various steps, and by ensuring that if an exception is propagated, then it's not swallowed randomly at a higher level.
The proposed solution applies the following logic:
- `ModelingStep`s don't handle exceptions and propagate them wherever they occur: model building validation, or training.
- `ModelingStepsExecutor` catch all job statuses, including the ones failing with exceptions, but collect them into a `ResultState` object that can be used at a higher level to make a decision. By swallowing all exceptions at this level, it should guarantee the robustness of AutoML, that won't fail on model or grid errors by default.
- `AutoML` can analyse the `ResultState` and decide to propagate some specific exceptions, or to throw new ones, for example if too many consecutive step failures as it's a sign a bad configuration.

On grids side: following changes apply:
- cancelled grids are not thrown away anymore and models trained until cancellation are accessible: a user (or AutoML) may want to train a long grid, interrupt it manually and still check the trained models.
- in a similar way to AutoML, if too many consecutive model fail, then the grid itself is aborted and raise a new `H2OGridException`.

### DIFF
--- a/h2o-core/src/main/java/hex/grid/Grid.java
+++ b/h2o-core/src/main/java/hex/grid/Grid.java
@@ -16,6 +16,7 @@ import java.io.InputStream;
 import java.lang.reflect.Array;
 import java.net.URI;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static hex.grid.GridSearch.IGNORED_FIELDS_PARAM_HASH;
 
@@ -128,15 +129,20 @@ public class Grid<MP extends Model.Parameters> extends Lockable<Grid<MP>> implem
     
     private void appendWarningMessage(String[] hyper_parameter, String checkField) {
       if (hyper_parameter != null && Arrays.asList(hyper_parameter).contains(checkField)) {
-        String warningMessage = "Adding alpha array to hyperparameter runs slower with gridsearch.  This is " +
-                "due to the fact that the algo has to run initialization for every alpha value.  Setting the alpha " +
-                "array as a model parameter will skip the initialization and run faster overall.";
-        Log.warn(warningMessage);
-        // Append message
-        String[] m = _warning_details;
-        String[] nm = Arrays.copyOf(m, m.length + 1);
-        nm[m.length] = warningMessage;
-        _warning_details = nm;
+        String warningMessage = null;
+        if ("alpha".equals(checkField)) {
+          warningMessage = "Adding alpha array to hyperparameter runs slower with gridsearch. " +
+                  "This is due to the fact that the algo has to run initialization for every alpha value. " +
+                  "Setting the alpha array as a model parameter will skip the initialization and run faster overall.";
+        }
+        if (warningMessage != null) {
+          Log.warn(warningMessage);
+          // Append message
+          String[] m = _warning_details;
+          String[] nm = Arrays.copyOf(m, m.length+1);
+          nm[m.length] = warningMessage;
+          _warning_details = nm;
+        }
       }
     }
 
@@ -335,9 +341,9 @@ public class Grid<MP extends Model.Parameters> extends Lockable<Grid<MP>> implem
       final String stackTrace = StringUtils.toString(t);
       final Key<Model> searchedKey = modelKey != null ? modelKey : NO_MODEL_FAILURES_KEY;
       SearchFailure searchFailure = _failures.get(searchedKey);
-      if ((searchFailure == null)) {
+      if (searchFailure == null) {
         searchFailure = new SearchFailure(_params.getClass());
-          _failures.put(searchedKey, searchFailure);
+        _failures.put(searchedKey, searchFailure);
       }
       searchFailure.appendFailedModelParameters(params, rawParams, failureDetails, stackTrace);
       searchFailure.appendWarningMessage(_hyper_names, "alpha");
@@ -440,6 +446,10 @@ public class Grid<MP extends Model.Parameters> extends Lockable<Grid<MP>> implem
     }
     searchFailure.appendWarningMessage(_hyper_names, "alpha");
     return searchFailure;
+  }
+  
+  public int countTotalFailures() {
+    return _failures.values().stream().mapToInt(SearchFailure::getFailureCount).sum();
   }
 
   /**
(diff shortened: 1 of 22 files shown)


sebhrusen:
@michalkurka : as @tomasfryda is off, do you mind having a look at this PR? Would appreciate, thx.

h2o-core/src/main/java/hex/grid/Grid.java
```diff
@@ -332,7 +332,9 @@ public Model getModel(MP params) {
    */
   private void appendFailedModelParameters(final Key<Model> modelKey, final MP params, final String[] rawParams,
                                            final Throwable t) {
-      final String failureDetails = isJobCanceled(t) ? "Job Canceled" : t.getMessage();
+      final String failureDetails = isJobCanceled(t) 
+              ? "Job Canceled, keeping the incomplete grid with all the models trained before cancellation."
```
michalkurka, line 336: this should describe individual model failure, it shouldn't have any relation to grid

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -67,14 +69,18 @@
...
   public final Key<Grid> _result;
   public final Job<Grid> _job;
   public final Recovery<Grid> _recovery;
   public final int _parallelism;
+  public int _maxConsecutiveModelFailures = DEFAULT_MAX_CONSECUTIVE_MODEL_FAILURES;
 
   /** Walks hyper space and for each point produces model parameters. It is
    *  used only locally to fire new model builders.  */
   private final transient HyperSpaceWalker<MP, ?> _hyperSpaceWalker;
+  private final AtomicInteger _consecutiveModelFailures = new AtomicInteger();
```
michalkurka, line 83: Do we want this to be a property of the grid search itself? I think it is conceptually a feature of the walker - the walker already receives notifications about failed models and it could be its responsibility to say there are no more models to train because nothing we tried succeeded (hasNext method).

Also the walker can be injected from outside of GS and any behavior can thus be introduced without GS knowing.

sebhrusen: > I think it is conceptually a feature of the walker

imo, the responsibility of the walker is limited to providing new model builders: I agree that the walker can stop the walking but it can not fail the job (or it would be in a very unusual way, like throwing on `hasNext()`), and I think that interrupting abruptly the entire job is a responsibility of the entity that owns the job, ie. GS.
Another way to look at it: if a model fails, there could be a policy to retry the training. Who would be responsible for this, the walker or the grid search? Imo, it's the search that would be responsible for this, as retrying is outside the scope of iterating. So, in general, I would keep any decision outside the iteration logic per se outside the walker.

h2o-core/src/main/java/hex/grid/Grid.java
sebhrusen: right, will revert

h2o-core/src/main/java/hex/grid/GridSearch.java
michalkurka: @sebhrusen do you need all this functionality in grid search? 

For example - can it be limited to eg. 10 consecutive failed models with no model trained? This is something you can control from the walker and it is still useful, failures can be handled on AutoML side.

To be honest I don't feel comfortable reviewing changes in GS of this kind. If we could find a way to make it useful in AutoML without structural changes in GS then that would be ideal but if the scope has to stay this broad I think we will have to hold off for after the major release.

sebhrusen: @michalkurka I updated grid to fail on consecutive failures iff no model has been trained yet. Is it ok for you?
I could also change the default `_maxConsecutiveModelFailures` to 0 for grids, and set it to 10 only in `AutoML` but this would mean that users can still pass wrong hyperparams and have a grid running forever without producing anything.

michalkurka: I think that looks good

I would keep the default "unlimited", we can then introduce a parameter for grid search to expose it to users and pick a reasonable default

h2o-core/src/main/java/hex/grid/GridSearch.java
```diff
@@ -227,12 +244,19 @@ public void onBuildSuccess(final Model finishedModel, final ParallelModelBuilder
...
       try {
-        grid.appendFailedModelParameters(null, modelBuildFailure.getParameters(), modelBuildFailure.getThrowable());
+        grid.appendFailedModelParameters(null, modelBuildFailure.getParameters(), ex);
       } finally {
         parallelSearchGridLock.unlock();
       }
-      attemptBuildNextModel(parallelModelBuilder, null);
+      if (!Grid.isJobCanceled(ex) 
+              && grid.getModelCount() == 0 
+              && _consecutiveModelFailures.incrementAndGet() >= _maxConsecutiveModelFailures) {
```
michalkurka, line 255: I feel like we should be able to get rid of this counter since we introduced "&& grid.getModelCount() == 0 "

the reason why I think it is an issue is that it introduces a bug for "recovered" grids - that is grids that are reloaded from disk and the building continues - the state is lost - or at least it seems like it could be lost

because we have model count == 0 we can just check how many parameters failed - `SearchFailures getFailures ` and see how many failed paramters are there

this would simplify the code, get rid off state and get me some sleep at night ;)

sebhrusen: good point! I forgot that we had access to the history of failures, will change this and care of your sleep patterns ;)

sebhrusen: I don't know why `GridSearch` API is so closed but it's a real pita to configure it from AutoML with it's 200 `startGridSearch` methods as confusing as each other.
Thinking about making a builder quickly, wdyt?

michalkurka: An additive change like make a lot of sense

sebhrusen:
build is fully green, any concern left? @michalkurka @tomasfryda @ledell

ledell (approved):
This looks good, thanks!

tomasfryda (approved):
Looks good to me, thank you @sebhrusen !

PR #5940, opened 2021-12-08
Buildscript fix: fix exludes after migrating to Gradle 7 (affects CDP) [nocheck]

michalkurka:


valenad1 (approved):
Aa excludes.. thank you for the fix!

koniecsveta (approved):

PR #5941, opened 2021-12-08
Make sure steam-extension is tested just like everything else [nocheck]

michalkurka:
This is needed for proper property propagation (eg. to enable specific java version) and also log and test result collection


michalkurka:
Screen Shot 2021-12-08 at 7 09 15 PM

valenad1 (approved):
Looks great! 👍 Java failed for another reason. Sorry for the delayed approve

PR #5947, opened 2021-12-10
Avoid direct number object instantiation - causes warning on Java 16+ [nocheck]

michalkurka:


h2o-jetty-8/src/main/java/org/eclipse/jetty/server/Response.java
```diff
@@ -342,7 +342,7 @@ public void sendError(int code, String message) throws IOException
                 error_handler = _connection.getConnector().getServer().getBean(ErrorHandler.class);
             if (error_handler!=null)
             {
-                request.setAttribute(RequestDispatcher.ERROR_STATUS_CODE,new Integer(code));
+                request.setAttribute(RequestDispatcher.ERROR_STATUS_CODE, code);
```
valenad1, line 345: I presume that primitive int is good enough here

valenad1 (approved):

h2o-jetty-8/src/main/java/org/eclipse/jetty/server/Response.java
michalkurka: yep, it will get auto-boxed into Integer

PR #5948, opened 2021-12-10
[PUBDEV-8452][FOLLOWUP] Rename lambda Field on RuleFitParameters to _lambda

mn-mikke:
This exception in naming conventions for fields on parameter classes breaks API generation in SW. Fields on parameter classes (e.g.`RuleFitParameters`) serve as a source of default values for SW API.

### DIFF
--- a/h2o-algos/src/main/java/hex/rulefit/RuleFit.java
+++ b/h2o-algos/src/main/java/hex/rulefit/RuleFit.java
@@ -167,8 +167,8 @@ public class RuleFit extends ModelBuilder<RuleFitModel, RuleFitModel.RuleFitPara
             glmParameters._weights_column = "linear." + _parms._weights_column;
         }
         glmParameters._auc_type = _parms._auc_type;
-        if (_parms.lambda != null) {
-            glmParameters._lambda = _parms.lambda;
+        if (_parms._lambda != null) {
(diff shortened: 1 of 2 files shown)


maurever (approved):
LGTM, I can't see any problem, but @koniecsveta would have a more meaningful review.

michalkurka:
@koniecsveta was it not generating any warning before? I would expect this would be caught

koniecsveta:
> @koniecsveta was it not generating any warning before? I would expect this would be caught

No. It was a reserved word in python api, hence I had to adapt it there. I did that the way it is done in glm api. But no warning for the java field, this did not cross my mind.

koniecsveta (approved):

michalkurka:
@koniecsveta thanks for confirming, I thought violating the underscore convention would trigger a warning but looks like it doesn't - there is only a check for the other direction (schema -> pojo)

```
          if (name.startsWith("_"))
            Log.warn("Found schema field which violates the naming convention; name starts with underscore: " +
                     meta.name + "." + name);
```

michalkurka (approved):
Thanks for catching this and making a fix @mn-mikke

michalkurka:
We should probably extend the checks on the H2O side if a violation of the convention causes issues to SW.

PR #5951, opened 2021-12-11
PUBDEV-8416, PUBDEV-8417: AutoML propagate early stopping criteria to SEs + GLM

sebhrusen:
https://h2oai.atlassian.net/browse/PUBDEV-8416
https://h2oai.atlassian.net/browse/PUBDEV-8417


h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -332,7 +336,6 @@ private void notifyModelListeners() {
   public Job<M> trainModelOnH2ONode() {
     if (error_count() > 0)
       throw H2OModelBuilderIllegalArgumentException.makeFromBuilder(this);
-    this._input_parms = (P) this._parms.clone();
```
sebhrusen, line 335: @koniecsveta I think this one was a mistake as we're cloning AFTER validation, during which many parms are set or modified internally.

ledell:
We will revisit this after the holidays after reviewing the benchmarks more carefully.

PR #5952, opened 2021-12-13
PUBDEV-7858: Expose distribution parameter in AutoML

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-7858

Flow PR: https://github.com/h2oai/h2o-flow/pull/179

## Overview

This PR exposes the `distribution` parameter in AutoML. It is supposed to work in the following way:
 Algos that support the specified distribution will use the specified distribution, other algos use the default distribution for the dataset.
 
Supported distributions by algos:
Screen Shot 2022-01-05 at 14 46 51

 
 ### API Proposal
 
 #### Python
```python
H2OAutoML(distribution="laplace")
H2OAutoML(distribution=dict(type="tweedie", tweedie_power=1.5))
H2OAutoML(distribution="tweedie")  #also works when you don't need to set additional args
```

#### R
```r
h2o.automl(distribution = "laplace", ...)
h2o.automl(distribution = list(type = "tweedie", tweedie_power = 1.5), ...)
h2o.automl(distribution = "tweedie", ...)  #also works when you don't need to set additional args
```

## Current Issues
* Unsupported distributions in AutoML:
    * `ordinal` - only GLM without lambda search supports it and we don't have such model in AutoML
    * `fractionalbinomial`, `quasibinomial` - only GLM (and GBM for quasibinomial) support it, and the default distribution for the rest of the models is gaussian which yields incompatible metrics
    * `modified_huber` - we say we support it but we don't (https://github.com/h2oai/h2o-3/blob/0d22bdb16ea0504e5e99c454f86a278161e8fa69/h2o-core/src/main/java/hex/ModelBuilder.java#L1413-L1415)

## TODO:
* bump the version of flow once flow gets merged and built
* add option to enforce strict usage of given constraints such as distribution or monotonicity constraints - this will be in *another PR* as it will probably take some time to come up and agree on an appropriate API and it does influence more than just distributions


h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -92,6 +100,25 @@ public boolean haveMojo() {
...
+      if (multinomial.equals(distribution) || ordinal.equals(distribution) || AUTO.equals(distribution))
+        return _nclass;
+      if (bernoulli.equals(distribution) || quasibinomial.equals(distribution)
+              || fractionalbinomial.equals(distribution))
+        return 2;
+      return 1;
+    }
+    return super.nclasses();
+  }
+
```
tomasfryda, line 121: Fix problem with quasibinomial/fractionalbinomial distributions in SE - they need to be interpreted as classification not regression (which is the current state).

ledell:
@tomasfryda I wonder if we can find a place in the user guide to talk about the `distribution` arg?  Since it's generic to all algos, then maybe we could put it on the [Training Models](https://docs.h2o.ai/h2o/latest-stable/h2o-docs/training-models.html) page?  I like your chart above and I think it would be great to include it in the docs.  I think it might make sense to include `X` for rather than `NA` for the RF column since they are all really non-applicable in theory to the RF algorithm.  

I found the reason that we actually have a `distribution` arg in RF to begin with (since it doesn't apply to the algo in any way).  It was a compatibility thing... we included it originally by mistake, then it was briefly removed because it didn't belong there, but removing it was a breaking change, so that change was [reverted](https://h2oai.atlassian.net/browse/PUBDEV-5335).  I made a related [ticket](https://h2oai.atlassian.net/browse/PUBDEV-8510) to add a warning when it's set by the user.

ledell:
@tomasfryda I was wondering what happens when you pass in the "wrong" distribution.  I tried passing in `"bernoulli"` for a multiclass problem and there were no warnings or errors.  What kind of behavior do we want to have in these cases?  

I wonder if we should consider dividing the distributions up into task type and throwing an error (or warning?) if they don't match up?  e.g. if user passed "gaussian" to an AutoML with a classification training frame, where response is a factor.

```
library(h2o)
h2o.init()

 aml <- h2o.automl(y = "Species", training_frame = as.h2o(iris), max_models = 10, distribution = "bernoulli")
 ```

**Comment by tomasfryda** [comment]
Thank you @ledell. That a good point, I added that error. What surprised me is that when I ran your example I noticed we can train DeepLearning with `Bernoulli` distribution on multinomial data. I didn't investigate that further but it's a strong reason for having that error.

File: h2o-core/src/main/java/hex/Model.java [hunk 2]
```diff
@@ -713,6 +713,10 @@ private void addToUsedIfColumn(Set<String> usedColumns, Set<String> allColumns,
       return values;
     }
 
+    public DistributionFamily[] supportedDistributions() {
```
michalkurka, line 716: Why is a list of supported distributions defined on `Parameter` object?

It seems the reason is that the output depends on the Parameters itself - eg.: https://github.com/h2oai/h2o-3/pull/5952/files#diff-1b45f7fd4b0c10d1411dcd1399033042b4916dca63fc6a5f881372d44b3f250dR165

This makes the overall semantic difficult to understand and it seems it is purpose build for AutoML use case and might be difficult to translate to other use cases as well.

It is definitely not obvious I need to fill all the other parameters and then call this function to see if I can also set the desired distribution. This duplicates what `init(false)` is meant for.

Instead of this custom function, can you set parameters including distribution, call init(false) and check if the '_distribution' had an error? or if there was any other error?

h2o-algos/src/main/java/hex/rulefit/RuleFit.java
```diff
@@ -122,40 +123,7 @@ private void initGLMParameters() {
...
-                case fractionalbinomial:
-                    glmParameters = new GLMModel.GLMParameters(GLMModel.GLMParameters.Family.fractionalbinomial);
-                    break;
-                case negativebinomial:
-                    glmParameters = new GLMModel.GLMParameters(GLMModel.GLMParameters.Family.negativebinomial);
-                    break;
-                default:
-                    error("_distribution", "Distribution not supported.");
-            }
+            glmParameters = new GLMModel.GLMParameters(distributionToFamily(_parms._distribution));
```
michalkurka, line 126: thank you!

michalkurka (changes requested):
@tomasfryda thank you for this effort - a lot of work went into this. 

I have some concerns about the `supported distributions` function:
- it is public but the semantics is not clear (it needs parameter object partially filled)
- it duplicates what init(false) is meant to do
- it is only used in AutoML and it might get out of sync with further changes in the algos

If only meant for AutoML then Model should allow all current possible distributions to be supported as opposed to just AUTO. This would allow the general developer audience to call supportedDistributions without seeing errors on 3rd party models.

h2o-core/src/main/java/hex/Model.java
michalkurka: The validations in the `init` functions are likely faulty in some cases (there are inconsistencies you pointed out) and we should fix those.

> Could it possibly be due to some reuse in RuleFit or IsolationForest? (rhetorical, I will see once I try it)
Not for Isolation Forest, possibly RuleFit but I doubt it - likely an omission.

I think we could split the PR into 2 steps:
- first adapt AutoML to respect the validations in the models (init method) - this is what users can do anyway
- fixing all validations in ModelBuilders with your findings (that doesn't have to be you who fixes this)

ledell:
> What surprised me is that when I ran your example I noticed we can train DeepLearning with Bernoulli distribution on multinomial data. I didn't investigate that further but it's a strong reason for having that error.

@tomasfryda I guess that's a DL bug, and also an RF bug?  Looks like all the other models are failing (as they should).  It would be a breaking change if we added an error for DL and RF, so that's something to consider, but maybe a useful one.

```r
> lb <- aml@leaderboard
> print(lb, n = nrow(lb))
                                              model_id mean_per_class_error    logloss      rmse
1 DeepLearning_grid_3_AutoML_1_20220111_231419_model_2           0.02000000 0.16138286 0.1528847
2 DeepLearning_grid_3_AutoML_1_20220111_231419_model_1           0.02666667 0.10365598 0.1508232
3 DeepLearning_grid_1_AutoML_1_20220111_231419_model_2           0.02666667 0.07837208 0.1530357
4 DeepLearning_grid_1_AutoML_1_20220111_231419_model_1           0.03333333 0.18943732 0.1823144
5 DeepLearning_grid_2_AutoML_1_20220111_231419_model_1           0.04000000 0.11371872 0.1865889
6 DeepLearning_grid_1_AutoML_1_20220111_231419_model_3           0.04000000 0.15215178 0.1844747
7                       DRF_1_AutoML_1_20220111_231419           0.04666667 0.32990601 0.1930501
8                       XRT_1_AutoML_1_20220111_231419           0.06666667 0.34003373 0.2040754
9              DeepLearning_1_AutoML_1_20220111_231419           0.11333333 0.23167989 0.2738521
         mse
1 0.02337372
2 0.02274764
3 0.02341991
4 0.03323854
5 0.03481542
6 0.03403092
7 0.03726834
8 0.04164676
9 0.07499499
```

sebhrusen (changes requested):
more to come

h2o-algos/src/main/java/hex/deeplearning/DeepLearning.java
```diff
@@ -70,6 +70,20 @@ public ToEigenVec getToEigenVec() {
...
+            DistributionFamily.multinomial,
+            DistributionFamily.gaussian,
+            DistributionFamily.poisson,
+            DistributionFamily.gamma,
+            DistributionFamily.laplace,
+            DistributionFamily.quantile,
+            DistributionFamily.huber,
+            DistributionFamily.tweedie,
+    };
+    if (!(Arrays.stream(allowed_distributions).anyMatch(_parms._distribution::equals)))
```
sebhrusen, line 85: contains

h2o-algos/src/main/java/hex/ensemble/Metalearners.java
```diff
@@ -134,7 +133,9 @@ protected void validateParams(Model.Parameters parms) {
             super.validateParams(parms);
 
             // Check if distribution family is supported and if not pick a basic one
-            if (!ArrayUtils.contains(_supportedDistributionFamilies, parms._distribution)) {
+            ModelBuilder mb = ModelBuilder.make(parms);
+            mb.init(false);
+            if (mb.error_count() > 0) {
```
sebhrusen, line 138: how do you know that those errors where distribution related? psychic?

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLV99.java
```diff
@@ -4,6 +4,7 @@
 import ai.h2o.automl.events.EventLog;
 import ai.h2o.automl.leaderboard.Leaderboard;
 import ai.h2o.automl.StepDefinition;
+import hex.genmodel.utils.DistributionFamily;
```
sebhrusen, line 7: unused?

h2o-core/src/main/java/hex/Model.java
```diff
@@ -747,6 +747,14 @@ public final String getTreatmentColumn(){
...
       return _max_categorical_levels;
     }
+
+    public void setDistributionFamily(DistributionFamily distributionFamily){
+      _distribution = distributionFamily;
+    }
+
+    public DistributionFamily getDistributionFamily() {
+      return _distribution;
+    }
```
sebhrusen, line 757: shouldn't we make the `_distribution`, `_family`, `_link`... properties private now that we have those accessors? How otherwise can we enforce that those properties are always set through this setter?

h2o-py/h2o/automl/_estimator.py
```diff
@@ -404,11 +411,45 @@ def __validate_algo_parameters(self, algo_parameters):
...
+    def __validate_distribution(self, distribution):
+        if is_type(distribution, str):
+            distribution = distribution.lower()
+            if distribution == "custom":
+                raise H2OValueError('Distribution "custom" has to be specified as a '
+                                    'dictionary with their respective parameters, e.g., '
+                                    '`dict(distribution = \"custom\", custom_distribution_func = \"...\"))`.')
+            return distribution
+        if is_type(distribution, dict):
+            dist = distribution["distribution"].lower()
```
sebhrusen, line 423: isn't it weird to use `distribution=dict(distribution=…)`? twice the same identifier? can confuse the user.
Usually you use something like`type` in that case:
```python
distribution=dict(type="tweedie", tweedie_power=0.5)
```

h2o-py/h2o/automl/_estimator.py
```diff
@@ -404,11 +411,45 @@ def __validate_algo_parameters(self, algo_parameters):
...
+            ALLOWED_DISTRIBUTION_PARAMETERS = dict(
```
sebhrusen, line 424: why the upper case? isn't your Py linter complaining? it's just a normal local variable, but apparently an arrogant one :)

h2o-py/h2o/automl/_estimator.py
```diff
@@ -404,11 +411,45 @@ def __validate_algo_parameters(self, algo_parameters):
...
     _extract_params_doc(getdoc(__init__))
     project_name = _aml_property('build_control.project_name', types=(None, str), freezable=True,
                                  validate_fn=__validate_project_name)
     nfolds = _aml_property('build_control.nfolds', types=(int,), freezable=True,
                            validate_fn=__validate_nfolds)
+    distribution = _aml_property('build_control.distribution', types=(str, dict), freezable=True, validate_fn=__validate_distribution)
+    custom_distribution_func = _aml_property('build_control.custom_distribution_func', types=(str,), freezable=True)
+    huber_alpha = _aml_property('build_control.huber_alpha', types=(numeric,), freezable=True)
+    tweedie_power = _aml_property('build_control.tweedie_power', types=(numeric,), freezable=True)
+    quantile_alpha = _aml_property('build_control.quantile_alpha', types=(numeric,), freezable=True)
```
sebhrusen, line 452: can you make those properties "private" (prefixed with underscore)?
Otherwise it provides 2 official ways to set those properties: the dict keys on `distribution` property, or directly through those properties, except that there won't be proper validation in the latter case.

sebhrusen (commented):
next is the last

h2o-algos/src/main/java/hex/deeplearning/DeepLearning.java
```diff
@@ -70,6 +70,20 @@ public ToEigenVec getToEigenVec() {
...
+      error("_distribution", _parms._distribution.name() + " is not supported for XGBoost in current H2O.");
```
michalkurka, line 86: xgboost?

h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
```diff
@@ -147,6 +152,25 @@ protected void setMetalearnerParameters(StackedEnsembleParameters params) {
...
+                } catch (H2OIllegalArgumentException e) {
+                    return false;
+                }
+                // Just to get the correct response
+                setCommonModelBuilderParams(params);
+                ModelBuilder mb = ModelBuilder.make(params);
+                mb.init(false);
+                return Arrays.stream(new String[]{"_distribution", "_family"})
+                        .allMatch((field) ->
+                                mb.getMessagesByFieldAndSeverity(field, Log.ERRR).length == 0);
```
sebhrusen, line 173: if this is the only reason why you need this `supportsDistribution` method, then there must be a better alternative: this is too complex and redundant.

You already have `StackedEnsembleModel.setDistributionFamily()` forwarding to the metalearner. So standard validation logic should check everything correctly in `StackedEnsemble.init()`: nothing that looks specific to SEs in AutoML, it should be done for any standard SE.

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
```diff
@@ -105,6 +106,45 @@
...
+            mb.init(false);
+            return Arrays.stream(fields)
+                    .allMatch((field) ->
+                            mb.getMessagesByFieldAndSeverity(field, Log.ERRR).length == 0);
+        } catch (H2OIllegalArgumentException e) {
+            return false;
+        }
+    }
+
+    protected void setDistributionParameters(Model.Parameters parms) {
```
sebhrusen, line 129: imo, this should be applied directly for all models in `ModelingStep.ModelStep.trainModel`, and `ModelingStep.GridStep.hyperparameterSearch` as that's where user input parameters are applied, not in each individual step.

h2o-py/tests/testdir_algos/automl/pyunit_automl_distributions.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def test_python_api():
+    def test_parameterized_distribution_without_param():
+        try:
+            aml = H2OAutoML(distribution=dict(distribution="huber"))
+            aml = H2OAutoML(distribution="tweedie")
+            aml = H2OAutoML(distribution="quantile")
+        except AssertionError:
+            assert False, "should not have failed"
```
sebhrusen, line 121: that's a strange assertion: why not let the error be propagated normally?

h2o-py/tests/testdir_algos/automl/pyunit_automl_distributions.py
```diff
@@ -0,0 +1,144 @@
...
+        df["ordinal_factors"] = df["ordinal"].asfactor()
+        return df
+
+    def get_distribution(model_id):
+        ap = h2o.get_model(model_id).actual_params
+        if "metalearner_params" in ap.keys():
+            ap = h2o.get_model(model_id).metalearner().actual_params
+        distribution = ap.get("distribution", ap.get("family"))
+        if distribution == "binomial":
+            distribution = "bernoulli"
```
sebhrusen, line 60: curious: why do you need this additional conversion here? is it only because of GLM or sth?

h2o-py/tests/testdir_misc/pyunit_explain.py
```diff
@@ -618,6 +618,8 @@ def test_learning_curve_for_algos_not_present_in_automl():
     assert isinstance(if_model.learning_curve_plot().figure(), matplotlib.pyplot.Figure)
     matplotlib.pyplot.close()
 
+# Remove unnecesary warnings
+matplotlib.rcParams["figure.max_open_warning"] = 0
```
sebhrusen, line 622: unrelated?

sebhrusen (commented):
the end. Thanks for the effort @tomasfryda .
Most of my comments are syntactic except for my concerns regarding where distribution param should be set and validated in AutoML.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -609,14 +614,49 @@ private void optionallySplitTrainingDataset() {
...
-    throw new RuntimeException("Number of domains is equal to 1.");
+    if (_buildSpec.build_control.distribution == DistributionFamily.AUTO) {
+      if (numOfDomains == 0)
+        return DistributionFamily.gaussian;
+      if (numOfDomains == 2)
+        return DistributionFamily.bernoulli;
+      if (numOfDomains > 2)
+        return DistributionFamily.multinomial;
+
+      throw new RuntimeException("Number of domains is equal to 1.");
```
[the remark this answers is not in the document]
ledell: @sebhrusen We actually use the term "domain" in the model code (it's exposed in the model JSON in Python, but not stored in R models for some reason).  I actually think we should expose this in R too because it's handy to have this info sometimes.  However, if we want to use this idea in an error message, I might use a different, more obvious term like "number of classes".

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -74,6 +73,7 @@ private static LeaderboardExtensionsProvider createLeaderboardExtensionProvider(
                 new ScoringTimePerRow(model, aml.getLeaderboardFrame(), aml.getTrainingFrame()),
 //                new ModelSize(model._key)
                 new AlgoName(model),
+                new Distribution(model),
```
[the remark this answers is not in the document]
ledell: I agree.

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
```diff
@@ -44,6 +45,13 @@
     public float max_after_balance_size = 5.0f;
 
     public int nfolds = -1;
+
+    public DistributionFamily distribution = DistributionFamily.AUTO;
+    public String custom_distribution_func;
+    public double tweedie_power = 1.5;
+    public double quantile_alpha = 0.5;
+    public double huber_alpha = 0.9;
```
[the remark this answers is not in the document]
ledell: I also couldn't figure out what this is for... Are these meant to be passed down to some of the algos?

h2o-py/h2o/automl/_estimator.py
ledell: Oops, this is from the original API spec, but now it seems obvious that this is a bit weird/redundant.  This is what we had:

 #### Python
```python
H2OAutoML(distribution="laplace")
H2OAutoML(distribution=dict(distribution="tweedie", tweedie_power=1.5))
H2OAutoML(distribution="tweedie")  #also works when you don't need to set additional args
```

#### R
```r
h2o.automl(distribution = "laplace", ...)
h2o.automl(distribution = list(distribution = "tweedie", tweedie_power = 1.5), ...)
h2o.automl(distribution = "tweedie", ...)  #also works when you don't need to set additional args
```

h2o-docs/src/product/data-science/algo-params/distribution.rst
```diff
@@ -160,6 +160,120 @@ Where:
 
 **Note**: Inversion, Ologit, Ologlog, and Oprobit are not associated with concrete distributions but can be used with a custom distribution.
 
+AutoML
+~~~~~~
```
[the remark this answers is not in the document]
ledell: @sebhrusen I'm not sure that I've seen that before, but it makes sense to have it's own section.  Yeah there might be some other places we could do this.

h2o-py/tests/testdir_algos/automl/pyunit_automl_distributions.py
ledell: Yeah, we have a naming conflict between GLM which is based on `family` values of glmnet, but "bernoulli" is a better name for distribution.  More info: https://www.statology.org/bernoulli-vs-binomial/

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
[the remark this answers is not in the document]
tomasfryda: The magic comes from `Model.Parameters` declaration. 

https://github.com/h2oai/h2o-3/blob/89a96768cdba79b10fcf530d4b55e0a09ea75b21/h2o-core/src/main/java/hex/Model.java#L393-L395

There isn't logic that would change the `-1` to the default value (yet). 

I think the main reason is to provide some information to the h2o flow users - so the corresponding fields are filled in with a value that will be used, `-1` could be confusing there I think.

h2o-automl/src/main/java/ai/h2o/automl/leaderboard/Distribution.java
```diff
@@ -0,0 +1,37 @@
+package ai.h2o.automl.leaderboard;
+
+import hex.Model;
+import water.Iced;
+import water.Key;
+
+public class Distribution extends Iced<Distribution> implements LeaderboardCell<String, Distribution> {
```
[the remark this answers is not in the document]
tomasfryda: Good point, thank you! I added this mainly for debugging the issue with deviances that lead to [this PR](https://github.com/h2oai/h2o-3/pull/6050).

h2o-algos/src/main/java/hex/ensemble/StackedEnsemble.java
```diff
@@ -92,6 +97,25 @@ public boolean haveMojo() {
...
+    if (_parms._metalearner_parameters != null) {
+      if (_parms._metalearner_parameters instanceof GLMModel.GLMParameters)
+        distribution = familyToDistribution(((GLMModel.GLMParameters) _parms._metalearner_parameters)._family);
+      else
+        distribution = _parms._metalearner_parameters._distribution;
+
+      if (multinomial.equals(distribution) || ordinal.equals(distribution) || AUTO.equals(distribution))
+        return _nclass;
+      if (bernoulli.equals(distribution) || quasibinomial.equals(distribution)
+              || fractionalbinomial.equals(distribution))
```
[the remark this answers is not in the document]
tomasfryda: > I know I'm annoying

On the contrary, I'm glad I'm getting this kind of feedback as well and it does make the code easier to read.

h2o-core/src/main/java/hex/Model.java
tomasfryda: The `DistributionFamily` has no information of `link` as it is used only together with `family`. (The same goes for the distribution parameters (`twedie_power`, etc).) This is ok if we use it to set `family` inferred from `distribution` but it would be limiting for other use-cases. 

I think the benefit of this setter and getter is limited to AutoML and "compound" algos (SE, rulefit) so I wouldn't enforce its usage.

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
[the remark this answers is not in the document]
tomasfryda: Changed to `"Number of classes is equal to 1."`. And in the following messages changed to use "classes" instead of "domains".

h2o-automl/src/main/java/ai/h2o/automl/AutoML.java
```diff
@@ -689,6 +729,10 @@ private void learn() {
     }
     for (ModelingStep step : getExecutionPlan()) {
       if (!exceededSearchLimits(step)) {
+        if (!step.supportsDistribution(_distributionFamily)) {
+          _eventLog.warn(Stage.ModelTraining,"Step " + step._algo.name() + "." + step._id +
+                  " doesn't support " + _distributionFamily.name() + " distribution. Using AUTO distribution instead.");
```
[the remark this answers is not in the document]
tomasfryda: Changed it to `info` level.

h2o-automl/src/main/java/ai/h2o/automl/ModelingStep.java
```diff
@@ -105,6 +106,45 @@
...
+        if (aml().getDistributionFamily().equals(DistributionFamily.custom))
+            parms._custom_distribution_func = aml().getBuildSpec().build_control.custom_distribution_func;
+        if (aml().getDistributionFamily().equals(DistributionFamily.huber))
+            parms._huber_alpha = aml().getBuildSpec().build_control.huber_alpha;
+        if (aml().getDistributionFamily().equals(DistributionFamily.tweedie))
+            parms._tweedie_power = aml().getBuildSpec().build_control.tweedie_power;
+        if (aml().getDistributionFamily().equals(DistributionFamily.quantile))
+            parms._quantile_alpha = aml().getBuildSpec().build_control.quantile_alpha;
```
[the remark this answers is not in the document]
tomasfryda: Done

h2o-algos/src/main/java/hex/ensemble/Metalearners.java
tomasfryda: Fixed, thank you!

h2o-automl/src/main/java/water/automl/api/schemas3/AutoMLV99.java
tomasfryda: Removed.

h2o-py/h2o/automl/_estimator.py
tomasfryda: Changed to use `list(type = `/`dict(type =` instead of `list(distribution = `/`dict(distribution=`.

h2o-py/h2o/automl/_estimator.py
tomasfryda: My Py linter is complaining about a lot of things including this one. It's lower-cased now.

h2o-py/h2o/automl/_estimator.py
tomasfryda: Done. Thank you for the explanation!

h2o-algos/src/main/java/hex/deeplearning/DeepLearning.java
tomasfryda: Thank you!

h2o-r/h2o-package/R/automl.R
```diff
@@ -222,6 +224,38 @@ h2o.automl <- function(x, y, training_frame,
   build_control$stopping_criteria$stopping_metric <- ifelse(length(stopping_metric) == 1,
                                                             match.arg(tolower(stopping_metric), tolower(formals()$stopping_metric)),
                                                             match.arg(stopping_metric))
+  if (!is.null(distribution) && !missing(distribution)) {
+    if (is.list(distribution)) {
+      build_control$distribution <- distribution$distribution
```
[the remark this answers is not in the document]
tomasfryda: Changed to `distribution$type`.

h2o-automl/src/main/java/ai/h2o/automl/modeling/StackedEnsembleStepsProvider.java
tomasfryda: Good point, thank you! `supportsDistribution` is no longer used so I removed it completely.

h2o-py/tests/testdir_misc/pyunit_explain.py
tomasfryda: Removed, thank you!

h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java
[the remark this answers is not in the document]
sebhrusen: I see, it's mainly for the prefilled field in Flow, right?

sebhrusen (approved):
Looks great, thanks again @tomasfryda

michalkurka (approved):

michalkurka:
@tomasfryda LGTM - there are however failed test that I've never seen fail before - maybe re-run first

tomasfryda:
All tests passed => merging.

PR #5954, opened 2021-12-13
After S3 libraries upgrade we don't need to open sun.security.ssl module [nocheck]

michalkurka:


michalkurka:
@valenad1 please review, after you approve I will revert the changes in the test pipeline definition

valenad1 (approved):
👍

PR #5956, opened 2021-12-14
HEXDEV-0779: Added a test for log4j2 vulnerability

michalkurka:


h2o-py/tests/testdir_misc/pyunit_log4j_cve.py
```diff
@@ -0,0 +1,84 @@
...
+    t.start()
+
+    conn, _ = sock.accept()
+    with conn:
+        data = conn.recv(10)
+        # if we do receive any data - it was H2O's log4j that called us
+        called_by_log4j = len(data) > 0
+        conn.close()
+
+    assert not called_by_log4j
```
sebhrusen, line 78: I would wrap this with 
```python
try:
    …
finally:
    sock.close()
```
otherwise, on failing test, the socket remains open

sebhrusen: and it makes me doubt about the utility of the poison pill

sebhrusen: I see, the poison pill is indeed useful on a fixed jar :)

sebhrusen: FYI, a simpler test, without the poison pill, but relying on timeout when waiting for the incoming connection:
```python
def test_log4j_cve():
    sock = None
    for res in socket.getaddrinfo("localhost", 50001, socket.AF_INET,
                                  socket.SOCK_STREAM, 0, socket.AI_PASSIVE):
        af, socktype, proto, canonname, sa = res
        try:
            sock = socket.socket(af, socktype, proto)
        except OSError:
            sock = None
            continue
        try:
            sock.bind(sa)
            sock.settimeout(5)
            sock.listen(1)
        except OSError:
            sock.close()
            sock = None
            continue
        break
    assert sock is not None

    try:
        t = threading.Thread(target=malicious_log)
        t.start()

        conn, _ = sock.accept()
        with conn:
            data = conn.recv(10)
            print(data.decode('utf-8'))
            # if we do receive any data - it was H2O's log4j that called us
            called_by_log4j = len(data) > 0
            conn.close()
            
        assert not called_by_log4j
    except socket.timeout:
        pass
    finally:
        sock.close()
```

michalkurka:
thank you @sebhrusen - I will add the try-catch before merging

PR #5959, opened 2021-12-14
Upgrade to log4j 2.16.0 to address CVE-2021-45046 [nocheck]

michalkurka:


h2o-assemblies/main/build.gradle
```diff
@@ -83,6 +83,8 @@ shadowJar {
   exclude 'test.properties'
   exclude 'cockpitlite.properties'
   exclude 'devpay_products.properties'
+  exclude 'org/apache/logging/log4j/core/lookup/JndiLookup.class'
+  exclude 'org/apache/logging/log4j/core/lookup/JmxRuntimeInputArgumentsLookup.class'
```
mmalohlava, line 87: do you need to remove 2nd class?

mmalohlava (approved):
ACK, THX!

mmalohlava:
@achraf-mer for info

h2o-assemblies/main/build.gradle
mmalohlava: after discussion with @coolsat there is a concern that the change could impact "innocent" customers (who needs JNDI)

michalkurka: ok, reverting back to just version upgrade

PR #5961, opened 2021-12-15
PUBDEV-7752 Disable centroid stats for CV

maurever:
JIRA: https://h2oai.atlassian.net/browse/PUBDEV-7752

### DIFF
--- a/h2o-core/src/main/java/hex/ModelBuilder.java
+++ b/h2o-core/src/main/java/hex/ModelBuilder.java
@@ -903,13 +903,13 @@ abstract public class ModelBuilder<M extends Model<M,P,O>, P extends Model.Param
     mainModel._output._cross_validation_models = _parms._keep_cross_validation_models ? cvModKeys : null;
     Key<Frame>[] predKeys = new Key[N];
     mainModel._output._cross_validation_predictions = _parms._keep_cross_validation_predictions ? predKeys : null;
-
+    
     for (int i = 0; i < N; ++i) {
-      if (i > 0) mbs[0].reduce(mbs[i]);
+      if (i > 0) cv_reduceCVBuilders(mbs[0], mbs[i]);
       cvModKeys[i] = cvModelBuilders[i]._result;
       predKeys[i] = Key.make(cvModelBuilders[i].getPredictionKey());
     }
-
+    
     Frame holdoutPreds = null;
     if (_parms._keep_cross_validation_predictions || (nclasses()==2 /*GainsLift needs this*/ || mainModel.isDistributionHuber())) {
       Key<Frame> cvhp = Key.make("cv_holdout_prediction_" + mainModel._key.toString());
@@ -933,8 +933,7 @@ abstract public class ModelBuilder<M extends Model<M,P,O>, P extends Model.Param
       int count = Model.deleteAll(predKeys);
       Log.info(count+" CV predictions were removed");
     }
-
-    mainModel._output._cross_validation_metrics = mbs[0].makeModelMetrics(mainModel, _parms.train(), null, holdoutPreds);
+    cv_makeMainModelMetrics(mainModel, mbs[0],  holdoutPreds);
     if (holdoutPreds != null) {
       if (_parms._keep_cross_validation_predictions) Scope.untrack(holdoutPreds.keysList());
       else holdoutPreds.remove();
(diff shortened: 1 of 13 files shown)


h2o-core/src/main/java/hex/ModelBuilder.java
```diff
@@ -978,6 +977,14 @@ public void cv_mainModelScores(int N, ModelMetrics.MetricBuilder mbs[], ModelBui
     // Now, the main model is complete (has cv metrics)
     DKV.put(mainModel);
   }
+  
+  public void cv_reduceCVBuilders(ModelMetrics.MetricBuilder mbsMain, ModelMetrics.MetricBuilder mbsReduced){
```
michalkurka, line 981: Do you need this method? It just delegates to "reduceForCV"? Instead of calling this on line https://github.com/h2oai/h2o-3/pull/5961/files#diff-892e6d71b6905c34c9126b46bf544491e10b6c6394a83b58de5a2ead0eca6ca7R908 - can you directly call reduceForCV? It seems a bit more cleaner to make the method "reduceForCV" part of the interface of MetricBuilder rather than ModelBuilder.

h2o-core/src/main/java/hex/ModelMetricsClustering.java
```diff
@@ -134,35 +129,58 @@ public void reduce(MetricBuilderClustering mm) {
...
       assert m instanceof ClusteringModel;
-      ClusteringModel clm = (ClusteringModel) m;
       ModelMetricsClustering mm = new ModelMetricsClustering(m, f, _customMetric);
+      setOverallStatToModelMetrics((ClusteringModel) m, f.numRows(), mm);
+      setCentroidsStatToModelMetrics(mm);
+      return m.addMetrics(mm);
+    }
 
-      mm._size = _size;
+    public ModelMetrics makeModelMetricsForCV(Model m, Frame f) {
```
michalkurka, line 152: it seems like this method is only needed because the MetricBuilder (mbs[0]) has essentially incorrect information (_size and _within_sumsqe are calculated for one of the models and later other partial information is reduced into it)

perhaps one way to simplify the whole flow and make sure we don't have inconsistent objects is to

define a method on MetricBuilder that would handle the whole aggregation for the per-model metrics, that is create a method that implements this loop:
```    
for (int i = 0; i < N; ++i) {
      if (i > 0) mbs[0].reduceForCV(mbs[i]);
mbs[0]._size = null
mbs[0]._within_sumsqe = null
return mbs[0]
```

then you don't need to differentiate between makeModelMetrics and makeModelMetricsForCV and can just have a one version (that is able to handle the nulls)

maurever: > it seems like this method is only needed because the MetricBuilder (mbs[0]) has essentially incorrect information (_size and _within_sumsqe are calculated for one of the models and later other partial information is reduced into it)

No, the reason is, that the arrays could have a different size because of `estimate_k=True` parameter. But you are correct, that this method could be merge something like that (because mbs[0]._size and mbs[0[._within_sumsqe are null from initialization):


```
    @Override
    public ModelMetrics makeModelMetrics(Model m, Frame f) {
      assert m instanceof ClusteringModel;
      ClusteringModel clm = (ClusteringModel) m;
      ModelMetricsClustering mm = new ModelMetricsClustering(m, f, _customMetric);
      setOverallStatToModelMetrics((ClusteringModel) m, f.numRows(), mm);
      if(_size != null && _within_sumsqe != null) {
           setCentroidsStatToModelMetrics(mm);
      }
      return m.addMetrics(mm);
    }
```

h2o-core/src/main/java/hex/ModelBuilder.java
maurever: It is overridden in ClusteringModelBuilder.java. The implement reduceForCV in MetricBuilder is possible, but this way made more sense to me. Thanks for the suggestion.

michalkurka: yes, I know it is overriding but it is only doing that to delegate to some other method - which is what I do not understand

if it was definited on metric builder there would be no delegating necessary

maurever:
@michalkurka, I implemented your suggestions here (https://github.com/h2oai/h2o-3/pull/5961/commits/844323a1405eaf575c2ed57e7d1002131b9f1121), but I am not sure if it is what you really want. :)

michalkurka (approved):
Thank you @maurever 

Do we need to also make a documentation change?

maurever:
Good point @michalkurka.

@hannah-tillman, could you review this commit, please? https://github.com/h2oai/h2o-3/pull/5961/commits/14db3bc00430ec0537600b89f0886a225e6ee070 Thank you!

hannah-tillman (approved):
just minor updates -- LGTM :)

maurever:
Thanks @michalkurka  and @hannah-tillman!

PR #5963, opened 2021-12-16
HEXDEV-0779: Dummy replacement for JndiLookup class affected by zero-day exploits 

michalkurka:
One of Log4j's suggested mitigations of recent vulnerabilities is to delete JndiLookup class from
affected jars, from https://logging.apache.org/log4j/2.x/security.html#CVE-2021-44228:

> Log4j 2.x mitigation: Implement one of the mitigation techniques below.
> ...
> - Otherwise, remove the JndiLookup class from the classpath: zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class

H2O produces error messages when this class is removed, to avoid such error
messages we suggest putting a dummy version of the JndiLookup class in place
of the original class. This class will print out the build is not affected
when it is being loaded (in our case that is alway on H2O startup).

This is verification to the users that the build is safe to use.


michalkurka:
@mn-mikke this is not meant to be a solution for current versions (those are patched by upgrading log4j) - this jar is meant for old versions and for users who chose to implement mitigation of removing `org/apache/logging/log4j/core/lookup/JndiLookup.class` class from the jar

mn-mikke (approved):
I see, thank you!

PR #5964, opened 2021-12-16
ITA-1003 - use most recent image, workaround for image version 40 is no longer needed

valenad1:
https://h2oai.atlassian.net/browse/ITA-1003


michalkurka (approved):
Thank you for the clean up!

PR #5972, opened 2021-12-20
PUBDEV-8477: EC2 Instances Update

hannah-tillman:
For: [PUBDEV-8477](https://h2oai.atlassian.net/browse/PUBDEV-8477)

Since the [AWS Standalone Instance](https://docs.h2o.ai/h2o/latest-stable/h2o-docs/cloud-integration/ec2-and-s3.html#aws-standalone-instance) and [Minio Instance](https://docs.h2o.ai/h2o/latest-stable/h2o-docs/cloud-integration/ec2-and-s3.html#minio-instance) both follow the **s3://** format, I removed reference to the `-hdfs_config` flag from both of them.


michalkurka (changes requested):
There is no change in the hdfs_config argument

the issue pointed out in jira is that S3 is not actually configured using this parameter, that only affects S3A/S3N

PR #5982, opened 2021-12-23
demo_rulefit_boston [nocheck]

koniecsveta:
Rulefit demo that could be linked in the release blogpost. This is better known and it shows that it mimics drf results. Apart of that it is better extendable by the other features that are on the way, because it will be possible to show local/global importances / interactions on it in the same way as it is done in paper.

### DIFF
--- a/h2o-py/demos/rulefit-boston-housing-demo.ipynb
+++ b/h2o-py/demos/rulefit-boston-housing-demo.ipynb
@@ -493,12 +493,12 @@
     }
    ],
    "source": [
-    "    from h2o.estimators.rulefit import H2ORuleFitEstimator\n",
-    "    rf = H2ORuleFitEstimator(seed=123, lambda_=.5 )#,  model_type=\"rules\")\n",
-    "    rf.train(y=y, x=x, training_frame=train)\n",
-    "    print(\"RMSE: \")\n",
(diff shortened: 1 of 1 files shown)


maurever (approved):
LGTM, thanks @koniecsveta. 

I would maybe add some plot of importances after the Rulefit algorithm is applied, to compare it with the DRF.

maurever:
Please, fix the demo for Python 2.7. ;) It looks like a small issue. 

```
...
  File "<string>", line 14
    from h2o.estimators.rulefit import H2ORuleFitEstimator
    ^
IndentationError: unexpected indent
```

koniecsveta:
The Boston Housing dataset includes a variable, `B`, which makes it sensitive regarding Fairness and Ethics in general.

This dataset has been used and then removed by several ML OSS libraries:
- [sklearn](https://github.com/scikit-learn/scikit-learn/issues/16155)
- [facet, from BCG-Gamma](https://github.com/BCG-Gamma/facet/issues/202)
- [causalnex](https://github.com/quantumblacklabs/causalnex/issues/91)

To avoid unnecessary negative feedbacks, the dataset should be changed. sklearn (see [PR](https://github.com/RGF-team/rgf/pull/357)) and facet has switched to the diabetes dataset

michalkurka (approved):
Thank you!

PR #5985, opened 2021-12-27
S3A token refresh - refactor, move logic out of PersistHdfs [nocheck]

michalkurka:


h2o-persist-hdfs/src/main/java/water/persist/S3ATokenRefresherFactory.java
```diff
@@ -0,0 +1,79 @@
...
+            return false;
+        }
+
+        // Important make sure the cloud is locked in order to guarantee that the leader will distribute credentials
+        // to all nodes and don't do refresh only for itself (which can happen if cloud is not yet locked)
+        Paxos.lockCloud("S3A Token Refresh");
+
+        if (H2O.CLOUD.leader() != H2O.SELF) {
+            // we are not a leader node in a locked cloud, give up
+            return true;
```
koniecsveta, line 52: It's not clear to me - why this returns `true`? If I am not a leader node and I've locked the cloud, then it means that the leader haven't been here and the token is not generated yet, because the leader would have locked the cloud already, otherwise I would return false on L34. So I return `true` but I don't have the token. And in getFileSystem I could potentially break from `for` cycle and do `FileSystem.get(path.toUri(), CONF);` even though I don't have the token, isn't it?

michalkurka: The semantics of the method is probably not ideal: the result means - was the `path` handled or do we need to keep trying other "refresher factories"?

Line `34` returns false because we don't know yet if given `path` is an `S3A` path or something else. That is why it returns false - keep trying other refreshers.

On line `52` we already know we are dealing with `S3A` and we are not the leader. We are therefore saying the `path` was handled (by not doing anything because we are not leader).

> So I return true but I don't have the token. And in getFileSystem I could potentially break from for cycle and do FileSystem.get(path.toUri(), CONF); even though I don't have the token, isn't it?

Yes, because it is built on the assumption that the leader is handling everything and distributing the tokens. Every interaction with the filesystem starts on the leader node (that is the one called by the user), the leader then can submit tasks that do access the filesystem on non-leader nodes but these nodes already assume the token exists.

koniecsveta: I see now, thanks for explanation 👌

koniecsveta (approved):

h2o-persist-hdfs/src/main/java/water/persist/S3ATokenRefresherFactory.java
michalkurka: Thanks for the comment @koniecsveta - I added javadoc and explained the true/false at each exit point of the method

PR #5986, opened 2021-12-27
Fix HDP 3.1 public build [nocheck]

michalkurka:
Versioning of artifacts changed with introduction of Data Platform 3.1.0.0


valenad1 (approved):
It's working on my local. What about a test run on hadoop images? Wdyt?

PR #5987, opened 2021-12-28
PUBDEV-8468: Bypass Java 16+ limitation and invoke keytool as an external command

michalkurka:


valenad1 (approved):
LGTM, thanks for the fix 👍

PR #5989, opened 2021-12-30
Update explain.rst

gpablo6:
The python tab for AutoML Explain example had a duplicated url as in the following statement:

```
f = "https://h2o-public-test-data.s3.amazonaws.com/smalldata/wine/winequality-redwhite-no-BOM.csv"
df = h2o.import_file("https://h2o-public-test-data.s3.amazonaws.com/smalldata/wine/winequality-redwhite-no-BOM.csv")
```

Which left the `f` variable unutilized. Thus I modify it so it is passed down as an argument to `h2o.import_file()`


michalkurka (approved):
Thank you

PR #5991, opened 2022-01-03
PUBDEV-8481: Fix upload MOJO in Python

michalkurka:
This PR only fixes the Python side

It also doesn't address issue https://h2oai.atlassian.net/browse/PUBDEV-8489 (that will be its own PR)

### DIFF
--- a/h2o-py/tests/testdir_generic_model/pyunit_mojo_import.py
+++ b/h2o-py/tests/testdir_generic_model/pyunit_mojo_import.py
@@ -5,8 +5,9 @@ from tests import pyunit_utils
 import os
 from pandas.testing import assert_frame_equal
 
+
 # Test of MOJO convenience methods
-def mojo_conveniece():    
+def mojo_convenience():    
     # Train a model
     airlines = h2o.import_file(path=pyunit_utils.locate("smalldata/testng/airlines_train.csv"))
     model = H2OGradientBoostingEstimator(ntrees = 1)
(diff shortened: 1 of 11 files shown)


michalkurka:
@maurever just fiy at this point, still WIP

h2o-py/tests/testdir_generic_model/pyunit_mojo_import.py
```diff
@@ -29,16 +29,20 @@ def mojo_conveniece():
...
+        # Load the model from the temporary file
+        mojo_model = h2o.upload_mojo(original_model_filename)
+        assert isinstance(mojo_model, H2OGenericEstimator)
+    
+        # Test scoring is available on the model
+        predictions = mojo_model.predict(airlines)
+        assert predictions is not None
+        assert predictions.nrows == 24421
+    finally:
+        pyunit_utils.set_forbidden_paths([])
```
maurever, line 45: First, it was not clear to me, that setting an empty list will clear the sandbox. It is a little bit confusing for me. IMO, some documentation would be useful, or splitting this function into two would be a more clear solution. 

Otherwise it looks good to me.

michalkurka: Thank you for the feedback, I fixed it - please take another look

maurever (approved):
Thanks, @michalkurka, for this fix!

PR #5992, opened 2022-01-05
ITA-1005: Fold SecurityUtils test stage back into JUnit tests stage

michalkurka:
It is not needed because it is passing right now, it was only separated because it was not working on Java 16


valenad1 (approved):
👍

PR #5993, opened 2022-01-06
PUBDEV-8488: Add check before producing a bad Mojo

tomasfryda:
https://h2oai.atlassian.net/browse/PUBDEV-8488


sebhrusen (commented):
cool! will work in combination with TE detection in https://h2oai.atlassian.net/browse/PUBDEV-8504

sebhrusen (approved):

PR #5994, opened 2022-01-06
PUBDEV-8489: Use original model ID for Imported MOJO model [POC]

michalkurka:


michalkurka:
@sebhrusen @maurever this is a POC, I would like to get your opinion on this change and also if you think it is worth it to modify the backend in this way. The main issue is that I needed to make Job#result mutable, this violates prior assumptions.

h2o-core/src/main/java/water/Job.java
```diff
@@ -60,6 +61,19 @@ public void setWarnings(final String [] warns){
     }.apply(this);
   }
 
+  public void setResult(final Key<T> result) {
+    String resultClassName = TypeMap.className(_typeid);
+    if (!"hex.generic.GenericModel".equals(resultClassName)) {
+      throw new UnsupportedOperationException("Result key cannot be modified for " + 
+              (_typeid == 0 ? "unknown result type." : "result of type " + resultClassName + "."));
+    }
```
sebhrusen, line 69: It's great to have a safeguard here, but as `_result` is public and not final anymore, the protection is relative.

as search of `_result` read-access shows 52 results, not as large I'd have expected, so… what about making `_result` private and expose a public getter instead? Backward compatibility concerns? Would be safer though.

sebhrusen (commented):
> The main issue is that I needed to make Job#result mutable, this violates prior assumptions.

I agree that it's a serious change, otherwise I'd have no pb with it: see suggestion below.

h2o-algos/src/main/java/hex/generic/Generic.java
```diff
@@ -138,6 +143,16 @@ private GenericModel importPojo(ByteVec pojoBytes, Key<Frame> pojoKey) throws IO
...
     }
 
+    private void setResultKey(Key<GenericModel> newResultKey) {
+        _job.setResult(newResultKey);
+        _result = _job._result; // careful - we have to make sure _result == _job._result (not just object equals)
+    }
+
+    private boolean isDefaultResultKey() {
+        final String genericPrefix = _parms.algoName() + "_";
+        return _result.toString().startsWith(genericPrefix);
```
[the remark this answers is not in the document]
michalkurka: yes, this is a best-effort approach - far from ideal

`_result` is set from ModelHandler, a flag would be better (or even better "future key") but it is an extensive change in API

I will think about it some more.

h2o-core/src/main/java/water/Job.java
michalkurka: Another approach would be to keep the final modifier and just 
- replace the whole job instance
- use reflection to mutate the final field

I would prefer to keep the typical case the same, don't affect the API - for this very specific need. Maybe I need a new different Job implementation for that.

sebhrusen: Sure `Key` class is currently `final` but it doesn't appear as a necessity

PR #6000, opened 2022-01-10
Tests: Fix Target Encoder scripts - not running in single node [nocheck]

michalkurka:
- Tests: Fix Target Encoder scripts - not running in single node
- Makefile clean-up: junit-1x tests are actually only running junit-11


sebhrusen (commented):
Congrats for 6kth PR: you won a review.

This is just cleanup, right? Not intended to fix the "TE tests not running" issue.

michalkurka:
>Congrats for 6kth PR: you won a review.

I didn't even notice! Thanks!

> This is just cleanup, right? Not intended to fix the "TE tests not running" issue.

It should fix it, the grep was wrong.

sebhrusen (approved):
ok, was it applying `grep -v` on empty string? this should never return anything

sebhrusen:
Ah right, this are the TE scripts!!! For some reason I thought these were generic Jenkins scripts, so I didn't understand why it failed only for TE

PR #6003, opened 2022-01-11
PUBDEV-8386: Customize FormAuthenticator to use relative redirects

michalkurka:
- Copy Jetty's FormAuthenticator into our code base, original code unmodified
- PUBDEV-8386: Customize FormAuthenticator to use relative redirects


mn-mikke (approved):
LGTM, What about Jetty 9? Are there any problems with relative redirects?

michalkurka:
@mn-mikke Jetty 9 is not affected; the original fix on Jetty 9 was verified to be working with a customer.

thank you for your review

PR #6006, opened 2022-01-13
PUBDEV-8513: Fix import of Parquet files from S3

michalkurka:
The issue that caused PUBDEV-8513 was that H2O was falling back to HDFS "s3a"-style import for reading parquet files footers, this worked fine as long as "s3a" was also configured. However, didn't work for community users.

### DIFF
--- a/h2o-persist-s3/src/test/java/water/persist/PersistS3Test.java
+++ b/h2o-persist-s3/src/test/java/water/persist/PersistS3Test.java
@@ -3,6 +3,8 @@ package water.persist;
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.model.AmazonS3Exception;
+import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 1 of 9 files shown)


h2o-persist-s3/src/test/java/water/persist/PersistS3Test.java
```diff
@@ -470,5 +475,68 @@ public void testCustomCredentialsProvider_ignoreInvalid() {
...
+      Scope.track(localVec);
+
+      assertEquals(localVec.length(), fromS3.anyVec().length());
+    } finally {
+      Scope.exit();
+    }
+  }
+
+  @Test
+  public void testParseParquetFromS3() throws IOException {
```
koniecsveta, line 511: locally this test is failing for me, looks like it still guesses CSV:
```junit.framework.ComparisonFailure: 
Expected :PARQUET
Actual   :CSV```

michalkurka: I think you will need to refresh Idea dependencies in your intellij - most likely it doesn't see Parquet parser (this was added as a test dependency in this PR)

Also, it expects the credentials in environment properties - similarly to other tests.

koniecsveta: that was it 👌

koniecsveta (approved):

